PR #15, opened 2014-11-03
The Road to Ember 2.0 RFC

tomdale:
- Start Date: 2014-12-03
- RFC PR: (leave this empty)
- Ember Issue: (leave this empty)

# The Road to Ember 2.0

## Intro

Today, we're announcing our plan for Ember 2.0. While the major version bump gives us the opportunity to simplify the framework in ways that require breaking changes, we are designing Ember 2.0 with migration in mind.

This is **not** a big-bang rewrite; we will continue development on the `master` branch, and roll out changes incrementally on the 1.x release train. The 2.0.0 release will simply remove features that have been deprecated between now and then. Our goal is that you can move your Ember app to 2.0 incrementally, one sprint at a time.

This RFC captures the results of the last two core team face-to-face meetings, where we discussed community feedback about the future of the project. While it explains the high-level goals and tries to paint a picture of how all the pieces fit together, this document will be updated over time with links to individual RFCs that contain additional implementation detail.

We plan to flesh out these more-detailed RFCs in the next few weeks, as the discussion here progresses, before finalizing this plan.

We are announcing Ember 2.0 through our community RFC process in advance of a release, both so our proposals can be vetted by the community and so the community can understand the goals and contribute their own ideas back.

### Stability without Stagnation

Ember is all about identifying common patterns that emerge from the web development community and rolling them into a complete front-end stack.  This makes it easy to get started on new projects and jump into existing ones, knowing that you will get a best-of-breed set of tools that the community will continue to support and improve for years to come.

In the greater JavaScript community, getting the latest and greatest often means rewriting parts of your apps once a year, as the community abandons existing solutions in search of improvements. Progress is important, but so is ending the constant cycle of writing and rewriting that plagues so many applications.

The Ember community works hard to introduce new ideas with an eye towards migration. We call this "stability without stagnation", and it's one of the cornerstones of the Ember philosophy.

Below, we introduce some of the major new features coming in Ember 2.0. Each section includes a transition plan, with details on how we expect existing apps to migrate to the new API.

When breaking changes are absolutely necessary, we try to make those changes ones you can apply without too much thought. We call these "mechanical" refactors. Typically, they'll involve a change to syntax without changing semantics. These are significantly easier to adopt than those that require fundamental changes to your application architecture.

**To further aid in these transitions, we are planning to add a new tab to the Ember Inspector that will list all deprecations in your application**, as well as a list of the locations in the source code where the deprecated code was triggered. This should serve as a convenient "punch list" for your transitional work.

Every member of the core team works on up-to-date Ember applications, and we feel the tension between stability and progress acutely. We want to deliver cutting-edge products, but need to keep shipping, and many companies that have adopted Ember for their products tell us the same thing.

### Big Bets

In 2014, we made big bets in two areas, and they've paid off.

The first bet was on open standards: JavaScript modules, promises and Web Components. We started the year off with globals-based apps, callbacks and "views", and incrementally (and compatibly) built towards standards-based solutions as those standards solidified.

The second bet was that the community was as tired as we were of hand-rolling their own build scripts for each project. We've invested heavily in Ember CLI, giving us a single tool that unifies the community and provides a venue for disseminating great ideas.

**In Ember 2.0, Ember CLI and ES6 modules will become first-class parts of the Ember experience.** We will update the website, guides, documentation, etc. to teach new users how to build Ember apps with the CLI tools and using JavaScript's new module syntax.

While globals-based apps will continue to work in 2.0, we may introduce new features that rely on either Ember CLI or ES6 modules. **You should begin moving your app to Ember CLI as soon as possible.**

All of the apps maintained by the Ember core team have been migrated to Ember CLI, and we believe that most teams should be able to make the transition incrementally.

### Learning from the Community

We're well aware that we don't have a monopoly on good ideas, and we're always analyzing competing frameworks and libraries to discover great ideas that we can incorporate.

For example, AngularJS taught us the importance of making early on-ramp easy, how cool directives/components could be, and how dependency injection improves testing.

We've been analyzing and discussing React's approach to data flow and rendering for some time now, and in particular how they make use of a "virtual DOM" to improve performance.

Ember's view layer is one of the oldest parts of Ember, and was designed for a world where IE7 and IE8 were dominant. We've spent the better part of 2014 rethinking the view layer to be more DOM-aware, and the new codebase (codenamed "HTMLBars") borrows what we think are the best ideas from React. We cover the specifics below.

React's "virtual DOM" abstraction also allowed them to simplify the programming model of component-based applications. We really like these ideas, and the new HTMLBars engine, landing in the next Ember release, lays the groundwork for adopting the simplified data-flow model.

**In Ember 2.0, we will be adopting a "virtual DOM" and data flow model that embraces the best ideas from React and simplifies communication between components.**

Interestingly, we found that well-written Ember applications are already written with this clear and direct data flow. This change will mostly make the best patterns more explicit and easier for developers to find when starting out.

### A Steady Flow of Improvement

Ember 1.0 shipped over a year ago and we have continued to improve the framework while maintaining backwards-compatibility. We are proud of the fact that Ember apps tend to track released versions.

You might expect us to do Ember 2.0 work on a separate "2.0" branch, accumulating features until we ship. We aren't going to do that.

Instead, **we plan to do the vast majority of new work on `master` (behind feature flags), and land new features in 1.x as they become stable.**

The `2.0.0` release **will simply remove the cruft** that naturally builds up when maintaining compatibility with old releases.

If we add features that change Ember idioms, we will add clear deprecation warnings with steps to refactor to new patterns.

Our goal is that, as much as possible, people will be able to boot up their app on the last `1.x` version, update to the latest set of idioms by following the deprecation prompts, and have things working on `2.0`.

Because going from the last version of Ember 1.x to Ember 2.0 will be just another six-week release, there simply won't be much time for us to make it an incredibly painful upgrade. ;)

## Simplifying Ember Concepts

Ember evolved organically from a view-layer-only framework in 2011 into the route-driven, complete front-end stack it is today. Along the way, we've accumulated several concepts that are no longer widely used in idiomatic Ember apps.

These vestigial concepts make file sizes larger, code more complex, and make Ember harder to learn.

**Ember 2.0 is about simplification**. This lets us reduce file sizes, reduce code complexity, and generally make Ember apps easier to pick up and maintain.

The high-level set of improvements that we have planned are:

* More intuitive attribute bindings
* New HTML syntax for components
* Block parameters for components
* More consistent template scope
* One-way data binding by default, with opt-in to mutable, two-way bindings
* More explicit communication between components, which means less implicit communication via two-way bindings
* Routes drive components, instead of controller + template
* Improved actions that are invoked inside components as simple callbacks

In some sections, we provide estimates for when a feature will land.  These are our best-guesses, but because of the rapid-release train model of Ember, we may be off by a version or two.

However, all features that are slated for "before 2.0" will land before we cut over to a major new version.

## More Intuitive Attribute Bindings

Today's templating engine is the oldest part of Ember.js. Under the hood, it generates a string of HTML and then inserts it into the page.

One unfortunate consequence of this architecture is that it is not possible to intuitively bind values to HTML attributes.

You would expect to be able type something like:

```handlebars
<a href="{{url}}">Click here</a>
```

But instead, in today's Ember, you have to learn about and use the `bind-attr` helper:

```handlebars
<a {{bind-attr href=url}}>Click here</a>
```

The new HTMLBars template engine makes `bind-attr` a thing of the past, allowing you to type what you mean. It also makes it possible to express many attribute-related concepts simply:

```handlebars
<a class="{{active}} app-link" href="{{url}}.html">Click here</a>
```

### Transition Plan

The HTMLBars templating engine is being developed on master, and parts of it have already landed in Ember 1.8. Doing the work this way means that the new engine continues to support the old syntax: your existing templates will continue to work.

The improved attribute syntax has not yet landed, but we expect it to land before Ember 1.10.

**We do not plan to remove support for existing templating syntax (or no-longer-necessary helpers like `bind-attr`) in Ember 2.0.**

## More Intuitive Components

In today's Ember, components are represented in your templates as Handlebars "block helpers".

The most important problem with this approach is that Handlebars-style components do not work well with attribute bindings or the `action` helper. In short, a helper that is meant to be used inside an HTML tag cannot be used inside a call to a component.

Beginning in Ember 1.11, we will support an HTML-based syntax for components. **The new syntax can be used to invoke existing components, and new components can be called using the old syntax.**

```handlebars
<my-video src={{movie.url}}></my-video>

<!-- equivalent to -->

{{my-video src=movie.url}}
```

### Transition Plan

The improved component syntax will (we hope) land in Ember 1.11. You can transition existing uses of `{{component-name}}` to the new syntax at that time. You will likely benefit by eliminating uses of computed properties that can now be more tersely expressed using the interpolation syntax.

**We have no plans to remove support for the old component syntax in Ember 2.0.**

## Block Parameters

In today's templates, there are two special forms of built-in Handlebars helpers: `#each post in posts` and `#with post as p`. These allow the template inside the helper to retain the parent context, but get a piece of helper-provided information as a named value (such as `post` in the previous examples).

```
{{#with contact.person as p}}
  {{!-- this block of code is still in the parent's scope, but
        the #with helper provided a `p` name with a
        helper-provided value --}}
  <p>{{p.firstName}} {{p.lastName}}</p>

  {{!-- `title` here refers to the outer scope's title --}}
  <p>{{title}}</p>
{{/with}}
```

Today, this capability is hardcoded into the two special forms, but it can be useful for other kinds of components. For example, you may have a calendar component (`ui-calendar`) that displays a specified month.

The `ui-calendar` component may want to allow users to supply a custom template for each day in the month, but each repetition of the template will need information about the day it represents (its day of the week, date number, etc.) in order to render it.

With the new "block parameters" feature, any component will have access to the same capability as `#each` or `#with`:

```handlebars
<ui-calendar month={{currentMonth}} as |day|>
  <p class="title">{{day.title}}</p>
  <p class="date">{{day.date}}</p>
</ui-calendar>
```

In this case, the `ui-calendar` component iterates over all of days in `currentMonth`, rendering each instance of the template with information about which date it should represent.

We also think that this feature will be useful to allow container components (like tabs or forms) to supply special-case component definitions as block params. We are still working on the details, but believe that an approach along these lines could make these kinds of components simpler and more flexible.

### Transition Plan

Block parameters will hopefully land in 1.12, and at that point the two special forms for `{{each}}` and `{{with}}` will be deprecated.  You should refactor your templates to use the new block parameters syntax once it lands, as it is a purely mechanical refactor.

**We have no plans to remove support for the `{{each}}` and `{{with}}` special forms in Ember 2.0.**

## More Consistent Handlebars Scope

In today's Ember, the `each` and `with` helpers come in two flavors: a "context-switching" flavor and a "named-parameter" flavor.

```handlebars
{{#each post in posts}}
  {{!-- the context in here is the same as the outside context,
        and `post` references the current iteration --}}
{{/each}}

{{#each posts}}
  {{!-- the context in here has shifted to the individual post.
        the outer context is no longer accessible --}}
{{/each}}
```

This has proven to be one of the more confusing parts of the Ember templating system. It is also not clear to beginners which to use, and when they choose the context-shifting form, they lose access to values in the outer context that may be important.

Because the helper itself offers no clue about the context-shifting behavior, it is easy (even for more experienced Ember developers) to get confused when skimming a template about which object a value refers to.

In Ember 1.10, we will deprecate the context-shifting forms of `#each` and `#with` in favor of the named-parameter forms.

### Transition Plan

To transition your code to the new syntax, you can change templates that look like this:

```hbs
{{#each people}}
  <p>{{firstName}} {{lastName}}</p>
  <p>{{address}}</p>
{{/each}}
```

with:

```hbs
{{#each people as |person|}}
  <p>{{person.firstName}} {{person.lastName}}</p>
  <p>{{person.address}}</p>
{{/each}}
```

**We plan to deprecate support for the context-shifting helpers in Ember 1.10 and remove support in Ember 2.0.** This change should be entirely
mechanical.

## One-Way Bindings by Default

After a few years of having written Ember applications, we have observed that most of the data bindings in the templating engine do not actually require two-way bindings.

When we designed the original templating layer, we figured that making all data bindings two-way wasn't very harmful: if you don't set a two-way binding, it's a de facto one-way binding!

We have since realized (with some help from our friends at React), that components want to be able to hand out data to their children without having to be on guard for wayward mutations.

Additionally, communication between components is often most naturally expressed as events or callbacks. This is possible in Ember, but the dominance of two-way data bindings often leads people down a path of using two-way bindings as a communication channel. Experienced Ember developers don't (usually) make this mistake, but it's an easy one to make.

When you use the new component syntax, the `{{}}` interpolation syntax defaults to creating one-way bindings in the components.

```handlebars
<my-video src={{url}}></my-video>
```

In this example, the component's `src` property will be updated whenever `url` changes, but it will not be allowed to mutate it.

If a template wishes to allow the component to mutate a property, it can explicitly create a two-way binding using the `mut` helper:

```handlebars
<my-video paused={{mut isPaused}}></my-video>
```

This can help ease the transition to a more event-based style of programming.

It also eliminates the boilerplate associated with an event-based style when working with form controls. Instead of copying state out of a model, listening for callbacks, and updating the model, the `input` helper can be given an explicit mutable binding.

```handlebars
<input value={{mut firstName}}>
<input value={{mut lastName}}>
```

This is similar to the approach taken by [React.Link][1], but we think that the use-case of form helpers is sufficiently common to make it ergonomic.

[1]: http://facebook.github.io/react/docs/two-way-binding-helpers.html

### Transition Plan

The new one-way default is triggered by the use of new component syntax.  This means that component invocations in existing templates will continue to work without changes.

When transitioning to the new HTML-based syntax, you will likely want to evaluate whether bindings are actually being mutated, and avoid using `mut` for values that the component never changes. This will make it easier for future readers of your template to get an understanding of what properties might be changed downstream.

To preserve the same semantics during a refactor to the new HTML-based syntax, you can simply mark all bindings as `mut`.

```handlebars
{{!-- these are semantically equivalent --}}

{{my-video src=movie.url paused=controller.isPaused}}

<my-video src={{mut movie.url}} paused={{mut controller.isPaused}}>
</my-video>
```

While the above example preserves the same mutability semantics, it should be clear that the video player component should never change the `url` of the `movie` model.

To make sure you get an exception should this ever happen, simply remove the `mut`:

```handlebars
<my-video src={{movie.url}} paused={{mut controller.isPaused}}>
</my-video>
```

**We have no plans to remove the old-style component syntax in Ember 2.0, so the semantics of existing component invocations will not change.**

## Separated Component Parameters

In today's Ember, parameters passed to components as attributes become properties of the component itself, putting them in the same place as other internal state.

This can be somewhat confusing, because it may not be obvious to the reader of a component's JavaScript or template which values are internal, and which are passed in as part of the public API.

To remind themselves, many Ember users write their components like this:

```js
  /* Public API */

  src: null,
  paused: null,
  title: null,

  /* Internal */
  scrubber: null
})
```

It can also be unclear how to react to a change in the external properties. It is possible to use observers for this purpose in Ember, but observers feel low-level and do not coordinate very well with the rendering process.

To reduce confusion, we plan to move external attributes into a new `attrs` hash.

If you invoke a component like this:

```handlebars
<my-video src={{movie.url}}></my-video>
```

then the `my-video` component accesses the passed-in `src` attribute as `this.attrs.src`.

We also plan to provide lifecycle callbacks (modelled after [React's lifecycle callbacks][react-lifecycle]) for changes to `attrs` that will integrate with the rendering lifecycle. We plan to supplement the API with callbacks for changes in individual properties as well.

[react-lifecycle]: http://facebook.github.io/react/docs/component-specs.html#lifecycle-methods

### Transition Plan

In Ember 1.10, we will begin installing provided attributes in the component's `attrs` hash. If a provided attribute is accessed directly on the component, a deprecation warning will be issued.

In applications, you should update your component JavaScript and templates to access provided attributes via the component's `attrs` property.

**In Ember 2.0, we will stop setting attributes as properties on the component itself.**

We will also provide a transitional mixin that Ember addons can use that will make provided attributes available as `attrs.*`. This will allow add-ons to move to the new location, while maintaining support for older versions of Ember. We expect people to upgrade to Ember 1.10 relatively quickly, and do not expect addons to need to maintain support for Ember 1.9 indefinitely.

## Routeable Components

Many people have noticed that controllers in Ember look a lot like components, but with an arbitrary division of responsibilities. We agree!

In current versions of Ember, when a route is entered, it builds a controller, associates a model with it, and hands it off to an (old-style) view for rendering. The view itself is invisible; you just write a template with the correct name.

We plan to transition to: when a route is entered, it renders a **component**, passing along the model as an `attr`. This eliminates a vestigial use of old-style views, and associates the top-level template with a regular component.

### Transition Plan

Initially, we will continue to support routing to a controller+template, so nothing will break. Going forward, routes will route to a component instead.

In order to do that refactoring, several things will change:

* Instead of referring to model properties directly (or on `this`), you will refer to them as `model.propName`.
* Similarly, computed properties that move to your component will need to depend on `model.propName` if they are migrated from an `ObjectController`.
* In both cases, the short version is that you can no longer rely on the proxying behavior of `ObjectController` or `ArrayController`, but you can remedy the situation by prefixing `model.` to the property name.
* Unlike controllers, top-level components do not persist across navigation. Persistent state should be stored in route objects and passed as initial properties to routable components.
* In addition to the asynchronous `model` hook in routes, routes will also be able to define a `attrs` hook, which can return additional asynchronous data that should be provided to the component.
* Routeable Components should be placed in a "pod" naming convention. For example, the component for the `blog-post` route would be `app/blog-post/component.js`.

**We plan to land support for routeable components in Ember 1.12, and deprecate routeable controllers at the same time. We plan to remove support for routeable controllers in Ember 2.0.** This will allow you to move your codebases over to routeable components piecemeal before making the jump to 2.0.

**We will also provide an optional plugin for Ember 2.0 apps that restores existing behavior.** This plugin will be included in the Ember automated test suite to ensure that we do not introduce accidental regressions in future releases on the 2.x series.

We realize that this is the change has the largest transitional cost of all the planned features, and we plan to dedicate time to the precise details in the full RFC on this topic.

## Improving Actions

Today's components can communicate with their parent component through actions. In particular, the `sendAction` method allows a child component to invoke a named action on the parent (inside of the `actions` hash).

Part of the reason for this API was a limitation in the original Handlebars syntax:

```handlebars
{{!-- we can't get too fancy with the value of key-press --}}
{{input key-press="valueChanged"}}
```

In this example, when the `input` component calls `this.sendAction('key-press')`, it invokes the `valueChanged` action on its parent component.

With the new HTML syntax for components, we have more flexibility:

```handlebars
<input key-press={{action "valueChanged"}}>
```

This will package up the parent's `valueChanged` action (in the `actions` hash) as a callback function that is available to the child component as `this.attrs['key-press']`.

```js
  keypress: function(event) {
    this.attrs['key-press'](event.target.value);
  }
});
```

The benefit of this approach is twofold:

* Actions are no longer treated specially in the component API. They are simply properties packaged up to be called by the child component.
* It is possible to pass an alternative function as the `key-press`, reducing the child component's knowledge of what the callback is doing. This has testing and abstraction benefits.

### Transition Plan

We will continue to support the `sendAction` API for the forseeable future in today's Handlebars syntax.

When calling an existing component with new HTMLBars syntax, you do not need to change your existing `actions` hash. You should change syntax that looks like this:

```handlebars
{{video-player playing="playingBegins"}}
```

To this:

```handlebars
<video-player playing={{action "playingBegins"}}>
```

The `video-player` component's internal use of `sendAction` will work with both calling styles.

New components should use `this.attrs.playing()`, but existing components that want to continue supporting legacy callers should continue to use `sendAction` for now. The `sendAction` API will seamlessly support both calling styles, and will be supported for the forseeable future.

```js
// instead of
this.sendAction('progress', value);

// new code can use
this.attrs.progress(value);
```

## Onward

Version 2.0 marks the transformation of Ember from simply an MVC framework to a complete front-end stack. Between Ember's best-in-class router, revamped components with virtual DOM, easy-to-use build tools, and a growing ecosystem that makes taking advantage of additional libraries a breeze, there's no better way to get started and stay productive developing web apps today.

Hopefully, this plan demonstrates that staying on the cutting-edge can be done without rewriting your app. There are a huge number of Ember apps in production today, and we're looking forward to a time in the very near future where they can start to take advantage these new features.

Expect to see many more RFCs covering these features in depth soon (including a roadmap for Ember Data 1.0). We look forward to hearing your feedback!

### DIFF
--- a/active/0000-the-road-to-ember-2-0.md
+++ b/active/0000-the-road-to-ember-2-0.md
@@ -1,0 +1,703 @@
+- Start Date: 2014-12-03
+- RFC PR: (leave this empty)
+- Ember Issue: (leave this empty)
+
+# The Road to Ember 2.0
+
+## Intro
+
+Today, we're announcing our plan for Ember 2.0. While the major version
+bump gives us the opportunity to simplify the framework in ways that
+require breaking changes, we are designing Ember 2.0 with migration in mind.
+
+This is **not** a big-bang rewrite; we will continue development
+on the `master` branch, and roll out changes incrementally on the 1.x
+release train. The 2.0.0 release will simply remove features that have
+been deprecated between now and then. Our goal is that you can move
+your Ember app to 2.0 incrementally, one sprint at a time.
+
+This RFC captures the results of the last two core team face-to-face
+meetings, where we discussed community feedback about the future of the
+project. While it explains the high-level goals and tries to paint a
+picture of how all the pieces fit together, this document will be
+updated over time with links to individual RFCs that contain additional
+implementation detail.
+
+We plan to flesh out these more-detailed RFCs in the next few weeks,
+as the discussion here progresses, before finalizing this plan.
+
+We are announcing Ember 2.0 through our community RFC process in advance
+of a release, both so our proposals can be vetted by the community and
+so the community can understand the goals and contribute their own ideas
+back.
+
+## Motivation
+
+### Stability without Stagnation
+
+Ember is all about identifying common patterns that emerge from the web
+development community and rolling them into a complete front-end stack.
+This makes it easy to get started on new projects and jump into existing
+ones, knowing that you will get a best-of-breed set of tools that the
+community will continue to support and improve for years to come.
+
+In the greater JavaScript community, getting the latest and greatest
+often means rewriting parts of your apps once a year, as the community
+abandons existing solutions in search of improvements. Progress is
+important, but so is ending the constant cycle of writing and rewriting
+that plagues so many applications.
+
+The Ember community works hard to introduce new ideas with an eye
+towards migration. We call this "stability without stagnation", and it's
+one of the cornerstones of the Ember philosophy.
+
+Below, we introduce some of the major new features coming in Ember
+2.0. Each section includes a transition plan, with details on how
+we expect existing apps to migrate to the new API.
+
+When breaking changes are absolutely necessary, we try to make those
+changes ones you can apply without too much thought. We call these
+"mechanical" refactors. Typically, they'll involve a change to
+syntax without changing semantics. These are significantly easier to
+adopt than those that require fundamental changes to your application
+architecture.
+
+**To further aid in these transitions, we are planning to add a new tab to the
+Ember Inspector that will list all deprecations in your application**,
+as well as a list of the locations in the source code where the
+deprecated code was triggered. This should serve as a convenient
+"punch list" for your transitional work.
+
+Every member of the core team works on up-to-date Ember applications,
+and we feel the tension between stability and progress acutely. We want
+to deliver cutting-edge products, but need to keep shipping, and many
+companies that have adopted Ember for their products tell us the
+same thing.
+
+### Big Bets
+
+In 2014, we made big bets in two areas, and they've paid off.
+
+The first bet was on open standards: JavaScript modules, promises and
+Web Components. We started the year off with globals-based apps,
+callbacks and "views", and incrementally (and compatibly) built towards
+standards-based solutions as those standards solidified.
+
+The second bet was that the community was as tired as we were of
+hand-rolling their own build scripts for each project. We've invested
+heavily in Ember CLI, giving us a single tool that unifies the community
+and provides a venue for disseminating great ideas.
+
+**In Ember 2.0, Ember CLI and ES6 modules will become first-class parts
+of the Ember experience.** We will update the website, guides, documentation,
+etc. to teach new users how to build Ember apps with the CLI tools and
+using JavaScript's new module syntax.
+
+While globals-based apps will continue to work in 2.0, we may introduce
+new features that rely on either Ember CLI or ES6 modules. **You should
+begin moving your app to Ember CLI as soon as possible.**
+
+All of the apps maintained by the Ember core team have been migrated to
+Ember CLI, and we believe that most teams should be able to make the
+transition incrementally.
+
+### Learning from the Community
+
+We're well aware that we don't have a monopoly on good ideas, and we're
+always analyzing competing frameworks and libraries to discover great
+ideas that we can incorporate.
+
+For example, AngularJS taught us the importance of making early on-ramp
+easy, how cool directives/components could be, and how dependency
+injection improves testing.
+
+We've been analyzing and discussing React's approach to data flow and
+rendering for some time now, and in particular how they make use of a
+"virtual DOM" to improve performance.
+
+Ember's view layer is one of the oldest parts of Ember, and was designed
+for a world where IE7 and IE8 were dominant. We've spent the better part
+of 2014 rethinking the view layer to be more DOM-aware, and the new
+codebase (codenamed "HTMLBars") borrows what we think are the
+best ideas from React. We cover the specifics below.
+
+React's "virtual DOM" abstraction also allowed them to simplify the
+programming model of component-based applications. We really like these
+ideas, and the new HTMLBars engine, landing in the next Ember release, lays
+the groundwork for adopting the simplified data-flow model.
+
+**In Ember 2.0, we will be adopting a "virtual DOM" and data flow model
+that embraces the best ideas from React and simplifies communication
+between components.**
+
+Interestingly, we found that well-written Ember applications are already
+written with this clear and direct data flow. This change will mostly
+make the best patterns more explicit and easier for developers to find
+when starting out.
+
+### A Steady Flow of Improvement
+
+Ember 1.0 shipped over a year ago and we have continued to improve the
+framework while maintaining backwards-compatibility. We are proud of the
+fact that Ember apps tend to track released versions.
+
+You might expect us to do Ember 2.0 work on a separate "2.0" branch,
+accumulating features until we ship. We aren't going to do that.
+
+Instead, **we plan to do the vast majority of new work on `master` (behind
+feature flags), and land new features in 1.x as they become stable.**
+
+The `2.0.0` release **will simply remove the cruft** that naturally
+builds up when maintaining compatibility with old releases.
+
+If we add features that change Ember idioms, we will add clear
+deprecation warnings with steps to refactor to new patterns.
+
+Our goal is that, as much as possible, people will be able to boot up
+their app on the last `1.x` version, update to the latest set of idioms
+by following the deprecation prompts, and have things working on `2.0`.
+
+Because going from the last version of Ember 1.x to Ember 2.0 will be
+just another six-week release, there simply won't be much time for us to
+make it an incredibly painful upgrade. ;)
+
+## Simplifying Ember Concepts
+
+Ember evolved organically from a view-layer-only framework in 2011 into
+the route-driven, complete front-end stack it is today. Along the way,
+we've accumulated several concepts that are no longer widely used in idiomatic
+Ember apps.
+
+These vestigial concepts make file sizes larger, code more complex, and
+make Ember harder to learn.
+
+**Ember 2.0 is about simplification**. This lets us reduce file sizes,
+reduce code complexity, and generally make Ember apps easier to pick up
+and maintain.
+
+The high-level set of improvements that we have planned are:
+
+* More intuitive attribute bindings
+* New HTML syntax for components
+* Block parameters for components
+* More consistent template scope
+* One-way data binding by default, with opt-in to mutable, two-way bindings
+* More explicit communication between components, which means less
+  implicit communication via two-way bindings
+* Routes drive components, instead of controller + template
+* Improved actions that are invoked inside components as simple callbacks
+
+In some sections, we provide estimates for when a feature will land.
+These are our best-guesses, but because of the rapid-release train model
+of Ember, we may be off by a version or two.
+
+However, all features that are slated for "before 2.0" will land before
+we cut over to a major new version.
+
+## More Intuitive Attribute Bindings
+
+Today's templating engine is the oldest part of Ember.js. Under the
+hood, it generates a string of HTML and then inserts it into the page.
+
+One unfortunate consequence of this architecture is that it is not
+possible to intuitively bind values to HTML attributes.
+
+You would expect to be able type something like:
+
+```handlebars
+<a href="{{url}}">Click here</a>
+```
+
+But instead, in today's Ember, you have to learn about and use the
+`bind-attr` helper:
+
+```handlebars
+<a {{bind-attr href=url}}>Click here</a>
+```
+
+The new HTMLBars template engine makes `bind-attr` a thing of the past,
+allowing you to type what you mean. It also makes it possible to express
+many attribute-related concepts simply:
+
+```handlebars
+<a class="{{active}} app-link" href="{{url}}.html">Click here</a>
+```
+
+### Transition Plan
+
+The HTMLBars templating engine is being developed on master, and parts
+of it have already landed in Ember 1.8. Doing the work this way means
+that the new engine continues to support the old syntax: your existing
+templates will continue to work.
+
+The improved attribute syntax has not yet landed, but we expect it to
+land before Ember 1.10.
+
+**We do not plan to remove support for existing templating syntax (or
+no-longer-necessary helpers like `bind-attr`) in Ember 2.0.**
+
+## More Intuitive Components
+
+In today's Ember, components are represented in your templates as
+Handlebars "block helpers".
+
+The most important problem with this approach is that Handlebars-style
+components do not work well with attribute bindings or the `action`
+helper. In short, a helper that is meant to be used inside an HTML tag
+cannot be used inside a call to a component.
+
+Beginning in Ember 1.11, we will support an HTML-based syntax for
+components. **The new syntax can be used to invoke existing components,
+and new components can be called using the old syntax.**
+
+```handlebars
+<my-video src={{movie.url}}></my-video>
+
+<!-- equivalent to -->
+
+{{my-video src=movie.url}}
+```
+
+### Transition Plan
+
+The improved component syntax will (we hope) land in Ember 1.11. You can
+transition existing uses of `{{component-name}}` to the new syntax
+at that time. You will likely benefit by eliminating uses of computed
+properties that can now be more tersely expressed using the
+interpolation syntax.
+
+**We have no plans to remove support for the old component syntax in
+Ember 2.0.**
+
+## Block Parameters
+
+In today's templates, there are two special forms of built-in Handlebars
+helpers: `#each post in posts` and `#with post as p`. These allow the
+template inside the helper to retain the parent context, but get a piece
+of helper-provided information as a named value (such as `post` in the previous examples).
+
+```
+{{#with contact.person as p}}
+  {{!-- this block of code is still in the parent's scope, but
+        the #with helper provided a `p` name with a
+        helper-provided value --}}
+  <p>{{p.firstName}} {{p.lastName}}</p>
+
+  {{!-- `title` here refers to the outer scope's title --}}
+  <p>{{title}}</p>
+{{/with}}
+```
+
+Today, this capability is hardcoded into the two special forms,
+but it can be useful for other kinds of components. For example,
+you may have a calendar component (`ui-calendar`) that displays a
+specified month.
+
+The `ui-calendar` component may want to allow users to supply a custom
+template for each day in the month, but each repetition of the template
+will need information about the day it represents (its day of the week,
+date number, etc.) in order to render it.
+
+With the new "block parameters" feature, any component will have
+access to the same capability as `#each` or `#with`:
+
+```handlebars
+<ui-calendar month={{currentMonth}} as |day|>
+  <p class="title">{{day.title}}</p>
+  <p class="date">{{day.date}}</p>
+</ui-calendar>
+```
+
+In this case, the `ui-calendar` component iterates over all of days
+in `currentMonth`, rendering each instance of the template with
+information about which date it should represent.
+
+We also think that this feature will be useful to allow container
+components (like tabs or forms) to supply special-case component
+definitions as block params. We are still working on the details,
+but believe that an approach along these lines could make these
+kinds of components simpler and more flexible.
+
+### Transition Plan
+
+Block parameters will hopefully land in 1.12, and at that point the
+two special forms for `{{each}}` and `{{with}}` will be deprecated.
+You should refactor your templates to use the new block parameters
+syntax once it lands, as it is a purely mechanical refactor.
+
+**We have no plans to remove support for the `{{each}}` and `{{with}}`
+special forms in Ember 2.0.**
+
+## More Consistent Handlebars Scope
+
+In today's Ember, the `each` and `with` helpers come in two flavors: a
+"context-switching" flavor and a "named-parameter" flavor.
+
+```handlebars
+{{#each post in posts}}
+  {{!-- the context in here is the same as the outside context,
+        and `post` references the current iteration --}}
+{{/each}}
+
+{{#each posts}}
+  {{!-- the context in here has shifted to the individual post.
+        the outer context is no longer accessible --}}
+{{/each}}
+```
+
+This has proven to be one of the more confusing parts of the Ember
+templating system. It is also not clear to beginners which to use,
+and when they choose the context-shifting form, they lose access to
+values in the outer context that may be important.
+
+Because the helper itself offers no clue about the context-shifting
+behavior, it is easy (even for more experienced Ember developers)
+to get confused when skimming a template about which object a value
+refers to.
+
+In Ember 1.10, we will deprecate the context-shifting forms of
+`#each` and `#with` in favor of the named-parameter forms.
+
+### Transition Plan
+
+To transition your code to the new syntax, you can change templates
+that look like this:
+
+```hbs
+{{#each people}}
+  <p>{{firstName}} {{lastName}}</p>
+  <p>{{address}}</p>
+{{/each}}
+```
+
+with:
+
+```hbs
+{{#each people as |person|}}
+  <p>{{person.firstName}} {{person.lastName}}</p>
+  <p>{{person.address}}</p>
+{{/each}}
+```
+
+**We plan to deprecate support for the context-shifting helpers in Ember
+1.10 and remove support in Ember 2.0.** This change should be entirely
+mechanical.
+
+## One-Way Bindings by Default
+
+After a few years of having written Ember applications, we have observed
+that most of the data bindings in the templating engine do not actually
+require two-way bindings.
+
+When we designed the original templating layer, we figured that making
+all data bindings two-way wasn't very harmful: if you don't set a
+two-way binding, it's a de facto one-way binding!
+
+We have since realized (with some help from our friends at React), that
+components want to be able to hand out data to their children without
+having to be on guard for wayward mutations.
+
+Additionally, communication between components is often most naturally
+expressed as events or callbacks. This is possible in Ember, but the
+dominance of two-way data bindings often leads people down a path of
+using two-way bindings as a communication channel. Experienced Ember
+developers don't (usually) make this mistake, but it's an easy one to
+make.
+
+When you use the new component syntax, the `{{}}` interpolation syntax
+defaults to creating one-way bindings in the components.
+
+```handlebars
+<my-video src={{url}}></my-video>
+```
+
+In this example, the component's `src` property will be updated whenever
+`url` changes, but it will not be allowed to mutate it.
+
+If a template wishes to allow the component to mutate a property, it can
+explicitly create a two-way binding using the `mut` helper:
+
+```handlebars
+<my-video paused={{mut isPaused}}></my-video>
+```
+
+This can help ease the transition to a more event-based style of
+programming.
+
+It also eliminates the boilerplate associated with an event-based style
+when working with form controls. Instead of copying state out of a
+model, listening for callbacks, and updating the model, the `input`
+helper can be given an explicit mutable binding.
+
+```handlebars
+<input value={{mut firstName}}>
+<input value={{mut lastName}}>
+```
+
+This is similar to the approach taken by [React.Link][1], but we think
+that the use-case of form helpers is sufficiently common to make it
+ergonomic.
+
+[1]: http://facebook.github.io/react/docs/two-way-binding-helpers.html
+
+### Transition Plan
+
+The new one-way default is triggered by the use of new component syntax.
+This means that component invocations in existing templates will
+continue to work without changes.
+
+When transitioning to the new HTML-based syntax, you will likely want to
+evaluate whether bindings are actually being mutated, and avoid using
+`mut` for values that the component never changes. This will make it
+easier for future readers of your template to get an understanding of
+what properties might be changed downstream.
+
+To preserve the same semantics during a refactor to the new HTML-based
+syntax, you can simply mark all bindings as `mut`.
+
+```handlebars
+{{!-- these are semantically equivalent --}}
+
+{{my-video src=movie.url paused=controller.isPaused}}
+
+<my-video src={{mut movie.url}} paused={{mut controller.isPaused}}>
+</my-video>
+```
+
+While the above example preserves the same mutability semantics, it
+should be clear that the video player component should never change the
+`url` of the `movie` model.
+
+To make sure you get an exception should this ever happen, simply remove
+the `mut`:
+
+```handlebars
+<my-video src={{movie.url}} paused={{mut controller.isPaused}}>
+</my-video>
+```
+
+**We have no plans to remove the old-style component syntax in Ember
+2.0, so the semantics of existing component invocations will not
+change.**
+
+## Separated Component Parameters
+
+In today's Ember, parameters passed to components as attributes become
+properties of the component itself, putting them in the same place as
+other internal state.
+
+This can be somewhat confusing, because it may not be obvious to the
+reader of a component's JavaScript or template which values are
+internal, and which are passed in as part of the public API.
+
+To remind themselves, many Ember users write their components like this:
+
+```js
+export default Component.extend({
+  /* Public API */
+
+  src: null,
+  paused: null,
+  title: null,
+
+  /* Internal */
+  scrubber: null
+})
+```
+
+It can also be unclear how to react to a change in the external
+properties. It is possible to use observers for this purpose in Ember,
+but observers feel low-level and do not coordinate very well with the
+rendering process.
+
+To reduce confusion, we plan to move external attributes into a new
+`attrs` hash.
+
+If you invoke a component like this:
+
+```handlebars
+<my-video src={{movie.url}}></my-video>
+```
+
+then the `my-video` component accesses the passed-in `src` attribute as
+`this.attrs.src`.
+
+We also plan to provide lifecycle callbacks (modelled after [React's
+lifecycle callbacks][react-lifecycle]) for changes to `attrs` that will
+integrate with the rendering lifecycle. We plan to supplement the API
+with callbacks for changes in individual properties as well.
+
+[react-lifecycle]: http://facebook.github.io/react/docs/component-specs.html#lifecycle-methods
+
+### Transition Plan
+
+In Ember 1.10, we will begin installing provided attributes in the
+component's `attrs` hash. If a provided attribute is accessed directly
+on the component, a deprecation warning will be issued.
+
+In applications, you should update your component JavaScript and
+templates to access provided attributes via the component's `attrs`
+property.
+
+**In Ember 2.0, we will stop setting attributes as properties on the
+component itself.**
+
+We will also provide a transitional mixin that Ember addons can use that
+will make provided attributes available as `attrs.*`. This will allow
+add-ons to move to the new location, while maintaining support for older
+versions of Ember. We expect people to upgrade to Ember 1.10 relatively
+quickly, and do not expect addons to need to maintain support for Ember
+1.9 indefinitely.
+
+## Routeable Components
+
+Many people have noticed that controllers in Ember look a lot like
+components, but with an arbitrary division of responsibilities. We
+agree!
+
+In current versions of Ember, when a route is entered, it builds a
+controller, associates a model with it, and hands it off to an
+(old-style) view for rendering. The view itself is invisible; you just
+write a template with the correct name.
+
+We plan to transition to: when a route is entered, it renders a
+**component**, passing along the model as an `attr`. This eliminates a
+vestigial use of old-style views, and associates the top-level template
+with a regular component.
+
+### Transition Plan
+
+Initially, we will continue to support routing to a controller+template,
+so nothing will break. Going forward, routes will route to a component
+instead.
+
+In order to do that refactoring, several things will change:
+
+* Instead of referring to model properties directly (or on `this`), you
+  will refer to them as `model.propName`.
+* Similarly, computed properties that move to your component will need
+  to depend on `model.propName` if they are migrated from an
+  `ObjectController`.
+* In both cases, the short version is that you can no longer rely on the
+  proxying behavior of `ObjectController` or `ArrayController`, but you
+  can remedy the situation by prefixing `model.` to the property name.
+* Unlike controllers, top-level components do not persist across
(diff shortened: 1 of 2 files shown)


ahacking:
@tomdale Thanks for the description of where things are heading.

It all sounds interesting but I'm still not seeing the high level picture of how the framework will hang together or how a 2.0 application will be composed.  I would like to see a high level overview/diagram of the main parts with their responsibilities, and clearly see how both data, client events and server events flow through this proposed 2.0 architecture.

There has been mention of services previously and now rout-able components and deprecation of controllers but I see no mention of services in this RFC.  It would be immensely useful to have the big picture/straw-man so we can see if it will stand up to the use cases.

Currently my controllers need to fetch data for dynamic model/data lookups, perform model buffering, dirty tracking, undo/redo and also orchestrate complex model persistence. I can't see components fulfilling these functions but they need to be taken care of somewhere.  In React/Flux these concerns fall clearly into action-creator/dispatcher/store, but where do these things live in an Ember 2.0 application?

I haven't seen the moving parts required to support a uni-directional Flux-like architecture:
-  Are there high level actions?
- Is there a dispatcher?
- Are there stores/state managers?
- How are actions that update state coordinated?
- How would stores/state be updated in response to an action whether originating from the client or a server?
- What in Ember 2.0 will support a unidirectional data flow?
- Is Ember 2.0 an asynchronous or synchronous architecture from some dispatcher like concern down?
- Has any thought been given to immutable datatypes as used in Om and React/Morearty.js? The reason I ask is that entire component tree rendering can be avoided completely via a simple === state test, and undo/redo is practically free. 

On the DOM side of things has consideration been given for a DOM builder api, ala React?  Whilst I concede handlebars templates may be more designer friendlier (with some caveats), I personally dislike using handlebars templates with their limited expressibility and issues of scope, context and inability to reference nested components.

I realise that handlebars/htmlbars being declarative was an approach to being able to efficiently and selectively update the DOM on change but that's largely historical baggage with a DOM diff approach. Using real JS and a DOM builder API I have no limitations on how I compose my components as the expressiveness is the full power of JS. Scope and context is pure JS, there is never any mystery. I can even use something like JSX to have HTML-like markup right in my code and avoid using helpers or the overhead, inconvenience and the bear traps of computed properties, which an approach based on declarative handlebars templates forces you into.  I believe its imperative to have an official DOM builder API and let developers choose for themselves what html composition approach they want to use.

Whilst every detail in this RFC sounds mostly positive its not clear that Ember 2.0 based on this RFC has communicated a clear high level architecture, despite the significant detail provided.

ef4:
> I realise that handlebars/htmlbars being declarative was an approach to being able to efficiently and selectively update the DOM on change but that's largely historical baggage with a DOM diff approach. 

@ahacking, that's a misunderstanding of the reason for declarative templates. Upstream Handlebars literally doesn't even support selective DOM updates, so that can hardly be the reason it exists.

Declarative templating is an architectural decision. It enforces separation of concerns. It helps ensure that no business logic accidentally starts leaking into templates where it doesn't belong.

This is one of those freedom-vs-integrity issues where programmers will always vehemently disagree, so I'm not trying to convince anyone. I'm just pointing out that we consider declarative templates a deliberate feature, not a performance optimization.

ahacking:
@ef4 thanks for your clarification of handlebars proper and the ember philosophy.

Its clear that Ember would not be able to do selective update efficiently if it didn't make templates declarative and the dom update driven off property changes.

I've also seen it stated that Ember will be moving towards a DOM diff approach based on what React has proven. Perhaps that is not the case? If so I understand the need to stick with templates and the lack of expressivity that comes with that, all the special helpers to do iteration and poor mans conditional logic plus the need to also use CPs to shape and prepare the data and provide logic values to those conditional helpers.

Given that components are the way forward and are about UI rendering of state and UI event handling, the template as an answer to separation of concerns argument seems a bit stretched.

The separation of concerns I would like to see in this facet of Ember are Component, DOM specification, DOM update (the real rendering), where DOM specification is a builder api and can be used by regular JS code or various flavours of templating be it handlebars, emblem, JSX or something else.

Forcing templates, computed properties and helpers to patch up for the lack of expressiveness in templates for Ember 2.0 seems a little anachronistic given the intent of components. I accept they will be needed to help people migrate existing code but that doesn't justify handlebars templates as the only way forward or even the right abstraction.

If I'm at odds with the Ember direction then so be it, ideally I'd like to see a lot of stuff removed from Ember that drive wtf moments such as CP's and observers and the need to use run loop deferred updates to fix surprises and feedback loops that pop up all the time.

My key concern is data flow and events, which CPs and observers really hurt. Templates drive me down that road and I don't want to go down it any further. Unless Ember has a good end to end solution for data flow which is easy to reason about (which it currently doesn't IMO) I believe the ability to  develop and maintain ambitious web apps will remain a challenge, as will the ability for developers to get up to speed with Ember compared with competing frameworks.

I see 2.0 as an opportunity to paint a superior architecture and I'd like to see that picture clearly defined end to end so that all these feature changes and additions being discussed can be clearly evaluated. State management, data flow and event handling is immensely important for a predictable and easy to use yet powerful framework.

kingpin2k:
Ditching controllers doesn't seem that big of a deal.  Code can be easily shared with mixins or using a service layer for what was past controller state.

I'm far more concerned about the conglomerate that we're calling routeable components.  The name alone sounds like a future collision of undestanding.  Unless they end up merging into the same thing it introduces yet another point of confusion where we're forced to distinguish between components and routeable components (component/view, arraycontroller/objectcontroller/controller).

Additionally it sounds like we're moving away from good clean separation of concern and just saying place everything in the routeable component.  Feel free to fetch your model from the route as usual, but then put everything else in the component.  I'm probably making a bigger deal of this than is necessary, it just seems like we're squeezing things down to a point where we end flipping the bird at the single responsibility principle and creating ginormous components.

The other issue I know has been mentioned a few times in here, is communication between components is a pain in the ass.   With the concept of actions and bubbling and keeping components contained it feels like we will fall into tech support mode of having to repeat over and over "well routeable components allow actions to bubble, but normal components don't."   It's unlikely that when I build a routeable component I want to have to force someone to define the portions of the API I want exposed.  I probably want every communication that's sent to route out of my component.  

Those are my thoughts, keep up the awesome work.

intuitivepixel:
:+1:  @kingpin2k  - 99,99999% Agreement!

trek:
@ahacking 

>I've also seen it stated that Ember will be moving towards a DOM diff approach based on what React has proven. Perhaps that is not the case? If so I understand the need to stick with templates and the lack of expressivity that comes with that, all the special helpers to do iteration and poor mans conditional logic plus the need to also use CPs to shape and prepare the data and provide logic values to those conditional helpers.

DOM diffing is an internal approach to tracking and applying data updates to DOM. It has nothing to do with using or not using templates.

> Templates drive me down that road and I don't want to go down it any further.

It sounds like what you want is is JSX or its JavaScript version. Why not just use React?

klclee:
:+1: @trek

ahacking:
@trek I certainly have been experimenting with React because of problems I have encountered in the Ember paradigm. React allows immutable data and unidirectional data flow which are very important architectural approaches IMO.

I also have an investment in Ember and I'd like to not throw that away if it can evolve in the right direction quickly enough.

Its the synchronous unidirectional flow I want, the so called " the data down" and state changing actions from the top , the ability to use immutable data types. Its about proper control over the state and state flow that doesn't require CPs and observer chain reactions as the basis of state proprogation and preparing of data for declarative templates.

If Ember can support true data down flow under the control of the application and immutable types then great. But CPs and observers don't work that way under the covers even though there may be the illusion of data down from what you express in templates.

I think this is within Embers grasp, I'm not a React fan, but a fan of deterministic state management and data flow that doesn't take indeterminate and surprising paths outside the control of my application to propagate state as Ember currently does. CPs and observers guarantee random state propagation and templates mandate their use.

@klclee well that was value add.

gtrak:
@trek in principle ember doesn't force you to use handlebars.  In practice, all the observer glue makes it hard to use anything else, without knowing the details of observers.

We're also experimenting with React in a pre-existing ember codebase, but I was admittedly more a fan of React's data-agnostic approach before I even started using ember ~6 months ago.

ahacking:
@trek. Embers use of CPs and observers allows it to know when state has been changed so it can rerender parts of the DOM that are dependent on those changes. If Ember takes a Dom diff approach then declarative templates, helpers and CPs are not a critical component of achieving efficient Dom update anymore. Hence my previous statement, declarative templates and CPs/observers could become completely optional.

stefanpenner:
> if it can evolve in the right direction quickly enough.

OSS is an ecosystem which enables scratching ones own itch, and collaborating with others to work together in scratching those itches. Ultimately many hands make light work, and it is often quite fun.

@ahacking 
Unfortunately this comment thread has become somewhat unwieldily. 

Rather then long comments that repeat previous comments. Let me recommend, focused proposals and patches to identify and solve the pain points near and dear to your heart.

stefanpenner:
@ahacking if you have time to chat about this, ping me on IRC or on google hangouts. I would love to try and work with you so we can continue to evolve front-end web app development.

heat:
How nice it will be a hang out on air to talk about that +1 if it happens. Count me in as a viewer

-----Mensagem Original-----
De: "Stefan Penner" <notifications@github.com>
Enviada em: ‎05/‎01/‎2015 14:35
Para: "emberjs/rfcs" <rfcs@noreply.github.com>
Cc: "Onezino Gabriel" <zinogabriel@gmail.com>
Assunto: Re: [rfcs] The Road to Ember 2.0 RFC (#15)

@ahacking if you have time to chat about this, ping me on IRC or on google hangouts. I would love to try and work with you so we can continue to evolve front-end web app development.

ahacking:
@stefanpenner thanks, much appreciated. I'd like to get into some detail with yourself and anyone else who's  genuinely interested in a productive discussion. I'm on IRC currently, my time-zone is GMT+10, hopefully I can catch you.

klclee:
sorry @ahacking i was just agreeing.. indeed i should be more thoughtful... I would also love to listening to this discussion if it happens. With my experience on react the data down approaches (using ```props``` to affect state of children elements). i have found that to be limiting somewhat. I ended up with a bunch of custom event listeners fired triggered from the child element to affect the parent owed props. Many of this became unmanageable for me, and could end up having sync issues? Maybe there is a better pattern for that in react but i am no expert in it. Using observers which I think is managed in the run loop makes me sleep a little better i guess... However I love to hear what the different approaches can be...

ahacking:
@klclee no worries glad to have your input and the experiences in the mix.

In Ember 2.0 I am certain sending actions to parent will be very similar to how you do it in React.  If you don't pass the handler into the component as an attr the child won't be able to call a handler on an ancestor. This mirrors how Ember components currently work assuming you're not using parentView hacks.

The patterns I'm most interested in from React is a the flux like approach of calling functors as top level action creators from ui event or action handlers within components. Those actions are handled by top level services/apis or stores. Once a store handles an action it should be only synchronous propagation of state to top level components and from those down to leaf components. I also want to be able to support immutable data types and see two way bindongs and observers minimised as much as possible because of the problems they create.

I think this is all doable and closely aligned with where things are headed. Just need some consensus and agreement within core of the best way forward.

ahacking:
@mgenev, @aceofspades there are a valid reasons for localised action handling as well as global actions that change application state. I agree that actions up all the way as per current ember, and having to pass lots of action handlers down contingently is painful and that might be intentional.

In localised action handling the component doesn't have enough context to know what should happen so it calls the handler given to it by its parent on attrs. This is typical for collection as parent and item as child scenarios and for general purpose UI components.

At some point up the chain you reach a component that needs to change application state., eg 'add a comment to a post'. This is where I want to see global semantic actions used to effect application state much like the React reflux addon.  Those global actions are handled by services and stores to change app state "at the top", which then starts the state propagation from top level components in the app hierarchy down to leaves, after which the render phase begins. Those top level components are what I believe routable components will be as route/URL is after all an important part of application state.

If you end up with state change logic in a component then that is a red flag that you should be putting that logic in a service or store (flux concept) and using global semantic actions like 'addComment'. State change logic in components will require long handler chains down and action propagation up so that each conponent along thr way can implement its piece of logic and that is painful. I think that pain is good as its the wrong thing to do and mixes concerns. Components shouldn't be managing application state or contain logic on how app state should be mutated, that belongs in services/stores. Components say what should happen, not how.

I'm just proposing how I see that things could work, not necessarily how it will in actuallity since not all of the vision of core has been set down in stone yet. I don't believe it's out of step with where things are heading.

We need different topics for these discussions as per @stefanpenner previous suggestion. Can we break out 2.0 into some themes? Do they need to be separate RFCs?

aceofspades:
Is there a strong advantage to using the same `Component` class to handle not only abstract, reusable, library objects, but also app-level concerns, and now routable ones? Are these responsibilities distinct enough to warrant special naming and behavior? 

To me, it seems clear to that declarations make sense when crossing domain boundaries, i.e. library to app. You're mapping one namespace into another, and choosing which data to publish and which events to subscribe to. Once in the app domain, you've got a controlled scope and event namespace, and the advantages of convention-over-configuration apply.

mgenev:
@aceofspades I agree fully about declarations making space when crossing domain boundaries. It comes down to this,  in ember 1.0 we have _the choice_ - if we want something to be tied to the app (domain) we'd use controller/view which was accessible from anywhere in the app. If we wanted something to be isolated and therefore portable between domains, we'd use a component.  This worked well and was dry and great (minus the annoyance of nesting components, I'd personally much rather use a global pub-sub/dispatcher than pass an action 4 times). 

Now that we're getting forced into components instead of controllers and views, in my opinion they have to be able to assume either role. They have allow both easy access by bubbling events to the top automatically and inheriting data to the bottom automatically, or if the user chooses so they can become isolated and force the declaration which serves as a handshake between 2 domains.

This would solve the problem.

ef4:
Parts of this discussion are veering in weird directions that don't seem to understand how Ember works today, nevermind what's proposed for 2.0. Firstly:

> Also using router to communication between different components (controllers?) does not make too much sense either (what if the component does not need to touch the server?) 

That doesn't really make sense. Whether you'd involve the Ember router has nothing to do with whether you need to touch the server. The router exists to manage global application state and tie it to the URL, bidirectionally. If a component is going to alter the global state, it almost certainly does want to involve the router, otherwise you have broken web semantics.

Secondly, @mgenev's JSBin is ignoring how templates work in Ember. All of those component action handlers are completely unnecessary, even in today's Ember, because the innermost part of this template:

````hbs
    {{#level-1}}
        {{#level-2}}
          {{#level-3 }}
            <button {{action 'handleAction'}}>
              Click me (yielded).
              "Inside a yield, the target points at the original target""
            </button>
          {{/level-3}}
        {{/level-2}}
      {{/level-1}}
````
Is still invoked in the context of it's owner, which in this example is the top-level application controller. So the handleAction event goes straight to the application controller anyway, and all those component action handlers are superfluous.

I would guess what you're really trying to illustrate is what happens if all those templates have their own separate templates, and are not actually nested like this in a single template. In that case it's true that you'd need to be explicit about event propagation, but given the isolation you're *choosing* to use by giving each component its own isolated template, being explicit it entirely appropriate.

Also the proposed 2.0 syntax significantly streamlines the syntax for explicit event propagation. You can do it all within templates without creating action handlers in JS files at all.

mgenev:
"I would guess what you're really trying to illustrate is what happens if all those templates have their own separate templates, and are not actually nested like this in a single template. "

Yes that's the whole point.

" In that case it's true that you'd need to be explicit about event propagation, but given the isolation you're choosing to use by giving each component its own isolated template, being explicit it entirely appropriate."

I'm not choosing to. Right now if I want to not have to rewrite everything I write in the next 6 months, I have to avoid controllers and views hence to be dry and make my hierarchy, I have to be nesting components instead and I end up in that situation...

ef4:
> Do you have a better suggestion?

Yes. It's nuts to try to avoid controllers entirely at this early stage. The code to support that is literally not written yet. This RFC exists so that the whole community can be involved in that process from the beginning -- not because it's imminently about to be done.

The stated recommendation is to use `Controller` (but not `ObjectController` or `ArrayController`, which are already deprecated on canary).

Everybody, including the core team, is literally writing apps today that still using Controllers, because that's how Ember apps are written today. We are not going to trash all of our code and yours by making it suddenly all completely broken. The deprecations are planned in very gentle steps so that you're never spending more than a little bit of effort making almost entirely mechanical changes, and nobody gets left behind or slammed with a giant chunk of rewriting.

mgenev:
I see. Use controllers and rewrite when the new syntax hits which will alleviate these issues which answers my question. Thanks for the explanation.

MiguelMadero:
@mgenev a common misconception when it comes to components is that they're all about re-use, when IMO, they're all about encapsulation, being re-use and portability a side benefit. Even nested components are often domain specific. 

I rarely see the need to bubble actions by more than 1-3 levels deep, and we nest components, a lot. However, we often encapsulate the actions at some sub-level. Often these actions result in a change to a model, if other parts of the application depend on the same model, they're automatically updated without having to know about the action that triggered the change.

trek:
@MiguelMadero this. Components are about both, but people forget the isolation part. 

In training I tend to tell people to think of components as functions: they have their own state (which is like a functions `var` statements) and passed in data (an function's arguments). You can use a function to isolate behavior and reuse behavior. If you _only_ used it to reuse behavior, we'd all write many-hundred-line functions all the time which would make testing (and understanding) difficult.

knownasilya:
I like how Vue.js explains the terms that are used in the libaray: http://vuejs.org/guide/#Concepts_Overview

hipertracker:
@knownasilya  "Each Vue instance is associated with a corresponding DOM element." That's must be damn slow. React is using Virtual DOM, much better concept.

knownasilya:
@hipertracker good and well, but we're talking about documentation and the ambiguity and connotation that some words carry when describing core concepts, and Vue.js does that well, without many assumptions. I suggest we do something similar.

ResentedHook:
@knownasilya and @mgenev 

As it turns out, you can trick the new syntax by doing something like:

`{{#each this in paged itemController='messages.message-list-item'}}`

it is strange, to me, that just giving it "this" would satisfy the syntax and controller.

changeweb:
If `Array & Object` controllers are deprecated in Ember 2.0, then how can we manage dependencies between Components by getting computed alias in Components? Something like:

    Ember.computed.alias('controllers.application');

joostdevries:
@changeweb From what I know there are two cases here:

- Access to properties from an arbitrary other controller (using `needs` in 1.x)

  For these cases, I think [services](http://discuss.emberjs.com/t/services-a-rumination-on-introducing-a-new-role-into-the-ember-programming-model/4947) offer a very nice pattern. You isolate any type of state or functionality that goes beyond the component at hand in a service. 

- Parent/child relations

  This should be doable with components but mutations and functional calls will have do be done using actions.

changeweb:
@joostdevries It looks interesting. Could you please give a small demo in [jsbin](http://emberjs.jsbin.com/) for a better & clear understanding of *Services*?

joostdevries:
@changeweb http://emberjs.jsbin.com/quranekafa/1/

changeweb:
This injection is working for controllers. If I remove `App.inject('route', 'session', 'service:session');`, this still works. But if I remove `App.inject('controller', 'session', 'service:session');`, this doesn't.

joostdevries:
@changeweb that's correct. Instead of "needing" the dependency you now have to inject it.

samdelagarza:
My two cents, depending on observables and computed properties across SRP boundaries (I.e. controllers) can lead to a mess. I've experienced this first hand and have preferred clear API boundaries across objects/services/interests. 

Sam

changeweb:
One may need to use lots of dependencies between components. `needs` in controller manages dependencies between controllers pretty well. We need something like `needs` or better dependency handler for route as well as components in Ember 2.0.

knownasilya:
@changeweb Have you seen http://emberjs.com/api/classes/Ember.html#method_inject_controller and http://emberjs.com/api/classes/Ember.html#method_inject_service?

changeweb:
@knownasilya Sorry I've not seen them before. I used `needs` which is functionally equivalent to `Ember.inject.controller()`.

Thanks for mentioning `Ember.inject.service()`. I think `Ember.inject.service()` will do the work by making properties easily accessible in the model hook.

mgenev:
@changeweb this stuff is just coming out now in 1.10 that's why you haven't seen it

kristianmandrup:
@hipertracker I totally agree with sentiments. Time has run out for "complete" MVC frameworks. Ember should be split up into smaller parts. Why be forced into a specific template language such as Handlebars. I just don't get it. It made sense at the time I guess (~4 years ago). All about being lightweight and flexible and build the architecture to your specific needs... Travel light ;)

drogus:
> Why be forced into a specific template language such as Handlebars.

You're not forced to use Handlebars.

kristianmandrup:
@drogus oh cool, well I assumed only Handlebars since I have yet to see an example of another Template language  being used with Ember. How easy is it to opt out and use your own favourite (Reactive) rendering engine? Opens up some new exciting possibilities :)
I know there is a React-Ember library, but assumed it hooked into Handlebars as well... Thanks!

changeweb:
@kristianmandrup Have you read this [can-reactjs-be-used-as-a-view-within-emberjs](http://discuss.emberjs.com/t/can-reactjs-be-used-as-a-view-within-emberjs/3470). For short [Using React](http://discuss.emberjs.com/t/can-reactjs-be-used-as-a-view-within-emberjs/3470/6) and [Using Ember view](http://discuss.emberjs.com/t/can-reactjs-be-used-as-a-view-within-emberjs/3470/7).

I think you can do almost all react thing with Ember view.

kristianmandrup:
@changeweb Awesome! Must admit I had been a bit disillusioned about both Ember and Angular lately... will be interesting to see how Ember 2.0 will shape up in the end. It is key that it is suficiently lightweight/flexible and doesn't "lock you in" IMO

taras:
@kristianmandrup 

> Why be forced into a specific template language such as Handlebars.

You can use [Emblem.js](http://emblemjs.com) with Ember instead of Handlebars, but I'm not sure what it's future is. Ember is all about creating happy paths. Once HTMLBars if fully baked into Ember, it will probably be possible to have other syntaxes use it to provide other templating syntaxes. 

> Ember should be split up into smaller parts.

Ember is already made out of many small parts. Templating is provided by [Handlebars.js](http://handlebarsjs.com), Promise implementation is done with [RSVP.js](https://github.com/tildeio/rsvp.js/). The run loop is [backburner.js](https://github.com/ebryn/backburner.js/). The router is [router.js](https://github.com/tildeio/router.js/). 

>  All about being lightweight and flexible and build the architecture to your specific needs... Travel light ;)

Ember never aimed to be lightweight. That's why it's a framework and not a library. It's always been created to provide everything that developers need to be productive.

kylecoberly:
I agree with Taras- there are plenty of roll-your-own options out there,
and Ember is already pretty pretty modular.

knownasilya:
The only thing missing is tree-shaking, to remove the unused parts and make your app speedier.

mgenev:
@knownasilya +111111

heat:
> Ember never aimed to be lightweight. That's why it's a framework and not a library. It's always been created to **provide everything that developers need to be productive**. 
I agree
It's what I looking for. It sounds great when I hear about decoupled parts and pluggable. But nowadays what I want is productivity.

davidlormor:
So I just wanted to chime in again on the discussion around routable components - I think I've seen the light, so to speak.  After migrating my "very large" ember project to ember-cli and adopting a "component-centric" development approach, I "get it"...actually it feels painful when I have to spin up a View to wrap my route-associated templates now.

While I still feel there is definite validity to the arguments around the separation of concerns, object/array controller associations, and service controller, I've heard a lot from core team members lately via various channels and am confident that they're driving this forward in a way that will properly deal with those concerns.  The development momentum I've built the past couple months using this approach can't be denied, and am super excited about the future of this framework!

QuantumInformation:
best news of all: 

>Instead, we plan to do the vast majority of new work on master (behind feature flags), and land new features in 1.x as they become stable.

adrianboston:
NSArrayController->selectedObject cici n'est pas une cocoa

wycats:
This was a really awesome discussion that significantly informed the Ember 2.0 direction. From this point forward, it makes more sense to open new RFCs for new features. If there's a feature you want implemented, definitely start a conversation about it.

Pieces of this RFC have since been expanded to full RFCs, or have been implemented, so definitely continue any relevant discussion on those RFCs.

For the big enchilada, expect an RFC for Routable Components to come in the next few days. You can also follow along with the new component implementation at ember.js/ember.js#10501

PR #16, opened 2014-11-03
Embedded handlebars draft 1

machty:
Prepared w @mmun

Pretty: 
https://github.com/machty/rfcs/blob/embedded-handlebars/active/0000-embedded-handlebars.md


chadhietala:
After 3 months of thinking about this, I wouldn't mind exploring this. @machty or @mmun have you guys done any work on this?

mixonic:
I believe this can be pretty easily done with an addon that compiles the template then creates a module or entry in `TEMPLATES` with that output. Would be really neat to see!

PR #17, opened 2014-11-04
ember data testing helpers

fivetanley:


fivetanley:
I think @samselikoff's ember-cli-mirage ended up working better for us in the long term. I don't have the bandwidth to get this RFC over. If someone else wants to do this, they could make an addon or open a new PR by copying/pasting/editing this RFC.

samselikoff:
By the way it is in my roadmap for Mirage to add the ability to generate ED models from your Mirage factories in integration & unit tests. Hopefully that will address some more of these points.

PR #21, opened 2014-11-13
Provide a way to mark a model as dirty when relationships or related model properties change.

timrwood:
# Summary

Provide a way to mark a model as dirty when relationships or related model properties change.
# Motivation

There is a common pattern to conditionally show a save button if a model needs saving.

For most cases, using the `isDirty` flag on the model is enough to determine whether a model needs saving or not. However, this flag currently ignores relationships, making it unusable for anything other than simple properties.

Ideally, the user would be able to specify which relationships are saved when the model is saved, and thus, which ones should mark the model as being dirty.
# Detailed design

In order to prevent dirtiness from spreading across the entire object graph, dirtying would have to be opt in. Users would specify on each relationship whether it would dirty the model or not.

``` js
var Comment = DS.Model.extend({
  post: DS.belongsTo('post', { dirties: true }),
  user: DS.belongsTo('user')
});

var comment = this.store.find('comment', 1);
comment.get('post');    // null
comment.get('user');    // null
comment.get('isDirty'); // false

comment.set('user', someUser);
comment.get('isDirty'); // false

comment.set('post', somePost);
comment.get('isDirty'); // true

comment.set('post', null);
comment.get('isDirty'); // false
```

For `hasMany` relationships, adding or removing a relationship would dirty the model.

``` js
var Post = DS.Model.extend({
  tags:     DS.hasMany('tag', { dirties: true }),
  comments: DS.hasMany('comment')
});

var post = this.store.find('post', 1);
post.get('isDirty');         // false
post.get('tags.length');     // 0
post.get('comments.length'); // 0

post.get('comments').pushObject(someComment);
post.get('comments.length'); // 1
post.get('isDirty');         // false

post.get('tags').addObject(someTag);
post.get('tags.length'); // 1
post.get('isDirty');     // true

post.save();
post.get('isDirty'); // false

post.get('tags').removeObject(someTag);
post.get('tags.length'); // 0
post.get('isDirty');     // true
```

Sometimes, entire related models are serialized when a model is saved. In those circumstances, it would be useful to allow the model to be dirtied when any of the related model properties are saved.

``` js
var User = DS.Model.extend({
  address: DS.belongsTo('address', { dirties: 'properties' })
});

var Address = DS.Model.extend({
  street: DS.attr('string')
});

var user = this.store.find('user', 1);
user.get('isDirty'); // false
user.get('address.street'); // 4th Ave

user.set('address.street', 'Main Street');
user.get('isDirty'); // true

user.set('address.street', '4th Ave');
user.get('isDirty'); // false
```
# Drawbacks

This would significantly increase the complexity of the `isDirty` computed property.

There would probably need to be checks in place to prevent recursive dirty checking. In the example below, once either of the models is made dirty, they would never become clean again by reverting a property, as each one's `isDirty` state is dependent on the other's.

``` js
var A = DS.Model.extend({
  b: DS.belongsTo('b', { dirties: true })
});
var B = DS.Model.extend({
  a: DS.belongsTo('a', { dirties: true })
});
```
# Alternatives

[Paul Ferrett's isFilthy computed property](http://paulferrett.com/2014/ember-model-isdirty-when-belongsto-changes/)

[Steven Lindberg's Dependent Relationships](https://gist.github.com/slindberg/8660986)

[Tim Wood's Dirty Relationships Mixins](https://gist.github.com/timrwood/c58b667650bee749d5af)
# Unresolved questions

The syntax feels like it could use some bikeshedding.

``` js
DS.belongsTo('other', { dirtiesOn: 'relationship' })
DS.belongsTo('other', { dirtiesOn: 'properties' })
```

Conceptually, a relationship should be considered dirty if the relationship is being saved by the model, so it could be named something related to saving.

``` js
DS.belongsTo('other', { savesRelation: true })
DS.belongsTo('other', { savesModel: true })
```

### DIFF
--- a/active/0000-ember-data-dirty-relationships.md
+++ b/active/0000-ember-data-dirty-relationships.md
@@ -1,0 +1,140 @@
+- Start Date: 2014-11-08
+- RFC PR:
+- Ember Issue:
+
+# Summary
+
+Provide a way to mark a model as dirty when relationships or related model properties change.
+
+# Motivation
+
+There is a common pattern to conditionally show a save button if a model needs saving.
+
+For most cases, using the `isDirty` flag on the model is enough to determine
+whether a model needs saving or not. However, this flag currently ignores relationships,
+making it unusable for anything other than simple properties.
+
+Ideally, the user would be able to specify which relationships are saved when the model
+is saved, and thus, which ones should mark the model as being dirty.
+
+# Detailed design
+
+In order to prevent dirtiness from spreading across the entire object graph,
+dirtying would have to be opt in. Users would specify on each relationship
+whether it would dirty the model or not.
+
+```js
+var Comment = DS.Model.extend({
+  post: DS.belongsTo('post', { dirties: true }),
+  user: DS.belongsTo('user')
+});
+
+var comment = this.store.find('comment', 1);
+comment.get('post');    // null
+comment.get('user');    // null
+comment.get('isDirty'); // false
+
+comment.set('user', someUser);
+comment.get('isDirty'); // false
+
+comment.set('post', somePost);
+comment.get('isDirty'); // true
+
+comment.set('post', null);
+comment.get('isDirty'); // false
+```
+
+For `hasMany` relationships, adding or removing a relationship would dirty the model.
+
+```js
+var Post = DS.Model.extend({
+  tags:     DS.hasMany('tag', { dirties: true }),
+  comments: DS.hasMany('comment')
+});
+
+var post = this.store.find('post', 1);
+post.get('isDirty');         // false
+post.get('tags.length');     // 0
+post.get('comments.length'); // 0
+
+post.get('comments').pushObject(someComment);
+post.get('comments.length'); // 1
+post.get('isDirty');         // false
+
+post.get('tags').addObject(someTag);
+post.get('tags.length'); // 1
+post.get('isDirty');     // true
+
+post.save();
+post.get('isDirty'); // false
+
+post.get('tags').removeObject(someTag);
+post.get('tags.length'); // 0
+post.get('isDirty');     // true
+```
+
+Sometimes, entire related models are serialized when a model is saved.
+In those circumstances, it would be useful to allow the model to be dirtied
+when any of the related model properties are saved.
+
+```js
+var User = DS.Model.extend({
+  address: DS.belongsTo('address', { dirties: 'properties' })
+});
+
+var Address = DS.Model.extend({
+  street: DS.attr('string')
+});
+
+var user = this.store.find('user', 1);
+user.get('isDirty'); // false
+user.get('address.street'); // 4th Ave
+
+user.set('address.street', 'Main Street');
+user.get('isDirty'); // true
+
+user.set('address.street', '4th Ave');
+user.get('isDirty'); // false
+```
+
+# Drawbacks
+
+This would significantly increase the complexity of the `isDirty` computed property.
+
+There would probably need to be checks in place to prevent recursive dirty checking.
+In the example below, once either of the models is made dirty, they would never become
+clean again by reverting a property, as each one's `isDirty` state is dependent on the other's.
+
+```js
+var A = DS.Model.extend({
+  b: DS.belongsTo('b', { dirties: true })
+});
+var B = DS.Model.extend({
+  a: DS.belongsTo('a', { dirties: true })
+});
+```
+
+# Alternatives
+
+[Paul Ferrett's isFilthy computed property](http://paulferrett.com/2014/ember-model-isdirty-when-belongsto-changes/)
+
+[Steven Lindberg's Dependent Relationships](https://gist.github.com/slindberg/8660986)
+
+[Tim Wood's Dirty Relationships Mixins](https://gist.github.com/timrwood/c58b667650bee749d5af)
+
+# Unresolved questions
+
+The syntax feels like it could use some bikeshedding.
+
+```js
+DS.belongsTo('other', { dirtiesOn: 'relationship' })
+DS.belongsTo('other', { dirtiesOn: 'properties' })
+```
+
+Conceptually, a relationship should be considered dirty if the relationship is being
+saved by the model, so it could be named something related to saving.
+
+```js
+DS.belongsTo('other', { savesRelation: true })
+DS.belongsTo('other', { savesModel: true })
+```


dbackeus:
+1 for the `saves...` suggestion rather than `dirties...`.

Soliah:
@ahacking I'd really love to see the implementation of your buffering of changes on controllers. I'm not sure if its just my inexperience with Ember speaking, but I found it quite troublesome to achieve exactly what you have describe - reading from the current model in Ember Data but not reflecting changes on the client (templates) until the changes has been persisted on the server.

MiguelMadero:
@ahacking I've been achieving something similar by "cloning" the object and then flushing the changes. I'm curious about your solution as well.

lolmaus:
I encountered the problem discussed here and resolved it before i found this thread.

Here's my solution: https://github.com/lolmaus/ember-cli-stained-by-children

Please tell me what you think of it. Star it if it helps you.

It will crash on a loop (`foo<->bar`), but it works fine with hierarchies (e. g. in `foo<-bar<-baz` staining `baz` will stain `foo`).

lolmaus:
My solution has a big drawback: it's unable to detect when a relationship itself changes, i. e. when a non-dirty child is replaced by another non-dirty child. The `child` and `children.[]` properties on a parent record trigger their observers just on parent record creation, and i can't figure out how to distinguish initial relationship populating from subsequent relationship changes.

calvinmetcalf:
> My solution has a big drawback: it's unable to detect when a relationship itself changes, i. e. when a non-dirty child is replaced by another non-dirty child. The child and children.[] properties on a parent record trigger their observers just on parent record creation, and i can't figure out how to distinguish initial relationship populating from subsequent relationship changes.

this is the thing I actually need, when the relationship changes (aka a new id is in the relationship field on the model) I need to change it in the database, I don't need to change it in the database if it's children change (I need to change them instead)

decasia:
Just wanting to weigh in here on what seems like one of the simpler points in a complex discussion:

@ahacking wrote:
> I think setting a belongsTo on a model should dirty the model that has the foreign key

I agree with this; it seems like a relatively non-controversial way to cover a certain set of simple cases. Perhaps this behavior could be supported in an add-on, if it isn't possible to have this supported by the library itself.

Kilowhisky:
I'd like to voice my approval for this at the very least,

@ahacking wrote:

> I think setting a belongsTo on a model should dirty the model that has the foreign key

The foreign key reference is a property of the model it lives in. If i redirect it to another object i have CHANGED the model. This not working is very, very counter intuitive.

smcclstocks:
Agreed. When I first started using ember-data it seemed totally logical & expected that a belongsTo change would dirty the parent since the ID has changed. A rollback would then implicitly set the belongsTo back to whatever it was when the parent was in a non-dirty state just like the other attrs on the parent.

mk-conn:
+1 
Will this be implemented in the near future? Any words on this?

MiguelMadero:
It would be good for this to be part of ED, in the meantime, I there is an addon for this:




https://github.com/lolmaus/ember-cli-stained-by-children/blob/0.x/README.md





I have not used it, bu this could help you. 



—

calvinmetcalf:
@MiguelMadero that only helps if the content of the child model changes which is larger problem space than this which is more about wanting the model to become dirty when it points to a different child.

MiguelMadero:
Sorry. I am not entirely familiar with the addon it just seemed related. It could probably help as a reference to implement what you need until

we get an official solution. 



—

Kilowhisky:
Anyone have a workaround for this right now? I've been avoiding it while working on other parts of my app but now i actually need cancel buttons to do what they say.

decasia:
@Kilowhisky I've used something like this in a mixin that I insert in my models:

```javascript
  rollbackAttributes: function() {
    this._super();
    this.rollbackRelationships();
  },

  rollbackRelationships: function() {
    var model = this;
    model.eachRelationship(function(name, meta) {
      if (meta.kind === 'belongsTo') {
        var oldId = model.get(`${name}Id`);
        if (oldId) {
          model.get(name).then(function(child) {
            if (!(child && child.get('id') === oldId)) {
              model.store.findRecord(name, oldId).then(function(originalRecord) {
                model.set(name, originalRecord);
              });
            }
          });
        }        
      }
    });
  },

  // keep a cache of all belongsTo ids before there are any user-initiated changes
  // so we can roll back manually in `rollbackAttributes`
  cacheRelationships: function() {
    var model = this;
    model.eachRelationship(function(name, meta) {
      if (meta.kind === 'belongsTo') {
        model.get(name).then(function(child) {
          if (child && child.get('id')) {
            model.set(`${name}Id`, child.get('id'));
          }
        });
      }
    });
  },

  ready: function() {
    this.cacheRelationships();
  }
```

Keeo:
@decasia I made [gist](https://gist.github.com/Keeo/8e900ed063e06000d06a) that supports sideloaded entities with async false based on your code.

pjcarly:
I just ran into this issue, I would expect changing a model's belongsTo Id field would flag the model as dirty, and rolling back the attributes would revert the change and put the original id back into the belongsTo relationship.

Thanks for sharing the mixin @decasia I'll give it a try.

Keeo:
@pjcarly careful with that. It doesn't work with ```async:false```.

pjcarly:
@Keeo  Ok cheers, isn't async: true the default?

Keeo:
@pjcarly Yes, but you may want to change it to false for *static* models like languages, states,... and that opens a can of worms. I was trying to modify the code and make gits out of it but I failed heavily.

pjcarly:
Cheers, thanks for the tips, I'll keep them in mind.

oswaldoacauan:
Run into this belongsTo trap today.

:+1: for this RFC

GeoffreyBooth:
I ran into this problem today, and I wrote an extension of DS.Model based on @decasia's [code](https://github.com/emberjs/rfcs/pull/21#issuecomment-135134132) above and [this StackOverflow post](http://stackoverflow.com/a/27184207/223225), updated for Ember 2.1. See my version [here](https://gist.github.com/GeoffreyBooth/cc0e7db6696649435a2a#file-model-coffee). Feel free to fork and rewrite in ES2015/Ember CLI-style, it should be nearly identical.

mk-conn:
Still no progress on this in the official ember-data? Why is it such a big deal (at least it seems like that? to dirty a model if the relationship changes?

bnetter:
Having the same problem today...

nucleartide:
Ran into this problem, :+1: for the RFC as well. :)

Kilowhisky:
Still struggling with this. I'm currently trying a modified version of what @decasia has but its not very elegant since i cannot get at the raw id that ember thinks the relationship correlates to. It seems very, very odd that you can't just access the identifier for a relationship without loading the relationship from the damn server. 

After some more thinking at this point the only way i think this feature would work with existing code would be something like the following. 

* Add a method `rollbackRelationships()` to `ds.model`
* Add a property `hasDirtyRelationships` to `ds.model`
* Add a method `changedRelationships()` to `ds.model`
* Change the `dirtyType` property when a relationship is dirtied. This may or may not break existing code?
* BONUS: Add a property `relationshipIds` that has a hash of the relationships and the locally known ids.

webark:
@Kilowhisky Does http://emberjs.com/blog/2016/03/13/ember-data-2-4-released.html#toc_code-ds-references-code get you anywhere with the ids? It has a `post.hasMany('comments').ids();` method on it.

Kilowhisky:
@webark Sweet! that's pretty much what i need. No more cluster frack of promises just to compare record equality! This should work nicely with the belongsTo fix that @decasia has. 

For now though i'm back to just saving the relationship in a temp variable after the model hook as my attempt at using the mixin didn't quite work well enough.

Awem:
OK, so we have https://github.com/emberjs/data/pull/4361 and https://github.com/emberjs/data/pull/3698. Both try to provide relationship rollback. However, there is no indication that either of them is going to be merged soon. That is bad. I would like to help, but without feedback from ember-data devs that is impossible. Anyone willing to provide an outlook?

meszike123:
There was not much progress on this ticket for a long time, so I would like to open the discussion once again, because I believe relationships bring a lot of value to ember-data, and they are very important. 

We have found an interesting and quite simple way of marking relationships dirty on model, where the basic idea is that the relationships have canonicalValue, which only changes, when the model is saved, or is reloaded. 

As our models are quite complicated, we needed two levels of dirty checking. Sometimes for us it is enough to know of the reference (belongsTo) or the array of references have changed (hasMany), but sometimes we need to dirty the model, if the data in the referenced model has changed. And of course not only mark the model dirty, but rollback it, without reloading it from backend

A simple implementation of this approach is implemented in the https://github.com/meszike123/ember-managed-relationships addon. 

I understand that this is probably not the best approach, and it would be better if relationships themselves will be refactored, but it might help somebody, till this is implemented in Ember.

vigpatrick:
@meszike123 I had severall problem with [ember-data-route](https://github.com/DockYard/ember-data-route) and the fact that my model wasn't set to dirty when i edit an relationship record. I used ember-managed-relationship as you suggest and it does the job pretty well. I had to change some code in ember-data-route for the 'isDirty'  and the rollback, but it work.

Thank you

danielspaniel:
I have new add on that does change tracking and full on relationship rollback. It is fast and efficient. 

https://github.com/danielspaniel/ember-data-change-tracker

johnunclesam:
Any news on this? I need to know in my model if some relationship (children) is still dirty...

danielspaniel:
@johnunclesam  https://github.com/danielspaniel/ember-data-change-tracker  also does dirty tracking for relationships. You can setup computed property isDirtyRelations or just use isDirty

johnunclesam:
Thanks @danielspaniel, I don't need all of that code, (BEUATIFUL, AMAZING CODE!), because I need this just once in one model. Maybe you can make this code: http://paulferrett.com/2014/ember-model-isdirty-when-belongsto-changes/ actually in Ember 2? How to?

sukima:
I don't know if this matters now but for me I did this trick to avoid more addons:

```js
// app/model/my-model.js
export default Model.extend({
  foo: hasMany('bar'),

  registerSnapshot() {
    let foo = this.get('foo').toArray();
    this.set('_currentSnapshot', { foo });
  },

  rollbackSnapshot() {
    let relations = this.get('_currentSnapshot');
    if (relations) {
      this.get('foo').setObjects(relations.foo);
    }
    this.set('_currentSnapshot', null);
    this.rollbackAttributes();
  }
});
```

```js
// app/components/my-form.js
  saveForm() {
    let model = this.get('model');
    return model.save().then(() => {
      model.registerSnapshot();
    });
  },

  didInsertElement() {
    this._super(...arguments);
    this.get('model').registerSnapshot();
  },

  willRemoveElement() {
    this._super(...arguments);
    Ember.run.scheduleOnce('afterRender', () => {
      this.get('model').rollbackSnapshot();
    });
  }
});
```

runspired:
Dirty tracking of relationships is now possible to be implemented by an addon via the `RecordData` RFC (#293). Many addons now exist that handle this and which can be further improved by making use of this feature.

I believe we should close this in favor of encouraging RecordData approaches or improving RecordData interfaces to make it smoother to do.

runspired:
After discussing with the rest of the team, there is agreement on the points I made above.

hjdivad:
@igorT @runspired reopening to remind us to give more detailed guidance + test re custom `RecordData`

runspired:
I've reached out to ember-data-change-tracker to offer review/guidance for reimplementing via RecordData. We need to expand the documentation as well to include how to use RecordData. That is being tracked here https://github.com/emberjs/data/issues/6359

navels:
Seems that ember-data-change-tracker isn't actively maintained and given the outstanding issues probably not something we want to incorporate into our app.  Here in 2023, assuming Ember 4+ (we are on 3.28 so that would be even better but we are upgrading soon), is there a recommended approach or addon?

runspired:
@navels implementing a custom cache in 4.12 is probably the nicest way

seanCodes:
@runspired Are there any examples or documentation explaining how to use custom caches? Using something like the following I’m getting errors because the `requestManger` is not defined (Ember 4.12 + Ember Data 4.12):

```js
// app/services/store.js

import Cache from '@ember-data/json-api'
import Store from '@ember-data/store'

class CustomCache extends Cache {
  mutate(mutation) {
    console.debug('CustomCache.mutate', mutation)
    return super.mutate(mutation)
  }
}

export default class extends Store {
  createCache(wrapper) {
    return new CustomCache(wrapper)
  }
}
```

PR #24, opened 2014-11-27
HTMLBars Bound Attributes

mixonic:
[Read as HTML](https://github.com/emberjs/rfcs/blob/attr-node/active/0000-bound-attributes.md)


rwjblue:
@mixonic - This has been accepted (and is enabled by default in Ember now). Can you confirm that this description is up to date so we can merge it?

zeppelin:
One thing popped up during my experiments.

*Not sure this is the right place to discuss, so I'm happy to move this elsewhere if needed.*

The issue is, when using the `ember-htmlbars-component-generation` feature, I'm unable to pass literals other than strings to an attribute. With the current mustache component syntax this works fine:

```hbs
{{x-video autoplay=false}}
{{x-image width=100 height=100}}
```

The new component syntax fails to support this, booleans and numbers are all get converted to strings:

```hbs
<x-video autoplay=false></x-video>
<x-image width=100 height=100></x-image>
```

Wrapping the literals in mustaches unfortunately won't compile:

```hbs
<x-video autoplay={{false}}></x-video>
<x-image width={{100}} height={{100}}></x-image>
```

It can be worked around using helpers, but it's far from being clean & elegant:

```hbs
<x-video autoplay={{unbound false}}></x-video>
<x-image width={{unbound 100}} height={{unbound 100}}></x-image>
```

/cc @mixonic

mixonic:
@zeppelin I'd suggest a discussion about component attributes should go elsewhere. This RFC is limited to attribute bindings against the DOM. I don't believe we have an RFC for tag-based components yet.

zeppelin:
@mixonic Sure, I'll open a new one.

wagenet:
@mixonic what's the status of this? We have this functionality, right?

PR #26, opened 2014-12-24
RFC: toggle-visibility

jamesarosen:
[Rendered](https://github.com/jamesarosen/rfcs/blob/toggle-visibility/active/0000-toggle-visibility-api.md)

### DIFF
--- a/active/0000-toggle-visibility-api.md
+++ b/active/0000-toggle-visibility-api.md
@@ -1,0 +1,49 @@
+- Start Date: 2014-12-24
+- RFC PR:
+- Ember Issue:
+
+# Summary
+
+Expose as public API the places where views set or toggle their visibility.
+
+# Motivation
+
+In the current implementation, Ember Views set initial visibility via
+`buffer.style('display', 'none')` and toggle it via
+`$el.toggle(isVisible);`. That is, the implementation uses `visibility`.
+HTML5 introduced a new `hidden` attribute for visibility toggling. The primary
+intent is provide a more semantic marker for *hiddenness*. As an added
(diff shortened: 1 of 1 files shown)


jamesarosen:
Realted Twitter thread: https://twitter.com/paul_irish/status/564577651609059329

jamesarosen:
https://github.com/emberjs/ember.js/issues/9634 has some more information, including some points on accessibility.

active/0000-toggle-visibility-api.md
```diff
@@ -0,0 +1,49 @@
...
+# Alternatives
+
+ 1. Clients could monkey-patch `applyAttributesToBuffer` and
+    `_toggleVisibility`. These methods are fairly long, though, and
+    would have to be copied wholesale.
+ 2. `Ember.View` could switch to `hidden` by default. Not all
+    browsers have sensible default behavior for this attribute, so
+    if Ember goes this route, it should probably also inject a
+    style like `[hidden] { display: none; }` that works welll,
+    but is easily overridden.
```
jamesarosen: I would prefer alternative 2, but I have two reservations:

 1. is it OK for Ember to inject `<style>[hidden] { display: none; }</style>` into the page on boot? Is that likely to break apps? Could it at least inject it *before* any other `<link>` or `<style>` tags so it's easy to override with simple CSS?
 2. are there apps or addons that rely on `el.prop('style').display` being `"none"` for hidden views? (As opposed to them just not being visible.)

tomdale:
We are trying to pare down the number of built-in APIs that abstract DOM features. In favor of this, we'd recommend people use an `{{#if}}` or bind this attribute using the new syntax described in #60.

PR #29, opened 2015-01-18
HTMLBars Component Attribute Types

zeppelin:
[Rendered](https://github.com/zeppelin/rfcs/blob/htmlbars-component-attribute-types/active/0000-htmlbars-component-attribute-types.md)

### DIFF
--- a/active/0000-htmlbars-component-attribute-types.md
+++ b/active/0000-htmlbars-component-attribute-types.md
@@ -1,0 +1,109 @@
+- Start Date: 2015-01-18
+- RFC PR:
+- Ember Issue:
+
+# Summary
+
+Allow HTMLBars components to use `number` and `boolean` literals for attribute
+values.
+
+# Motivation
+
+When using the new component syntax introduced in the
+`ember-htmlbars-component-generation` feature, it's not possible to pass
+literals other than strings to an attribute. With the current mustache component
+syntax this works fine: `num`, `bool`, `str` get converted to `number`,
+`boolean` and `string`, respectively:
+
+```hbs
+{{x-foo num=100 bool=false str="bar"}}
+```
+
+The new component syntax, however, only support `string` type, even though the
+values are unquoted.
+
+```hbs
+<x-foo num=100 bool=false str="bar"></x-foo>
+```
+
+The behavior of Handlebars is very convenient, most of the projects rely on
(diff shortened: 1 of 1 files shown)


mmun:
This should be fine. Given how the Handlebars parser works currently we can get `{{100}}` working but not `{{true}}`. I spoke with the Handlebars maintainer and he was OK with accepting a PR to accommodate the bool case.

mixonic:
Oh man that is exiting.

mmun:
This should be relatively straightforward to implement and will involve some minor changes to the Handlebars and HTMLBars parser. Anyone interested? cc @oneeman @marcioj

marcioj:
@mmun I would like to help. I will try to send something in this weekend

zeppelin:
So the resolution is to always use curlies ex. `attr={{true}}`?

I suspect this can cause some confusion, as people might wonder why `attr=false` is not working, not realizing they're in fact passing a string. Shouldn't there be at least a warning when attribute values are neither quoted, nor wrapped inside curlies?

mmun:
@zeppelin Ah I misunderstood the proposal. I think a warning is a good idea for unquoted literal tokens. The alternative, to simply rewrite it to {{true}} at parse times, feels bad because it would mean change the semantics of HTML.

zeppelin:
@mmun I'm fine with `{{true}}`. Even though it's slightly more ugly than just saying `true`, it's clear and unambiguous from Handlebars' perspective.

jonathanKingston:
Shouldn't the meaning of an attribute come from the component definition rather than it's usage?

That way a component:
```
<my-element truthy="1" number="1" />
```
Could have the following attributes:
```
{truthy: true, number: 1}
```

That is more of the direction I was going with #35 that the JavaScript has precedence over the template usage. I'm happy to expand my idea to allow for types.

mmun:
You can use `{{null}}`, `{{undefined}}`, `{{true}}`, `{{"foo"}}` (oh god why?!) and `{{3.14}}` now in HTMLBars. This should arrive in Ember in 1.13.

jonathanKingston:
@mmun as I mentioned in #35 has there been any thought around using arrays within the dom? This is pretty much required to duplicate the behaviour of `link-to` helpers in a component (which there is many use-cases for this).

PR #30, opened 2015-01-23
Don't update query params if there was no transition

drogus:
[Rendered](https://github.com/drogus/rfcs/blob/dont-update-query-params-if-there-was-no-transition/0000-dont-update-query-param-on-controller-if-there-was-no-transition.md)


wagenet:
This seems reasonable to me. Unfortunately, I don't know enough about router internals to know how easy it would be to make this happen.

trek:
This likely no longer applies for 2.0. Query params configuration is moving to the Route. I suspect in this particular use case you'd debounce the function that causes the transition?

PR #31, opened 2015-01-26
Ember data partial record loading

drogus:
This is a bit vague at the moment, but I wanted to put it out for discussion and I'll try to add more details as I get feedback

[Rendered](https://github.com/drogus/rfcs/blob/ember-data-partial-record-loading/0000-ember-data-partial-record-loading.md)

### DIFF
--- a/0000-ember-data-partial-record-loading.md
+++ b/0000-ember-data-partial-record-loading.md
@@ -1,0 +1,107 @@
+- Start Date: 2015-01-03
+- RFC PR:
+- Ember Issue:
+
+# Summary
+
+Partial record loading is a method of loading incomplete representation of the
+record and fetching the rest of the data when it's needed.
+
+# Motivation
+
+Partial record loading is useful for APIs that may return an incomplete
+representation of the record (for example due to performance optimisation).
+Let's take a `repository` resource which returns `last_build` as a part of a
+response:
+
+```json
+{
+  "repository": {
+    "id": 1,
+    "owner": "emberjs",
+    "name": "ember.js",
+    "last_build": {
+      "type": "build",
+      "id": 999,
+      "state": "passed",
+      "number": 333
+    }
+  }
+}
+```
+
+The `last_build` hash is an incomplete representation of a record of a `build`
+type. It can be beneficial to load such a record to the store right away and
+load the rest of the data when it's accessed.
+
+For example:
+
+```javascript
+store.push('build', repository['last_build'])
+var build = store.find('build', repository['last_build']['id'])
+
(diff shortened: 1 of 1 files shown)


knownasilya:
I really like this idea, especially since I make [model]Simple models just to help with performance. Usually this means returning an `id` and some kind of pretty identifier, e.g. `name`.

rkh:
:+1:

Partial entities allow nice API optimisations without having to create entity types just made for Ember.js (our current API is quite leaky in this regard unfortunately).

joostdevries:
@drogus I'd add a option to `DS.attr` which allows people to explicitly enable/disable this.

drogus:
@joostdevries could you elaborate? I don't understand what would be the benefit.

drogus:
I added a note about saving incomplete records.

joostdevries:
@drogus Right now you're suggesting to implement partial loading by default and allowing people to specify what properties might be missing from their API that should not trigger a server query. 

That approach makes this a breaking change because people with API's that somtimes omit properties will get additional server queries all of a sudden. 

If you'd instead add an option `{async:true}` to `DS.attr()` which defaults to `false`, people could opt-in to partial record loading and the change would not be breaking.

drogus:
> Right now you're suggesting to implement partial loading by default

I wouldn't turn it on by default, because of backwards compatibility, as you already mentioned. What's more, this can vary between the models, it may make sense for only part of your models, especially if you're working with an API, which you don't control. With that in mind, as soon as you enable partial record loading for a model (or all of the models), it should work automatically out of the box.

That said, I agree that there should be an API for a more granular control. I described that in the proposal: https://github.com/emberjs/rfcs/pull/31/files#diff-a99783a2dc2a593ac1cac37a386037c2R86, I believe that your use case (ie. whitelisting or blacklisting properties) is very similar to this one.

joostdevries:
:+1:

workmanw:
:+1: This would be so so so awesome for us. We have a noSQL backend and we have a lot of data denormalization. Right now we have "denormed" versions for all of our models. Having partial loading / entity projections would be a welcomed feature.

drogus:
I've updated the proposal with a bit more specifics, particularly the public API.

makmanalp:
:+1:

igorT:
@drogus and I refactored this a lot after EmberConf, and were working on an addon that would enable this. We should paste the updated version, and I will close this for now.

pangratz:
> @drogus and I refactored this a lot after EmberConf, and were working on an addon that would enable this

:heart_eyes: :+1:

PR #4, opened 2014-08-20
RFC: Ember Data url templates

amiel:
This is my third draft after feedback from @igorT.

There are still some details that need to get ironed out, but I thought it was about time to get community discussion on this.

Thanks!

(RFC content pasted in here by @machty)
- Start Date: 2014-08-20
- RFC PR:
- Ember Issue:
## Summary

`url-templates` improves the extesibility of API endpoint urls in `RESTAdapter`.
## Motivation

I think that Ember Data has a reputation for being hard to configure. I've often
heard it recommended to design the server API around what Ember Data expects.
Considering a lot of thought has gone in to the default RESTAdapter API, this
is sound advice. However, this is a false and damaging reputation. The adapter
and serializer pattern that Ember Data uses makes it incredibly extensible. The
barrier of entry is high though, and it's not obvious how to get the url you need
unless it's [a namespace](http://emberjs.com/guides/models/connecting-to-an-http-server/#toc_url-prefix)
or something [pathForType](http://emberjs.com/guides/models/customizing-adapters/#toc_path-customization)
can handle. Otherwise it's "override `buildURL`". `RESTSerializer` was recently
improved to make handling various JSON structures easier; it's time for url
configuration to be easy too.
## Detailed Design

`buildURL` and associated methods and properties will be moved to a mixin design
to handle url generation only. `buildURL` will use templates to generate a URL
instead of manually assembling parts. Simple usage example:

``` javascript
export default DS.RESTAdapter.extend({
  namespace: 'api/v1',
  urlTemplate: '{host}{/namespace}/posts{/id}'
});
```
### Resolving template segments

Each dynamic path segment will be resolved on a singleton object based on a `urlSegments`
object provided in the adapter. This `urlSegments` object will feel similar to defining
actions on routes and controllers.

``` javascript
// adapter
export default DS.RESTAdapter.extend({
  namespace: 'api/v1',
  pathTemplate: '{/:namespace}/posts{/:post_id}{/:category_name}{/:id}',

  pathSegments: {
    category_name: function(type, id, snapshot, requestType) {
      return _pathForCategory(snapshot.get('category'));
    }

    post_id: function(type, id, snapshot, requestType) {
      return snapshot.get('post.id');
    };
  }
});
```
#### Psuedo implementation

``` javascript
export default Adapter.extend({
  buildURL: function(type, id, record, requestType) {
    var template = this.compileTemplate(this.get('urlTemplate'));
    var templateResolver = this.templateResolverFor(type);
    var adapter = this;

[...]

PR #32, opened 2015-02-05
Ember-Data and non-CRUD endpoints

seanpdoyle:
Ember Data wraps typical CRUD actions nicely. However, it is lacking when it comes to endpoints that mutate the state of the world without expecting typical request bodies. We could develop an API that makes interacting with those endpoints easier.

### DIFF
--- a/active/0000-ember-data-actions.md
+++ b/active/0000-ember-data-actions.md
@@ -1,0 +1,90 @@
+- Start Date: 2015-02-04
+- RFC PR: (leave this empty)
+- Ember Issue: (leave this empty)
+
+# Summary
+
+Ember Data wraps typical CRUD actions nicely. However, it is lacking when it comes to endpoints that mutate the state of the world without expecting typical request bodies. We could develop an API that makes interacting with those endpoints easier.
+
+# Motivation
+
+To expose an API to interact with non-CRUD endpoints.
+
+# Detailed design
+
+Imagine that we have a commenting API.
+
+Administrators can curate comments, marking spammy or offensive records as spam.
+
+* `GET /comments` exposes the collection
+* `GET /comments/:id` exposes a particular comment
+* `POST /comments/:id/spam` marks a comment as spam
+
+What is the best way to interact with the `/comments/:id/spam` endpoint?
+
+http://emberjs.jsbin.com/befonozine/1/edit
+
+```javascript
+App.Comment = DS.Model.extend({
+  name: DS.attr("string"),
+  body: DS.attr("string"),
+  
+  reportSpam: function() {
+    var adapter = this.store.adapterFor("comment");
+    var url = adapter.buildURL("comment", this.get("id"), this);
+    
+    return adapter.ajax(url + "/spam", "post");
+  }
+});
+
+App.CommentAdapter = App.ApplicationAdapter.extend({
+  reportSpam: function(comment) {
+    var url = this.buildURL("comment", comment.get("id"), this);
+    
+    return this.ajax(url + "/spam", "post");
+  }
+});
+
+App.IndexRoute = Ember.Route.extend({
+  model: function() {
+    return this.store.find("comment");
+  },
+  
+  actions: {
+    reportSpam: function(comment) {
+      // first choice
+      comment.reportSpam().then(function() {
+        // do something
+      });
+      
+      // second choice
+      this.store.adapterFor("comment").reportSpam(comment).then(function() {
+        // do something
+      });
+    }
+  }
+});
+```
+
+
+This JS Bin explores some possible approaches.
+
+Neither are particularly pretty, and both might even depend on private / internal methods.
+
+# Drawbacks
+
+If you argue that `spam` should be a `DS.attr("boolean")` on the comment, and toggling that flag should be handled on the server, this RFC seems like a bad idea.
+
+From an API design standpoint, exposing an additional, focussed endpoint is easier and cleaner than monitoring the state and mutation of a flag on a record, buried in an otherwise conventional controller action.
+
+# Alternatives
+
+If we don't address this, interacting with non-CRUD endpoints could lead to heinous workarounds on the front-end, and/or heinous workarounds on the backend.
+
+# Unresolved questions
+
+What could we expose to make this easier?
+
+Should the model itself know about its URL?
+
+Should the store know how to translate special model functions to special API endpoints? Should the adapter? 


active/0000-ember-data-actions.md
```diff
@@ -0,0 +1,90 @@
...
+    
+    return adapter.ajax(url + "/spam", "post");
+  }
+});
+
+App.CommentAdapter = App.ApplicationAdapter.extend({
+  reportSpam: function(comment) {
+    var url = this.buildURL("comment", comment.get("id"), this);
+    
+    return this.ajax(url + "/spam", "post");
```
seanpdoyle: I didn't mention it in the RFC itself, which I'll have to fix, but I don't think `this.ajax` will work if the response body is empty (which could happen?)

MiguelMadero: If the server is returning JSON it should return an empty object (`{}`) or empty array depend on what makes more sense. An empty string isn't valid JSON that's why you're seeing the failure. There are arguments in favor and against using 204s to indicate an empty body response.

MiguelMadero:
I'll add two more options: 

1. At the instance level. 
2. At the adapter
3. At the store
4. At the class level. 


I think all operations at the instance level (1), should be performed by a function at the instance level. This is consistent with calling save, rollback, reload. I can think of actions like `reportSpam`, `favorite`, `upvote`, `etc`. We could try executing the same method at the class, adapter or store and give it the instance, but that seems inconsistent. 

For actions at the collection level, I like to think of the other 3 options. I think actions at this level are less common, specially if we don't have nested URL/Resources, so I can't think of any good example that would apply at a top level resource other than bulk updates, search or paging, but I think we have other good options for those already. 

I've my doubts about using the adapter (2), since normally we don't use it directly, but through the store. A possible API to do this at the store level. Instead of `this.store.adapterFor("comment").reportSpam(comment)` a possible APIs are 

```
this.store.action("comment", "reportSpam", comment)
// or
this.store.actions("comment").reportSpam(comment)
// or 
var reportSpam = this.store.actions("comment", "reportSpam")
reportSpam(comment);
```

The first one is consistent with find, filter, create, push, etc. They take the type as a first argument and operate on the collection, but it feels too loosey-goosey specifying the action name as a string and having the additional arguments after it. I'm not happy with the other 2. 

The 4th alternative is to re-open the model and so we can invoke the action at the class level. 

```
Comment.reopenClass({
  reportSpam: function () {...} 
})

import Comment from 'models/comment'
Comment.reportSpam(comment);
```

Even though I personally like the last option. We moved away from functions at the class level (find et.al use to be there in an old version and were left only at the store). In this case, an argument for *not* having them at the store is that they don't really modify or depend on the store, like find, fetch, etc.

workmanw:
We have this pattern dozens of times in out app. Everything from "like", to "spam", to other things involving transactional updates. Our implementation has three key parts. Adapter function that actually makes the ajax call. A `BaseModel` (which all models extend) which has a helper function that makes it easy to use. And lastly "helper functions" for the models that make to push and AJAX responses into the store and return models.

(1) Here is the Application adapter pattern for making the request.
```javascript
var ApplicationAdapter = DS.RESTAdapter.extend({
  ajaxForRecord: function(record, relPath, type, dataHash) {
    var url = [];
    var recordPath = this.pathForType(record.type);
    
    url.push(this.urlPrefix());
    url.push(recordPath);
    url.push(record.get('id'));
    url.push(relPath);
    url = url.join('/');

    return this.ajax(url, type, dataHash);
  }
});
```

(2) Here is the base model demonstrating the ajax helper method.
```javascript
var BaseModel = DS.Model.extend({
  ajax: function (path, type, data) {
    var adapter = this.store.adapterFor(this.constructor);
    return adapter.ajaxForRecord(this, path, type, { data: data });
});
```

(3) Here is one of several helper methods for processing ajax responses
```javascript
function pushPayloadAndFindRecord(store, typeKey, payload) {
  payload = extractPayload(typeKey, payload);
  var recordData = payload[typeKey] || payload[typeKey.pluralize()] || [];

  // Store the key
  var id = recordData[0].id;

  // Push the data
  store.pushPayload(payload);

  // Return the record promise
  return store.find(typeKey, id);
}
```

Lastly, here is how all the pieces fit together.
```javascript
var DocumentModel = BaseModel.extend({
  cloneDocument: function(retainPermissions) {
    return this.ajax('/clone', 'POST', { retainPermissions: retainPermissions }).then(function(resp) {
      pushPayloadAndFindRecord(this.store, 'document', resp.data);
    });
  } 
});
```

I certainly wouldn't argue that this approach is perfect. We've found having the ajax function helper function on the model makes the most sense because it allows helper functions like `cloneDocument` to easily do their job and it makes testing very easy. I'm sure the other options work just as nicely, just my $0.02.

knownasilya:
Would love something like this as well. For example, if you have a `user` model, and you want to download a csv of users, you can't do `this.store.csv('user')`.

Or if you have a user model, and you want to download that same model as a pdf, you might want to do:

```js
model.fetch('pdf')
  .then(data => console.log(data);
```

```hbs
{{! or get the url for easy download }}
<a {{bind-attr href=model.fetchUrls.pdf}}>Download PDF</a>
```

sandstrom:
I agree, there should be better support for these things. Two things that come to mind: *multiple end-points* for the same record and *non-CRUD actions*. 

**Multiple end-points for the same record**

We've got a few instances where records exist at multiple endpoints.

*Filtering encoded in endpoint path*

- `/archived`
- `/users/2/todos`
- `/pending`

may all return Todo records (current user, archived; given user; and pending respectively).

This could be reorganized into `/todos?filter=archived`, `/todos?user=2` etc. but I think this is common (the Github API has a few).

*Search results*

In search, say Elastic Search, the query may contain a json payload, but any updates/writes would go to another endpoint.

*Write/read and replication*

There may be a read-only replica with data [geographically] closer to the user. E.g. `eu.example.com/todos` may handle both read and writes, and `us.example.com/todos` may be read-only.

*Possible solution*

Perhaps having named endpoints in the adapter would work. With the default one being inferred from the record as it is now. Such that:

- `store.find('todo') => /todos`
- `store.find('todo', { endpoint: 'user', userId: 2 }) => /users/2/todos`

Naming is tentative, I'm not sure 'endpoint' is the ideal word here.

**Non-CRUD actions**

Don't know if this falls under the scope of this proposal, but sometimes non-CRUD actions have dedicated endpoints. Here is an example from the Stripe API. We've got similar ones in ours.

Our current solution is an out-of-band request, using pushPayload on the response, to update the model state. This works, but being able to define these actions on the adapter would be neat.

*Possible solution*

```
// in router
record.operation('refund');

// in adapter
var ChargeAdapter = RESTAdapter.extend({
  operations: {
    'refund' : { path: '/charges/:id/refund', method: 'put' }
  }
})
```

Naming is tentative, I'm not sure 'operation' is the ideal word here.

For example, the Strips API has a bunch of these (https://stripe.com/docs/api/curl#refund_charge).

davewasmer:
I built out an addon ([ember-data-actions](https://github.com/davewasmer/ember-data-actions)) to play around with this concept and different APIs.

The readme goes into depth, but short version is that it adds two types of APIs: resource level (i.e. `book.publish()` hits `/books/:id/publish`) and collection level (i.e. `store.action('book', 'published', {})` hits `/books/published`). The default behaviors are overridable on the model or adapter layer. Not a huge fan of my naming ("actions" is already a term in the Ember view layer), but this is just a proof of concept.

knownasilya:
@davewasmer very nice! I'll give it a try. Maybe you could use the template urls addon here https://github.com/emberjs/rfcs/pull/4#issuecomment-94130057

davewasmer:
good idea @knownasilya, I'll look into that. FYI, I haven't published the addon yet, since I haven't tested it at all, it was more of a code spike. But if people are interested, I could polish it up a bit and ship it.

knownasilya:
@davewasmer Is the code on github the latest? If so, I can install from github..

davewasmer:
Yep, that's latest. PRs welcome :wink:

MiguelMadero:
very nice



—

ppcano:
We have been using this pattern at the store level.

```
store.perform('cancel', subscription)
```

The [implementation](https://gist.github.com/ppcano/92e204f85f188e6fbfdd) is simple but I am in favour ED provides this feature out of the box.

amiel:
I like this idea. I often end up using a regular ajax call in a service or something like that when I need to make a call that does not fit with CRUD.

This could also integrate with #4 by calling `buildURL` with the name of the action in the `requestType` parameter.

BenjaminHorn:
@ppcano  I would like to play around with your implementation, but in line https://gist.github.com/ppcano/92e204f85f188e6fbfdd#file-store-js-L13 you have 'data' argument, which is not defined. 
thanks

ppcano:
@BenjaminHorn gist updated.

mike-north:
FWIW, there is an addon that aims to address a lot of the use cases in this RFC. https://github.com/mike-north/ember-api-actions

seanpdoyle:
Closing in favor of @mike-north's addon.

workmanw:
I know this is an old RFC, but I'm not sure it was worth closing. I personally feel this is something that does belong in ember-data core. Even though @mike-north has a great addon, so many people have this use case (myself included). JSONApi even offers this as mechanism in the form of links. 

I guess maybe another RFC would be in order then, IDK. I guess I would like to know what the ember data core team thinks about this concept.

rwjblue:
I also wish the addon supported collection based links a bit better (having to define collection level things on the model is not great IMO).  I do not view that as a deficiency of the addon, but of Ember Data not exposing a way to augment the collection object that will be used for a given relationship.

fivetanley:
> I do not view that as a deficiency of the addon, but of Ember Data not exposing a way to augment the collection object that will be used for a given relationship.

This is what references are for. If you can give me a dream/pseudocode API we could figure out how to implement them in references. Might help drive the documentation process on references in ember data  bit better too.

fivetanley:
I agree with @workmanw, there is definitely a need for this outside of an addon. Feels like a core addition.

PR #33, opened 2015-02-13
Remove records from record arrays only in deleted.saved state

drogus:
[Rendered](https://github.com/drogus/rfcs/blob/ember-data-deleted-records-disappear-from-record-arrays/0000-deleted-records-disappear-from-record-arrays.md)

### DIFF
--- a/0000-deleted-records-disappear-from-record-arrays.md
+++ b/0000-deleted-records-disappear-from-record-arrays.md
@@ -1,0 +1,46 @@
+- Start Date: (fill me in with today's date, YYYY-MM-DD)
+- RFC PR: (leave this empty)
+- Ember Issue: (leave this empty)
+
+# Summary
+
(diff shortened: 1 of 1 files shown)


alexspeller:
:+1: this is something I've hit before, it makes much more sense this way around

drogus:
I've talked about this with @igorT and it can be implemented now thanks to payload snapshots. Previously it could cause unexpected behaviour with relationships. Given post that has many comments, when a comment was deleted, but not persisted, keeping it in record arrays could result in sending the comment when a save was triggered on a post.

I'm not sure if this needs to be merged, because this is about Ember Data, not Ember.js itself.

runspired:
We now hide record from record arrays with a bit more clarity and it covers all the needs and desires expressed in this RFC: https://github.com/emberjs/data/blob/master/addon/-legacy-private/system/model/internal-model.js#L278-L293

Thank you so much for the time and the suggestion 🎉

PR #34, opened 2015-02-15
RFC: Default Container Values

jamesarosen:


active/0000-default-registrations.md
```diff
@@ -0,0 +1,107 @@
...
+looked up via `lookup` or `lookupFactory`), it couldn't be "freely" overridden.
+
+### Default Until Lookup
+
+Treat *all* entries as if they had `{ default: true }` from the previous section.
+Any entry can be freely overriden until `lookup` or `lookupFactory` is called on
+it.
+
+# Drawbacks
+
```
stefanpenner: This overlaps with the resolver based fallback system of ember-cli / module based apps. Is their a way we can utilize that flow instead of introducing something new?

jamesarosen: The `Registry` fallback system happens with a linked list of parent registries. If there were a registry *just for the addon* that Ember inserts in the chain, and if the addon could access its own registry somehow, we could do it there:

```js
export default {
  name: 'i18n',
  initialize: function(container, application) {
    const i18nRegistry = container.lookup('registry:ember-i18n');
    i18nRegistry.alias('i18n:plural-form', 'ember-i18n@util:plural-form');
  }
};
```
or possibly
```js
const PUBLIC_API = {
  'i18n:plural-form': 'ember-i18n@util:plural-form',
  'i18n:store': 'ember-i18n@store:main'
};

i18nRegistry.normalize = function(fullName) {
  return PUBLIC[fullName] || fullName;
};
```
Aesthetically, I like the idea of reusing `normalize`. As an addon author, I like the idea of a public API method on `Container` or `Registry` that handles aliasing for me.

bcardarella:
I don't agree with the nature of this RFC. It appears to promote the idea of hacking around namespaces. I think this can be very error prone.

You can get what you want *today* with addons. For example, if you have an addon asset called `ember-i18n/addon/plurals/en.js`

```js
import default Ember from 'ember';
export default Ember.Object.export({
  ...
});
```

if you expect this to be a class that the end-user is intended to override then the end-user *should* be using a class in the app namespace and not the addon namespace. So you should include an app wrapper class that is the default intended to be overridden `ember-i18n/app/plurals/en.js`

```js
import En from 'ember-i18n/plurals/en';

export default En.extend({
  ...
});
```

Now the end-user can simply provide their own version of this file in their own app: `my-app/app/plurals/en.js`

```js
import En from 'ember-i18n/plurals/en';

export default En.extend({
  ...
});
```

I have been using this pattern in several of my addons and it has worked well.

Alternatively, I am selectively falling back to the namespace as default lookup in ember-validations: https://github.com/dockyard/ember-validations/blob/master/addon/mixin.js#L43-L63 but in this case there is a more complex lookup algorithm than just app => addon.

jamesarosen:
> If you expect this to be a class that the end-user is intended to override then the end-user should be using a class in the app namespace and not the addon namespace.

I like this. It's at odds with something I was trying to do yesterday, but I can change course.

I've been trying to extract lots of pure functions into their own `utils/*.js` modules in Ember-I18n for better readability and testability -- for example, `compileTranslationTemplate`. I wanted the user to be able to override some of these default implementations. I came up with a few options:

 1. put them in `app/util/*.js` -- this seemed dangerous because of the possibility of name collision without a namespace
 2. put them in `app/util/i18n/*.js` -- a decent option; it reminds me of Rubygems' `lib/my_library_name/*.rb`
 3. put them in `addon/` and expose them as hanging off some object -- requires passing that object around to look up the implementation
 4. put them in `addon/` and use an initializer to inject them into the container under their public names -- what this PR would better enable

jamesarosen:
@bcardarella I think I've come around to exporting things in `app/` and looking them up via the container in my addon.

PR #36, opened 2015-02-24
RFC: Optional Segments in Routing

SlexAxton:
Hi Embereños!

[RFC](https://github.com/stripe/ember-rfcs/blob/stripe-optional-route-segments/active/0000-optional-route-segments.md)

We here at Stripe on the front-end team ( @michelle / @danwang / @slexaxton ) have been working on an RFC for implementing a feature we'd like in our app. We think it's fairly generic, but have hopefully explained our specific use-case as well.

We'd be happy to implement entirely or help implement this feature as a Pull Request (likely to a mix of Ember/RouteRecognizer/Router.js) if there is agreement that it should exist.

### DIFF
--- a/active/0000-optional-route-segments.md
+++ b/active/0000-optional-route-segments.md
@@ -1,0 +1,288 @@
+- Start Date: 2015-02-23
+- RFC PR: (leave this empty)
+- Ember Issue: (leave this empty)
+
+# Optional Segments in Ember routing
+
+## Summary
+Introduce optional segments to the routing specification so that unset parameters can map to URLs with missing segments.
+
+## Motivation
+### tl;dr
+For our application, there are a lot of routes/resources that can be accessed in two modes, default mode and test mode. We'd like the URLs to look like `/test/resource` for test mode and `/resource` for the default mode, which is where our users will spend the majority of their time.
+
+### Real World Stripe Example
+
+As a mechanism for showing why we might like something like this, we'd like you to consider our payments dashboard (which is being re-implemented in Ember). However, we feel the feature could be useful for many generic cases when specificity of top level hierarchies isn't necessary.
+
+It helps to think of the URL as a "selector" of hierarchical data. Because of this, CSS selectors are a good corollary. In our case, we'd like the option to avoid over-specifying the selector. `$('#foo .bar')`is not preferable to `$('.bar')` if `.bar` only exists inside of `#foo` -- however, it's extremely important to specify `#foo` if there are other `.bar` elements present in the DOM.
+
+#### The Stripe [User -> Business -> Mode] Hierarchy
+
+A Stripe user can have (but doesn't have to have) many businesses, and each business can view their data in test mode or live mode (two disjoint data sets). As good stewards of the web, we want users with the correct permissions to be able to directly link their teammates to a resource that is for a specific business in a specific mode.
+
+In the world where a user only has one business and one 'mode' of data, the urls map well to the Ember router system.
+
+`/payments/some_id`
+
+This would retrieve a specific payment from the user's one business in their only "mode." However, if we add the constraint that a user can have live-mode and test-mode datasets, we'd want to reflect that somewhere in the url. This could be a query parameter, but we think that's a little ugly for something so prevalent on our site, and doesn't map well to the data hierarchy (more on this later). Instead we'd like to optionally include the `test/` segment for test mode. This maps most closely to our "Boolean" handling of Optional Segments.
+
+`https://dashboard.stripe.com/payments/some_id`
+
+`https://dashboard.stripe.com/test/payments/some_id` 
+
+This works more nicely in our minds as a way of representing the url, and the hierarchy of the data. You might notice that we could also make this non-optional by always being explicit about the mode:
+
+`https://dashboard.stripe.com/live/payments/some_id`
+
+`https://dashboard.stripe.com/test/payments/some_id`
+
+But we also feel that the default case of 'live' data shouldn't need to be represented in the url. It's "over-specifying" in our opinion since it's not strictly necessary. We'd love to always strive for a 'minimum specific url'.
+
+Additionally, we'd like to handle the business identity. This maps more closely to our 'Dynamic' handling of Optional Segments. Some of our users have many businesses, but the long-tail of users has just one business. We'd like to not trouble users with a single business with an extra url parameter, but still allow deep linking for users with many.
+
+`https://dashboard.stripe.com/payments/some_id`
+
+This would be a valid 'minimum significant url' for a user with a single business in live mode, but in our most complex case (many businesses, many modes):
+
+`https://dashboard.stripe.com/skylight/test/payments/some_id`
+
+This would be the 'minimum significant url' for the 'skylight' business's test data for the payment with an id of `some_id`
+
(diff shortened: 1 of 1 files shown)


mde:
I was just asking @fivetanley about this on Twitter. (My frame of reference for this is the Rails router, and Barista (http://kieran.github.io/barista/), which is the router that @kieran pulled out of Geddy and made awesome.)

I was curious about the need to define a separate, singular resource route for single items vs. the collection (from the docs):

```javascript
App.Router.map(function() {
  this.resource('posts');
  this.resource('post', { path: '/post/:post_id' });
});

App.PostRoute = Ember.Route.extend({
  model: function(params) {
    return this.store.find('post', params.post_id);
  }
});
```
When I asked about optional dynamic segments, I guess was I was really asking about was 'optionally present' dynamic segments (in this case, the post_id). Seems weird to need two different base routes for the same model, but I guess this is because we have multiple controllers serving in lieu of an action/controller combo? Or is this just a convention thing?

In any case, truly optional dynamic segments seem pretty useful, but I'd much prefer the parenthesis to a trailing question mark. It visually implies that the thing is optional, and there's prior art.

kieran:
Yeah, I've found this to be a common sticking point in Ember. We've used Ember at Universe since the days of the old router, which in many ways I still kinda prefer.

In Ember, routes resolve 1-1 with states. In the old router, routes were actually extended states, adding the path property and a few methods for serialization / deserialization / etc. but the concept remains essentially unchanged: a route resolves to one state in a potentially nested state tree.

Consider the following example:

````
/posts
      /:id
          /comments
````
In Rails land, the router would resolve this by its endpoint (comments), passing the rest of the params (post_id) to the comments controllers index action. One controller is instantiated to handle the request, and one action is dispatched.

Now there are two ways we can go in Ember with a similar hierarchy.

### First (not idiomatic Ember):
````
/posts               <- posts state
      /:id           <- post state
          /comments  <- comments state
````
In Ember, each of these states would be resolved, entered and activated. 
- A posts index state, which would naively want to load a collection of posts
- A post show state, which would want to load the post
- A comments index state, which would naively want to load a *potentially unscoped* collection of comments

### Second (idiomatic-ish Ember):
````
/posts               <- posts state
/posts/:id           <- post state
          /comments  <- nested comments state
````
A hit to /posts in this case would activate and end in the posts index state. Totally normal behaviour so far.

A hit to /posts/123 would bypaass the first state tree, instead resolving directly to a shallow "post show" state. We now have the desired behaviour of simply loading the post with the relevant id.

We'll get a bit fancy and nest the comments state on the last line. This makes the comments index state a sub-state of the post show state, which allows us to access interrogate the post state for  params before loading the collection comments.

-----
There's obviously a bit more to it than that (considering nested views & outlets, etc...), but that's generally the reason ember routes are the way they are.

I'd personally prefer to see a differentiation between having entered a state (htough in a substate of that state) and being **in** a state itself (as a leaf node). I think that might be enough to let the controllers decide which cases warranted model or collection loads.

----
btw @mde that reminds me, I have updates to Barista (including named generators & new docs) I need to get pushed - will be in touch with a PR :-)

**Comment by mde** [comment]
Sorry to hijack this thread with my question -- @kieran, that was good context. I still wonder why we have the singular/plural split in documentation for the resource routes -- is that just convention?  Is there a better place to have this discussion? :)

**Comment by kieran** [comment]
I think the gist of it is around trying to maintain a clear mapping between states and routes, while pulling the route definitions out into a separate file.

There are quite a few discussions on http://discuss.emberjs.com around nested routes, probably with much better explanations than I can offer ;-)

**Comment by mde** [comment]
:+1:

**Comment by alexspeller** [comment]
@kieran seems like you can accomplish that by leveraging index routes
@mde singular routes are for showing one item, plural for multiple. In general. I mean, they are just names, you can show whatever you want on a route.

However, all of these comments seem to have missed the point of the rfc, which doesn't actually do what anyone has asked, and only makes dynamic segments optional on a single route. Which seems like it makes sense to me

**Comment by alexspeller** [comment]
Thinking about this, I don't think the idea of having the segment be a boolean makes sense at all - it seems very tied to Stripe's specific use case. It would be a lot better if you could access the value as with other dynamic segments. A good use case would be localization:

```
/posts/1
/fr/posts/1
/ru/posts/1
```

michelle:
@alexspeller Thanks for taking a look!

Yeah, the boolean option definitely has more limited usefulness than the other option, "Treat Optional Segments as Dynamic Segments when present." In fact, for the more complex use case Stripe has that Alex described (with business names), we'd also need "optional dynamic segments" rather than "optional boolean segments."

wycats:
I really like the syntax we used in Merb and Rails 3: `/foo(/bar)(/:baz)`. In order words, the parentheses mark optional segments. One benefit is that you can nest them.

SlexAxton:
I think I like the power that that syntax adds re: nesting. It does make it not-insignificantly more complex to parse (not-regexable anymore, etc), but I don't think that's necessarily a blocker pending some perf gut-checks.

kieran:
You can still validate nested routes with a regex, but the actual parsing / generation is a touch more complicated.

You can see an example of the Merb / Rails syntax here: http://kieran.github.io/barista/#keys

Example route: `/:controller(/:action(/:id))(.:format)`
Regex that validates the route: `/^(/([a-zA-Z_][\w\-]*)(/([a-zA-Z_][\w\-]*)(/([\w\-\s]+))?)?(\.([\w\-\s]+))?)(\?.*)?$/`

If the regex matches, it will then attempt to parse it properly using http://kieran.github.io/barista/docs/route.html#section-49 though it subs out to specific parse methods per part type.

Of course this is not directly compatible with client-side routes, since HTTP methods are assumed.

SlexAxton:
Ah great. I assumed the optimization here could usually just be `if (route.indexOf('(') >= 0) ...` then we'd go into "parser" mode or something. Essentially avoid anything complex unless we know we'd need it.

It seems like Batista is doing more or less the same thing but better.

kieran:
Yeah, essentially I lean on the native regex implementation to take care of that kind of optimization. I figure it's faster / more battle-tested than any hand-rolled opts on my part would be.

SlexAxton:
Hey @wycats - was wondering if you guys had a chance to discuss this at one of the team meetings recently? We're just eager to get started working on if it's something folks are interested in.

tomdale:
We discussed this at the core team meeting and people were generally :+1:. @wycats only request is that we add an optional parens syntax, as you may want to group multiple optional segments.

SlexAxton:
Great. We'll get working on an implementation and attempt the parens syntax with things defaulting to working like Merb and Rails 3 where relevant.

/cc @michelle @danwang

wycats:
@SlexAxton I think @tomdale was a little unclear. I'd prefer *just* the parentheses syntax. Do you feel very strongly about `?` (I don't like it both because of grouping issues and because `?` already has meaning in URLs).

SlexAxton:
I got what he meant I think. We're moving forward with parens.

While I have you though, do you feel strongly of we also implemented the "constraints" stuff from rails/merb that helps reduce ambiguity? (As a separate but very related PR?)

wycats:
@SlexAxton I would love to see an RFC. In particular, what would the constraint matcher function get

SlexAxton:
Kk. We'll go one at a time. Will rfc some constraints when this is further along.

tomdale:
@SlexAxton What's the status of this?

knownasilya:
Would love for this to move forward as well :+1:

wagenet:
I don't love closing down an active ticket, but nothing has happened here in 6.5 years! If there's still interest in moving this forward, please let me know and I'll do what I can to help it progress.

It's also worth noting that some new router explorations are being done for Polaris which may help address these concerns.

PR #37, opened 2015-02-27
Element and Fragment RFC

ef4:
[RFC](https://github.com/ef4/rfcs/blob/element-and-fragment/active/0000-element-and-fragment.md)

### DIFF
--- a/active/0000-element-and-fragment.md
+++ b/active/0000-element-and-fragment.md
@@ -34,8 +34,8 @@ template **must** be defined with the `<element>` keyword.
 If you place this template in `my-component.hbs`:
 
 ````handlebars
-<element class="special {{class}}" data-my-id={{id}}>
-  Hello {{name}}.
+<element class="special {{attrs.class}}" data-my-id={{attrs.id}}>
+  Hello {{attrs.name}}.
 </element>
 ````
 
@@ -48,7 +48,7 @@ It can be invoked like this:
 And will result in this output:
 
 ````html
-<div class="special magical" data-my-id="1">
+<div class="special magical" data-my-id="1" id="1" name="Tomster">
   Hello Tomster.
 </div>
 ````
@@ -69,6 +69,57 @@ The output changes to
   Hello Tomster.
 </span>
 ````
+
+## `<element>` HTML Attribute Semantics in Detail
+
+There are three places that may legitimately need to set a component's
+HTML attributes:
+
+ 1. The caller of a component may intend one of more of the provided
+    attributes to go directly onto the HTML element:
+
+        <my-component title="Tomster" model=hamster>
+
+    Every attribute passed into a component in this way gets set on
+    the component's `attrs` object, which is how components receive
+    arguments in general. Any attribute with a primitive value
+    (string, number, boolean) *also* gets set by default as an HTML
+    attribute on the component's element. So in the example above,
+    `title="Tomster"` will appear by default on the component's
+    element, but `model` will not, assuming `model` is an `object`.
+
+ 2. The component itself may set its own HTML attributes:
+
+        <element title="Mr {{attrs.title}}">
+
+    This is logically equivalent to calling `setAttribute` directly on
+    the component's element in the `didInsertElement` hook, and
+    setting up a corresponding observer to keep the value up to
+    date. **It takes precedence over any values set by the caller.**
+
+    It is important to understand that `<element>` only sets HTML
+    attributes, it does not set `attrs`. So the above example leaves
+    the original caller provided `attrs.title` alone, but overrides
+    the `title` HTML attribute to have a modified value.
+
+ 3. The component's parent class may *also* have attribute settings in
+    its `<element>` that should be inherited. A child components takes
+    precedence over its parent component's attributes.
+
+So we can summarize the complete semantics as this series of logical
+steps:
+
+ 1. First, any caller-provided attributes with primitive values are
+    set as HTML attributes.
+
+ 2. Next, the component asks its parent class to set HTML attributes
+    based on what appears in the parent's `<element>`. This recurses
+    back so that the most primitive ancestor is setting first, and
+    descendants have the chance to override their ancestors.
+
+ 3. Finally, the component sets attributes based on its own
+    `<element>`.
+
 
 ## `<fragment>` semantics
 
@@ -107,28 +158,7 @@ with.
 
 ## What happens to `attributeBindings`, `classNames` and `classNameBindings`
 
-In the vast majority of cases, users won't need these anymore, and
-will be able to get the same control directly from their template
-using `<element>`.
-
-However, we will still keep them to handle trickier inheritance
-situations, where a base class or mixin needs to control element
-attributes in all derived classes. Only library authors are likely to
-need to know about them, and most users should be steered toward using
-`<element>` instead.
-
-We can also put:
-
-````js
-  attributeBindings: ['tagName', 'class']
-````
-in the base `Component` implementation so that these properties will
-propagate from the caller to the `<element>` by default.
-
-Any attributes set directly on `<element>` in a component's definition
-**take precedence over** `attributeBindings`. We make a special
-exception for `class`: `classNameBindings` and `classNames` will be
-merged with any value for `class` provided on the `<element>`. 
+They get deprecated.
 
 ## Fragment names
 
@@ -168,11 +198,6 @@ it can define behaviors common to both.
 
 Adds a tiny amount of arguable boilerplate to templates that was not
 there before.
-
-This proposal doesn't eliminate `attributeBindings`,
-`classNameBindings`, and `classNames`. I think it would be a good idea
-to eliminate them, but we need a detailed design for some other way to
-do inheritance -- one idea would be template macro expansion.
 
 # Alternatives
 
--- a/active/0000-element-and-fragment.md
+++ b/active/0000-element-and-fragment.md
@@ -1,0 +1,198 @@
+- Start Date: 2015-02-27
+- RFC PR:
+- Ember Issue:
+
+# Summary
+
+Introduce `Ember.Fragment` and the `<element>` and `<fragment>`
+template keywords.
+
+# Motivation
+
+This RFC addresses two key problems:
+
+ - Ember has no clear concept of a "tagless component". We need such a
+   concept now that we're moving to a component-centric world.
+
+ - Today users are forced to learn two separate APIs for setting
+   element attributes, one that works only for a component's top-level
+   element (and cannot be used from inside templates), and one that
+   works everywhere else.
+
+# Detailed design
+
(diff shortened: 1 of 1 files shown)


mmun:
I like this API because it can be implemented under a feature flag pretty cleanly with the presence of `<element>` opting you in.

Is there a use case for components without templates? I'm thinking no. Things like `{{#each}}` are handled at the helper level of abstraction. There's a tug-of-war between the in-template APIs and in-class APIs for components. With the <element> tag I agree that we can likely do away with the in-class APIs (classNameBindings, attributeBindings, elementId, tagName, etc).

Since <element> is new API we can disable these in-class APIs. I think it would be a shame not to take that opportunity.

ef4:
The main issue with getting rid of `classNameBindings` and friends is inheritance. For example, we would need a good way to do the kind of things that Ember's TextSupport Mixin does using `attributeBindings`, without forcing every component that mixes it in to copy all the attributes into their own template's `<element>`.

There are several possibilities for this. One would be to make attributeBindings be an argument to `<element>`, so you can say `<element attributeBindings=textSupportAttributes>`. Another would be a compile-time macro system so you can put `<text-support-element>` in your template and it would expand to `<element autocapitalize={{autocapitalize}} autocorrect={{autocorrect}} ... >`.

mmun:
Alternatively, if each component embeds it's template (rather than being lookedup, which adds complications because it requires a name) AND templates exposed metadata related to `element`, we could implement "template" inheritance.

mmun:
I like the potential of

```js
export Ember.Component.extend(TextSupport, {
  template: hbs`<element class="ember-text-field" tagName="input">`
})
```

ef4:
I don't quite see how that's different than just keeping attributeBindings. What part of TextSupport are we inheriting? It's template? If so, how does that fit with our template?

active/0000-element-and-fragment.md
```diff
@@ -0,0 +1,198 @@
...
+
+Introduce `Ember.Fragment` and the `<element>` and `<fragment>`
+template keywords.
+
+# Motivation
+
+This RFC addresses two key problems:
+
+ - Ember has no clear concept of a "tagless component". We need such a
+   concept now that we're moving to a component-centric world.
```
stefanpenner: can we explain why this is neded?

mmun:
I don't see the point of attribute bindings at all, at least in the one-way case. We shouldn't need to specify that something is bound to a DOM attribute - it should always happen. `{{input placeholder=foo}}` should "just work" without having to specify `'placeholder'` somewhere.

The mutable attribute case should be fairly rare. For example, updating a class-name of a component based on its current state. I believe react completely punts on this (you'd have to manually mutate the DOM).

I just think of TextSupport as mixing event handlers to fire actions.

wagenet:
@mmun See https://github.com/emberjs/ember.js/blob/v1.10.0/packages/ember-views/lib/mixins/text_support.js#L108. It seems very useful to have that behavior be inheritable.

mmun:
Nowhere am I suggesting that the behaviour not be inheritable. I am saying one way attribute bindings should work without explicitly listing them.

active/0000-element-and-fragment.md
rwjblue: @stefanpenner - We need this because we use this concept internally. There are cases where you might want to implement things like our internal `{{each}}` and `{{if}}` helpers as components, and you wouldn't need/want there to have to be an extra element.

wagenet:
@mmun but how do you compose that? Are you suggesting a way in which templates can be combined?

active/0000-element-and-fragment.md
```diff
@@ -0,0 +1,198 @@
...
+<div class="special magical" data-my-id="1">
+  Hello Tomster.
+</div>
+````
+
+`<element>` accepts an optional `tagName`, so if we changed the
+template to:
+
+````handlebars
+<element class="special {{class}}" data-my-id={{id}} tagName="span">
```
stefanpenner: is `tagName` bindable?

active/0000-element-and-fragment.md
```diff
@@ -0,0 +1,198 @@
...
+
+
+## Why are `<element>` and `<fragment>` mandatory?
+
+If we want to distinguish between Component templates and Fragment
+templates, at least one of them needs to be mandatory. It makes sense
+for `<element>` to be the mandatory one, since `<element>` does useful
+work and represents a real DOM element.
+
+`<fragment>` is only needed due to upgrade-compatibility
```
stefanpenner: should we propose a timeline for this, or is it pre-mature.

active/0000-element-and-fragment.md
rwjblue: It should not be.

active/0000-element-and-fragment.md
```diff
@@ -0,0 +1,198 @@
...
+definition from template invocation.
+
+We could avoid introducing `<fragment>` and treat any template without
+an `<element>` as a fragment, but this makes the upgrade painful.
+
+We could detect whether a template has a single top-level element or
+not and automatically make it by a Component or Fragment. This idea
+was rejected because it's a footgun -- for example, accidentally
+changing your template from one kind to the other would change the
+sematnics of your `$` method.
```
stefanpenner: typo: sematnics -> semantics

active/0000-element-and-fragment.md
```diff
@@ -0,0 +1,198 @@
...
+
+# Unresolved questions
+
+ - Should `Ember.Component` extend `Ember.Fragment`?
```
stefanpenner: yes, I believe so.

active/0000-element-and-fragment.md
```diff
@@ -0,0 +1,198 @@
...
+
+ - Should `Ember.Fragment` have a `$` method that matches its DOM
```
stefanpenner: as its dom range is dynamic, I don't believe so. Imagine the following scenario, `this.$()` can easily become stale, and we have to mechanism by which to update it.

```hbs
<fragment>
  <h1>asdf</h1>
  {{#if theSkyIsFalling}}
    <p>Run!!</p>
  {{/if}}
</fragment>
```

stefanpenner:
further clarification:

both:

`<element></element>` and `<fragment></fragment>` must be top level, with no siblings.

For example, the following:
```hbs
<h1> something </h1>
<element> ...</element>
```

should raise

wagenet:
As it turns out, every attribute binding declared in `TextSupport` could just be explicit as part of the `<element>`. To make the issue more clear, here's an example I put together: https://gist.github.com/wagenet/989b759848e8ee8043a8. If it wasn't for inheritance, I'd be happy to ditch the whole notion of attribute bindings. However, without some sort of equivalent we do run into the problem I demonstrate.

mmun:
Yeah, that's a good example to keep in mind!

drogus:
I really like the first part of the proposal (ie. `Ember.Fragment` and `Ember.Component` distinction), but I'm not so sure about explicit definition of `<element>` or `<fragment>`. I agree that having 2 APIs for the same thing may be confusing, but as other commenters noted, we will still be left with 2 APIs.

A possible solution would be to allow to "merge" inherited attributes in a template, although I'm not sure if it's technically doable, because I don't know HTMLBars too well. It could look like:

```javascript
var FooComponent = Ember.Component.extend(),
    BarComponent = FooComponent.extend();
```

```hbs
{{! templates/components/foo-component.hbs }}
<element class="foo">Something</element>
```

```hbs
{{! templates/components/bar-component.hbs }}
<element class="bar" {{super-attributes}}>{{super-content}}</element>
```

Using `{{bar-component}}` would render `<div class="foo bar">Something</div>`

ef4:
I incorporated feedback on how to handle HTML attributes in a way that lets us deprecate attributeBindings and friends, while still handling inheritance.

jerel:
At first I disliked the idea of `<element>` and `<fragment>` in templates but the more I ponder it the more I think that having the element defined in the template instead of in JS makes components easier to reason about. I'm :+1: 

There's another use case that I think `<fragment>` can solve nicely: integrating third party libraries with the Ember component structure and actions. I've been doing this in my own apps but the current incarnation of components doesn't make it ideal (even with `tagName` set to `''` a component still goes through the render steps and outputs a textNode) and performance is OK but not great. Future example:

A `map` component template that doesn't create any HTML itself:

    <fragment>
        {{#each location in locations}}
            {{#if location.onMap}}
                {{marker latLng=location map=map}}
            {{/if}}
        {{/each}}
    </fragment>


And the `marker` component:

        click: function() {
            this.sendAction('openPopup', this.marker);
        },
        render: function() {
            var map = this.get('map'),
                latLng = this.get('latLng');
            this.marker = this.map.addMarker([latLng[0], latLng[1]]).addTo(map);
            this.marker.on('click', this.click());
        },
        destroy: function() {
            this.marker.off('click');
            this.get('map').removeMarker(this.marker);
        },
    });

Hope this isn't off-topic for this RFC but figured I'd add it so you can see how it might be used/abused.

drogus:
@ef4 do you think that ability to use parent's template content would be beneficial? An API for overriding parent's attributes looks good, but I still don't see anything for content. I'm not sure how important that is, but it was one of the things that I remember using, for example to add an attribute to a component, but still leave the parent's template.

Also, it's still missing how could mixins be used without `attributeBindings`. Here, too, I'm not sure how important that is, but probably without covering this case `attributeBindings API shouldn't be deprecated.

mixonic:
Great! It took some thinking, but I understand why the both the two classes and two markups are needed to manage backwards compat. Very good.

### top-level element tag

Stef says:

> both <element></element> and <fragment></fragment> must be top level, with no siblings.

This follows how component layouts work today. Seems correct in all ways.

ernesto-jimenez:
This RFC came to mind while watching the video from @ebryn's [minitalk in EmberConf][minitalk] about [ember-component-css][ember-component-css].

The add-on allows you to drop component-specific CSS in a `styles.css` in a pod along with `template.hbs` and `component.js`, then it automatically adds a namespace class to the component's root tag and namespaces all the CSS.

In order to style the component's root tag it uses this syntax:
```css
& {
  padding: 2px;
}
```

If this RFC gets implemented, `&` could be changed to `element` like this:
```
element {
  padding: 2px;
}
```

I'm just commenting it here because it seems the RFC introduces new conventions that would help namespacing component's CSS in the future, which is a problem apps will run into as more components get shared between apps.

[minitalk]: https://www.youtube.com/watch?v=K4fyHxg3Ldk&feature=youtu.be&t=33m00s
[ember-component-css]: https://github.com/ebryn/ember-component-css

ebryn:
@ernesto-jimenez Yes, my `ember-component-css` project is built with routable components/pods in mind!

I"m going to be working on the implementation of routable components. I've actually got some hacks available for use right now in an addon: https://github.com/ebryn/ember-routable-components

tomdale:
Superseded by #60 (sorry @ef4)

PR #38, opened 2015-02-27
Routable Components RFC

ef4:
[RFC](https://github.com/ef4/rfcs/blob/routeable-components/active/0000-routeable-components.md)

### DIFF
--- a/active/0000-routeable-components.md
+++ b/active/0000-routeable-components.md
@@ -1,0 +1,488 @@
+- Start Date: 2015-02-27
+- RFC PR:
+- Ember Issue:
+
+# Summary
+
+Eliminates Controllers.
+
+# Motivation
+
+As Ember has evolved toward a component-centric model, controllers
+have become redundant and vestigal. This unnecessarily burdens
+developers with learning an additional concept and forces them to make
+unnecessary choices.
+
+The goal of this design is to describe how to eliminate
+`Ember.Controller` while maintaining sufficiently rich APIs to do all
+the important things controllers do.
+
+# Scope
+
+- Eliminate `Controller` by designing alternative APIs for routing to
+  a component, maintaining long-lived non-URL state, and using query
+  parameters.
+
+- Simplify the behavior of `Route`'s model hooks. While this is not
+  strictly needed to achieve the goal of eliminating controllers, it
+  makes sense to address it while we're redesigning the API between
+  routes & components anyway.
+
+- Consider how routeable components relate to asynchronous components,
+  because they may share concepts or implementation.
+
+- Clarify the distinction between Components and Fragments ("tagless
+  components").
+
+# Definitions
+
+ - **routeable component** - a component whose identity and attributes
+   are determined from the URL at runtime. Routeable components are
+   what go inside `{{outlet}}`s.
+
+ - **asynchronous component** - a component with built-in support for
+   asynchronously loading data, with a friendly API for reflecting
+   loading and errors states.
+
+Both of the above are conceptual definitions for the purposes of this
+design document. They are not necessarily actual classes or concepts
+that would need to be taught to new Ember devs.
+
+# Detailed design
+
+## Routing to a component
+
+Our current API for letting a `Route` decide which things to render
+and where they should go is fairly rich and complex. We can simplify
+it significantly.
+
+### Choosing what component(s) to render
+
+The current `Route.render` method is where a `Route` expresses what
+should actually get rendered. It has a fair bit of complexity that we
+can deprecate. Today you can say things like:
+
+````js
+this.render('post', {
+  view: 'special-post',
+  controller: 'other-controller' ,
+  model: myPost
+})
+````
+
+Which lets you pick and choose which template, view, controller, and
+model to render. There are fairly rich semantics for choosing defaults
+for each of those things, and allowing views to choose their own
+template, etc.
+
+Instead, we can deprecate the use of the `view`, `controller`, and
+`model` arguments, and accept a name which will be used to lookup a
+*component*, not a template. We will also accept arbitrary attributes
+to set on the component.
+
+````js
+this.render('post', attributes)
+````
+
+The current semantics of `into` and `outlet` would still be preserved:
+
+````js
+this.render('choices', attributes, { into: 'application', outlet: 'sidebar' })
+````
+
+We can maintain transitional compatibility by failing to detect a
+component and looking for a template instead, using the old behavior
+for that case.
+
+Nested child routes have names like 'parent-route/child-route'. Their
+corresponding component would also be named
+'parent-route/child-route', implying that we must support component
+resolution in a non-flat namespace.
+
+Since standalone route templates are going away in favor of
+components, the current distinction between the `templates` and
+`templates/components` directories becomes unhelpful. We want to put
+all of them into just `templates`. Thanks to the introduction of
+`<element>` and `<fragment>` (see the
+[Element and Fragment RFC](element-and-fragment)), we can make this
+transition gracefully.
+
+  - A template in `templates` with an `<element>` or `<fragment>` tag is defining a Component or Fragment.
+  - A template in `templates` with neither is a legacy standalone template.
+  - A template in `templates/components` with neither is a legacy component.
+
+### Specifying component attributes
+
+We will introduce a new route hook named `attributes`. It receives the
+route's positional parameters (the same ones that are currently passed
+to the `model` hook), plus any query parameters defined for the route,
+and must return a POJO. The POJO's properties may be promises, which
+will all be resolved in the same way we currently resolve the `model`
+hook (including invoking `loading` actions as needed). The default
+`attributes` implementation is:
+
+````js
+attributes: function(params, queryParams) {
+    return {
+        model: this.model(params)
+    }
+}
+````
+
+(Possible variation for this default implementation: any queryParams
+that are present get stuffed directly into the attributes as well,
+alongside `model`.)
+
+Unlike today's `model` hook, `attributes` gets called *every* time
+your route is entered, whether or not `transitionTo` was passed a
+preexisting model. This eliminates the need for `beforeModel` and
+`afterModel`, because you can implement either behavior from within
+your own `attributes` or `model` functions. Therefore, `beforeModel`
+and `afterModel` are deprecated.
+
+This also means we deprecate `refresh`, because routes essentially
+always refresh. 
+
+### When Routes can call render
+
+Today, `render` is normally called from the `renderTemplate` hook, but
+it can also be called from action handlers on the route. Calling
+`render` from action handlers -- outside of a route transition --
+makes routes unnecessarily stateful and introduces an additional way
+to do things that are better off done directly from within components
+and their templates. We will therefore deprecate calling `render`
+outside of transitions. See Appendix A below for an example of
+updating this old behavior.
+
+Today, `renderTemplate` receives the default controller and the
+model. But since we don't intend to have controllers anymore, we will
+deprecate `renderTemplate` in favor of `renderComponents`, which would
+have a default implementation like:
+
+````js
+renderComponents: function(attributes) {
+    this.render(this.routeName, attributes);
+}
+````
+
+The `attributes` argument comes from the new `attributes` hook as
+described in the previous section. It is resolved with `RSVP.hash`
+before passing the result to `renderComponents`.
+
+We also deprecate `setupController`, because there is no longer a
+controller to setup -- all state is passed to the component(s) in
+`renderComponents` instead.
+
+### Idempotence
+
+Your `attributes` and `renderComponents` hooks are required to be
+idempotent (and therefore `model`, which is called from the default
+implementation of `attributes`, must also be idempotent if you're
+using it with the default `attributes` implementation). This keeps
+routes truly stateless and easy to understand. You should not rely on
+knowing exactly when or how often a route will be re-invoked, leaving
+it up to Ember to optimize.
+
+For cases where you really need to do something different on initial
+render, the existing `activate` action will still reliably fire only
+when a route is first entered, after Ember invokes your `attributes`
+method.
+
+We will add a complementary `update` action that fires after your
+`attributes` hook only when the route has not been initially entered.
+
+
+### Loading and error states
+
+`Routes` will still fire the `loading` and `error` actions as they do
+now.
+
+The default handlers for those actions currently look for
+correspondingly-named templates. We would change them to look for
+correspondingly-named components instead. For example, if before you
+had a template named `'people/loading'`, you would instead have a
+component named `'people/loading'`.
+
+This is actually a fairly small change, since adding a template is
+sufficient to define a component. But it unifies the API, and gives
+you the full capability of components for implementing these
+substates.
+
+### Sending actions from component up to route
+
+Components are deliberately isolated and don't send arbitrary actions
+upward unless you explicitly tell them to. Routeable components
+preserve this same behavior.
+
+Therefore, if you want to send an action from a component up to the
+Route that rendered it, you should pass an action handler function as
+one of your component's attributes.
+
+### Fragments: "Tagless Components"
+
+Today, when you render a template into an outlet, neither the template
+nor the outlet introduces extra DOM elements. We want to be able to do
+the same thing with routeable components, which means we need to
+officially support "tagless components".
+
+But instead of having two flavors of components (tagged and tagless)
+with subtly different behavior, we will introduce a new concept:
+`Ember.Fragment`. A Fragment is a "tagless component". It still
+receives attributes and can maintain internal state, and it renders a
+template, but it does not have a top-level element and it may generate
+zero, one, or more DOM elements.
+
+`Fragments` have no `element` property. They *do* have a `$` property,
+but it represents a DOM range, not a single element.
+
+Everywhere else in this document where I say "component", I really
+mean `Component` *or* `Fragment`.
+
+To distinguish `Component` templates from `Fragment` templates, we
+will introduce the `<element>` and `<fragment>` keywords, which are
+described in a [separate RFC document](element-and-fragment).
+
+## Query Parameters
+
+Today, query parameters live on controllers, which means we need to
+find them a new home. Neither Routes nor Components are intended to be
+stateful, so we will not bind the values of queryParams directly to
+either of those.
+
+queryParams will be *declared and configured* on Routes, with the same
+API we currently have on Controllers except for two exceptions. The
+first exception is default values -- a query parameter like this:
+
+````js
+Ember.Controller.extend({
+  queryParams: {
+    foo: { as: 'bar' }
+  }
+  bar: 0
+});
+````
+
+would look like this on a Route:
+
+````js
+  queryParams: {
+    foo: { as: 'bar', default: 0 }
+  }
+````
+
+because the Route itself does not a `bar` property.
+
+The second difference is that we no longer need a `scope`
+option. queryParam stickyness is not managed at the Route level, it's
+managed inside components using sessions (which have arbitrarily
+controllable scopes). More on this below under "Stickyness".
+
+### Data down
+
+When query parameters are changed, their values will arrive in the
+queryParams argument to the `attributes` hook. Note that this implies
+you will now always be opting in to a "full transition" as described
+in the current Guides. This should be ok because each layer is
+supposed to be smart about rerendering only things that really need it
+-- the change will propagate down through the component hierarchy,
+triggering diffing at each step.
+
+### Actions up
+
+To change queryParams, the `Route` will call
+`transitionTo`. Components will not know anything about queryParams as
+such -- they are just like any other `attributes`. A component that
+needs to change a queryParam will use an action handler to communicate
+upward to the route as described in the section titled "Sending
+actions from component to route".
+
+We can provide a default implementation of a mutable attribute that
+can be passed to the component whose setter will trigger the
+`transitionTo`. For example:
+
+````js
+attributes: (params, queryParams) {
+  model: this.model(params),
+  filterBy: mut(queryParams, 'filterBy')
+}
+````
+
+would give the components a mutable attribute that behaves the same as
+if the component was invoked directly from a template with:
+
+````handlebars
+<the-component model=model filterBy={{mut queryParams.filterBy}}>
+````
+
+### Default values
+
+queryParams have default values, which determine when they should
+appear in the URL and determine what formatting rules will apply (for
+example, a default integer value will cause future values to be parsed
+as ints). This behavior can remain the same -- it will just be
+configured on Routes instead of Controllers.
+
+
+### pushState vs replaceState
+
+This can also remain the same. When a Route sets a queryParam it will
+use either pushState or replaceState based on the configuration for
+that parameter.
+
+### Stickyness
+
+Sticky query params are a specific case of a more general need for
+statefulness. Without singleton controllers, this kind of state will
+be managed by session services.
+
+Components can use session services to store and retrieve
+arbitrarily-scoped state. They can explicitly push that state upward
+into the queryParams when they want to.
+
+If you transition into a route that causes a component to render, and
+that component has session state for the value of a `mut` attribute,
+the component is free to set the `mut` attribute, propagating the
+change upward. That change may propagate all the way to the route and
+cause a queryParam to be set.
+
+## Idempotent Model Hooks
+
+Today we have `beforeModel`, `model`, and `afterModel` hooks on
+`Route`. The distinction between them, and the timing of when they're
+called, is complicated by the original goal of avoiding unnecssary
+reloading of models. If you `transitionTo('person', 1)` (or directly
+set the corresponding URL), all three hooks will fire. But you can
+also `transitionTo('person', withThisPersonModel)`, and the `model`
+hook will not fire, since a model is already available. `beforeModel`
+and `afterModel` exist mostly to deal with this case.
+
+However, we can simplify all of this by moving the problem of avoiding
+model reloading into the data layer. `ember-data` is already capable
+of sufficiently intelligent caching. 
+
+Therefore, we can collapse the three hooks down to a single `model`
+hook *that always fires* regardless of whether the transition
+references a model or an id. This gives the user full control over
+whether or not they want their data to be served from cache, refreshed
+on demand, or a combination of both.
+
+## Asynchronous Components
+
+We already have a nice API for dealing with asynchronous data loading
+during route transitions. But we lack a similarly nice API for loading
+asynchronous data that does not involve a route transition.
+
+A canonical example of a component that needs this capability is a
+file hierarchy browser. Each time the user clicks a node to expand it,
+we may need to retrieve additional data, without involving a route
+transition.
+
+Today it is relatively straightforward to add this capability to a
+component on your own, but it requires every dev to think about the
+problem and put together their own solution. We can make a shared
+solution that's easier to learn and use.
+
+Once we have asynchronous components, it may be tempting to handle
+*all* data loading through them, eliminating traditional `Routes`, and
+essentially moving the `model` hook onto `Components`. But we think
+there are still strong architectural reasons for having `Routes` as a
+separate concept. `Routes` orchestrate transitions, and often have
+important work to do before you even know what components need to be
+on the page.
+
+The key questions for deciding to load data in a `Route` vs in an
+asynchronous `Component` are:
+
+  1. If you're also changing the URL, use a `Route`.
+  
+  2. If you can't decide exactly where you're going until after
+     looking at the (asynchronously-loaded) data, use a `Route`.
+
+  3. Otherwise it's probably fine to use an asynchronous `Component`.
+
+
+# Drawbacks
+
+There are clearly upgrade costs, but I've attempted to make it
+possible to gradually convert a large app.
+
+This is a significant change in mental model that makes the framework
+somewhat more opinionated about what Routes are allowed to do. For
+example, they can no longer render into outlets at arbitrary times. I
+don't think that's a drawback, but it should be considered.
+
+I did not attempt to radically simplify the semantics of the `into`
+and `outlet` arguments to Routes's `render` method. I think there's an
+argument to be made for doing so, but we would need more design work
+to do that.
+
+# Alternatives
+
+It may be possible to achieve some of this RFCs goals with something
+like the Engines RFC, by making it possible to "mount" components into
+the route hierarchy and then delegate the remainder of the URL for
+them to internally route to their children.
+
+# Unresolved questions
+
+- Should the default implementation of the `attributes` hook pass
+  queryParams directly as component attributes automatically?
+
+- This proposal makes some assumptions about the semantics of `mut`, a
+  concept that has been discussed and sketched out in the Ember 2.0
+  RFC but not fully specified.
+
+- I think implementing sticky query params via component sessions and
+  mutable attributes is a good idea, but we need to confirm that we
+  can make the timing work out such that we don't trigger unnecessary
+  double-renders or URL history entries.
+
+# Appendix A: Cross-outlet Communication
+
+The current way to alter another outlet in response to user input is
+to call `render` (or its opposite, `disconnectOutlet`) from an action
+handler on the Route. For example:
+
+````js
+actions: {
+  showUploadingStatus: function(status) {
+    this.render('upload-status', {
+      model: status,
+      into: 'application',
+      outlet: 'statusbar'
+    })
+  }
+}
+````
+
+This RFC deprecates calling `render` outside of route
+transitions. Instead, you should render what you need into both
+outlets during the initial transition, and link the two components
+explicitly via action handlers. Conceptually:
+
+````js
+renderComponents: function(attributes) {
+
+  /* This is a hypothetical way to construct a mutable
+    attribute. It's intended to reflect the same semantics as as
+    the `mut` template helper described in the Ember 2.0 RFC. In
+    this case we're just initializing one with an undefined value.
+  */
+  var uploadStatus = mut();
+
+  // Our main component receives an action in its attributes that
+  // it can invoke to set the value of the mutable attribute.
+  attributes.setUploadStatus = uploadStatus.setter();
+  this.render(this.routeName, attributes);
+
+  // Our upload-status component receives the mutable attribute,
+  // which means the component will be invalidated when the
+  // attribute changes.
+  this.render('upload-status', {
+    status: uploadStatus
+  }, {
+    into: 'application',
+    outlet: 'statusbar'
+  });
+}
+````


chancancode:
:clap: Great work!

So just earlier today, I was mentioning the async component use case to @rwjblue and @ef4.

A friend was asking for ideas on how to implement this, and he mentioned something along the lines of "you almost want a component with its own mini-route". That sounded bizarre at first, but as I think about it a bit more, it is starting to make a lot of sense.

Of course, a "route" for a component doesn't make any sense. What we actually want is the "almost free"  asynchrony handling and state management offered by the Ember router. Async and states are much more generic than the problem of routing, and since this approach worked out so well for routing, it makes a lot sense that you would desire the same kind of mechanisms in other places that you need to deal with these problems.

So, I think I would be interested in something along the lines of http://emberjs.jsbin.com/bujufi/1/edit. Basically, like the router, you have a `promise` hook (I would have called it `model` like the router, but that clashes with the property that holds the actual resolved value) where you can optionally implement and return a promise. If you did return a promise from the hook, it will put the component into the loading substate and render the `component-name/loading` template, if there is one. When the promise resolves, it will assign the resolved value to the `model` property and render the regular `component-name` template. If it rejects, it will render the `component-name/error`.

This is not a formal proposal, of course, and there are many nuances to be addressed. But overall, I feel that it fits in quite naturally with the rest of the framework.

I would also like to draw your attention to this part of the code:

```javascript
App.YelpReviewComponent = Ember.Component.extend({

  businessId: null, // Passed in

  promise: function() {
    if (this.get('businessId')) {
      return GetYelpReview( this.get('businessId') );    
    } else {
      return null;
    }
  }.property('businessId'),

  ...

});
```

The highlight is that the `promise` hook is actually just a regular computed property, so you just declare its dependencies like you normally would. When one or more of its dependencies changed, this will "re-fire", and the framework can just observe this computed property to determine whether the component has changed state.

To me, this feels very natural and aligns very well with other parts of the framework (unlike `refreshModel`, etc, which are special-occasion concepts that I need to learn).

> Once we have asynchronous components, it may be tempting to handle
> *all* data loading through them, eliminating traditional `Routes`, and
> essentially moving the `model` hook onto `Components`. 

Is this really so bad?

My gut feeling is that the routable component use case might just be a special case of async components. I think data-fetching async components should actually a pretty quite common thing, and the reason we aren't thinking/implementing them this way today is mostly just a side-effect of the poor ergonomics. If you think about it, `` `<audio>` `<video>` `<iframe>` etc are all data-fetching async components – with `window.fetch` coming to browsers this should feel much more natural and second-nature over time.

So, I think it would *need* to have an a solid API for async components regardless. Since people need to learn/do that anyway, it is perhaps not a bad idea at all to put most of this responsibility (async management, substates) in the component, and implement routing *in terms of* async components. Most people coming from server-side MVC have NFI what a "route" object is (to be honest, I still don't know if I understand what it /really/ is, as opposed to just understanding how to use them), so I think it is not necessarily a bad thing if you can get away with not doing much at all with them in the 80% common cases.

I am completely thinking out loud here, so bullshit alert on please. I feel like I am not smart enough to put together all the pieces, but there seems to be some interesting stuff in there somewhere. Substates, for example, are interesting. Since they don't actually change the URLs anyway (and in fact, we iterated a few times on whether they should eagerly update the URLs and stuff like that), it perhaps suggest that it is not intrinsically a routing concern. The CP + dependencies `promise` hook is also interesting. Then perhaps `{{outlet}}s` are really just `{{yield}}s`. Maybe all the route need to do is just instantiate the `WhateverRouteComponent`, set all the params as attributes on it and let it figure out what to do (promise, substates and whatnot). Perhaps the component (or links, etc) will call `setState` on the component to change between its child states/routes/components/templates, and the router just observe
these states to change URL accordingly...

Anyway. I don't really have much concrete suggestions at the moment, and definitely haven't considered any backwards compatibility things at all. Just thought that I would throw this out there and let people smarter than me decide if it's pure snake oil or what :smile:

I should also mention, it's almost 5AM here, so take it with many grains of salt :speak_no_evil:

workmanw:
Thanks a lot! This is quite the RFC. Overall I'm pretty happy with how this shakes out. I love the idea of async components and the use of fragments.

But I do have a few concerns based on app's usage of Ember. I've grouped them into sections below. These are grouped by priority of concern or impact to our application (biggest impact to smallest). 

-------

What about `controllerFor` or `componentFor`? We have dozens of cases where a route needs to communicate with an ancestor resource's controller. Sometimes it's as simple as toggling state (like showing/hiding elements on when another route's active/deactivate are called). Other times those controllers house large reusable logic pertaining to their model's state.

Similarly, what about `Ember.inject.controller` or `Ember.inject.component`? Very similar to above, our controllers frequently contact parent resource controllers. This type of thing happens a lot with our global nav-bars. Different routes can add/remove/enable/disable different nav options. Also when a controller creates a new model and requires one or more ancestor models to build a relationship. 

-------

It's not clear from this RFC if actions will still exist as a hash on routes. It's also not clear if there will be any method of action bubbling on the route will exist. We have countless scenarios where we leverage action bubbling, anything sending realtime messages, to contextually handling alerts, to opening modal with different options depending on where the user is in the app. It would be a huge blow to loose action bubbling on routes.

-------

> Therefore, if you want to send an action from a component up to the Route that rendered it, you should pass an action handler function as one of your component's attributes.

We have near a hundred cases of controller actions that bubble to the route. Most of the time the action originates in a component and we want it to behave differently depending on where it is used in the app. This feels like it will add a lot of boilerplate. 

-------

Storage for transient application state? We have some places in our app where we did get value from controllers being singletons. Where it was good to keep "temporary data". If you imagine Gmail, you can select a few emails, click to see the details of one, then navigate back and your selection is maintained. I know the goal of routes are to be stateless and I agree with that. I was just curious if there was a recommendation for this? Should I make a service just to store temp-selection? 

------

> This RFC deprecates calling render outside of route transitions.

We, like many other people, have been using this approach to modals: http://emberjs.com/guides/cookbook/user_interface_and_interaction/using_modal_dialogs/ This would break this pattern. That's okay with me, I didn't really love it. But I'm wondering if there is an alternative. Maybe using a ContainerView as modal manager.

------

Thanks again! I hope this didn't come across as negative. I appreciate the amount of thought that went into this RFC by everyone and I'm excited to see this come to fruition.

wycats:
I really think it's important to differentiate between "I know what component I'm going to but I have work to do before I can show it" (async component) and "I have to do some work to do before I even know where I'm going" (route).

ef4:
Yes, @wycats has hit on the main reason that I think `Route`s still needs to be a thing.

The problem with doing all the routing in components is that there is often important work to do before you can even establish a component hierarchy. Authorization decisions, for example, are really nicely done in routes.

I think this will become even more significant as we work on lazily loading chunks of application. Routes are a natural place to manage decisions around what source code is even going to be needed, without having the components themselves present yet.

zeppelin:
Based on choice of words in the RFC is it safe to assume that naming routable components won't have to be consistent with the requirements of regular components - like [how it proposed for fragments](https://github.com/ef4/rfcs/blob/element-and-fragment/active/0000-element-and-fragment.md#fragment-names)?

ef4:
@zeppelin Yes, definitely. We don't want everyone to need to rename their routes to contain dashes just became Web Component names need dashes.

jerel:
Since the 2.0 plan was announced I started focusing on components and services with minimal usage of controllers in my apps and I feel this RFC is well done and covers the pain points that I've felt.

A couple questions I still have:

1. What about the use case of manipulating (filtering, sorting, etc) data before it is displayed? Sorting, filtering, etc used to be done in a route's ArrayController. Will the documented path be to create a service for this data or to do it in a parent component with the `{{#each}}{{/each}}` in that component's template?

2. With async components is there something that will keep new developers from eliminating all routes except the application route and breaking the url behavior (back button) like so many other SPA frameworks allow?

ef4:
@workmanw So we almost certainly don't want something like `Ember.inject.component`, because one of the key things about components is their isolation. They can be understood and re-rendered reliably precisely because all their incoming state comes via their attributes or via services.

Where you're using `controllerFor` or `needs` to access a parent controller, you would instead have the parent component explicitly pass attributes or action handlers downward.

Where you're using `controllerFor` or `needs` to access some completely separate controller elsewhere, that should probably become a shared service that is injected into both.

I should probably update the document to be explicit about the fate of `controllerFor` and `modelFor`. I think they go away entirely.

ef4:
@jerel Any data manipulation an ArrayController was doing before can be done the same on the component instead. For example, if your component receives a `model` that is a list of people, it can derive a `filteredPeople` property to be rendered.

Your second question is harder to answer. I agree that we don't want to make it easy to break the URL. We already run this risk today as people build their own ad-hoc asynchronous components, but by making asynchronous components easier we may get more of them, in scenarios that really should have been reflected in the URL.

workmanw:
> Where you're using controllerFor or needs to access a parent controller, you would instead have the parent component explicitly pass attributes or action handlers downward.

I guess the reason I struggle with this is because I don't know where this data would live. Like let's say I have a left nav that can be collapsed/expanded. Right now that is a boolean value on a controller high up in the ancestor chain. The controller for a leftnav uses `needs` and some of the routes use `needs` to get to the aforementioned controller. Now any controller can just do `this.set('controllers.application.leftNavOpen', false');`. But since I can't do `needs` and I can't do `Ember.inject.component`, where does that state live? How can components way down the chain change his application state.

chancancode:
> I really think it's important to differentiate between "I know what component I'm going to but I have work to do before I can show it" (async component) and "I have to do some work to do before I even know where I'm going" (route).

I think that differentiate makes sense.

> The problem with doing all the routing in components is that there is often important work to do before you can even establish a component hierarchy. Authorization decisions, for example, are really nicely done in routes.
> 
> I think this will become even more significant as we work on lazily loading chunks of application. Routes are a natural place to manage decisions around what source code is even going to be needed, without having the components themselves present yet.

Those seems perfectly legit uses, too.

I think, what I am really getting at is the former ("I have work to do before I can show it") seems like a very common need for components in general, not just routable ones. Therefore, it appears (again, I am very much just speculating) to be a nice thing to be able to reconcile and always use the same pattern in both the routable and non-routable cases, whatever that API ended up looking like. 

So, I agree there's very much still a need for routes. "80% common cases" is probably overstating it by a lot, but I mean "in the simple examples you see in the guides today (the blog app, or a github issue browser)", ideally you wouldn't/shouldn't have to touch the routes much, because those really just fall under "have work to do before I can show it".

I agree with @wycats that those two are separate things that are currently conflated. If we can encourage a good separation of the two (something like, "routing decision should go into the routes", "data fetching should be in the component"), it seems to make a lot of sense.

This is basically the difference between what you put in your `before_action` in Rails controllers vs what you do in the body of your controller action. You can technically just have one or the other, but that would be mega confusing.

It could be that the work you to figure out your routing problems *just so happens* to give you the right data for display purposes, too. It should be possible to re-use those precious bits you downloaded for both purposes, but that feels it's technically more like an optimization (even though it is quite common) rather than the default way to reason about them ( => "since it's so common you should always do that in the routes").

By the way, I think the re-using data thing will be a problem outside of routable components, too. For example, in http://emberjs.jsbin.com/bujufi/1/edit, the `{{ related-yelp-reviews }}` component fetches a list of related reviews, and wants to reuse the `{{ yelp-review }}` component to display them. Since the list already has the data for displaying the reviews, you would want to pass those materialized models down instead of the IDs.

I'm not sure if the strawman API in the jsbin is really that great of an idea to accomplish this, but I am just saying this would need to be solved anyway, and that solution (whatever it turned out to be) is probably reusable in this case.

* * *

Anyway! I don't usually look at or comment on these things, and I think my comments has very little to do with the concrete proposal you are having, so I hope they aren't too distracting from the real work that is going on here :smile: If this is counter-productive, I will very gladly move it to somewhere more appropriate :smile:

ef4:
@workmanw that's a good example. I see two possibilities, depending on your particular situation.

The default thing to do when you want to share state from a parent component down to some children is to pass `mut` attributes explicitly. The benefit is that this is completely transparent and there's never any mystery about where a state change is coming from. The downside is that it does require each intermediate level to explicitly pass the value onward to its own children.

The second option is for cases where you have some widely-shared state that doesn't make sense to pass everywhere explicitly, and you create a service to manage it. Services are intended to be the well-defined escape hatch for sharing state "sideways" through the application. You could make a "leftNav" service that is injected into all the components that need it, and they can all read and write the shared state through the service. 

I think we need to publish more concrete details about services, because they are clearly critical for routeable components.

workmanw:
@ef4 Thanks for the back and forth. The first option would not be desirable for us, it would just create too much boilerplate because we have so many cases where descendant controllers contact ancestor controllers. The second option would be more ideal for us, but I'm not exaggerating when I say that 2/3 of my routes would end up with their own services. 

It's possible that our app is in the minority, we have a lot of "dense" UI interactions, but the more I try to apply this API conceptually to our app, the more I feel like it's putting handcuffs on me. Things like action bubbling, `controllerFor`, `modelFor`, `needs`, `Ember.inject.controller` all made it easy to communicate top down and bottom up. It feels like this API proposal decouples and insulates the routes and components along the ancestry chain.

Again, hopefully this doesn't come off as negative. I'm just trying to provide honest feedback based on our needs and how we've used Ember. Thanks a lot for hearing me out! See you guys at Emberconf.

ef4:
@workmanw your concerns are totally appropriate and valid. We absolutely need to make sure the API is great for those scenarios. This discussion is the whole point of publishing RFCs.

jesse-black:
My questions:
* Will it be possible to pass actions on the route to nested components without polluting the implementation of outer components to manually relay unrelated actions?
* How will it handle reusable (from multiple routes) components that require operations on the ember-data store such as creating records? Right now the store is injected to controllers and we could create a controller to be reused from multiple routes. I assume the store won't be injected into components, so instead they will have to be handled on the top level application route if the components are to be used from multiple routes. It seems like this will lead to a bloated application route full of unrelated functionality to support multiple components like this.

davidlormor:
First off, let me say great job with RFC :+1:  - this deals with some of the issues I've run into moving to a component-oriented structure (on v1.10 right now).  In particular, async components is a much needed addition - this is a friction point I've run into many times now.  Also, glad to see the transition to a singular `attributes` hook vs. the current `beforeModel`, `model`, `afterModel` implementation with no clear way to "force" reload of model data.

Just wanted point out one method I've come across that never seems to be mentioned (this might be of use for @workmanw re: action bubbling). There seems to be quite a bit of  confusion/worry/apprehensiveness in the discussions around having to explicitly pass in methods to a component.

There is another way...although I seem to be the only one advocating this (wondering if I'm missing something...): using the 'targetObject' property allows you to access the current parent when a component is inserted into the DOM.  The example that I use (a common pattern in my application) is the [cancel button (JS Bin)](http://emberjs.jsbin.com/rixuso/1/edit?html,js,output).  As you can see, I don't have to pass in the action, but rather the action is fired from the component with assumption that it will be handled up the chain. Curious to hear any thoughts on this method of firing actions, and what might be done to improve upon this way of action "bubbling." Perhaps borrowing from the ideas presented in the Flux architecture about having a single "action dispatcher"?

Also...very much this (from @ef4):

>I think we need to publish more concrete details about services, because they are clearly critical for routeable components.

I can definitely see the value in the services concept, but details are vague at best around how they work...would be happy to help make them a "first-class citizen" if someone could point me in the right direction.

sandstrom:
Interesting, I think this is great overall – very much looking forward to it!

Two thoughts:

1. `beforeModel` has quite a few uses besides those mentioned in the RFC. One example is pausing to setup authentication[1], which must occur before calling the model hook. And moving that to say an initializer isn't ideal, because it's common that only certain routes are authenticated (requires auth).

    Will there still be a promise-aware hook to handle such cases?
    (if not I vote for keeping `beforeModel`)

    One great thing with Ember is that it's easy to customize (hooks are a part of this). Perhaps one could keep both `beforeModel` and `afterModel` (possibly under different names) since they have use-cases not directly tied to models.

2. Under *drawbacks* it's mentioned that:
    > "[it] makes the framework [..] opinionated about [...] Routes. For example, they can no longer render into outlets at arbitrary times."

    I think modals are commonly handled by having the application route (or another route) rendering into a specific 'modal outlet'. I don't have a clear idea of how that should be handled, but it would be great if that use-case is given some thought.

    *I'm not a fan of modals and think they should be avoided, but to be fair they also have their use-cases.*

[1] https://github.com/simplabs/ember-simple-auth/blob/master/packages/ember-simple-auth/lib/simple-auth/mixins/authenticated-route-mixin.js#L42

tim-evans:
I have a few thoughts around the `mut` keyword. First of all, it's strange to consider something mutable v. not mutable, since Ember doesn't use immutable objects. Is mutable the wrong word for what's being expressed here?

In addition, I'm not sure if I buy the use cases. Can someone explain to me the pain points that are solved by using `mut`? For a developer like myself, it adds a lot of friction with development. I'm all on board with making Ember great for new devs, but handicapping experienced devs seems like it's not the right approach. I'm confident that I know whats mutating what, and I'm not scared of 3rd party addons that consume my data, for two reasons. 1) The pool of addons is small and 2) They can access any data in my app by rummaging through my app's container and registry.

I think the question here is, what perks do experienced Ember devs get out of this change?

slindberg:
Huge thanks to @ef4 for tackling the most challenging of all changes in 2.0. :beers:

> However, we can simplify all of this by moving the problem of avoiding model reloading into the data layer. ember-data is already capable of sufficiently intelligent caching.

I really like the idea of the `model` hook being idempotent; it makes complete sense take advantage of a data store's identity map. However I don't see this working well for routes that have model hooks that find record collections. For instance, say a route does something like this:

```javascript
// app/routes/things/index.js
  model: function() {
    return this.store.findAll('things');
  }
});
```

Does this mean that when query params are updated on the `things.index` route, a transition is triggered and all records are re-requested? Ember Data could do a slightly better job here of knowing if all records have been fetched previously, but the point still stands for things like `store.findQuery({ ... })`.

> Therefore, we can collapse the three hooks down to a single model hook that always fires regardless of whether the transition references a model or an id

Does this mean that the first parameter to `model` is a params hash sometimes and a model other times? This seems backwards incompatible.

kimroen:
> Does this mean that the first parameter to model is a params hash sometimes and a model other times? This seems backwards incompatible.

@slindberg I think this was referring to the way `link-to`, `transitionTo` and friends currently either fire or don't fire the `model` hook depending on if the passed in reference are params or a model, and that it will now fire either way. I don't think the parameters passed to `model` changes.

slindberg:
> I don't think the parameters passed to model changes.

You're right, they just can't change. I guess this just means that the params will need to be extracted from the model when passed in via `link-to` or `transitionTo`.

kimroen:
> You're right, they just can't change. I guess this just means that the params will need to be extracted from the model when passed in via link-to or transitionTo.

Yes, just like [Ember.Route#serialize](http://emberjs.com/api/classes/Ember.Route.html#method_serialize) already does to construct the URL from the passed in model.

ef4:
In response to @sandstrom 's question:

> beforeModel has quite a few uses besides those mentioned in the RFC. One example is pausing to setup authentication[1], which must occur before calling the model hook. 

As long as the model hook is *always* called, you can encapsulate that behavior inside the model hook itself:

```js
  model: function(params) {
    return setupMyAuthentication().then(() => MyModel.find(params));
  }
```

The only reason we had to have a separate `beforeModel` in the current system is that you couldn't rely on `model` always running.

sandstrom:
@ef4 I agree that's a way of solving it.

However, I like that Ember has hooks for common things, e.g. `willDestroy`, `didInsertElement` and `didTransition`. I think it's convenient with hooks for common things, instead of having to call `this._super()` in the model method of all child classes (needed if using a mixin or ancestor class).

I argue in favor of keeping these hooks.

mitchlloyd:
@workmanw I think that particular modal pattern is going to be simpler now that we have better components and better thought technology around them ("data down actions up").

> We, like many other people, have been using this approach to modals: http://emberjs.com/guides/cookbook/user_interface_and_interaction/using_modal_dialogs/ This would break this pattern. That's okay with me, I didn't really love it. But I'm wondering if there is an alternative. Maybe using a ContainerView as modal manager.

I was planning to update that guide as soon as the `foo={{action 'bar'}}` syntax lands (it's noted on the Glimmer branch). Here is a hacky version: http://emberjs.jsbin.com/cisoko/5/edit?html,js,output

workmanw:
@mitchlloyd Well I was referring to the fact that you can no longer call `render` outside of the normal render hook. With the cook book example you'd call `render` in an action handler to insert a modal into an outlet. With this change, you'll have to have all modals rendered into outlets up front.

So I've been experimenting with building a `ModalManager` which is really just a ContainerView that binds to a `ModalService`. Allowing anyone with access to the service to open one or more modals that end up stacked.

ef4:
@workmanw I think that's a good approach. Liquid Fire already uses a modal service in the same way. It's probably something we can abstract into a shared addon.

active/0000-routeable-components.md
```diff
@@ -0,0 +1,488 @@
...
+and must return a POJO. The POJO's properties may be promises, which
+will all be resolved in the same way we currently resolve the `model`
+hook (including invoking `loading` actions as needed). The default
+`attributes` implementation is:
+
+````js
+attributes: function(params, queryParams) {
+    return {
+        model: this.model(params)
+    }
```
stefanpenner: ;

active/0000-routeable-components.md
```diff
@@ -0,0 +1,488 @@
...
+}
+````
+
+(Possible variation for this default implementation: any queryParams
+that are present get stuffed directly into the attributes as well,
+alongside `model`.)
+
+Unlike today's `model` hook, `attributes` gets called *every* time
```
stefanpenner: :+1:

active/0000-routeable-components.md
```diff
@@ -0,0 +1,488 @@
...
+Your `attributes` and `renderComponents` hooks are required to be
+idempotent (and therefore `model`, which is called from the default
+implementation of `attributes`, must also be idempotent if you're
+using it with the default `attributes` implementation). This keeps
+routes truly stateless and easy to understand. You should not rely on
+knowing exactly when or how often a route will be re-invoked, leaving
+it up to Ember to optimize.
+
+For cases where you really need to do something different on initial
+render, the existing `activate` action will still reliably fire only
```
stefanpenner: seems strange for this to be an action that could also bubble up

active/0000-routeable-components.md
```diff
@@ -0,0 +1,488 @@
...
+when a route is first entered, after Ember invokes your `attributes`
+method.
+
+We will add a complementary `update` action that fires after your
```
stefanpenner: same as ^

active/0000-routeable-components.md
```diff
@@ -0,0 +1,488 @@
...
+`attributes` hook only when the route has not been initially entered.
+
+
+### Loading and error states
+
```
stefanpenner: same as ^

seems like these should be methods, not actions/

active/0000-routeable-components.md
```diff
@@ -0,0 +1,488 @@
...
+you the full capability of components for implementing these
+substates.
+
+### Sending actions from component up to route
+
+Components are deliberately isolated and don't send arbitrary actions
+upward unless you explicitly tell them to. Routable components
+preserve this same behavior.
+
+Therefore, if you want to send an action from a component up to the
```
stefanpenner: so like ?
```js

attributes( ) {
  return {
    save: this.actions.save
  }
}
```

active/0000-routeable-components.md
```diff
@@ -0,0 +1,488 @@
...
+  }
+  bar: 0
+});
+````
+
+would look like this on a Route:
+
+````js
+  queryParams: {
+    foo: { as: 'bar', default: 0 }
```
stefanpenner: we should describe the behaviour around unsetting and re-defaulting if it the value becomes `undefined`

mixonic:
@ef4 awesome writeup, very clear and thoughtful thanks :-)

#### attributes vs. attrs

At the component level, we talk about a component having `this.attrs` as a bucket for data passed down. I find the distinction between `attrs` (the data bucket) and `attributes` (the attributes of your shadowed elements) to be really useful when building a mental model. The word attributes is already so overloaded that I'm wary of adding it again at the model level.

Would you consider changing the route hook to be `getAttrs` or `initialAttrs`, or generally something that ties it more directly to the component-level concept? Conventionally it would mean most of your uses of "attibutes" in this document would become "attrs".

Happy to PR something against  your branch.

#### semantic argument for beforeModel

Merging `beforeModel`, `model`, and `afterModel` is definitely possible. Where do you see something like authentication going? It seems a poor pattern to suggest that devs throw authorization into a hook named `attributes`. I can foresee some nightmarishly complex `attributes` functions.

But since `attributes` runs on every transition, it doesn't seem the correct place to validate entrance into a route anyway. `activate` is not an asynchronous hook on the routes, though it fires at the correct time. Could `activate` be made async? Or can we preserve a hook semantically equivalent to `beforeModel`? `validateEntrance`?

stefanpenner:
@mixonic enter/exit?

mixonic:
@stefanpenner made async, sounds good!

2468ben:
@workmanw @ef4 I just moved a whole modal/panel system around that whole this.render(...), and only survived using a service and 1.11's component helper. Instead of a 'modal' outlet that you'd usually render into whenever you felt like it, you could replace the outlet with a template like this:

```
{{#if modalService.componentName}}
  // modal container stuff like a header, close button, etc. that only shows up when there's a modal
  {{component modalTypeService.componentName
    model=modalTypeService.componentModel
    newModal='newModal'
    manualTransition='manualTransition'
  }}
}}
```

Then whenever I want to show a modal, I have an action on the application route called newModal. It sets modalService.componentName and modalService.componentModel to null, then sets them to the new values you want. That way the whole component/DOM gets cleared out and builds up fresh.

It also means you can set those values immediately, instead of waiting until the template with {{outlet 'modal'}} is up and ready. This specifically helped my life a ton with query params that were trying to throw stuff in outlets before the initial page transition even finished. 

The annoying part of this way is listing all the possible actions or props you'd wanna pass in to any modal component you have, but keeping with the 2.0 flavor I just passed all my props in a big object into componentModel. Otherwise I have a couple actions to pick another modal or make another transition, and anything other actions I sneak into a service.

2468ben:
@ef4 I use liquid-fire for smaller dialogs too and am moving our big stuff to it, thank you times ten. I think between liquid-fire and the component helper, we can solve most of the problems we used to solve with this.render() or dynamic outlets that won't be supported in 2.0.

williamsbdev:
If the `afterModel` hook is being deprecated, when is the `model` hook promise resolved and where would my logic go for when I have an actual model object.

ef4:
@williamsbdev If you want to do something with the model after its promise resolves, you can just do it directly:

```js
model(params) {
  return loadMyModel(params).then(theModel => {
    doSomethingWith(theModel);
    return theModel;
  });
}
```

This is better than having a special `afterModel` hook because it's one less piece of Ember-specific API to learn. It's just a standard promise-returning function, and you can chain whatever behavior you want inside it.

The timing of the resolution doesn't really change from the way it's done now.

williamsbdev:
@ef4 thanks for the tutorial. I like the idea of just chaining a `.then` on the model. So just to clarify the model will be fully resolved before being setting it on the component, just like it is now before setting it on the controller?

ef4:
Yes. There has also been separate discussion around asynchronous components that have their own support for resolving promises, but that it beyond the scope of this particular RFC.

Panman8201:
Just to verify, are the other Route hooks remaining? (activate, deactivate, and redirect) Also, not replacing `setupController` with `setupComponent` seems a bit harsh. I know a _lot_ of people use that hook to do other things as well. Separation of logic is a good thing, and removing all the hooks limits this.

ef4:
Removing an imperative `setupComponent` is actually extremely important. We want the view layer to be free to create and destroy components on its own schedule -- the routes should never be able to reach directly into the component hierarchy and touch component instances. Here's a concrete example of why:

```handlebars
{{#if showingDetails}}
  {{outlet}}
{{/if}}
```

When the user toggles `showingDetails` on and off, we want to be able to create and destroy the component that's sitting inside that `{{outlet}}`. We're not going to trigger any new transitions. We also don't want to reuse the same component instance, because that makes component lifetimes significantly more difficult to reason about -- it's easier if you treat them as disposable, both from a performance standpoint (we can just nuke the whole DOM tree underneath `showingDetails` without traversing it to see if anything should be saved) and from a reasoning standpoint (it's nice to be able to trust that your component's `didInsertElement` will only happen once, and no state will leak across multiple instantiations).

All of this also applies to making animation easier. If the view layer is free to create and destroy components without trying to preserve instances that have been carefully crafted by route hooks, it can do a lot of animation magic without getting entangled in the router's logic.

Panman8201:
@ef4 Understandable. I guess my point was with the collapse of all these hooks, lots of logic will be moved into one, `attributes` hook. I suppose it will be up to the developers to create other functions to separate that logic.

Again, just to confirm, are the other Route hooks remaining? (activate, deactivate, and [redirect](http://emberjs.com/api/classes/Ember.Route.html#method_redirect)) I think they all serve their own purpose that is separate from the `attributes` hook.

2468ben:
Will the new model function still pass the transition in as an argument (like how afterModel currently does it)? In case I want to redirect based on the resolved model.

balinterdi:
@2468ben I don't think need a separate hook for redirection. Taking @ef4's above example, you could redirect like this:

```js
model(params) {
  return loadMyModel(params).then(theModel => {
    if (theModel.length === 1) {
      this.transitionTo('books', theModel);
    } else {
      return theModel;
    }
  });
}
```

Even today, the current transition is implicitly aborted when you transition to another route.

2468ben:
@balinterdi that's perfect, thanks for responding. It would be still be great to have access to the transition (for its properties or for sending actions before the transition's done), but I'll take what I can get.

Panman8201:
@balinterdi I've seen use cases where you want to "save" the current transition only to "retry" the transition later on.

@2468ben This is where the [`redirect`](http://emberjs.com/api/classes/Ember.Route.html#method_redirect) hook will remain useful, as it gets both model and transition. Thus my question about those hooks in Ember 2.0.

2468ben:
@Panman8201 Definitely, `redirect` will be great for replacing `afterModel` stuff. If I wanted to access/bother the transition before getting a model like in `beforeModel`, then I'd need it in the new `model` function anyway. Unless there's still going to be a function before `model` where I can access the transition.

active/0000-routeable-components.md
```diff
@@ -0,0 +1,488 @@
...
+have a default implementation like:
+
+````js
+renderComponents: function(attributes) {
+    this.render(this.routeName, attributes);
+}
+````
+
+The `attributes` argument comes from the new `attributes` hook as
+described in the previous section. It is resolved with `RSVP.hash`
```
chadhietala: Is it possible to have an escape valve for waiting on promises?   The behavior described above forces the transition to wait on the slowest call.  Currently the way you get around this is returning a hash that contains invoked promises.  If this is the default, then maybe the escape valve is "just use an async component".

jayphelps:
image

According to @tomdale's 2.0 RFC, the plan was to ship routable components in `v1.12.0` but it looks like no one is working on this; or at least, no WIP PR exists that I can find.

Is anyone working on this in secret or do we still need a champion? I know @ebryn has [ember-routable-components](https://github.com/ebryn/ember-routable-components) but he called it a "hack".

I realize most of the team is working on the big Glimmer push, but this questions is asked of me a lot and I don't have a good answer.

active/0000-routeable-components.md
```diff
@@ -0,0 +1,488 @@
...
+for each of those things, and allowing views to choose their own
+template, etc.
+
+Instead, we can deprecate the use of the `view`, `controller`, and
+`model` arguments, and accept a name which will be used to lookup a
+*component*, not a template. We will also accept arbitrary attributes
+to set on the component.
+
+````js
+this.render('post', attributes)
```
jayphelps: From your [previous comment](https://github.com/emberjs/rfcs/pull/38#issuecomment-76535592) it sounds like the solution is just "they are still components....but don't follow the dash rule". That seems like it will be quite confusing for some, is my concern unfounded?

rwjblue: Components that are rendered from the route will not have a dash requirement. The dash requirement is mostly to deal with DOM compatibility when invoking the component. In the rout able component context, there is no invoking template (it is the route).

jayphelps: @rwjblue right, I get that but isn't this inconsistency 1. confusing, 2. not future proof. If Ember uses real web components with custom elements, that rule will apply to every component, regardless of how it is created.

jayphelps: Hope I'm not being rude. Just wanna be sure we're making the right decision since we'll be locked in until 3.0.

rwjblue: 1) We have to land somewhere.
2) We won't run out of version numbers. :stuck_out_tongue_closed_eyes:

mixonic: it would be neat to make routable components fragments by default, and thus not have a conflict with custom elements. This might be a 3.0 thing anyway though.

knownasilya: Hum, it's new API, why not in 2.0?

mixonic: element/fragment RFC needs to be considered as well. If that was in, I would say that the default layout for a routable component would be `<fragment></fragment>`.

Not being able to listen for events and not having `this.$()` seem possibly confusing. I hope we can move away from attribute bindings anyway re: element/fragment RFC. If you want an element, just add an element in the template.

DingoEatingFuzz: Please don't make routeable components require a hyphen in the name. It's already crippling when I need to find a way to stick a hyphen in components that have one-word names. This would also mean routes need hyphens, and I don't want to start identifying ember apps by silly urls such as `/application-users/userid`

jayphelps: Just to clarify--The URL isn't directly tied to the route name. You can override the ` { path: '/users' }` but also ember could have a convention based approach, where routable components are suffixed but that suffix is removed when auto-creating the path for you. e.g. `<users-route>` -> `/users`.

It does seem core isn't down for such a thing, so I don't think you need to worry though. I just wanted to clarify that it wouldn't mean your URLs need to suffer.

DingoEatingFuzz: Sure it could be overridden, but as you mentioned, the convention of path = route = component would suffer.

sandstrom: I'd also prefer that routable components doesn't require a hyphen (for the same reason as Dingo mentions).

Personally I don't think it's confusing if they differ slightly from template-invoked components (they'll presumably be different in a few other ways regardless).

MiguelMadero: +1 to the no hyphen. @jayphelps I don't think it would be that confusing. It's when we're writing markup where the hyphen is a requirement.

wycats: @MiguelMadero :+1:

tomdale:
Needs update to take into account #60 and other recent changes in thinking.

chadhietala:
Linking from https://github.com/emberjs/ember.js/issues/11721

It's about two weeks out from the 2.0 stable release and I'm wondering what the status is on routable components. I'm aware that routable components were pushed back into the a 2.x release, however it appears the soonest they will land is 2.2.0. While this is fine, it's becoming confusing to people as to why they need controllers still and the ergonomics around it are pretty lousy. I've had to explain to numerous people who have taken "2.0.0 training" from @ebryn that routable components are not a thing but they are Coming Soon™. The sooner these can land the better. I don't consider the 2.0.0 programming model to be complete until they are there. Perhaps I should just wait 8 months like I did for HTMLBars. :trollface:

knownasilya:
Last I knew, they still needed a champion. It is tough, especially with on-boarding and saying that controllers are only needed for query params. 

It's hard from the OSS side as well, because not many people can tackle a feature this large due to the size and complexity of the framework's architecture. That's one thing to be desired, better documentation on the internals for contributors, but once again that takes time and needs to be updated every time a major refactor lands (glimmer, etc.). Sigh.

mixonic:
There was a spike either @trek or @ebryn were working on. I believe there are still one or two unknowns design-wise.

I highly doubt they are 8 months out.

SaladFork:
1) **Sending actions from component up to route** states:

>  if you want to send an action from a component up to the Route that rendered it, you should pass an action handler function as one of your component's attributes.

What would this look like?
```
attributes: function (params, queryParams) {
  return {
    save: this.actions.save
  };
}
```
Would we need a helper to get the closure action behavior of `{{action`? Perhaps it would look more like:
```
attributes: function (params, queryParams) {
  return {
    save: action('save')
  };
}
```

---

2) In **Query Parameters** the `bar` property is removed "because the Route itself does not have a `bar` property." If this is the case, what is the purpose of the `as` option then?

---

3) slindberg touched upon this briefly but I feel it requires more attention. **Data Down** states:

> Note that this implies you will now always be opting in to a "full transition" ...

> However, we can simplify all of this by moving the problem of avoiding model reloading into the data layer. ember-data is already capable of sufficiently intelligent caching ...

> This gives the user full control over whether or not they want their data to be served from cache, refreshed on demand, or a combination of both.

There are many cases one would always want the latest data whenever a route is entered. Logically I'd approach this by making my `attributes` implementation always get the latest data via the appropriate option to ember-data's store API.

Additionally, it is often useful to use query params to maintain UI state in the URL (is a modal shown, what element is selected, what tab is being shown, sorting order, client-side filtering values, etc.).

These conflict, however, as UI state query params changing shouldn't result in the `attributes` hook being called and causing AJAX/data requests.

knownasilya:
3) concerns me most. I do that a lot in most of the apps I've worked on, and a full transition would be completely unacceptable.

ef4:
@knownasilya the meaning of "full transition" is probably not that clear. Even a full transition will do very little work if your model hook is idempotent and Glimmer sees that most of the page is unchanged. This is central to the whole idea of the react-inspired, data-down programming model.

Under the hood, Glimmer *always* revalidates your whole page from the top down whenever anything changes. So the distinction between "just change one value on my controller" and "do a full transition" is not a particularly meaningful distinction anymore.

knownasilya:
Thanks for the clarification @ef4. It make sense, since the view is smarter now, and only updates the DOM that needs to be updated.

workmanw:
Recently I've been trying to practice the routable components pattern. One thing that's come up that's been pretty tough to handle is figuring out how the route should ask the component if there is pending changes. Historically, if a user tried to transition away from an "edit page" we would use the `willTransition` hook to ask the controller if there were unsaved changes. Then prompt them.

With the `attrs` hook and this new paradigm, I'm not quite sure how the route would ask the component in this case. It might be possible to check `model.isDirty` sometimes, but even that isn't always reliable.

knownasilya:
What's wrong with the `willTransition` hook?

workmanw:
@knownasilya There is nothing wrong with the `willTransition` hook. The problem is within the `willTransition` hook, how can I ask the component if there are pending changes?

Here is what it use to look like:
```
willTransition: function(transition) {
  var controller = this.controllerFor(this.routeName);
  if(controller.get('hasPendingChange')) {
    transition.abort();
    this.get('modalService').promptChanges(/* ... */);
  }
}
```

Since there aren't controllers and I don't think there is a provided way to lookup a component ... how should I figure out if there are pending changes. I can't just do `this.componentFor(this.routeName)` like you could with `this.controllerFor(this.routeName)`.

Does that make sense?

knownasilya:
@workmanw if nothing exists along those lines (it should), maybe you could do at least `this.component` like `this.controller` to at least get access to the current component. You could always lookup the component on the container, e.g. `var myComponent = this.container.lookup('component:my-component');`.

mixonic:
I am wading into this without too much context, but I will offer two thoughts.

First, components are not singletons. As you depart a route, all the components will be thrown away. Thus unlike singleton controllers, there is no need to reset state on them as the user navigates away.

Second, if some state is required by the route to control if the user can depart, it seems like there are two options. First, the route can have a way to inspect and directly call methods on the component. This feels funky, since it violates the isolation of the component. The second approach would be that state should be kept on the route itself. Basically the `attrs` object that becomes hash params for the routed component should encompass anything you would need as state at the routing level. Of course this comes with other pain, like do we need a `mut` helper if you want to pass mutable state down.

knownasilya:
Good point, so container lookup wouldn't work either.. I guess something like a service would be useful to keep the state. So the route and the component both have access to the service and other routes can subscribe to it as well.

workmanw:
@mixonic 

Option one (component access in the route) -- This seems the most flexible and to be the easiest to explain and use, but I agree it seems very funky. 

Option two feels more inline with the goal or routable components, but comes with a handful of side effects. `{{mut}}` as you mentioned, but also in many of my cases `hasUnsavedChanges` is a computed property that looks at several different pieces of state. Additionally, if the user chooses to discard their changes we'd like to call `rollbackChanges` function on the controller [now component].

Option three is a service as @knownasilya just mentioned -- But I'm not kidding when I say I'd have services for like 2/3rds of my routes. Then I'd either have to have a mutable/two-binding/alias mechanism to be constantly propagating the state of `hasUnsavedChanges` to the service; or I'd have to ultimately store all of the data on the service and have `hasUnsavedChanges` live there. But if I went through all that trouble Services would be the new Controllers and Components would be the new Views. 

Blah. Well at least I've conveyed what I hope is a pretty common use case.

mixonic:
@workmanw you've raised a great use-case to consider. :+1: Thanks!

mitchlloyd:
One nice thing about components over controllers is that we have a lot of control of their lifecycle. It might be interesting if a component could send an action (or call a method in the routing service) in its willDestroy hook to abort the route transition.

Another idea, maybe the routing service could trigger an event that would give the component a chance to abort the transition and send an action to show the modal.

2468ben:
@mitchlloyd I prefer your second idea, sending an event from the service to any subscribed components on willTransition. Either you could pass along the transition and let a component abort it.

I'm also not looking forward to the "almost every service has a route" like @workmanw, since we're really talking about a route-related event. I'd rather just subscribe to the route for those events, and subscribe to the service for the non-route-specific stuff.

Since all components will re-created under each specific route, could they include a reference to the current route? Or is that something we'd have to manually pass down our component chains like `{{post-content routeRef=attrs.routeRef}}`?

ebryn:
@workmanw I believe the solution to your problem is to inject the routing service and subscribe to a willTransition event on your component

ef4:
Yes, it's the editor component that holds the unsaved state, and that's where you want to be configuring rules for not accidentally dropping that state. It's not really a property of the current route at all.

workmanw:
@ebryn @ef4 @mitchlloyd Yea, I completely agree that it's the component who is really responsible for making the decision about preventing route transition, I just wasn't sure how that API would look. Using the router service makes sense to me. Hopefully it will be publicized around the same time as routable components. 

Thanks for weighing in!

knownasilya:
What's the router service? Any RFCs or discussions I can have a look at?

2468ben:
Oh, the "routing service" :)

This year of Ember has been like going on a long trip to somewhere, sometime, for a while, and we'll tell you more as we get there but **all your friends are going**

You should pack everything you can think of and expect to throw out *almost* all of it once we get there. But your tour guides are friendly and the New World is gonna be *awesome*

Oh, did you pack for the snow we forgot to mention the snow

And the bears

wycats:
@2468ben I'm really sorry you feel that way. We have worked somewhat hard to try to avoid regressions from what you're already doing (we were careful not to deprecate anything without a way to immediately transition to something else), while laying out a somewhat long-term sense of where we're going.

In retrospect, all of that long-term discussion has created the "road trip" sense. I'm curious: is there some way that we have broken an existing pattern that you were using that requires the "routing service" to work again? Or is the frustration just that the solution for certain questions that are not yet supported often ends up being a long-term proposal that we haven't yet started to work on?

Either of those is a problem, of course.

ef4:
@2468ben This repo is for discussing the future. It's entirely appropriate for concepts to pop up here that aren't documented or implemented anywhere else yet.

> but all your friends are going

It's true that we celebrate teams that manage to stay on the bleeding edge. But we celebrate them for a reason: it's *hard to do*.

People should *not* be feeling pressured to live on the bleeding edge. There is zero risk in sticking with stable 1.x APIs for as long as you need to. We are committed to providing security support for the 1.x branch for the foreseeable future, because we know many apps will continue to run it until 2.x has time to bake and stabilize.

2468ben:
Don't worry, I'm fine! I was hoping that by typing a smiley face plus the phrase "and the bears" into a github thread, I could communicate that I'm being lighthearted about this whole situation. Ember is a community-created cutting-edge open-source project, and everyone expects a ever-evolving thing with ever-shifting goalposts. I'd say I'm 1% frustrated at Ember, and 99% frustrated at how it's impossible to sound lighthearted online.

mitchlloyd:
@2468ben You had me laughing out loud and I took it as lighthearted :)

2468ben:
-- Take all of this with the same lighthearted tone please --

To be fair, there is absolutely pressure to live on the bleeding edge, and I'm not even living there. I'm on Ember 1.11, Ember Data beta 15, ember cli 0.2.3, and a couple addons that are still compatible.

I could have stayed away from ember-cli, but after a year of being told in every official and non-official resource why I should totally use it, why the future would be devoted to it, why it would be in lockstep with core Ember, and why all useful addons/plugins would move to it, I spent the weeks to migrate. For two different companies. That's a decent hunk of pressure especially for software that was still technically alpha.

I could have stayed away from the regular requests to run in beta and canary whenever possible to help fix bugs, although that was only slight pressure. I helped out with 1.0-1.7 but couldn't handle it this time around and checking the daily changelogs.

I could stay away from 1.13 and all of the breaking changes needed upgrading Ember Data and my addons, with the only downside being that the addons have some bugs that probably won't be fixed for 1.11. The whole Ember addon "ecosystem" is tremendous, but by the time an addon bug is fixed, it might only work under a newer collection of lockstep Ember core libraries, in which case we work to update all of them together. In that sense there is upgrade pressure, with the alternative being to fork addons and fix the bugs for your version. Of course this applies everywhere in software, just faster nowadays.

This is an amazing framework, and not surprisingly those of us who use it are building big, important, job-dependent things. And it's a job priority to reach for a stable framework whenever possible. After riding the Ember data beta train and ember-cli alpha train for months/years, there is a lot of external company pressure to tell your customers you're running the stable versions of code. And internal company pressure that you're running code versions that a coworker can find docs or stackoverflow answers for, in case they're new to the game. 

There might be a little *peer* pressure in the Ember community to try the latest and greatest things, but it's the stability + "does any answer match my stack" that is the real *vocational* pressure. For everyone not at the 1.13-for-everything, if I say I'm under pressure, I mean overwhelming vocational pressure, and I thank the community for making the 1.13 ecosystem and taking that pressure off. Until everyone gets there, some of us may get cranky and need a nap (or alcohol).

For @wycats and the core team, here's why the "routing service" (which, great idea, thanks to all involved) and other soon-to-be-released features made me write a joke on Github: I'm trying to plan for the future just like everyone else, even if I'm sticking on 1.11 for a while. But when I'm in the code building a big new feature, **of course** I want to write it with components and services and in a way that won't need a big rewrite or a feature freeze! When I read the Road to 2.X or all the blog posts on preparing for the new architecture, I'd be a bad engineer if I said "to hell with my company's future, I'll write things the deprecated way". So I build my new features in a way that seems compatible for the (1 month or 1 year later) migration 2.x, in case my company ever really needs something in 2.x. Not all the features are there yet so I make a few mixins or utilities that serve as unofficial bridges, knowing that eventually Ember will release the official way.

So when I've built much of my app towards what Ember evangelizes is the future, I'm always guessing a little. And when the answer appears out of nowhere like the "routing service", I have to laugh because I can't believe I never guessed there would be a routing service! And I laugh because some custom mixins and utilities I made in preparation are now totally disposable. And I laugh because I have no idea if the routing service will disappear before 2.2, or if it will appear but not until 2.4, or what else will change by then. It's the joy and slight craziness of building with the future in mind: every time I write new code, I imagine I'm playing mini-golf with my eyes closed :)

mitchlloyd:
@2468ben Just know that you are not alone here, Ember users feel this pain every day. We're building mental models around how Ember ought to work and then it's up to the community and the core team to get us there. The value proposition of Ember is that we can change the framework without throwing it away, but it requires a lot of careful work and experimentation.

The best case scenario is that when you feel this pain, you would have the time and motivation to get involved and familiar with Ember and then help move it to where we want it. I know that not everyone can do that but in my experience, if you do have the desire, the Ember core team is really good at getting people involved. Maybe you could hold the torch for getting the routing service in a better place?

No pressure :wink:

2468ben:
@mitchlloyd Thanks for the response, I honestly am fine with wherever Ember wants to go, and however they decide to get there. I've built a few frameworks before this but I decided to focus 100% on end products, and to work with whatever big tools are out there. I have zero disagreement with any architecture or implementation decision that Ember has made; my passion is using their great decisions to build my product.

So instead of influencing the eventual routing service implementation, I really just want to know if there's something I could do to get 1.x code more compatible with the parts that are already ironed out.
Lots of changes are "mechanical" or easily explained in a few lines of blog post, but routeable components is a big one and the RFC hasn't been updated since February, so maybe there's a really quick update someone could give, even if it's "not sure yet". Something to tide us over for a few months until there's a blog post.

* Currently planning to land in 2.2, 2.3 or later?
* There was an announcement somewhere that a 'first step' landed in Canary, any link or short answer if that's for public use?
* Is the routing service a pretty sure thing, and will it be pretty much like the old Router or something alongside it?
* Anything else in the RFC you'd say has changed a lot and is worth mentioning? If there is, do you have a quick idea/suggestion for a 'prefactor' or pattern that might help ease the breaking change? Like "move your controller logic to a component if you want to get a headstart".

ef4:
@2468ben There is already a routing service in Ember, and has been for a while. It's just still marked private, because it needs a public design discussion before people start depending on it. But I expect whatever we finally mark as public to be pretty similar. See:

https://github.com/emberjs/ember.js/blob/5a023f33afa2cedb5a652cd2efd354352aa3d7c7/packages/ember-routing/lib/services/routing.js

It's not a replacement for the router, it's just a standardized way for components to communicate with the router.

tomdale:
> But I expect whatever we finally mark as public to be pretty similar.

I don't think I can endorse that. As the person who worked on extracting the routing service from the existing code, there are still a _lot_ of rough edges that I would like to get sanded down when working on the public routing service API.

Feel free to use the private API, but I would peg the odds of code that relies on it breaking in a future release at close to 100%.

2468ben:
@tomdale That's honestly great to hear, a "don't count on X" keeps me and others from refactoring because we keep hearing about X as the future. I think what I'm poking at above is if there's also "do count on X" for any bit of the Routeable Components RFC. It's mentioned constantly in Ember conversations as the future architecture and a set of big changes landing at any moment; I'm happy if the only update right now is "stay back, hide your children, nothing is certain". Or "this one tiny part will is definitely going in".

We're all gonna refactor through umpteen deprecations, so while we're in there, any yes/nos on something solidifying since February could keep a lot of folks from planning based on the old RFC. Until then I'll just putt around a little more.

berlincityboy:
Hey,
how can I pass initial attributes for a routable component?

The way which is described in the RFC (https://github.com/ef4/rfcs/blob/routeable-components/active/0000-routeable-components.md#specifying-component-attributes) seems not to work for me. Has something changed or has been implemented in a different way?

knownasilya:
@berlincityboy according to the tests, it looks the same as it is now: https://github.com/emberjs/ember.js/blob/5351b9458f306398f6168acaa6f5eafe7a0bce2b/packages/ember/tests/routing/basic_test.js#L388

ming-codes:
I would love to see there's a way to pass arguments into `{{outlet}}`. This way it'll feel like the outlet actually is the routed component. Or rather, the `{{outlet}}` helper can just be like the `{{component}}` helper except the component name is supplied by the route.

```
{{#x-component as |yield1}}
  {{outlet param1=yield1}}
{{/x-component}}
```

mellatone:
@tomdale Is there a sense of where the public routing service is at? I feel that it is a pretty important aspect to routable components, and one that the community can gather around to move forward. The {{link-to}} helper has been something of a stand-in for older versions of Ember, but it only works in the simplest of use cases.

jcope2013:
@mellatone there is an open RFC for one https://github.com/emberjs/rfcs/pull/95

RGL9032:
@chadhietala 
> Perhaps I should just wait 8 months like I did for HTMLBars.

@mixonic 
> I highly doubt they are 8 months out.

Surely it won't be another 8 months before Ember delivers on their promise, right? I mean `Ember.View` is being deprecated soon so surely we won't be left dangling without any guidance or insight on future direction. Ember would never do that to the devs who have supported them over the years ...

locks:
@RGL9032 `ember-legacy-views` will be supported until [Ember 2.8 LTS](http://emberjs.com/blog/2016/02/25/announcing-embers-first-lts.html#toc_legacy-view-and-controller-addons), _if_ the refactoring work is concluded by then ("will likely be").
If you wish to continue supporting Ember, words of encouragement will be more effective. If you feel frustrated, imagine the core developers that are in the core specifically for their commitment to improving the framework.

sandstrom:
@RGL9032 there is a ton of effort behind Ember. Sure, a thing or two may have been slower out the door than initially expected, but that's pretty common with any project (open-source or not). I think Ember kept great pace recently!

buschtoens:
It would be super useful if attributes set on the `{{outlet}}` would be propagated to the component. Take CSS classes as an example.

grapho:
@buschtoens  I imagine that case will be handled by classNames and/or classNameBindings on the routable component js

mitchlloyd:
@grapho Could you expand on those other ways? Feels odd to me that we wouldn't be able pass things from a template context into a routable component. 

One use case is where one would normally use `this.modelFor('parent')` and `model()` or `setupController()` inside of a route to get access to a parent model inside of a template. We already have a natural API to pass attributes from a parent context into a component in our templates, why not for routable components? 

```hbs
{{route-component parent=parent}}
```

grapho:
@mitchlloyd tbh routable components are still fuzzy for me, i need to re-read some of the RFC and other videos I have seen.

The final implementation is still undecided as far as I know.  There are still many pieces missing before it can happen.

What I have _heard_ at least is, setupController() and the 3 model hooks will be replaced by a more semantic hook more specific to passing attrs into the routable component... something like an `attributes()` hook (confirm?).  That is where you will be "passing in" the attributes and actions to the routable component.

Will the route template contain an actual invocation to the routable component like your example above?  Or perhaps a combination of the new attr hook and renderTemplate() as the original rfc suggests? or maybe the routable component's _template_ itself will be responsible to setting things like classNames?  I dont think there is a clear answer on any of that yet... otherwise it might already be more common knowledge.

:P

RGL9032:
Is there anything that can be done to advance this? Is this still a goal for the core team? There is such a vacuum about the state of ember, the framework goals and the timeline for things. Engines and FastBoot are looking great and have gotten a lot of attention, but those features benefit primarily larger ember users. Glimmer 2 is also looking great, and it does benefit us all, but it's taken so much time and at this point it does not bring any new features beyond performance.

Features like routable components, enhanced pods, routing service, and test unification are foundational improvements that benefit all of us. Beyond RFCs and old PRs, those features seem to be mostly stuck. What can the community at large do to help these goals move forward? I worry things have gotten so big and so complex that it's hard for the devs with only a few hours of availability to contribute. Additionally, with the bigger efforts like Glimmer 2 and Engines underway for months, I've sensed an unwillingness to take on medium and smaller features until the larger ones had completed. I definitely understand where that comes from, but some how we have to work past that. 

The amount of time that's gone into building larger features has me really worried for the future. Competing frameworks have done a much better job at iterating core parts of their API over the past 2 years, while ember has mostly focused on the edges. If you compare the framework changes from `Ember-1.0` thru `Ember-1.10` to `Ember-2.0` thru `Ember-2.10` it's night and day.

ef4:
@RGL9032 we have more activity and contributors now than ever before. I agree that it takes time and effort to stay abreast of what's going on, but that's because there's so much of it, not so little.

Please realize that when an account with no participation history jumps onto an old RFC to complain even though three major initiatives are "looking great", that doesn't leave much productive room for a helpful response.

If you have only "a few hours" but want to help I suggest spending that time looking through [the latest notes](https://github.com/emberjs/core-notes) to see what's already ongoing and who is working on it, and using that are a starting point to talk to the people involved in areas that interest you.

mirague:
Routable components look very promising1

wycats:
This RFC was opened more than three years ago and literally predates the release of Ember 2.0.

At the time, we were in the process of removing Ember 1.x-era "views", and we finally did eliminate legacy support for them after Ember 2.4. During the end of the 1.x timeframe, we deprecated and removed a lot of features very quickly. We also introduced a series of React-inspired data-flow APIs in Ember 1.13 (the last release of Ember 1.x) that became canonical APIs in Ember 2.0.

While we were making all of these changes, we assumed that we would quickly deprecate controllers and unify the concept of controllers with components. There are two reasons that didn't happen.

First, soon after Ember 2.0, we very quickly came to regret the pace of changes in the Ember 1.13 era, and resolved to avoid doing anything similar again. This more-or-less meant that any plans we had to continue deprecating, removing and consolidating features around that time were put on hold. The community had enough churn, and we wanted to let things settle before creating any more.

Second, Ember 1.13 introduced the first version of Glimmer, and at the time we were highly optimistic that the Glimmer 1 architecture would make it easy to implement Glimmer components (with angle bracket invocation) quickly. We assumed that the new "routeable components" would want to default to the new "Glimmer component" semantics in the case where a routeable component didn't have a JavaScript class, which at the time, made it seem that routeable components blocked on finishing the Glimmer component work.

Unfortunately, my personal work attempting to implement Glimmer components quickly ran into roadblocks, which ultimately led to the Glimmer 2 work, which resulted in a rather large and time consuming initiative.

Between our desire to avoid introducing churn in the early 2.x era and the fact that our efforts in the view layer got redirected into implementing and integrating the Glimmer 2 engine, our ideas for routeable components in that era got put on the shelf.

So much time has gone by since then, and the Glimmer architecture has evolved so much, that the original ideas that we had at the time aren't even relevant anymore. And in fact, the term "routeable component" has become a punchline that is permanently tainted with the long and sordid history of the exact details of this feature.

---

I think that we should still do a series of features with these characteristics:

- Allowing a top-level template for a route to have a component class
- Making it possible for routes to provide arguments to a its template other than `model` (and using the new `@name` syntax)
- Making it possible to use query params in Ember without controllers

But I don't think we should call that feature "routeable components" and I don't think that feature shares much in common with the thinking we had at the time @ef4 originally wrote this RFC. I also think that we should look to emulate @chancancode's recent work in breaking down Glimmer component into multiple chunks when we work on this feature. Instead of one monolithic, entangled feature, I think we should break it apart into small bite-sized chunks we can land a bit at a time.

Finally, with @chancancode's latest work breaking apart the Glimmer component feature (#276, #278, #280), I no longer believe that allowing a route's template to have a component blocks on Glimmer components (but there are some design questions, like what should happen if a route has both a controller and a component, and whether it should be allowed at all).

---

For all of these reasons, I'm closing this RFC.

PR #39, opened 2015-03-10
Changing top level dynamic segments while preserving others

kellyselden:
[RFC](https://github.com/kellyselden/rfcs/blob/transition-to-any-dynamic-segment/active/0000-transition-to-any-dynamic-segment.md)

### DIFF
--- a/active/0000-transition-to-any-dynamic-segment.md
+++ b/active/0000-transition-to-any-dynamic-segment.md
@@ -1,0 +1,67 @@
+- Start Date: 2015-03-09
+- RFC PR: (leave this empty)
+- Ember Issue: (leave this empty)
+
+# Summary
+
+Provide a way to change top level dynamic segments without supplying deeper nested dynamic segments.
+
+# Motivation
+
+For something like a locale in the URL that sits at the top level (`/:locale/products/:product_id/items/:item_id`), there is no easy way to swap it out while preserving the rest of the URL. Here is a [stack overflow question](http://stackoverflow.com/questions/27710301/ember-js-swap-out-a-nested-resource-from-any-route) and an [open issue](https://github.com/emberjs/ember.js/issues/10094).
+
+# Detailed design
+
+This is implemented in an [addon](https://github.com/kellyselden/ember-cli-transition-to-dynamic/). The API there is `this.transitionToDynamic(currentRouteName, { routeWithSegment: { segment: newValue } });`, or `this.transitionToDynamic('item', { product: { id: 1 } });` for the route `/products/:id/items/:id`. This implementation does not cover all cases of `transitionTo`.
+
+Possible API:
+
+```js
+this.transitionTo(currentRouteName, {
+  segments: {
+    routeName: {
+      segmentName: newSegment
+    }
+  }
+});
+```
+
+For `/:locale/products/:product_id/items/:item_id` it could be:
+
+```js
+this.transitionTo(currentRouteName, {
+  segments: {
+    home: {
+      locale: 'en-gb'
+    }
+  }
+});
+```
+
+For `/products/:id/items/:id` it could be:
+
+```js
+this.transitionTo('item', {
+  segments: {
+    product: {
+      id: 2
+    }
+  }
+});
+```
+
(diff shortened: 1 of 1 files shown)


davewasmer:
Would changing a top level dynamic segment cause all the child routes to refire their model hooks, as if you had directly transitioned to that leaf route?

kellyselden:
I'm not sure. Perhaps if possible not refire the model hooks of unchanged segments. I'm not familiar with the complexities of the router.

active/0000-transition-to-any-dynamic-segment.md
```diff
@@ -0,0 +1,67 @@
...
+      segmentName: newSegment
+    }
+  }
+});
+```
+
+For `/:locale/products/:product_id/items/:item_id` it could be:
+
+```js
+this.transitionTo(currentRouteName, {
```
jamesarosen: This line should probably be something like `this.transitionTo('locale.product.item', {`, right?

kellyselden: You are right. Although I would assume a resource reset after locale. How about `this.transitionTo('product.item', {`?

kellyselden: Hmm. I think i was thinking that this was a nav component action that would be handled by the locale or application route. So in that case, `currentRouteName` was assumed to be always calculated.

The following example `this.transitionTo('item', {` was "I want to look at this same item in a different product's context", so the route would be static.

ef4:
This was reviewed by the core team today.

When the child routes are data-dependent on the parent route, this new option creates a footgun. It is incorrect to try to swap the parent route's value without rerunning all the children.

When the child routes are *not* data-dependent on the parent route, we consider it a bug that any expensive rerendering is done, and we think Glimmer plus the ongoing routable component work should address those bugs.

We agree that there is still a need for a public API that makes it easier to transition to the existing URL with only one segment changed, and would welcome suggestions for that API. Most likely it should go on the routing service.

SaladFork:
> We agree that there is still a need for a public API that makes it easier to transition to the existing URL with only one segment changed, and would welcome suggestions for that API. Most likely it should go on the routing service.

@ef4: A suggestion for something to this effect:

Today through the private API [`Ember.Route#paramsFor`](http://emberjs.com/api/classes/Ember.Route.html#method_paramsFor) a route can get at the parameters of any route (including itself). What if this or a similar mechanism could be used for changing a parameter?

One idea for implementation: the current `paramsFor` returns a POJO whose keys are dynamic segment names merged with query parameter names. What if instead of returning a POJO, an `Ember.Object` was returned instead to facilitate data binding / observation.

Example usage:
```javascript
this.paramsFor('locale').set('locale', 'en-gb');
this.paramsFor('products').set('product_id', 2);
```

This approach averts the concern @igorT raised in [the issue](https://github.com/emberjs/ember.js/issues/10094#issuecomment-68450066) by requiring the route name be specified in addition to the segment/parameter name.

Drawbacks:
* Not obvious from seeing a `.set` that it will result in a transition.
* Potentially no longer being able to assume you can read values off the returned value (which was a POJO, now `Ember.Object`) without using `Ember.get`.

dschmidt:
Can't `Route.serialize()` be used for swapping out parameters? It could be called after the model resolved or `Route.reserialize()` could be manually triggered similar to how `recompute()` works in the new Helper API.

That's a very natural approach to correcting typos in urls - e.g. wrong casing of usernames - this would even work automatically.
It would also be very easy to swap out parameters on every level with this - it just needs an action that changes a property on a model on the right level and then calls `reserialize()` (or an observer that does it automatically when the property changes). 

The downside is that you can't generate a proper link to a different locale, just "something" that triggers an action...

wagenet:
Given the seven(!) years of inactivity on this issue I think it's fair to close it. The last relevant state was the core teams conclusion:

> We agree that there is still a need for a public API that makes it easier to transition to the existing URL with only one segment changed, and would welcome suggestions for that API. Most likely it should go on the routing service.

If someone wants to make an RFC for this specific part then I believe that would still be welcome.

PR #41, opened 2015-03-18
Add glossary section of detailed

chadhietala:
For majority of detailed designs it seems very helpful to introduce terminology upfront that will be used in the underlying detailed design.

### DIFF
--- a/0000-template.md
+++ b/0000-template.md
@@ -13,7 +13,7 @@ Why are we doing this? What use cases does it support? What is the expected outc
 # Detailed design
 
 ### Glossary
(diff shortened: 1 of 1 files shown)


0000-template.md
```diff
@@ -12,6 +12,9 @@ Why are we doing this? What use cases does it support? What is the expected outc
 
 # Detailed design
 
+### Glossary
+This is where you can definitions of terms used in the underlying detailed design.
```
locks: "you can" find?

mmun:
It's confusing how this bleeds into the next paragraph without a new heading.

PR #45, opened 2015-04-04
Solicit feedback about IE8 and IE9 support in Ember 2.x

wycats:
[RFC](https://github.com/emberjs/rfcs/blob/drop-ie8/active/0000-internet-explorer.md)

### DIFF
--- a/active/0000-internet-explorer.md
+++ b/active/0000-internet-explorer.md
@@ -1,0 +1,147 @@
+- Start Date: (fill me in with today's date, YYYY-MM-DD)
+- RFC PR: (leave this empty)
+- Ember Issue: (leave this empty)
+
+# Summary
+
+Solicit feedback about the support timeframe for Internet Explorer 8 and Internet Explorer 9.
+
+# Motivation
+
+As Ember heads towards version 2.0, it is a good time to evaluate our browser support matrix. Ember follows Semantic Versioning, and we consider browser compatibility to be under the umbrella of those guarantees. In other words, we will continue to support whatever browsers we officially support in Ember 2.0 until Ember 3.0.
+
+Ember 1.x did not have an official browser support matrix, but we would like to correct this for Ember 2.0.
+
+We want to make this decision on the basis of the browsers that our community still needs to support, while weighing that against the costs we bear as a community to support older browsers. This RFC will lay out some of those costs, so we can decide what tradeoff is most appropriate.
+
+Members of the core team maintain many different kinds of apps across many different kinds of companies. Some of us work on applications with small, agile teams, while others work inside of large corporations with many engineers. When this topic came up amongst the team, we discovered that, across all these different companies and Ember apps, no one was still supporting IE8.
+
+Because of this, the core team's impression is that the costs of IE8 support now far exceed the benefits, and we are considering dropping support
+for IE8 in Ember 2.0. Before we make the decision, we want to hear from the rest of the community. Supporting IE8 incurs significant cost, both in terms of features and maintenance, and we want the community to help us think through the cost-benefit analysis.
+
+Ember is more than just the framework's code. When people use Ember, they expect to be able to use Ember's tooling, read Ember's documentation, find solutions to problems on Stack Overflow, and read tutorials produced by community members. All of these are shackled to the limitations of IE8, and by dropping support for IE8, people can begin to rely on the improved baseline of ES5.
+
+Below, we outline the costs of continuing to support IE8, so that you can help us make a considered decision.
+
+# Detailed design
+
+## IE8
+
+### Eliminate `get()`
+
+Currently, accessing properties on an Ember object requires using the `.get()` method. By using this abstraction, we have been able to implement several powerful features, such as proxies and computed properties, even on older browsers like IE8 that lack getters and setters.
+
+However, ECMAScript 5, which shipped in **2009**, added support for getters to JavaScript itself, and we would like to use this feature in Ember to eliminate the explicit calls to `get`. Developers new to the framework tell us that having to remember to use `.get()` is a big source of confusion. More seasoned developers get used to it, but moving the Ember object model closer to the pure JavaScript object model is a major goal for Ember 2.x. While many of the features of ES6 classes can be transpiled, getters and setters require engine support, and could not be used if we needed to support IE8.
+
+### More ES6 Features, Today
+
+While much of ES6 can be transpiled correctly to ES3 (the version of JavaScript included with IE8), transpiling ES6 modules and classes requires `defineProperty`.
+
+Continued support for IE8 limits our ability to adopt new ES6 features in the internals of Ember, and to talk about them in our documentation.
+
+One example: In ES6, classes define their methods as non-enumerable properties. Transpiling this to existing browsers is only possible with `defineProperty`, which is not included in IE8. Trying to transpile ES6 classes to work on IE8 would lead to apps exhibiting subtly different behavior that would be painful to debug. IE8 users would discover that the larger Ember ecosystem was incompatible with their apps in hard-to-predict ways, and we think the ecosystem is one of the biggest advantages Ember offers.
+
+In other words, we don't think we can make the full transition to JavaScript classes a first-class part of the Ember experience if we still support IE8. As we did with modules, we would like to move more of our core to JavaScript features in the future, which would be significantly stymied by the lack of `defineProperty` in IE8.
+
+### Remove the jQuery Dependency
+
+For its entire lifetime, Ember has relied on jQuery to smooth the rough edges of browser compatibility when interacting with the DOM. When people think about that dependency, they often assume that we could just replace calls to things like `.attr` with their more verbose DOM counterpart and call it a day.
+
+jQuery does more than just patch over IE8 rough spots; it also serves as the central place for normalizing behavior that can differ significantly across browsers. If we tried to pick-and-choose pieces of jQuery to pull into Ember, we would also be responsible for backporting any changes made to jQuery. We'd rather just rely on jQuery directly; that's what dependencies are for.
+
+The jQuery dependency has helped us with a few cross-browser areas:
+
+* Portable `DOMContentLoaded` (via `jQuery.ready`)
+* Support for event delegation across a wide variety of events.
+* Attribute and property normalization, which has already been implemented by HTMLBars
+* HTML parsing, which has also been implemented by HTMLBars
+
+Of these, proper support for event delegation is the largest remaining reason to rely on jQuery. IE9's support for the capture phase of events makes it simpler to support event delegation properly across all event types without a normalization layer.
+
+### Support More Event Types
+
+Many newly specified events in the web platform (such as the media events) do not bubble, which is a problem for frameworks like Ember that rely on event delegation. However, the capture API, which was added in IE9, is invoked properly for all events, and does not require a normalization layer. Not only would supporting the capture API allow us to drop the jQuery dependency, but it would allow us to properly handle these non-bubbling events. This would allow you to use events like `playing` in your components without having to manually set up event listeners.
+
+### CSS Improvements in Ember 2.x
+
+Today, the main Ember framework does very little to directly help with CSS. We expect that to change in the 2.x series, as we explore ways to help tame the CSS beast.
+
+However, a number of important CSS features landed in IE9: CSS3 selectors, full support for `querySelectorAll`, `getComputedStyle`, `calc()` to name a few. Productively tackling the CSS problem without these features would be like fighting with both hands tied behind our backs, and it may be impossible for us to robustly tackle the problem until Ember 3.0 if we needed to continue to support IE8.
+
+While it may be theoretically possible to implement some form of this feature in IE8, it is likely that the cost of doing so in a backwards-compatible way would significantly add to development time; perhaps so significantly it would be better to wait until we drop support for IE8 than attempt to bolt it on to a browser released half a decade ago.
+
+### Maintenance Costs
+
+While it's very easy to weigh the costs of features that we could not implement at all due to IE8, there is a much more pernicious cost that is harder to see.
+
+Support for IE8 adds costs, sometimes significant, to every new feature we work on. For example, broken support for text nodes in IE8 significantly impeded early work on Glimmer. Every new area of work requires budgeting a significant amount of time for IE8 support.
+
+This is not surprising. When asked many years ago what jQuery could do when IE6 was gone, John Resig replied that we would gain little from dropping IE6, and that the benefits would not come until jQuery could drop IE8, the last version of IE featuring the bugs that made IE6 so difficult to develop for.
+
+Quite often, we will assume that a feature is ready to ship, and only discover subtle issues in IE8 very close to the release once it has been tested. We estimate that support for legacy Internet Explorer slowed down the development of HTMLBars by 2x.
+
+In short, we would be able to implement more features more quickly without the burden of bugs that were first introduced 15 years ago.
+
+## What About IE9?
+
+In the first decade of 2000, browsers were updated very slowly, and every new release took a long time to be supplanted by the next release. As the last version of Internet Explorer supported by Windows XP, IE8 is a relic of this bygone era. In contrast, IE9 usage was quickly supplanted by IE10, and that pattern continues with IE11.
+
+The public trackers have IE9 at a lower share of total usage than IE8, so it might be worth considering dropping them together. Our decision for Ember 2.0 will likely hold until late 2016, so it's worth considering more than just the current moment when making the decision.
+
+While IE9 added support for the ES5 features we need to move into the future for JavaScript, IE10 added support for the last great wave of web features. Here is a sampling:
+
+* Flexbox and Grid Layout
+* Offline storage (IndexedDB, File, Blob)
+* Web Workers
+* Typed Arrays
+* Web Sockets
+* App Cache
+* History API
+
+Several of these features are required for asm.js, and in total, they make the web platform a capable application runtime. While we don't have any immediate plans to take advantage of these web features right now, the best experiments that people are doing today rely on them. By assuming IE10 as the baseline across the entire ecosystem, we would be able to do much more aggressive experimentation on the web platform.
+
+# Drawbacks
+
+Many users have told us that they chose Ember because of the community's commitment to backwards compatibility. When we announced in early 2014 that we would continue to support IE8 for at least another year, other libraries and frameworks had already dropped support. That being said, there will always be organizations using Ember that exist on the tail-end of browser adoption patterns. We risk alienating or upsetting those users by dropping support for a browser that, while on the way out, is not yet completely gone.
+
+However, in many cases, the requirement of IE8 support is driven by non-technical management who do not have a strong sense of the experience of using apps in IE8. In practice, many applications are not rigorously tested in older browsers, and the performance of IE8 is so bad that applications written using any framework perform poorly. Techniques that framework and application developers use to make Chrome fast quite often have pathological characteristics on browsers with a DOM and JavaScript engine written in the 90s.
+
+Still, some people make it work, and dropping IE8 support may prevent those teams from staying with the community as it migrates to Ember 2.0.
+
+# Alternatives
+
+## Drop IE8 Support During 2.x
+
+One alternative we have considered is deprecating IE8 support prior to releasing 2.0, but still maintaining it for a few point releases to give IE8 more time to lose market share.
+
+After discussing with the core team, we believe that this would be a violation of our Semantic Versioning commitment to users. Specifically, we want to avoid a large group of apps getting stuck midway through the 2.x cycle. Version numbers are an important tool for developers, maintainers and ecosystems to communicate compatibility. Tools such as package managers rely on version numbers correctly indicating breaking changes.
+
+We consider browser compatibility to be a feature of Ember, and dropping IE8 support in a minor release would be akin to stripping out any other major feature. While the ecosystem would muddle along in either case, such a move would cause exactly the kind of ecosystem fragmentation that Semantic Versioning is designed to prevent.
+
+If we want to communicate the idea that changing versions comes with a reduction in functionality, we should do that the same way we always do, by incrementing the major version.
+
+## Early 3.0
+
+Another option is to release 3.0 in six months, rather than the nearly two years between Ember 1.0 and Ember 2.0.
+
+Correctly tuning the cadence of major releases is a delicate tradeoff. Semantic Versioning allows us to easily communicate about breaking changes, and some take this as a license to make them frequently. However, a robust ecosystem relies on a certain measure of stability.
+
+We believe that the frustration of breaking changes every six months (or even a year) would outweigh whatever benefits it would provide. Ember's biggest goal is building a shared foundation for our ecosystem to build on, and this requires a careful commitment to stability.
+
+While we could make a "small" breaking release soon after 2.0, breaking changes inherently fragment the ecosystem, and we hope that the years to come bring more stability for add-on authors and tool-makers, not less.
+
+## Bring Your Own Compatibility
+
+Some libraries attempt to thread the needle of IE8 compatibility by asking users to bring their own compatibility libraries. They write the internals of their framework as if IE8 did not exist, and require end users to use polyfills to make the environment look equivalent to newer browsers. For example, React asks users to bring libraries such as `es5-shim`, `es5-sham`, `console-polyfill` and `html5shiv` if they want IE8 support.
+
+Facebook.com supports IE8, and uses React, so there is a path to using React with IE8. This path is partially documented on the React website. This gives us a perfect opportunity to evaluate the impact of this strategy in the real world. We admire the React team's work in this area: support for IE8 is difficult and triaging and fixing IE8 bugs requires diligent effort.
+
+After reviewing the [IE8-compatibility issues filed on React.js tracker](https://github.com/facebook/react/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+ie8), we believe there are significant user experience costs to this strategy.
+
+We have spent considerable effort on first-class IE8 support in Ember 1.x, and we feel that users who require IE8 support will have a better experience using Ember 1.14 (with the subset of the ecosystem that supports 1.x) than trying to cobble together a solution that works reliably in a version of Ember with second-class, bring-your-own-compatibility support.
+
+# Unresolved questions
+
+We are relying on the community to help us weigh the above tradeoffs. The more data you can provide about the browser makeup of your customers (especially as it affects revenue), the better we can reason whether now is the time to remove IE8 (and possibly IE9) support.
+
+If you cannot share the information publicly, please email whatever information you consider useful to browserusage@emberjs.com. We will keep it in the strictest of confidence.
--- a/text/0045-internet-explorer.md
+++ b/text/0045-internet-explorer.md
@@ -1,5 +1,5 @@
-- Start Date: (fill me in with today's date, YYYY-MM-DD)
-- RFC PR: (leave this empty)
+- Start Date: 2015-06-07
+- RFC PR: https://github.com/emberjs/rfcs/pull/45
 - Ember Issue: (leave this empty)
 
 # Summary


lukemelia:
In Yapp's own products, we have not supported IE8 for a few years. In Yapp Labs' Ember consulting projects, I would estimate that 5-10% of our clients require IE8 support for their applications. After reading this RFC, I am in favor of dropping IE8 support for Ember 2.x. I'm less sure about IE9.

cport1:
IE10+

SebastianEdwards:
:+1: we've already dropped support for anything less than < IE10.

ryanflorence:
Microsoft is dropping support January 2016. Is 8 months worth it?

https://support.microsoft.com/en-us/lifecycle/search?sort=PN&alpha=internet%20explorer

chrisbateman:
@ryanflorence beat me to it. Additionally - IE9 will only be supported on Vista, which has pretty small market share.

workmanw:
At Batterii we haven't supported IE 8 for years. All of our customers are enterprise and we find roughly about 15% of our users are on IE 9. Loss of IE 9 support would be tough for us because it could mean the loss of a few big customers. 

Whatever you decide, the more advanced notice, the better.

stclair:
IE 10+. Our experience dropping support for IE8 and limiting support for IE9 has been overwhelmingly positive.  Customer criticism has been almost nonexistent.

ryanflorence:
> about 15% of our users are on IE 9

Sometimes I've confused "visits" with "users". Just because you have 15% of visits on IE 9 doesn't mean that 15% of your users will be hosed.

Obviously, everybody's customers are different. I've generally been developing software for university students and teachers, its totally normal for them to use multiple browsers. For others, maybe 15% IE9 visits really does mean 15% of users. Just something to consider when discussing browser support.

nchase:
dropping IE8 and 9 sounds like a great idea.

Pradeek:
How long will 1.x with IE8/9 be supported after 2.0 is released?

workmanw:
@ryanflorence Nope, these are individual users. Tracked with Intercom.io. Since we only have larger enterprise customers, individual users often don't have a choice of which browser they use. Internal IT handles locks their computers down. We're starting to see the tides turn on this topic, but a few of our biggest customers are IE 9 only.

brandonweiss:
Yessss. Kill it with fire.

bradly:
IE 8/9 contribute to a significant part of our user's browser usage, but that being said, I think as long as it is clear on not only what you support, but how you plan on depreciating browers in the future all will be well. For instance, Angular dropped IE 8 support when moving from 1.2 to 1.3 which was unexpected and now we have multiple teams already using Angular that have to stick to 1.2.x or move to a different. Not a great place to be in.

ryanlower:
:+1: for dropping

opsb:
The majority of our users are unfortunately using IE9 (corporate clients). Very happy to see support for IE8 dropped though.

RandomEtc:
Dropping IE8 seems compelling. Why not start there and drop `.get` etc?

If there are significant feature gains to be had by dropping IE9 support during 2.x, why not mint Ember 3.0 at that time when those features are the next most compelling thing? Adherence to semantic versioning is honorable, but there are plenty of numbers higher than 2, you won't run out.

aaronbhansen:
We stopped supporting IE 8 and IE9 in our apps after continuing to run into issues. I support dropping both.

ryanbillingsley:
Microsoft in some of their own projects are going to IE10+ for browser matrixes, so at the very least, dropping IE8 is a good idea.

landongn:
IE8, specifically, feels like something that will _painfully_ impact a small subset of the overall community, for an otherwise great gain for the rest of us.  

For new projects, in the real world, I would say a lack of IE8 support would be pretty understandable, and pegging those projects at 1.14.x seems like a logical solution, however imperfect it would be. 

jQuery dropping support for > 1.9 had a near 12% [1] drop in line count, just from those compatibility fixes.  jQuery today (2.1.3) tags in around just under 9400 lines of code, for a reduction of around 1000 lines.  

Ember canary (debug) is sitting around 48,000 lines, and if similar results were typical, would see a net reduction in 5,700 lines. 

Less code is better code, in my book.  It's a big, fat, :+1: from me. 

[1]: http://blog.jquery.com/2013/04/18/jquery-2-0-released/

evrowe:
While we have to continue supporting IE8 at HealthSparq for the time being, based on usage statistics, the front end engineering team has been pushing to drop support for 8 (and potentially 9, which has very low usage for us) for some time. Reading through the RFC makes it clear that the costs of supporting IE8 in Ember 2.0 are higher than the benefits said support would provide.

I'm in favor of dropping it, for the good of the framework (and hopefully the ecosystem), and as it might provide us with the ammunition we need to move our internal browser support matrix forward.

jmurphyau:
Our application is an app to help manage the technical side of contact centres - targeted towards enterprises. I just asked a colleague should we/will we support IE8 and sent him the link and he's response was this:

"We can alway package [app] as a thick client application for these people"

Maybe not everyone is in a position of being able to install a thick client app in an organisation that is large and still uses IE8 - but it's something to worth considering. There are a whole heap of tools that enable packaging web apps as thick client apps

felixrieseberg:
As a Microsoft engineer I have to point out that everything above Windows XP can and should upgrade to at least IE9 today. In addition, it's important to consider that still running Windows XP is outright dangerous, as it is no longer supported from our end.

Personally, I'd even appreciate dropped support. It feels silly to keep technology more than a decade old on life support, especially if we already dropped support.

jrobeson:
i really appreciate how much you folks care about this. I personally want IE8 gone, but the RFC reflects really well on the Ember community. So, I'm sure the best decision will be made in the end. Thanks everyone!

tylr:
I would strongly encourage the ember core team to drop IE8 support.

In January of 2014 IE8 accounted for 0.91% of bustle.com's session traffic. Today, that number is down to 0.11% even though we have continually improved our IE8 support. What kind of organization is making high cost technical decisions to support such a small minority of users familiar with a bad web experience?

Even if you look outside of the ember ecosystem, IE8 support no longer seems practical. Without features like CORs, support imposes impractical methods to modern API and service development. The origin of services and their hosts should be transparent. You could make similar arguments for every modern browser feature not in IE8.

*Alternative IE8 support idea*
I believe a better experience could be provided by way of ember-cli-fastboot. _modern progressive enhancement_???

Why not offer static content to older browsers that aren't capable of the modern web UI interaction? If the argument _for_ IE8 is simply political, this is a viable solution. Especially considering that the value proposition for js frameworks is to provide more interactive tailored product experiences.

I believe the burden of continued support of IE8 would outweigh the cost of looking into alternatives for both ember developers and ember product developers.

tl;dr Just saying ember supports IE8 doesn't mean anything. The ecosystem that makes ember great never actually will. Nobody is building great products with modern UIs that support it. It does not seem practical nor conducive to the ember community to make IE8 support a part of the scope of the framework.

mutewinter:
At Beatport, we only target IE10 and up.

kencheeto:
We don't support IE8 in our Ember app at Zendesk. IE9 is probably on the way out as well.

nanuxbe:
I guess we all want to drop ie8 support. Some of our corporate customers dont't. People have always been reluctant to change and keeping ie8 support is enabling them. So my vote is for dropping ie8-9 support.

ShogunPanda:
Same here. We don't support anything below 10. Drop this ancient.

bakura10:
I do not support IE8 and 9 anymore. Lack of CORS on those browsers is an additional reason why it makes it annoying to use for us. On our end, IE9 usage is also nearly zero.

in all honesty, Ember 1.x is now extremely stable, efficient and has all features I ever need! I can't see why it would be bad for people that require IE8 to simply use the 1.x branch.

As a PHP framework developer we had this same issue of dependency considering how fast new version are released. Most framework dropped support for what were considered "the enterprise version"... and it went smoothly without very few protests.

So +1 for me to drop IE8.

cibernox:
I would drop support for both. The current app I'm building still has a 15% of IE9 users(it targets primary and secondary schools) , but we're considering package the app as a desktop NW app for those schools. Surprisingly system administrators are more willing to install new software that to update the browser version due to the amount of security and parental control that schools have. It might also be a solution for other people in this thread

topaxi:
:+1: on dropping IE8 and IE9 support. We only support up to the latest two browser versions of each browser.

joeruello:
Echoing the common voices in this thread, wewe only support 10+.  Maybe it may be viable to backport some features to 1.15 for people who are stuck needing to support?

 If not, we have found "just install Chrome" to be a simple enough workaround for our clients.

karlfreeman:
Working within the pharmaceutical sector the browser support matrix we proposed for a large ember application was IE9+. This, the client felt, was a suitable compromise given IE8 current lifespan. So I would be in favour of dropping support for IE8 but not IE9. I suspect IE9 will be the minimum for a while.

Perhaps additional weight could be sought be collating browser support across other frameworks for building applications.

A rising tide lifts all boats.

leifcr:
Since fastboot will prerender on the server, that should be good enough for ie8 and ie9. Any clients stuck on ie8 or ie9 should be aware of the security risk by running a browser that won't get updates... 

Dropping at least ie8,and probably ie9 as well.

Fed03:
+1 drop!

trek:
I've privately shared a breakdown of Groupon's browsers % by visits, users, and (most importantly) revenue.

For those reading here, IE8's % across the board falls below our browser support threshold. IE9 is still well above this threshold, but sinking. IE8's low use was surprise. We weren't closely monitoring and the last time we re-evaluated support, IE8 use was still strong. This wasn't even that long ago.

We've added a monthly report about browser %s so we can aggressively cut IE9 when it drops below our support threshold.

I strongly encourage anyone who needs IE Classic support to gather and share these numbers publicly or privately to help inform our support decisions.

kellyselden:
In my large Ember app, I have to support down to IE9 because that's the lowest version media queries will work. The app is 50% internal and 50% external users, and the internal PCs are locked at IE9. My company is dropping IE9 for IE10 at the end of the year I believe.

mixonic:
Good discussion. I've been thinking about this for a while, and would like to share some additional context.

* IE8 persists largely because it is the last viable browser on the XP operating system.
* Microsoft has provided a migration strategy away from IE8 since early 2014 ([source](http://blogs.windows.com/itpro/2014/04/09/internet-explorer-11-now-helps-customers-stay-up-to-date/), [source](http://blogs.msdn.com/b/ie/archive/2014/04/02/stay-up-to-date-with-enterprise-mode-for-internet-explorer-11.aspx)).
* In January 2016, Microsoft will officially end the IE8 "extended support" phase ([source](http://blogs.msdn.com/b/ie/archive/2014/08/07/stay-up-to-date-with-internet-explorer.aspx), [source](http://www.zdnet.com/article/microsoft-to-drop-support-for-older-versions-of-internet-explorer/)). No additional bug or security fixes.
* In June, the number of Ember developers who support IE8 is projected to be less than 10% of the community ([source](http://www.201-created.com/ember-community-survey-2015#browser-requirements)), and I believe this percentage is inflated by developers claiming to support IE8 but without any actual IE8 users.
* Globally, IE8 usage is well below 4% ([source](http://gs.statcounter.com/#browser_version-ww-monthly-201402-201502)). Enterprise numbers certainly differ.

The decision to remove IE8 support does not apply to codebases we work on today, it impacts *new* codebases that would be started after June. I find it nearly unimaginable that a new project would be started (with Ember especially) requiring IE8 six months before the browser is EOLed.

Ember's six-week release schedule opens to door for IE8 supporters to use a deferred-upgrade strategy. They can continue to use Ember 1.14.x (or start on it) until IE8 support is dropped, then presuming all deprecations have been addressed can move forward to 2.x.

I'm in full support of dropping IE8 support in June. Confirming this now (in April) gives enterprise users plenty of notice.

They question of IE9 is more difficult, but I want to cast it in the light of where web development is moving.

Mobile web development is increasingly important ([north america desktop v mobile](http://gs.statcounter.com/#all-comparison-na-monthly-201303-201503), [global desktop v mobile](http://gs.statcounter.com/#all-comparison-ww-monthly-201303-201503)). **My clients demand responsive layouts and mobile-friendly experiences more than they demand legacy IE support**. To support great mobile experiences there are a few requirements:

* Responsive layout
* Offline capability
* Leverage multiple CPUs
* Low latency/bandwidth data transfer
* Fast, smooth rendering and animation

And IE10 is the browser that added solutions for these challenges:

* CSS3, Flexbox and Grid Layout
* IndexedDB, AppCache
* Web Workers
* Websockets
* requestAnimationFrame, Canvas, CSS3

Adding these capabilities to the underlying platform of Ember lets us focus on solutions that work for desktop applications and the mobile web. It places our bet on the future of the web instead of the past.

jamm:
Support of a11y tools can increase number of visitors much more than IE8,9 support :) In any framework. Just saying :)

aethant:
Definite thumbs down on IE8. Take ol' yeller behind the barn, finally.

bsclifton:
What is the biggest obstacle preventing folks who are using XP/IE8 from upgrading/using another browser? Licensing costs for the OS? Genuinely curious for the folks who have said their org is reliant on IE8

douglawrence:
Our products built on Ember.js are currently targeted for US based users. In Q1 2015, usage of both IE 8 and IE 9 is <1% of our total traffic. We dropped product support for IE 9 in early 2014 and as a company, our Customer Success team aggressively campaigns to encourage customers to switch to modern browser technology. A larger concern for us is Android 4.x browser fragmentation which represents a higher volume of traffic than IE 9 on our consumer facing sites.

rtablada:
TLDR;

Not supporting IE8 in 2.0 core would be a good move: it cleans up the API, gives likely performance wins, reduces some of the uglier shims and repetitive code, and reduces the payload size to achieve the current core features.

The newest release should be looking for some BC but also have an eye to reduce the pain of the past. For those that need to support IE8 (maybe 9 too) there should be some sort of community driven LTS for 1.X. Hopefully this could be spear headed by a larger company that still requires this level of backwards compatibility.

---

My experience:

At our company, we are mostly building apps that are long lived dashboards that are replacing desktop applications in the healthcare, financial, and music industries.

We've been surprised by the amount of clients that still do have IE8 installed, but when building long lived applications, we make a comparison in the fact that we are replacing what would have been a desktop app 2 years ago. So when looked at in these terms we say "hey install our app by installing Chrome", the fact that the rest of the internet runs smoother is just a perk.

Otherwise, there has been some effort into wrapping these sorts of apps in Frameless Chromium or Nodewebkit but we haven't had to dive too far down that path.

I also realize that our apps differ than some of the more organic traffic and more web feeling sites like Bustle and a few of the other companies I have talked to using Ember (will not name them since I know a few of them have asked for discretion so I will let them give their own experiences and thoughts).

adrianba:
Working on the IE team, we're doing everything we can to move people off IE8 (and 9 and 10). As has already been noted, support ends in January 2016 but of course that doesn't mean usage stops (as we've seen with WIndows XP).

We're working on providing Enterprise Mode for customers with IE8 applications that otherwise block them from upgrading. Later this year, Windows 7 and 8 customers will be able to upgrade to Windows 10, which comes with IE11 and the new "Project Spartan" browser.

I recommending [moving on from IE8](http://adrianba.net/archive/2015/03/12/itrsquos-time-to-move-on-from-ie8.aspx) and supporting IE11 and "Project Spartan". It's also interesting to figure out how to deal with fragmentation in the [WebKit](http://www.quirksmode.org/blog/archives/2015/02/android_webview.html) and [Chromium](http://www.quirksmode.org/blog/archives/2015/02/counting_chromi.html) space.

lmcardle:
We are in the process of deciding between Angular and Ember. One of the objections for going with Angular is that 2.0 only supports IE11+.  Obviously this is only one factor in the decision, but if Ember dropped IE9 support it would remove that particular value proposition between Ember and Angular.

slindberg:
I have yet to see this brought up in the discussion: as much as I'd love to see [`get()` go away in favor of property descriptors](https://github.com/emberjs/rfcs/blob/drop-ie8/active/0000-internet-explorer.md#eliminate-get), is IE8/9 really the biggest blocker?

It's my understanding that `get()` enables object proxying, which isn't possible with or without `defineProperty` until [`Proxy`](https://kangax.github.io/compat-table/es6/#Proxy) is reliably available. I realize that [`ObjectController` has been officially deprecated](http://emberjs.com/deprecations/v1.x/#toc_objectcontroller), but that [does not currently apply to `ObjectProxy`](https://github.com/emberjs/ember.js/blob/v1.11.1/packages/ember-runtime/lib/mixins/-proxy.js#L79). Controllers aside, proxying is a powerful pattern that we make use of in our app, and one that [Ember Data relies on heavily](https://github.com/emberjs/data/blob/v1.0.0-beta.16.1/packages/ember-data/lib/system/promise-proxies.js#L64). Unless I'm mistaken, IE8 is not the biggest factor in being able to remove `get()`, and should not be considered here.

That said, I believe the remaining arguments in favor of removing IE8 support are overwhelming. The company I work for does not support IE8 at all and IE9 support will likely be dropped soon as well.

cibernox:
When

cibernox:
When is the estimated release date for Ember 3.0?because right now I might feel *slightly* resilent on stop supporting IE9, but I don't see myself supporting it beyond the end of this year. It would be a shame to have to support IE9 until 2017

mixonic:
@cibernox I believe it is mentioned elsewhere, but the assumption is that the 2.x cycle will be about a year and a half long. I personally prefer that the 2.x cycle be shorter than the 1.x cycle (which was about two years).

trek:
> When is the estimated release date for Ember 3.0

This was touched on in the RFC. Our cadence so far has been ~24 months for major revision. I think we'd like to stick in the 16-24 month range if possible.

rtablada:
@slindberg not confirming this but I think that it may be possible to shim Proxy? I think there was also discussion of moving away from implicit Proxy use for performance gains and a closer move toward simplicity.

cibernox:
@trek @mixonic In that case I personally don't see IE9 being in my life beyond this summer. If there is any benefit dropping it I'm :+1: on going only with modern browsers

sandstrom:
**We'd vote for dropping both IE 8 & 9 support (even okay with IE11+ only).**

We have an Ember-application used by mostly medium-sized companies (a few hundred customers in total). Most are Microsoft-centric, the archaic kind. Despite that, we haven't had much issues requiring an evergreen browser or IE11.[1]

Many are vocal about dropping old browsers (managers, users and some developers). But I think the cost of supporting old browsers is often vastly under-estimated. It consumes a tons of time and limits the architecture.[2]

***

[1] Some will complain loudly, but in reality getting them to upgrade to IE11 or use Chrome/FF is fairly easy (a IE8 -> IE11 has tons of other benefits, and there are methods to run specific sites in IE8-mode).

[2] https://github.com/emberjs/ember.js/search?utf8=%E2%9C%93&q=IE+OR+IE8+OR+IE9+OR+IE10&type=Code

rtablada:
@sandstrom I think even more telling is this search (same query but all the time trying to debug IE issues) https://github.com/emberjs/ember.js/search?q=IE+OR+IE8+OR+IE9+OR+IE10&type=Issues&utf8=%E2%9C%93

blimmer:
I think IE8 is definitely fine to drop. We haven't supported it for over a year. We still support IE9, technically, but there are a plethora of known issues with our site on IE9 so it wouldn't be a terribly hard sell to drop support for it with 2.0. :+1:

jrjohnson:
:+1: Especially since as I understand it 1.14 will have all of the goodness that is 2.0.  If I get stuck there for a few months while we work on the politics to officially drop support for IE8 I'm ok with that.  I particularly like the idea of removing browsers at a major version bump.  As the 2.x releases start to add new things we need then it becomes easier to make a business case for dropping IE8 support in order to gain those features.

coding-horror:
I created a topic on the official Ember discussion site to... discuss... this, with some comments based on our http://www.discourse.org usage of Ember.js

http://discuss.emberjs.com/t/ember-may-drop-support-for-ie8-and-ie9-in-2-0/7721

Let's see what the rest of the Ember community has to say, but TL;DR fuck IE8 and IE9 right in their browser-holes. And I am writing this on a Microsoft OS, as a primarily Microsoft-centric dev, so I say that with the deepest love and affection.

jonathanKingston:
For me this was the biggest argument winners of picking Ember in a corporate setup (Granted it isn't the only positive Ember brings to the table at all), stopping an argument with "Yeah but they won't support x% of our customers" is a no brainier for a company.
Angular has always tried to push it is for the corporate world, yet with version 2.x it shot itself in the foot. Some corporates are still using IE7 with XP SP1 which I am not advocating at all however from a marketing perspective I think IE support should be considered as a selling point.

IE8 might make sense too but I'm not really sure of the advantages of dropping 9. At a previous employer 2%+ of business was worth the effort to maintain the extra code. My current firm takes a hard line on security so out of the box we have a brilliant argument to stop using older browsers.

All the arguments here are pro of features that I don't really see being core to ember anyway so it can be down to libraries of implementations to make that choice.

@trek makes a great point about sharing stats (this isn't something I can do for my old employer whom needed the legacy) however we did notice the steep drop off at the end of last year for IE8.

@mixonic makes some great points of why IE10 is better, however surely these features should be progressively enhanced. Although IE8/9 are much older browsers I have found it helps support fringe device browsers on phones,consoles etc that have poor support for additional features like this. I would prefer there to be a way to opt into these features, shim or code branch within the sections of code for non supporting browsers.
The other point around this is that, Ember doesn't have to support all the same features in all browsers. We created a browser grading for support for 'nice to have' features, so IE8 didn't get rounded corners nor were MS filters used as they are slow. Similar degraded support was given for sockets, appcache etc. Users of older browsers don't expect the support for new features, they just want the web to work at the same speed it did way back when their browser was new.

How much are these versions really holding you back? Is it possible to provide a bolt on set of addons to solve the IE issue support (I'm pretty sure it would be easy to re-compile the offending files with working IE replacements and ensure they pass the same tests)?

@maban might be able to back me up here but I am fairly certain console browsers follow the same sort of issues IE8/9 have and perhaps a different approach could be reached for older devices.

Browser share for my previous company matched this article pretty well: http://www.sitepoint.com/browser-trends-january-2015-ie8-usage-triples/
I'm pretty sure bigger corporates might sponsor Ember to support older browsers due to the profit they bring - despite Microsoft wanting them dead.

btecu:
I'm one of those who has to support IE 8 and IE 9. We're mostly .NET but this past week we've released our first feature (page) with Ember. We plan to migrate to Ember, although currently Ember is inside our .NET application.
We have two applications, one for corporate clients and one for general public. The following graphs are for a period of 30 days. As you can guess the general public barely uses IE, however the corporate users do use it. Surprisingly, nobody uses IE 10 or IE 11.

general

corporate

As someone mentioned before, we don't do heavy testing on IE 8 but just make sure it somewhat works and definitely don't worry about how it looks.

I'm an avid supporter of dropping IE 8 from Ember as I plan on do a campaign where I target individual clients and ask them to upgrade to a modern browser. However, I think dropping just IE 8 and not IE 9 would be a mistake, as I look at the styles, polyfills and extra libraries I have to pull in for both IE 8 and IE 9. So I would say **drop both IE 8 and IE 9** or keep both if dropping IE 9 is not possible.

2468ben:
@wycats The RFC does a great job how rough the "Bring Your Own Compatibility" approach to IE8 would be, and I'm also in the camp of dropping support after reading all of trouble you've gone through.

Could you add a "Bring Your Own Compatibility" paragraph for just IE9 support in Ember 2.0, or a TL;DR on how rough it would be?

opsb:
I'm curious about the mention of fastboot as a route to supporting IE8/9. My understanding of fastboot is that it's only responsible for the initial render, after that the app runs as usual in the browser. It seems unlikely that ember is being used to build static sites, how does it solve the problem of running an ember app in unsupported browsers?

lmcardle:
Adding to the FastBoot option as a means of support, in the corporate world we do not always have control of the server which will be used to host the application. In our case, FastBoot would not be an option, as we are not allowed to use Node as the server for deployment.

wycats:
> I'm curious about the mention of fastboot as a route to supporting IE8/9. My understanding of fastboot is that it's only responsible for the initial render, after that the app runs as usual in the browser. It seems unlikely that ember is being used to build static sites, how does it solve the problem of running an ember app in unsupported browsers?

FastBoot would help content-rich sites that were mostly concerned with making sure that their content was accessible on older browsers, but for the most part not full, logged-in experiences.

Harris-Miller:
I would like to see a jQuery 3.0 style compatible version of Ember 2.0 if at all possible. As someone who builds financial websites, our metrics show that a lot of our core users are stuck using IE8 at their jobs. Although we are 8 months away from Microsoft's end of support, it does not mean user won't be stuck with it. Graceful degrades are still our main model for design, as our clients still support the use of IE8.

mikejerome:
I don't see the problem with using Ember 1.x for apps that need to support old browsers. It seems silly to let a legacy browser hold back the entire framework moving forward. Drop IE8 and IE9 for 2.0

stclair:
@mikejerome Well said.

tipsqueal:
Drop IE 8 and 9. As some other people stated, the longer we enable IE 8/9 to stick around the longer it will stick around. Stop enabling them. If someone wants IE 8/9 support they can stick to Ember 1.x

opsb:
There seems to be a lot of support for dropping both IE8/9 (except perhaps from those with enterprise customers - which includes us). It seems IE8 is the blocker with regards to making syntactic changes. If the desire to drop IE9 is purely to free the core team from having to handle IE9 specific bug fixes perhaps that role could be filled by other members of the community.

ernesto-jimenez:
:+1: to dropping support of IE8 and IE9. Specially since it's currently expected Ember 3.0 would take 16-24 months.

I agree with @mixonic, nowadays mobile is more important than IE8 or IE9 for most people, and it's going to be even more important. In my opinion, it would be better to spend that support time looking at mobile: performance, responsive apps, etc. Dropping support for old IE and cutting down the size of Ember by removing jQuery and compatibility code sounds like a good place to start.

As some people already mentioned, those who have customers in IE8 can stick to 1.x for a few months while shepherding customers through the transition. IMHO, as a community it might be more helpful to collect some resources about how to help customers transition: successful arguments, case studies of people who have transitioned, thick clients, etc.

@adrianba, does the IE dev rel team have any resources to help those with big corporate customers make the transition?

> Ember 1.x did not have an official browser support matrix, but we would like to correct this for Ember 2.0.

:+1: :+1: :+1: any hints on what browsers would be in the matrix? I'm specially thinking about mobile browsers.

jdhoek:
Dropping IE8 and IE9 makes complete sense if you consider only non-corporate consumers. Anyone still using these old browsers should by now be nudged towards newer browsers, preferably evergreens such as Firefox and Chrome. Not just because we — as developers — have a lot of good reasons to do so from our end, but also because an up-to-date browser can help in keeping the end user safe from malicious websites and such. Websites like http://browsehappy.com/ are a great help here.

So what remains is our corporate clientèle. What @nanuxbe says is pretty much spot on. As developers we tend to bend over backwards to facilitate legacy browsers, because it seems impossible to get corporate clients to upgrade to newer browsers.

But with IE8 and IE9 we risk repeating the IE6 scenario, where by supporting these old browsers, we are *enabling* them to continue using them until the next big update, where corporate IT invariable pegs the browser to the next fixed version of IE for a couple of years.

There are already plenty of ways to solve this issue that don't require all major libraries and frameworks to keep supporting these browsers way beyond their lifespan. For example you could:

* Keep using Ember 1.x until IE8/IE9 usage is no longer an issue
* Distribute your Ember application with a modern web browser included (e.g. [NW.js](https://github.com/nwjs/nw.js))
* Have IT supply an enterprise friendly modern browser in addition to IE8/IE9 (e.g. [Firefox ESR](https://www.mozilla.org/en-US/firefox/organizations/))

In light of this I believe that Ember should **drop support for IE8 and IE9 in Ember 2.0**.

g13013:
In  my opinion, since Ember 1.x already supports IE8/IE9, for companies that use older browsers, Ember 1.x provides enough features to build excellent apps. for this reason I suggest that Ember 2.x should not keep supporting IE8/IE9.

sandstrom:
Given that: 

1. there seems to be a majority in favor of axing IE 8 & 9 (based on the comments above), 
2. time spent on legacy IE is time that could have been spent making Ember better,
3. IE 10 has a tiny market share (which will be even lower when 2.0 comes out) and
4. the Ember 1.x line will continue to work with old IE.

➣   I argue it makes sense to axe IE 10 too!

***

ie-stats

https://clicky.com/marketshare/global/web-browsers/internet-explorer/

screen shot 2015-04-06 at 14 28 06

http://www.sitepoint.com/browser-trends-january-2015-ie8-usage-triples/

al3x-edge:
:+1: on dropping IE8 and IE9 support. Although we are currently supporting IE9, we will be dropping support for it at the end of the year. I would rather plan on it being removed now with a plan going forward to help users migrate or fill in holes where needed.

ghedamat:
We actively support IE8/9 only for one of our clients as their app targets mainly corporate users and the usage statics are VERY different (> 20%)..

That said I fully agree with the general line of thought, drop support for both IE8 and IE9 and users with corporate need will have to stay on the 1.x branch.

DavidTPate:
:+1: on dropping support for IE8
:-1:  on dropping support for IE9

I work with a lot of large enterprises who for some unknown reason still use IE9. I definitely understand not support IE8 since it is not a HTML5 browser, but I don't see why IE9 would need to be dropped now without any real reasons, but instead just the potential of future use. It's likely appropriate to say that you would support it for ~1 year similar to IE8, I think that is reasonable as hopefully marketshare of IE9 will continue to diminish.

jschilli:
The case for dropping IE8 is much more clear than IE9 - IE8 market share continues to diminish.

Every OS that supports IE8 can upgrade to IE9 or above.  The same is not true for IE9 on Vista.

@ryanflorence suggested that IE9 is EOL a/o 1/12/16 but that is not what the doc says.  If you click through microsoft is saying that only the most recent version of the browser available on a given OS will be actively patched.   The EOL for Vista is 4/11/2017.

https://support.microsoft.com/en-us/lifecycle/search?sort=PN&qid=&alpha=Windows%20Vista&Filter=FilterNO

Personally would love to see support for IE < 11 go away.  Our base still runs ~10% IE9 so dropping IE9 would be more painful.  We believe that number will drop dramatically in the coming year - but IE9 users on Vista have to decamp vista to move forward (again not arguing that they shouldn't)

WRT locking ember apps to 1.x releases to support IE9 that's a bit of a bitter pill.  We've invested a lot to stay current - with the rapidity with which the Ember ecosystem moves, locking into version of Ember.* and staying there creates a huge effort/barrier to getting or staying current.

If IE8 is blocking fundamental progress on core then dropping it becomes more straightforward.  As others have suggested if IE9 support is a question of additional effort then drafting community support for that additional effort makes sense.

nanuxbe:
@jschilli WinXP users can't upgrade past IE8 (and there are more XP users tha Vista users)

jschilli:
Windows XP is EOL whereas Vista doesn’t go EOL until 4/2017.

The EOLness is significant IMO with respect to making the decision.

dehuszar:
+1 for killing IE8 support.  I'd be down for removing IE9 support as well if there was a way to polyfill support back in with an addon.

laynegt:
My previous employer's client base was Fortune 100, and we had planned on dropping IE8 support as soon as early this year.  IE8 usage last summer was 10+% and tangibly dropping every month.  We were also finding that the conversation with corp IT about installing alternate browsers was often pretty painless, so we were confident we could drive usage down farther.  Incidentally, our UX team was a big proponent of dropping IE8 since the experience was so poor.

2468ben:
@dehuszar yeah that's the main unanswered question on this RFC and thread for me: how hard would a polyfill just for IE9 be? IE9 support is clearly not unanimous, but if a polyfill isn't a nightmare, then Ember 2.0 can stay fully lean while products stuck supporting IE9 can focus on a good polyfill.

Even if it's a separately maintained IE9 addon that can't keep compatibility after "Ember 2.5" starts taking advantage of more web tech, it would still get those enterprise products to 2.5 instead of 1.x, and won't bother anyone else's code.

Only if it's not a nightmare.

mixonic:
@wycats talks about a polyfill approach specifically in the RFC. I remain hesitant that it provides any real value. Core team members and contributors would still be expected to support IE9, and moving the code to another repo or addon will not necessarily change the amount of actual effort we expend solving tricky bugs.

FWIW, I'm also uncertain of "corporate stepping up" for IE9. We're seen few large companies interested in sponsoring or contributing to IE6, 7, or 8 support and I would not expect much from them with IE9. Ideas for how to incentivize this are welcome, though we should stay on topic.

andekande:
Since Win7 SP1 shipped IE9 it still is the browser to go for Users that don't know anything about Windows update or have it disabled for some reason (the MS conspiracy).
My company supports IE9 but we see our clients do allow the operation of Firefox as 2nd Browser on XP, or have upgraded to modern IE on Win7.

2468ben:
@mixonic Thanks for the response! The RFC's polyfill section talks only about IE8, not IE9 so I'm glad to hear from someone in the core trenches like you.
I agree that the extra effort (or reliance on a community addon) would be a pain in the ass, I just wondered where it is on like a 1-10 scale, 10 being the pain of IE8 support.
Either way I know that if our enterprise apps had to stay on 1.x, it would still be great, as long as when ember-cli hits 1.0, there's still some way the latest 1.x branch could be supported by it instead of stuck on the alpha track.

ntodd:
Considering the benefits, I'd support dropping IE8.  Our biggest app has 2.4% of users on IE8, but 26.4% of users on IE9 (most of the rest are Mobile Safari).  It's an enterprise app for a medical company and they are *very* slow to update desktop browsers.  Dropping IE9 would be a deal breaker, but IE8 would be fine.

IE9 usage in general may be low, but many corporate environments are still running it.

awaer:
:+1: for dropping IE8 support
:-1:  for dropping IE9 support

rwjblue:
I am fairly opposed to "partial" support or "support by addon".  We either support a browser or we don't. If we support it, all changes and releases are tested against it (which is the only way to ensure a platform does not end up completely broken).

SamSaffron:
"I work with a lot of large enterprises who for some unknown reason still use IE9. I definitely understand not support IE8 since it is not a HTML5 browser, but I don't see why IE9 would need to be dropped now without any real reasons,"

http://discuss.emberjs.com/t/ember-may-drop-support-for-ie8-and-ie9-in-2-0/7721/9

> Microsoft itself is ceasing support of old versions of Internet Explorer in January

Seems completely insane to me to keep supporting IE9 for the next 3 years when Microsoft are dropping support in Jan, the 6 month gap can easily be filled by the old corporate dinosaurs running on an old Ember in the interim.

weegeekps:
I work on software written for the Legal & Finance sector, a sector which is definitely lagging behind others in terms of decommissioning IE8.  A fair sized portion of our customers are still on IE8 due to dependencies on old ActiveX intranet applications that are essential to their business.  _That said_, :+1: for dropping support for IE8.  The maintenance cost is too great for continuing to support IE8 long-term and it is unlikely support for IE8 will continue much longer.  We've already run into this problem with Angular and that is already started a drive to no longer support IE8.

I do have one request re: the jQuery dependencies.  Please try to keep a upgrade path clear for those of us who are using "Ember.$" and "this.$()" in our codebases.  I've got codebases littered with "this.$()" and it will significantly increase the difficulty of upgrading to Ember 2.0 if we have to change every instance of those.

nathanhammond:
Miscellaneous thoughts:

- In the 2006-2007 time-frame jQuery would fix crashing bugs and uncaught errors such that it was *always safe* to include jQuery whether or not that browser was supported by jQuery. If we remove support for older browsers this is an ideal to target because it allows for some fallback code to run which could save us from the "White Screen of Death." A somewhat-graceful failure pattern should be something that we add as we deprecate support for older mainstream browsers (give ourselves a chance to tell users what went wrong).
- Continuous incremental improvements to FastBoot (such that it might allow for some trivial state manipulation) provides hints at a possible way forward for the next time we have this conversation about Safari, Mobile Safari, AOSP, Xbox Browser, or who knows what else.
- I agree with @rwjblue that browser support should not come with a "Bring Your Own Compatibility" approach. Without directly accounting for browsers during library feature development we *will* accidentally break browser compatibility without awareness.
- The users who will likely have the largest problems with IE8/IE9 are selling into numerous corporations (100+) whose IT departments are *well* outside of the scope of their relationship. That appears to be a very small minority in the Ember community.
- Support for IE8 and IE10 ends on January 12, 2016 for user-facing operating systems. Support for IE9 ends on April 11, 2017 for user-facing operating systems. If we elect to maintain support for IE8, IE9, and IE10 those dates should be hard caps as to how long we support them in Ember. (Amusingly, IE9 support is longer than IE10 support.)
- Completely in favor of dropping IE8 and IE9 when wearing my "personal projects" hat.

mupkoo:
> In short, we would be able to implement more features more quickly without the burden of bugs that were first introduced 15 years ago.

This is the part that got me. I would really prefer if Ember's team have more time to focus on new features instead of supporting old software.

rtablada:
@weegeekps I'd like to know more about the jQuery plan but my assumption would be that this.$() would be available if jquery (or any jquery compat library) is available, but would not ship as standard and would not be used for core framework code.

jherdman:
> I'd like to know more about the jQuery plan but my assumption would be that this.$() would be available if jquery (or any jquery compat library) is available

Simply returning the DOM node would be enough, wouldn't it?

weegeekps:
> I'd like to know more about the jQuery plan but my assumption would be that this.$() would be available if jquery (or any jquery compat library) is available, but would not ship as standard and would not be used for core framework code.

@rtablada I would assume so as well, but assumptions are a dangerous thing.  Better to call it out now than wish I had later.

> Simply returning the DOM node would be enough, wouldn't it?

@jherdman No, this.$() should return a jQuery object for the DOM node.  Returning the DOM node directly would end up breaking any code using jQuery functions.

ernesto-jimenez:
@weegeekps @rtablada @jherdman when the time comes, I'm sure the core team will do a graceful transition to remove jQuery, adding deprecation warnings to `this.$` and maybe even starting a RFC if the change is significant.

Until then, we should probably keep this thread about deprecating IE8/IE9 in Ember 2 :)

tonycoco:
Does anyone have any data on government or the banking industry? I'd be curious to find out the percentages there as well.

nathanhammond:
@tonycoco I've provided out-of-band information to the included email address regarding banking.

mgenev:
Drop IE8 and make Ember lighter please :)

zyllorion:
Please commit to maintaining IE8 / IE9 support for 1-2 years more.  Yes I am serious.

We have numerous clients that have no published plans to move away from IE8 and definitely not IE9.  We have no control over this, and these clients are worth significant amounts of revenue. We cannot use an unmaintained Ember for this length of time and will be forced to invest in another framework.

Ember offers something different to others, and the main thing is platform support.  There are dozens of fancy frameworks offering the latest and greatest browser features and we are still a long way from auto-updating browsers in any commercial setting.  Ember was chosen because it had a commitment to providing widespread support along with some internal knowledge, Angular specifically drops IE8 in 1.3 and ext.js did not have an existing knowledge pool for us to draw upon.  A recent migration to ember-cli was essentially a re-write and that was a considerable investment.

So many IE haters are quick to say drop it, quoting percentages of IE users.  Clearly they do not have financial losses by doing so, the only numbers that have any significance to business are monetary.  I would ask that when calculating impact, the statistics do not include non commercial users.  Few like developing for IE so platform abstraction that draws on a global knowledgebase is a key benefit of a framework.

My view is that if using Ember dictates OS dependencies then corporates will jump ship and use ext.js etc.  If Ember loses support from those developers causing development to , will the hobbyists and trend-jumpers like the product so much?

tomdale:
@weegeekps: We would definitely continue to support `this.$()` in components so long as jQuery was present. We rely on this heavily in our apps; even if Ember didn't require jQuery, I don't anticipate we'd drop it from Skylight.

stefanpenner:
> We cannot use an unmaintained Ember for this length of time and will be forced to invest in another framework.

just a heads up, I believe ember is the last of the big frameworks with official support for IE8. 

> Clearly they do not have financial losses by doing so,

I can appreciate this, but be mindful that Ember.js is built and maintained almost entirely by volunteers and to a lesser degree short term contracting engagements.

If there are financial motivations for some institutions maybe a paid LTS version for ember 1.x that continues to support IE8 can exist far into the future.

Ultimately this is a tricky balance. Of the maintainers no-one's applications are still supporting these older browsers, but we go back and invest considerable effort to ensure we maintain compatibility and reduce the potential of new features by reducing to what is technically feasible in an ES3 environment.

OSS works best when you are scratching your own itch. If old IE8 does actually represent a significant portion of our consumers then our active maintainers are not an accurate representation. If some form of support for IE8 continues to exist within the timeline you suggest, we need such a champion.

weegeekps:
> My view is that if using Ember dictates OS dependencies then corporates will jump ship and use ext.js etc. If Ember loses support from those developers causing development to , will the hobbyists and trend-jumpers like the product so much?

@zyllorion As someone coming from a corporation that does software exclusively for the Legal & Finance sector (with Government contracts as well), I would be surprised if you have more than 10% of your customers left on IE8 who can't upgrade to IE9 or use a different browser with your product.  Less than 10% and you're likely losing money if you continue to support the IE8 folks long term given the extra cost of maintenance required to support IE8's quirks.  We've actually had quite a bit of success telling folks "use Firefox or Chrome if you have to keep IE8" with very little complaints.  Occasional grumblings from lawyers but nothing significant as I understand it.

At the end of the day nobody benefits by continuing to support IE8.  The browser is EOL'ed in January 2016, and apps that don't have the latest and greatest features tend to fail financially rather quickly these days, even for well established players in the sector.

tomdale:
One thing we should be clear about:

We intend to make the last release in the 1.x series an LTS release. That means that, while it won't get new features, it will get critical bug fixes and security patches.

If you're in an enterprise environment that still requires IE8, *you probably also want the stability and battle-testedness of the 1.x series anyway*. Sticking with 1.x for enterprise environments gives you the best of both worlds: IE8 support and an API you know won't break.

This allows 2.0 to proceed at a much faster clip, and companies that do not have customers on older browsers will not have to pay a penalty. Even better, because of our strong commitment to Semantic Versioning, once you _do_ drop IE8 (likely before Ember 3.0), you will be able to make the jump from 1.14 to 2.x without any backwards compatibility problems.

2468ben:
@tomdale :+1: thanks for responding to all this, and discussing the possibility of an 1.x LTS.

johnnyshields:
If ie8 and 9 can be supported by fastboot, good enough for me

yonjah:
From the many people commented here about there company still requires support of IE8, how many have actually using latest stable 1.x branch ? 
At least from my experience Ember 1.11 IE8 support is too limited to be useful ( debug build completely breaks [10519](https://github.com/emberjs/ember.js/issues/10519), and alto core components work some tags  might break components IE8 support [10520](https://github.com/emberjs/ember.js/issues/10520) )

I guess not being able to debug code without patching ember core is the most serious issue, but the fact that it's so easy to create a component that breaks IE8 support doesn't help much either.

Since our app targets enterprise clients we assume we'll have to support IE8 (we are still in the prototype stage), but I think ember 1.x had two conflicting declarations -
1. pushing the web forward by aligning to standard like ES6 and Components.
2. keeping support for IE8

I think if ember really want to help pushing the web forward dropping default support for IE8-9 is a must, but since ember promise to keep IE8 support there should be a straight forward way to opt-in for IE8.
I think keeping the 1.x branch LTS is a great idea but there are some thing that should have extra consideration when doing so -
1. ember-cli latest should work with both versions, since ember-cli is the de facto  way to build emebr apps, it should support Emebr 1.x branch and IE8 build even if 2.x doesn't
2. addons should be built to support Ember 1.x and IE8, since this is generally up to the addon developer, official documents should make it clear where a feature or pattern might break Ember 1.x or IE8 support and what needs to be done to support both.

I think the ultimate implementation would be if you could load Ember 2.x for modern browser and Ember 1.x for IE8-9 for the same ember app, this will not only improve developing apps that support IE8 but will make the upgrade from 1.x latest to 2.x latest transparent

johnnyshields:
I'm planning to release a public-facing app that uses FastBoot. Until I read about FastBoot I was planing on using a traditional server backend, but FastBoot creates a new possibility I wouldn't have considered otherwise. As long as the "server mode" of FastBoot works on IE8/9 I'm happy even if the Ember JS itself never loads... I'm based in Asia and IE8/9 seems to be lingering here more than in US.

sandstrom:
@yonjah just to chime in, our customers are all enterprise (of the old, conservative kind).

It's actually not a major issue to only support the recent version of IE (and if they really need to keep an old IE version, they are typically fine with using Chrome/Firefox in tandem).

If I were you I'd target IE 10 (or 11) plus. You'll save a ton of time and I don't think it'll affect sales.

yonjah:
@sandstrom A few weeks ago we decided were not going to keep testing IE8 for our first release, but there still very high chances we'll need to support it after the initial release.

We are targeting very limited and conservative audience so one client requiring IE8 support will mean we'll have to add support for it.

I hope your right but we wont be able to know that until we officially go online

Panman8201:
Speaking from the K-12 public education sector, it would be safe to drop IE8/9 support. At the school districts I've worked at we've been fairly flexible on installing browsers. Any computers holding onto IE8/9 due to other software needs typically have an alternate browser installed. In fact, most computers have an alternate browser installed anyway. So, +1 for dropping IE8/9. I'd hate to see it hold back Embers potential.

tomdale:
Hey folks,

This has been an amazing discussion. Big thanks to everyone for contributing your perspectives.

After reviewing this thread, it seems clear that the vast majority of Ember users who have responded, including people working at large corporations, are comfortable with dropping IE8 support in Ember 2.0. On the other hand, while there is enormous support for dropping IE9 support as well, a number of people still rely on support for IE9, and the benefits of dropping IE9 in Ember 2.0 are not as strong.

After reviewing this thread, many in-person conversations with Ember users in large companies, and reviewing the private data sent to us via email, **we have decided that Ember 2.0 will support IE9+**.

So how are we going to manage this transition, and what should you do if your business still requires IE8 support for the time being?

### 1.13 with Extended Browser Support

The core team will continue to periodically release point releases in the 1.13 series to patch security bugs and browser compatibility issues, including issues in IE8.

No new features will be added, and we should be clear that **we do not intend people to stay on this release unless they must support IE8**. Our Semantic Version guarantees mean that the vast majority of the community should migrate to the 2.x series as soon as possible.

It is important to note that Ember 1.13 will come with deprecation warnings for everything that we will break in Ember 2.0. As a result, if you are running Ember 1.13 without any deprecation warnings, you should be able to easily upgrade to Ember 2.0. And because of the Semantic Versioning guarantees in the Ember 2.x series, it should be relatively simple to upgrade from Ember 1.13 to the most recent version of Ember 2.x when you are able to drop IE8 support.

For example, imagine you build the Ember app for Big Widget Enterprise Co. that requires IE8 support. You upgrade to 1.13 (the last release in the 1.x series) and, over time, refactor code to eliminate all deprecation warnings. Periodically, you apply 1.13 patch releases to maintain browser compatibility and to fix potential security issues.

Then, in April of 2016, management decides that enough customers have moved off IE8 that you no longer need to support it. At that time, Ember 2.6 will be the most recent stable release. Because 1.13 without deprecation warnings is forwards-compatible with Ember 2.6, you can upgrade from 1.13 to 2.6 with little hassle.

With the integration of Ember CLI and Ember Data into the Semantic Versioning guarantees, many of your dependencies will follow a similar upgrade path.

#### Ecosystem

Of course, the above guarantees only apply to Ember, Ember Data, Ember CLI, and the rest of the core-supported packages. Addon authors are free to define their own support matrices. We encourage those who depend on older browsers to contribute back by submitting PRs to the addons they use with compatibility patches. Likewise, we encourage authors of existing addons to work with users to offer a browser compatibility matrix as close to the core projects as possible.

If you require support for IE8 (and as a result, Ember 1.13), make sure to make your voice heard across the addon ecosystem.

That said, you should expect that **new** addons that come out after Ember 2.0 will not target Ember 1.13, and you should factor that into your decision to remain on the 1.13 Extended Browser Support release of Ember.

### FastBoot

FastBoot, our effort to bring server-side rendering to all Ember apps, is designed to offer even users with slow, low-feature browsers a fast experience. While most people think of this as a benefit to mobile users, IE8 certainly qualifies as a slow, low-feature browser.

Because Ember applications are written "route first", any idiomatic Ember content app that uses links as the primary mode of navigation will be able to provide a passable experience for users with an unsupported version of JavaScript, or no JavaScript at all.

It is worth noting that FastBoot, in the medium term, will have good support for read-only content sites. However, while it is possible to support forms pretty easily, forms without JavaScript (using cookie authentication) introduce the prospect of CSRF attacks. A good solution for FastBoot forms that is also secure is probably a longer-term project. We would encourage the community to experiment with a secure approach to forms that works with FastBoot.

### jQuery Compatibility

In our RFC, we mentioned that dropping IE8 will give us the opportunity to remove jQuery as a strict dependency. We should have been clearer that we have no intent to remove the Ember APIs that delegate to jQuery (such a `Ember.$` and `this.$()` inside components).

Because these APIs will remain in 2.0, both for ease of upgrade and because we have not yet made the jQuery dependency optional, Semantic Versioning prohibits us from removing them until at least Ember 3.0.

On a personal note, we rely on jQuery heavily in our own apps. We think it's a great library that remains hugely valuable to smooth over clunky DOM APIs and browser quirks (even in modern browsers). For those users who need the absolute smallest payload size, we don't want to saddle you with a dependency that you don't need. But we expect the majority of users to continue using jQuery, and we have no plans to remove the Ember/jQuery integration at this time.

Thank you again for everyone who took the time to help us make this decision, and thank you so much for being a part of the Ember community.

PR #46, opened 2015-04-09
Registry / Container reform

dgeb:
[RFC](https://github.com/dgeb/rfcs/blob/registry-reform/active/0000-registry-reform.md)

### DIFF
--- a/active/0000-registry-reform.md
+++ b/active/0000-registry-reform.md
@@ -1,0 +1,109 @@
+- Start Date: 2015-04-09
+- RFC PR:
+- Ember Issue:
+
+# Summary
+
+Fully encapsulate and privatize the `Container` and `Registry` classes by
+exposing a select subset of public methods on `Application` and
+`ApplicationInstance`.
+
+# Motivation
+
+The `Container` and `Registry` classes currently lead a confusing life of
+semi-private exclusion within Ember applications. They are undocumented
+publicly but not fully private either, as knowledge of their particulars is
+required for developing both initializers and unit tests. This situation has
+become untenable as the new `Registry` class has been extracted from
+`Container`, and the complexity of their usage has grown across
+`Application` and `ApplicationInstance` classes.
+
+We can bring sanity to this situation by continuing the work started at the
+`Application` level to expose methods such as `register` and `inject` from the
+internally maintained `Registry`.
+
+Furthermore, once `Container` and `Registry` are fully private, their
+architecture and documentation can be cleaned up. For instance, a
+`Container` can freely reference its associated `Registry` as `registry`
+rather than `_registry`, as it can be assumed that only framework developers
+will reference this property.
+
+# Detailed design
+
+`Application` will expose the following methods from its internally maintained
+registry:
+
+* `register`
+* `inject`
+* `registerOptions` - mapped to `Registry#options`
+* `registerOptionsForType` - mapped to `Registry#optionsForType`
+
+`ApplicationInstance` will also expose the the same methods. However, these
+methods will be exposed from its own internally maintained registry, which
+has the associated `Application`'s registry configured as a "fall back". No
+direct path will be provided from the `ApplicationInstance` to the
+`Application`'s registry.
+
+`ApplicationInstance` will also expose the following methods from its
+internally maintained container:
+
+* `lookup`
+* `lookupFactory`
+
+`ApplicationInstance` will cease exposing `container`, `registry`, and
+`applicationRegistry` publicly.
+
+`Application` initializers will receive a single argument to `initialize`:
+`application`.
+
+Likewise, `ApplicationInstance` initializers will receive a single argument
+to `initialize`: `applicationInstance`.
+
+`Container` and `Registry` will be made fully private and documented as
+such. Each `Container` will freely reference its associated `Registry` as
+`registry` rather than `_registry`.
+
+[ember-test-helpers](https://github.com/switchfly/ember-test-helpers)
+will provide an `isolatedApplicationInstance` method instead of an
+`isolatedContainer` for unit testing. A mechanism will be developed to specify
+which initializers should be engaged in the initialization of this instance.
+In this way, we can avoid duplication of registration logic, as is currently
+done in a most un-DRY manner in the [isolatedContainer](https://github.com/switchfly/ember-test-helpers/blob/master/lib/ember-test-helpers/isolated-container.js#L56-L79).
+
+# Drawbacks
+
+This refactor will require maintaining backwards compatibility and
+deprecation warnings until Ember 2.0. This will temporarily increase
+internal code complexity and file sizes.
+
+# Alternatives
+
+The obvious alternative is to make `Container` and `Registry` fully public
+and documented. An application's registry would be available as a `registry`
+property. An application instance's container would remain available as
+`container`.
+
+We could still pass an `Application` into application initializers
+and an `ApplicationInstance` into application instance initializers.
+
+If this alternative is taken, I would suggest that `Application` should
+deprecate `register` and `inject` in favor of calling the equivalents on its
+public `registry`.
+
+Regardless of which alternative is chosen, we should ensure that the public
+aspects of container and registry usage are well documented.
+
+# Unresolved questions
+
+* Are the public methods listed above sufficient or should any others be
+exposed?
+
+* What mechanism should be used to engage initializers in unit and
+integration tests? Should test modules simply have an `initializers` array,
+similar to the current `needs` array?
+
+* Given the semi-private nature of containers and registries, we may not need
+to worry about semver for deprecations. However, we should be good citizens
+and properly deprecate as much as possible. Some real world use cases in
+initializers will no doubt be a surprise, so we need to tread carefully.
+
--- a/text/0046-registry-reform.md
+++ b/text/0046-registry-reform.md
@@ -1,6 +1,6 @@
 - Start Date: 2015-04-09
-- RFC PR:
-- Ember Issue:
+- RFC PR: https://github.com/emberjs/rfcs/pull/46
+- Ember Issue: https://github.com/emberjs/ember.js/pull/11440
 
 # Summary
 
@@ -106,4 +106,3 @@ similar to the current `needs` array?
 to worry about semver for deprecations. However, we should be good citizens
 and properly deprecate as much as possible. Some real world use cases in
 initializers will no doubt be a surprise, so we need to tread carefully.
-


active/0000-registry-reform.md
```diff
@@ -0,0 +1,109 @@
...
+`Container` and `Registry` will be made fully private and documented as
+such. Each `Container` will freely reference its associated `Registry` as
+`registry` rather than `_registry`.
+
+[ember-test-helpers](https://github.com/switchfly/ember-test-helpers)
+will provide an `isolatedApplicationInstance` method instead of an
+`isolatedContainer` for unit testing. A mechanism will be developed to specify
+which initializers should be engaged in the initialization of this instance.
+In this way, we can avoid duplication of registration logic, as is currently
+done in a most un-DRY manner in the [isolatedContainer](https://github.com/switchfly/ember-test-helpers/blob/master/lib/ember-test-helpers/isolated-container.js#L56-L79).
```
mixonic: For Ember-Data we use `DS._setupContainer`:

  * https://github.com/emberjs/data/blob/87dec3e56b6b1744c49103417cc8cacc8f1edb23/packages/ember-data/lib/setup-container.js#L7

If Ember only needs to register/inject, then I would prefer such a hook since they can be easily layered onto a single registry in an arbitrary order.

However I note that you suggest the name be `isolatedApplicationInstance` - does this imply you will return a real instance and not simply a container with a configured registry?

dgeb: Yes. If `Container` is private and only accessible through an `ApplicationInstance`, then I believe that's the level at which it should be available for testing.

workmanw:
With the recent container / registry split we were burned pretty badly with our test helpers. We had created quite a few helpers to do proper testing setup/mocking/teardown/etc. I personally am a little weary of the idea of isolating these given that interaction with the container had previously been encouraged for testing and other initializers.

I guess I have no real concrete reason as to this shouldn't be done, other than I don't really want to refactor all my test helpers again.

tomdale:
Big, big :+1:. I think this proposal kills several :bird:s with one stone:

1. Simplifies initializers: it was mega confusing knowing whether to call register/inject on the application or the container/registry.
2. Gives us a public API we can support in a forwards-compatible way, while giving us the freedom to continue to tweak the container and registry implementations.

> I personally am a little weary of the idea of isolating these given that interaction with the container had previously been encouraged for testing and other initializers.

I can't speak for everyone, of course, but @wycats and I have been very openly against anyone treating the container API as public. There are cases where you have to use it, but there is a reason it is marked private. This proposal allows us to (finally) provide an API we can encourage people to use.

workmanw:
@tomdale Certainly my primary use case is around testing. Using containers to initialize and mock stuff. But also, it is one of the arguments provided when creating an initializer. And for a period of time, before services people were encourage to used injection this way. This is one of dozens examples that come up when you Google ember dependency injection. http://ember.zone/ember-application-initializers/

Any ways, just to make my position clear, I'm not against it. Just pointing out it might be a source of pain for some folks.

tomdale:
@workmanw Point taken. With Dan's RFC, assuming you were indeed sticking to initializers and coloring within the lines, the refactor is *extremely* mechanical: just update all of your initializers and replace `container` with `applicationInstance`.

igorT:
`container.has`/`registry.has` have been useful for me in the past, so might be good to expose those as well. They don't seem particularly hard to maintain.

dgeb:
> container.has/registry.has have been useful for me in the past, so might be good to expose those as well. They don't seem particularly hard to maintain.

Seems reasonable. Perhaps `has` could be exposed as `hasRegistration`.

dgeb:
@tomdale Thanks for your support of this proposal.

@workmanw I'm sorry that the churn in this area has already affected you. I initiated much of this churn with the Registry / Container split as part of the SSR refactor. That was a necessary precondition to the Application / ApplicationInstance split that @tomdale and @wycats performed to enable SSR. Only now that their work is nearing completion is it possible to come back to rethink Registry and Container access. I hope the end result of all this work is a clean set of interfaces that shield developers from any complexity introduced by these refactors. And since these interfaces will finally be thoroughly public, we can avoid similar churn and breakage in the future.

active/0000-registry-reform.md
```diff
@@ -0,0 +1,109 @@
...
+
+# Drawbacks
+
+This refactor will require maintaining backwards compatibility and
+deprecation warnings until Ember 2.0. This will temporarily increase
+internal code complexity and file sizes.
+
+# Alternatives
+
+The obvious alternative is to make `Container` and `Registry` fully public
```
stefanpenner: most of the in-app usages of container are the result of not having:

1. class level injection rules (not global injection rules, which don't inherit)
2. factory injections are backwards, instances should get references to factories and not the other way around.

beyond that, there exist dynamic situations where `has/lookupFactory` should be exposed, but `lookup` should never be exposed as this should have been the result of an injection.

active/0000-registry-reform.md
```diff
@@ -0,0 +1,109 @@
...
+
+Regardless of which alternative is chosen, we should ensure that the public
+aspects of container and registry usage are well documented.
+
+# Unresolved questions
+
+* Are the public methods listed above sufficient or should any others be
+exposed?
+
+* What mechanism should be used to engage initializers in unit and
```
stefanpenner: Unit don't get initializers. By definition this is then no-longer a unit test. Unit level tests can get collaborates provided as constructor Args, but should have no container awareness. My above comment should help remedy most use-cases where users app code violated this.

dgeb: All of Ember's "unit" tests rely on the concept of a container that is initialized. The [isolatedContainer](https://github.com/switchfly/ember-test-helpers/blob/master/lib/ember-test-helpers/isolated-container.js) is not truly "isolated", but relies on registrations performed by Ember and even Ember Data. I'm sure we can all agree that the code in `isolatedContainer` is un-DRY and a temporary hack, so let's solve this problem more elegantly.

As [the discussion about integration vs. unit testing components](https://github.com/switchfly/ember-test-helpers/pull/38) makes clear, we need a broader discussion about dependency injection, what's exposed to tests, and ember-cli's conventions regarding integration and unit tests. This goes beyond the scope of this RFC, but definitely overlaps it.

@stefanpenner I'll ping you soon to try to set up that overdue hangout / discussion.

drogus:
I'm not sure if this is something that fits the scope of this RFC, if that's the case just ignore me.

Lately I've been digging into testing story for Ember.js and one thing bothers me about testing an application with services. At the moment there's no easy way to inject mocked services during acceptance testing, or at least there's no way that I know of. `Ember.inject.service` looks up factory on a container, which will try to resolve it, so even if a user registers a mock service in the tests setup, it won't be picked up.

It would be great if new public API allowed something like this:

```
var application = startApp();
var mockService = Ember.Object.extend({
  fetchPosts: function() {
    return [{ title: 'a post' }];
  }
});
application.register('service:storage', mockStorage);
```

This would make acceptance tests use a mocked service.

There are multiple ways to handle such behaviour, but all that I know of require preparing code for substitution (like creating adapters in ember-data).

technomage:
Ideally we will have a way to register a replacement service as described above, using a public API. 

In testing comma components this has proven very necessary to isolate issues in the front end vs. the back end code.

technomage:
That was comms components.

igorT:
@drogus I am also very interested in the acceptance testing story. I think there are at least two blocking issues:

1. `startApp` helper in acceptance tests returns an `Application` and not an `ApplicationInstance`, so its impossible to overwrite a registration, you can just register in the fallback case.
2. `resolver` always has precedence over the registered classes https://github.com/emberjs/ember.js/issues/11174

drogus:
@igorT 

Yeah, the only workaround I found to overcome this issues is to reopen resolver:

```javascript
application.registry.resolver.__resolver__.reopen({
  resolve: function(fullName) {
    if(fullName === 'service:storage') {
      return mockService;
    } else {
      return this._super.apply(this, arguments);
    }
  }
});
```

which is clearly not very elegant and will break with any bigger internal changes.

dgeb:
Here's one way to solve the acceptance testing story:

* The `resolver` is set only on the `Application`, not the `ApplicationInstance`
* Acceptance tests receive an `ApplicationInstance` instead of an `Application`
* Custom registrations can be made on the `ApplicationInstance`.

In this way, registrations will be resolved in order by:

* The application instance's registry
* The application's resolver
* The application's registry

In this way, mocks can be registered with the application instance and those registrations will take precedent when resolved.

active/0000-registry-reform.md
```diff
@@ -0,0 +1,109 @@
...
+methods will be exposed from its own internally maintained registry, which
+has the associated `Application`'s registry configured as a "fall back". No
+direct path will be provided from the `ApplicationInstance` to the
+`Application`'s registry.
+
+`ApplicationInstance` will also expose the following methods from its
+internally maintained container:
+
+* `lookup`
+* `lookupFactory`
```
bf4: Great, so code in ember-islands that finds `App.DocPrinterComponent` would change as below?

```diff
-var container = App.__container__;
-var componentLookup = container.lookup('component-lookup:main');
-var component = componentLookup.lookupFactory('doc-printer',  container);
+var component = App.lookupFactory('doc-printer');
```

igorT:
@dgeb that means that there is parallel `register` on the container that delegates to the resolver if there is one but caches locally if there isn't?

dgeb:
@igorT there is no (non-deprecated) `register` method on the container. `register` exists only on registries.

There are two registries now: one on the application instance and one on the application. Each has a `register` method for custom registrations. 

When resolving, the instance's registry "falls back" to the application's registry if it can't resolve. If a resolver is associated with either registry, then it will be consulted before the custom registrations in that registry.

The solution here (which I'm about to PR) is to skip setting the resolver on the instance's registry and only set it on the application's registry. This allows custom registrations made through the instance's registry to take precedence over both the resolver and registrations made on the application's registry.

igorT:
My bad I misread `The resolver is set only on the Application, not the ApplicationInstance` as `The registry is set only on the Application, not the ApplicationInstance`

active/0000-registry-reform.md
bf4: Or would it be like

```javascript
App.instanceInitializer({

  initialize: function(instance) {
    instance.container.lookupFactory('doc-printer');
  }
});
```

(h/t al3xnull in slack https://embercommunity.slack.com/archives/needhelp/p1434479693008755 )

dgeb: Sorry to miss your earlier question. So the `lookup` and `lookupFactory` methods would be exposed directly on the instance, in the same way as `register` and `inject`:

```
App.instanceInitializer({

  initialize: function(instance) {
    instance.lookupFactory('doc-printer');
  }
});
```

(Note: @stefanpenner has concerns about `lookupFactory` as it's currently implemented. It may be replaced entirely for 2.0)

bf4: alrighty, thanks

dgeb:
Implemented and merged via https://github.com/emberjs/ember.js/pull/11440

Should this RFC just be closed now?

mixonic:
Yes :-)

The filename should be changed to `text/0046-registry-reform.md` and the `RFC PR` and `Ember Issue` should be added at the top of the text.

dgeb:
Thanks @mixonic! It's ready for review now.

PR #47, opened 2015-04-13
Named Template Blocks

taras:
[RFC](https://github.com/embersherpa/rfcs-1/blob/master/active/0003-named-template-blocks.md)

Ideas are @krisselden's, I wrote it up to open the discussion.

### DIFF
--- a/active/0003-named-template-blocks.md
+++ b/active/0003-named-template-blocks.md
@@ -50,36 +50,38 @@ Here is what that might look like,
 {{#ember-table}}
   {{^header as |name|}}
   	{{name}}
-  {{^}}
   {{^cell as |value|}}
   	{{value}}
   {{^}}
+  Nothing to show.
 {{/ember-table}}
 ```
+
 Specifying a named block inside of the component's block template would override default implementation of that block inside of the component's layout.
 
 # Detailed design
 
-## New Handlebars syntax
+## Expand ^ syntax
 
-To enable this functionality we would need to introduce a new syntax that would deferentiate a  named template block from a regular helper with a block. This would be done with the introduction of `^` syntax that indicates that this is a named template block. 
+Currently, Handlebars implements `^` which means `else`. Infact, `else` is aliased to `^`. We would expand this syntax to allow named portions of template. 
 
 Here is an example of what a template with this syntax might look like.
 
 ```
-{{#table-component}}
+{{#table-component items as |item|}}
+	// default block
 	{{^header}}
-		Header Content goes here
-	{{^}}
+	// header content
 	{{^footer}}
-		Footer goes here
+	// footer content
 	{{^}}
+	// empty view
 {{/table-component}}
 ```
 
 ## Refactor `{{else}}` helper
 
-One possible implementation would be to refactor `{{else}}` implementation into a more generic implementation that allows the name of the helper to be modified. 
+One possible implementation would be to refactor `{{^}}` implementation into a more generic implementation that allows the name of the helper to be modified. 
 
 Handlebars has several built in blocks that are availble on helper's `options` argument,  namely `options.fn` & `options.inverse`. These would be changed to `options.blocks.default` & `options.blocks.inverse` respectively. Every other named template block would be available on `options.blocks` hash. The above example would have `options.blocks.header` & `options.blocks.footer` in addition to it's default blocks.
 
@@ -101,9 +103,9 @@ The component must be able to determine programmatically if it should consume it
 
 Similar pattern can be followed to specify the ability to customize body and footer.
 
-## Named blocks can yield
+## Named blocks have block params
 
-Named template blocks need to be able to yield values into the scope in the same way as regular template blocks can. This would allow the component to expose template friendly values to be used in the template block.
+Named template blocks need to be able to receive block params and yield values into the scope in the same way as regular template blocks can. This would allow the component to expose template friendly values to be used in the template block.
 
 ## Named blocks must be portable
 
@@ -116,7 +118,6 @@ Why should we *not* do this?
 # Alternatives
 
 What other designs have been considered? What is the impact of not doing this?
-`<table-header></table-header>`
 
 # Unresolved questions
 
--- a/active/0003-named-template-blocks.md
+++ b/active/0003-named-template-blocks.md
@@ -1,0 +1,123 @@
+- Start Date: 2015/04/12
+- RFC PR: 
+- ember-cli Issue: 
+
+# Summary
+
+Introduce a new Handlebars syntax that would denote beginning and end of a template block. 
+
+# Motivation
+
+In Ember 2.0, we're moving towards components becoming primary mechanisms for encapsulating functionality with templates being the mechanism that wires these components together. It would be helpful to allow developers the ability to customize certain portion of component's template from it's block template without having to extend a component or overwrite it's layout.
+
+Currently, we have the ability to do this but it's limited to `{{else}}` helper.
+
+For example, we can customize what `{{each}}` helper presents when the array passed to it is empty.
+
+```
+{{#each model as |item|}}
+	{{item.name}}
+{{else}}
+  	No items are available.
+{{/each}}
+```
+
+`{{if}}` helper has the same behaviour.
+
+We are currently lacking the ability to specify a custom template block that we can consume inside of the component or a helper. This RFC proposes that we introduce the ability to specify custom named template blocks that can be used to customize the default behaviour of a component or a helper. 
+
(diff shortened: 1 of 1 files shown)


active/0003-named-template-blocks.md
```diff
@@ -0,0 +1,126 @@
...
+{{view Ember.Table.ScrollContainer}}
+{{view Ember.Table.ColumnSortableIndicator}}
+```
+
+To customize any of these areas, the developer has to extend the component, change specify a new layout and make changes inside of this new layout. If we had the ability to specify custom named template blocks then we could allow the developer to customize one of these areas from the template where the component is being used. 
+
+Here is what that might look like,
+
+```
+{{#ember-table}}
```
stefanpenner: an alternative syntax.

```hbs
<ember-table as |t|>
  <t.header as |name|> {{name}} </t.header>
  <t.cell as |value|> {{value}} </t.cell>
</ember-table>
```

active/0003-named-template-blocks.md
```diff
@@ -0,0 +1,126 @@
...
+
+Named blocks will have block params and will receive values with a new `{{yield-to}}` helper. `{{yield-to}}` will take name of a block as a first parameter and yield properties as `{{yield}}` does. This will allow the component to expose template friendly values to be used in the named blocks.
+
+```{{yield-to 'header' headerContent}}``` will yield `headerContent` to `header` block.
+
+# Drawbacks
+
+Why should we *not* do this?
+
+# Alternatives
```
stefanpenner: `yield-to` has no idea what to do if there are multiple `named` template blocks, imagine the following.

```hbs
<my-form as |f|>

  <f.input> </f.input>
  <f.input> </f.input>
  <f.input> </f.input>

  <f.save> Save</f.save>
</my-component>
```

or a table example:

```js
<my-table as |t|>
  <t.column> my first column </t.column>
  <t.column> my second column </t.column>
  <t.column> my thirds column </t.column>
</my-table>

```

active/0003-named-template-blocks.md
mixonic: Stef's suggestion looks very similar to the proposed nested helper syntax :-/

balinterdi: I also noticed that this RFC and https://github.com/emberjs/rfcs/pull/43 aim to resolve the same issue from different aspects. Is that what you refer to, @mixonic ?

taras: @stefanpenner one of the goals of the proposal is to allow the user to overwrite a portion of the default layout. Does what you're suggesting allow for this?

active/0003-named-template-blocks.md
taras: > yield-to has no idea what to do if there are multiple named template blocks, imagine the following.

`yield-to` would not have to deal with multiple scopes because it's scope is limited to the layout. Here is an example,

in *app/components/ember-table.hbs*
```
{{#if blocks.header}}
  {{yield-to 'header' header}}
{{else}}
  {{#each header as |column|}}
    {{#if blocks.header-column}}
      {{yield-to 'header-column' column}}
    {{else}}
      {{column.name}}
    {{/if}}
  {{/each}}
{{/if}}
{{#if blocks.rows}}
  {{yield-to 'rows' rows}}
{{else}}
  {{#each rows as |row|}}
    {{if blocks.row}}
         {{yield-to 'row' row}}
      {{else}}
         {{#each row as |column|}}
           {{#if blocks.column}}
             {{yield-to 'column' column}}
           {{else}}
             {{column.name}}
           {{/if}}
        {{/each}}
    {{/if}}
  {{/each}}
{{/if}}
```

When using the component, the user can specify a custom `header-column` and `column` blocks.

```
{{#ember-table}}
  {{^header-column as |column|}}
    <button {{action 'sort' column.sortKey}}>{{column.name}}</button>
  {{^column as |column|}}
    {{column.value}}
{{/ember-table}}
```

technomage:
Stef's syntax is easier to follow in terms of 'where is this coming from'. The ^ strikes me as trying to save keystrokes, but at the cost of clarity. Explicit data/control flow is better when reading the template IMO.

taras:
@technomage `^` is syntax that's built into Handlebars. `{{else}}` is an alias to `{{^}}`.

active/0003-named-template-blocks.md
stefanpenner: > Stef's suggestion looks very similar to the proposed nested helper syntax :-/

yes, i would like to unify these two things.

krisselden: @stefanpenner this is a different use case and a different RFC, yielded helpers are about the block placing the content.   Named templates are about the parent controlling the rendering via its layout.  We need to keep these things separate.

ef4:
I strongly favor the block params solution over named blocks, for all the reason @stefanpenner mentioned.

I disagree that it handles a different case. Yielded helpers are general enough that users can actually implement their own named blocks when they need that:

```hbs
{{#my-parent as |p|}}
  {{#p.contentFor "header"}}
    Put this in the header
  {{/p.contentFor}}
  {{#p.contentFor "footer"}}
    Put this in the footer.
  {{/p.contentFor}}
{{/my-parent}}
```

taras:
@ef4 can you please describe how your code example works or what it does?

ef4:
In my example, the user is overriding default blocks in the parent's layout, just like in the RFC. I'll pick apart a slightly different example, because I think I can make the syntax even nicer:

```handlebars
{{#my-parent as |customize|}}
  {{#customize "header"}}
    My customized header
  {{/customize}}
  Hello world
{{/my-parent}}
```

Once you have the ability to yield and pass helpers (or components, which is not so different), you don't need any other special support from the framework, and can do all of the rest in normal components. Which means less syntax for people to learn. One possible example for the parent's template would be:

```handlebars
{{#customizable-content use=header}}
  Default header content goes here.
{{/customizable-content}}
...other parent layout stuff...
{{yield customize}}
...other parent layout stuff...
{{#customizable-content use=footer}}
  Default footer content goes here
{{/customizable-content}}

```

`customize` is a helper defined by the parent. Whatever blocks you pass to it get set as properties on the parent. A critical step here is that a block is treated as just a normal value that can be passed around and ultimately bound into the DOM where you want it to go. This is not a big step -- under the hood they already work that way. In fact, we probably don't even need a separate type for "block", because it has the same API as a yieldable helper.

`customizable-content` is just a normal component. (We could make the syntax even better if we let components take positional arguments, eliminating the `use=`. That's a feature I definitely intend to implement for 2.0. But for now I'm showing normal component syntax.)

`customizable-content`'s template would be:

```handlebars
{{#if use}}
  {{use}}
{{else}}
  {{yield}}
{{/if}}
```

So putting it all together, this example would output:

```html
<div class="this-is-the-parent">
  My customized header
  ...other parent layout stuff...
  Hello world
  ...other parent layout stuff...
  Default footer content goes here
</div>
```

because the header has been overridden by the child, and the footer has not.

krisselden:
@ef4 customize helper does not get executed unless you yield, this is ignoring the scoping and execution rules of htmlbars and glimmer.  You'd have to hack around a ton of stuff in order to get this to work.

krisselden:
There is another strategy I think we could take that still plays well with scoping rules that could solve both problems that I'm going to try to convince @mmun to help me out making the RFC since I'm not very good at writing this stuff up, but I also think is easier to do via html tags.

rwjblue:
Can you transfer this into an issue so that we can continue to iterate on the right way forward?  We have had a few discussions on this topic at the most recent face to face, and might have a path forward to achieve the intent.

taras:
What would you like me to copy into issue?

jamesarosen:
I just came across what I think is a good use-case for this. It doesn't seem to be *exactly* covered by the discussion above, but it's quite close.

```hbs
{{#filtered-list filter="Bob" items=users as |list|}}
  {{#list.empty}}
    No users matching {{list.filter}}
  {{/list.empty}}

  {{#list.item as |user|}}
    <li>{{user.name}}</li>
  {{/list.item}}
{{/filtered-list}}
```

ef4:
@jamesarosen that case can already be done pretty elegantly with the `{{yield to="inverse"}}` capability that is in 1.13.

```handlebars
{{#filtered-each filter=myFilter items=users as |user|}}
  {{user.name}}
{{else}}
  No users matching {{myFilter}}
{{/each-filtered-list}}
```

The template of the `filtered-each` component would look like:

```handlebars
{{#if matchingItems}}
  {{#each matchingItems as |item|}}
    {{yield item}}
  {{/each}}
{{else}}
  {{yield to="inverse"}}
{{/if}}
```

jamesarosen:
@ef4 :)

PR #48, opened 2015-04-14
Query params that reloadModel should not trigger loading route.

danfinlay:
- Start Date: 2015-04-13
- RFC PR: (leave this empty)
- Ember Issue: (leave this empty)
# Summary

Currently, query parameters set to reloadModel will trigger a relevant `LoadingRoute` when changed.  Given the most common uses of query params, which are state-changing but without general context switching, like filters, it makes more sense to simply set the model to the resolving promise, allowing more localized loading displays, and a less jarring filtering experience.
# Motivation

This allows for a more graceful default behavior when query-params are used to reload a route's model.  Without this change, any application that has query params used to refresh a model have their filters hidden during the filtered request, or cannot use the LoadingRoute feature of Ember.
# Detailed design

This should be a fairly simple change.  Right now the `Route` `queryParams` hash allows query-param keys to provide options.  In those options, if `reloadModel` is set to true, when the query param is changed, the full route transition is triggered, which gives any `LoadingRoute` in the hierarchy above the current route the opportunity to be displayed until the model is fulfilled.

Instead of a full route refresh, I think fewer of the route's initialization steps should be triggered.  I'm not really familiar with what steps are in this process, but I think for someone familiar with it, it should be easy enough to find.
# Drawbacks

If there are people who have Ember apps that have query-params that `reloadModel` and rely on a `LoadingRoute` to display loading indication during that request, and had no other way of representing the model in a loading state, their template would have a period where it had no model.
# Alternatives

The query params hash on the route could provide an additional option flag to enable this feature, avoiding any breakage.

App creators can simply not use the LoadingRoute feature, but that seems like a shame, since it's a useful tool.

App creators can manage their model on the controller, to avoid the `reloadModel` default behavior.
# Unresolved questions

What part of the redirection would have to be changed?:

Would this be a change in the the route? Somewhere else? I really don't know many Ember internals are implemented.:x

### DIFF
--- a/active/0000-query-params-should-not-trigger-loading-route.md
+++ b/active/0000-query-params-should-not-trigger-loading-route.md
@@ -1,0 +1,35 @@
+- Start Date: 2015-04-13
+- RFC PR: (leave this empty)
+- Ember Issue: (leave this empty)
+
+# Summary
+
+Currently, query parameters set to reloadModel will trigger a relevant `LoadingRoute` when changed.  Given the most common uses of query params, which are state-changing but without general context switching, like filters, it makes more sense to simply set the model to the resolving promise, allowing more localized loading displays, and a less jarring filtering experience.
+
+# Motivation
+
+This allows for a more graceful default behavior when query-params are used to reload a route's model.  Without this change, any application that has query params used to refresh a model have their filters hidden during the filtered request, or cannot use the LoadingRoute feature of Ember.
+
+# Detailed design
+
+This should be a fairly simple change.  Right now the `Route` `queryParams` hash allows query-param keys to provide options.  In those options, if `reloadModel` is set to true, when the query param is changed, the full route transition is triggered, which gives any `LoadingRoute` in the hierarchy above the current route the opportunity to be displayed until the model is fulfilled.
+
+Instead of a full route refresh, I think fewer of the route's initialization steps should be triggered.  I'm not really familiar with what steps are in this process, but I think for someone familiar with it, it should be easy enough to find.
+
+# Drawbacks
+
+If there are people who have Ember apps that have query-params that `reloadModel` and rely on a `LoadingRoute` to display loading indication during that request, and had no other way of representing the model in a loading state, their template would have a period where it had no model.
+
+# Alternatives
+
+The query params hash on the route could provide an additional option flag to enable this feature, avoiding any breakage.
+
+App creators can simply not use the LoadingRoute feature, but that seems like a shame, since it's a useful tool.
+
+App creators can manage their model on the controller, to avoid the `reloadModel` default behavior.
+
+# Unresolved questions
+
+What part of the redirection would have to be changed?:
+
+Would this be a change in the the route? Somewhere else? I really don't know many Ember internals are implemented.:x


kamalaknn:
+1, In my use case we have a `PostRoute` for post details nested inside a `PostsRoute` which displays the list. Changing the pagination or filters on the list will transition to `loading`. But I want the list to rerender without affecting the details (PostRoute). We override `queryParamsDidChange` as workaround now.

Panman8201:
-1 When I use the `reloadModel` option, the query-param is used in the REST call. Therefore I need an indication for the users that something is going on, while in the loading state. It is my understanding that any filtering/sorting/etc. going on in Ember-land should be done on the Controller, or perhaps the `setupController` hook.

flyswatter:
> I need an indication for the users that something is going on, while in the loading state.

You still would have one, your model could be an unresolved promise, allowing you to create a simple `{{isLoading}}` block, providing more local loading state, without obscuring the route's entire interface.  Furthermore, this entire feature could be hidden behind a flag in the route's `queryParams` options, something like `triggersFullRouteTransition` with a default of `true`, so it wouldn't need to break your app's behavior to be implemented.

> It is my understanding that any filtering/sorting/etc. going on in Ember-land should be done on the Controller, or perhaps the setupController hook.

This is fine for client-side sorting and filtering, but for server-side filtering, sorting, and searching, reloading the model on the route is really the correct Ember solution.  Making AJAX calls from the Controller and assigning them to the `model` property is hacky, and leads to memory leaks, as the store doesn't know to dispose of the old models.

Panman8201:
> You still would have one, your model could be an unresolved promise, allowing you to create a simple `{{isLoading}}` block, providing more local loading state, without obscuring the route's entire interface.

I suppose if you were using `RSVP.hash` to return multiple models then that would make sense (only reloading one of the many models). In most of my cases the route relies on one model so if it's reloading then it's fine that the entire interface is obscured/loading.

> This is fine for client-side sorting and filtering, but for server-side filtering, sorting, and searching, reloading the model on the route is really the correct Ember solution.

I guess that's my point, if something is being loaded from the server then you'll want to indicate so to your users. Sounds like you are using block level loading states, instead of the loading route. 

I'd say your alternate of an addition option would be the best solution for this particular RFC.

flyswatter:
> In most of my cases the route relies on one model so if it's reloading then it's fine that the entire interface is obscured/loading.

I guess the gambit of this feature (if it weren't hidden behind an option flag) would be that while your route relies on a model, your route's query parameters probably don't, and so there's value gained by not hiding them with a loading route.

If we believe that for most users, the options provided by query params represent a set of options of the route's state independent of the model, effecting it but not effected by it, then reloading the entire route is discarding potentially valuable information from the current view (the query params), which could be a set of filters, a search bar, or just visual consistency.

But yeah, I concede this is a gambit.  For some users, the Loading substate may already be used effectively, and so a breaking change for some users may not be worth the benefit, although I do believe this would be a better default, if there wasn't already a legacy in place.

Panman8201:
> while your route relies on a model, your route's query parameters probably don't

Yeah, I see where you're going with it. From a UX perspective, it would be helpful show a loading state just for the affected data, rather than the entire route. So if I have a list of filters and click on one of them the list will appear in a loading state but the filters will remain visible. In that case, when should the `{{link-to}}` query-param become active?

For the default, IMO it should remain going to the loading route as it's easier to implement (globally even). And if someone wants to get into fancier UI design then they can use the option/flag and make their own loading blocks.

Good stuff, I like the focus on UX as of late.

flyswatter:
> In that case, when should the {{link-to}} query-param become active?

I'm unsure whether the `{{link-to}}` would get the `active` class, or just its normal [loadingClass](http://emberjs.com/api/classes/Ember.LinkView.html#property_loadingClass) property.

The `active` class would achieve my own personal needs, but it seems like using the `loadingClass` property might be more in line with the established Ember workflow.

Now if I can just get `loadingClass` to work right...

> Good stuff, I like the focus on UX as of late.

Thanks!  I guess I'm just coincidentally part of that, since this is my first RFC!

sarus:
Definitely could use this.  We use query params for filtering, soprting, and paging and can't use loading templates because it is way too jarring to have the entire page show a loading template when the user pages data.  It's made even worse in our case because we use nested routes and so the user is only paging one set of data on the page but the entire page "flashes" to show the loading template when the user performs a paging, sorting or filtering operation.  

Is there some other way around this?

wagenet:
Given the six years of inactivity here I think we can close it. If there is something here still worth considering, please let me know.

PR #50, opened 2015-05-07
Add improved actions RFC

mixonic:
[View as HTML](https://github.com/emberjs/rfcs/blob/action/active/0000-improved-actions.md)

### DIFF
--- a/active/0000-improved-actions.md
+++ b/active/0000-improved-actions.md
@@ -1,0 +1,478 @@
+- Start Date: 2014-05-06
+- RFC PR: (leave this empty)
+- Ember Issue: (leave this empty)
+
+# Summary
+
+The `{{action` helper should be improved to allow for the creation of
+closed over functions that can be passed between components and passed
+the action handlers.
+
+See [this example JSBin from @rwjblue](http://emberjs.jsbin.com/rwjblue/466/edit?html,js,output)
+for a demonstration of some of these ideas.
+
+# Motivation
+
+Block params allow data to be passed from one component to a downstream
+component, however there is currently no way to pass a callback to a downstream
+component.
+
+# Detailed design
+
+First, the existing uses of `{{action` will be maintained. An action can be attached to an
+element by using the helper in element space:
+
+```hbs
+{{! app/index/template.hbs }}
+{{! submit action will hit immediate parent }}
+<button {{action "submit"}}>Save</button>
+```
+
+An action can be passed to a component as a string:
+
+```hbs
+{{! app/index/template.hbs }}
+{{my-button on-click="submit"}}
+```
+
+```js
+// app/components/my-button/component.js
+export default Ember.Component.extend({
+  click: function(){
+    this.sendAction('on-click');
+  }
+});
+```
+
+Or a default action can be passed:
+
+```hbs
+{{! app/index/template.hbs }}
+{{my-button action="submit"}}
+```
+
+```js
+// app/components/my-button/component.js
+export default Ember.Component.extend({
+  click: function(){
+    this.sendAction();
+  }
+});
+```
+
+In all these cases, `submit` is called on the parent context relative to the scope `action` is
+attached in. The value `"submit"` is attached to the component in the last two as
+`this.attrs.on-click` or `this.attrs.action`, although it is not directly used.
+
+### Creating closure actions
+
+Closure actions are created in a template and may be used in all places a string
+action name can be used. For example, this current functionality:
+
+```hbs
+<button {{action "submit" on="click"}}>Save</button>
+```
+
+Would be written using a closure action as:
+
+```hbs
+<button {{action (action "submit") on="click"}}>Save</button>
+```
+
+The functionality is exactly the same as the string-based action example.
+How does that happen?
+
+* `(action "submit")` reads the `submit` function off the current scope's
+  `actions.submit` property.
+* It then creates a closure to call that function.
+* `{{action` receives that function as a param. It registers a listener (in
+  this case on click) and when fired calls the closure function.
+
+Consider usage on the calling side. With the current string-based actions:
+
+```hbs
+{{my-component action="submit"}}
+```
+
+```js
+export default Ember.Component.extend({
+  click: function(){
+    this.sendAction(); // submit action, legacy
+    // this.attrs.action is a string
+    this.attrs.action; // => "submit"
+  }
+});
+```
+
+With closure actions, the action is available to call directly. The `(action` helper
+wraps the action in the current context and returns a function:
+
+```hbs
+{{my-component action=(action "submit")}}
+```
+
+```js
+export default Ember.Component.extend({
+  click: function(){
+    this.sendAction(); // submit action, legacy
+    // this.attrs.action is a function
+    this.attrs.action(); // submit action, new style
+  }
+});
+```
+
+A more complete example follows, with a controller for context:
+
+```js
+// app/index/controller.js
+export default Ember.Controller.extend({
+  actions: {
+    submit: function(){
+      // some submission task
+    }
+  }
+});
+```
+
+```hbs
+{{! app/index/template.hbs }}
+{{my-button save=(action 'submit')}}
+```
+
+```js
+// app/components/my-button/component.js
+export default Ember.Component.extend({
+  click: function(){
+    this.attrs.save();
+    // for backwards compat, you may also this.sendAction('save');
+  }
+});
+```
+
+### Hole punching with a closure-based action
+
+The current system of action bubbling falls down quickly when you want to pass a message through multiple
+levels of components. A closure based action system helps address this.
+
+Instead of relying on bubbling, a closure action wraps an action from the current context's
+`actions` hash in a function that will call it on that context. For example:
+
+```hbs
+{{! app/index/template.hbs }}
+{{my-form submit=(action 'submit')}}
+```
+
+```hbs
+{{! app/components/my-form/template.hbs }}
+{{my-button on-click=attrs.submit}}
+```
+
+```hbs
+{{! app/components/my-button/template.hbs }}
+<button></button>
+```
+
+```js
+// app/components/my-button/component.js
+export default Ember.Component.extend({
+  click: function(){
+    this.attrs['on-click']();
+    // for backwards compat, you may also this.sendAction();
+  }
+});
+```
+
+A closure action can also be called by an action handler:
+
+```hbs
+{{! app/index/template.hbs }}
+{{my-form submit=(action 'submit')}}
+```
+
+```hbs
+{{! app/components/my-form/template.hbs }}
+{{my-button on-click=submit}}
+```
+
+```hbs
+{{! app/components/my-button/template.hbs }}
+<button {{action on-click}}></button>
+```
+
+Lastly, closure actions allow for yielding an action to a block. For example:
+
+```hbs
+{{! app/index/template.hbs }}
+{{my-form save=(action 'submit') as |submit reset|}}
+  <button {{action submit}}>Save</button>
+  {{! ^ goes to my-form's save attr property, which
+        is the submit action on the outer scope }}
+  <button {{action reset}}>Reset</button>
+  {{! ^ goes to my-form }}
+  <button {{action "cancel"}}>Cancel</button>
+  {{! ^ goes to outer scope }}
+{{/my-form}}
+```
+
+```hbs
+{{! app/components/my-form/template.hbs }}
+{{yield attrs.save (action 'reset')}}
+```
+
+```js
+// app/components/my-form/component.js
+export default Ember.Component.extend({
+  actions: {
+    reset: function(){
+      // rollback
+    }
+  }
+});
+```
+
+### Currying arguments with a closure-based action
+
+With string-based actions, an argument can be passed to the called function. For
+example:
+
+```hbs
+<button {{action "save" model}}></button>
+```
+
+```js
+export default Ember.Component.extend({
+  actions: {
+    save: function(model) {
+      model.save();
+    }
+  }
+});
+```
+
+Closure actions allow for another opportunity to curry arguments. Arguments
+set by an element action helper simply add to the end of the arguments list:
+
+```hbs
+{{! app/index/template.hbs }}
+{{my-component save=(action "save" model)}}
+```
+
+```hbs
+{{! app/components/my-component/template.hbs }}
+<button {{action attrs.save prefs}}></button>
+```
+
+```js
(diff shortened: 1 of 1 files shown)


active/0000-improved-actions.md
```diff
@@ -0,0 +1,278 @@
...
+action name can be used. For example, this current functionality:
+
+```hbs
+<button {{action "submit" on="click"}}>Save</button>
+```
+
+would be written using a closure action as:
+
+```hbs
+<button {{action (action "submit") on="click"}}>Save</button>
```
rwjblue: This is pretty confusing, why double wrapping?  What does the closure wrapped action get attached to?

active/0000-improved-actions.md
```diff
@@ -0,0 +1,278 @@
...
+With the current string-based actions:
+
+```hbs
+{{my-component action="submit"}}
+```
+
+```js
+export default Ember.Component.extend({
+  click: function(){
+    this.attrs.action === "submit";
```
rwjblue: This seems like a comment informing the reader of what happens when `this.sendAction()` is called below, can it be made into a JS comment so the highlighting makes it clearer?

active/0000-improved-actions.md
```diff
@@ -0,0 +1,278 @@
...
+wraps the action in the current context and returns a function:
+
+```
+{{my-component action=(action "submit")}}
+```
+
+```js
+export default Ember.Component.extend({
+  click: function(){
+    typeof this.attrs.action === "function";
```
rwjblue: Can make this a comment or add a comment explaining why it is here?

active/0000-improved-actions.md
```diff
@@ -0,0 +1,278 @@
...
+  actions: {
+    submit: function(){
+      // some submission task
+    }
+  }
+});
+```
+
+```hbs
+{{! app/components/my-button/template.hbs }}
```
rwjblue: Seems like this should actually be in `{{! app/templates/index.js }}` since it is invoking `my-button` (and passing the closure wrapped `submit` action in to the `{{my-button}}` component).

active/0000-improved-actions.md
mixonic: Hm. It is not double wrapped- the return value of `(action "submit")` is a function, which the element-space `{{action` helper calls.

String-based actions bubble like before. Function actions are closures to the scope they were created in.

active/0000-improved-actions.md
```diff
@@ -0,0 +1,278 @@
...
+  click: function(){
+    this.attrs.save();
+    // for enhanced backwards compat, you may also this.sendAction('save');
+  }
+});
+```
+
+### Hole punching with a closure-based action
+
+The current system of action bubbleing falls down quickly when you want to pass a message through multiple
```
rwjblue: I am not certain, but I think this should be `bubbling`.

rwjblue: Also, you might mention it as "manual action bubbling". @danmcclain wrote a nice article about the current technique at http://reefpoints.dockyard.com/2015/01/28/bubbling-actions-through-components.html...

active/0000-improved-actions.md
```diff
@@ -0,0 +1,278 @@
...
+  <button {{action reset}}>Reset</button>
+  {{! ^ goes to my-form }}
+  <button {{action "cancel"}}>Cancel</button>
+  {{! ^ goes to outer scope }}
+{{/my-form}}
+```
+
+```hbs
+{{! app/components/my-form/template.hbs }}
+{{yield save (action 'reset')}}
```
rwjblue: Should this be `attrs.save`?

active/0000-improved-actions.md
```diff
@@ -0,0 +1,278 @@
...
+Currently `{{action` is only used in an element space:
+
+```hbs
+<button {{action "booyah"}}>Fire</button>
+```
+
+The closure usage is a new, perhaps `action` is not the right word. However the two
+behaviors are pretty similar in their conceptual behavior.
+
+* `{{action` in an element space attaches an action from the current scope to an event on the element,
```
rwjblue: "attaches an action name" (not the action itself, which I think of as the action function)

active/0000-improved-actions.md
```diff
@@ -0,0 +1,278 @@
...
+* `(action` closes over an action from the current scope so it can be attached as an action later.
+
+Additionally, there may be developers who still have `{{action someActionName}}` instead
+of the quoted version. This is long deprecated, but these apps may see some
```
rwjblue: AFAIK, it is actually unsupported since 1.10 or 1.11.  We do support bound action names (where the name resolves to a string when clicked) though, and that is not deprecated (and I think it should stay).

active/0000-improved-actions.md
rwjblue: * `<button {{action 'submit'}}>Save</button>` would invoke the local `submit` action as the button elements on click handler.
* `<button {{action (action "submit")}}>Save</button>` would wrap the local submit action in a function, then attach that function to the button elements on-click handler.

So, yes, technically it would work, but why would you ever do this?  If you are calling an action from your own local scope (which would have to be true for `(action "submit")` to work), why would you then prefix with `{{action` again?

rwjblue:
I am hugely :+1: on this RFC, when can I have it?

workmanw:
What about `target` and action arguments? Will those be supported? Example: 

```hbs
{{user-inline-edit submit=(action "saveUser" userModel target="bulkSaveManger")}}
```

active/0000-improved-actions.md
mixonic: We discussed, will update in RFC to clarify the steps happening. Robert is correct that these cases would have identical behavior, that was what I intended to show.

mixonic:
@workmanw interesting. It is technically possible. `target` would be specifying an alternative target for the closure. I will add.

This does raise the issue of `<button {{action someClosureAction target=otherComponent}}>` which should just throw IMO. You cannot have a closure action and a target for that action defined at attach-time.

workmanw:
@mixonic I'm not sure how others feel, but for our app this would be greatly beneficial. One of our existing pain points with the current action system is the need for a single component to have different targets for different actions. Wrapping up the `target` in the `action` sub-expression would save us a lot of "action daisy chaining".

Thanks for this RFC! Super excited for this.

mixonic:
@workmanw Added here: https://github.com/emberjs/rfcs/blob/action/active/0000-improved-actions.md#re-targeting-the-scope-of-a-closure-action

workmanw:
:+1: Thank you.

active/0000-improved-actions.md
```diff
@@ -0,0 +1,396 @@
...
+`actions` hash in a function that will call it on that context. For example:
+
+```hbs
+{{! app/index/template.hbs }}
+{{my-form submit=(action 'submit')}}
+```
+
+```hbs
+{{! app/components/my-form/template.hbs }}
+{{my-button on-click=attrs.submit}}
```
drogus: Would `{{my-button on-click=(action 'submit')}}` work here as well?

mixonic: `{{my-button on-click=(action 'submit')}}` would target the submit action on MyForm. The current scope is always what is looked to for an `(action` string.

rwjblue: Only if you wanted to provide `my-button` with the action from `my-form`.  As it is written here, `my-button` is getting the wrapped action directly from `index` controller/route, and you do not need to handle the action and `this.sendAction` again (the current manually bubbling).

drogus: Ah, right, I have misunderstood the use case here, thanks for explanation!

active/0000-improved-actions.md
```diff
@@ -0,0 +1,439 @@
...
+- RFC PR: (leave this empty)
+- Ember Issue: (leave this empty)
+
+# Summary
+
+The `{{action` helper should be improved to allow for the creation of
+closed over functions that can be passed between components and passed
+the action handlers.
+
+See [this example JSBin from @rwjblue](http://emberjs.jsbin.com/rwjblue/223/edit?html,js,output)
```
rwjblue: I updated the JSBin to work (using 1.11.3): http://emberjs.jsbin.com/rwjblue/466/edit?html,js,output

The original doesn't work (doesn't use template compiler).

mixonic:
Updated with a description of `value=`.

Seems good to merge.

rwjblue:
:👍:

al3x-edge:
:+1: What would yielding multiple actions look like?

Current example:
```hbs
{{! app/components/my-form/template.hbs }}
{{yield attrs.save (action 'reset')}}
```

Multiple yield?:
```hbs
{{! app/components/my-form/template.hbs }}
{{yield attrs.save (action 'reset') (action 'update')}}
```

mixonic:
@al3x-edge like:

```hbs
{{yield attrs.save (action 'reset') (action 'update')}}
```

Your third example would curry the argument `update` onto the action. When `this.attrs.reset()` was called (with no argument), the handler would receive `"update"` as the first argument.

al3x-edge:
@mixonic Ah. Makes sense. Didn't read well enough apparently -- Thanks Matt

mgenev:
Does this address all the problems expressed in this RFC? https://github.com/emberjs/rfcs/pull/42

If yes, yay, if not, meh

mixonic:
@mgenev it mitigates the challenge you pointed out in the [Summary](https://github.com/mgenev/rfcs/blob/3fa8ffd6e41bd99d88653497abc0ebfe25a358f0/active/0000-global-action-bubbling-for-components.md#summary), although how it fixes that problem is very different than the solutions suggested in that discussion.  

What used to be:

```hbs
{{! app/index/template.hbs }}
{{my-component openModal="openModal"}}
```

```hbs
{{! app/components/my-component/template.hbs }}
<button {{action 'popSomething'}}>
```

```js
/// app/components/my-component/component.js
import Ember from "ember";

  actions: {
    popSomething() {
      this.sendAction('openModal');
    }
  }
});
```

Is now simply:

```hbs
{{! app/index/template.hbs }}
{{my-component openModal=(action "openModal")}}
```

```hbs
{{! app/components/my-component/template.hbs }}
<button {{action openModal}}>
```

However I'm not sure this addresses your desire to have "openModal" be an action on the route reachable from any arbitrary point in the component hierarchy. Closure actions read the function off `this.actions` and call it- they do not use `send` at all and thus will *never* bubble.

For the specific case you laid out, I would strongly suggest using a service:

```hbs
{{! app/components/my-component/template.hbs }}
<button {{action (action 'open' target=modalService)}}>
```

```js
/// app/components/my-component/component.js
import Ember from "ember";

  modalService: Ember.inject.service('modal')
});
```

```js
/// app/services/modal.js
import Ember from "ember";

export default Ember.Service.extend(Ember.Evented, {
  actions: {
    open() {
      this.trigger('didOpen');
    }
  }
});
```
```js
/// app/components/my-modal/component.js
import Ember from "ember";

  modalService: Ember.inject.service('modal'),
  init() {
    this._super.apply(this, arguments);
    this._onOpen = () => {
      // do something to display popup
    };
    this.get('modalService').on('didOpen', this_onOpen);
  },
  destroy() {
    this.get('modalService').off('didOpen', this._onOpen);
    this._super.apply(this, arguments);
  }
});
```

In this example, the `my-modal` component just reflects the UI state of the `modalService`. The action `open` can be called from any component that has the service available.

It is off topic for this RFC, but I believe there are at least two follow-up RFCs that would help make this pattern easier:

* Managing the event subscription to `didOpen` manually is error prone. @igorT and I were just talking about making `Ember.on('modalService.didOpen', () => {})` work, which would remove the need for manual setup and teardown.
* `<button {{action (action 'open' target=modalService)}}>` screams for a better way to attach a closure action to a click event. We've toyed with `<button on-click={{action 'open' target-modalService}}>` but have not yet reconciled that with components.
* Oh yeah, maybe all services should be evented. I'm starting to use that mixin a ton.
* There is un-RFC'd discussion of making `outlet.foo` a way to access the routable component of a given component. This provides the "skip to the top" functionality you want without a concept of global bubbling. You would just make the target of the action `outlet`.

Still improvements to go, but I'm feeling *much* better about the patterns we can create with these new actions and with the direction our conversations are going.

mgenev:
@mixonic thank you for the great detailed explanation and examples. Being able to omit the redundant sending of the intermediate action is indeed a big improvement.
Regarding services, in the case of a modal, I can see how it makes a lot of sense to have a modal service.

What about other instances though in which I'm simply nesting components 4-5-6 levels deep in a route, which I do find myself doing now in the "everything's a component" architecture. Am I to create a service for every route then? This strikes me as a hassle. I'd rather have some kind of global proxy service or a pub-sub pattern which allows me to take actions to the top in every case...

mixonic:
@mgenev I find that most cases where I need access to the "top" level scope (by that we usually mean the controller today, in the future the routable component) can be handled by using `yield`:

```hbs
{{! app/index/template.hbs }}
{{my-component}}
```

```hbs
{{! app/components/my-component/template.hbs }}
{{my-other-component passedVal=fromComponent}}
```

```hbs
{{! app/components/my-component/template.hbs }}
<button {{action omgHowDoIGetToController}}></button>
```

Composing these component with yield keeps access to the parent scope easier:

```hbs
{{! app/index/template.hbs }}
{{#my-component as |val|}}
  {{#my-other-component passedVal=val}}
    <button {{action omgHowDoIGetToController}}></button>
  {{/my-other-component}}
{{/my-component}}
```

```hbs
{{! app/components/my-component/template.hbs }}
{{yield fromComponent}}
```

```hbs
{{! app/components/my-component/template.hbs }}
{{yield}}
```

But I definitely acknowledge this isn't always viable. I think the `outlet.` API is the best bet for accessing "top level" things when components are deeply nested. It is always going be a bit awkward, because the ask is a tough one. A deeply nested component being aware of the ambient state for the whole application violates the goal of having components be scoped, isolated things that are portable.

mgenev:
@mixonic thanks, I'll explore the pattern with the block component and yielding, @runinspired was also telling me that as a solution in the slack.

For reference, here's how Aurelia solves these issues very simply:
http://aurelia.io/docs.html#event-modes
http://aurelia.io/docs.html#eventing

Raghs:
@mixonic 

I have tried to implement modal dialogue using service pattern as you mentioned in you post. but i am not able to achieve , if you don't mind could you please share the working example in jsbin

Thanks

Raghs:
@mixonic

I have tried to implement modal dialogue using service pattern as you mentioned in your post, but i am not able to achieve the same ,it would be great help if you please share the working code in JSBIN

Thanks

PR #52, opened 2015-05-17
Expression RFC

mixonic:
[View as HTML](https://github.com/emberjs/rfcs/blob/expression/active/0000-expression.md)


rlivsey:
Huge :+1. 

A hook for cleaning up after ourselves would be useful but if `destroy` gets called at that point then that's fine.

jamesarosen:
Will there be a way to register expressions that don't have a `-` in their name? For example, [ember-can](https://github.com/minutebase/ember-can)'s `can`, [ember-cli-i18n](https://github.com/dockyard/ember-cli-i18n/)'s `t`, and [ember-i18n v4.0](https://github.com/jamesarosen/ember-i18n/pull/213)'s `t`.

tikotzky:
Whats the reasoning behind requiring a `-` in the expression name?
As I understand its required for components to align with web components. Is it just to help differentiate between helpers and expressions?

mixonic:
After discussion and consideration, I've decided to advocate for another direction. Instead of a wholly new type of template api, I'd like to introduce class-based helpers with a `compute` function having the same call signature as function helpers.

PR forthcoming, likely tomorrow morning at this point.

mixonic:
Regarding the dash- the new proposal will put the burden to change that requirement back on helpers themselves. With the glimmer refactor I think it is possible, will bring it up in discussion.

mixonic:
Supplanted by https://github.com/emberjs/rfcs/pull/53

PR #53, opened 2015-05-18
Helpers

mixonic:
[View as HTML](https://github.com/emberjs/rfcs/blob/improved-helpers/active/0000-helpers.md)

Pardon any spelling snafus etc, written on a :bus: :-p

### DIFF
--- a/text/0053-helpers.md
+++ b/text/0053-helpers.md
@@ -1,0 +1,327 @@
+- Start Date: 2015-05-17
+- RFC PR: https://github.com/emberjs/rfcs/pull/53
+- Ember Issue: https://github.com/emberjs/ember.js/pull/11278
+
+# Summary
+
+Ember.js 1.13 will introduce a new API for helpers. Helpers will come in two
+flavors:
+
+**Helpers** are a class-based way to define HTMLBars subexpressions. Helpers:
+
+  * Have a single return value.
+  * Must have a dash in their name.
+  * Cannot be used as a block (`{{#some-helper}}{{/some-helper}}`).
+  * Can store and read state.
+  * Have lifecycle hooks analogous to components where appropriate. For
+    example, a helper may call `recompute` at any time to generate a new
+    value (this is akin to `rerender`).
+  * Are a superset of shorthand helpers, the function-based syntax described
+    below. They can do more, but in many cases a shorthand helper is appropriate.
+
+**Shorthand helpers** are a function-based way to define HTMLBars
+subexpressions. Helpers written this way:
+
+  * Have all the limitations of regular helpers.
+  * Have no instance associated with them, cannot store or read state.
+  * Have no lifecycle hooks. The function is simply re-computed when any input
+    changes.
+
+These improved helpers fill a gap in Ember's current template APIs:
+
+|                   | has positional params | has layout (shadow DOM) | can yield template | has lifecycle, instance | can control rerender |
+|-------------------|-----------------------|-------------------------|--------------------|-------------------------|----------------------|
+| components        | Yes                   | Yes                     | Yes                | Yes                     | Yes                  |
+| helpers           | Yes                   | No                      | No                 | Yes                     | Yes                  |
+| shorthand helpers | Yes                   | No                      | No                 | No                      | No                   |
+
+An example helper:
+
+```js
+// app/helpers/full-name.js
+import Ember from "ember";
+
+export default Ember.Helper.extend({
+  nameBuilder: Ember.inject.service(),
+  compute(params) {
+    const builder = this.get('nameBuilder');
+    return builder.fullName(params[0], params[1]);
+  }
+});
+```
+
+An example shorthand helper:
+
+```js
+// app/helpers/full-name.js
+import Ember from "ember";
+
+export default Ember.Helper.helper(function(params, hash) {
+  let fullName = params.join(' ');
+  if (hash.honorific) {
+    fullName = `${hash.honorific} ${fullName}`
+  }
+  return fullName;
+});
+```
+
+Helpers can be used anywhere an HTMLBars subexpression is valid:
+
+```hbs
+{{full-name 'Bigtime' 'Beagle'}}
+{{input value=(full-name 'Gyro' 'Gearloose') readonly=true}}
+{{#if (eq (full-name 'Webbigail' 'Vanderquack') selectedFullName))}}
+  You have chosen wisely.
+{{/if}}
+```
+
+# Motivation
+
+Ember.js 1.13 make a private API change that removed the ability to access
+application containers. `Ember.HTMLBars._registerHelper` was previously passed
+the `env` object, and this was removed as it is an internal implementation
+detail.
+
+Ember's helper API has not kept pace with improvements possible
+after the introduction of HTMLBars. This has resulted in the community using
+a variety of private APIs, many of which leak information about the outer
+context of a helpers invocation as well as the render layer implementation.
+
+The current public API is:
+
+* [Ember.Handlebars.makeBoundHelper](http://emberjs.com/api/classes/Ember.Handlebars.html#method_makeBoundHelper)
+
+This API is sorely lacking in functionality required by addon authors.
+
+* Has no access to other parts of the app, like services
+* Leaks a private API for dealing with blocks
+* Results in less efficient helpers due to the Handlebars compatibility layer
+* Has poor support for hash arguments
+
+Additionally it remains difficult to write a helper that recomputes due to
+something besides the change of its input.
+
+Specifically, this RFC addresses many of the concerns in
+[emberjs/ember.js#11080](https://github.com/emberjs/ember.js/issues/11080).
+Libraries such as [yahoo/ember-intl](https://github.com/yahoo/ember-intl),
+[dockyard/ember-cli-i18n](https://github.com/dockyard/ember-cli-i18n), and
+[minutebase/ember-can](https://github.com/minutebase/ember-can) will be
+provided a viable public API to couple to.
+
+# Detailed design
+
+Helpers must have a dash in their name. In an Ember-CLI app, they can be named
+according to the `app/helpers/full-name.js` convention (`app/full-name/helper.js`
+in pods mode). For a globals app, naming a helper `App.FullNameHelper` is
+sufficient.
+
+### Definition and lifecycle
+
+A helper is defined as a class inheriting from `Ember.Helper`. For
+example:
+
+```js
+// app/helpers/hello-world.js
+import Ember from "ember";
+
+// Usage: {{hello-world}}
+export default Ember.Helper.extend({
+  compute() {
+    return "Hello Helper World";
+  }
+});
+```
+
+Upon initial render:
+
+* The helper instance is created.
+* The `compute` method is called. The return value is outputted where the
+  helper is used. For example in `<div class={{some-helper}}></div>` the return
+  value is set to the class.
+
+The `compute` function is always called with `params` (the bare, ordered
+arguments) and `hash` (the named arguments). For example:
+
+```js
+// app/helpers/greet-someone.js
+import Ember from "ember";
+
+// Usage: {{greet-someone 'bob' greeting='say hello'}}
+export default Ember.Helper.extend({
+  compute(params, hash) {
+    return `Hello ${params[0]}, nice to ${hash.greeting}`;
+  }
+});
+```
+
+Which functions the same as this shorthand:
+
+```js
+// app/helpers/greet-someone.js
+import Ember from "ember";
+
+// Usage: {{greet-someone 'bob' greeting='say hello'}}
+export default Ember.Helper.helper(function(params, hash) {
+  return `Hello ${params[0]}, nice to ${hash.greeting}`;
+});
+```
+
+When the `params` or `hash` contents change, the `compute` method is called
+again. The instance of the helper is preserved across rerenders of the parent.
+A shorthand helper, having no instance, is called every time a bound
+argument changes.
+
+The `init` and `destroy` methods can be subclassed for setup and teardown.
+
+### Consuming a helper
+
+Helpers can be used anywhere an HTMLBars subexpression can be
+used. For example:
+
+```hbs
+{{#if (can-access 'admin')}}
+  {{link-to 'login'}}
+{{/if}}
+{{#if (eq (can-access 'admin') false)}}
+  No login for you
+{{/if}}
+<my-login-button isAdmin={{can-access 'admin'}} />
+Can access? {{can-access 'admin'}}
+```
+
+Passing a helper to a `{{`- invoked component skips the auto-`mut` behavior:
+
+```hbs
+{{my-login-button isAdmin=(can-access 'admin')}}
+```
+
+Let's step through exactly what happens when using an helper like this:
+
+```hbs
+<my-login-button isAdmin={{can-access 'admin'}} />
+```
+
+Upon initial render:
+
+* The helper `can-access` is looked up on the container
+* The helper is identified as a full helper, not a shorthand helper function
+* The helper is initialized (`init` is called)
+* The `compute` function is called on the helper.
+* The return value from `compute` is passed as an `attr` to `my-login-button`.
+* The helper instance remains in memory.
+
+If the parent scope is rerendered:
+
+* The `compute` function is called again.
+* The return value from `compute` is passed as an `attr` to `my-login-button`.
+
+Upon teardown:
+
+* The helper is destroyed, calling the `destroy` method.
+
+### Returning a value
+
+The return value of helper is passed through to where their subexpression
+is called. For example, given a helper (this one a shorthand helper):
+
+```js
+// app/helpers/full-name.js
+import Ember from "ember";
+
+export default Ember.Helper.helper(function fullName(params, hash) {
+  return params.join(' ');
+}
+```
+
+The following are effectively the same:
+
+```hbs
+<div data-name={{full-name "Fenton" "Crackshell"}}></div>
+<div data-name={{"Fenton Crackshell"}}></div>
+```
+
+```hbs
+{{my-component name=(full-name "Magica" "De Spell")}}
+{{my-component name="Magica De Spell"}}
+```
+
+```hbs
+<p>{{full-name "Bentina" "Beakley"}}</p>
+<p>{{"Bentina Beakley"}}</p>
+```
+
+An exclusion to this pattern is the following form:
+
+```hbs
+<div {{full-name "Webbigail" "Vanderquack"}}></div>
+```
+
+This is a legacy form of mustache usage. Helpers will throw an exception when
+used in this manner.
+
+### Consuming services and recompute
+
+Helpers are a valid target for service injection. For example:
+
+```js
+// app/helpers/current-user-name.js
+import Ember from "ember";
+
+export default Ember.Helper.extend({
+  // Same API as components:
+  session: Ember.inject.service(),
+  compute() {
+    return this.get('session.currentUser.name');
+  }
+});
+```
+
+However consuming a property from a service does not bind the data being
+displayed to that property. After `{{current-user-name}}` has been computed
+and rendered, it will never be invalidated.
+
+For this reason, helpers are granted some control over their
+computation lifecycle. A helper will recompute when:
+
+* A value passed via the template changes (`params` or `hash`)
+* The `recompute` method is called
+
+For example, this helper checks if the current use has access to a
+resource type:
+
+```js
+// app/helpers/can-access.js
+import Ember from "ember";
+
+// Usage {{if (can-access 'admin') 'Welcome, boss' 'Heck no!'}}
+export default Ember.Helper.extend({
+  session: Ember.inject.service(),
+  onCurrentUserChange: Ember.observes('session.currentUser', function() {
+    this.recompute();
+  }),
+  compute(params) {
+    const currentUser = this.get('session.currentUser');
+    return currentUser.can(params[0]);
+  }
+});
+```
+
+# Drawbacks
+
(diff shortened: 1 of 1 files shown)


active/0000-stateful-helpers.md
```diff
@@ -0,0 +1,276 @@
...
+    value.
+  * Are a superset of traditional function-based helpers. They can do more,
+    but in many cases a traditional helper is appropriate.
+
+They fill a gap in Ember's current template APIs:
+
+ |has positional params|has template|has lifecycle, instance|can control rerender
+---|---|---|---|---
+components|Yes|Yes|Yes|Yes
+function helpers|Yes|No|No|No
```
rwjblue: Function helpers can yield to templates.  `if` for example is a function helper today...

jamesarosen:
@mixonic you've made my day 😀

My two favorite points are making `registerHelper` public and the `recompute` method.

I'm definitely for `init` and `willDestroy` hooks, but its *far* from a deal-breaker to have to override `destroy`.

active/0000-stateful-helpers.md
```diff
@@ -0,0 +1,276 @@
...
+    return "Hello Helper World";
+  }
+});
+```
+
+Upon initial render:
+
+* The helper instance is created.
+* The `compute` method is called, and its return value is treated like the
+  return value of a function-based helper.
```
rwjblue: It is not obvious to me what exactly is done with the return value of a function-based helper. Can you potentially list a few in context examples for:

* sub-expr: `{{#each (get model recordType) as |record|}} {{/each}}`
* element space: `<div {{get model recordType}}></div>` (I don't like this and it should be discouraged for sure)
* attribute assignment: `<div data-name="{{get model recordType}}"></div>`
* "normal" (aka none of the above): `<p>{{get model recordType}}</p>`

mmun:
This is going to be very popular with addon authors :) I don't like `recomputeObserves`. Maybe just `dependencies`?

active/0000-stateful-helpers.md
```diff
@@ -0,0 +1,276 @@
...
+    options.template.yield();
+  } else {
+    options.inverse.yield();
+  }
+};
+```
+
+Porting this usage to a stateful helper only changes the boilerplate:
+
+```
```
rwjblue: js block

active/0000-stateful-helpers.md
```diff
@@ -0,0 +1,276 @@
...
+// app/helpers/custom-if.js
+import Ember from "ember";
+
+// Usage: {{if-is-yes 'no'}}not rendered{{else}}rendered!{{/if-is-yes}}
+export default Ember.Helper.extend({
+  compute(params, hash, options) {
```
rwjblue: Can we call this `blocks`?

active/0000-stateful-helpers.md
```diff
@@ -0,0 +1,276 @@
...
+    if (params[0]) {
+      options.template.yield();
```
rwjblue: Can this be `blocks.default.yield()` (which is what `template` is IIRC)?

rwjblue:
Looks wonderful.  A couple points:

* Can you add verbiage to indicate the lifespan of a stateful helper?  Are they singleton? Are they created/destroyed along the same lines as components?  
* Does the stateful helper have knowledge of its surroundings?  For example: is it aware of its containing component/view?
* What internal helpers can be restructured to extend from `Ember.Helper` (at least the new `get` helper that is being worked on)?

mmun:
The `get` helper is not amenable to this kind of refactoring because it needs to munge the params to create a custom stream. As I understand it, the only additional dependencies you have should come from  global singletons, so basically services.

cibernox:
I do like this proposal better than the subexpresions one. 

This kind of helpers remains as close as possible to the existing helpers, that are the simplest abstraction you can have for getting an output from some input, just adding the minimum amount of functionality to let them `watch` other subsystems.

I'd like the to be one way and not having access to the context. That would limit side efects changes to things you're injecting, and ideally you should keep that to a minimum.

active/0000-stateful-helpers.md
mixonic: helpers can yield but do not have a template. Do you think it would be more accurate to say they do not have a layout?

rwjblue: I thought you were saying they cannot "accept a block" (aka the template) which is not correct.  The terminology of `layout` and `template` are annoying, and I am sorry that you have to thread the needle here...

active/0000-stateful-helpers.md
mixonic: I've added a section on return values, and ran some tests. I discovered that function-based helpers don't work in some of the places I describe they should. I'd like to make them work in those cases.

active/0000-stateful-helpers.md
```diff
@@ -0,0 +1,322 @@
...
+  * Can store and read state
+  * Have lifecycle hooks analagous to components where appropriate. For
+    example, a stateful helper may call `recompute` at any time to generate a new
+    value.
+  * Are a superset of traditional function-based helpers. They can do more,
+    but in many cases a traditional helper is appropriate.
+
+They fill a gap in Ember's current template APIs:
+
+|                  | has positional params | has layout | can yield template | has lifecycle, instance | can control rerender |
```
wycats: I don't necessarily feel like this table is adding a lot to understanding the problem, and it makes it look like each of the columns is a legitimate "primitive knob" that you might want to tune separately. I also think the comparison with Component makes it seem even more like there are knobs being tuned.

The way I'd talk about it is:

* Components are stateful and manage an area of DOM
* Helpers produce a single value that can be passed into other helpers, but don't directly control an area of DOM
  * `Ember.Helper`s have a full lifecycle and access to services, and have the ability to trigger a recomputation
    when something external changes.
  * There is a shorthand for `Ember.Helper` that you can use if you have no dependencies on external services.

active/0000-stateful-helpers.md
```diff
@@ -0,0 +1,322 @@
...
+|------------------|-----------------------|------------|--------------------|-------------------------|----------------------|
+| components       | Yes                   | Yes        | Yes                | Yes                     | Yes                  |
+| stateful helpers | Yes                   | No         | Yes                | Yes                     | Yes                  |
+| function helpers | Yes                   | No         | Yes                | No                      | No                   |
+
+Example usage:
+
+```js
+// app/helpers/full-name.js
+import Ember from "ember";
```
wycats: This example doesn't show how recomputation is triggered.

As a side note, I kind of want `Ember.on` to support paths:

```js
export default Ember.Helper.extend({
  nameBuilder: Ember.inject.service(),
  didUpdate: on('nameBuilder.change', function() { this.recompute(); }),
  compute(params) {
    const builder = this.get('nameBuilder');
    return builder.fullName(params[0], params[1]);
  }
});
```

tomdale:
Awesome start, @mixonic.

I'd prefer to descope using stateful helpers in block form and focus solely on the subexpression case. Once we've seen how that shakes out, we can circle back around and better understand the use cases for the block form.

Personally, I'd rather push people in the direction of components if they need block semantics, especially now that @ef4 has implemented positional params. The biggest downsides of Components today is that they cannot be used in subexpressions, which IMO should be the major focus of this RFC.

Lastly, the block form tempts people to make a comparison of stateful helpers to Components. By eliminating the block form for now, you can dramatically simplify and clarify the narrative: Stateful helpers are just regular helpers that have access to services, and here's the syntax for building them.

I agree with the above comments that `recomputeObserves` isn't the ideal name. We should also consider a syntactic sugar for recomputing in response to events on services and not just property changes, although perhaps that's outside the scope of this RFC. (Supporting a `.on('service.eventName')` syntax, perhaps?).

active/0000-stateful-helpers.md
wycats: As another side note, I think we should be exploring using decorators to make APIs like this nicer:

export default Ember.Helper.extend({
  @recompute({ on: 'change' })
  nameBuilder: Ember.inject.service(),
  compute(params) {
    const builder = this.get('nameBuilder');
    return builder.fullName(params[0], params[1]);
  }
});

active/0000-stateful-helpers.md
```diff
@@ -0,0 +1,322 @@
...
+
+```hbs
+{{full-name 'Bigtime' 'Beagle'}}
+{{input value=(full-name 'Gyro' 'Gearloose') readonly=true}}
+```
+
+# Motivation
+
+Helpers in Ember are pure functions. This make them easy to reason about, but
+also overly simplistic. It is difficult to write a helper that recomputes due to
```
wycats: I think I would say this a little bit differently.

Historically, helpers in Ember have had access to a lot of ambient information in the scope. This made them powerful, but also error-prone. In practice, they were used by addon authors to implement powerful functionality.

Glimmer does not want to expose the guts of the scoping system to user code, so our initial stab at helpers was to make them pure functions: data in, data out. That works well for a lot of cases, and is very ergonomic, but is a regression for addons that need the value they emit to be based on external state. This proposal is an attempt to recoup enough expressiveness so that these addons can retool for Glimmer without exposing the guts of the scoping system to addons again.

active/0000-stateful-helpers.md
```diff
@@ -0,0 +1,322 @@
...
+
+* Helper names should use a `-`.
+* Helper names that do not use a `-` must be registered via
+  `Ember.Helpers.registerHelper`.
+* In Ember-CLI, helpers live in `app/helpers/full-name.js`, or with pods in
+  `app/full-name/helper.js`.
+
+### Defininition and lifecycle
+
+A stateful helper is defined as a class inheriting from `Ember.Helper`. For
```
wycats: I assume that the term "stateful helper" is meant to be used in this RFC, but that in documentation we would just talk about "helpers", and talk about the function form as a shorthand.

active/0000-stateful-helpers.md
```diff
@@ -0,0 +1,322 @@
...
+  {{link-to 'login'}}
+{{/if}}
+{{my-login-button isAdmin=(can-access 'admin')}}
+<my-login-button isAdmin={{can-access 'admin'}} />
+Can access? {{can-access 'admin'}}
+```
+
+In all these cases, the helper is considered one-way. That is, it is a
+readable value and not two-way bound (toggling `isAdmin` does not impact the
+helper).
```
wycats: > toggling `isAdmin` does not impact the helper

I don't understand this.

active/0000-stateful-helpers.md
```diff
@@ -0,0 +1,322 @@
...
+
+// Note that what is called "options" in other docs is
+// called "blocks" here. This terminology should become the
+// documented norm.
+export default function full-name(params, hash, blocks) {
+  return params.join(' ');
+}
+```
+
+Or its equivilent stateful version, the following are effectively the same:
```
wycats: Typo.

active/0000-stateful-helpers.md
```diff
@@ -0,0 +1,322 @@
...
+An exclusion to this pattern is the following form:
+
+```hbs
+<div {{full-name "Webbigail" "Vanderquack"}}></div>
+```
+
+This is a legacy form of mustache usage. In 2.0 both forms of helpers should
+throw an exception when used in this manner.
+
+### Block form
```
wycats: I think we should de-scope this section for 1.13 and discuss it again for 2.1.

tomdale: Agree with @wycats about :arrow_up:. Also, I'd love to better understand the use cases for block helpers that are impossible to model with components, now that we have positional arguments.

jamesarosen: As an add on author, I agree. Subexpression use is far more important to me than block use.

active/0000-stateful-helpers.md
```diff
@@ -0,0 +1,322 @@
...
+However consuming a property from a service does not bind the data being
+displayed to that property. After `{{current-user-name}}` has been computed
+and rendered, it will never be invalidated.
+
+For this reason, stateful helpers are granted some control over their
+computation lifecycle. A stateful helper will recompute when:
+
+* A value passed via the template changes (`params` or `hash`)
+* The `recompute` method is called
+* Any key in the `recomputeObserves` array changes
```
wycats: I don't think we should add observation sugar right now. We can add sugar later (I personally think events / observables are going to be a more common pattern from services), and in the interim we will get more comfortable with even more ergonomic decorator approaches.

active/0000-stateful-helpers.md
```diff
@@ -0,0 +1,322 @@
...
+    this.recompute();
+  }),
+```
+
+# Drawbacks
+
+Stateful helpers may superficially appear similar to components, but in
+practice they have none of the special behavior of components. The intent of
+this RFC is that they remain conceptually very close to function helpers,
+but despite this intent they are a new concept for the framework.
```
wycats: I think we can avoid concept explosion by documenting function helpers as a shorthand for `Ember.Helper`.

active/0000-stateful-helpers.md
```diff
@@ -0,0 +1,322 @@
...
+After discussion and consideration it was clear that a third kind of template
+API would be very challenging to document and teach well.
+
+# Unresolved questions
+
+Helpers are one-way. Is it possible to have a two-way helper (data
+flowing upward like a mut)?
+
+Perhaps there should be hooks in place for the lifecycle, instead of relying on
+`init` and `destroy`.
```
wycats: :+1:

active/0000-stateful-helpers.md
```diff
@@ -0,0 +1,322 @@
...
+
+A [previous RFC](https://github.com/emberjs/rfcs/pull/52) explored creating a new class called Expressions, which would have
+more closely modeled the API of components (using positional params, attrs).
+After discussion and consideration it was clear that a third kind of template
+API would be very challenging to document and teach well.
+
+# Unresolved questions
+
+Helpers are one-way. Is it possible to have a two-way helper (data
+flowing upward like a mut)?
```
wycats: Can you show an example?

active/0000-stateful-helpers.md
tomdale: :+1: :arrow_up:

active/0000-stateful-helpers.md
```diff
@@ -0,0 +1,322 @@
...
+as helpers today. They have similar lifecycle hooks to components.
+
+### Naming and file placement
+
+The requirement for naming a stateful helper are the same as those for existing
+helpers:
+
+* Helper names should use a `-`.
+* Helper names that do not use a `-` must be registered via
+  `Ember.Helpers.registerHelper`.
```
stefanpenner: If ember-cli, we should likely just "auto-register" helpers in the conventional helper locations. Regardless of dash or not. That doesn't preclude having this helper, but i figured I would make this suggestion and seek some feedback.

active/0000-stateful-helpers.md
```diff
@@ -0,0 +1,322 @@
...
+```
+
+When the `params` or `hash` contents change, the `compute` method is called
+again. The instance of the helper is preserved across rerenders of the parent.
+
+The `init` and `destroy` methods can be subclassed for setup and teardown.
+
+`compute` is called with the same arguments as a function helper.
+
+### Consuming an stateful helper
```
stefanpenner: Consuming **a** stateful helper

active/0000-stateful-helpers.md
stefanpenner: maybe he meant the `mut` scenario?

active/0000-stateful-helpers.md
```diff
@@ -0,0 +1,322 @@
...
+of the HTMLBars subexpression they are powering. For example, given a helper:
+
+```js
+// app/helpers/full-name.js
+import Ember from "ember";
+
+// Note that what is called "options" in other docs is
+// called "blocks" here. This terminology should become the
+// documented norm.
+export default function full-name(params, hash, blocks) {
```
stefanpenner: I don't `full-name` is valid function name syntax

active/0000-stateful-helpers.md
```diff
@@ -0,0 +1,322 @@
...
+
+Let's step through exactly what happens when using an helper like this:
+
+```hbs
+<my-login-button isAdmin={{can-access 'admin'}} />
+```
+
+Upon initial render:
+
+* The helper `can-access` is looked up on the container
```
stefanpenner: i suspect before this, it is looked up in the `helper` registry that `registerHelper` populates.

active/0000-stateful-helpers.md
stefanpenner: :+1:

active/0000-stateful-helpers.md
mmun: :+2:

active/0000-stateful-helpers.md
mmun: I agree.

mixonic:
Thanks for the great notes all, I will be editing and writing tomorrow.

@tomdale I wanted to address or prompt you for more comments regarding:

> I'd prefer to descope using stateful helpers in block form and focus solely on the subexpression case. Once we've seen how that shakes out, we can circle back around and better understand the use cases for the block form.

The stateful helper syntax (the class) gives developers access to services. This is absolutely central to the initial reason for this RFC, addressing specific addon author pain and private API abuse. Stripping this to only describe function helpers would yield a very different RFC, and not address that pain.

Maybe I'm missing something?

jamesarosen:
@mixonic my reading of Tom's words is that this RFC should address
```hbs
{{foo "bar"}}
```
and
```hbs
{{#if (foo "bar")}}
```
but not
```hbs
{{#foo "bar"}}baz{{/foo}}
```
And that in both the former cases, `foo` has access to services and can control re-evaluation.

rlivsey:
What's the difference between a stateful block helper and a tagless component?

jamesarosen:
@wycats said,

> I assume that the term "stateful helper" is meant to be used in this RFC, but that in documentation we would just talk about "helpers", and talk about the function form as a shorthand.

That suggests that this RFC probably needs to consider the public API for that shorthand.

ember-cli/ember-cli#3710 may be of interest. I opened that because the guides and ember-cli differ about how to build pure-functional helpers, and both recommend using APIs that are currently private (though widely assumed to be public).

active/0000-stateful-helpers.md
mixonic: I don't think we need to explore this in the RFC and I will remove the note. The idea would be to define semantics for:

```
<my-component isAdmin={{mut (can-access 'admin')}}>
```

This is not immediately useful, to my imagination.

mixonic:
I've made a significant set of changes this morning:

* Re-introduce the dash requirement **WAIT DONT KILL ME**. @rwjblue and I have a more complete solution for this we still plan to include in 1.13, as it is required for us to have feature parity with 1.12 for addon authors. I am very aware this is a requirement, its omission here is just to let us focus on solving one problem at a time.
* Remove the ability for helpers to `yield` and otherwise interact with blocks. They now only return a value.
* Change some terms:
  * "Stateful Helper" -> "Helper"
  * function-based helpers are called "shorthand helpers"
* Remove the `recomputeObserves` sugar.

mixonic:
@rlivsey "What's the difference between a stateful block helper and a tagless component?"

This is exactly Tom's thought, and is why I've removed the block stuff.

mixonic:
See https://github.com/emberjs/rfcs/pull/58 for a detailed proposal about dashless helpers.

tim-evans:
How would you deal with helpers that take in promises?

jamesarosen:
@tim-evans 

> How would you deal with helpers that take in promises?

With a (full) *Helper*, this is straightforward: just call `this.recompute()` when the promise is fulfilled. For a *shorthand helper*, this wouldn't work unless you could somehow convert `promise._state` to a `Stream`.

@mixonic the current guides use `Ember.Handlebars.registerHelper`, which allows for [dependencies](http://guides.emberjs.com/v1.10.0/templates/writing-helpers/#toc_dependencies):

```js
Ember.Handlebars.helper('fullName', function(person) {
  return person.get('firstName') + ' ' + person.get('lastName');
}, 'firstName', 'lastName');
```

I see two options for this in the new world:

 1. *shorthand helpers* can *only* recompute on properties directly passed to them; everything else requires a full *Helper*
 1. the manual registration mechanism in #58 could allow for adding dependencies to *shorthand helpers*

rwjblue:
@mixonic - Can you double check the final details on this (as compared to the implementation PR), rename the file (increment the number) so we can merge?

mixonic:
@jamesarosen for now, I'd like to say you need to use the full helper syntax for additional bindings. We can improve this, just need to debate the sugar separate from shipping in 1.13 :-)

text/0053-helpers.md
```diff
@@ -0,0 +1,327 @@
...
+```js
+// app/helpers/can-access.js
+import Ember from "ember";
+
+// Usage {{if (can-access 'admin') 'Welcome, boss' 'Heck no!'}}
+export default Ember.Helper.extend({
+  session: Ember.inject.service(),
+  onCurrentUserChange: Ember.observes('session.currentUser', function() {
+    this.recompute();
+  }),
```
seanpdoyle: If this is a common occurrence, could it be achieved in a more declarative manner?

Maybe something like:

```js
export default Ember.Helper.extend({
  recomputedWhenTheseThingsChange: ['session.currentUser', 'session.isLoggedIn'],
});
```

wycats: I would like to see what cases are common before adding sugar, especially since ES7 decorators will shortly make many of these things much nicer to express.

Eptis:
Will this be in Ember 1.13 in the beta cycle or will it be in the Ember 2.0 (beta) cycle?

cibernox:
@Eptis This is already in 1.13 (although probably a 1.13.beta3 should be released)

rwjblue:
@Eptis - It will be in 1.13

PR #57, opened 2015-05-23
RFC for reference unification

wycats:
[Rendered][rendered]

[rendered]: https://github.com/emberjs/rfcs/blob/ember-data-reference-unification/active/0000-ember-data-reference-unification.md

### DIFF
--- a/active/0000-ember-data-reference-unification.md
+++ b/active/0000-ember-data-reference-unification.md
@@ -1,0 +1,362 @@
+- Start Date: 2015-05-20
+- RFC PR: (leave this empty)
+- Ember Issue: (leave this empty)
+
+# Summary
+
+This RFC adds a unified way to perform meta-operations on records, has-many relationships and belongs-to relationships:
+
+* get the current local data synchronously without triggering a fetch or producing a promise
+* notify the store that a fetch for a given record has begun, and provide a promise for its result
+* similarly, notify a record that a fetch for a given relationship has begun, and provide a promise for its result
+* retrieve server-provided metadata about a record or relationship
+
+# Motivation
+
+When we initially designed the Ember Data API for relationships, we focused on consumption and mutation of the relationship data. For example, you can retrieve the value of a `belongsTo` relationship via `get('post')`, or adding new records to a has-many relationship via `get('comments').pushObject(newComment)`.
+
+The top-level reading operations are designed to be [zalgo][zalgo]-proof: regardless of whether or not the record or relationship has been loaded already, you get back a promise for the result. Behind the scenes, this will trigger a fetch if needed, or simply return the current value if it has already been fetched. From a programming model perspective, this simplifies your code because you can handle locally-available data and remotely-available data in a single code path.
+
+[zalgo]: http://blog.izs.me/post/59142742143/designing-apis-for-asynchrony
+
+However, in sophisticated applications, there is often a need to refer to a record without triggering side effects.
+
+For example, you may want to initiate the fetch for a record or relationship yourself, and provide Ember Data with a promise representing the result of that fetch. That use-case is supported by the `store.push` API, but it has a few problems:
+
+* The `store.push` API supports pushing data once the fetch has completed, but no way of telling
+  Ember Data that a fetch has begun. As a result, any calls to `store.find` in the interim will
+  trigger unnecessary fetches.
+* The `store.push` API works only for top-level records with already-known types and IDs. It does
+  not support any way of "feeding" the data for a relationship to Ember Data.
+
+In sum, this makes it difficult to front-load work (especially asynchronous work). Instead, Ember Data is currently optimized for reacting to requests from the application layer, which is sometimes a very awkward way of structuring your code.
+
+Second, Ember Data was originally designed with APIs that refer to data and operations on data. Over time, we have come to realize that people quite often need to look at metadata about records or relationships, as well as perform meta-operations on them.
+
+Some examples:
+
+* getting the expected count of a has-many relationship before it has been fetched
+* learning whether a relationship is already loaded or not
+* examining server-sent metadata
+* working with pages of records in a has-many relationship, especially when pages are loaded asynchronously ("pagination")
+
+Third, because has-many relationships are represented as a `RecordArray`, we have been able to kludge around some of these issues by adding meta-operations to the has-many relationship itself. In contrast, belongs-to relationships have remained anemic. For example, there is no way to trigger a reload of a belongs-to relationship, whereas has-many relationships can be reloaded by calling `.reload()` on the `RecordArray`.
+
+# Detailed design
+
+This RFC proposes the addition of three new **public** APIs:
+
+* `RecordReference`
+* `HasManyReference`
+* `BelongsToReference`
+
+## Getting References
+
+* `store.getReference(type, id)`
+* `record.getReference(name)`
+
+## References
+
+### `push`
+
+```js
+/**
+  This API allows you to provide a reference with new data. The simplest usage
+  of this API is similar to `store.push`: you provide a normalized hash of data
+  and the object represented by the reference will update.
+
+  If you pass a promise to `push`, Ember Data will not ask the adapter for the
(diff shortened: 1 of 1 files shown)


igorT:
Several concerns that were raised in the Slack room:
  1. It seems like people would want to manipulate relationships based on just having the reference, and not having the full object. Is this possible?

  2. It will be easy for references to end up in the dom, see for example:
  ```  
  {{each post.comments as |comment|}}
       {{linkTo 'comment' comment.id}}
  {{/each}}
  ```
  If we want to, how do we stop this? How do we enable this use case?

  3. Lets say you want to filer all the comments whose id is not equal to 1. That CP is going to be annoying/hard to write, because you will need to be filtering on relationshipReferences, but the CP will not be able to observe those, it can only observe `comments.@each.id` or something like that.

igorT:
https://github.com/emberjs/data/issues/2264 seems relevant

thomassnielsen:
I like this a whole lot, and it would solve many of the use cases where we today have unnecessary fetches against the server and/or usage of private APIs. I also agree that the drawbacks seems worth it as described. 

What I'm missing currently is details on how (and if) this will affect current APIs. Especially what Igor describes in point 2, about references not leaking into record arrays (see https://github.com/emberjs/data/issues/2375 for a similar issue). As I see it, the current APIs should maintain only showing loaded records, and references should only show up when using the new reference APIs. Any and all template usage should only show loaded records. I think it would be helpful if this is more explicitly stated in the RFC (if this is indeed the intent).

I also think what Igor describes in point 1 should be implemented if possible without increasing complexity too much. Being able to manipulate relationships based on reference would save a lot of API calls and increase performance in several apps I'm currently working on.

sandstrom:
Looks great! We have a few cases of determining if a reference exist without triggering a server-fetch. This would allows us to move off the private APIs we're currently using.

***

Would it make sense to allow the adapter to sub-class these?

For example, a JSON-API adapter may want to sub-class `HasManyReference` to add a `loadNext` method, utilizing [pagination metadata](http://jsonapi.org/format/#fetching-pagination) to load, say, the next 100 comments for a post into the existing `RecordArray`.

wycats:
> It seems like people would want to manipulate relationships based on just having the reference, and not having the full object. Is this supported behavior? Basically setting a belongsTo to a reference.

The current proposal does not support mutation, but I would be happy to add it if people need it. Perhaps: `reference.update({ ... })`. I'm not sure if it makes sense to support updating everything (like `id`) but maybe it is?

> It will be easy for references to end up in the dom, see for example:

I don't understand how your example puts references in the DOM. Can you give me the JS code that produces this?

> How do we enable this use case?

Can you flesh out the use-case?

> Lets say you want to filer all the comments whose id is not equal to 1. That CP is going to be annoying/hard to write, because you will need to be filtering on relationshipReferences, but the CP will not be able to observe those, it can only observe comments.@each.id or something like that

This proposal doesn't address that case. Do you think it's important to solve it now? If so, I can give it some thought.

igorT:
```
{{each post.commentReferences as |comment|}}
     {{linkTo 'comment' comment.id}}
{{/each}}
```

fivetanley:
More discussion on getting the ids in a template here as well: worth reviewing https://github.com/emberjs/data/issues/3025

cc @chancancode

igorT:
Another use case is to want to bind to `isLoading` or `isReloading` of the relationship

wecc:
Comment from @igorT:
> @tomdale and @wycats are to write up a nice example on how to get IDs and how to use isLoaded

jdurand:
I used to do this the non-public (dirty) way with `province.data.country_id`. I upgraded to ember 1.13.2 and ember-data 1.13.4 and the foreign key ids are not accessible through `model.data` anymore.

I'd like to quick-fix this with a dirty (or not) solution.
Did anyone figure out how to do this without having to patch their API server?

active/0000-ember-data-reference-unification.md
```diff
@@ -0,0 +1,362 @@
...
+
+```js
+/**
+  If the `remoteType` is `link`, the URL to use to load the relationship.
+
+  @property
+*/
+```
+
+### `ids`
```
fivetanley: i think this should be `id` as it is a belongsTo and will only have one id, thus not being plural ids

fivetanley:
I really like the direction this is headed. It seems like a lot of things that were not possible using public APIs are now possible in JavaScript land. I'm wondering what this means in templates though; it seems like it may be a bit cumbersome to get these references yourself every time and define them in a routable component.

API of my dreams:

```handlebars
{{#if comment.isLoaded}}
  your commentID is {{comment.id}}
{{else}}
  <button {{action "load comment"}}>
    Load comment
  </button>
{{/if}}
```

With this current RFC, I don't think this is quite possible without always defining isLoaded, etc yourself in every component or every model.

arenoir:
@jdurand I am also looking for a quick fix to accessing foreign keys.  Did you have any success?

jdurand:
@arenoir no, I haven't had time to dig into this yet.

arenoir:
@jdurand okay here is my "quick fix". Added overrode the normalizeAttributes hook in the serializer to copy the foreign key.  IMO the foreign_key should be public and not be thrown away.

```
//models/checklist-item.js
  timeEntry: belongsTo('time-entry', {async: true}),
  timeEntryId: attr('string'), //set in serializer
});

//serializers/checklist-item.js
export default DS.ActiveModelSerializer.extend({
  normalizeAttributes(typeClass, hash) {
    var foreignKey = hash['timeEntry'];
    
    if (foreignKey) {
      hash['timeEntryId'] = foreignKey;
    }

    return this._super(typeClass, hash);
  }
});
```

I think this a pretty standard use case. Hope this helps.

sandstrom:
This would be a great addition to Ember Data! :boat:

@wycats do you want to wait for more feedback, or is this RFC ready for activation?

wycats:
@sandstrom I'm happy with it, assuming it gets further iterated on over time to handle use-cases we missed in the initial design.

sandstrom:
@wycats Sounds great! I've tried to compile what people have said above (please add if I've missed something).

- [ ] Should mutation be added, or is reading enough for the first version?
- [x] References ending up in the DOM[A]. As I've understood it this won't happen with this API.
- [ ] Use-case: "Filtering all comments whose id is not equal to 1". Should this be supported? Current proposal doesn't cover this.
- [ ] How will `isLoaded` be handled? It was said above that Tom and Yehuda would write up an example. Is such an example still needed?

This would be *very* useful in our app, we're currently going deep into private APIs to work around this[B]. I'd rather see a smaller implementation soon, than wait for the utopian RFC covering all use-cases. :smile: 

Carbon copy: @igorT @wecc @fivetanley 

***

*A. Leaking references into DOM*
```handlebars
{{! will this leak references? I don't think it will, but please correct me if I'm wrong}}
{{each post.commentReferences as |comment|}}
  {{linkTo 'comment' comment.id}}
{{/each}}
```

*B. Brittle Workaround*
```js
// brittle workaround
model._internalModel._relationships.initializedRelationships.branches.canonicalState.length
```

sandstrom:
Friendly ping :smile: @igorT @wecc @fivetanley @wycats 

This is a great proposal (that would solve a concrete pain in our app).

On the checklist above, what are your thoughts? (It's a list of all things mentioned in this thread, I think this should be ready as soon as you guys feel that they are resolved; can add to the list if you feel something is missing)

sandstrom:
Should this be closed now, since https://github.com/emberjs/data/pull/3303 is merged?

fivetanley:
merged as 074c798138fc9d4206c609182614cf3bf24b5fc9. thanks for the reminder @sandstrom!

ef4:
FYI, I just edited the above original post to make "[Rendered](https://github.com/emberjs/rfcs/blob/ce19612e447e0ec393f842bfb31add96edaba9a1/text/0057-ember-data-reference-unification.md)" be a valid permalink. It was broken because it pointed at a file under `active` on `master`.

PR #58, opened 2015-05-24
Helper listing (dashless helpers)

mixonic:
[Rendered as HTML](https://github.com/mixonic/rfcs/blob/helper-listing/active/0000-helper-listing.md)

A proposed solution for dashless helpers in 1.13. You should likely read [#53](https://github.com/emberjs/rfcs/pull/53) before this.

### DIFF
--- a/active/0000-helper-listing.md
+++ b/active/0000-helper-listing.md
@@ -1,0 +1,166 @@
+- Start Date: 2015-05-24
+- RFC PR: (leave this empty)
+- Ember Issue: (leave this empty)
+
+# Summary
+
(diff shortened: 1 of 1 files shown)


jamesarosen:
One place where this might not work is in component-integration tests. Unless the "boot time discovery" phase is run on constructing a container (rather than during an actual app boot), those integration tests won't be able to find helpers.

rwjblue:
It will not require app boot. Component integration/unit tests are very much awaited by this RFC.

jamesarosen:
Veronica Mars Over-the-Moon Face

rwjblue:
Implemented in https://github.com/emberjs/ember.js/pull/11393.

PR #60, opened 2015-05-25
Component Unification (angle brackets)

wycats:
[Rendered](https://github.com/emberjs/rfcs/blob/angle-bracket-components/active/0000-component-unification.md)

### DIFF
--- a/active/0000-component-unification.md
+++ b/active/0000-component-unification.md
@@ -1,0 +1,215 @@
+- Start Date: 2015-05-24
+- RFC PR: (leave this empty)
+- Ember Issue: (leave this empty)
+
+# Summary
+
+This RFC proposes a unification of the way that components and their containing element are defined. It also unifies "element" components and "fragment" or "tagless" components.
+
+# Motivation
+
+The motivation of this RFC is to unify:
+
+* a component's template with its outer element
+* element-style components with fragment-style components
+
+In particular, it is an attempt to eliminate the gap between templates and APIs like `attributeBindings`, `classNames`, `classNameBindings`, `tagName`, `elementId` and others on the component itself. It is also an attempt to allow fragment-style components ("tagless components") to specify one of its elements to use as the root element for event handling and attribute shadowing.
+
+In brief, an invocation of an angle bracket component would be replaced with the rendered contents of its layout in all cases:
+
+```handlebars
+{{!-- application.hbs --}}
+<my-button content="hello" />
+```
+
+```handlebars
+{{!-- my-button.hbs --}}
+<button>{{attrs.content}}</button>
+```
+
+This invocation would result in (roughly):
+
+```handlebars
+<button>hello</button>
+```
+
+This also works if you have multiple top-level elements in your component layout:
+
+```handlebars
+{{!-- my-button.hbs --}}
+
+<p class="open"></p>
+<p>{{attrs.content}}</p>
+<p class="close"></p>
+```
+
+The same invocation as above would result in:
+
+```handlebars
+<p class="open"></p>
+<p>hello</p>
+<p class="close"></p>
+```
+
+At the basic level, this mental model makes it easy to use components for all kinds of different abstractions, including abstractions in special HTML contexts, like `<table>`s, `<select>`s and SVG.
+
+# Detailed design
+
+When a component is invoked using angle brackets, that invocation is replaced by the component's template.
+
+> This is different from today's curly-brace semantics, which replace
+> the invocation (`{{my-button}}`) with an element managed by Ember,
+> unless the component specifies `tagName: ''`. If a curly-brace
+> component is "tagless", it cannot handle events, and has no access
+> to its DOM element.
+
+The component's template (sometimes referred to as its "layout"), has access to the attributes passed in by the invocation as `attrs`, and the JavaScript implementation of the component can access the attributes as `this.attrs`.
+
+## The Root Element
+
+If there is only a single element at the top-level of a component's template, that element:
+
+* handles events for the component via declarative event handlers
+  (methods and `.on`)
+* is available at `component.element` and `component.$`
+
+If there are multiple elements at the root, or if there is a dynamic control flow at the root (`#if` / `else` or `#each`), an element can be marked as the root element:
+
+```handlebars
+<hr>
+<div {{root}}>
+  {{yield}}
+</div>
+```
+
+The element marked as `root` will then become the target of declarative event handlers and be available via `component.element` and `component.$()`
+
+## Propagating Attributes
+
+If there is a single root element, it will also shadow any string attributes onto itself.
+
+These templates:
+
+```handlebars
+{{!-- application.hbs --}}
+<my-box aria-role="button" title="jump ahead">Jump</my-box>
+```
+
+```handlebars
+{{!-- my-box.hbs --}}
+<div class="box-outer"><div class="box-inner">{{yield}}</div></div>
+```
+
+Would produce:
+
+```html
+<div class="box-outer" aria-role="button" title="jump ahead">
+  <div class="box-inner">Jump</div>
+</div>
+```
+
+The benefit of this shadowing is that it allows the invoker of a component to install common attributes on the root element of the content that will replace it.
+
+> In practice, we have observed that components quite often forget to shadow important attributes, and this leads to a barrage of pull requests as people request specific attributes to be added. In practice, allowing the invoker of a component some control over the root element seems like a much better default than asking every component author to remember to shadow attributes.
+
+This shadowing only applies to attributes provided as strings in original invocation. This means that an invoker of a component can avoid shadowing any attribute by writing `name={{"Yehuda"}}` rather than `name="Yehuda"`.
+
+The resulting rule is very simple, and gives total control over shadowing, by default, to the invoker:
+
+* Any attribute *specified as a string* will be shadowed onto the root
+  element, if one exists.
+
+### Shadowing Attributes Elsewhere
+
+A component author may wish to specify the exact element to shadow the outer attributes because there is no single root element:
+
+```handlebars
+{{!-- application.hbs --}}
+<my-box aria-role="button" title="jump ahead">Jump</my-box>
+```
+
+```handlebars
+{{!-- my-box.hbs --}}
+<hr><div class="box" {{splat-attrs}}>{{yield}}</div><hr>
+```
+
+Would produce:
+
+```html
+<hr>
+<div class="box" aria-role="button" title="jump ahead">Jump</div>
+<hr>
+```
+
+### Disabling Attribute Shadowing
+
+In some cases, a component might want to completely disable attribute shadowing. While it is always possible for a component invocation to avoid shadowing even string attributes (via `attr={{"value"}}`), it may sometimes be useful for a structural component built together with the rest of its app to disable shadowing.
+
+> The reason to make shadowing the default and not vice versa is that **the risks of the wrong default are not symmetrical**. If we chose to make shadowing off-by-default and an addon didn't explicitly enable it, the only solution for apps would be to submit a pull request to the original addon. (This adds additional, potentially fatal friction to attributes that are used for accessibility.) In contrast, by making it on-by-default, an app can avoid shadowing by using a non-string attribute. While annoying, it can be done without changing the code of the component, and the resulting code is reasonably terse and understandable.
+
+To explicitly disable shadowing attributes:
+
+```handlebars
+{{!-- my-box.hbs --}}
+<div class="box" {{splat-attrs false}}>{{yield}}</div>
+```
+
+This should be done with care, especially in addons, as it prevents the invoker of a component from including useful attributes on the root element. **This is especially problematic for accessibility, because it makes it difficult to install ARIA attributes, `alt`, `title` and others that are critical for assistive technologies.**
+
+## Pedagogy
+
+In order for this to work, it is important to teach the component model matter-of-factly. "When you use a component it is replaced with its contents." "If it has a single root element that element is used for event handling and automatically shadows attributes."
+
+We should design educational materials alongside the implementation work of this RFC to see whether the differences between single-element components and multi-element components overwhelm the similarities for the purposes of explanation.
+
+If they do, it may be worth front-loaded the differences and go with the `<element>` alternative, but if the similarities dominate (as this proposal expects), the pedadogy benefit will be strong.
+
+It will be important to attempt to teach the multi-element case as being the same as the single-element case, but without the benefit of the default root (which is a somewhat obvious limitation).
+
+# Drawbacks
+
+The first major drawback is the change from Ember 1.0 curly-brace components, which had an Ember-managed element. However, one major drawback of that system is that it was necessary to manage the attributes, classes, tag name, and ID through additional JavaScript APIs, despite the fact that other APIs existed in the template.
+
+As the templating layer has gotten better (most notably with the removal of `bind-attr`), the gap between the templating layer and the medley of APIs to do the same thing in JavaScript has grown. This proposal re-unifies them.
+
+Second, this proposal attempts to provide good defaults for root element handling and attribute splatting. While the risks of the wrong default are not symmetrical (as described above), it is possible that the default behavior is too aggressive.
+
+Finally, this proposal makes Ember components inconsistent with web components, which manage an outer element. However, the fact that this makes web components unsuitable for abstracting content in special contexts (tables, select boxes, and SVG) and poorly suited for abstracting multiple elements in general is an argument in favor of the "replacement" strategy.
+
+This is the strategy used by React.js, and it works quite well. The mental model of "when you invoke this component, it is replaced with its content" is intuitive, quite easy to teach, and works well with the CSS we have today.
+
+# Alternatives
+
+One major alternative would be to make the attribute splatting opt-in rather than opt-out. It is reasonably easy to teach people how to splat attributes, and adding additional implicit behavior is often frustrating.
+
+The reason this proposal chose to make it opt-out is described in "Disabling Attribute Shadowing": the risks associated with forgetting to opt-in are much higher than the risks associated with forgetting to opt out.
+
+That said, since there is an easy way to opt into splatting arguments (`{{splat-attrs}}`), it might be enough to encourage the authors of reusable attributes to include a splat in their definitions. On the flip side, if everyone has to remember to write it, maybe it's a good default!
+
+Another alternative would be to stick with the original design from Ember 1.x and have components manage their own element. However, the drawbacks of this system are acute, and this solution is almost strictly better than the `attributeBindings`, `classNameBindings`, etc. world of Ember 1.x.
+
+Yet another alternative would be to wrap templates with `<element>`, which would mark the part of the template owned by the component. The `<element>` tag would be able to specify attributes like any other tag, which would eliminate the need for the Ember 1.x parallel API in JS, and make it easy to see the entire structure of the component in the template.
+
+However, it's not clear why this:
+
+```handlebars
+<element tagName="content" class="box">{{yield}}</element>
+```
+
+Is better than this:
+
+```handlebars
+<content class="box">{{yield}}</content>
+```
+
+One of the goals of this proposal is to make the basic mental model simple, and "the component invocation is replaced with its template" is simple and easy to understand.
+
+When precisely to handle events, and whether to splat arguments are worth exploring. In the opinion of this proposal, those questions are easily amenable to the usual analysis of defaults: what is the common case, and what are the relative costs of each decision.
+
+With all of that said, opt-out can sometimes be more frustrating to people who did not expect the default behavior, even when they understand the rationale for the tradeoff.
+
+# Unresolved questions
+
+* Should we go with `<element tag="foo">` as an explicit marker of
+  the root element and an opt-in for auto-splat?
+* Should we make splatting opt-in instead of opt-out?


rwjblue:
I love the direction with this, but had a few questions while reading:

* Whitelist API for using `<some-thing-as-div></some-thing-as-div>`: see https://github.com/emberjs/ember.js/issues/11269 for an example.
* What happens to the existing JS-land API's for these things (`attributeBindings`, `classNameBindings`, etc)?
* Does this mean that every component must have a layout? If not, would the existing JS API's take effect?
* Interop with current curly brace invocation semantics:
  * How can component authors handle both cases since it is up to the caller to use the "right" way?
  * Will helper methods be available to know if a given component was invoked as angle brackets vs curlies?
  * What will happen if `{{root}` or `{{splat-attrs}}` is seen when invoked via curlies?

wycats:
> What happens to the existing JS-land API's for these things (attributeBindings, classNameBindings, etc)? Are they additive to the marked {{root}} element? Are they ignored in angle bracket invocation?

I think they would be an error when used with angle-bracket components.

> Does this mean that every component must have a layout? If not, would the existing JS API's take effect?

If you want to describe anything about the component (tags, classes, attributes), a layout is much more expressive than the JS APIs, so "must have a layout" -> "doesn't need the JS APIs" :wink: 

> Interop with current curly brace invocation semantics:

This is the stickiest question you raised and I don't have great answers yet. To flesh it out:

> If I'm a component author and I want to be compatible with 1.12/1.13 curly bracket components, 2.1 curly brace invocation, and 2.1 angle bracket invocation, how do I do that?

I don't have a good solution in mind that supports doing all of the above in the same codebase; I kind of feel like you may need to maintain a separate branch for pre-2.0 components :frowning: 

I'm open to anything though, and would love some brainstorming here.

rwjblue:
> I kind of feel like you may need to maintain a separate branch for pre-2.0 components

The main issue isn't with maintaining branches for supporting different Ember versions (that may just be a fact of life in some cases), but for maintaining branches for supporting the two different invocation styles within the *same* Ember version.

stefanpenner:
Another point we likely need to investigate is what happens with double curly components today, that contain a slash. `{{users/my-form}}` is valid today, but does not translate nicely to DOM. We could transition / to : which does translate. But that assumes we don't want to use : for something else.


Also what's the order of precedence between native WebComponents and ember component lookup. I suspect ember components should take a higher precedence. But what is the strategy when a collision occurs? What tagName is used? I suspect these collisions will be frustrating to debug if we do not message and provide escape hatches.


A finally point is the proposed scoped helper/component yielding + dot syntax. As the dot form is valid, it may be an WebComponents (unless they are more limited then document.createElement)

```hbs
<my-form as |f|>
  <f.input>
</my-from>
```

How are theoretical collisions with `<f.input` dealt with?

I suspect the dot versions of tags will not be used, as css rules against them get pretty gnarly

igorT:
Does the class get overridden or do they concat when shadowed? It seems like concatting the classes would be very useful for customizing styling of the component, without having to copy all the original styles.

jonathanKingston:
I'm struggling to see the motivation to unifying 'element' components with 'fragment' components.

For me this adds further risks that large changes will come to the component stack when keeping them conceptually separate (perhaps unifying the code base under the hood) will alleviate the risk for change later. Separating a unified concept seems harder than unifying in my opinion.

I like the idea of:
- Components using custom tags by default
- Their templates not requiring the top level tag, which is specified in the template
- The tag is injected into the DOM with the properties the template specified it with

This would remove the need for all the bindings as:
```
{{!-- application.hbs --}}
<my-component class="thing" other-attr="thing" />
```

```
{{!-- my-component.hbs --}}
<div>
  {{other-attr}}
</div>
```

Renders:
```
<my-component class="thing" other-attr="thing">
  <div>
    thing
  </div>
</my-component>
```

Partials or fragments could mop up the rest as they currently do.


This roughly relates to my attrTypes RFC too: https://github.com/emberjs/rfcs/pull/35 which is designed to help filter attributes passed into the component template (So template authors have to conform to a specified API). This mapping could also be responsible for attributes that don't really make much sense being visible to the DOM like actions (unless a textual alternative could be used instead like a uuid which might help with debugging).

jonathanKingston:
Also is it worth waiting for the slots syntax to calm down a little as I think this would likely solve the other issues here too.

For an easier migration could there be a property I could set on the components JavaScript to disable the replacement? All my components extend from a base level component so it would make a simple transition to disable this (However I would prefer the separation of fragment/partial style components to normal ones).

wycats:
> the slots syntax

which syntax do you have in mind?

wycats:
> Does the class get overridden or do they concat when shadowed?

They concat. It would make sense to concat styles too.

jgwhite:
> This shadowing only applies to attributes provided as strings in original invocation. This means that an invoker of a component can avoid shadowing any attribute by writing name={{"Yehuda"}} rather than name="Yehuda".
>
> - The resulting rule is very simple, and gives total control over shadowing, by default, to the invoker:
> 
> Any attribute specified as a string will be shadowed onto the root element, if one exists.

I may have missed something, but with this rule how would one bind a dynamic property to an attribute?

i.e.

```hbs
<x-foo aria-label={{someComputedLabel}} />
```

wycats:
@jamiewhite simple:

```handlebars
<x-foo aria-label="{{someComputedLabel}}" />
```

jgwhite:
@wycats ah, let me check I’ve got this straight:

```
foo={{bar}} ---> this.attrs.foo = bar
foo="{{bar}}" ---> this.element.foo = bar
```

wycats:
@jamiewhite both cases would assign to attrs, but string attrs would also setAttribute.

jgwhite:
Okay cool, that makes sense. Another thing I’m not 100% clear on from the RFC:

If I want the in-DOM result of invoking:

```hbs
<x-foo>Bar baz</x-foo>
```

to be:

```html
<x-foo><span>Bar baz</span></x-foo>
```

What should `x-foo.hbs` look like?

wycats:
@jamiewhite

```handlebars
<x-foo><span>Bar baz</span></x-foo>
```

jgwhite:
@wycats and if I wanted to recursively invoke `<x-foo />` ?

wycats:
@jamiewhite recursive invocations always require control flow so they wont be at the top of the template.

jgwhite:
@wycats this is making more and more sense as I reason it through.

My initial concern was that in templates like this:

```hbs
{{! x-foo.hbs}}

<x-foo>
  {{yield}}
  {{#if somePredicate}}
    <x-foo>Some consequent</x-foo>
  {{/if}}
</x-foo>
```

The difference in behaviour between the root `<x-foo>` and inner `<x-foo>` would cause confusion. Looking at it now however, the abstraction seems solid.

wycats:
@jamiewhite yeah I worked through it a bunch and came to the conclusion that any plausible recursion case requires some kind of control flow in order to avoid stack overflow.

You can think of the outermost tag as the component's signature, and any inner tags as recursive calls. @tomdale thinks we should make this the default conventional approach for templates that don't explicitly require a special tag for some reason.

jgwhite:
> You can think of the outermost tag as the component's signature, and any inner tags as recursive calls.

Yeah, that’s very clear and *very* declarative.

> @tomdale thinks we should make this the default conventional approach for templates that don't explicitly require a special tag for some reason.

Could you elaborate on this point? Do you mean that a conventional component template wouldn’t have an explicit signature element?

jonathanKingston:
@wycats the slots syntax as is being pushed for Shadow DOM:
https://lists.w3.org/Archives/Public/public-webapps/2015AprJun/0187.html

This appears to be what will land in browsers and I think will solve a lot of the same use cases this is solving.

---

I understand the merit of the duplicate tag within the template of the DOM, however personally that is a whole lot of components that need changing for something that improve only the 'partial' kind. This improvement isn't (as far as I am aware) ever going to be native to HTML. Much like services are essentially just `Ember.Object`s wouldn't it be safer to leave the concepts distinct but perhaps share the same code bases?

wycats:
@jonathanKingston I'll reply to this more later, but fwiw, I don't think trying to map Ember's semantics very closely to hypothetical web components semantics is a winning strategy.

Web Components are still at a very early stage, and I personally think React's experience (and Ember 1.x's, for that matter), bears more heavily on the problem space than a future WC spec. If Ember and React both land on something inexpressible in WC, and it works out smashingly, that would be a good opportunity to attempt to persuade the standards process to make it expressible.

jonathanKingston:
@wycats My issue isn't in driving the standards development at all with extend the web forward principles. However collapsing both concepts into one before they are complete seems a little premature.

Also as far as I was aware there was consensus from major browsers on this syntax.

jayphelps:
I'm personally against allowing multiple top level elements w/ marking via `{{root}}`. I feel like it's not inline with the future of W3C web components and something we'd just end up eventually deprecating. For consumers of your components, it also makes it more confusing for CSS selectors since the sibling elements extraneously appear...rather unintuitive IMO.

jayphelps:
> However, the fact that this makes web components unsuitable for abstracting content in special contexts (tables, select boxes, and SVG) and poorly suited for abstracting multiple elements in general is an argument in favor of the "replacement" strategy.

I did read this btw. I just haven't concluded this is indeed the best approach. Still digesting..will likely have more targeting queries and comments.

active/0000-component-unification.md
```diff
@@ -0,0 +1,215 @@
...
+However, it's not clear why this:
+
+```handlebars
+<element tagName="content" class="box">{{yield}}</element>
+```
+
+Is better than this:
+
+```handlebars
+<content class="box">{{yield}}</content>
```
jayphelps: *the `<content>` tag name here is unintentionally confusing to people familiar with the unrelated Shadow DOM*

I completely agree that using the tag vs. `<element tagName="x-foo">` is superior. After all, it's what will output, why abstract it?

jamesarosen:
If a single running app can only have curly components *or* angle-bracket components, but not both, then upgrading a project will require doing all of the following simultaneously:

 * turn on angle bracket components, either by upgrading ember or setting a flag
 * upgrade all addons to their angle-bracket versions
 * upgrade all components in the project

Given what I know of addon maintenance and app schedules, that will never happen for any but the smallest projects.

Is there no way of letting individual components opt-in to the new behavior? It would be very nice if Ember could raise an exception when an app used an angle-bracket component via curlies.

igorT:
My understanding was that an app would be able to mix and match, but that a single component is either an angle bracket or curly component

workmanw:
> that a single component is either an angle bracket or curly component.

Do you mean a single instance of a component or a single component class?

If you're referring to the later, that would be a pretty rough upgrade. Instead of migrating templates one by one, you'd have to cross the other vertice and upgrade every usage of a single component one by one. This would be much more tedious and make testing (actual QAing) much harder.

jamesarosen:
@workmanw The component's template either includes the outer HTML or the inner HTML. It *might* be possible to do per-instance selection with some sort of additional flag like
```hbs
<my-component exclude-wrapping-tag=true>
```
or
```hbs
{{my-component include-wrapping-tag=true}}
```
but that seems... *difficult* at least.

slindberg:
For those interested, [here is the issue in React](https://github.com/facebook/react/issues/2127) discussing whether components should be able to return multiple top-level elements (they currently do not). The TL;DR is that it's useful for avoiding wrapper elements for certain CSS and event handling situations, but comes at a cost ([from that issue](https://github.com/facebook/react/issues/2127#issuecomment-106810565)):

> A big issue is that you lose the technical, practical and intuitive assumption of one component = one element/node

**IMO, the decision boils down to whether or not it's important to maintain a direct conceptual mapping between DOM elements and components.**

> One of the goals of this proposal is to make the basic mental model simple, and "the component invocation is replaced with its template" is simple and easy to understand.

Multiple root elements run counter to this goal by making rendered DOM difficult to reason about. I think React's one component = one element assumption has immeasurable value that outweighs the headaches that people will experience with the edge cases.

However, it's worth considering that since the argument for a single root node is architectural and the argument against is born out of real developer frustration, there will *always* be people pushing for it.

tim-evans:
For me, this feels like the correct solution. However, I'm a bit reticent to this change because of the scope. If this lands, this should be before angle brackets become a thing in Ember, otherwise there is an overhead to those using angle brackets having to convert from an "old" syntax to a "new" synatax (where the old syntax is barely months old).

Could there be a discussion with this RFC on the migration process? I assume at some point ember will have an end of life for curly brackets, and we'll have to jump on the angle bracket train. If we're to do so, it would be a boon to have tools that aid in the conversion of these.

As for the conversation about `{{root}}` and whatnot, I am of the opinion that components that have multiple root elements should have to manage that in the component. I think having `this.$()` return multiple children is reasonable and less surprising. `this.element` on the other hand, should probably be handled by the author of the component. I'm still waffling on how I feel specifically on these points, but I feel like it might be best to go with a single root, then have another RFC for multiple roots. That way we can move forward the implementation of the unification and still discuss and work out the details of multiple roots.

rwjblue:
> If this lands, this should be before angle brackets become a thing in Ember, otherwise there is an overhead to those using angle brackets having to convert from an "old" syntax to a "new" synatax (where the old syntax is barely months old).

Yes, this is exactly why we decided to disable angle-bracket invocation for 1.13 and 2.0.

> I assume at some point ember will have an end of life for curly brackets, and we'll have to jump on the angle bracket train.

Perhaps, but that is a LOOOONNGGGG way off (curly bracket invocation for components will have to continue to work until at least 3.0).

jayphelps:
After having discussed this in depth with mixonic, ebryn, and stef, my original position still stands. I firmly believe losing 1:1, component:element is not a good idea. i.e. allowing multiple top-level elements. Everything else about this proposal I am 100% behind.

Summary of my biggest concerns:

* It doesn't make HTML/angle-bracket sense, now or IMO any possible near future HTML spec.
  * `<x-foo>` implies HTML (hence why we're moving towards it) but in every single case of true HTML, no element will magically be replaced with multiple elements. `::before` and `::after` exist to (partially) fill in this gap in the HTML world, without causing selector havoc and making things unintuitive.
* It makes the Component class behave differently depending on whether there is a "root" or not (one top level element or one marked via `{{root}}`).
  * Complicates mixins. Now you have to know whether a mixin works with root-ed components or not, e.g. expecting `component.element` or `component.$` and/or the author of the mixin has to take both cases into account.
* The use case where multiple top-level elements is needed can be solved with `{{partial}}` (or similar) or even better, a new helper specific for fragments. Something like `{{fragment "x-foo"}}` which could certainly still be backed by a class and have lifecycles, but both the author and consumer of the fragment clearly knows it doesn't have `this.$`, `this.element`, possibly has multiple top-level elements, etc. Again, there is no declarative HTML equivalent and can't imagine one that behaves like this; this belongs in curly braces.

balinterdi:
Don't want to derail the –very interesting– discussion so just one tiny question: would angle-brackets components still introduce one-way bindings as it was planned before?

stefanpenner:
> Summary of my biggest concerns:

Ya, I have been mildy un-interested in the fragment case. I am likely missing something, but it feels quirky, and the cost/benefit doesn't yet make sense to me.

I would love to see someone thoroughly enumerate the motivation, must-have scenarios, and finally the pitfalls.

tomdale:
@stefanpenner I'm surprised to hear you say you don't understand the case for fragment functionality, just because in my mind it's one of _the_ most asked questions when developers are learning how to build components.

Previously, usage of the `Ember._Metamorph` mixin was extremely common. Only in the last one or two releases did we make that (widely-used) private API into a public API by supporting `tagName: ''` to get the same effect. Now that we are eliminating `tagName` et al. and favoring an in-template description of the root element, not allowing the fragment case would be a regression in functionality that would leave many users frustrated.

For me, there are two major motivating cases (maybe more, but I haven't had my coffee yet ;):

1. Not wanting to break CSS, where DOM structure is important. For example, in Skylight I wrote a `feature-flag` component that you can use to wrap sections of the template in to feature flag them. Before I made this a tagless component, it was common for the feature flag to break our styles (e.g. `section > h2`) because it inserted a new DOM element into the hierarchy.
2. Cases where a wrapping element would be illegal. For example, imagine a component that wants to put two `<tr>`s into a table: `<table><two-rows></two-rows><tr><td>blah</td></tr></table`. It is impossible to use a wrapping element here and without this functionality there is no way to accomplish something that won't be treated as invalid by the browser.

tim-evans:
@tomdale I ended up writing a `feature-flag` helper for Archiv8. This allowed for:

```handlebars
{{#if (feature-flag 'exhibitions')}}{{/if}}
```

Just as another solution to that particular problem :wink:

DingoEatingFuzz:
> Cases where a wrapping element would be illegal. For example, imagine a component that wants to put two `<tr>`s into a table: `<table><two-rows></two-rows><tr><td>blah</td></tr></table>`. It is impossible to use a wrapping element here and without this functionality there is no way to accomplish something that won't be treated as invalid by the browser.

This and the classic `<p><p>inner p is invalid</p></p>`, and `<a><a>outer a is ignored in favor of inner a</a></a>` problems.

I think it's an unfortunate reality that no matter how angle bracket components turn out, developers will still have to think in HTML from time to time.

As to the multiple table rows, one work-around is to have the component be a `<tbody>`, since multiple table bodies is valid, even though it's :rainbow:semantically:rainbow:  different.

Although I think a best practice should be to strive to make components have a single element (for instance, I think the `<hr><div {{root}}>` example could always be avoided), there will be times when it is only logical to have two elements in one component.

If this is to be a best practice, I agree with the `{{root}}` API because it is a sane default. By doing nothing with `{{root}}`, you are following the best practice. It's only when you have to do something unusual that the tiny headaches of what element is `root`, what is `.element`, how do I get the other elements in the component comes into play.

@jayphelps, your mixins concern sounds like a real one, but after giving it a little bit of thought, I can't think of a practical mixin that may break because it needs to operate on single-element and multi-element mixins. Do you have an example?

stefanpenner:
> @stefanpenner I'm surprised to hear you say you don't understand the case for fragment functionality, just because in my mind it's one of the most asked questions when developers are learning how to build components.

I use the fragment functionality today, but it basically puts my component into a very different mode. A mode with many more caveats. 

1. this.$() becomes stale, so `this.$().plugin();` works in only a subset of scenarios.
2. event listeners are dubious
3. consumers of the component must realize the component may not be compatible with ***
  * tagName
  * id
  * class
  * any other attribute that works on regular components but not on fragments
  * event handlers
  * css
  * etc
  * animations

----

**3.** is the cause for my current hesitation, as it pokes a hefty hole in the abstraction. More concerning, is that as a consumer both forms appear identical, but in-fact are quite different.

I appreciate the intent to unify concepts, but due to the caveats the goal of having both a simple & easy universal primitive here are at odds.

jonathanKingston:
My issue with the angle bracket syntax is where additional HTML features are required like pipes and namespacing etc; this functionality should remain as Moustaches to prevent it appearing as native code.

Usage in page templates makes sense when the HTML syntax doesn't need to be bent to have additional syntaxes.

Removal of having to specify tagName (this could stay if you wanted to disable this feature perhaps) and routing through the attributes to the DOM by default makes a lot of sense too.

---

Would a neater API that is separate to components remove the risk of joining the concepts. The code could be reshaped to both inherit from the same underlying code if there is in fact lots of shared 
functionality.

---

@DingoEatingFuzz any mixin that basically works on the root element has this issue.
```
this.$.attr('id', 'thing');
```
But conceptually any could break in ways like that as @stefanpenner has mentioned.

stefanpenner:
> My issue with the angle bracket syntax is where additional HTML features are required like pipes and namespacing etc; this functionality should remain as Moustaches to prevent it appearing as native code.

I don't believe I agree with this, as pipes and such are merely a superset.

This is more similar to objective-c/c relationship vs the c++/c relationship. For those unfamiliar, object-c leaves C alone, merely augmenting additional concepts. So any C code is valid objective-c code, whereas any C code is not valid C++ code

My specific concern mentioned above is the result of me becoming concerned that the fragment (that looks identical to a component) approach puts us on the path of an misaligned/improper superset. 

I believe my concerns would be totally mitigated if a component and a fragment could at a glance be visually differentiated.

jonathanKingston:
@stefanpenner I wasn't saying you agreed with that part at all :smiley:. I get that it is a superset however layering functionality on top of what looks like native HTML is all the same sort of badness that ASP, Angular and ColdFusion has done. I get that it could probably be transpiled away when there is native functionality for it however for me it's important functionality that gets lost to the eye because it is hidden within the bracket construct.
Also there is the possibility that it would never get adopted and lots of components then need rewriting conceptually.

jayphelps:
@DingoEatingFuzz I've written and seen hundreds of mixins that would break because of this..One example is a mixin that adds some common d3 ability to your component, which currently rightfully assumes there is only one element at the top and that a root always exists.

`tagName: ''` has been technically possible, so you could have theoretically had this problem already but in the real world, I personally have met very few devs who were aware of it and the ones that did thought it was private-API (or just an accidental ability) and it wasn't ever used. What it accomplishes is desirable in certain cases, I'm not arguing against that, i.e. the fragment case, backed by a class with hooks, etc. I just believe it should be a separate primitive from a component.

tomdale:
@stefanpenner I agree with you on issues 1 & 2, and is why we intend to have very clear error messages with links to guides on how to handle those issues. I think as long as the documentation is clear we will be good.

In terms of 3, `tagName=` will no longer work and the other attributes you list will be reflected on to the root element. If you know enough about the structure of a component that you want to set attributes like class on it, there is probably some inherent coupling there anyway. IMO setting classes and other things from the calling side of the component tends to be a dubious practice. However, there are legit use cases, and if there is a legit use case that involves fragments, those attributes will still be reflected on the conceptual "root" element, which is what you want.

jayphelps:
@tomdale can you address my concerns? I feel like I'm invisible in this conversation.

tomdale:
@jayphelps The issues you raise are good ones, but I am mostly focused on improving the on-ramp to Ember for new developers. Having multiple styles of components depending on whether they have a root element or not is going to confusing for new developers, in my opinion in the same vein as the confusion in Angular between factories, services and providers.

As I said to @stefanpenner, we will have very clear, explicit error messages if you try to add event listeners or use `this.$()` for the fragment case. In the future we can explore ways to make them work, with caveats that will have to be message appropriately (we have a few ideas here).

tomdale:
@jayphelps We are working the Web Components folks, as well as other framework/library authors, to better align the standard with the needs of real web development. Web Components was a very design-upfront spec and it is currently failing to meet the needs of any of the major frameworks/libraries.

jayphelps:
@tomdale I don't think it's confusing if you just don't call fragments components and don't use them with the same syntax. ala `{{fragment "x-foo"}}` or whatevs. IMO this is actually easier for new devs because they probably won't care about fragments initially so won't need to learn the additional complexity of "a component sometimes acts this way, other times differently"

tim-evans:
>  I am mostly focused on improving the on-ramp to Ember for new developers.

jayphelps:
@tomdale 

> Having multiple styles of components depending on whether they have a root element or not is going to confusing for new developers

...isn't that exactly what this RFC is proposing...?

tomdale:
> This deeply concerns me.

@tim-evans You shouldn't let it concern you. It's easy to fall into a trap where you can add incremental complexity that is easy for advanced users like yourself to absorb, but becomes a huge barrier for new entrants. Obviously new users are not my only concern; this feature is just as much for me as it is for them, and I hope you'll agree I'm a somewhat advanced user.

> Having multiple styles of components depending on whether they have a root element or not is going to confusing for new developers

@jayphelps No, the transition between the two styles is very seamless and the mental model you can describe in documentation is very simple: "Whatever you put into the component template is what you will see in DOM." I know this model feels very different to what you're used to, but having been using it for the past few weeks, I promise it is more intuitive than the existing system. I hope you can give it a try in your own apps to see what I mean.

jamesarosen:
> Whatever you put into the component template is what you will see in DOM.

This rule is great for app authors. It's less good for addon authors (especially those writing mixins for components) *if it allows multiple root elements*. If components *require* multiple root elements -- by validating the template at some point -- then it's a great rule for both.

In the past, `Ember.Component` served two purposes: *single root element* (`tagName="anything"`) and *fragment* (`tagName=""`). There are many benefits to restricting `Ember.Component` to the former, as @stefanpenner has described.

There are also benefits to having *something* to support the *fragment* case. That case would have looser restrictions on the template, but less support for other things like CSS and event handlers. For this, I like the `{{fragment}}` helper and `Ember.Fragment` class suggested above.

tomdale:
@jayphelps Pinged you in community slack.

tomdale:
@jamesarosen Why are two concepts better than one concept with clear error messages?

stefanpenner:
> In terms of 3, tagName= will no longer work and the other attributes you list will be reflected on to the root element.

My concerns are with the fragment case, in which no root or single root for attribute reflection would exist.

Let me illustrate what I believe to be a common scenario producers of HTML today would assume to be reasonable, but in the world of fragments would not.

```hbs
<div class="{{mode}}">
  <maybe-fragment-maybe-not class="hide-in-presentation-mode" />
  <video></video>
</div>
```

```css
.presenting .hide-in-presentation-mode {
  display: none;
}
```

or via the `hidden` attribute:

```hbs
<div class="{{mode}}">
  <maybe-fragment-maybe-not hidden=isPresenting />
  <video></video>
</div>
```

Now, in-order to get the expected functionality. The developer may be required wrap the `maybe-fragment-maybe-not` in an additional tag, which then would carry the class/attribute. This requires either memorizing which components are well mannered and which are imposters, or by doing a dive, and hoping the component doesn't shift out from under them.

If instead some visual differentiator existed, the consumer would not be caught unexpected.
Note: I lack the creativity to come up with anything new, so I will like use `{{` to represent something not guaranteed to behave as a well mannered tag.

```hbs
<div class="{{mode}}">
  {{i-am-clearly-a-fragment}}
  <i-am-clearly-a-well-mannered-component />
</div>
```

tomdale:
@stefanpenner Thanks for the motivating use case. This is actually the original design we came up with, but the thing that made us deviate from that plan was that semantics for curly components are already defined, and we need some syntax to opt people into the new DDAU-style components.

I am open to alternative syntax suggestions, and of course, we could implement the described behavior in Ember 2.0 if we can deprecate curly component invocation in the 2.x series.

jamesarosen:
> Why are two concepts better than one concept with clear error messages?

It depends on whether they really are two concepts. `Helper` and `Component` are two concepts because `Helper` is limited to emitting just strings. Is `Fragment` different enough? From an app-developer's standpoint, probably not. But from a `Mixin`'s standpoint, I really think so.

If the modes are distinct enough, I always prefer two classes or methods to one class or method with two modes. *Flag-avoidance* is the same reason Rails changed from `.find(:all, query)` and `.find(:first, query)` to `.where(query)` and `.where(query).first()`.

DingoEatingFuzz:
Up until this point, I thought specifying `{{root}}` was a requirement for control-flow/multiple-element components. Reading back through the RFC, it's clear that it is **not** required, so put me in the concerned camp :)

Having less API surface-area by unifying components and fragments is definitely appealing, but I think in the long run the WTFs related to property assignments falling through and mixins not behaving is greater than the WTFs related to a fragment being kinda like a component, but technically not a component.

At least in this way you can avoid fragments if you don't like the extra confusion/cognition/consequences whereas in the case of a multi-modal component you cannot.

Some examples of unavoidable confusion/cognition/consequences:
 1. How will/should this mixin work with a fragment that has a root node that may vary in tag name?
 2. How will/should this mixin work with a fragment that has no root node
 3. Where all these extra elements in my DOM coming from?
 4. Why isn't a class being set on this component?
 5. Why is the class sometimes being set on the first element in the component, but other times on a random other element?
 6. [During code review] I forget, is `<onboarding-messaging>` always a single component or is it variable? Are there circumstances where these attributes don't do what the author thinks it will do?

jayphelps:
Just an FYI @tomdale and I talked things over on slack, I think we both understand each other's points but are still in the same spot. :trollface: 

My suggested course of action: finalize angle brackets with 1:1 component, some form of separate `{{frag "x-foo"}}`/`Ember.Fragment` primitive, see how the dust settles in the real world then reevaluate unification. It's easier to merge the two than it is to break them out later, which would be a massive FU.

jonathanKingston:
>@jayphelps We are working the Web Components folks, as well as other framework/library authors, to better align the standard with the needs of real web development. 

@tomdale is this a public conversation, If so where?

ef4:
@stefanpenner: most of your list of objections I actually see as features.

> event listeners are dubious

Event listeners on the Component class are *fundamentally* dubious now that we have the component's own element in the template. You can avoid having two APIs for the same thing and just use `{{action}}` (or the newer alternatives, like `on-click=`) on the component's top-level element. This has the benefit of being 100% unambiguous, no matter how many top-level elements are in the template.

> consumers of the component must realize the component may not be compatible with: tagName, id, class...

I think moving away from those APIs is positive. When somebody wants a particular tag, it is better to let them just write the tag. So instead of:

```handlebars
<some-component tagName="tr" class="tall" />
```

You can lift the top element out into the caller's context, and implement the component as a fragment:

```handlebars
<tr class="tall">
  <some-component />
</tr>
```

The reality of the DOM as it works today is that it is not arbitrarily flexible.  There are plenty of contexts with strict rules, and plenty of ways to break existing layouts just because you wanted to add a new layer of abstraction. 

Components are a unit of abstraction that *does* need to be arbitrarily flexible. They are the function calls in your template. Splitting a big function into smaller functions should not force you to change the output of the function, but that is exactly what happens if components and DOM are forced to be one-to-one.

I am also unsympathetic to the web components argument. Ember Components are a superset of web components. We already do much that they don't. But we are still completely interoperable, and you can choose to implement a component in Polymer, and it will work fine with Ember.

jamesarosen:
> most of your list of objections I actually see as features...

@ef4 this post of yours really turned me around on this RFC

stefanpenner:
> event listeners are dubious
> Event listeners on the Component class are fundamentally dubious now that we have the component's own element in the template. You can avoid having two APIs for the same thing and just use {{action}} (or the newer alternatives, like on-click=) on the component's top-level element. This has the benefit of being 100% unambiguous, no matter how many top-level elements are in the template.

I totally agree, also as event-listeners are already contextual, a video tags `canplaythrough` is not expected to work on a `div`, so I see no problem.

> consumers of the component must realize the component may not be compatible with: tagName, id, class...
> I think moving away from those APIs is positive. When somebody wants a particular tag, it is better to let them just write the tag. So instead of:
> 
> You can lift the top element out into the caller's context, and implement the component as a fragment:
> 


I have un-intentionally lumped together tagName and class (legitimate DOM attrs). I wish for us to separate these two, as I also believe tagName usage to be an anti-pattern and your lifting example to be best practices.

So lets continue to refine these thoughts, but lets focus on the example I provided.

Given:
```hbs
<div class="{{mode}}">
  <maybe-fragment-maybe-not hidden=isPresenting />
  <video></video>
</div>
```

The lifting strategy, results in defensive structural wrapping.

```hbs
<div class="{{mode}}">
  <span hidden=isPresenting >
    <maybe-fragment-maybe-not />
  </span>
  <video></video>
</div>
```

 *note: `hidden` is just one example, another would be the accessibility improving `aria-` labels.*

It feels wrong, that applying accessibility hints results in potentially breaking defensive structural wrapping

Now, after investigation of the components implementation (or via trial & error) a developer may infer that the defensive wrapping is unneeded. Unfortunately Overtime this inference may not continue to stand true as the single-root may transition to multi-root.

The transition from single-root to multi root (or back) should be considered poor practice, but the potential does have a non negligible cost.

Finally, and what I believe to be the root of my concerns. We have two concepts, range and element, although related they both expose very different and incompatible capabilities. All while occupying the same syntax.

```hbs
<div class="{{mode}}">
  {{fragment-thing}}
  <video></video>
</div>
```

```hbs
<div class="{{mode}}">
  <not-a-fragment />
  <video></video>
</div>
```

aside: I wonder if today `<` is the new hotness, now everything looks better as an `<`.

ef4:
I like keeping curlies for fragments and angles for components so that the invocation is unambiguous, but it has practical upgrade challenges. We would need a way for the curlies to opt in to one-way binding and own-element-in-template.

stefanpenner:
> but it has practical upgrade challenges. We would need a way for the curlies to opt in to one-way binding and own-element-in-template.

I absolutely agree, @tomdale also voiced this concern.

One idea, was to introduce (until we can fully re-claim `{{`)  `{{fragment ...`. This is obviously an unfortunate interim step, but it feels like a worthwhile short term cost.

jmurphyau:
Is there any way to opt in to one way bindings and no attr proxying in curly `{{components}}`? 

I'm finding myself write `(readonly prop)` and would love being able to disable attr proxying to the component object.

I know angle brackets will opt in to this but they aren't available yet.. Aside from getting the functionality sooner, would there be any other benefit in having these options available in curlys?

If this isn't currently possible and seems like something reasonable I may make an attempt at a PR..? Turning off attr proxying doesn't seem too difficult but will need to look into `readonly` by default rather than `mut` ...

ohnnyj:
Tried adding a comment to https://github.com/emberjs/ember.js/commit/9e5f072734cef1b71b5682dc7566f515e47f9b8a but no response so thought I would repost here, apologies if this is not the right forum.

Trying to use this new implementation but running into an issue (using https://github.com/emberjs/ember.js/commit/04dd84c479c76953e70cc3916bdfa5e65013b633).

I have a component template:

```html
<my-component>
  <div class="container"></div>
</my-component>
```

and use it in a route/top level template:

```html
<my-component model={{model}} />
```

From the description mentioned in the above commit I would expect that I would end up with a <my-component /> tag in the html but instead I get the error mentioned in https://github.com/emberjs/ember.js/commit/9e5f072734cef1b71b5682dc7566f515e47f9b8a.

If I remove the `<my-component>` from my component template no error but not the markup I would like. Maybe I am misunderstanding the correct way to do this.

And, BTW, this is probably an invalid usage pattern but changing the <my-component> to {{my-component}} in the component template results in a stack overflow.

chadhietala:
I believe there should be a build time warning or failure if you are using mutative APIs on attrs in angle bracket components. In React props can be mutated as well but their is tooling around this suggests that you should not mutate props.

The example below shows mutating both arrays and objects. Arrays are clearly worst that mutating objects as object properties are not propagated. It becomes hard to reason if the attrs can be mutated as they are passed along, we should do our utmost to shorten the conceptual gap between the static program and the dynamic process :wink:.

http://emberjs.jsbin.com/xahidifube

https://github.com/emberjs/ember.js/issues/11757

tomdale:
@chadhietala Setting up property descriptors on `attrs` would be slow but is probably worth it in development mode.

stefanpenner:
> @chadhietala Setting up property descriptors on attrs would be slow but is probably worth it in development mode.

in theory we can make attrs the instance of a class with those descriptors on the proto. So that would only be expensive the first time the class is derived. I suspect in practice this will be few, it would not longer be per snapshot or per component instance.

tomdale:
@stefanpenner The problem is that `attrs` aren't known on a per-component basis; they're known per-invocation. So e.g. I may have a component `{{my-component}}` that sometimes I invoke as `{{my-component foo="bar"}}` and sometimes as `{{my-component bar="baz"}}`. The `{{component}}` helper introduces even more complexity here.

Your proposal would require us to cache the class with descriptors on a per-invocation basis, which is not impossible but we don't have any infrastructure for it currently, as best I can tell. Maybe our friends on the Chrome team will hurry up and implement `Proxy` and then these concerns would be moot, anyway. ;)

stefanpenner:
> @stefanpenner The problem is that attrs aren't known on a per-component basis; they're known per-invocation.

yup, but this is statically known at HTMLBars build time.

chadhietala:
Like I mentioned, I believe tooling can be the happy path.  In general I don't really believe in the notion of "in dev it's slower".  I would rather have consistent performance characteristics regardless of environment.

mitchlloyd:
Very minor cocern here but with ye olde components, passing in an attr was an effective way to write tests for components that use services. 

    this.render("{{my-component session=sessionStub}}");

Any thoughts on whether angle bracket components mean the end of this pattern? This may be another reason that we need a well-paved path for stubbing services.

rwjblue:
@mitchlloyd - I think https://github.com/switchfly/ember-test-helpers/pull/96 is a much better pattern for this (and works with ember-qunit@0.4.10+).

workmanw:
While working to upgrade our app to 2.0 and plan for some future updates, one question came up relating to `this.attrs`, how will default values be configured? Right now it's very easy to provide sensible defaults for components by declaring them on the component class, but with `this.attrs` will that be possible?

As a use case, we often provide default values labels for many components, but allow them to be overridden.

I'm not even sure if this is the right place to ask this since `this.attrs` isn't exactly component unification. I just wanted to be sure to raise the concern of default values. If that concern belongs on another RFC or as an issue I'd be happy to raise it their.

mitchlloyd:
@rwjblue Thanks for the heads up!

mmun:
@workmanw You can just do

```js
this.foo = 'foo' in this.attrs ? this.attrs.foo : default
```

in whatever hook makes sense to you.

workmanw:
@mmun Okay yea, I was also thinking I could build a computed macro for this use case too. Something like:

```javascript
function attrsDefault(key, defaultValue) {
   return Ember.computed('attrs.'+key, function() {
     return (key in this.attrs) ? this.attrs[key] : defaultValue;
   });
}

Ember.Component.extend({
 foo: attrsDefault('foo', 'bar')
});
```

But I wasn't sure if this would be unadvised or potentially result in a naming conflict for a period of time.

mmun:
@workmanw That's definitely something I'm hoping someone experiments with in an addon so we can gather consensus on how useful it is.

tim-evans:
How would angle bracket components work with inverses?

For example, I'd write a `{{toggle-button}}` with an `{{else}}` to separate the enabled and disabled state. How would I write this with Ember.GlimmerComponent?

rwjblue:
The conversation in this RFC has been great! I think at this point this RFC doesn't reflect current lines of thinking, and either a rewrite or a new RFC will be needed when we are ready to start professing forward here. 

I propose that we close this PR...

rwjblue:
Thanks for everyone's hard work here. We have made a ton of progress, and are in the middle of integrating the updated glimmer engine that is required to start working on angle bracket components again. 

I'm going to close this for now while we continue pushing forward on the known path. We should reopen (or entertain a new RFC) when ready to work on angle bracket components again.

PR #10, opened 2014-10-25
RFC: Engines

tomdale:
[RFC](https://github.com/tomdale/rfcs/blob/master/active/0000-engines.md)

PR #64, opened 2015-06-12
Contextual components RFC

mmun:
Written together with @mixonic.

[Rendered view](https://github.com/emberjs/rfcs/blob/contextual-component-lookup/text/0000-contextual-component-lookup.md)

### DIFF
--- a/text/0064-contextual-component-lookup.md
+++ b/text/0064-contextual-component-lookup.md
@@ -1,0 +1,233 @@
+- Start Date: 2015-06-12
+- RFC PR: https://github.com/emberjs/rfcs/pull/64
+- Ember Issue: (leave this empty)
+
+# Summary
+
+The goal of this RFC is to allow for better component composition and the
+usage of components for domain specific languages.
+
+Ember components can be invoked three ways:
+
+* `{{a-component`
+* `{{component someBoundComponentName`
+* `<a-component` (coming soon!)
+
+In all these cases, attrs passed to the component must be set at the place of
+invocation. Only the `{{component someBoundComponentName` syntax allows for the name
+of the component invoked to be decided elsewhere.
+
+All component names are resovled to components through one global resolution
+path.
+
+To improve composition, four changes are proposed:
+
+* The `(component` helper will be introduced to close over component attrs in
+  a yielding context.
+* The `{{component` helper will accept an argument of the object created by
+  `(component` for invocation (as it invokes strings today).
+* Property lookups with a value containing a dot will be considered for
+  rendering as components. `{{form.input}}` would be considered, for instance.
+  Helper invocations with a dot will also be treated like a component if the
+  key has a value of a component, for instance `{{form.input value=baz}}`.
+* A `(hash` helper will be introduced.
+
+# Motivation
+
+When building a complex UI from several components, it can be difficult to
+share data without breaking encapsulation. For example this template:
+
+```hbs
+{{#great-toolbar role=user.role}}
+  {{great-button role=user.role}}
+{{/great-toolbar}}
+```
+
+Causes the user to pass the `role` data twice for what are obviously related
+components. A component can yield itself down:
+
+```hbs
+{{! app/components/great-toolbar/template.hbs }}
+{{yield this}}
+```
+
+```hbs
+{{#great-toolbar role=user.role as |toolbar|}}
+  {{great-button toolbar=toolbar}}
+{{/great-toolbar}}
+```
+
+And `great-button` can have knowledge about properties on `great-toolbar`, but
+this break the isolation of components. Additionally the calling syntax is not
+much better, `toolbar` must still be passed to each downstream component.
+
+Often `nearestOfType` is used as a workaround for these limitations. This API
+is poorly performing, and still results in the downstream child accessing the
+parent component properties directly.
+
+Consequently there is a demand by several addons for improvement. Our goal
+is a syntax similar to DSLs in Ruby:
+
+```hbs
+{{#great-toolbar role=user.role as |toolbar|}}
+  {{toolbar.button}}
+  {{toolbar.button orWith=additionalProperties}}
+{{/great-toolbar
+```
+
+As laid out in this proposal, the `great-toolbar` implementation would look
+like:
+
+```hbs
+{{! app/components/great-toolbar/template.hbs }}
+{{yield (hash
+  button=(component 'great-button' role=user.role)
+)}}
+```
+
+# Detailed design
+
+### The `(component` helper and `{{component` helper
+
+Much like `(action` creates a closure, it is proposed that the `(component`
+helper create something similar. For example with actions:
+
+```hbs
+{{#with (action "save" model) as |save|}}
+  <button {{action save}}>Save</button>
+{{/with}}
+```
+
+The returned value of the `(action` nested helper (a function) closes over the
+action being called (`actions.save` on the context and the `model` property).
+The `{{action` helper can accept this resulting value and invoke the action
+when the user clicks.
+
+The `(component` helper will close over a component name. The
+`{{component` helper will be modified to accept this resulting value and invoke
+the component:
+
+```hbs
+{{#with (component "user-profile") as |uiPane|}}
+  {{component uiPane}}
+{{/with}}
+```
+
+Additionally, a bound value may be passed to the `(component` helper. For
+example `(component someComponentName)`.
+
+Attrs for the final component can also be closed over. Used with yield, this
+allows for the creation of components that have attrs from other scopes. For
+example:
+
+```hbs
+{{! app/components/user-profile.hbs }}
+{{yield (component "user-profile" user=user.name age=user.age)}}
+```
+
+```hbs
+{{#user-profile user=model as |profile|}}
+  {{component profile}}
+{{/user-profile}}
+```
+
+Of course attrs can also be passed at invocation. They smash any conflicting
+attrs that were closed over. For example `{{component profile age=lyingUser.age}}`
+
+Passing the resulting value from `(component` into JavaScript is permitted,
+however that object has no public properties or methods. Its only use would
+be to set it on state and reference it in template somewhere.
+
+### Hash helper
+
+Unlike values, components are likely to have specific names that are semantically
+relevent. When yielded to a new scope, allowing the user to change the name
+of the component's variable would quickly lead to confusing addon documentation.
+For example:
+
+```hbs
+{{#with (component "user-profile") as |dropDatabaseUI|}}
+  {{component dropDatabaseUI}}
+{{/with}}
+```
+
+The simplest way to enforce specific names is to make building hashes
+of components (or anything) easy. For example:
+
+```hbs
+{{#with (hash profile=(component "user-profile")) as |userComponents|}}
+  {{component userComponents.profile}}
+{{/with}}
+```
+
+The `(hash` helper is a generic builder of objects, given hash arguments. It
+would also be useful in the same manner for actions:
+
+```hbs
+{{#with (hash save=(action "save" model)) as |userActions|}}
+  <button {{action userActions.save}}>Save</button>
+{{/with}}
+```
+
+### Component helper shorthand
+
+To complete building a viable DSL, `.` invocation for `{{` components will be
+introduced. For example this `{{component` invocation:
+
+```hbs
+{{#with (hash profile=(component "user-profile")) as |userComponents|}}
+  {{component userComponents.profile}}
+{{/with}}
+```
+
+Could be converted to drop the explicit `component` helper call.
+
+```hbs
+{{#with (hash profile=(component "user-profile")) as |userComponents|}}
+  {{userComponents.profile}}
+{{/with}}
+```
+
+A component can be invoked like this only when it was created by the
+`(component` nested helper form. For example unlike with the `{{component`
+helper, a string is not acceptable.
+
+To be a valid invocation, one of two criteria must be met:
+
+* The component can be called as a path. For example `{{form.input}}` or `{{this.input}}`
+* The component can be called as a helper. For example `{{form.input value=baz}}` or `{{this.input value=baz}}`
+
+And of course a `.` must be present in the path.
+
+# Drawbacks
+
+This proposal encourages aggressive use of the `(` nested helper syntax.
+Encouraging this has been slightly controversial.
+
+No solution for angle components is presented here. The syntax for `.`
+notation in angle components is coupled to a decision on the syntax for
+bound, dynamic angle component invocation (a `{{component` helper for angle
+components basically).
+
+`(component 'some-component'` may be too verbose. It may make sense to simply
+allow `(some-component`.
+
+Other proposals have leaned more heavy on extending factories in JavaScript
+then passing an object created in that space. Some arguments against this:
+
+* Getting the container correct is tricky. Who sets it when?
+* Properties on the classes would not be naturally bound, as they are in this proposal.
+* As soon as you start setting properties, you likely want a `mut` helper,
+  `action` helper, etc, in JavaScript space.
+* Keeping the component lookup in the template layer allows us to take advantage
+  of changes to lookup semantics later, such as local lookup in the pods
+  proposal.
+
+# Alternatives
+
+All pain, no gain. Addons really want this.
+
+# Unresolved questions
+
+There has been discussion of if a similar mechanism should be available for
+helpers.


mmun:
cc @stefanpenner @rwjblue

text/0000-contextual-component-lookup.md
```diff
@@ -0,0 +1,119 @@
...
+This problem seems unavoidable as long as {{foo}} is ambiguous between a property
+lookup and a helper call.
+
+# Alternatives
+
+I don't know of any alternative strategies that support contextualized components.
+
+# Unresolved questions
+
+- Should this also be available for helpers?
```
stefanpenner: I would love this, but seems to be slightly more involved. Especially when taking sub-expressions into account.

heycarsten:
:muscle: _STRONG MOVE_ :muscle:

text/0000-contextual-component-lookup.md
stefanpenner: * what do HTMLTags look like ?

text/0000-contextual-component-lookup.md
```diff
@@ -0,0 +1,119 @@
...
+{{this.component}}
+{{this.component name=name}}
+<this.component>
+<this.component name="{{name}}">
+```
+
+# Drawbacks
+
+The only drawback I see is the requirement for the dot separator. The reason for
+enforcing the dot separator is to avoid accidental fallback to global components.
```
stefanpenner: IMHO this is a +1, as it also allows "private/local" components.

stefanpenner: We do now "claim" the dot syntax, whereas before it was not reserved.

text/0000-contextual-component-lookup.md
```diff
@@ -0,0 +1,119 @@
...
+and
+
+```js
+// components/form-for.js
+
+import Ember from 'ember';
+import curry from 'factory-utils';
+
+export default Ember.Component.extend({
+  helpers: Ember.computed(function() {
```
stefanpenner: helpers may be a confusion term here :P

mmun: oops, good call.

text/0000-contextual-component-lookup.md
mmun: I listed below:

```
<this.component>
<this.component name="{{name}}">
```

This is semi-future hostile but I don't know of any web proposals that use dots in tag names.

mmun: I added some text in drawbacks addressing HTML tags.

stefanpenner: I was not clear, what is the outputted HTML.

mmun: Hmm, yes good point! That is problematic. The factories would need to know about their name/tag name.

text/0000-contextual-component-lookup.md
```diff
@@ -0,0 +1,129 @@
...
+
+```js
+// components/form-for.js
+
+import Ember from 'ember';
+import { curry } from 'factory-utils';
+
+export default Ember.Component.extend({
+  controls: Ember.computed(function() {
+    var inputFactory = this.container.lookupFactory('container:form-for-input');
```
rwjblue: `lookupFactory('component:form-for-input')`

mmun:
I need to address what happens when the property (e.g. `this.component`) changes. This should simply trigger a revalidation (and subsequent rerender) like the existing dynamic component helper does.

stefanpenner:
> STRONG MOVE  I wish components could just use angle-bracket syntax.

they can/will

stefanpenner:
> I need to address what happens when the property (e.g. this.component) changes. This should simply trigger a revalidation (and subsequent rerender) like the existing dynamic component helper does.

good catch, this must also work in the re-yielding case.

text/0000-contextual-component-lookup.md
stefanpenner: I believe this is actually just part of a larger problem. `<` components, may or may not reflect there visual tagName.

text/0000-contextual-component-lookup.md
DingoEatingFuzz: Isn't `this.container.lookupFactory` private?

text/0000-contextual-component-lookup.md
```diff
@@ -0,0 +1,129 @@
...
+
+## `form-for` Example
+
+This section describes an example of how you'd use this technique to
+yield contextualized components in a form building component.
+
+Let's say we want our users to write
+
+```hbs
+{{#form-for post as |f|}}
```
stefanpenner: should we transition this to the `<` syntax ?

stefanpenner:
we should be sure the following example works:

outside
```hbs
  <my-select data=model.entries as |s entry|>
    <s.option value={{entry.value}}></my-option>
  </my-select>
```

inside
```hbs
  {{#each entries as |entry|}}
     {{yield formHelpers entry}}
  {{/each}}
```

MiguelMadero:
This is really nice, we have a few places where we're doing this, while nearestOfType solves the problem nicely, this will make it simpler and more intuitive going forward. 

In most cases where we use this, the child-components only make sense inside that parent-component. I'm wondering if we need to consider a way to "hide" the components from being used directly so they're only available through lookup properties. 

Following your example, I could use `<form-for-input>` (or `{{form-for-input}}`) directly, without being inside of a `form-for`. That means that instance won't have a reference to a form. In the places I used nearestOfType, I simply throw an exception at runtime if the child-component is not inside the expected parent. Since we're rethinking this approach it might make sense to limit how components are resolved so we can only get them through the parent.One possible solution could be to nest them in a folder like `form-for/input`. We could still specify the full path to the container of the injector.

stefanpenner:
> I'm wondering if we need to consider a way to "hide" the components from being used directly so they're only available through lookup properties.

just store them some-where non-resolvable

mmun:
@MiguelMadero You could just throw an exception in `init` of `form-for-input` if `this.form` isn't set, but I agree with @stefanpenner. For instance, `-form-for-input` (leading hyphen).

stefanpenner:
> @MiguelMadero You could just throw an exception in init of form-for-input if this.form isn't set, but I agree with @stefanpenner. For instance, -form-for-input (leading hyphen).

we can also come up with a injectable strategy that is non addressble. This allows for type injections etc.

MiguelMadero:
Nice. Thanks



—

juggy:
Can't you use the html namespace? 
```
 <my-select data=model.entries as |s entry|>
    <s:my-option value={{entry.value}}></my-option>
  </my-select>
```

stefanpenner:
> <my-select data=model.entries as |s entry|>
>     <s:my-option value={{entry.value}}></my-option>
>   </my-select>

may have issues with SVG?

juggy:
I was reading https://github.com/emberjs/rfcs/pull/10#discussion_r19474370 and it seems this RFC is not the only one where the syntax will be a struggle.

It should be unified that is for sure. Namespaces would be a good solution for namespaced components than this use case.

MiguelMadero:
Well this isnt a namespace, it is more of a property accesor. I think thats why the dot makes more sense. 



—

stefanpenner:
> It should be unified that is for sure. Namespaces would be a good solution for namespaced components than this use case.

i don't believe this is actually a namespace, as it is contextual. There is no parallel in the DOM.

juggy:
The parallel in the DOM would be table/tr/td/th.
tr or td by itself does not mean anything or their own, within table they come to life. 

Might not be what we are looking for but it makes sense. Might be easier to resolve the components from the parent component first then global, without the "." 

That does not solve all use case (what if you have a component inbetween the parent and its child for example).

stefanpenner:
> That does not solve all use case

it is also brittle to change.

TR/TD isn't really the same either, they are contextual, but actually they do reside on the global namespace of all DOMElements.

mmun:
There is a concern that users will write embedded subcomponents:

```
  childComponent: Ember.Compontent.extend({
    // ...
  })
});
```

This is not a pattern we want to promote, particularly because this use case will likely be covered better with pods. The resolution is to assert that the factory was obtained via injection or lookupFactory (possibly using a private branding).

stefanpenner:
> (possibly using a private branding).

:+1:

mixonic:
I've several specific points of feedback here, which I'd tried to form into a cohesive system (with some feedback from @mmun). If he is interested and people agree in the general direction, I'm happy to make a PR to this RFC.

### Motivation

The motivation is to allow the yielded of some components into the scope of another template. The example that seems simple enough to understand is:

```
<my-form on-submit={{action 'save' model}} as |form|>
  {{form.input value=model.name}}
</my-form>
```

100% on-board. However there are a few aspects I strongly dislike about the current proposal:

* The proposal for looking up a component class is `this.container.lookupFactory('component:form-for-input')`. As the container itself is still not public (though close), this seems extremely poor. Explaining what is happening to a new developer seems unrealistic.
* Templates in Ember (thus far) provide a fairly good barrier between the setup/teardown of individual components. By moving the work of building hash arguments and `curry` of the factory into JS, we lose many utilities already provided for in Ember templates. For example:
  * We would shortly want a `mut` helper in JS to pass mutable values
  * We would shortly want an `action` helper in JS to curry scope for actions
  * Making the `controls` property cause a rerender when a value being `curry`'d over changes seems painful with this proposal (would love to see it addressed if my ideas here are rejected).
* To support pods, core has considered a concept called "local lookup" that means component lookup first goes to the current directory of a POD component, then to the global scope. How would JS-space have access to this?
* In general, user-space helpers (often already used for passing arguments to a component) would not be available to these component definitions. (for example, how does `{{my-input placeholder=(format-currency cents)}}` translate to JS-space so it can be curried?)

### Proposal for an API in Templates

Instead, I propose that we only need to make a few changes to support this use-case in templates.

**A `hash` helper should be added to Ember**

This would allow creation of objects:

```hbs
{{yield (hash name="bob") (hash name="nancy")}}
{{my-comp options=(hash title="Mr")}}
```

**Components used in subexpressions should curry arguments**

A component name used as a subexpression or in attribute space should curry to a new class. For example:

```hbs
<some-comp factory={{other-component}}>
{{yield (hash
  other=(other-component)
)}}
{{yield (hash
  input=(my-input parent=this)
  checkbox=(some-checkbox placeholder="Hi")
)}}
```

When used in this way the class would simply be extended, so:

```hbs
<some-comp factory={{other-component name="bob"}}>
```

Could be used with a JS class of:

```js
// app/components/some-comp.js
  didReceiveAttrs() {
    var componentClass = this.get('other-component');
    Ember.Component.detect(componentClass); // => true
    var component = componentClass.create({container: this.container});
    Ember.Component.detectInstance(componentClass); // => true
    component.get('name'); // => 'bob'
  }
});
```

Basically, calling a component in subexpression would curry. These would be bound like any other template helper, and would remove the need for an API in JavaScript.

**The component helper should also curry**

This is a natural extension of component lookup currying, and would allow for complex flexibility.

```
{{yield (hash
  input=(component dynamicInputOption **inputHash parent=this)
)}}
```

Here I've also used the proposed [splat operator](https://github.com/wycats/handlebars.js/issues/1050) to demonstrate a completely dynamic component yield.

### Invocation API Feedback

I think this RFC should consider three types of invocation, but not address solutions for them all.

**traditional** component lookup is when we are referencing a global component or one with the new local lookup semantics. For example:

```hbs
{{some-component}}
<some-component />
```

**passed** components are representative of a component class, and are passed like variables.

```hbs
<my-form as |form|>
  {{form.input}} {{! as stated in this RFC, b/c this path has a . it is possibly a component}}
</my-form>
```

**dynamic** component lookup uses a binding to a string. *I propose it should also accept a component class*:

```hbs
{{component "dashlessComponentPath"}}
{{component somePath}}
<my-form as |form|>
  {{component form.input}}
</my-form>
{{component (another-component)}}
```

For this RFC, **I propose we completely ignore the question of how to use a passed component with angle bracket invocations**, and instead defer whatever solution we come up with for **dynamic angle components**. Some examples of what that might look like:

```hbs
{{! NONE OF THIS SHOULD BE PROPOSED IN THIS RFC }}
<{{somePath}} />
<my-form as |form|>
  <{{form.input}}></{{form.input}}>
  {{! an alternative }}
  <ember-component is={{form.input}}></ember-component>
</my-form>
```

The use cases described in this RFC should instead inform the discussion about what dynamic component lookup with angles looks like. I imagine paths with dots like `<this.foo` would be a strong contender, however approving that syntax without considering other dynamic lookup strategies is short-sighted.

mmun:
Re: Dynamic angle bracket components (DABC). I agree that this RFC should not discuss that topic, but this RFC  ought to be blocked until we have reached consensus that there is a viable solution to DABC if this RFC is merged. I'll make a separate issue for that.

@mixonic I prefer your solution because its much closer to how components are used today. I would like @stefanpenner's approval. For posterity, the example in my original RFC

```hbs
{{! templates/components/form-for.js }}

{{yield controls}}
```

and

```js
// components/form-for.js

import Ember from 'ember';
import { curry } from 'factory-utils';

  controls: Ember.computed(function() {
    var inputFactory = this.container.lookupFactory('component:form-for-input');
    var submitFactory = this.container.lookupFactory('component:form-for-submit');

    return {
      input: curry(inputFactory, { form: this }),
      submit: curry(submitFactory, { form: this })
    };
  })
});
```

could be replaced with

```hbs
{{! templates/components/form-for.js }}

{{yield (hash
  input=(form-for-input form=this)
  submit=(form-for-submit form=this)
)}}
```

which is significantly less code and one less file.

mmun:
@mixonic I assume your `didReceiveAttrs` example is just a "thought experiment", because no one should be accessing the factory from attrs directly :P It would be a bit awkward with streams and the view hierarchy - you could stash the factory on a service and invoke it in some other subtree. I don't want to give people the opportunity to do crazy things.

mixonic:
> For instance, you could stash the factory on a service and then invoke the component it in some other subtree.

Upon further discussion, it sounds like the optimal structure would *not* be a raw extended factory as proposed above. Instead, we should create what I will call a `ComponentCell`. This would be responsible for keeping track of the original factory (the constructor) and a closure that created the component with the proper attrs.

This is really an implementation detail. The impact on my above suggestions is that **the factory would not be available in JS-space**. The cell itself would be passed instead, and have basically zero public API surface.

mixonic:
See https://github.com/emberjs/rfcs/pull/85 for some small additional commentary on the new proposal

mitchlloyd:
This all looks great and I'm going to try using a hash helper today. However, the fact that the "shorthand" syntax will require a period is a little disappointing. It seems a shame that the following would not work:

```handlebars
{{#with (component "user-profile") as |userProfile|}}
  <p>Where to put the profile?</p>
  <p>How about right here: {{userProfile}}</p>
{{/with}}
```

I'm not expert on these internals, but it _feels_ as if we should be able to to recognize that `userProfile` is a component at runtime and render it correctly. Conceptually it feels correct that I've bundled up this component with the `component` helper and now I'm going to stick it in a template somewhere. I'm assuming this is not trivial since you guys added the dot requirement, but I would love to know more.

mmun:
@mitchlloyd You're right. There's no technical obstructions. In fact, the block params case _should_ be supported. Here's an example that _is_ problematic:

parent-component.hbs: `{{child-component user-profile=(component "user-profile")}}`
child-component.hbs: `{{user-profile}}`

It's not clear here where `user-profile` is being resolved to. If there's a bug, you need to look in several possible places. In this case, we've compromised that `{{this.user-profile}}` is good enough to break this ambiguity.

mixonic:
@mmun ah right we should add the use-case for `this.` to the RFC

mitchlloyd:
@mmun, @mixonic Looks like you've got this well thought out. That all sounds great to me!

MiguelMadero:
Is `this.` only needed when we have to dissambiguate or for every property accessor on the current context? It's rare and unconventional to have a property called `user-profile` a property would be `userProfile`. Also for single word properties wouldn't clash either since components require dashes. 

Either way, I think it's important to be explicit in the use of `this.` it could also be more efficient since we can save the property lookup, but I'm not sure that the argument for it is ambiguity.

mixonic:
@MiguelMadero I'm not clear what point you are making.

If the common name for a component property is `userProfile`, then the `this.` would be needed to treat that property as a component invocation. There is no `-` in `userProfile` that would otherwise make it clear there may be a component invocation happening.

This is important to get right so let me know if I missed something. Additionally, it is also important, but not a reason to disregard your concerns, to note that `userProfile` seems very rare to me. In most cases you would use a `hash` with a block param, or with angle you must access the passed-in attr as `attrs.userProfile`. The `this.` case will be fairly rare in the long run.

MiguelMadero:
My points it that there's no need to disambiguate if we follow the conventions. A component needs an underscore and a properties should not have them. So, there's clash unless we deviate from the convention for property names which isn't enforced. 

I think it's worth always using `this.` not because it's ambiguous with the normal component resolution, but to be explicit and even consistent, we could say every contextual lookup needs a dot and it's always a property path.

mmun:
@MiguelMadero I agree with what you're saying. It's worth considering a hyphen to be the disambiguator. There are other concerns too - if you set a property called `outlet` would it shadow the outlet helper?

MiguelMadero:
Oh, yes, I forgot about helpers. It makes sense to disambiguate :)

stevesims:
Really liking the prospect of a component helper, to the point that I'm actively disappointed it doesn't work in Ember already.  :-)

mixonic:
Core discussed this last Friday (21st of August). The consensus is that the API described here has significant intersection with work on other parts of the framework that are on-going, for example angle components. This means the final implementation may diverge from what is proposed here.

However, after the rather extensive conversations here and internally the direction is generally approved.

RFCs are intended to spark discussion, build consensus, and show intent. I'm merging this RFC tonight, which means this proposal is ready for implementation behind a feature flag and real-world experimentation.

PR #65, opened 2015-06-14
Deprecation warning handlers

pangratz:
[Rendered](https://github.com/pangratz/rfcs/blob/publicize-log-levels/text/0000-deprecation-warning-handlers.md)

### DIFF
--- a/text/0065-deprecation-warning-handlers.md
+++ b/text/0065-deprecation-warning-handlers.md
@@ -1,0 +1,176 @@
+- Start Date: 2015-06-30
+- RFC PR: https://github.com/emberjs/rfcs/pull/65
+- Ember Issue: (leave this empty)
+
+# Summary
+
+Deprecations and warnings in Ember.js should have configurable runtime handlers.
+This allows default behavior (logging, raise when `RAISE_ON_DEPRECATION` is true)
+to be overridden by an enviornment (Ember's tests), addon, or other tool
+(like the Ember Inspector).
+
+Ember-Data and the Ember Inspector have both requested a public
+API for changing how deprecation and warning messages are handled. The requirements
+for these and other requests are complex enough that deferring the message
+behavior into a runtime hook is the suggested path.
+
+# Motivation
+
+`Ember.deprecate` and `Ember.warn` usually log messages. With `ENV.RAISE_ON_DEPRECATION`
+all deprecations will throw an exception. In some scenarios, this
+is less than ideal:
+
+* Ember itself needs a way to silence some deprecations before their usage
+  is completely removed from tests. For example, many view APIs in Ember 1.13.
+* The Ember inspector desires to raise on specific deprecations, or silence
+  specific deprecations.
+* Ember-Data also desires to silence some deprecations in tests
+
+In [PR #1141](https://github.com/emberjs/ember.js/pull/11419)
+a private log level API has been introduced, which allows finer grained control
+if specific deprecations should be logged, throwing an error or be silenced
+completely. For example:
+
+```js
+Ember.Debug._addDeprecationLevel('my-feature', Ember.Debug._deprecationLevels.LOG);
+// ...
+Ember.deprecate("x is deprecated, use Y instead", false, { id: 'my-feature' });
+```
+
+Initially a public version of this API was discussed, but it quickly became
+clear that a runtime hook provided more flexibility without incurring the
+cost of a complex log-level API.
+
+Note that "runtime" refers to Ember itself. A custom handler could be injected
+into Ember-CLI's template compilation code. "runtime" in this context still
+refers to handling deprecations raised during compilation.
+
+# Detailed design
+
+A handler for deprecations can be registered. This handler will be called
+with relevent information about a deprecation, including guarantees about
+the presence of these items:
+
+* The deprecation message
+* The version number where this deprecation (and feature) will be removed
+* The "id" of this deprecation, a stable identifier independent of the message
+
+Additionally, an application instance may be passed with the options. An example
+handler would look like:
+
+```js
+import { registerHandler } from "ember-debug/deprecations";
+
+registerHandler(function deprecationHandler(message, options) {
+  // * message is the deprecation message
+  // * options.until is the version this deprecation will be removed at
+  // * options.id is the canonical id for this deprecation
+  if (options.until === "2.4.0") {
+    throw new Error(message);
+  } else {
+    console.log(message);
+  }
+});
+```
+
+Warnings are similar, but will not recieve an `until` value:
+
+```js
+import { registerHandler } from "ember-debug/warnings";
+
+registerHandler(function warningHandler(message, options) {
+  // * message is the warning message
+  // * options.id is the canonical id for this warning
+  if (options.id !== 'view.rerender-on-set') {
+    console.log(message);
+  }
+});
+```
+
+##### chained handlers
+
+Since several handlers may be registered, a method of deferring to a previously
+registered handler must be allowed. A third option is passed to handlers, the
+function `next` which represents the previously registered handler.
+
+For example:
+
+```js
+import { registerHandler } from "ember-debug/deprecations";
+
+registerHandler(function firstDeprecationHandler(message, options, next) {
+  console.warn(message);
+});
+
+registerHandler(function secondDeprecationHandler(message, options, next) {
+  if (options.until === "2.4.0") {
+    throw new Error(message);
+  }
+  next(...arguments);
+});
+```
+
+The first registed handler will receive Ember's default behavior as `next`.
+
+##### new assertions for deprecate and warn
+
+Ember's APIs for deprecation and warning do not currently require any information
+beyond a message. It is proposed that deprecations be **required** to pass
+the following information:
+
+* Message
+* Test
+* Canonical id (with a format of `package-name.some-id`)
+* Release when this deprecation will be stripped
+
+For example:
+
+```
+import Ember from "ember";
+
+Ember.deprecate("Some message", false, {
+  id: 'ember-routing.query-params',
+  until: '3.0.0'
+});
+```
+
+If this information is not present and assertion will be made.
+
+Warnings likewise will be required to pass a canonical id:
+
+```
+import Ember from "ember";
+
+Ember.warn("Some warning", {id: 'ember-debug.something'});
+```
+
+##### default handlers
+
+The default handler for deprecation should be quite simple, and mirrors current
+behavior:
+
+```js
+function defaultDeprecationHandler(message, options) {
+  if (Ember.ENV.RAISE_ON_DEPRECATION) {
+   throw new Error(format(message, options));
+  } else {
+   console.log(format(message, options));
+  }
+}
+```
+
+The default handler for warnings would be simple `console.log`.
+
+# Drawbacks
+
+By not providing a robust log-level API, we are punting complexity to the
+consumer of this API. For a low-level tooling API such as this one, it seems
+and appropriate tradeoff.
+
+# Alternatives
+
+Each app can stub out `deprecate` and `warn`.
+
+# Unresolved questions
+
+`RAISE_ON_DEPRECATION` could be considered deprecated with this new API.


mixonic:
This should propose module names for this API

bantic:
I'd like to propose using dots to namespace deprecation ids, and allowing one to set levels for groups of related deprecations by specifying a substring of the path in `setLevel`. Similar to the way [Ember Instrumentation](http://emberjs.com/api/classes/Ember.Instrumentation.html) works.

So for example:
```
Ember.Debug.setLevel('view', Ember.Debug.logLevels.SILENCE);

Ember.deprecate('this is silenced', false, {id: 'view.helper.select'});
Ember.deprecate('this is silenced, too', false, {id: 'view.keyword'});
```

krisselden:
Why levels? Shouldn't specific warnings and deprecations be silenceable for during a section of code?

eccegordo:
As I mentioned over in 
https://github.com/emberjs/rfcs/pull/56#issuecomment-113049597

I think "categories" of deprecations would be nice to have. That way I could selectively disable whole classes of deprecations as I am working through the upgrade process. Especially if I need to focus on time sensitive deprecations and defer more long range deprecations for later.

My mental model looks like this, but I am sure there there are others.

I would expect that every deprecation warning in the code base would be registered with at least 1 or more categories.

```
PRIVATE    --> Your code uses private API and is subject to change, don't say we didn't warn you
URGENT     --> Your code directly uses API that is being removed next release
IMPORTANT  --> Framework change, or your code relies on API behavior that will change eventually in long term support release
SUGGESTION --> Your code triggers legacy feature or is suboptimal for performance, no change immediately required but you might want to look into it.
```

mixonic:
We have:

* warnings, deprecations, and assertions
* raise, log, and silence

I am not crazy about adding four more levels of configuration and maintenance. I would like to consider how we can flag either a) the release a given thing is expected to be dropped in or b) the public/private nature of the API. In my mind these are coupled, since presuming we go with extended-support releases private API deprecations will be dropped on a minor release and public API deprecations only on majors. Having a way to include that expected removal release would be nice (though I'm sure sometimes the planned release would not pan out as expected).

Basically I'd rather couple things to the release number than bike shed on when something is `IMPORTANT` vs. `URGENT`. I'll try to have a proposal for how the logLevel managers might interact with this data included here.

eccegordo:
Number of categories is not so important I suppose. And if coupled to release version you could in theory even compute the relative urgency.

The high order bit for me personally is the distinction between deprecation and SUGGESTION.

A deprecation means some code or api is changing or being removed. Heads up! You could even be smart about with something like

> Hey by the way this deprecation, if you refactor you could remove this API entirely by applying the Svelte build. 

That way if there is just one or a few places in the code base where that particular feature is used they could opt to refactor away. It might even serve as a powerful tool to assess whether I want to use a particular addon.

The "suggestion" category is more nuanced. I would technically say it is not even a deprecation. But more like a warning when there is a particular code smell or known problematic code pattern.

Like:
> Hey this code might trigger a performance regression, or legacy code path. The API is not necessarily changing but you would be advised to address it

An example is the warning about mutating state within a `didInsertElement` hook and potential performance problems.

As an app developer I might not actually care about the performance in some cases. Because if it works it works. But I really do care if the feature goes away and I need to consider upgrade path.

pixelhandler:
:+1: I Iike this a lot :)

pangratz:
@mixonic I updated the RFC to use import statements...

---

I think the actual issue/problem is to silence deprecations which are not relevant/important at the moment; since *all* deprecation messages should be addressed eventually. Having a possibility to silence specific deprecation messages using the API proposed by @bantic, is one way of dealing with deprecations. This could be useful in the Ember-Inspector where there might be the possibility to hide messages from a certain namespace (`view.*` for example).

Deciding which deprecation messages are important at the moment is crucial. One way could be to deprecate stuff as done now via `deprecate("deprecation message", condition, { id: 'identifier' })`. When a new release is cut (say `2.1`), all calls to `deprecate` are replaced with:

    deprecatedSince("2.1", "deprecation message", condition, { id: 'identifier' })

This would allow us to calculate the urgency of the hit deprecation. Say if Ember.js `v2.4` is used and the method is deprecated since `2.1`, an error like **You hit a deprecated path which will be removed in the next minor release. Fix ASAP if you want to to upgrade**, whereas a deprecated method since `2.3` would only be logged in the Ember-Inspector...

mixonic:
@pangratz Right. I actually think the removal version is more pertinent though.

```
deprecate(
  "some long and annoying message that is usually on a line alone",
  condition,
  { id: 'view.some-thing', until: '3.0.0' }
);
```

Another valid `until` value would be `2.4.0`, for example. This does not explicitly address private vs. public deprecation flagging, but attempts to skirt the issue by leaning on version numbers.

It also seems to be agreed that there should be another level of logging for "urgent" deprecations, ones the break in the next release. I remain unsure this concept is practically useful (setting a flag that silenced all but urgent deprecations would seem to be an anti-pattern, setting up apps to continue to rely on a dead-end feature until the last minute), but I'm going to explore some options right now.

mixonic:
Ok, like most things as soon as I thought we were done it became clear we had designed the wrong thing. It sounds like there are ambitious enough plans for the inspector that it requires more of a hook. You call:

```js
deprecate(
  "some long and annoying message that is usually on a line alone",
  condition,
  { id: 'view.some-thing', until: '3.0.0' }
);
```

And Ember's test runner, or the inspector, or Ember-Data's test runner, or some addon @rwjblue cooks up can handle it like it chooses:

```js
import { registerDeprecationHandler } from "ember-debug";

registerDeprecationHandler((message, options, next) => {
  if (nextVersion(options.until)) {
    throw new Error(message);
  }
  next(...arguments);
});
```

The `next` function would allow a handler to defer to a previously registered handler (for example to get default logging behavior).

* The same should exist for `warn`. `warn` should enforce the presence of `id`.
* `deprecate` should enforce the presence of `id` and `until`.
* `ENV.RAISE_ON_DEPRECATION` will be honored by the default handler in Ember.

This solution has the benefit of giving addon and app authors *complete run-time control* over how a deprecation is handled. And without needing to understand an overly complex log-level API.

pixelhandler:
So I have about a handful of apps … when I upgrade to the current release we fix the deprecations. I tried an update from 1.11.3 to 1.12.1 today just to see if it was possible. Normally we don't merge in an updated version of Ember unless we actually fix the deprecations. This idea of silencing deprecations will help my team to upgrade then fix the deprecations.

Some context: 

In one app, on the first page following login, there are 999+ messages in the console. And the console is not even usable. That is why we can't live with deprecations warnings. Who does development without a console? I do when I update Ember till I fix the deprecations. 

Literally, the volume of warnings I get is prohibitive toward merging in a current release - unless we can budget enough time to fix the code that we've been warned about. 

The updates from 1.10 to 1.11 took a week or two. So I'm at the point were I'd have to plan 4 weeks of work then 2 weeks of upgrades. So for kicks I tried a update from 1.12.1 to 1.13.2 including the stable 1.13.4 ember data and boom the app is broke, error can't destroy a view. or something like that. 

I'll try the suggestion to redefine the methods as mentioned here http://discuss.emberjs.com/t/is-this-possible-to-turn-off-some-deprecations-warnings/8196/3 

Just saying, I understand the deprecation warnings are important and that I should fix them if I don't want to get left behind. But I can't have the warnings effectively disabling my console. Some how this looks appealing:

```
Ember.deprecate = function() { };
Ember.warn = function() { };
```

But seems silly to stick that in my application. I'll be very happy when I can set the log level and disable warnings that we can't afford to fix at the same time as upgrading or beta testing .

Here is some more dialog https://twitter.com/ebryn/status/610997797530136576 "upgrading is easy" well the deprecations make it both easy and super hard it depends on how much code is in your project.

rwjblue:
@pixelhandler - I have felt this pain as well, and that is **exactly** what this RFC is trying to help us solve.  Once we have the proper infrastructure in place, tooling will be able to selectively silence a blacklist or whitelist of "known" deprecations so that you can work through them (console clear) one at a time either via an Ember CLI addon (that I am working on) or via the Ember Inspector.

rwjblue:
@mixonic - I am :+1: on your latest proposal, this would allow maximum flexibility for tooling developers to fill the massive void that we have here at the moment.  Tools like the inspector, or an ember-cli addon that helps manage deprecations can make much of the pain that @pixelhandler mentioned away.

mixonic:
New RFC is in a PR pending review by @pangratz at https://github.com/pangratz/rfcs/pull/2. This PR should likely be renamed if/when it merges.

pangratz:
@mixonic I very much like the newly proposed handler API :clap: . Having the `until` information seems useful! But I would still add a `since` version as well, so potential deprecation handlers can sort the warnings by urgency, if that makes sense ...

rwjblue:
@pangratz - I do not believe that the `since` information is useful for determining urgency.  The reason a thing is urgent is not because it has been deprecated for a long time, but because it will be removed soon.

pangratz:
> ... is urgent is not because it has been deprecated for a long time, but because it will be removed soon.

Ah, THAT helped a lot understanding the the reasoning behind it! Thank you sir.

pixelhandler:
@rwjblue another thing that affects my regarding deprecations is that I get failures when I deploy using asset pipeline, precompiling fails for me (on 1.12.1)

```
Tasks: TOP => assets:precompile
rake aborted!
JSON::ParserError: 757: unexpected token at 'DEPRECATION: Using {{with}} without block syntax is deprecated. Please use standard block form (`{{#with foo as |bar|}}`) instead. See http://emberjs.com/deprecations/v1.x/#toc_code-as-code-sytnax-for-code-with-code for more details.

bundle exec rake assets:precompile ) exit status: 1
```

I wonder how the deprecations break that, weird.

stefanpenner:
Woah what is parsing stdout here?

rwjblue:
> I wonder how the deprecations break that, weird. So basically I can't merge an update to ember unless I fix all the deprecation warnings cause I can't precompile the assets for the apps. So can't deploy.

I'm not really sure what is going on with your precompilation process, but parsing stdout seems kinda bizarre.  Happy to chat with you about it in slack when you have time.

stefanpenner:
That seems like something an issue should be tracking. Maybe ember-rails or w/e the common library exists

pixelhandler:
@rwjblue rake assets:precompile fails for me locally as well. It will be easier to fix the deprecation warning than to debug asset pipeline. I wonder if the backtics in the warning have anything to do with it? oh well.

pixelhandler:
@stefanpenner @rwjblue other deprecations like `DEPRECATION: Using the same function as getter and setter is deprecated. See http://emberjs.com/deprecations/v1.x/#toc_computed-properties-with-a-shared-getter-and-setter for more details.` don't break the build (with precompile)

rwjblue:
@pixelhandler - I am sorry you are having issues with template precompilation, but those issues are not really related to this RFC or deprecations in general.  I would be more than happy to help you track things down in slack, but I'd like to keep this RFC on topic so we can fix the deprecation "spew" issue for folks...

pixelhandler:
@rwjblue Ok I'll say it like this. How about a SHUTTING OFF deprecations during template compilation too, please.

mixonic:
@pixelhandler Again, that would be implementable by a consumer of the API described in this RFC, such as Ember-CLI.

pixelhandler:
@mixonic what is the benefit for deprecations during compilation of templates? for me it breaks the build depending not the message, e.g. maybe having backtics or something like that. So we have a core lib with templates that need to be compiled and various apps that need to be compiled too I'm not sure that a consumer API will actually handle that for example the barber gem handles precompiling with asset pipeline (I think). So would need a way for that gem to use the template compiler to also consume this consumer API. I think the issues I have are relevant to the motivation no the need to change how deprecations are handled. @rwjblue asked to to comment on this RFC so I did, hope it helps.

rwjblue:
> for me it breaks the build depending on the deprecation message

This is a bug, plain and simple. I pinged you on a related issue in ember-rails (https://github.com/emberjs/ember-rails/issues/479), and have potentially tracked down where it is happening.

> what is the benefit for deprecations during compilation of templates

Because we can tell you everything that you are using that is deprecated, as opposed to doing it at runtime where you will only see deprecations for the sections of templates that you happen to test (many people do not have 100% template test coverage for example).  We can also give you the specific filename, line, and column of each deprecation which is not possible after precompilation (because the compiled templates do not retain source information).

The difference is basically `"You are using xyz, please update to 1234."` vs `"You are using xyz (app/templates/application @ L3:C15), please update to 1234."`.  Clearly the second one is **much** more useful, and is only possible at precompilation time.

>  I'm not sure that a consumer API will actually handle that for example the barber gem handles precompiling with asset pipeline (I think)

Barber can plug into the tooling API that we are setting up here just as easily as ember-cli.

> I think the issues I have are relevant to the motivation - the need to change how deprecations are handled.

Without a doubt, but the fact that it just blows up is a bug and should definitely be fixed.  This RFC is about making it possible to deal with deprecations in piecemeal fashion.  We definitely do not need an RFC for new public API to fix a straight up bug in template compilation.

pixelhandler:
@rwjblue @mixonic This is an RFC, sorry my comments include issues, from my perspective it was those issues that motivated me to comment here so the future would be brighter and maybe less gloomy than the present. So here is my comment... (or story if you will)

As a developer I want to configure the deprecation warnings and have the ability to set when the deprecations are fired, either at **runtime or compile time**, so that I have the flexibly to deploy my applications without chasing bugs resulting from the deprecation *warnings*.

I want the ambitions apps built with ember also to be sustainable with ember. ;)

Hope the above comments are received here on this topic. That is my comment on what I think is needed for handing deprecations warnings. I will be happy to take the rest of any issue followup elsewhere.

text/0000-deprecation-warning-handlers.md
```diff
@@ -0,0 +1,172 @@
+- Start Date: 2015-06-23
+- RFC PR: (leave this empty)
+- Ember Issue: (leave this empty)
+
+# Summary
+
+Deprecations and warnings in Ember.js should have configurable runtime handlers.
```
pixelhandler: I would suggest also configurable compile time handlers too.

text/0000-deprecation-warning-handlers.md
```diff
@@ -0,0 +1,172 @@
...
+
+# Drawbacks
+
+By not providing a robust log-level API, we are punting complexity to the
+consumer of this API. For a low-level tooling API such as this one, it seems
+and appropriate tradeoff.
+
+# Alternatives
+
+Each app can stub out `deprecate` and `warn`.
```
pixelhandler: Can the template complier also have configuration to stub `deprecate` and `warn` too? seems like it would be a big advantage for example that would be ideal compared with debugging the asset pipeline.

mmun: Yes, and I agree.

rwjblue: This RFC applies equally to the template compiler (they share the same Ember.deprecate logic).

rwjblue:
> This is an RFC, sorry my comments include issues, from my perspective it was those issues that motivated me to comment here so the future would be brighter and maybe less gloomy than the present. So here is my comment... (or story if you will)

@pixelhandler - I hope that you have not taken my replies as negative or discouraging, as I mentioned earlier I want to make sure that the actual proposed API here can be used to solve the real problems that you have been describing. We absolutely appreciate your comments, and are definitely thinking of these exact scenarios you have been detailing when we decided to work on this proposal.

pixelhandler:
@rwjblue it's all good, thanks for the dialog. After some research I think I found the [perfect build scenario](https://www.youtube.com/watch?v=dQw4w9WgXcQ)

mixonic:
I've added a note about the meaning of "runtime" in this document. The suggested API should be viable for template-compile-time warnings and deprecations.

I consider this vaguely +1 by core, but will add it to this week's agenda for a final :+1:. If there are remaining concerns, please raise them!

mixonic:
Merged!

About a million years ago, several folks including @pangratz @bantic and @rwjblue offered to implement this feature. If anyone would like to dive in, I'm happy to provide context and do reviews along the way.

Additionally, @rwjblue has suggested that this API could be back-ported to 1.13 (via overwriting Ember's deprecate function on the global namespace) and used by addons and tooling during the 1.13 -> 2.0 transition.

rwjblue:
Just a little follow-up on this RFC:

* Implemented behind a feature flag in Ember in https://github.com/emberjs/ember.js/pull/11833
* All deprecations updated to include `id` and `until`.  Tracked in https://github.com/emberjs/ember.js/issues/11835.
* Polyfill for usage with prior Ember versions: https://github.com/rwjblue/ember-debug-handlers-polyfill

PR #67, opened 2015-06-19
Better record polymorphism

igorT:
https://github.com/emberjs/rfcs/blob/polymorphism/text/0000-find-polymorphism.md

### DIFF
--- a/text/0000-find-polymorphism.md
+++ b/text/0000-find-polymorphism.md
@@ -141,13 +141,14 @@ This approach has several advantages:
 It would be much easier for `line.save()` to automagically go to /shapes
 
 However, I believe that the drawbacks are more significant:
-1.This approach is much more restrictive, you can either always include or not include subclasses,
-   but cannot decide at invocation time if you want to opt out
-2. In larger teams it might be surprising that one day `store.findAll('comments')` is returning other
+
+  1.This approach is much more restrictive, you can either always include or not include subclasses,
+    but cannot decide at invocation time if you want to opt out
+  2. In larger teams it might be surprising that one day `store.findAll('comments')` is returning other
 random models just because somebody on the team added an abstractBaseClass: true to the `comment` model.
 This is a bigger problem than forgetting to add `{ polymorphic: true }` because errors can happen in a completely
 different part of the app.
-3. This approach is incosistent with our current relationship polymorphism. It would require us to either deprecate
+  3. This approach is incosistent with our current relationship polymorphism. It would require us to either deprecate
 `hasMany({ polymorphic: true })` (and we just released 1.0 :( ) or to use an inconsistent mixed approach.
 
 # Unresolved questions
--- a/text/0000-find-polymorphism.md
+++ b/text/0000-find-polymorphism.md
@@ -1,0 +1,154 @@
+- Start Date: (fill me in with today's date, YYYY-MM-DD)
+- RFC PR: (leave this empty)
+- Ember Issue: (leave this empty)
+
+# Summary
+
+Add a `{ polymorphic: true }` flag to `store.findAll`, `store.findRecord`, `store.peekAll`,
+`store/peekRecord` which would include subclasses in the results.
+
+# Motivation
+# Current State
+Model Polymorphism is a popular and widely used feature of Ember Data. It is also supported
+out of the box in JSONApi.
+ Polymorphism is a natural way to model various domains, such as having `Comment` superclass
+and `TextComment`, `VideoComment` subclasses, or a `Shape` superclass that has `Line` and `Circle` as subclasses.
+
+A more advanced use case which is supported today is using Mixins for polymorphism.
+Consider a Social Network app, where several different object can be commented on.
+
+```js
+var Commentable = Ember.Mixin.create({
+  comments: hasMany('comment')`
+});
+```
+
+```js
+var BlogPost = DS.Model.extend(Commentable, {
+  title: DS.attr();
+});
+```
+
+```js
+var Comment = DS.Model.extend({
+  parent: DS.belongsTo('commentable', { polymorphic: true });
+});
+```
+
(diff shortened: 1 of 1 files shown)


igorT:
It seems like people like the alternative better, I will update the RFC

trabus:
Having tried to use polymorphic models before (eventually dropping them because they weren't working the way we needed), I feel that the alternative proposal is not necessarily any simpler or clearer for developers. I also agree with the drawbacks being a big enough issue to not go with the alternative, and prefer the rfc proposal instead.

knownasilya:
I like the first approach, it's explicit and clear.

Finhas:
Any update on this? I agree with @trabus and @knownasilya on this one. 

In the drawbacks, you mentioned that if the `polymorphic` flag is not used or forgotten, you would not get data. But could a query for shape that gets
 
`{ data: { id:1, type: 'line' }, { id:2, type: 'circle' }, { id:2, type: 'shape'}}`

return

`{ data: { id:2, type: 'shape'}}`??

mmoonport:
Have we given this any more thought. I like the second approach as it closely mirrors how our backend models work. It will provide consistent results with no code changes to the calls that are previously out there. And in my experience if you have a polymorphic model you are most likely calling the sub class models in finds because you can't call the parent and get all the models you'd expect back. This would greatly simplify some of our code if we could just pull the trigger on this

btecu:
I'd be glad to have either of the approaches working. Any news on this?

ffaubry:
Progress?

leifcr:
Will the ids-and-type strategy lay parts of the groundwork for this? See http://emberjs.com/blog/2016/05/03/ember-data-2-5-released.html#toc_upcoming-features-in-ember-data-2-6-beta-1 should be

miguelcobain:
@leifcr I doubt it. That feature is basically a polymorphic relationship extension. It was motivated by the fact that an id isn't enough to uniquely identify an object.

xaqq:
Something like that is desperately needed to avoid horrible workaround.

As is, `findAll` / `peekAll` are **completely broken** from an OOP point of view.
Inheritance means "is-a" so it makes zero sense to not load subtypes instances when doing `findAll('baseClass')`.

I hope support for something like this will be implemented soon.

buritica:
hey @igorT is there something we can do to help move this forward?  We hit this problem and are having issues handling some polymorphic relationships. If you give us some guidance, we could take a stab at implementing it with @sescobb27 & @nicollejimenez

BryanCrotaz:
I've got an interest in this as well. We built our own addon but it uses private api that's broken in ED 2.10. What's the current state?

BryanCrotaz:
Added a proposal for automatic hierarchy discovery, including code that we use in our ED 2.8.0 addon

https://github.com/emberjs/rfcs/pull/244

courajs:
We could improve the "refactor hostile" drawback of the `{ polymorphic: true }` approach.
`findAll('shape')` can warn in the console if shape has subclasses. Then, the warning can be dismissed by just adding `{ polymorphic: false }`.

BryanCrotaz:
@courajs sounds good. Default should definitely be polymorphic. Should be able to turn the warning off, otherwise it will flood a real polymorphic app.

BryanCrotaz:
3 years since the original issue is there any progress? I'm using an addon to fake some of it but it uses private API so is tied to 2.10. 2.10 stops me upgrading ember to use glimmer as pushPayload is broken with later ember. 

I'm. 
Stuffed. 
Would love to help but I have no idea what progress has been made or what help is needed

slimeate:
> I'm using an addon to fake some of it

@BryanCrotaz, do you think your addon might make a good starting point to implement this RFC? If it's available on github, I'd love to take a look. My team has been maintaining some custom adapter/serializer code in our app to essentially handle what this RFC proposes but I'd love to use and contribute to a community solution instead.

knownasilya:
Not sure if https://github.com/EmberMap/ember-data-storefront would be a good place for this, worth checking there too.

runspired:
I suspect now that Ember has true class inheritance that this is something we could support more easily and would be nice to revisit.

runspired:
I think we have a much simpler path to polymorphism for everything *except* findAll

- for `query` and `queryRecord` and `findRecord` we return whatever was in the primary data membership of the api response, we blindly assume any records in there are valid records, since the data key signifies that to us.
- for findAll (which returns peekAll), peekRecord and peekAll we continue to return only the narrowed sub-type. For these methods polymorphism is easily solved for in user space by joining two arrays or joining multiple peeks spanning the desired types.
- we can *however* also add a new API for peek! because of the work on `identifiers`, apps have a public API available to them by which to create secondary indeces for records. This can include polymorphic lookup indeces and emner-data's cache already keeps track of what polymorphic types exist for a given identifier: https://github.com/emberjs/data/blob/master/packages/-ember-data/tests/integration/identifiers/polymorphic-scenarios-test.ts#L96. We could likely introduce a new peek API that returns the identifiers for any records of a given base-type.

sly7-7:
@runspired Not sure my use case is related, but I would something like 'lazy load' records coming from an async polymorphic hasMany, one at a time, to be able to put placeholders while each record is loading.
For now, I'm using a private api from hasManyReference, 
 like the `ids()` function (https://github.com/emberjs/data/blob/master/packages/store/addon/-private/system/references/has-many.ts#L204), but using the whole identifiers, and then I can `store.findRecord(identifer.type, identifier.id)`.
Do you think something like `hasManyReference#identifiers()` could make sense ? (or adding a parameter to `ids()` to return the full identifier)

wagenet:
@runspired do you think this RFC can be revived? or does it make more sense to write a new one?

runspired:
@wagenet probably best as a new RFC

wagenet:
@runspired I'll go ahead and close. If you need a tracking issue will you be able to take care of that?

PR #70, opened 2015-06-29
Namespaced components & helpers lookup

bcardarella:


stefanpenner:
One thing worth noting, npm modules also use @ for scoped packages. It's likely worth enumerating how this does or doesn't conflict/confuse.

bcardarella:
@stefanpenner I would personally prefer `@` as well, but as I noted it introduces additional complexities with the implementation of this

bcardarella:
It actually appears using `@` in a tag name is flat our illegal:

rwjblue:
@bcardarella - Illegal in HTML is potentially good for us, because we can potentially use that to implement in HTMLBars parser and know for certain that we have no conflicts with default elements.

bcardarella:
@rwjblue how would that be represented in the DOM? If the element itself cannot exist in that form would it be re-written when inserting?

stefanpenner:
> @bcardarella - Illegal in HTML is somewhat good for us, because we can potentially use that to implement in HTMLBars parser and know for certain that we have no conflicts with default elements.

Ya, it is a pro IMHO. Although it means, it will never translate to a real DOM name, which is likely for the best.

rwjblue:
This largely works (to some extent) and is in use in a number of apps in the wild by way of https://github.com/rwjblue/ember-holy-futuristic-template-namespacing-batman which implements the `::` syntax proposed by the original module unification RFC.

I think more work will be forthcoming to make "real" template imports work, as well as using actual components and helpers directly as values (see #432 for details there).

PR #73, opened 2015-07-06
Make jQuery Optional

sandstrom:
[Link to rendered RFC](https://github.com/sandstrom/rfcs/blob/master/text/0000-jquery-optional.md)

### DIFF
--- a/text/0000-jquery-optional.md
+++ b/text/0000-jquery-optional.md
@@ -1,0 +1,75 @@
+- Start Date: 2015-07-06
+- RFC PR: (leave this empty)
+- Ember Issue: (leave this empty)
+
+# Summary
+
+Make jQuery optional.
+
+# Motivation
+
+Why are we doing this? What use cases does it support? What is the expected outcome?\
+
+I use jQuery heavily and like it! But decoupling it from Ember have some benefits: it's not always needed,
+some people think payload size is important (I don't) and it may have minor performance contributions.
+
+# Detailed design
+
+I've tried to track down the places where jQuery is used and its implications.
+There is some primary usage (around events) as well as some auxiliary usage, where it should be fairly easy
+to replace it (I hope).
+
+**Primary use** *(will require insight from people with good knowledge of browser difference)*
+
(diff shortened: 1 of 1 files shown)


text/0000-jquery-optional.md
```diff
@@ -0,0 +1,75 @@
...
+
+# Alternatives
+
+Wait 1-2 years, then axe jQuery.
+
+# Unresolved questions
+
+- How would the core team like to handle the 'make it optional' switch? May affect the build process,
+library requirements etc.
+- The FF11 requirement, what is the minimum supported version of FF? Is it fine to drop? (FF is on ~38 now)
```
dukex: FF11

mike1o1:
One approach the Angular team takes is to use Google's [jqLite](https://code.google.com/p/jqlite/) library if jQuery itself isn't loaded. Is it worth investigating to see if a similar approach can be taken for Ember? 

This can also be exposed for developers. If a developer invokes `Ember.$` or `this.$()` within a view and jQuery is not available, Ember would return an instance of jqLite. Angular's equivalent to `Ember.$` is `angular.element`, which is documented [here](https://docs.angularjs.org/api/ng/function/angular.element), in case you wanted some prior art to look at.

locks:
Related: https://github.com/emberjs/ember.js/issues/3791

runspired:
Related https://github.com/emberjs/rfcs/pull/86 removes most of the reliance on jQuery for eventing.

In my current implementation, `jQuery.event.fix(e)` is used to normalize the event. This is something that could be pulled out as a micro-lib exposing the needed jquery AMD module.

https://github.com/runspired/ember-capture-eventing/blob/master/addon/event_dispatcher.js#L379

ming-codes:
Related: https://github.com/rwjblue/ember-native-dom-event-dispatcher

Haven't tried it yet, but assuming it works, it's a bridge to be rid of jQuery.

sandstrom:
Closing this RFC, since work is [now](https://github.com/emberjs/ember.js/issues/16058) [underway](https://github.com/emberjs/data/issues/5320) to make jQuery optional.

PR #75, opened 2015-07-06
[meta] Add note about filing issues before full RFC PRs

davewasmer:
It might be helpful in some cases, as @mixonic [suggested](https://github.com/emberjs/ember.js/issues/11645#issuecomment-118868091), to hold high-level conceptual discussions and organize feedback in an issue prior to diving into the implementation details that a full RFC requires.


rwjblue:
Thank you!

mmun:
:+1: :+1: :+1:

PR #79, opened 2015-07-31
Improve CPs ergonomics by changing caching strategy

cibernox:
[Link to rendered](https://github.com/cibernox/rfcs/blob/computed_property_caching_strategy/text/more-friendly-cp-caching.md)

### DIFF
--- a/text/more-friendly-cp-caching.md
+++ b/text/more-friendly-cp-caching.md
@@ -1,0 +1,125 @@
+- Start Date: 2015-07-31
+- RFC PR: (leave this empty)
+- Ember Issue: (leave this empty)
+
+# Summary
+
+At the moment, the setter function in computed properties caches the return value of the function,
+even when the return value is undefined. While undefined is a perfectly valid value, cases where
+you really want to cache undefined are very rare, and most of the time is just the user forgetting
+to return from the function.
+
+I think there is a better (or even completely) retrocompatible way to improve user experience
+and be closer to regular javascript.
+
+# Motivation
+
+I've found that computed properties with getters and setters keep bitting newcomers because they
+don't behave like regular JS getters and setters, and I think they can be more user friendly and
+also more aligned with regular javascript.
+
+This is the kind of code that keeps confusing users:
+
+```js
+height: 100,
+goldenRatioWidth: computed('height', {
+  get() {
+    return this.get('height') * 1.618;
+  },
+  set(_, value) {
+    this.set('height', value / 1.618);
+  }
+})
+```
+
+I've explained myself half a docen times in the last month to confused programmers why this
+snippet doesn't work as expected, and it's easy to see why they are confused. The equivalent
+code with regular getters and setters whould look like this:
+
+```js
+heigth: 100,
+get goldenRatioWidth() {
+  return this.height * 1.618;
+},
+set goldenRatioWidth(value) {
+  this.height = value / 1.618;
+}
+```
+
+No return, but the code works as expected because there is no caching, and the next time the
+property accessed the getter will be invoked and the expected result will be computed.
+
+I think that computed properties should be as close to this behavior as they possibly can,
+specially when code with decorators will look like this:
+
+```js
+@computed('height')
+get goldenRatioWidth() {
+  return this.height * 1.618;
+},
+set goldenRatioWidth(value) {
+  this.height = value / 1.618;
+}
+```
+
+In this snipped above I'd expect the code to behave the same with and without the decorator except
+for the "caching with declarative expiration policy" that the decorator adds.
+
+# Detailed design
+
+The reason why people gest confused is that most people doesn't expect setter to cache.
+Although your getter and setter can do any crazy stuff, in most sanely defined computed properties
+developers look for "idempotence" (not sure if it's the better term). When you invoke
+`set('myProp', 200)` they expect the next time they get the property with `get('myProp')` to
+get the same value they just set.
+
+When someone implements a getter in a computed property is because setting state in this
+computed property:
+
+A) Has to set state in another property, like the example above. Tipically that other property
+is one of the dependent keys.
+
+B) Has to trigger some callback-like logic. Tipically this is way of replacing an observer.
+
+In computed properties for case A, after setting the state in a dependent key the users, used to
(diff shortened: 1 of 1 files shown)


stefanpenner:
i like this approach, upgrade path seems kinda rough.

cibernox:
@stefanpenner I added a 4th idea involving decorators that can be a real option.

Having the `@computed` decorator have different semantics is not technically a breaking change because it's new syntax, so it could be doable and at the same time allow to deprecate old usages of `computed`.

stefanpenner:
True

rwjblue:
I personally like the idea that calling the setter would clear the cache and mark as dirty. I believe we can and likely should fix this within the 2.x cycle.

I propose the following path:

* Add a new feature flag to enable the new behavior.
* Allow that feature flag to be enabled at runtime throughout the 2.x cycle (value in `features.json` stays `null` throughout all of 2.x). 
* After feature has hit `2.x.0-beta.1`, add a deprecation when the feature is not enabled to canary (will become `2.(X+1).0-beta.1` after 6 weeks of deprecations).
* Enable the flag by default in newly generated ember-cli applications.
* Enable flag by default in 3.0.0.
* Remove old code.

Following the above path, we satisfy both our SemVer requirements and our requirement that 3.0.0 not have any non-deprecated changes.

cibernox:
If I hear another positive voice from anyone else on the core I'll start a spike of the implementation.

stefanpenner:
@rwjblue I'm unconvinced we can actually make this change in app CP as docs/guides/learning hazards would likely prevail not to mention addons that want to work in both worlds, now need to do special magic, but i do think we can use the decorators as a signal for a refresh.

rwjblue:
FWIW, this exact path was followed by the `ember-routing-drop-deprecated-action-style` feature that removed looking up actions on the root of a controller.  That flag was introduced around 1.5.0, and ultimately the deprecated code path was removed in 1.8.0.

---

@stefanpenner:

>  I'm unconvinced we can actually make this change in app CP as docs/guides/learning hazards would likely prevail

> not to mention addons that want to work in both worlds

The majority of new features in the last few cycles have also implemented a nice polyfill-like addon that addon authors can use to get the new behavior across versions.  This has been a pretty successful technique, and includes things like `ember-new-computed` addon (poorly named, but highly used), `ember-debug-handlers-polyfill`, and `ember-getowner-polyfill`. 

> I do think we can use the decorators as a signal for a refresh

Sure, I do not disagree, but I think that the path forward is fairly simple/straightforward as I laid out in https://github.com/emberjs/rfcs/pull/79#issuecomment-176863246.

stefanpenner:
@rwjblue the polyfil approach, and enabling on a per computed property basis would address my concerns totally.

rwjblue:
>  enabling on a per computed property basis would address my concerns

Ya, this is a great idea.  Using the polyfill would already allow upgrading on a per CP basis, but we could also add an option flag to be used to force the setters return value to be cached (the existing/older behavior) even when the feature flag is enabled.  Something like:

```
goldenRation: computed('height', {
  cacheSetterReturnValue: true,

  get() {
    return this.get('height') * 1.618;
  },
  set(_, value) {
    return this.set('height', value / 1.618);
  }
```

cibernox:
@stefanpenner The behaviour of the new CPs is more or less a subset of the original one. 
Basically this removes caching from setters, and will most likely only affect people caching undefined. 

I might be wrong, but seems compatible enough.

Also, what about something like

```js
foo: computed('bar', 'baz', {
  get() { /* ... */ },
  set() { /* ... */ }
}).uncachedSetter();  // Enables the new behaviour when globally disabled

foo: computed('bar', 'baz', {
  get() { /* ... */ },
  set() { /* ... */ }
}).cachedSetter(); // Restores the old behaviour when globally enabled
```

This *perhaps* is pollyfillable.

lukemelia:
I agree that the proposed approach would make the lessen the learning curve, though I'm quite sure there is a lot of code out there that relies on the current behavior.

If we decide not to change the behavior but want to improve user education, one idea would be to use the function.toString() approach mentioned in development builds to warn on setters that don't contain `return`.

rwjblue:
@lukemelia - Yes, agreed, even if we do not want to change the go forward behavior (which I personally do want to change) we should make development ergonomics better.

stefanpenner:
> though I'm quite sure there is a lot of code out there that relies on the current behavior.

Yes, this. I know for a fact, large swaths of code depend on this (accidentally or intentionally). Many scenarios the get/set are not symmetric, and changes this would be quite painful.

Incrementally upgrading via new syntax and on a per-cp basis could mitigate these concerns. I would love to get @tomdale / @wycats input early though.

workmanw:
> I personally like the idea that calling the setter would clear the cache and mark as dirty. I believe we can and likely should fix this within the 2.x cycle.

I absolutely agree. 

We have a few macros that have very complex CP logic for validating and marshalling data. In some cases we end up having to make a 3rd function that just calculates the return value so it's reusable for  `get` and `set`. It's really frustrating to look at. This would be a welcomed changed.

caseywatts:
This bit me again today lol. Here's my example in case it helps :)

[ES6 getter/setter docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/set) suggests we don't use a return value during `set()`, and I thought Ember would behave the same.

I ended up with this silly code 
```
  priceDollars: computed('priceCents', {
    get() {
      return centsToDollars(this.get('priceCents'));
    },
    set(attribute, value) {
      const priceCents = dollarsToCents(value);
      this.set('priceCents', priceCents);
      return centsToDollars(priceCents);
    }
  })
```

I also tried `return this.get('priceDollars');` but that returns the old value lol

I'd rather have done this, with no return value:
```
  priceDollars: computed('priceCents', {
    get() {
      return centsToDollars(this.get('priceCents'));
    },
    set(attribute, value) {
      this.set('priceCents', dollarsToCents(value));
    }
  }),
```

locks:
@caseywatts do you think https://guides.emberjs.com/v3.0.0/object-model/computed-properties/#toc_setting-computed-properties could be updated to make the behaviour clearer?

cibernox:
🙌🏻

rwjblue:
Thank you for helping push this forward!

PR #85, opened 2015-08-15
Heavy petting for the contextual component RFC

mixonic:
@mmun for you to consider.

### DIFF
--- a/text/0000-contextual-component-lookup.md
+++ b/text/0000-contextual-component-lookup.md
@@ -4,126 +4,230 @@
 
 # Summary
 
-Currently, you can only invoke components using a global name. The goal of this
-RFC is to allow better composition of components by allowing components
-to be referenced through property lookups on the current scope. For example,
+The goal of this RFC is to allow for better component composition and the
+usage of components for domain specific languages.
 
-```hbs
-{{this.localInput value=post.title}}
-```
+Ember components can be invoked three ways:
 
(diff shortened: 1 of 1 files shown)


workmanw:
:+1: I could definitely see several places in our we could use this.

--------------

One question, what about proving class references, rather than strings, to the component helpers. For private component classes? For example:

```js
// app/components/special-button.js
var MyPrivate = Ember.Component.extend({ /* ... */ });

  myPrivateComponent: MyPrivate

  /* ... */
});
```

```hbs 
{{!-- app/templates/components/special-button.hbs --}}
{{!-- passed by class, not string --}}
{{component myPrivateComponent}} 
```

We've got a lot of components in our app that aren't used in block form and have "private" sub-components. It gets kind of messy to have every component in it's own .js file sometimes. 

--------

I was secretly hoping to see some mention of a splat-like-operator, but I realize it's probably outside the scope of this RFC.

mixonic:
@workmanw splat is being discussed at https://github.com/wycats/handlebars.js/issues/1050 We're waiting for wycats to clear his stack. Hopefully we green light soon :-)

@mmun has concerns about how we handle the container (and therefor injections etc) on bare classes. I think it may be surmountable. If he is hesitant I would at least like to get the un-controversial version of this rolling so we can land it soon, maybe 2.2 if core approves on Friday.

PR #87, opened 2015-08-26
Ember.computed.sortBy

workmanw:
[Rendered](https://github.com/workmanw/rfcs/blob/computed-sort-array-argument/active/0000-computed-sort-array-argument.md)

It's not really a big RFC. Happy to add any further detail, I just didn't want to draw it out when it was straight forward.

### DIFF
--- a/active/0000-computed-sort-array-argument.md
+++ b/active/0000-computed-sort-array-argument.md
@@ -1,0 +1,56 @@
+- Start Date: 2015-08-26
+- RFC PR: (leave this empty)
+- Ember Issue: (leave this empty)
+
+# Summary
+
+In Ember 2.x, `Ember.computed.sort` is the one and only computed macro for sorting a list. Its current API is built assuming that the sort order may change at runtime, leaving no way to easily create a declarative sort macro that you know is immutable. This RFC proposes adding an additional argument type for the `sortDefinition` parameter to make it easier to declare and to ensure sort order is immutable.
+
+# Motivation
+
+I'm hoping to satisfy two primary annoyances with this RFC. First, `Ember.computed.sort` can be used as a singular definition as the sort order will be the value of the second argument, you will not have to look for another definition on the class, thus improving code readability. In our app, we often have the following:
+
+```javascript
+var NewestPostsController = Ember.Controller.extend({
+  posts: null,
+  newestPosts: Ember.computed.sort('posts', '_newestPostsOrder'),
+  _newestPostsOrder: ['createdOn:desc']
+});
+```
+
+The second annoyance is I know that if `_newestPostsOrder` was inadvertently changed, it could cause `newestPosts` to be incorrect. I realize this would just be a bug in our app, but having the ability to define an immutable sort order would give the same peace of mind as a readOnly computed property.
+
+# Detailed design
+
+Right now the second argument for `Ember.computed.sort`, `sortDefinition`, accepts two values, a string and a function. I'm proposing we add array as a third allowed argument type. The array would expect the exact same sort properties format as it does when using a string as a dependent key. Reusing the above example,
+
+Before:
+```javascript
+var NewestPostsController = Ember.Controller.extend({
+  posts: null,
+  newestPosts: Ember.computed.sort('posts', '_newestPostsOrder'),
+  _newestPostsOrder: ['createdOn:desc']
+});
+```
+
+After:
+```javascript
+var NewestPostsController = Ember.Controller.extend({
+  posts: null,
+  newestPosts: Ember.computed.sort('posts', ['createdOn:desc'])
+});
+```
+
+This design would not break semver as it's an additional argument type that was previously not supported.
+
+# Drawbacks
+
+I can think of two drawbacks. First and foremost, the array argument type will become reserved for this functionality. It could not easily be reused for a new extension in the future. Second, it's ultimately just a more convenient way of doing something that is already functionally possible.
+
+# Alternatives
+
+No real alternatives were considered. The only thing that came to mind as an alternative was adding an additional computed macro for this use case (EG `Ember.computed.readOnlySort`).
+
+# Unresolved questions
+
+I don't believe there are any unresolved questions at this time.
--- a/active/0000-computed-sort-array-argument.md
+++ b/active/0000-computed-sort-array-argument.md
@@ -22,7 +22,7 @@ The second annoyance is I know that if `_newestPostsOrder` was inadvertently cha
 
 # Detailed design
 
-With this RFC I'm proposing we introduce an additional computed macro, `Ember.computed.sortBy`, to fill the gap. The argument order would be similar to that of `Ember.computed.sort`. Reusing the above example,
+With this RFC I'm proposing we introduce an additional computed macro, `Ember.computed.sortBy`, to fill the gap. The argument style would be similar to that of `Ember.computed.sort`. However, with `Ember.computed.sortBy` all non-primary sort orders (secondary, tertiary, etc) would be defined as additional arguments. `Ember.Enumerable.sortBy` serves as existing precedence for this pattern. Reusing the above example,
 
 Before:
 ```javascript
@@ -37,13 +37,8 @@ After:
 ```javascript
 var NewestPostsController = Ember.Controller.extend({
   posts: null,
-  newestPosts: Ember.computed.sortBy('posts', ['createdOn:desc', 'upvoteCount'])
+  newestPosts: Ember.computed.sortBy('posts', 'createdOn:desc', 'upvoteCount')
 });
-```
-
-Additionally, the sort order could be defined as a string instead of a single item array. Eg:
-```javascript
-  newestPosts: Ember.computed.sortBy('posts', 'createdOn:desc')
 ```
 
 
--- a/active/0000-computed-sort-array-argument.md
+++ b/active/0000-computed-sort-array-argument.md
@@ -4,17 +4,17 @@
 
 # Summary
 
-In Ember 2.x, `Ember.computed.sort` is the one and only computed macro for sorting a list. Its current API is built assuming that the sort order may change at runtime, leaving no way to easily create a declarative sort macro that you know is immutable. This RFC proposes adding an additional argument type for the `sortDefinition` parameter to make it easier to declare and to ensure sort order is immutable.
+In Ember 2.x, `Ember.computed.sort` is the one and only computed macro for sorting a list. Its current API is built assuming that the sort order may change at runtime, leaving no way to easily create a declarative sort macro that you know is immutable. This RFC proposes adding an additional computed macro for declarative and immutable sorting.
 
 # Motivation
 
-I'm hoping to satisfy two primary annoyances with this RFC. First, `Ember.computed.sort` can be used as a singular definition as the sort order will be the value of the second argument, you will not have to look for another definition on the class, thus improving code readability. In our app, we often have the following:
+I'm hoping to satisfy two primary annoyances with this RFC. First, providing a way to declare a computed sort macro as a singular definition (one line of code). As opposed to `Ember.computed.sort` which requires two declarations to use. It's my belief this will improve readability for some use cases. In our app, we often have the following:
 
 ```javascript
 var NewestPostsController = Ember.Controller.extend({
   posts: null,
   newestPosts: Ember.computed.sort('posts', '_newestPostsOrder'),
-  _newestPostsOrder: ['createdOn:desc']
+  _newestPostsOrder: ['createdOn:desc', 'upvoteCount']
 });
 ```
 
@@ -22,14 +22,14 @@ The second annoyance is I know that if `_newestPostsOrder` was inadvertently cha
 
 # Detailed design
 
-Right now the second argument for `Ember.computed.sort`, `sortDefinition`, accepts two values, a string and a function. I'm proposing we add array as a third allowed argument type. The array would expect the exact same sort properties format as it does when using a string as a dependent key. Reusing the above example,
+With this RFC I'm proposing we introduce an additional computed macro, `Ember.computed.sortBy`, to fill the gap. The argument order would be similar to that of `Ember.computed.sort`. Reusing the above example,
 
 Before:
 ```javascript
 var NewestPostsController = Ember.Controller.extend({
   posts: null,
   newestPosts: Ember.computed.sort('posts', '_newestPostsOrder'),
-  _newestPostsOrder: ['createdOn:desc']
+  _newestPostsOrder: ['createdOn:desc', 'upvoteCount']
 });
 ```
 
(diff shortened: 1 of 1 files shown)


opsb:
This would be nice. The common case is that you specify the ordering explicitly, it's rare that I've wanted to bind the sort properties themselves.

rlivsey:
:+1: it's extremely rare that I update the sort property so making the default be simpler is great.

I wonder if you could detect if it's a sort definition by the presence of ':' or whether that would be too confusing, so you could use a string instead of an array:

```javascript
newestPosts: Ember.computed.sort('posts', 'createdOn:desc')
```

That might be too easy to footgun yourself by leaving out the sort order and wonder why it doesn't work:

```javascript
newestPosts: Ember.computed.sort('posts', 'createdOn')
```

opsb:
How about introducing Ember.computed.sortBy?

    newestPosts: Ember.computed.sortBy('posts', 'createdOn') # default to ascending

    newestPosts: Ember.computed.sortBy('posts', 'createdOn:desc')

workmanw:
@opsb Yea, I had mentioned this in my alternatives, but my `Ember.computed.readOnlySort` was a far less attractive name. I'll update my RFC.

active/0000-computed-sort-array-argument.md
```diff
@@ -0,0 +1,66 @@
...
+# Drawbacks
+
+I can think of two drawbacks. First and foremost, the array argument type will become reserved for this functionality. It could not easily be reused for a new extension in the future. Second, it's ultimately just a more convenient way of doing something that is already functionally possible.
+
+# Alternatives
+
+An alternative would be to implement a new computed macro called: `Ember.computed.sortBy`. This would allow for more argument/API flexibility. Here a few examples:
+
+```javascript
+newestPosts: Ember.computed.sortBy('posts', 'createdOn') # default to ascending
```
mitchlloyd: You've been writing too much Python ;)

# => //

mitchlloyd:
To me the current alternative looks a little better than proposal. It lines up nicely with the Ember array methods.

This seems like a useful feature and a nice API. In the past people have been adverse to adding more CPMs to Ember, encouraging user-land libraries like https://github.com/cibernox/ember-cpm instead. It will be interesting to hear more feedback.

active/0000-computed-sort-array-argument.md
workmanw: Hahaha. My brain didn't even notice. :)

opsb: Ha, it's a ruby week for me...

workmanw:
@mitchlloyd The more I thought about it, the more that I do like the alternative. But I think we [ember] tend to add lots of little things, and then fold them back together. My inclination was to not increase the surface of the API.

But yea, we'll see how interested in this people actually are. If it gets a luke warm response, I'll open a pll to ember-cpm. :)

active/0000-computed-sort-array-argument.md
```diff
@@ -0,0 +1,66 @@
...
+
+newestPosts: Ember.computed.sortBy('posts', 'createdOn:desc')
```
MiguelMadero: minor: this doesn't need naturally, I would expect it to read `sortBy fieldNameImSorting` e.g. sortBy firstName. This is the case for array.sortBy and other similar ones like filterBy and mapBy. When we have the arrayPropertyName as the first parameter it breaks the readability. I think that extending the current API makes more sense.

lukemelia:
I'm in favor of introducing `Ember.computed.sortBy`. I've long found `Ember.computed.sort` to have a confusing API, and I think that leaving it as is for advanced scenarios, and having `sortBy` for simpler ones is a big usability win.

stefanpenner:
I like the idea of a sort by

active/0000-computed-sort-array-argument.md
mitchlloyd: For those of us who lean towards a more functional style when they can, this actually reads pretty normally. For instance with LoDash this would be `sortBy(posts, 'createdOn')`.

workmanw:
After further consideration, I too think `sortBy` is better. And it seems that's pretty much the consensus here. I'm looking for advice on how to transition this RFC to reflect that. Should I modify this RFC to reflect that, should I close this and open a new one, or should just leave this as is for now?

Also, FWIW, if there is a green light for `sortBy`, I'd be happy to implement it. I have some familiarity with the implementation of `Ember.computed.sort`.

active/0000-computed-sort-array-argument.md
MiguelMadero: sounds good

mitchlloyd:
@workmanw I've noticed that some RFC PRs here change dramatically before they are merged. For instance the approach in the "contextual components" RFC changed so much that the name of the branch (contextual-component-lookup) is at odds with where the proposal landed. Seems like it's nice to keep the history together even if the branch name isn't perfect.

workmanw:
@mitchlloyd Thanks!

I've updated this RFC to reflect `Ember.computed.sortBy` as the proposal.

rwjblue:
I am :+1:, I'll try to bring this up at the next core team meeting.  In the meantime, can you create an addon that provides this capability (which could be used as a polyfill once this feature lands)?

workmanw:
@rwjblue Sure thing, I'd be happy to get moving on it this weekend. Ultimately, I may have wait for emberjs/ember.js#12221 to be resolved (also see: emberjs/ember.js#12212). But I'll definitely see how far I can get.

workmanw:
@rwjblue Done!
I created [workmanw/ember-computed-sortby](https://github.com/workmanw/ember-computed-sortby) and published it to npm. I included basic docs in the README and code. And tests of course. 

For those interested in trying it you can use:  `ember install ember-computed-sortby`.

mixonic:
@workmanw could this just be an update to `sort` to accept an array as the second argument?

Or alternatively if you suggest `sortBy('posts', 'createdOn:desc')`, why keep the array syntax at all? Just allow multiple arguments after the first.

workmanw:
@mixonic 
> Could this just be an update to sort to accept an array as the second argument?

This was actually the initial suggestion of this RFC. `Ember.computed.sortBy` was the alternative. But there was consensus that `Ember.computed.sortBy` was more favorable. After further consideration, I started to agree. My primary reason was that having `sortBy` as a separate macro meant you could provide a nicer API and made it easier for someone reading the code to know exactly what it was doing.

 > Or alternatively if you suggest sortBy('posts', 'createdOn:desc'), why keep the array syntax at all? Just allow multiple arguments after the first.

This is a pretty good idea that hadn't occurred to me. It also matches the API of `Ember.Enumerable.sortBy`. I think this is probably worth adopting.

mixonic:
Nice. My other lingering thought, though I don't know where to go with it, is that microsyntaxes ala `prop:desc` generally become troublesome later on. Some wacky ideas:

```
sorted: Ember.computed.sortBy('thing', 'propA', ['propB', 'desc'], ['propC']);
sorted: Ember.computed.sortBy('thing', { property: 'propA' }, { property: 'propB', direction: 'desc' });
```

We don't need to avoid the microsyntax, but I encourage you to explore some other ideas before we lock it in.

workmanw:
I updated the RFC and the polyfill repo to remove the array syntax for the `sortDefinitions`. 

@mixonic I've given some thought to alternatives to the microsyntax, but I haven't come up with anything I loved. Everything just felt too verbose. I'm still open to exploring this, but given that `Ember.computed.sort` still relies on ':desc', I'm inclined to not break with the pattern. 

FWIW, the best alternate idea I came up with was:
```
sorted: Ember.computed.sortBy('thing', 'propA', 'propB', false, 'propC');
sorted: Ember.computed.sortBy('thing', 'propA', 'propB', -1, 'propC');
```

I had also thought about making a `symbol` for desc that could be specific to Ember. Something like:
```
Ember.descending = Symbol('desc');
sorted: Ember.computed.sortBy('thing', 'propA', 'propB', Ember.descending, 'propC');
```
But I loved this even less.

mitchlloyd:
I don't mind the tuple style arrays. They are still pretty concise without inventing syntax in a string.

```
Ember.computed.sortBy('thing', 'propA', ['propB', 'desc'], ['propC', 'asc']);
```

lukemelia:
I find the microsyntax easier to read and I'd posit it is just as likely to be future-friendly as a tuple-style array.

rwjblue:
To throw my hat in the ring:

I prefer the easy to understand but slightly more verbose versions:

```javascript
// option 1
Ember.computed.sortBy('thing', 'propA', {propB: 'desc', propC: 'asc'});

// option 2
Ember.computed.sortBy('thing', 'propA', {property: 'propB', direction: 'desc'}, { property: 'propC', direction: 'asc'});
```

I like "option 1" the best of these two.  

I am perfectly happy with supporting the microsyntax also, but would prefer for it to just be sugar on top of a better/easier to grok thing.  I find that these various microsyntaxes (that look similar but are different) are quite often a barrier to entry for new folks, and "Option 1" above is actually pretty close to the same number of total characters (slightly more, but not by much)...

kimroen:
I agree that considering `Ember.computed.sort` uses `:desc` I'd expect that to work here too. Then again, `sortBy` elsewhere does not have this syntax, which might be confusing. No matter what we come up with here, it's something we made up for this purpose, so I'm not sure it matters either way - anything would be experienced as different.

---

@rwjblue In your option 1, would these have the same behavior?

```js
Ember.computed.sortBy('thing', 'propA', {propB: 'desc', propC: 'asc'});
Ember.computed.sortBy('thing', 'propA', {propB: 'desc'}, 'propC');
Ember.computed.sortBy('thing', 'propA', {propB: 'desc'}, {propC: 'asc'});
```

Or to put it differently; what are the rules for the objects? Are they actually easier to understand?

workmanw:
I would say that I personally still lean toward the microsyntax because I find it the most concise and readable. Out of the ones proposed I liked this one the best:

```
Ember.computed.sortBy('thing', 'propA', ['propB', 'desc'], ['propC', 'asc']);
```

It feels like the easiest to read. The two alternates with pojos felt a little more jarring, especially with syntax highlighting. All that said, I don't have a strong opinion about the format at all. I'd be happy with anything that's been suggested here.

mitchlloyd:
One more variation on the options from @rwjblue. 

```javascript
Ember.computed.sortBy('thing', 'propA', { propB: 'desc' }, { propC: 'asc' });
```

Having the constraint of one object per property might make things more consistent given what @kimroen has shown. "micro:syntax", tuples, objects, no strong preference here either.

opsb:
As this is an exercise in ergonomics the microsyntax seems like the best option. It matches the existing syntax for Ember.computed.sort, it just provides a shortcut so that you don't have to use a separate property to specify the sort config.

rwjblue:
@kimroen - It was late and I mistyped (or copied), I meant to take two arguments the array and an object.  My example should have been:

```javascript
Ember.computed.sortBy('thing', { propA: 'asc', propB: 'desc', propC: 'asc' });
```

Also, the ergonomics here (IMHO) should not be about fewest characters typed, but readability/ease of understanding.  We can always start with something like I have listed above (which is still much better than having a second property that you have to go track down), and add a super slim string based microsyntax that desugars into the above later...

opsb:
I suppose the reason I'm drawn to the microsyntax is that I've already been using it in ember for years, by now it is the most readable syntax. Newcomers may have a different opinion though. I don't see any issue with supporting both syntaxes but I think it would be nice to include the microsyntax for those of us that already familiar with it.

mixonic:
As I began the bikeshedding, I will now put an end to it :-)

The place where the microsyntax is painful is not this proposed `sortBy`, but `sort`. `sort` is often used for headers that can change how the content is ordered when clicked, and in this case mucking with strings is unfortunate.

As `sortBy` is static, the microsyntax seems harmless. I support using it for `sortBy`. If anyone else feels strongly otherwise they should make a case advocating for their position. Thanks for the exploration folks!

workmanw:
@mixonic Alrighty :)

I'm happy to continue spending cycles where ever necessary to keep this RFC moving forward. With respect to the microsyntax, it seems like pretty split presently. It may just be one of those things the core team has to call the ball on.

workmanw:
It's been a few months since there has been any action on this. I have an addon that we've been using for the past few months in production. I currently have the bandwidth to submit an Ember PR if it seems valuable for this to reside in the core. If not, that's fine too, we can close this and keep it as an addon.

I'm not sure what the next step here is. The contributing.md is a little vague, but it sounds like I should just submit a PR? It was mentioned bringing this up at a core team meeting for discussion. So I'm not really sure what the next steps are or if the ball is even in my court.

mixonic:
@workmanw thanks for bringing it up. I will put this on the next core agenda for approval. Is the addon public? Can you link me up?

Sorry for the delay, but I will try to get you unblocked on Friday.

workmanw:
@mixonic Yeap! Here you go: https://github.com/workmanw/ember-computed-sortby

mmun:
Not really a fan of

```js
Ember.computed.sortBy('thing', { propA: 'asc', propB: 'desc', propC: 'asc' });
```

as it does not express clearly the fact that the order of the sort properties matters.

I don't love the microsyntax API but we should implement it because it is what people expect. If we can decide on a bette API later, we can uniformly deprecate the microsyntax API everywhere that it is used and transition to the new API.

workmanw:
Ping @mixonic :) 

I don't mean to keep bugging you, but I'm in a holding pattern right now :airplane:. I can just submit a PR if that would be a better place for this discussion to occur.

mixonic:
@workmanw core discussions have been waylaid by the holidays :-/ There is strong push-back regarding adding any new CP macros. See also: https://github.com/emberjs/ember.js/pull/12758

I will attempt to raise it again on Friday. Agreed that we should give you a solid direction.

workmanw:
> core discussions have been waylaid by the holidays.

@mixonic Yea that's what I figured, no worries :)

> There is strong push-back regarding adding any new CP macros. See also: emberjs/ember.js#12758

Oh interesting. I must say that I certainly understand wanting to keep the surface of the API as small as possible and if `sortBy` isn't right for the core that's fine by me. That said, from @stefanpenner's comment in the linked issue, it sounds like that moratorium has been in place for a while? But no one, including the core team members who have commented here, thought to mention it. That would be :(.

mixonic:
@workmanw! I owe you a direction, for better or worse.

We discussed this RFC and the general "CP moratorium" at today's meeting. The consensus is that we want to strictly limit introducing new macros, though there are some cases such as the [addition of uniqBy to make our APIs consistent](https://github.com/emberjs/ember.js/pull/12758) that are justified.

Additionally, there is a strong resistance to adding new microsyntax to Ember.

These two ideas (no new macros, no new microsyntax) mean we're going to close this RFC without accepting it. We are quite supportive of the [ember-computed-sortby addon](https://github.com/workmanw/ember-computed-sortby) though. Thanks for the effort on this, and I hope you feel the process made for a better addon if not for an addition to Ember itself.

workmanw:
Well I can't say I'm not a little disappointed, but I do appreciate it reaching a final conclusion. Thank you.

MiguelMadero:
Thanks @workmanw @mixonic  and everyone involved. I really appreciate the process, even when the outcome might seem negative, it sets a direction and precedent and there's good reasoning behind not doing it. Also, having the addon ecosystem enables innovation outside of core, which I think is perfect for this case.

mmun:
I agree with the outcome of this RFC but I am still strongly in favor of filling in this hole in the API.

There are three sorting methods and they should all share a uniform

- Ember.computed.sort
- Ember.computed.sortBy
- Ember.Enumerable#sortBy

Presently,

- Ember.Enumerable#sortBy can take a list of paths to sort by, but does not support the asc/desc microsyntax.
- Ember.computed.sort can take a list of paths to sort by and _does_ support asc/desc syntax.
- Ember.computed.sortBy does not exist in core.

Thus path forward is thus,

- Deprecate the current the microsyntax used by Ember.computed.sort.
- Find a new way to express asc/desc without using a microsyntax. It must be backwards compatible with the microsyntax.
- Ensure that all three APIs uniformly use the new API.

PR #88, opened 2015-09-05
Addon Version Compatibility

elwayman02:
View the rendered RFC [here](https://github.com/elwayman02/rfcs/blob/addon-compatibility/text/0000-addon-version-compatibility.md).

Thanks to @kategengler & others for the concept and @jkgisme for reviewing the RFC before submission.

### DIFF
--- a/text/0000-addon-version-compatibility.md
+++ b/text/0000-addon-version-compatibility.md
@@ -1,0 +1,79 @@
+- Start Date: 2015-09-04
+- RFC PR: (leave this empty)
+- Ember Issue: (leave this empty)
+
+# Summary
+
+This feature adds a standard of marking addons with version compatibility to core dependencies such as Ember, Ember-CLI, Ember-Data, etc.
+
+# Motivation
+
+Why are we doing this? What use cases does it support? What is the expected outcome?
+
+The Ember Addon ecosystem is booming, but it is also becoming increasingly crowded with a deluge of information. 
+Sites like Ember Addons and Ember Observer attempt to lessen the information overload by categorizing and scoring addons.
+However, it is still difficult to identify addons that are outdated or addons that only work past a certain version of Ember, for example.
+This RFC details a proposal to standardize tagging of addon compatibility by their authors, so that services such as Ember Addons or Ember Observer
+can consume that information to surface to developers in an easy-to-understand format. This could also enable features such as Ember-CLI warning users
+when they attempt to `ember install` a package that is listed as incompatible with the core package versions of their application.
+
+# Detailed design
+
+This is accomplished by adding a `compatibility` tag to the `ember-addon` section of an addon's `package.json`. 
+This tag should contain a hash of package names that are associated with range strings identifying the compatible versions.
+Version ranges should be formatted as outlined by [Node-SemVer](https://github.com/npm/node-semver).
+ 
+Example `package.json`:
+
+```json
+{
+    "name": "ember-my-addon",
+    "version": "0.0.1",
+    "description": "My Ember Addon",
+    "keywords": [
+        "ember-addon"
+    ],
+    "ember-addon": {
+        "configPath": "tests/dummy/config",
+        "demoURL": "username.github.io/ember-my-addon",
+        "compatibility": {
+            "ember": "^1.12.0",
+            "ember-cli": "1.13.1 - 1.13.8",
+            "ember-data": "1.0.0-beta.18"
+        }
+    }
+}
+```
+
+The Ember-CLI blueprint for new addons could possibly include some default compatibility that 
+appropriately indicates the compatibility of a the current Ember releases with other versions.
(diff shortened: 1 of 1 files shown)


JKGisMe:
I like it. It doesn't seem like it would be a super hard thing to implement. Just providing the step of 

"Hey this addon you want to `ember install` is known to be compatible with Ember versions x.x - x.x but this project is using y.y. Are you sure you want to continue?" 

would give users a heads up to check things out further. And users complaining to addon maintainers might be a bit of motivation for them to keep the version numbers up to date.

mike-north:
Nice work Jordan!

Any interest in handling both bower & npm?

jalcine:
[**@mike-north**](https://github.com/mike-north) commented on [Sep 6, 2015, 8:53 PM EDT](https://github.com/emberjs/rfcs/pull/88#issuecomment-138146928 "2015-09-07T00:53:36Z - Replied by Github Comment Enhancer"):
> Nice work Jordan!
> 
> Any interest in handling both bower &amp; npm?

This would be _tremendously_ helpful, as well as compatible versions of Node. However, thinking on that; could the Node versions and NPM versions be extracted from the module's `package.json`?

dukex:
I have one stupid question, Will be necessary for the addon developer make a release or some like that only to tell his addon is compatibility with a new ember version?

mike-north:
> could the Node versions and NPM versions be extracted from the module's package.json

Not reliably. The addon may be using a particular version of a dependency, but other versions may be compatible (and supported). For example, you're using a particular version of ember-cli, but it's likely that some set of older versions will work just fine.

jalcine:
@dukex: No question's stupid :). And tbh, I'm not qualified to answer this as I don't know how `ember cli` discovers add-ons (unless it's just doing an `npm install` under the hood). If it's an `npm install`; then a patch release would be necessary (to push out the necessary meta-data information).

jalcine:
@mike-north Gotcha.

elwayman02:
@mike-north in my mind you would put both bower and npm dependencies in this list, since the addon's `package.json` is the primary point that services would use to consume data about it.  Bower dependencies are usually added via blueprints, of course, and other dependencies tend to be historically harder to feel out since the `bower.json` isn't used for installing the addon in a consuming application. I am not sure of the potential drawbacks of combining both bower and npm packages in the same `compatibility` list under `ember-addon`, but perhaps there could be two sections for `npmCompatibility` and `bowerCompatibility`? I suppose this depends a lot on how Ember-CLI, EmberAddons, EmberObserver and others would intend to use such information.

@dukex yes, an addon developer would need to add this metadata into a new npm release in order for it to be picked up by consuming services, because each npm release only has the package.json that was released with, and wouldn't have the changes that add this meta until another release was cut. That said, it could be added as its own patch release relatively easily, or it could simply be included with the next release of new features/bugfixes.

JKGisMe:
Yeah, keeping a compatibility version updated when new ember versions come out is important. Part of the existing problem is that addon maintainers say things like "compatible with ember 1.8+" in the readme. Then users can't tell if it's _still_ true or if it was only true at the time they wrote the readme.

elwayman02:
@mike-north here is an example of how one Addon author has handled bower version compatibility:

https://github.com/jamesarosen/ember-i18n/blob/master/bower.json

text/0000-addon-version-compatibility.md
```diff
@@ -0,0 +1,79 @@
...
+The usage of this feature is not restricted to core Ember libraries, obviously. 
+Addon authors could use it to mark other dependencies which they have strict compability with.
+
+# Drawbacks
+
+This is still a manual process depending on other maintainers, and if Ember-CLI adds this to the default blueprint, 
+we may see a slew of addons that use the default but never get updated as things become incompatible.
+
+# Alternatives
+
```
stefanpenner: I believe the plan (and my current thought) is for us to use `peerDependencies` and make `ember install` more intelligent to resolve & download, or detect conflict and report.

Adding an additional concept like `compatibility` should be reserve if ^^ doesn't work for some reason. So far no blockers have been noticed.

text/0000-addon-version-compatibility.md
```diff
@@ -0,0 +1,79 @@
...
+appropriately indicates the compatibility of a the current Ember releases with other versions.
+Authors of addons should then update their package.json to reflect the compatibility of their package.
+
+The usage of this feature is not restricted to core Ember libraries, obviously. 
+Addon authors could use it to mark other dependencies which they have strict compability with.
+
+# Drawbacks
+
+This is still a manual process depending on other maintainers, and if Ember-CLI adds this to the default blueprint, 
+we may see a slew of addons that use the default but never get updated as things become incompatible.
```
Turbo87: Maybe it would make more sense to add a "Compatibility" section to the default addon blueprint instead?

PR #91, opened 2015-09-12
WeakMap RFC

stefanpenner:
[rendered](https://github.com/emberjs/rfcs/blob/weakmap/text/0066-template.md)

### DIFF
--- a/text/0066-template.md
+++ b/text/0066-template.md
@@ -34,6 +34,11 @@ listeners/caches/chains/descriptors etc. Unfortunately today, there is no
 public API for apps or addons to utilize this. `Ember.WeakMap` aims to be
 exactly that API.
 
+Some examples:
+
+* https://github.com/offirgolan/ember-cp-validations/blob/master/addon/utils/cycle-breaker.js
+* https://github.com/stefanpenner/ember-state-services/ (will soon utilize the user-land polyfil of this) to prevent common leaks.
+
 # Detailed design
 
 ## Public API
--- a/text/0066-template.md
+++ b/text/0066-template.md
@@ -80,6 +80,7 @@ in the key objects meta weak bucket. This will allow one object to belong in
 multiple weakMaps without chance of collision.
 
 Concrete Implementation: https://github.com/emberjs/ember.js/pull/12224
+Polyfill: https://www.npmjs.com/package/ember-weakmap
 
 # Drawbacks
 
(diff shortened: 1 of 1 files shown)


mixonic:
@stefanpenner This suggests that the Ember WeakMap would not support map -> key weakness. Correct?

If so, `destroy` on the weak map would still remove the data for all current members from memory?

stefanpenner:
Destroying the weakmap itself would have no affect. The data would just become inaccessible and require the keys be reclaimed before

stefanpenner:
In this case the WM is basically just a way to store private state on an object without polluting

mixonic:
"Motivation" should likely include the specific Ember features this would be used for. To quote Stef via chat:

* Fixing computed sort nicely
* Potential future feature of runtime DK for cp

There are some addons that would use this:

* ember-state-service
* ember-cp-validations

Though those could also consume WeakMap as a stand-alone library instead. The stand-alone version would be dependent on Ember `meta` however, which would definitely be reaching into something private. Seems like it should be mentioned under "Alternatives" (not that I'm advocating for this alternative).

stefanpenner:
Actually it isn't that meta is created out of band that is the problem. We can trigger its creation from userland. The issue is more of shaping. We can't ensure meta shapes correctly with this new slot. Unless we can insert the slot in the constructor of meta.

In theory run times should be able to adapt to afew dynamic slots added. In reality, we blow our budget of such flexibility far to easily.

We will publish a polyfil addon, so older code can use it. Unfortunately that addon will be stuck causing meta to have different slots

jasonmit:
@stefanpenner based on the WeakMap spec, `delete` returns a boolean on whether or not the delete found anything.  In your proposal, it returns an instance of the weakmap.  Just curious if intentional or does the RFC need to be updated.

Related: https://github.com/thoov/ember-weakmap/pull/3

stefanpenner:
we should follow the spec https://tc39.github.io/ecma262/#sec-weakmap.prototype.delete so returning true/false

stefanpenner:
looks like the one in ember does this: https://github.com/emberjs/ember.js/blob/master/packages/ember-metal/lib/weak_map.js#L89-L101

stefanpenner:
This RFC is still required to eventually make it public, as i believe the built-in weakmap is still private.

mmun:
Looks good :+1:

stefanpenner:
next step: https://github.com/emberjs/ember.js/issues/13668

PR #92, opened 2015-09-13
`url-for` helper

GavinJoyce:
[Link to rendered](https://github.com/GavinJoyce/rfcs/blob/gj/href-to/text/0000-url-for.md)

### DIFF
--- a/text/0000-url-for.md
+++ b/text/0000-url-for.md
@@ -1,0 +1,47 @@
+- Start Date: 2015-09-12
+- RFC PR: (leave this empty)
+- Ember Issue: (leave this empty)
+
+# Summary
+
+This RFC proposes the extraction of a new `url-for` helper from `link-to`.
+
+# Motivation
+
+Decomposing `link-to` into a set of helpers will allow us to simplify its internals while providing new low level primitives for use in performance sensitive applications. `url-for`, the first of these extracted helpers, will allow for simple URL rendering in an Ember application. Extracting separate helpers for dealing with link classes will be suggested in a future RFC.
+
+This proposal would also allow plain URLs to be used within an Ember application without triggering an application reload. For example, a url such as `/about` could be used within a template or helper.
+
+# Detailed design
+
+The `url-for` helper would provide a similar interface to the `link-to` helper in block form:
+
+```
+{{#link-to 'photo.comment' 5 primaryComment}}
+  Main Comment for the Next Photo
+{{/link-to}}
+```
+
+The `url-for` equivalent would be:
+
+```
+<a href="{{url-for 'photo.comment' 5 primaryComment}}">
+  Main Comment for the Next Photo
+</a>
+```
+
+The `url-for` helper would simply render the URL, no component is needed.
+
+An additional click handler would be setup in [EventDispatcher](https://github.com/emberjs/ember.js/blob/master/packages/ember-views/lib/system/event_dispatcher.js) to capture any clicks on anchors within the ember application. Any anchor URL recognized by the router will be handled by it.
+
+# Drawbacks
+
+This expands the API surface area of Ember.js by introducing a new helper. One additional global event handler would be required.
+
+# Alternatives
+
+The existing [`href-to` addon](https://github.com/intercom/ember-href-to) provides this capability and applications can currently opt-in if they desire.
+
+# Unresolved questions
+
+Is there a better name than `url-to`? What other features of `link-to` should we also support (eg. `replace=true`)?


rwjblue:
:+1: - I personally prefer 'url-for' (because this is useful for more things than just hrefs). 

I'd also like to see a 'url-is-active' helper for similar reasons (it could fit in this RFC, but I'm fine with it being a separate one also).

rwjblue:
Also, I'd like to see Ember.LinkComponent be changed around to provide all of its functionality via public template helpers (possibly when it is implemented as a glimmer component). This 'url-for'/'href-to' helper is a great start towards that goal...

GavinJoyce:
Thanks, I agree that `url-for` is a better name

text/0000-href-to.md
```diff
@@ -0,0 +1,47 @@
...
+- RFC PR: (leave this empty)
+- Ember Issue: (leave this empty)
+
+# Summary
+
+This RFC proposes the addition of a `href-to` helper which is a lightweight alternative to the `link-to` helper.
+
+# Motivation
+
+For performance sensitive applications like [Intercom](https://www.intercom.io/), `link-to` can be quite heavy. We've found many cases where all we want is to link to a route. `link-to` creates a component with a set of attribute and class bindings which we don't need. In contrast, the `href-to` helper would simply render a URL at render time.
```
stefanpenner: wouldn't the solution to ^ be to make link-to not slow?

text/0000-href-to.md
```diff
@@ -0,0 +1,47 @@
...
+
+In addition, this proposal would allow plain URLs to be used within an Ember application without triggering an application reload.
```
stefanpenner: This is more of an appealing motivation. link-to should be made of compositional units, that can be used to construct link-to itself and other new and novel things.

text/0000-href-to.md
GavinJoyce: Yeah, the faster that `link-to` is, the less of a need for something like this. We've found the `href-to` addon to be 6-12x times faster in Ember 1.11-1.13 though. `link-to` will always have much more to do, all `href-to` has to do is to generate a URL.

text/0000-href-to.md
```diff
@@ -0,0 +1,47 @@
+- Start Date: 2015-09-12
+- RFC PR: (leave this empty)
+- Ember Issue: (leave this empty)
+
+# Summary
+
+This RFC proposes the addition of a `href-to` helper which is a lightweight alternative to the `link-to` helper.
```
stefanpenner: i believe this may want to be named `url-for`


given:
```js
<a href=url ></a>
```

my interpretation is: 

`a` has a attribute (or property) named `href` that takes a URL as a value.

In other words, the value of the hypertext reference is the provided url.

text/0000-href-to.md
```diff
@@ -0,0 +1,47 @@
...
+</a>
+```
+
+The `href-to` helper would simply render the URL, no component is needed.
+
+An additional click handler would be setup in [EventDispatcher](https://github.com/emberjs/ember.js/blob/master/packages/ember-views/lib/system/event_dispatcher.js) to capture any clicks on anchors within the ember application. Any anchor URL recognized by the router will be handled by it.
+
+# Drawbacks
+
+This expands the API surface area of Ember.js by introducing a new helper. One additional global event handler would be required.
```
stefanpenner: i think this is fine.

stefanpenner:
I suspect a sibling RFC (or maybe part of this one) should explore the extracting other parts of link to. Such as the classes that are associated with  a URL. active/loading/current etc.

With that in-place, link-to may easily just become a sugaring of its component parts. It could even become just an HTMLBars transform, with nearly no overhead.

text/0000-href-to.md
GavinJoyce: :+1: @rwjblue had the same name suggestion

text/0000-href-to.md
rwjblue: We should definitely make link-to faster, but the two ideas are not opposed to each other. For example, one way to make it faster is to have it take advantage of template helpers like this one.

stefanpenner: my point was, we don't normally except joe turbo features, when the core primitive is still more or less unoptimized. So that being a primary motivation may give the wrong impression.

As i mentioned bellow, the main motivation should be to desugar link-to, this will:

* prevent link-to from continuing to be a junk drawer
* allow us to use new templating features to make the component parts faster
* enable more experimentation

GavinJoyce: I'll rewrite the motivation section to emphasise these points. Thanks

GavinJoyce:
I've renamed the helper to `url-for` and rewrote the motivations sections to better capture the motivations that you both suggested:

```
Decomposing `link-to` into a set of helpers will allow us to simplify its internals while providing new low level primitives for use in performance sensitive applications. `url-for`, the first of these extracted helpers, will allow for simple URL rendering in an Ember application. Extracting separate helpers for dealing with link classes will be suggested in a future RFC.

This proposal would allow plain URLs to be used within an Ember application without triggering an application reload. A relative url such as `/about` could be used within a template or helper.
```

In order to keep this small, I've kept this focused on `url-for`. The other class helpers can follow once we've got agreement on this one.

text/0000-url-for.md
```diff
@@ -0,0 +1,47 @@
...
+
+# Summary
+
+This RFC proposes the extraction of a new `url-for` helper from `link-to`.
+
+# Motivation
+
+Decomposing `link-to` into a set of helpers will allow us to simplify its internals while providing new low level primitives for use in performance sensitive applications. `url-for`, the first of these extracted helpers, will allow for simple URL rendering in an Ember application. Extracting separate helpers for dealing with link classes will be suggested in a future RFC.
+
+This proposal would also allow plain URLs to be used within an Ember application without triggering an application reload. For example, a relative url such as `/about` could be used within a template or helper.
```
rwjblue: Slight nit-pick: I personally wouldn't call this a relative url since it includes a leading slash.

GavinJoyce: thanks

stefanpenner:
> The other class helpers can follow once we've got agreement on this one.

:+1:

tomdale:
Awesome first step. My feeling is that this functionality should likely be specced out as part of a `routing` service, and the `url-for` helper should just be a small wrapper exposing that to Handlebars-land. Perhaps the first iteration of the public `routing` service just exposes `urlFor` and `transitionTo` methods and we can build from there.

chadhietala:
This will make links within i18n strings much easier so :+1: . For example we use a i18n string format based on Java xmsg (Yahoo uses [something similar](http://formatjs.io/guides/message-syntax/) but standardized) and our helper looks something like.

```
<!-- define the string -->
{{t-def 'hello_world' text='Hello {Hello :name, click this <a href={:link}>Link</a>'}}

<!-- Use string -->
<h1>{{t 'hello_world' name=name link=(url-for 'profile' 123) }}</h1>
```

Definitely a need for this type of primitive.

MiguelMadero:
I was thinking the same. Let's build on top of the routingService.

Turbo87:
has there been any progress on this? I need a similar solution to what @chadhietala described and it would be awesome if ember would provide this out of the box.

rwjblue:
Closing this in favor of #95

PR #94, opened 2015-09-21
RFC: `network` Service

tomdale:
[Rendered](https://github.com/tomdale/rfcs/blob/network-service/text/0000-network-service.md)

### DIFF
--- a/text/0000-network-service.md
+++ b/text/0000-network-service.md
@@ -266,7 +266,7 @@ exposing jQuery API that users may come to rely on.
 ## Built-in Services
 
 Should built-in services in Ember be namespaced somehow? As we introduce
-new services (`ajax` this case, things like `routing` in the future),
+new services (`network` this case, things like `routing` in the future),
 they may conflict with the names of services in apps or addons.
 Upgrading your app to the new version of Ember that contains the service
 could cause things to break, violating our Semantic Versioning concerns.
@@ -277,9 +277,9 @@ the new system service until they have renamed the existing user-defined
 service.
 
 It is unclear how this would work with addons, however. For example,
-imagine you have an `ajax` service in your app. You upgrade Ember to a
-new version that contains the built-in `ajax` service. You receive a
+imagine you have an `network` service in your app. You upgrade Ember to a
+new version that contains the built-in `network` service. You receive a
 warning about the conflict but choose to temporarily ignore it. Then,
-you install an addon that requires the built-in `ajax`. If it receives
-the user-defined `ajax` service, the app will fail in surprising and
+you install an addon that requires the built-in `network`. If it receives
+the user-defined `network` service, the app will fail in surprising and
 likely hard-to-diagnose ways.
--- a/text/0000-network-service.md
+++ b/text/0000-network-service.md
@@ -1,0 +1,285 @@
+- Start Date: 2015-09-21
+- RFC PR: (leave this empty)
+- Ember Issue: (leave this empty)
+
+# Summary
+
+Introduces a standard way of fetching data over HTTP in Ember by way of
+a built-in network service.
+
+# Motivation
+
+The motivations for introducing a built-in network service are two-fold:
+
+1. New users can fetch data over HTTP performantly out-of-the-box.
+2. It provides a standard way of fetching data on the server-side for
+   things like FastBoot.
+
+Because a route's `model()` hook works by "just returning a promise,"
+many new users to Ember start off by returning the result of
+`jQuery.ajax()` from this hook.
+
+While this works, jQuery's Promises implementation has historically left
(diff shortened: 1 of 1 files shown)


grapho:
This seems interesting. I am one of those users who, while learning, started with jquery-ajax and then _after_ it started to fail on me, switched to ic-ajax.  I have felt for a while that Ember should have its own simple built-in ajax-like function that actually fully works with RSVP.

tomdale:
@praxxis points out some prior art: https://github.com/ember-cli/ember-ajax

jcope2013:
throwing this discussion in here also, a con of fetch in my opinion is that fetch resolves successfully on 4xx and 5xx errors, but $.ajax rejects. https://github.com/ember-cli/rfcs/issues/19

taras:
@tomdale you can see the discussion on this [`ember-cli/rcfs` PR that lead to `ember-ajax` being added to Ember CLI organization](https://github.com/ember-cli/rfcs/issues/19#issuecomment-137952616).

mmun:
@jcope2013 I disagree that that is a con. The rejection state of a promise is only for exceptional cases (e.g. a network timeout). You should really think of it like a async try-catch construct.

oriSomething:
i don't think it suppose to be in the core simply because many time as surprise it might sound, i don't use `ajax` / `fetch` in project and i'm sure may don't use either.
but anyway, `fetch` doesn't support `abort` because there is still no **cancelable promises**, so until `fetch` doesn't allow this option, `ajax` wrapper is still preferable

stefanpenner:
For those interested in easily using fetch in ember apps today, checkout http://http://github.com/stefanpenner/ember-fetch

@tomdale. Maybe I missed it, but fetch and Ajax have different rejection semantics. Fetch does not reject on 500 rather it fulfills with a request object containing the status and errors. If we where to support fetch out of the box, without confusing people we may want to explore different defaults for the model hook. 

This confusion is actually why I didn't serious ember-Cli over to ember-fetch yet. In the rfc I opened, this ended up being a big community concern.

I suspect with some massaging the situation can be made much nice. And potentially nicer then what we have today.

ilkkao:
It'd be nice to have a some kind of option to opt-out from this. For example in case the app communicates with server only through socket.io.

stefanpenner:
> It'd be nice to have a some kind of option to opt-out from this. For example in case the app communicates with server only through socket.io.

thats covered in the proposal

jonnii:
If bloat is a concern perhaps the `fetch` and `ajax` functionality can be split into two, with `fetch` shipping out of the box (and being the blessed path in guides etc..) and the ajax shim being an addon? FWIW I (personally) think there are enough compelling reasons to have a network service available in core.

josemarluedke:
Github has a lib for the [fetch polyfill](https://github.com/github/fetch), maybe it's useable if is decided to implement this proposal.

jamesarosen:
> throwing this discussion in here also, a con of fetch in my opinion is that fetch resolves successfully on 4xx and 5xx errors, but $.ajax rejects. -@jcope2013

> I disagree that that is a con. The rejection state of a promise is only for exceptional cases (e.g. a network timeout). You should really think of rejection an an async try-catch construct. -@mmun

One somewhat sane possibility would be to have `network.fetch` obey `fetch()` semantics and `network.ajax` obey `$.ajax` semantics.

MiguelMadero:
@tomdale I share the concerns about the name conflicts. We already have a service and routing service. For routing is different and probably need the warnings or namespaces, but I'm wondering if this needs to be a service, since it doesn't hold to any state, can we just simply import it like `ic-ajax` today instead of "inject"ing it.

Binarytales:
This RFC was pretty eye opening to me as someone who is relatively new to Ember. I have recently started using jQuery ajax where it sounds like I should be using `ic-ajax`.

It seemed odd to me that Ember did not already include this functionality in core.

I had a thought about addressing the size drawback but perhaps I'm not fully understanding how Ember depends on jQuery. However, my thought was, if Ember has an ajax "light" implementation in core then perhaps Ember could depend on a selective build of jQuery that does _not_ include it's full implementation of ajax. Instead jQuery's full ajax implementation could be the part that's available as an optional addon to developers that require it.

tomdale:
For the moment, I'm retracting this RFC so that we can iterate a few ideas in the addon space. We're exploring whether the [ember-cli/ember-ajax addon](https://github.com/ember-cli/ember-ajax) can be used to solve the same set of problems without adding additional functionality to Ember core.

tomdale:
Wrote up some of my thinking about how the addon ecosystem might look to address a similar issue: https://gist.github.com/tomdale/203ea125acd0be8b5f1f

PR #95, opened 2015-09-24
Create Router Service

ef4:
[Rendered](https://github.com/emberjs/rfcs/blob/router-service/text/0000-router-service.md)

### DIFF
--- a/text/0000-router-service.md
+++ b/text/0000-router-service.md
@@ -131,10 +131,6 @@ A `url-for` helper can be implemented almost identically to the `is-active` exam
 `currentURL`: provides the serialized string representing `currentRoute`.
 
 
-### Deprecation
-
-I propose deprecating the publicly extensible `willTransition` and `didTransition` hooks. They are redundant with an observable `currentRoute`, and the arguments they receive leak internal implemetation from router.js.
-
 ## RouteInfo Type
 
 A RouteInfo object has the following properties. They are all read-only.
@@ -155,7 +151,7 @@ router.currentRoute.find(info => info.name === 'people').params
 
 ## Transition Object
 
-A `transition` argument is passed to `Route`'s `beforeModel`, `model`, `afterModel`, and `willTransition` hooks. Today it's public API is only really `abort()` and `retry()`.
+A `transition` argument is passed to `Route#beforeModel`, `Route#model`, `Route#afterModel`, `Route#willTransition`, and `Router#willTransition`. Today `transition`'s public API is only really `abort()` and `retry()`.
 
 ### New Properties: `from` and `to`
 
@@ -206,7 +202,7 @@ This RFC provides public API for doing the things people have become accustomed 
 
 # Drawbacks
 
-This RFC suggests only two small deprecations that are unlikely to effect many apps, so the API-churn burden may appear low. However, we know that use of the private APIs we're deliberately disabling is widespread, so users will experience churn. We can provide our usual deprecation cycle to give them early warning, but it still imposes some cost.
+This RFC doesn't deprecate any public API, so the API-churn burden may appear low. However, we know that use of the private APIs we're deliberately disabling is widespread, so users will experience churn. We can provide our usual deprecation cycle to give them early warning, but it still imposes some cost.
 
 This RFC doesn't attempt to change the existing and fairly rich semantics for initiating transitions. For example, you can pass either models or IDs, and those have subtle semantic differences. I think an ideal rewrite would also change the semantics of the route hooks and transitionTo to simplify that area.
 
--- a/text/0000-router-service.md
+++ b/text/0000-router-service.md
@@ -379,6 +379,7 @@ A more complete version that also matches models and queryParams can be written 
     {{outlet}}
   {{/with-dynamic-variable}}
 {{/liquid-bind}}
+```
 
 4. Prerendering of non-current routes becomes possible. You can use `recognizeAndLoad` to obtain a `RouteInfoWithAttributes` and then use `{{#with-dynamic-variable "routeInfo" myRouteInfo}} {{outlet}} {{/with-dynamic-variable}}` to render it.
 
--- a/text/0000-router-service.md
+++ b/text/0000-router-service.md
@@ -141,7 +141,7 @@ A RouteInfo object has the following properties. They are all read-only.
 
  - name: the dot-separated, fully-qualified name of this route, like `"people.index"`.
  - localName: the final part of the `name`, like `"index"`.
- - attrs: the attributes provided to this route's component, if any, like `{ model }`. For old-style routes that have a controller/view/template instead of a routable component, the implicit `attrs` shall contain `{ model, controller, view }`.
+ - attrs: the values that were resolved by this `Route`. Today, that means `{ model, controller }`. In the future it may become the attributes passed directly to a "routeable component".
  - params: the values of this route's parameters. Same as the argument to `Route`'s `model` hook. Contains only the parameters valid for this route, if any (params for parent or child routes are not merged).
  - queryParams: the values of any queryParams on this route.
  - parent: another RouteInfo instance, describing this route's parent route, if any.
--- a/text/0000-router-service.md
+++ b/text/0000-router-service.md
@@ -148,11 +148,48 @@ Example: this feature can replace [this use of private API in ember-href-to](htt
 
 Takes a string URL and returns a promise that resolves to a `RouteInfoWithAttributes` for the leafmost route represented by the URL. The promise rejects if the URL is not recognized or an unhandled exception is encountered. This method expects to receive the actual URL as seen by the browser _including_ the app's `rootURL`.
 
-### Improved Event Coverage
+### Deprecating willTransition and didTransition
 
 Application-wide transition monitoring events belong on the Router service, not spread throughout the Route classes. That is the reason for the existing `willTransition` and `didTransition` hooks/events on the Router. But they are not sufficient to capture all the detail people need. See for example, https://github.com/nickiaconis/rfcs/blob/1bd98ec534441a38f62a48599ffa8a63551b785f/text/0000-transition-hooks-events.md
 
-I do not believe an additional event is needed, we just need to ensure that the existing two events fire at every appropriate time. `willTransition` should fire _whenever_ the destination route changes, and that includes redirects, error handling, and loading states. A long chain of redirects should cause a sequence of `willTransition` events and then a final `didTransition` event.
+In addition, they receive handlerInfos in their arguments, which are an undocumented internal implementation detail of router.js that doesn't belong in Ember's public API. Everything you can do with handlerInfos can be done with the RouteInfo type that is proposed in this RFC, with the benefit of sticking to supported public API.
+
+So we should deprecate willTransition and didTransition in favor of the following new events.
+
+### New Events: routeWillChange & routeDidChange
+
+The `routeWillChange` event fires whenever a new route is chosen as the desired target of a transition. This includes `transitionTo`, `replaceWith`, all redirection for any reason including error handling, and abort. Aborting implies changing the desired target back to where you already were. Once a transition has completed, `routeDidChange` fires. 
+
+Both events receive a single `transition` argument as described in the "Transition Object" section below, which explains the meaning of `from` and `to` in more detail.
+
+Redirection example:
+
+ 1. current route is A
+ 2. user initiates a transition to B
+ 3. routeWillChange fires `from` A `to` B.
+ 4. B redirects to C
+ 5. routeWillChange fires `from` A `to` C.
+ 6. routeDidChange fires `from` A `to` C.
+
+Abort example:
+
+ 1. current route is A
+ 2. user initiates a transition to B
+ 3. routeWillChange fires `from` A `to` B.
+ 4. in response to the previous routeWillChange event, the transition is aborted.
+ 5. routeWillChange fires `from` A `to` A.
+ 8. routeDidChange fires `from` A `to` A.
+
+Error example:
+
+ 1. current route is A
+ 2. user initiates a transition to B.index
+ 3. routeWillChange fires `from` A `to` B.
+ 4. B throws an exception, and the router discovers a "B-error" template.
+ 5. routeWillChange fires `from` A `to` B-error
+ 6. routeDidChange fires `from` A `to` B-error
+
+These are events, not extension hooks -- now that we are exposing a Service, it makes more sense to subscribe to its events than extend it.
 
 ### New Properties
 
@@ -348,7 +385,7 @@ A more complete version that also matches models and queryParams can be written 
 
 # Drawbacks
 
-This RFC doesn't deprecate any public API, so the API-churn burden may appear low. However, we know that use of the private APIs we're deliberately disabling is widespread, so users will experience churn. We can provide our usual deprecation cycle to give them early warning, but it still imposes some cost.
+This RFC deprecates only two public extension hooks API, so the API-churn burden may appear low. However, we know that use of the private APIs we're deliberately disabling is widespread, so users will experience churn. We can provide our usual deprecation cycle to give them early warning, but it still imposes some cost.
 
 This RFC doesn't attempt to change the existing and fairly rich semantics for initiating transitions. For example, you can pass either models or IDs, and those have subtle semantic differences. I think an ideal rewrite would also change the semantics of the route hooks and transitionTo to simplify that area.
 
@@ -400,6 +437,3 @@ Should we introduce new API via the `Ember` global and switch to a module export
 
 They let you do spooky-action-at-a-distance stuff. We could instad choose to implement single-purpose methods like `(get-route-info)` and `{{#with-route-info someValue}}` that would solve the routing case without opening the door to arbitrary user-defined ones.
 
-## Possible semantic breakage of willTransition
-
-I proposed fixing willTransition to ensure that it fires during redirects. I see that as a clear bugfix, but if there is reason to believe it will break people's apps then we should deprecate `willTransition` and introduce a new event with the correct semantics.
--- a/text/0000-router-service.md
+++ b/text/0000-router-service.md
@@ -121,6 +121,13 @@ This takes the same arguments as `transitionTo`, but instead of initiating the t
 
 A `url-for` helper can be implemented almost identically to the `is-active` example above.
 
+### New Method: URL recognition
+
+`recognize(url)`
+
+Takes a string URL (relative to the application's baseURL) and returns a `RouteInfo` for the leafmost route represented by the URL. Returns `null` if the URL is not recognized.
+
+Example: this feature can replace [this use of private API in ember-href-to](https://github.com/intercom/ember-href-to/blob/b8cf0699eec6a65570b05e4fc22b27d8cea49c42/app/instance-initializers/browser/ember-href-to.js#L34).
 
 ### New Properties
 
--- a/text/0000-router-service.md
+++ b/text/0000-router-service.md
@@ -435,5 +435,5 @@ Should we introduce new API via the `Ember` global and switch to a module export
 
 ## Generic dynamically-scoped variables may be too powerful
 
-They let you do spooky-action-at-a-distance stuff. We could instad choose to implement single-purpose methods like `(get-route-info)` and `{{#with-route-info someValue}}` that would solve the routing case without opening the door to arbitrary user-defined ones.
+They let you do spooky-action-at-a-distance stuff. We could instead choose to implement single-purpose methods like `(get-route-info)` and `{{#with-route-info someValue}}` that would solve the routing case without opening the door to arbitrary user-defined ones.
 
--- a/text/0000-router-service.md
+++ b/text/0000-router-service.md
@@ -94,7 +94,7 @@ export default Helper.extend({
     }
     return this.get('router').isActive(routeName, ...models, hash.queryParams);
   },
-  observer('router.currentRoute', function() {
+  didTransition: observer('router.currentRoute', function() {
     this.recompute();
   });
 });
--- a/text/0000-router-service.md
+++ b/text/0000-router-service.md
@@ -10,6 +10,8 @@ This RFC proposes:
  
  - codifying and expanding the supported public API for the `transition` object that is currently passed to `Route` hooks.
 
+ - introducing the concept of dynamically-scoped routing info.
+
 These topics are closely related because they share a unified `RouteInfo` type, which will be described in detail.
 
 # Motivation
@@ -19,6 +21,16 @@ Given the modern Ember concepts of Components and Services, it is clear that rou
 The immediate benefit of having a `RouterService` is that you can inject it into components, giving them a friendly way to initiate transitions and ask questions about the current global router state.
 
 A second benefit is that we have the opportunity to add new capabilities to the `RouterService` to replace several common patterns in the wild that dive into private internals in order to get things done. There are several places where we leak internals from router.js, and we can plug those leaks.
+
+A `RouterService` is great for asking global questions, but some questions are not global and today we incur complexity by treating them as if they are. For example:
+
+ - `{{link-to}}` can use implicit models from its context, but that breaks when you're trying to animate to or from a state where those models are not present.
+
+ - `{{link-to}}` has a lot of complexity and performance cost that deals with changing its active state, and the precise timing of when that should happen.
+
+ - there is no way to ask the router what it would do to handle a given URL without actually visting that URL.
+
+All of the above can be addressed by embracing what is already internally true: "the current route" is not a single global, it's a dynamically-scoped variable that can have different values in different parts of the application simultaneously.
 
 # Detailed design
 
@@ -67,7 +79,7 @@ That API will be carried over verbatim to `RouterService`, and the publicly acce
 
 ```js
 transitionTo(routeName, ...models, queryParams)
-replaceWith(routeName, ...models, queryParms)
+replaceWith(routeName, ...models, queryParams)
 ```
 
 These two have the same semantics as the existing methods on `Ember.Route`:
@@ -112,12 +124,11 @@ export default Helper.extend({
 
 ```
 
-
 ### New Method: URL generation
 
 `urlFor(routeName, ...models, queryParams)`
 
-This takes the same arguments as `transitionTo`, but instead of initiating the transition it returns the resulting URL as a string.
+This takes the same arguments as `transitionTo`, but instead of initiating the transition it returns the resulting root-relative URL as a string (which will include the application's `rootUrl`).
 
 A `url-for` helper can be implemented almost identically to the `is-active` example above.
 
@@ -125,9 +136,23 @@ A `url-for` helper can be implemented almost identically to the `is-active` exam
 
 `recognize(url)`
 
-Takes a string URL (relative to the application's baseURL) and returns a `RouteInfo` for the leafmost route represented by the URL. Returns `null` if the URL is not recognized.
+Takes a string URL and returns a `RouteInfo` for the leafmost route represented by the URL. Returns `null` if the URL is not recognized. This method expects to receive the actual URL as seen by the browser _including_ the app's `rootURL`.
+
 
 Example: this feature can replace [this use of private API in ember-href-to](https://github.com/intercom/ember-href-to/blob/b8cf0699eec6a65570b05e4fc22b27d8cea49c42/app/instance-initializers/browser/ember-href-to.js#L34).
+
+
+### New Method: Recognize and load models
+
+`recognizeAndLoad(url)`
+
+Takes a string URL and returns a promise that resolves to a `RouteInfoWithAttributes` for the leafmost route represented by the URL. The promise rejects if the URL is not recognized or an unhandled exception is encountered. This method expects to receive the actual URL as seen by the browser _including_ the app's `rootURL`.
+
+### Improved Event Coverage
+
+Application-wide transition monitoring events belong on the Router service, not spread throughout the Route classes. That is the reason for the existing `willTransition` and `didTransition` hooks/events on the Router. But they are not sufficient to capture all the detail people need. See for example, https://github.com/nickiaconis/rfcs/blob/1bd98ec534441a38f62a48599ffa8a63551b785f/text/0000-transition-hooks-events.md
+
+I do not believe an additional event is needed, we just need to ensure that the existing two events fire at every appropriate time. `willTransition` should fire _whenever_ the destination route changes, and that includes redirects, error handling, and loading states. A long chain of redirects should cause a sequence of `willTransition` events and then a final `didTransition` event.
 
 ### New Properties
 
@@ -136,6 +161,19 @@ Example: this feature can replace [this use of private API in ember-href-to](htt
 `currentRouteName`:  a convenient alias for `currentRoute.name`.
 
 `currentURL`: provides the serialized string representing `currentRoute`.
+
+### Query Parameter Semantics
+
+Today, `queryParams` impose unnecessarily high cost because we cannot generate URLs or determine if a link is active without taking into account the default values of query parameters. Determining their default values is expensive, because it involves instantiating the corresponding controller, even in cases where we will never visit its route.
+
+Therefore, the `queryParams` argument to the new `urlFor`, `transitionTo`, `replaceWith`, and `isActive` methods defined in this document will behave differently.
+
+ - default values will not be stripped from generated URLs. For example, `urlFor('my-route', { sortBy: 'title' })` will always include `?sortBy=title`, whether or not `title` is the default value of `sortBy`.
+
+ - to explicitly unset a query parameter, you can pass the symbol `Ember.DEFAULT_VALUE` as its value. For example, `transitionTo('my-route', { sortBy: Ember.DEFAULT_VALUE })` will result in a URL that does not contain any `?sortBy=`.
+
+(Sticky parameters are still allowed, because they only apply when the destination controller has already been instantiated anyway.)
+
 
 
 ## RouteInfo Type
@@ -154,6 +192,10 @@ Notice that the `parent` and `child` properties cause `RouteInfos` to form a lin
 ```js
 router.currentRoute.find(info => info.name === 'people').params
 ```
+
+## RouteInfoWithAttributes
+
+This type is almost identical to `RouteInfo`, except it has one additional proprty named `attributes`. The attributes contain the data that was loaded for this route, which is typically just `{ model }`.
 
 ## Transition Object
 
@@ -212,6 +254,98 @@ Some of the private APIs we should mark and warn include:
  - `lookup('router:main')` (should use `service:router` instead)
 
 
+## Dynamically-Scoped Variables
+
+The next section introduces general-purpose, dynamically-scoped variables. Why am I doing this in "The Router Service RFC"? Because dynamically-scoped RouteInfo has several nice capabilities that will be introduced in the subsequent section. And once we have one dynamically-scoped variable, we have the option of exposing that capability for other uses.
+
+### In the general case
+
+A [dynamically-scoped variable](https://en.wikipedia.org/wiki/Scope_(computer_science)#Dynamic_scoping) takes its value from its calling context, as opposed to its lexical context. Since dynamically-scoped variables are powerful to the point of potential danger, their syntax is intended to be appropriately verbose.
+
+Reading a dynamically-scoped variable in handlebars:
+
+ ```hbs
+ {{!- retrieve the value of a dynamically scoped variable }}
+ (get-dynamic-variable "myVariableName")
+ ```
+
+Defining a component that reads a dynamically-scoped variable:
+
+```js
+let MyComponent = Ember.Component.extend({
+  didInsertElement() {
+    // Access to the variable here is intended to be indistinguishable
+    // from a normal, explicitly-passed input argument. 
+    doSomethingWith(this.get('myDynamicVariable'));
+  }
+});
+MyComponent.reopenClass({
+  // This is where the specialness happens.
+  getDynamicVariables: [ 'myDynamicVariable' ]
+});
+```
+
+Defining a helper that reads a dynamically-scoped variable:
+
+```js
+let MyHelper = Ember.Helper.extend({
+  compute(params, hash) {
+    return doSomethingWith(hash.myDynamicVariable);
+  }
+});
+MyHelper.reopenClass({
+  getDynamicVariables: [ 'myDynamicVariable' ]
+});
+```
+
+Setting a dynamically-scoped variable:
+
+```hbs
+{{#with-dynamic-variable "myVariableName" someValue}}
+  {{!-
+    within this block AND ALL ITS DESCENDANTS until
+    otherwise overridden by another with-dynamic-variable statement, 
+    `get-dynamic-variable "myVariableName"` returns someValue.
+  -}}
+{{/with-dynamic-variable}}
+```
+
+Note that there is no `set-dynamic-variable`. You can only introduce new scopes, not mutate your containing scope. There is also no way to set a dynamically-scoped variable directly from Javascript -- your component must use a `with-dynamic-variable` block within its handlebars template.
+
+User-defined, dynamically-scoped variables must be pre-declared before first render via `Ember.declareDynamicVariable("myVariableName")`. Attempting to set one that wasn't declared is an error. Attempting to read one that doesn't exist returns `undefined`. Predeclaration is an optimization that allows us to maintain consistent performance within Glimmer.
+
+### The specific case of routeInfo
+
+`routeInfo` is a dynamically-scoped variable provided by Ember as public API. Its value is always a `RouteInfoWithAttributes` object that is correct for the given route context. This enables several nice things, which I will illustrate with examples:
+
+1. Here is a simplified `is-active` helper that will always update at the appropriate time to match exactly what is rendered in the current outlet. It will maintain the correct state even during animations. Instead of injecting the router service, it consumes the `routeInfo` from its containing environment:
+
+```js
+Ember.Helper.extend({
+  compute([routeName], { routeInfo }) {
+    return !!routeInfo.find(info => info.name === routeName);
+  }
+}).reopenClass({
+  getDynamicVariables: ['routeInfo']
+});
+```
+
+A more complete version that also matches models and queryParams can be written in the same way. 
+
+2. We can improve `link-to` so that it always finds implicit model arguments from the local context, rather than trying to locate them on the global router service. This will fix longstanding bugs like https://github.com/ember-animation/liquid-fire/issues/347 and it will make it easier to test components that contain `{{link-to}}`. This would also open the door to relative link-tos.
+
+3. `liquid-outlet` can be implemented entirely via public API. It would become:
+
+```hbs
+{{#liquid-bind (get-dynamic-variable "routeInfo") as |currentRouteInfo|}}
+  {{#with-dynamic-variable "routeInfo" currentRouteInfo}}
+    {{outlet}}
+  {{/with-dynamic-variable}}
+{{/liquid-bind}}
+
+4. Prerendering of non-current routes becomes possible. You can use `recognizeAndLoad` to obtain a `RouteInfoWithAttributes` and then use `{{#with-dynamic-variable "routeInfo" myRouteInfo}} {{outlet}} {{/with-dynamic-variable}}` to render it.
+
+
 # Drawbacks
 
 This RFC doesn't deprecate any public API, so the API-churn burden may appear low. However, we know that use of the private APIs we're deliberately disabling is widespread, so users will experience churn. We can provide our usual deprecation cycle to give them early warning, but it still imposes some cost.
@@ -256,5 +390,16 @@ beforeModel(transition) {
 ```
 Possibly we *want* this to feel awkward because it's a weird thing to do.
 
+## Naming of Ember.DEFAULT_VALUE Symbol
 
+Should we introduce new API via the `Ember` global and switch to a module export once all the rest of Ember does, or should we just start with a module export right now? If so, what module?
 
+    import { DEFAULT_VALUE } from 'ember-routing';
+
+## Generic dynamically-scoped variables may be too powerful
+
+They let you do spooky-action-at-a-distance stuff. We could instad choose to implement single-purpose methods like `(get-route-info)` and `{{#with-route-info someValue}}` that would solve the routing case without opening the door to arbitrary user-defined ones.
+
+## Possible semantic breakage of willTransition
+
+I proposed fixing willTransition to ensure that it fires during redirects. I see that as a clear bugfix, but if there is reason to believe it will break people's apps then we should deprecate `willTransition` and introduce a new event with the correct semantics.
--- a/text/0000-router-service.md
+++ b/text/0000-router-service.md
@@ -220,6 +220,7 @@ A RouteInfo object has the following properties. They are all read-only.
  - name: the dot-separated, fully-qualified name of this route, like `"people.index"`.
  - localName: the final part of the `name`, like `"index"`.
  - params: the values of this route's parameters. Same as the argument to `Route`'s `model` hook. Contains only the parameters valid for this route, if any (params for parent or child routes are not merged).
+ - paramNames: ordered list of the names of the params required for this route. It will contain the same strings as `Object.keys(params)`, but here the order is significant. This allows users to correctly pass params into routes programmatically.
  - queryParams: the values of any queryParams on this route.
  - parent: another RouteInfo instance, describing this route's parent route, if any.
  - child: another RouteInfo instance, describing this route's active child route, if any.
--- a/text/0000-router-service.md
+++ b/text/0000-router-service.md
@@ -144,7 +144,6 @@ A RouteInfo object has the following properties. They are all read-only.
 
  - name: the dot-separated, fully-qualified name of this route, like `"people.index"`.
  - localName: the final part of the `name`, like `"index"`.
- - attrs: the values that were resolved by this `Route`. Today, that means `{ model, controller }`. In the future it may become the attributes passed directly to a "routeable component".
  - params: the values of this route's parameters. Same as the argument to `Route`'s `model` hook. Contains only the parameters valid for this route, if any (params for parent or child routes are not merged).
  - queryParams: the values of any queryParams on this route.
  - parent: another RouteInfo instance, describing this route's parent route, if any.
--- a/text/0000-router-service.md
+++ b/text/0000-router-service.md
@@ -10,7 +10,9 @@ This RFC proposes:
  
  - codifying and expanding the supported public API for the `transition` object that is currently passed to `Route` hooks.
 
- - introducing the concept of dynamically-scoped routing info.
+ - introducing the `get-route-info` template helper
+ - introducing the `#with-route-info` template keyword
+ - introducing the `readsRouteInfo` static property on `Component` and `Helper`.
 
 These topics are closely related because they share a unified `RouteInfo` type, which will be described in detail.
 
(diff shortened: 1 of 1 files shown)


text/0000-router-service.md
```diff
@@ -0,0 +1,252 @@
...
+
+### Deprecation
+
+I propose deprecating the publicly extensible `willTransition` and `didTransition` hooks. They are redundant with an observable `currentRoute`, and the arguments they receive leak internal implemetation from router.js.
+
+## RouteInfo Type
+
+A RouteInfo object has the following properties. They are all read-only.
+
+ - name: the dot-separated, fully-qualified name of this route, like `"people.index"`.
```
chadhietala: Having the name is great but do you think it would be worth giving the instance or the class as well?  I would imagine you could use ember with something that allows you to specify "query fragments" (Falcor) in your routes and potentially components if you have a way of collecting that fragments recursively.

For example you could imagine that in node you could do something like:

```js
  let indexHTML = fs.readFileSync(__dirname+'/index.html').toString();
  const dataRegex = /%DATA%/;

  const buildQuery = (handler, query = []) => {
    query = query.concat(handler.instance.collectQuery());
    if (handler.parent) {
      buildQuery(handler.parent, query)
    }
    return callUpstreamService(query)
  }
  app.colllectQueryFor('/profile/123').then(buildQuery).then(data => {
     let indexHTML = indexHTML.replace(dataRegex, JSON.stringify(data));
     res.writeHead(200, {
       'Content-Length': indexHTML.length,
       'Content-Type': 'text/html'
     });
     res.write(indexHTML);
     res.end();
  })
```
 
This would allow you to extract the data requirements without a lot of overhead. If not, this can be proved out in user land as you should be able to lookup the route instances from the info.

ef4: I don't want to encourage people to do anything remotely stateful with Route instances. I think it's dubious that they are instances at all.

Your overall scenario is an important one, and it comes down to enabling routes to resolve their promises in parallel instead of in series. That would let your data layer be Falcor or GraphQL-like and automagically batch and combine the requests generated in each of the `model` hooks. There would be no need to do a new top-down query building step -- the hooks we have are already where routes tell us what data they need. 

That whole feature is probably a different RFC. I think it would (1) cause routes to be non-blocking by default (meaning their children's model hooks begin to run even before their own model hook resolves), (2) make `modelFor` return a promise, so that you can re-introduce explicit dependencies when you need them, and (3) provide an optional way to declare a route as blocking, which can still be necessary if you're going to do data-dependent redirections.

chadhietala: Agree on all points.

text/0000-router-service.md
```diff
@@ -0,0 +1,252 @@
...
+  compute([routeName, ...models], hash) {
+    let allModels;
+    if (hash.models) {
+      allModels = models.concat(hash.models);
+    } else {
+      allModels = models;
+    }
+    return this.get('router').isActive(routeName, ...models, hash.queryParams);
+  },
+  observer('router.currentRoute', function() {
```
mixonic: Even in ES6 this needs a property name: `currentRouteDidChange: observer(` etc.

mixonic: Additionally, this should be `currentUrl`? `currentRoute` would remain unchanged if only models changed in the current route.

text/0000-router-service.md
```diff
@@ -0,0 +1,252 @@
...
+
+I propose deprecating the publicly extensible `willTransition` and `didTransition` hooks. They are redundant with an observable `currentRoute`, and the arguments they receive leak internal implemetation from router.js.
+
+## RouteInfo Type
+
+A RouteInfo object has the following properties. They are all read-only.
+
+ - name: the dot-separated, fully-qualified name of this route, like `"people.index"`.
+ - localName: the final part of the `name`, like `"index"`.
+ - attrs: the attributes provided to this route's component, if any, like `{ model }`. For old-style routes that have a controller/view/template instead of a routable component, the implicit `attrs` shall contain `{ model, controller, view }`.
```
mixonic: No `view`?

text/0000-router-service.md
```diff
@@ -0,0 +1,252 @@
...
+ - params: the values of this route's parameters. Same as the argument to `Route`'s `model` hook. Contains only the parameters valid for this route, if any (params for parent or child routes are not merged).
+ - queryParams: the values of any queryParams on this route.
+ - parent: another RouteInfo instance, describing this route's parent route, if any.
+ - child: another RouteInfo instance, describing this route's active child route, if any.
+
+Notice that the `parent` and `child` properties cause `RouteInfos` to form a linked list. So even though the `currentRoute` property on `RouterService` points at the leafmost route, it can be traversed to discover everything about all active routes. As a convenience, `RouteInfo` also implements `Enumerable` over all the reachable `RouteInfos` from topmost to leafmost. This makes it possible to say things like:
+
+```js
+router.currentRoute.find(info => info.name === 'people').params
```
mixonic: `find` traverses the entire list of active routes, including the `currentRoute`? Seems little funky, but obviously `router.find` is also funky.

A lazy `router.get('activeRoutes')` CP could transform the linked list into an array, thus allowing normal enumerable interaction that is free unless you use it.

mixonic: Ah, I see later on where you want to traverse `to` and `from` independent of what is `active`. Hm.

text/0000-router-service.md
```diff
@@ -0,0 +1,252 @@
...
+
+We could adopt some of the existing broadly used APIs as de-facto public. This avoids churn, but imposes a complexity burden on every new learner, who needs to be told "this is a weird API, but it's what we're stuck with".
+
+## Semver Lawyering
+
+I'm interepreting router.js's public/private documentation as out-of-scope for Ember's semver. The fact that we pass an instance of router.js's Transition as our `transition` argument is not documented. An alternative interpretation is that we need to continue supporting those methods marked as public in router.js's docs.
+
+## Optional Helpers
+
+I didn't propose shipping `is-active` and `url-for` template helpers -- I merely showed that they're easy to build using the router service. But we should arguably just ship them as part of the framework too.
```
mixonic: I am :+1: on having them as part of the framework. Leaving them out an this RFC and shunting them to an addon for the immediate future seems a good plan of attack though.

machty:
A few random thoughts/notes:

- Perhaps we should role in some of the details of the routeable components RFC into here, specifically with shrinking the Route hook API to just a `model` that's always called (with objects passed to transitionTo getting serialized into params passed to the always-called model). 

- I want to noodle a little more on what a transition is; transitions manage a mess of async and I would want to minimize via one mechanism or another the ability to reach into it and twiddle with its internals, particularly the progressively-resolving internal handlerInfos

- We need a way to grab the current router state such that it can be transitioned to (returned to) later

MiguelMadero:
@ef4 we have a similar service now. There're two things that we're using that could be part of the public interface. It's easy to build those on top of the current public interface, but something to consider. 

```
    /**
     * Transitions to an ancestor. For example from charts.
     */
    transitionUp: function () {
        var segments = this.get('currentRouteName').split('.');
        segments.pop();
        this.transitionToRoute(segments.join('.'));
    },

    _subscribeToRouterEvents: function () {
        var router = this.get('router');
        router.on('didTransition', this._handleDidTransition.bind(this));
    }.on('init'),

    _handleDidTransition: function () {
        this.trigger('routerDidTransition');
    }
```

Our RoutingService is also Ember.Evented, so a consumer can subscribe to the `routerDidTransition` event to know when there're route changes. We use this from an analytics service and to communicate with our native (ios/android) wrapper.

text/0000-router-service.md
```diff
@@ -0,0 +1,252 @@
...
+`currentRoute`: an observable property. It is guaranteed to change whenever a route transition happens (even when that transition only changes parameters and doesn't change the active route). You should consider its value deeply immutable -- we will replace the whole structure whenever it changes. The value of `currentRoute` is a `RouteInfo` representing the current leaf route. `RouteInfo` is described below.
+
+`currentRouteName`:  a convenient alias for `currentRoute.name`.
+
+`currentURL`: provides the serialized string representing `currentRoute`.
+
+
+### Deprecation
+
+I propose deprecating the publicly extensible `willTransition` and `didTransition` hooks. They are redundant with an observable `currentRoute`, and the arguments they receive leak internal implemetation from router.js.
```
nathanhammond: Dropping `didTransition` and `willTransition` is likely going to be very painful. Basically anybody who ever implemented analytics used those hooks. Some people even (read, me) stored stuff on the transition object and pulled it back out later.

text/0000-router-service.md
```diff
@@ -0,0 +1,252 @@
...
+```
+
+However, this pattern feels awkward, and I think it justifies just adding a public `parentRouteName()` method to `Route` that would simplify to:
+
+```js
+beforeModel(transition) {
+  const parentModel = this.modelFor(this.parentRouteName());
+}
+```
+Possibly we *want* this to feel awkward because it's a weird thing to do.
```
nathanhammond: I'm asked how to do this all the time but usually there is a better solution for whatever the presented problem is than trying to determine the parent route name dynamically. I'm okay with it being awkward, it's still a lot cleaner than diving into `currentHandlerInfos`.

text/0000-router-service.md
```diff
@@ -0,0 +1,252 @@
...
+
+I'm interepreting router.js's public/private documentation as out-of-scope for Ember's semver. The fact that we pass an instance of router.js's Transition as our `transition` argument is not documented. An alternative interpretation is that we need to continue supporting those methods marked as public in router.js's docs.
+
+## Optional Helpers
+
+I didn't propose shipping `is-active` and `url-for` template helpers -- I merely showed that they're easy to build using the router service. But we should arguably just ship them as part of the framework too.
+
+## Branching Route Hierarchies
+
+I am implicitly assuming we will only ever have linear route hierarchies, where a given route has at most one child. I can imagine eventually wanting a way to support branching route hierarchies, where each branch can transition independently. I'm not trying to account for that future.
```
nathanhammond: What you describe here is basically the solution for modals we've all been waiting for.

text/0000-router-service.md
MiguelMadero: I agree. Maybe we need to provide a new event that doesn't expose private members. 

It also seems better than relying on an observer. 



—

text/0000-router-service.md
```diff
@@ -0,0 +1,252 @@
...
+`url(routeName, ...models, queryParams)`
+
+This takes the same arguments as `transitionTo`, but instead of initiating the transition it returns the resulting URL as a string.
+
+A `url-for` helper can be implemented almost identically to the `is-active` example above.
+
+
+### New Properties
+
+`currentRoute`: an observable property. It is guaranteed to change whenever a route transition happens (even when that transition only changes parameters and doesn't change the active route). You should consider its value deeply immutable -- we will replace the whole structure whenever it changes. The value of `currentRoute` is a `RouteInfo` representing the current leaf route. `RouteInfo` is described below.
```
nathanhammond: Implementing this as an observer will mean that you have to wire up your own memoization which was previously easy to do by having both the `willTransition` hook called just in time. Also, I believe that `willTransition` is fired before the transition is resolved making this new API "lossy" if you were trying to (easily) track aborted transitions.

(Not saying that these API changes would prevent any of the existing use patterns, just some of them would be a bit harder to implement.)

text/0000-router-service.md
workmanw: `willTransition` is especially helpful for globally preventing a route based on some global state. Say for example pending file uploads. Something the "routable component" might not be aware of.

text/0000-router-service.md
nathanhammond: :+1: on routes not being instances. Making them stateful prevents all sorts of cleverness for solving exactly the problem Chad brings up. The separate RFC is #97.

text/0000-router-service.md
```diff
@@ -0,0 +1,252 @@
...
+
+## Transition Object
+
+A `transition` argument is passed to `Route`'s `beforeModel`, `model`, `afterModel`, and `willTransition` hooks. Today it's public API is only really `abort()` and `retry()`.
+
+### New Properties: `from` and `to`
+
+I'm proposing we add `from` and `to` properties on `transition` whose values are `RouteInfo` instances representing the initial and final leafmost routes for this transition. Like all RouteInfos, these are read-only and internally immutable. They are not observable, because a  `transition` instance is never changed after creation.
+
+On an initial full-page load, the `from` property will be `null`. This creates a public API for distinguishing in-app transitions from full-page reloads.
```
blimmer: I love this idea. This is so needed. Right now I'm relying on two (private) layers deep in router.js to do this.

blimmer:
Overall, I really like this idea! I think a routing service is very helpful an will make my previous apps use less private APIs. I think the only downside is consumers relying on this service to not handle routing concerns in the `Ember.Route` they're currently on, which the current API encourages.

text/0000-router-service.md
```diff
@@ -0,0 +1,252 @@
...
+ - `willTransition`
+ - `didTransition`
+
+That API will be carried over verbatim to `RouterService`, and the publicly accessible `Ember.Router` class will *become* `RouterService`. In terms of implementation, I expect the existing `EmberRouter` class will continue to exist mostly unchanged. But public access to it will be moderated through `RouterService`.
+
+### New Methods: Initiating Transitions
+
+```js
+transitionTo(routeName, ...models, queryParams)
+replaceWith(routeName, ...models, queryParms)
```
chadhietala: I think it might be useful to also introduce more APIs here to get closer to parody with the underlying browser API. For instance it's currently not very clear how to go back a transition. Solving this in user space normally requires you to save off the previous transition to create an adhoc linked list across your app or create a service to keep track of this, which may be subject to getting out of sync. Things get more leaky when you are in a node environment and now need to guard against `window` references because the framework didn't provide a good way of handle cases.

kellyselden:
I got linked here in slack, that this RFC might be curious how people might be using `currentHandlerInfos`. Here is my relevant piece of code from [ember-cli-transition-to-dynamic](https://github.com/kellyselden/ember-cli-transition-to-dynamic):
```js
  this.router.router.currentHandlerInfos.forEach(route => {
    var routeSegments = routesWithSegments[route.name];

    // using route._names instead of route.params because
    // the latter's order is not guaranteed
    route._names.forEach(param => {
      var paramsSource = routeSegments && routeSegments[param]
        ? routeSegments : route.params;
      models.push(paramsSource[param]);
    });
  });
```

text/0000-router-service.md
stefanpenner: I suspect this would be a tangental RFC

text/0000-router-service.md
stefanpenner: does  `currentRoute` emit change events when QP other params change?

text/0000-router-service.md
stefanpenner: > Implementing this as an observer

observable property, doesn't mean observer. Rather it is just a property that is guaranteed which correctly emits change events. Allowing CP (or maybe observer) to depend.

stefanpenner: @ef4 currentRoute should likely be readOnly, we may also want to explore using Object.freeze or similar.

text/0000-router-service.md
stefanpenner: @ef4 I think we likely want both evented and observable here. It seems like different use-cases would prefer to "subscribe" in different ways.

text/0000-router-service.md
stefanpenner: I think they are fine being stateful, but not good being singletons.

text/0000-router-service.md
```diff
@@ -0,0 +1,252 @@
...
+ - localName: the final part of the `name`, like `"index"`.
+ - attrs: the attributes provided to this route's component, if any, like `{ model }`. For old-style routes that have a controller/view/template instead of a routable component, the implicit `attrs` shall contain `{ model, controller, view }`.
+ - params: the values of this route's parameters. Same as the argument to `Route`'s `model` hook. Contains only the parameters valid for this route, if any (params for parent or child routes are not merged).
+ - queryParams: the values of any queryParams on this route.
+ - parent: another RouteInfo instance, describing this route's parent route, if any.
+ - child: another RouteInfo instance, describing this route's active child route, if any.
+
+Notice that the `parent` and `child` properties cause `RouteInfos` to form a linked list. So even though the `currentRoute` property on `RouterService` points at the leafmost route, it can be traversed to discover everything about all active routes. As a convenience, `RouteInfo` also implements `Enumerable` over all the reachable `RouteInfos` from topmost to leafmost. This makes it possible to say things like:
```
stefanpenner: I'm uncertain about Enumerable future, I believe we should transition to `iterable`, as that isn't something we have done yet, i suspect we should leave out the `enumerable` part here.

text/0000-router-service.md
```diff
@@ -0,0 +1,252 @@
...
+router.currentRoute.find(info => info.name === 'people').params
+```
+
+## Transition Object
+
+A `transition` argument is passed to `Route`'s `beforeModel`, `model`, `afterModel`, and `willTransition` hooks. Today it's public API is only really `abort()` and `retry()`.
+
+### New Properties: `from` and `to`
+
+I'm proposing we add `from` and `to` properties on `transition` whose values are `RouteInfo` instances representing the initial and final leafmost routes for this transition. Like all RouteInfos, these are read-only and internally immutable. They are not observable, because a  `transition` instance is never changed after creation.
```
stefanpenner: when in a series of transitions, with aborts/redirects/retry's will `from` and `to` refer to larger transition, or the more local transition start/stop

text/0000-router-service.md
stefanpenner: having to write guards for `null` may be annoying, should it maybe follow a nullObject pattern?

text/0000-router-service.md
```diff
@@ -0,0 +1,252 @@
...
+
+
+### Actively discourage use of private API
+
+This RFC provides public API for doing the things people have become accustomed to doing via private properties on `transition`. To eliminate confusion over the correct way, we should hide all the private API away behind symbols, and provide deprecation warnings per our usual release policy around breaking "widely-used private APIs"
+
+
+# Drawbacks
+
+This RFC suggests only two small deprecations that are unlikely to effect many apps, so the API-churn burden may appear low. However, we know that use of the private APIs we're deliberately disabling is widespread, so users will experience churn. We can provide our usual deprecation cycle to give them early warning, but it still imposes some cost.
```
stefanpenner: I think those deprecations will likely affect analytics/instrumentation related add-ons (or apps that use them directly). and replicating that class of behavior via observable property feels quirky. I think the two hook deprecations need some further exploration.

text/0000-router-service.md
```diff
@@ -0,0 +1,252 @@
...
+
+This RFC doesn't attempt to change the existing and fairly rich semantics for initiating transitions. For example, you can pass either models or IDs, and those have subtle semantic differences. I think an ideal rewrite would also change the semantics of the route hooks and transitionTo to simplify that area.
```
stefanpenner: seems good to keep those separate.

text/0000-router-service.md
```diff
@@ -0,0 +1,252 @@
...
+
+# Alternatives
+
+## Less Churn
+
+We could adopt some of the existing broadly used APIs as de-facto public. This avoids churn, but imposes a complexity burden on every new learner, who needs to be told "this is a weird API, but it's what we're stuck with".
```
stefanpenner: We likely need to take a survey of some real apps, and see how many of those on-transition properties are wildly used. I know I have touched them, when nothing else worked.

The survey results likely wouldn't stop us from slowly phasing them out, rather to be sure we phase out "weird API" IFF a reasonable alternative is provided (unless we deem the scenario to esoteric  )

text/0000-router-service.md
```diff
@@ -0,0 +1,252 @@
...
+
+I am implicitly assuming we will only ever have linear route hierarchies, where a given route has at most one child. I can imagine eventually wanting a way to support branching route hierarchies, where each branch can transition independently. I'm not trying to account for that future.
+
+## Route.parentRoute
+
+This RFC makes it possible for a route to determine its parent's name dynamically via public API, and thus access its parent's model/params/controller:
+
+```js
+beforeModel(transition) {
+  const parentInfo = transition.to.find(info => info.name === this.routeName).parent;
```
stefanpenner: an alternative (or addition) to the general purpose `find` is the corresponding `findBy('name')`

text/0000-router-service.md
```diff
@@ -0,0 +1,252 @@
...
+  const parentModel = this.modelFor(parentInfo.name);
+}
+```
+
+However, this pattern feels awkward, and I think it justifies just adding a public `parentRouteName()` method to `Route` that would simplify to:
+
+```js
+beforeModel(transition) {
+  const parentModel = this.modelFor(this.parentRouteName());
```
stefanpenner: seems like it could be a property?

text/0000-router-service.md
rtablada: To check would the proposed observable/event be on the router service or the private router?

Seems like an interesting thought to have components subscribe to `beforeTransition` or whatever the event would be and could interject their thoughts.
The first thing that comes to mind is a `user-form` component within a `users.create` route, while the route does not know the `user-form`'s state until submission (DDAU) the form could listen for the transition and check to see if any data could be lost and alert/confirm, etc.

nickiaconis: @ef4 By deprecating the `willTransition` and `didTransition` _hooks_, do you mean that the `willTransition` and `didTransition` _events_ would still be available (or something akin to them at least)? Would this be the time/place to request a `changeTransition` event (name up for debate) that would be triggered for transitions initiated between `willTransition` and `didTransition`?

nickiaconis:
@ef4 -- By deprecating the `willTransition` and `didTransition` _hooks_, do you mean that the `willTransition` and `didTransition` _events_ would still be available (or something akin to them at least)?

Also, would this be the time/place to request a `changeTransition` event (name up for debate) that would be triggered for transitions initiated between `willTransition` and `didTransition` (i.e. when one route redirects to another before `didTransition` is fired)?

ef4:
I originally meant deprecating both the hooks and the events. But the feedback here has made it clear that people are relying on them and we need to account for that, either by keeping them as they are or designing something even better. Specifically, we may want to get more specific about what happens with the hooks during redirections and errors.

text/0000-router-service.md
```diff
@@ -0,0 +1,252 @@
...
+This RFC makes it possible for a route to determine its parent's name dynamically via public API, and thus access its parent's model/params/controller:
+
+```js
+beforeModel(transition) {
+  const parentInfo = transition.to.find(info => info.name === this.routeName).parent;
+  const parentModel = this.modelFor(parentInfo.name);
+}
+```
+
+However, this pattern feels awkward, and I think it justifies just adding a public `parentRouteName()` method to `Route` that would simplify to:
```
asakusuma: It would be useful to have every route in the hierarchy available via a public API. We use handlerInfos to get this information for tracking purposes right now.

nathanhammond: Depending upon when you access it nothing below the current pivot is guaranteed to be "resolved." `redirect` vs. any of the `model` hooks behave differently. Intricacies surrounding that make it hard to reason about what the "correct" behavior should be if you're accessing descendants.

asakusuma: I believe our particular use case only needs the routes in the hierarchy above the pivot. In fact we do work to figure out where the pivot is, so it would actually be nicer for us to only get the routes above the pivot.

text/0000-router-service.md
locks: agreed on both leaving them out, and requiring naming bikeshedding.

tim-evans:
For those interested, I have implemented a polyfill for this RFC for use in our Ember apps: https://gist.github.com/tim-evans/d9e07256baaa8d61944205a66a29336b

Hopefully this helps!

workmanw:
@tim-evans Honestly this would make a great addon. We have our own half-implemented polyfill for this so that components can do route transitions and check the route location. I've heard of others doing the same.

locks:
@workmanw I really advise against an addon with the amount of internals you need to touch.

workmanw:
@locks Well clearly there is a need here. Some basic level of information about the route state is often important. The `is-active` helper is about as fundamental of an example as it gets. Just thought it would be helpful, but I guess I'm wrong.  ¯\_(ツ)_/¯.

locks:
> Well clearly there is a need here.

Hence the RFC. Which hasn't shipped due to lack of implementation.

nickiaconis:
> Hence the RFC. Which hasn't shipped due to lack of implementation.

I was under the impression RFCs should ship before they are implemented. Is that not how it works? From the emberjs/rfcs README, emphasis mine:

> Of course, the RFC author (like any other developer) is welcome to post an implementation for review _after the RFC has been accepted_.

ef4:
Implementation and RFC can be simultaneous or not, as long as everybody is aware that the implementation is experimental and highly subject to change.

mixonic:
It is 100% appropriate (and helpful) for there to be a polyfill/addon to explore this. Thanks @tim-evans! Perhaps I can cajole you into pushing some of the work upstream, let us chat on Slack :-)

rwjblue:
Some feedback from the core team F2F:

- [ ] Rename router service method from `url` to `urlFor`
- [ ] Add capability to `isActive` so that you can specify that you want to match a given destination as active when it uses the default query param based on the bucket cache (previously known as model dependent state) for that route.
- [ ] Make it clear that the various API's do not use controller default values to strip them from query params:
  - [ ] `isActive`
  - [ ] `urlFor`
  - [ ] `recognize`
- [ ] Update `RouteInfo` section:
  - [ ] indicate that `params` are the serialized values (not full models if provided)
  - [ ] Add `models` (when during a transition and the model is known)
- [ ] Specifically discuss `willTransition` / `didTransition`.
- [ ] Fix semantics of `willTransition` / `didTransition`.
- [ ] Keep `router:main` (and its current private API's) around with deprecations.

nathanhammond:
@rwjblue Does this also imply that we're dropping default query params?

ef4:
@nathanhammond it means that explicitly setting to a default value would cause the queryParam to render in the URL -- at least when using newly added API methods. Possibly also for all link-tos if we can mitigate breakage. 

The cost of looking up the defaults is very high, because it instantiates controllers throughout the app that you may have otherwise never needed. And it doesn't play well with lazy loading engines. 

I will incorporate Rob's notes into the RFC. There are some extra details that will help, like adding a way to express the difference between "don't care what a particular QP is" vs "please set to default (without needing to know the value of default)".

nathanhammond:
@ef4 I'm in favor of dropping default query params from existence. It impacts engines, lazy bundle loading, and many other places (including your example of controller instantiation) with the small cost of a "slightly more verbose URL." I also like @sethkinast's thoughts here: https://github.com/dgeb/ember-engines/issues/6#issuecomment-225459660

ef4:
Yes, I can summarize how I feel about query params in general: they should be just a different way of formatting the URL. They're not fundamentally different from the regular `params` that get passed into the `model` hook. IMO it should be possible to refactor between `/my-route/:page_id` and `/my-route/?page=:page_id` without changing anything about the way the value is consumed.

ef4:
We need to remember to make the URL generation and recognition include the app's rootURL.

text/0000-router-service.md
```diff
@@ -0,0 +1,260 @@
...
+ - `EmberRouter` is the class that we already have today, defined in `ember-routing/system/router` and available publicly as `Ember.Router`
+ - `RouterService` is the new class I am proposing.
+
+`EmberRouter` has the following public API today:
+
+ - `map`
+ - `location`
+ - `rootURL`
+ - `willTransition`
+ - `didTransition`
```
nickiaconis: `EmberRouter`'s  public `willTransition` and `didTransition` API include the parameters `oldInfos` and `newInfos`, in addition to (and preceding) the `transition` parameter.

https://github.com/emberjs/ember.js/blob/v2.7.0/packages/ember-routing/lib/system/router.js#L292

Will these be carried over as well?

ef4: No, one of the explicit goals of this RFC is to codify public API so that people can stop messing around with handlerInfos (which are an internal implementation detail of router.js and are not Ember public API).

text/0000-router-service.md
```diff
@@ -0,0 +1,439 @@
...
+let MyComponent = Ember.Component.extend({
+  didInsertElement() {
+    // Access to the variable here is intended to be indistinguishable
+    // from a normal, explicitly-passed input argument. 
+    doSomethingWith(this.get('myDynamicVariable'));
+  }
+});
+MyComponent.reopenClass({
+  // This is where the specialness happens.
+  getDynamicVariables: [ 'myDynamicVariable' ]
```
rwjblue: Why not require it to be passed in via the template layer?  If this stays, we need to define exactly the semantics.

* What hooks does a change to a JS-land declared dynamic variable trigger?
* Is the variable observable via CP and/or observers?
* What happens if you `this.set('myDynamicVariable')`?  Does it throw, clobber, etc?

text/0000-router-service.md
```diff
@@ -0,0 +1,439 @@
...
+Defining a helper that reads a dynamically-scoped variable:
+
+```js
+let MyHelper = Ember.Helper.extend({
+  compute(params, hash) {
+    return doSomethingWith(hash.myDynamicVariable);
+  }
+});
+MyHelper.reopenClass({
+  getDynamicVariables: [ 'myDynamicVariable' ]
```
rwjblue: Same as with my comment above. Does a change to the dynamic variable for `compute` to be called again, or do we need to stash it off, observe it, and call `this.recompute()`?

text/0000-router-service.md
```diff
@@ -0,0 +1,439 @@
...
+    within this block AND ALL ITS DESCENDANTS until
+    otherwise overridden by another with-dynamic-variable statement, 
+    `get-dynamic-variable "myVariableName"` returns someValue.
+  -}}
+{{/with-dynamic-variable}}
+```
+
+Note that there is no `set-dynamic-variable`. You can only introduce new scopes, not mutate your containing scope. There is also no way to set a dynamically-scoped variable directly from Javascript -- your component must use a `with-dynamic-variable` block within its handlebars template.
+
+User-defined, dynamically-scoped variables must be pre-declared before first render via `Ember.declareDynamicVariable("myVariableName")`. Attempting to set one that wasn't declared is an error. Attempting to read one that doesn't exist returns `undefined`. Predeclaration is an optimization that allows us to maintain consistent performance within Glimmer.
```
rwjblue: Complete bikeshedding, but I'd rather not put the method on the root of the `Ember` global like this.  Something like `Ember.Templates` or `Ember.Rendering` or something with additional options seems better to me...

text/0000-router-service.md
```diff
@@ -256,5 +390,16 @@ beforeModel(transition) {
...
 
+## Naming of Ember.DEFAULT_VALUE Symbol
+
+Should we introduce new API via the `Ember` global and switch to a module export once all the rest of Ember does, or should we just start with a module export right now? If so, what module?
+
+    import { DEFAULT_VALUE } from 'ember-routing';
+
+## Generic dynamically-scoped variables may be too powerful
+
+They let you do spooky-action-at-a-distance stuff. We could instad choose to implement single-purpose methods like `(get-route-info)` and `{{#with-route-info someValue}}` that would solve the routing case without opening the door to arbitrary user-defined ones.
```
rwjblue: s/instad/instead/

text/0000-router-service.md
ef4: Agreed, we can use namespaces instead. I considered even going straight to ES2015 imports.

text/0000-router-service.md
ef4: I considered requiring doing it in a template, but that has poor ergonomics. We want `get-dynamic-variable` to be an internal implementation detail, not something that callers need to know about. So you want to hide it inside the implementation of your helper or component.

You can do that in a component by nesting another private component inside it, and that is merely awkward:

```hbs
{{-my-inner-component info=(get-dynamic-variable "routeInfo")}}
```

But helpers are worse because there's no way to implement one in terms of hbs. 

As for the detailed semantic questions: the answer to all of them is that these paramters should be indistinguishable from parameters that are passed explicitly in the template. In theory, you could implement them with an AST transform that rewrites:

```hbs
{{my-component foo=bar}}
```

into

```hbs
{{my-component foo=bar myDynamicVariable=(get-dynamic-variable "myDynamicVariable")}}
```

I deliberately didn't specify any different semantics because they should always follow what other normal parameters do -- and that will differ between classic and glimmer components.

text/0000-router-service.md
ef4: If by "we" you mean the person implementing the helper, no, it's definitely automatic.

As for implementation, we already have implementations that do this correctly for both current Ember and Glimmer2 -- that is how `{{outlet}}` works. To the template layer it's just another stream (today) or reference (g2) that can be invalidated.

text/0000-router-service.md
rwjblue: > I deliberately didn't specify any different semantics because they should always follow what other normal parameters do

Then those are the semantics we should say that we follow

> Will treat dynamic variables found in `getDynamicVariables` just like any other named argument to your helper/component.

This clarifies the mental model (at least mine) that a dynamic variable is roughly the same as the AST transform version that you mentioned above.

text/0000-router-service.md
rwjblue: This question was answered in the other section.  The thing that was unclear to me was that these are treated as if you had specified a hash argument of `variable-name=(get-dynamic-variable 'variable-name')`.  Thus `compute` is properly called when that dynamic variable is changed.

text/0000-router-service.md
```diff
@@ -400,6 +437,3 @@ Should we introduce new API via the `Ember` global and switch to a module export
 
 They let you do spooky-action-at-a-distance stuff. We could instad choose to implement single-purpose methods like `(get-route-info)` and `{{#with-route-info someValue}}` that would solve the routing case without opening the door to arbitrary user-defined ones.
 
-## Possible semantic breakage of willTransition
```
nathanhammond: Agreed that this is far too large of a breaking change. The reason we pushed to get #126 onto the agenda was because we reached a point in which we can't continue to maintain all of our "`willTransition` fires every time" hacks any longer. I know of at least four addons (three of which are internal) break along this change.

text/0000-router-service.md
```diff
@@ -148,11 +148,48 @@ Example: this feature can replace [this use of private API in ember-href-to](htt
...
+ 6. routeDidChange fires `from` A `to` C.
+
+Abort example:
+
+ 1. current route is A
+ 2. user initiates a transition to B
+ 3. routeWillChange fires `from` A `to` B.
+ 4. in response to the previous routeWillChange event, the transition is aborted.
+ 5. routeWillChange fires `from` A `to` A.
+ 8. routeDidChange fires `from` A `to` A.
```
nathanhammond: Instinct says that this is different from the current behavior, but I can't confirm without review.

text/0000-router-service.md
```diff
@@ -0,0 +1,440 @@
...
+
+On an initial full-page load, the `from` property will be `null`. This creates a public API for distinguishing in-app transitions from full-page reloads.
+
+### Example: testing whether route will remain active
+
+Here's an example showing how `willTransition` can figure out if the current route will remain active after the transition:
+
+```js
+willTransition(transition) {
+  if (!this.transition.to.find(route => route.name === this.routeName)) {
```
nathanhammond: `route.name` is not guaranteed to be unique, especially with the advent of engines. I'm aware that this is just an example, but it is something we should try to solve for in a friendly way. Whatever that Route object is should be loaded at this time, maybe we attach a reference to it inside of the `RouteInfo` object.

text/0000-router-service.md
ef4: This is deliberately different from current behavior, which is why I deprecated `willTransition` and introduced a new name, `routeWillChange`.

text/0000-router-service.md
nathanhammond: Alternatively, it is guaranteed to be unique, but we end up leaking the mount point of the engine into the engine allowing for all sorts of bad behavior.

text/0000-router-service.md
```diff
@@ -0,0 +1,440 @@
...
+
+### Deprecating willTransition and didTransition
+
+Application-wide transition monitoring events belong on the Router service, not spread throughout the Route classes. That is the reason for the existing `willTransition` and `didTransition` hooks/events on the Router. But they are not sufficient to capture all the detail people need. See for example, https://github.com/nickiaconis/rfcs/blob/1bd98ec534441a38f62a48599ffa8a63551b785f/text/0000-transition-hooks-events.md
+
+In addition, they receive handlerInfos in their arguments, which are an undocumented internal implementation detail of router.js that doesn't belong in Ember's public API. Everything you can do with handlerInfos can be done with the RouteInfo type that is proposed in this RFC, with the benefit of sticking to supported public API.
+
+So we should deprecate willTransition and didTransition in favor of the following new events.
+
+### New Events: routeWillChange & routeDidChange
```
nathanhammond: What should we do about the parallel events that appear on the `Route`s? They're subtly different, and firing a stream of `routeWillChange` events at a route is a _bit_ odd. I'm comfortable with this change only being at the `Router` level, but I worry about teachability in that scenario.

ef4: Yeah, they're not just subtly different, they're completely different.

I think the teaching story is that the events on a Route are for things that are specific to that route, and the global events on the router are for global concerns like analytics.

text/0000-router-service.md
```diff
@@ -0,0 +1,440 @@
...
+
+
+Example: this feature can replace [this use of private API in ember-href-to](https://github.com/intercom/ember-href-to/blob/b8cf0699eec6a65570b05e4fc22b27d8cea49c42/app/instance-initializers/browser/ember-href-to.js#L34).
+
+
+### New Method: Recognize and load models
+
+`recognizeAndLoad(url)`
+
+Takes a string URL and returns a promise that resolves to a `RouteInfoWithAttributes` for the leafmost route represented by the URL. The promise rejects if the URL is not recognized or an unhandled exception is encountered. This method expects to receive the actual URL as seen by the browser _including_ the app's `rootURL`.
```
nathanhammond: It seems like this must process the entire hierarchy's "model chain" in order to guarantee safety. I'm not sure how you can just load the leaf? We get around this by creating a separate `prefetch` hook which knows it isn't allowed to depend on anything unless it explicitly wires up its own promise chain. Nonetheless it still relies on all the `prefetch` hooks in the hierarchy being invoked in order so that the user can manually wire together a promise chain.

text/0000-router-service.md
nathanhammond: How would you distinguish between `Ember.Route#refresh` and `Transition#abort`?

text/0000-router-service.md
```diff
@@ -0,0 +1,440 @@
...
+ 8. routeDidChange fires `from` A `to` A.
+
+Error example:
+
+ 1. current route is A
+ 2. user initiates a transition to B.index
+ 3. routeWillChange fires `from` A `to` B.
+ 4. B throws an exception, and the router discovers a "B-error" template.
+ 5. routeWillChange fires `from` A `to` B-error
+ 6. routeDidChange fires `from` A `to` B-error
```
nathanhammond: Transition into loading and error states is done synchronously and not allowed to fail. I'm not sure that we want to expose the `transition` object at that point in time. I'm generally +1 on being able to inspect this behavior but I'm afraid that we may open up a can of worms if we support monitoring transitions via `intermediateTransitionTo`.

nathanhammond: (This also applies to the `application` route which is not allowed to reject from its model hook.)

nathanhammond: Also (implementation detail) the event handler must be invoked as a function otherwise the microtask queue will move happily along without it. 😜

text/0000-router-service.md
```diff
@@ -0,0 +1,440 @@
...
+
+`currentURL`: provides the serialized string representing `currentRoute`.
+
+### Query Parameter Semantics
+
+Today, `queryParams` impose unnecessarily high cost because we cannot generate URLs or determine if a link is active without taking into account the default values of query parameters. Determining their default values is expensive, because it involves instantiating the corresponding controller, even in cases where we will never visit its route.
+
+Therefore, the `queryParams` argument to the new `urlFor`, `transitionTo`, `replaceWith`, and `isActive` methods defined in this document will behave differently.
+
+ - default values will not be stripped from generated URLs. For example, `urlFor('my-route', { sortBy: 'title' })` will always include `?sortBy=title`, whether or not `title` is the default value of `sortBy`.
```
nathanhammond: Hallelujah.

text/0000-router-service.md
ef4: Yes, it necessarily must process the entire chain. A `RouteInfo` object always recursively contains a complete state for all active routes, not just one leaf. It's just that the leafmost one happens to be the handle you are given as a starting point.

If people are depending on controller state, we really can't help them here. But anything else should just work. And on the sweet future day when there are no longer singleton controllers, the problem goes away.

text/0000-router-service.md
```diff
@@ -0,0 +1,440 @@
...
+
+# Alternatives
+
+## Less Churn
+
+We could adopt some of the existing broadly used APIs as de-facto public. This avoids churn, but imposes a complexity burden on every new learner, who needs to be told "this is a weird API, but it's what we're stuck with".
+
+## Semver Lawyering
+
+I'm interpreting router.js's public/private documentation as out-of-scope for Ember's semver. The fact that we pass an instance of router.js's Transition as our `transition` argument is not documented. An alternative interpretation is that we need to continue supporting those methods marked as public in router.js's docs.
```
nathanhammond: `router.js` is on my list to tackle; in my approach to `route-recognizer` I treated anything that I could get away with without breaking Ember's public and intimate APIs (turned out to be very little, but whatever) as fair game to change. I'd like to take the same approach here. Breaking change to `router.js`; SemVer compatible change to Ember.

text/0000-router-service.md
ef4: The stream of events for each is distinct.

    willTransition A->B
    willTransition A->A
    didTransition A->A

vs

    willTransition A->A
    didTransition A->A

You can uniquely identify these sequences because they always end with didTransition.

text/0000-router-service.md
```diff
@@ -0,0 +1,440 @@
...
+
+## Optional Helpers
+
+I didn't propose shipping `is-active` and `url-for` template helpers -- I merely showed that they're easy to build using the router service. But we should arguably just ship them as part of the framework too.
+
+## Branching Route Hierarchies
```
nathanhammond: This is something that I want to revisit; I had a conversation with @machty this summer about his stack routing research. It seems likely we want to try and solve this coincidentally to "rip the bandaid off" so to say.

text/0000-router-service.md
```diff
@@ -0,0 +1,440 @@
...
+  const parentModel = this.modelFor(this.parentRouteName());
+}
+```
+Possibly we *want* this to feel awkward because it's a weird thing to do.
+
+## Naming of Ember.DEFAULT_VALUE Symbol
+
+Should we introduce new API via the `Ember` global and switch to a module export once all the rest of Ember does, or should we just start with a module export right now? If so, what module?
+
+    import { DEFAULT_VALUE } from 'ember-routing';
```
nathanhammond: Yes this.

text/0000-router-service.md
nathanhammond: You could also theoretically `abort` a `refresh` which would be just:
```
willTransition A->A
didTransition A->A
```

Thinking about it more, in this scenario doing all of the bookkeeping is probably best done by maintaining a reference to the initial transition in the external (tracking, whatever) event handler and if the `didTransition` fires with a different transition than you started with you've identified that it changed... it just means you can't solely rely on `from` and `to` to get there.

Seems fine.

text/0000-router-service.md
nathanhammond: Okay, we're on the same page. Maybe worth clarifying in the RFC as it isn't immediately clear.

text/0000-router-service.md
ef4: I think this RFC creates enough public API that you could implement these things in userspace without crazy hacks.

For example, to do a split-page app with two independent route hierarchies, your code could call `recognizeAndLoad` twice, and use `{{#with-dynamic-variable "routeInfo"}}` to render two separate blocks, one for each half. You could even have link-to equivalents that are smart enough to only navigate their own half of the app, by looking at the surrounding routeInfo.

text/0000-router-service.md
```diff
@@ -0,0 +1,440 @@
...
+{{#with-dynamic-variable "myVariableName" someValue}}
+  {{!-
+    within this block AND ALL ITS DESCENDANTS until
+    otherwise overridden by another with-dynamic-variable statement, 
+    `get-dynamic-variable "myVariableName"` returns someValue.
+  -}}
+{{/with-dynamic-variable}}
+```
+
+Note that there is no `set-dynamic-variable`. You can only introduce new scopes, not mutate your containing scope. There is also no way to set a dynamically-scoped variable directly from Javascript -- your component must use a `with-dynamic-variable` block within its handlebars template.
```
nathanhammond: What are the rules for shadowing?

```hbs
{{#with-dynamic-variable "common" "outer scope"}}
  1. {{common}}
  {{#my-component common="inner scope"}}
    2. {{common}}
  {{/my-component}}
{{/with-dynamic-variable}}
```

I suspect the answer is the following?

```
1. outer scope
2. inner scope
```

text/0000-router-service.md
```diff
@@ -0,0 +1,440 @@
...
+This RFC provides public API for doing the things people have become accustomed to doing via private API. To eliminate confusion over the correct way, we should hide all the private API away behind symbols, and provide deprecation warnings per our usual release policy around breaking "widely-used private APIs".
+
+Some of the private APIs we should mark and warn include:
+
+ - transition.state
+ - transition.params
+ - `lookup('router:main')` (should use `service:router` instead)
+
+
+## Dynamically-Scoped Variables
```
nathanhammond: This addresses `get-outlet-state` in terms of getting information. In ember-a11y I trigger side effects from this. I suppose I can still do that by triggering on the getter of something but it feels dirtier, somehow.

text/0000-router-service.md
mmun: Did you mean this?

```
  {{#my-component as |common|}}
    2. {{common}}
  {{/my-component}}
```

ef4: These variables live in a completely separate namespace from normal (lexically-scoped) variables. `{{common}}` can never refer to a dynamically-scoped variable. You would need to write `{{get-dynamic-variable "common"}}`. 

So there is no shadowing issue between dynamic and lexical variables.

Dynamic variables in inner scopes can shadow dynamic variables in outer scopes, which is the point of `#with-dynamic-variable`

text/0000-router-service.md
ef4: You don't need a getter-with-side-effects. You can have a component that takes `routeInfo` as an input parameter, and in `didRender` check to see whether the latest `routeInfo` is actually different enough to warrant the action.

(Doing a stability check is necessary because your component would see a new value even when only one of its child routes was rerendering, because `routeInfo` contains a comprehensive picture of the whole route hierarchy.)

text/0000-router-service.md
```diff
@@ -0,0 +1,440 @@
...
+
+The next section introduces general-purpose, dynamically-scoped variables. Why am I doing this in "The Router Service RFC"? Because dynamically-scoped RouteInfo has several nice capabilities that will be introduced in the subsequent section. And once we have one dynamically-scoped variable, we have the option of exposing that capability for other uses.
+
+### In the general case
+
+A [dynamically-scoped variable](https://en.wikipedia.org/wiki/Scope_(computer_science)#Dynamic_scoping) takes its value from its calling context, as opposed to its lexical context. Since dynamically-scoped variables are powerful to the point of potential danger, their syntax is intended to be appropriately verbose.
```
mmun: @ef4 So if you have an index route template like

```
{{#outer-component}}
  {{#inner-component}}
    hello
  {{/inner-component}}
{{/outer-component}}
```

whose rendering "call stack" looks something like

```
index route template
outer-component's template
outer-component's block
inner-component's template
inner-component's block
```

then setting a dynamic variable in `outer-component's template` would make it available in everything below that: `outer-component's block`, `inner-component's template`, `inner-component's block`?

ef4: Correct.

text/0000-router-service.md
nickiaconis: Does this means `willTransition` and `didTransition` will continue to exist on `Route` (i.e. not be deprecated) and only `Router` gets the new `routeWillChange` and `routeDidChange` events?

If so, the naming difference should help illustrate that there is a difference in the concerns covered by each. Some amount of teaching will still be necessary to explain which concerns are covered by which, but at least no one will be able to suggest that the use case for `Router#X` and `Route#X` is synonymous anymore (where `X` is `willTransition` or `didTransition`).

workmanw:
I know this poor timing given that some of this is actively in development. This week I found that I really needed `modelFor`, or some equivalent, from another component. 

Here is my use case. We use MixPanel to track detailed analytics. We have an intercom support button in the top corner of our app (located in a high up route template, e.g. `application.hbs`). We wanted to track, via MixPanel, if the user was in a specific section of the app when they clicked the support button. And if so, we needed to add state from a routable model to the tracking event.

Since this component was higher up in the route hierarchy, I couldn't simply bind the data down. We could have lookup the singleton controller via the owner API and got the model that way. Or we could have stored/cleared the model on a service in the route's `setupController`/`deactivate` hooks. 

So I thought maybe this is something solved by the router service, but it is not covered by this RFC. I wanted to call this out because it felt weird that with this service I can check to see if a route is active, but I can't figure out what the model associated with it is.

ef4:
@workman I think it is covered, in at least two different ways.

Option one is to make your support button component consume the `routeInfo` dynamic variable. Through that it has access to the entire current state of the route hierarchy.

Option two is to make a service that tracks the `routeDidChange` events on the router service and remembers the last one it heard.

workmanw:
@ef4 Okay cool. I had originally thought that the `routeInfo` was only relating to the leaf route. But upon further inspection I can see that, _"that the parent and child properties cause RouteInfos to form a linked list"_. Thank you.

text/0000-router-service.md
```diff
@@ -0,0 +1,440 @@
...
+
+This takes the same arguments as `transitionTo`, but instead of initiating the transition it returns the resulting root-relative URL as a string (which will include the application's `rootUrl`).
+
+A `url-for` helper can be implemented almost identically to the `is-active` example above.
+
+### New Method: URL recognition
+
+`recognize(url)`
+
+Takes a string URL and returns a `RouteInfo` for the leafmost route represented by the URL. Returns `null` if the URL is not recognized. This method expects to receive the actual URL as seen by the browser _including_ the app's `rootURL`.
```
arjansingh: In addition to returning `null` can `recognize` also call a hook on the `location` (e.g., `location.routeNotFound`)? 

One of the downsides of the current router is that it [throws an exception](https://github.com/tildeio/router.js/blob/b921fd3407064a77bd99f4f4eddc4377592ac93b/lib/router/transition-intent/url-transition-intent.js#L17-L22) when no route is found.

This makes it impossible to recover from an unrecognized route. In the case of FastBoot, it forces us to handle unrecognized routes at the server middleware layer, instead in the Ember application itself.

@tomdale or @danmcclain, Please add any other concerns that I might have forgot.

ef4: I'm not fully understanding this issue. If you want to handle unrecognized URLs within the app, don't we already have error routes that cover that? Wouldn't the ember app just render it's own error template within the fastboot server?

This does highlight a potential gap in the RFC. I didn't go into detail on whether recognizeAndLoad should go down the error handling path. I think it probably should.

arjansingh: We never get to an error handling route because we are using an `ApplicationInstance` to render the page.

Instead it will just [reject](https://github.com/emberjs/ember.js/blob/v2.7.0/packages/ember-application/lib/system/application-instance.js#L269-L279) with a [`UnrecognizedURLError`](https://github.com/tildeio/router.js/blob/b921fd3407064a77bd99f4f4eddc4377592ac93b/lib/router/transition-intent/url-transition-intent.js#L15-L22) whenever the Router cannot find the requested route.

That bubbles up to our [middleware](https://github.com/ember-fastboot/fastboot-express-middleware/blob/master/src/index.js#L60-L65) where we can do our handling. That's why we want a hook. You're right, this could be a larger issue with how routing is done in ApplicationInstance sandboxes as opposed to an app that's run in a full browser.

dgeb:
This RFC is now entering a ["final comment period"](https://github.com/emberjs/rfcs/pull/149). This period is intended to focus the community's attention for a final round of consideration and feedback before an RFC is merged.

If no significant blockers are raised by **October 7**, we plan to merge this RFC and make plans for its implementation. :shipit:

text/0000-router-service.md
```diff
@@ -0,0 +1,440 @@
...
+// If personId 5 is invalid, and the user visits /person/5/detail, they will get
+// redirected to /person/0/detail. And /person/5 will get redirected to /person/0.
+```
+
+
+### Actively discourage use of private API
+
+This RFC provides public API for doing the things people have become accustomed to doing via private API. To eliminate confusion over the correct way, we should hide all the private API away behind symbols, and provide deprecation warnings per our usual release policy around breaking "widely-used private APIs".
+
+Some of the private APIs we should mark and warn include:
```
kellyselden: You could add `router.router.currentHandlerInfos` and `router.router.recognizer.recognize` to this list I believe.

text/0000-router-service.md
kellyselden: I believe with this linked list you can get what `router.router.currentHandlerInfos` used to get you. I also use `route._names` as an ordered version of `route.params` mentioned [here](https://github.com/emberjs/rfcs/pull/95#issuecomment-147473764). Is this something that is covered in the RFC? Is the new `RouteInfo.params` ordered?

kellyselden:
I added some comments a couple weeks ago. I'm unfamiliar with final comment period. Are those comments now blocking the merge, or will the merge proceed with unaddressed comments?

ef4:
@kellyselden I missed your open question about param ordering. 

`params` is a POJO of keys and values (it's the same thing that gets passed into a Route's `model` hook), so it's not ordered. We can add a `paramNames` to RouteInfo that would b an array of string parameter names, so that it's possible to determine what parameters are required for the route and in what order.

This seems like a simple and uncontroversial bit of detail to add, thanks for pointing it out.

ef4:
After taking feedback and hearing a lot of uncertainty around general-purpose dynamic variables, I'm going to table them in favor of `get-route-info` and `set-route-info` which the RFC mentions in Alternatives under "Generic dynamically-scoped variables may be too powerful". They serve the exact same purpose for routing, without opening the door to non-routing uses. We can do general-purpose dynamic variables in a separate RFC.

nathanhammond:
Reviewing the original implementation of the `query-params` subexpression we have _always_ [disallowed passing of POJOs](https://github.com/emberjs/ember.js/commit/808419cc1c77cffa3a88d093d8e58ecbf742a2d8#diff-16c5478c4db7319cfe85de73182a8b4aR808).

I'd like to add a rider to this RFC which allows the subexpression to receive objects. It becomes incredibly annoying in the scenario where we don't investigate the final state to prune values from the query string inside of Handlebars files and makes it more difficult to accomplish [standardization of nested query param behavior](https://github.com/nathanhammond/query-string-spec/blob/master/SPEC.md) since everything must still be key/value pairs.

Current State:
```handlebars
{{#link-to 'some-route' (query-params
  key1="value1"
  key2="value2"
  key3=(if key3 "value3")
)}}Some Route{{/link-to}}
```

Proposed Approach:

```javascript
Ember.controller.extend({
  someQPObject: {
    key1: 'value1',
    key2: 'value2'
  }
});
```

```handlebars
{{#link-to 'some-route' (query-params someQPObject)}}Some Route{{/link-to}}
```

Sketch to update [current implementation](https://github.com/emberjs/ember.js/blob/aeeddf7c00b42a340494465252481b5414d87009/packages/ember-glimmer/lib/helpers/query-param.js):

```javascript
positional.forEach(function(positionalArg) { if (typeof positionalArg !== "object") { assert(); } });
var result = Ember.assign({}, ...positional);
```

ef4:
@nathanhammond that all sounds good but this RFC is already pretty big and thoroughly argued, and I was hoping not to have to restart final comment period (since my latest updates were all just incorporating feedback from the discussion so far and AFAIK there are no controversial questions outstanding).

Done separately, the query params change seems like a short and sweet RFC that can sail through.

text/0000-router-service.md
```diff
@@ -0,0 +1,434 @@
...
+
+### Query Parameter Semantics
+
+Today, `queryParams` impose unnecessarily high cost because we cannot generate URLs or determine if a link is active without taking into account the default values of query parameters. Determining their default values is expensive, because it involves instantiating the corresponding controller, even in cases where we will never visit its route.
+
+Therefore, the `queryParams` argument to the new `urlFor`, `transitionTo`, `replaceWith`, and `isActive` methods defined in this document will behave differently.
+
+ - default values will not be stripped from generated URLs. For example, `urlFor('my-route', { sortBy: 'title' })` will always include `?sortBy=title`, whether or not `title` is the default value of `sortBy`.
+
+ - to explicitly unset a query parameter, you can pass the symbol `Ember.DEFAULT_VALUE` as its value. For example, `transitionTo('my-route', { sortBy: Ember.DEFAULT_VALUE })` will result in a URL that does not contain any `?sortBy=`.
```
nathanhammond: Rereading this now with a bit more context and explaining it for future travelers (read: me, six months from now).

This is to ensure that our default behavior of "hydrating unsupplied query params" continues to work. Without this we effectively end up doing `Object.assign({}, unsupplied, queryParams);` which would make `queryParams` only additive, which is wrong. We can't use values _other_ than this symbol as it is possible that it may be invalid.

1. I'm concerned about the name as `DEFAULT_VALUE` implies more than we likely mean to convey. I'd nominate `UNSET_VALUE`, `UNBIND_VALUE`, or `DELETE_VALUE` instead.
2. Can we stash this somewhere not on the `Ember` global?

nathanhammond: This gets worse if you nest arrays of objects and want to UNSET a property on a nested value.

```
{
  characters: [
    { thing: "one", hat: false },
    { thing: "two", hat: false },
    { thing: "cat", hat: true }
  ]
}
```

Indices get included because otherwise the array is ambiguous.
```
characters[0][thing]=one&characters[0][hat]=0&
characters[1][thing]=two&characters[1][hat]=0&
characters[2][thing]=cat&characters[2][hat]=1
```

If I wish to unset `characters[2][hat]` and eliminate `characters[0][thing]=one` (never liked that one anyway) I'm unsure it's possible to do gracefully with this set of proposed tools.

ef4: I think structured query params are additive API, so unless there is something in this proposal that forecloses on your design, I think we don't need to completely solve it here.

nathanhammond: They're additive in that they're new but considering our current state has severe issues and I want to ensure that we feel good about our future direction.

For the basic scenario and only-keyed-objects I think we can get away with `DELETE_VALUE`. To support arrays it almost has to become a Lisp where you specify the transforms. To do that gracefully I feel like we would need to adopt:
- `push`, `pop`
- `shift`, `unshift`
- `slice`
- `splice`
- ...

And that's beyond what I would want to support as it would come with a parser and all sorts of other annoying problems. Throwing away that proposal, I think it's bad.

---

We could always stamp something as "pre-merged" and expose an API to get the `unsupplied` values. It'll make structured query param merging palatable for advanced users, doesn't impact this simple scenario, doesn't add a tremendous amount of complexity and another Lisp-y microsyntax.

For bonus points I don't think that needs to be specified at this time, but seems like a direction where this doesn't paint us into any corners.

/FIN

locks:
I would like to thank everyone involved for shaping this RFC into its current state through detailed and constructive feedback.

This service has been a long time coming, and we hope that transitioning from the current intimate API is both swift and easy. On top of that, it will unlock new capabilities and optimization opportunities.

In this Final Comment Period only one new concern was raised, and it was considered that it is an additive concern that can be addressed by a follow up RFC to the present one. So all that's left to do is to merge it. 🎉 

One final thank you to @ef4 for sticking with it for the last year and change!

MiguelMadero:
Would it make sense to consider `hasRoute` as part of the public API? 
Today it's available on the ember router, router.js and the recognizer. 


cc: @kazuzenefits

rwjblue:
@MiguelMadero - I think it seems fine, but it should be done as a follow-up RFC.  It should be small enough and be able to focus on just this one method...

MiguelMadero:
Thanks @rwjblue I can throw something together. I just wanted to check first.

Fed03:
I think would be handy if we can expose methods like `transitionBack` and `transitionForward` handled by the service in order to navigate inside the ember app boundaries. what do you think?

sdhull:
Not sure if this is the right place to add comments for continuing work on the router service or not (considering this was merged but doesn't represent the entirety of what was described in https://github.com/emberjs/rfcs/blob/master/text/0095-router-service.md).

Anyway my 2c:
Rendering links for **pagination** is currently difficult in ember because there is no equivalent of what you find in Rails `url_for(params.merge(page: new_page))`. We get close with `routerService.urlFor` and `RouteInfo`, however there is no easy way (afaict) to get _all_ the params you need to feed into `urlFor`.

What I would like to see is the addition of something like `RouteInfo#urlArgs` (or some other name?) that would return an array containing `RouteInfo#name` (dot-separated fully-qualified name, eg `'people.show.posts.show.comments'`) & any models or model ids used to generate this route. Then you could do:

```
const queryParams = routeInfo.queryParams;
queryParams.page = newPage;
routerService.urlFor(...routeInfo.urlArgs, {queryParams: queryParams});
```

I guess alternatively `RouteInfo#models` would also be sufficient, as long as it includes _all the models (or ids) required_ for the give `RouteInfo#name`.

PR #96, opened 2015-10-02
Assisted ember upgrades

GavinJoyce:
[Rendered](https://github.com/GavinJoyce/rfcs/blob/gj/assisted-upgrades/text/0000-assisted-upgrades.md)

### DIFF
--- a/text/0000-assisted-upgrades.md
+++ b/text/0000-assisted-upgrades.md
@@ -1,0 +1,45 @@
+- Start Date: 2015-10-02
+- RFC PR: (leave this empty)
+- Ember Issue: (leave this empty)
+
+# Summary
+
+Manually updating Ember apps to remove depreciated syntax is time consuming and
+troublesome. We should automate this process where possible and make it a first
+class citizen in Ember.
+
+# Motivation
+
+Simplifying the Ember upgrade process is good for everyone. Application developers spend less
+time upgrading their apps and reduce the likelihood of introducing bugs. The Ember can be
+more ambitious with syntax changes and possibly have shorter depreciation timeframes allowing
+for a leaner framework codebase.
+
+# Detailed design
+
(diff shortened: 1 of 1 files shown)


text/0000-assisted-upgrades.md
```diff
@@ -0,0 +1,45 @@
...
+process for introducing new syntax and APIs which depreciate old ones should include
+consideration for automating the transformation.
+
+# Drawbacks
+
+Introducing a deprecation would take a little more effort.
+
+# Alternatives
+
+The [ember-watson](https://github.com/abuiles/ember-watson) addon provides some commands for
```
stefanpenner: using ember-watson, and giving it access to the HTMLBars AST transforms etc seems like a reason path forward.

ef4:
I think ember-watson is the right place to solve the problems you're describing.

Last I knew, the barrier to doing htmlbars rewriting was creating a whitespace-preserving AST printer. We actually already implement some of the transformations you'd want, they just happen at compile time now, and can't be easily printed back out as template files.

GavinJoyce:
Fair enough, I'll close this an contribute to [ember-watson](https://github.com/abuiles/ember-watson). Perhaps we can revisit this once it has more comprehensive coverage of depreciations? It would be nice to explicitly include new depreciations in ember-watson as part of the depreciation process, include links in the depreciation documentation and possible include ember-watson in ember-cli

rwjblue:
I don't think we will block new deprecations from being added on having automated upgrade tooling, but we definitely are blocking deprecations without an entry in the deprecation guide for the version in question. My goal would be to ensure that this guide has all the info required for someone to be able to add upgrade automation to ember-watson.

GavinJoyce:
@rwjblue Sounds good, thanks

GavinJoyce:
The first of many (hopefully): https://github.com/abuiles/ember-watson/pull/87

PR #97, opened 2015-10-05
[Route Prefetching] Allow routes to request data in parallel

nickiaconis:
_**Updated** to @stefanpenner's prefetch proposal:_

[Rendered.](https://github.com/nickiaconis/rfcs/blob/master/text/0000-prefetch-hook.md)

The demo is especially insightful (and has been updated to match the new RFC).
http://nickiaconis.github.io/ember-parallel-model-demo/


rwjblue:
:+1: - I love the goal and the general idea.

> This is a major breaking change.

The API's that are designed in this RFC need to be completely backwards compatible or this change has to wait until Ember 3.0 (which is quite a long way off). FWIW, I believe that it is possible to make this migration in a backwards compatible way with enough thought and planning...

Serabe:
:+1: What would happen if a transition is done in a `beforeModel` hook?

text/0000-parallel-model-hooks.md
```diff
@@ -0,0 +1,115 @@
...
+        this.transitionTo('fiz');
+      }
+    }
+  }
+});
+```
+
+# Drawbacks
+
+- This is a major breaking change.
```
chadhietala: I think a way of getting around the incompatibilities would be to create a new object type, that opts you into new semantics.  There is a precedents with this type of feature introduction with `GlimmerComponent`.

text/0000-parallel-model-hooks.md
```diff
@@ -0,0 +1,115 @@
...
+- Tie this paradigm change to something opt-in like routable components where people will already have to change their behavior.
+- Maintain it behind a feature flag on the Canary branch which you can toggle on and off via config until we can land it.
+- Provide some built-in hooks so that we're able to reach in and change this behavior as an addon in a safer way.
+- The worlds most absurd monkeypatch–which would also be packaged as an addon.
+- Maintain a patch and build our own custom Ember.
+
+# Unresolved questions
+
+
+- Should `modelFor` always return a promise? Should we instead add a `modelPromiseFor` method?
```
ef4: If we have two names, `asyncModelFor` would read much better than `modelPromiseFor`. Especially given ES7 await:

    await this.asyncModelFor('other.route');

ef4:
I'm good with the idea of running many routes in parallel. But I'm not ok with running all the hooks *within* a single route in parallel. `beforeModel`, `model`, and `afterModel` should still resolve in order, and each should wait for the one before. Otherwise `afterModel` becomes a terrible name that will trip people up.

If you want your model hook to start immediately, don't return a promise from your beforeModel hook.

FWIW, there's already a proposal in the routable components RFC to unify all three hooks and require `model` to be idempotent. I think that change and this change could go together and get triggered by the same opt-in upgrade.

text/0000-parallel-model-hooks.md
stefanpenner: the migration to this sounds rough, but I would prefer modelFor to always return a promise. Todays world is extremely zalgo

text/0000-parallel-model-hooks.md
stefanpenner: `AsyncRoute`

chadhietala: :+1:

rwjblue: > There is a precedents with this type of feature introduction with GlimmerComponent.

Agreed.  Definitely a possibility, but we (mostly @wycats, @tomdale, and @chancancode TBH) worked very hard to **not** need to make a different base class and discovered through that pain that it was unavoidable. It was definitely a last resort...

stefanpenner: @rwjblue yup.

I think a separate name, and provided via community add-on (with minimal hooks in internals) may be an escape valve. I want us to be careful not to cram ideas (good or bad) too quickly into ember. As once they are added, it is tricky and extremely costly to unwind.

We can look to several hasty (although well thought on 1.13.x features) as a demonstration of this.

I do believe the motivation here is good, the implementation may also. But it is well known the more users we can get using this in the wild the faster we can explore the problem space, and the more unknown unknowns we can squash.

stefanpenner:
There is also another approach, all hooks stay the same. But additional hooks are added purely for eager loading / prefetching.

Something like:

```js
export default Ember.Route({
  async prefetch(params) {
     // executed upfront, allowing each route to opt into pre-fetching semantics.
     // this should be thought of as nearly entirely declarative
     // one could imagine also providing graphQL esq (or actually) declarations here, that can be composed at runtime.
     
     return {
       user: this.store.find('user', await this.prefetched('parent.route').user.id),
     }
  },

  model(params) {
    // .. executed normally, after the route is entered and the routes own prefetch has completed
    // relying on identify maps and such which have pre-fetched.
    return this.store.find('user', params.user_id)
  }
]);
```

This makes prefetching:

* purely additive
* opt-in
* a separate DSL that aims to NOT expose existing zalgo API.
* easy to migrate a code-base slowly, without totally parallel buy-in.

The algorithm would be as follows:

1. begin routing
2. recursively invoke all descendent routes `prefetch` assuming the developer has correctly de-zalgoified.
3. begin entering routes, as there corresponding pre-fetch and parents pre-fetches have been invoked

This should:

* provide the concurrency option that prevents unintentional sequential route waterfall. 
* work nicely given existing solutions and apps.

ef4:
@stefanpenner I think that whole idea can be done in userspace using public APIs once we ship the changes proposed in https://github.com/emberjs/rfcs/pull/95.
Your `application` route's `beforeModel` hook will receive a well-defined structure that tells you exactly what routes are about to get activated, and all their parameters. At that point it's easy to do something like call your own custom hook on each of the route classes.

I think that's a good solution for teams that are itching for a solution right away. We would still want to iterate to a fully baked-in solution that will Just Work.

text/0000-parallel-model-hooks.md
nathanhammond: To help explore the problem space I'd love to get an early version in behind a feature flag so we have fewer unknowns. It seems like it would be a good fit to come up with an API for this sort of behavior at the same time as trying to solidify routable components so that they play nicely together. It also gives us a `GlimmerComponent`/`"use strict";` type of escape valve for opting in if we can't come up with a better way to handle it.

The option of building an addon has community effects that will be quite painful. We see tight coupling between liquid-fire and Ember that an addon to accomplish this would make look positively decoupled. Self-trolling will occur all over the place.

text/0000-parallel-model-hooks.md
nickiaconis: :+1: to `asyncModelFor` as the method name if we go with a new method.

@stefanpenner I'm not sure a different subclass would help with the `modelFor` question. You'd still be calling `this.modelFor('foo.bar');`, which would return a model or a promise depending on whether `FooBarRoute` was a `Route` or an `AsyncRoute`. Seems rather opaque.

nathanhammond: One of the things that we're considering as part of this process is enforced disallowal to mutate state that can cause trouble during these hooks by creating a special `this` object/context. If we run things in this special context while this is feature-flagged we'll be able to identify the types of things people want to do and provide APIs for that behavior.

This is also a first step toward optional removal of dynamic dependency injection (think, addon) which will allow for fully-resolving dependency graphs at build time for things like app partitioning without strict development time partitioning with engines. (Otherwise anything which _could_ be dynamically injected will have to be part of the primary payload.)

nickiaconis:
@ef4 I agree with you that `beforeModel`, `model`, and `afterModel` should run sequentially or be replaced by a single idempotent hook.

Can you give an example of using the API from #95? I don't follow.

@stefanpenner Interesting idea to add an additional hook for eager loading. Would it rely on Ember Data caching to get the prefetched data in the `model` hook? I'm not sure I like that as it doesn't provide a mechanism for people using `store.query` or not using Ember Data at all.

mmun:
The prefetch result could be made available in the transition object passed to the model hook, but it feels like a lot of hoops to jump through.

nathanhammond:
> If you want your model hook to start immediately, don't return a promise from your `beforeModel` hook. -@ef4

This is the recommendation that we've given for a long time even with `model` hooks: if you want instant rendering effectively skip the hooks and reassign properties on the object you returned as promises resolve. We lose things like nice loading route behavior if we adopt that approach and it becomes unnecessarily complicated to know when you're able to do certain things (you have to pass around references to promises).

In my opinion the thing that distinguishes `afterModel` simply becomes that it has a reference to the `model` promise passed in as an argument. (This will be more important if you have limited context to prevent state mutation.)

As more of a power user I'm of course in favor of running all the things in parallel, but I recognize that the consequences of that for the majority of the Ember ecosystem are nontrivial. Unification under routable components neatly solves this problem.

nathanhammond:
@stefanpenner The API you propose is interesting, but also assumes that the only async behavior we're dealing with is loading of data. It could also be writing to disk (NW.js) or a long-running process that we handed off to a WebWorker. Granted, those are less likely and we should aim for making sure that the happy path is optimized. Even in the world where we accounted for all use cases it would still likely make sense to have a `data` hook which can be invoked on all routes to build the data dependency tree.

For example, @nickiaconis implemented multiplexing on top of Ember Data already which is one of the primary motivations for this RFC: collating across route boundaries. Discussions are being had about how to do client-side composition of data models in the vein of Falcor/GraphQL such that it is a fully-resolved data-dependency tree instead of still requiring round trips between tree "layers" (depth from root node(s)).

stefanpenner:
> @stefanpenner Interesting idea to add an additional hook for eager loading. Would it rely on Ember Data caching to get the prefetched data in the model hook? I'm not sure I like that as it doesn't provide a mechanism for people using store.query or not using Ember Data at all.

Either a identify map, or other hooks could use `this.prefetched` API as well, to scope data from the prefetched hooks

nathanhammond:
@stefanpenner We could turn the whole prefetch hook into a DSL which we can serialize into Ember Data calls, lisp-like functional transformations that could be passed to the server a la GraphQL, or into vanilla Ajax calls. We just design/adopt a way to describe what data you need as a fully resolved tree and then let some serialization of that definition inform your `DataAdapter` "here is what I need." How it goes and gets that data is then entirely an implementation detail which makes the public API surface much easier to use. (No need to know about `await this.prefetched` or whatever we come up with.)

stefanpenner:
@nathanhammond yes I believe `prefetch` (or very like some more aptly named `hooks` are primitives to allow for this.

My motivation is to find some stepping stones which bridge the gap between declarative/prefetch ideas and the existing world. I believe the existing world has many merits and strengths, we would be wise not to fall into the trap of believing a drastic departure is required.

The problem as I see it takes the following shape: 

* GlobalOptimization === new & missing
* local reasoning === existing & works tolerably well

It seems like both of these can complement each other nicely. In-fact it should be possible to minimize changes to the existing, while still reaping the rewards of a globally optimized system.

1. generic pre-fetching based on declarative DSL is often possible and can provide high value
2. generic pre-fetching in conjunction with turing complete language can fill in the remaining gaps
3. pre-fetching in conjunction with an existing identify map provides an additional win.
4. `activate` -> `before` -> `model` -> `after` -> `setupController` ... `deactivate` hooks provide a DSL for the per route domain, if 1-3 do there work 4's current latency concerns should be mitigated.

Doing this may prove some existing API / timings redundant, which means in 3.0 they can be removed.

dgeb:
> We could turn the whole prefetch hook into a DSL which we can serialize into Ember Data calls, lisp-like functional transformations that could be passed to the server a la GraphQL, or into vanilla Ajax calls. We just design/adopt a way to describe what data you need as a fully resolved tree and then let some serialization of that definition inform your DataAdapter "here is what I need." How it goes and gets that data is then entirely an implementation detail which makes the public API surface much easier to use.

@nathanhammond This is almost precisely what I'm working on in Orbit.js, which will soon have a lisp-like customizable query expression language. More details here: https://github.com/orbitjs/orbit.js/issues/212 

This approach allows for complex, customizable, and synchronous queries of in-memory data, which can be asynchronously "hydrated" from remote sources. When used with Ember, this will allow many model hooks to be resolved synchronously while still asynchronously fetching data. The actual fetch calls could be batched, as you suggest, depending on the capability of remote sources (GraphQL, JSON API, etc).

Of course, there are still going to be remote calls that should block route loading, in which case a promise should be returned to the model hook. But for many cases, synchronous resolution of in-memory queries combined with asynchronous fetching of remote data will provide the best user experience.

wycats:
Here's what I think we should do:

1. Add a new behind-a-flag escape valve that allows a route to ask Ember not to wait for its async hooks to complete before continuing with the next steps.
2. Add new behind-a-flag APIs (`asyncModelFor`, etc.) methods that allows a route to manually wait for a previous step to complete.

Once these behind-a-flag APIs are in place, we can get more real-world feedback on what patterns (other than the obvious ones) emerge, which we can feed into the design process for a new public API.

For what it's worth, I am very confident that the public version of this API requires `async function` to be ergonomic, which is at Stage 3 in TC39, supported by babel, and we are getting ready to adopt.

tomdale:
Came here to chime in that I think different _routes_ should load in parallel but am opposed to `beforeModel` and `afterModel` becoming forced into async as well, though I see @ef4 has already said the same thing more eloquently than I could. It seems like there is consensus on that point, which is great.

@wycats I am a little nervous about putting the functionality in behind a flag if we *know* that is not the final, desired API. Mostly because people will likely come to rely on it, and it may be a large refactor to go from e.g. a flag on existing routes to an entirely new `Route` base class, as someone proposed above.

wycats:
> @wycats I am a little nervous about putting the functionality in behind a flag if we know that is not the final, desired API. Mostly because people will likely come to rely on it, and it may be a large refactor to go from e.g. a flag on existing routes to an entirely new Route base class, as someone proposed above.

I think we need to learn more from users who get real benefits from parallelism before we can design the API. Today, users in that boat are monkey-patching into private APIs, which I think you can agree have a worse version of the problem you're worried about than a few known flagged APIs.

wycats:
TLDR, in TC39 terms: let's move this proposal to Stage 0 ;)

rwjblue:
> I think we need to learn more from users who get real benefits from parallelism before we can design the API.

Agreed.

> Today, users in that boat are monkey-patching into private APIs

When folks monkey patch private API's they are clearly taking on any future issues themselves. This is not true of feature flags that have landed in core (where folks report and expect bugs to be fixed by the magic OSS fairy).

> TLDR, in TC39 terms: let's move this proposal to Stage 0

Agreed, making an addon that provides the strawman proposal sounds wonderful!

nickiaconis:
> > TLDR, in TC39 terms: let's move this proposal to Stage 0

> Agreed, making an addon that provides the strawman proposal sounds wonderful!

There have been several suggested alternatives to the original proposal. What are we moving forward with? I can update the RFC to reflect what it is we want to be implemented as Stage 0.

nathanhammond:
> I think we need to learn more from users who get real benefits from parallelism before we can design the API.

I'm not sure how to get our target audience (mid-size applications, moderately familiar Ember devs) to test this since everybody commenting on this thread has large apps and already knows the internals of the router. We're all terrible as case studies.

> Agreed, making an addon that provides the strawman proposal sounds wonderful!

Our tradeoff:
- Addon which can immediately be used by anyone using Ember without specifically setting a feature flag or really understanding the risks. Broader test base. (This would be implemented by monkeypatching to accomplish exactly what @nickiaconis has as a diff, except with more copypasta.)
- Behind a feature flag on Canary for people who are already comfortable being on the bleeding edge and won't complain if things break. Smaller test base. (Maybe @nickiaconis can be the magic OSS fairy. :smile:)

I feel like the second one has less of a risk of becoming the next vendor prefix debacle, but that's just me.

wycats:
> When folks monkey patch private API's they are clearly taking on any future issues themselves. This is not true of feature flags that have landed in core (where folks report and expect bugs to be fixed by the magic OSS fairy).

I don't feel like this is what happens. When an addon that uses private APIs becomes popular, it becomes just as much an entitlement as an opt-in flag, but more so.

rwjblue:
> I don't feel like this is what happens. When an addon that uses private APIs becomes popular, it becomes just as much an entitlement as an opt-in flag, but more so.

I respectfully disagree.

wycats:
> I respectfully disagree.

Liquid Fire users have not felt like they were to blame for using private APIs, while users of the unstable `visit` API absolutely expected churn. I'm not sure where the disagreement is coming from.

rwjblue:
This is a great example, `liquid-fire` users clearly understand that issues they have/see are due to an issue with liquid-fire (there are very few issues reported to emberjs/ember.js that are due to issues in liquid-fire), while fastboot/visit API users are often reporting issues to Ember (rightfully so, as much of the code to blame for issues they are having is in Ember).

nickiaconis:
Ping.

> > > TLDR, in TC39 terms: let's move this proposal to Stage 0

> > Agreed, making an addon that provides the strawman proposal sounds wonderful!

> There have been several suggested alternatives to the original proposal. What are we moving forward with? I can update the RFC to reflect what it is we want to be implemented as Stage 0.

nickiaconis:
@stefanpenner -- The more I think about it, the more I like your idea of a `prefetch` hook. I'm going to update the PR based on that.

Also, to get a feel for it, I've taken a stab at implementing prefetch as an addon. It's almost working, but I've run into an issue. `Router#willTransition` isn't re-fired when a transition redirects. Any ideas how to work around that or maybe another Router hook from which to call the `prefetch` methods?

nickiaconis:
Updated to @stefanpenner's prefetch proposal.

[Rendered.](https://github.com/nickiaconis/rfcs/blob/master/text/0000-prefetch-hook.md)

The demo has also been updated to match the new RFC:
http://nickiaconis.github.io/ember-parallel-model-demo/

nathanhammond:
This is far less invasive, feels at home in the existing programming model, requires opt-in (and is thus backwards-compatible) and allows for us to be incredibly clever with the prefetch hook (think GraphQL).

@nickiaconis Can you address how some likely future patterns would work inside of this API? I'm not sure we've covered the use case of having child route nodes able to inspect and/or modify the "parent" `prefetch` or specify how the resolution should work.

To attempt turning that last statement into questions, if you were using this to build up a single outbound query which would be resolved via multiplexing (either streaming or blocking) how would that work? What if you were building up a single request a la GraphQL? How would you know which portion of the data object to feed into which model hook? (We can talk in person about the approach we've got internally for that behavior.)

nickiaconis:
@nathanhammond Feeding data to the right place should be the responsibility of the multiplexing or query language solution. I would imagine a call to a data fetching API would return a promise that resolves with the requested data. Ember should be data-API-agnostic; the only assumption `prefetch` makes is that the data API returns a promise that resolves with the data. This assumption is already made by the `model` hook.

nickiaconis:
The more I think about it, the less I'm convinced we need a `prefetched` method. All use cases I can think of for `prefetched`/`asyncModelFor` subvert the optimization gained by using the `prefetch` hook. If a route needs access to a parent route's data in order to make a request, it should just use the `model` hook, at which point `modelFor` fills this role.

The only exception is getting a route's own data, for which a property could be used. We could still call it `Route#prefetched`, but it would be a property. Its value would be a promise if anything was returned from the `prefetch` hook, otherwise `undefined`.

nickiaconis:
Feel free to try it out: https://www.npmjs.com/package/ember-prefetch

I've implemented it with `Route#prefetched` as a property rather than a method to see if anyone finds a need for the method version. The one caveat is that it will cause transitions to abort early when you redirect (in order to retrigger `Router#willTransition`).

cibernox:
I was about dropping an idea but I see that @wycats just read my mind.

It seems to me that the current `beforeModel`/`model`/`afterModel` makes perfect sense. 

It doesn't sound very backwards compatible but the way I'd implement this is with a feature flag that makes all routes to load in parallel (that defaults to false until Ember 3.0) and those routes that for some hierarchical reason need access to the model of some parent route in order to get their own can get it in an asynchronous fashion with a new `asyncModelFor`.

Perhaps it might come handy, specially for migration purposes, a way to flag a route as `blocking`. Particularly I'm thinking that this would be useful for the top-most route or the "authenticated", because is an extremely common pattern to load the current user there and just rely on it's existence in all other routes.

dgeb:
I also agree with continuing to use the current model hooks. I'd suggest `fetchModelFor` instead of `asyncModelFor`.

That leaves the question of how to indicate whether the loading of a route's model hook is blocking. Let's consider making this an option for route mappings (in `router.js` at first, `routes.js` once engines are implemented) instead of on routes themselves. It strikes me as an inter-route, rather than intra-route, concern. Of course, the default `blocking` option value would be `true` and could be overridden (I'm not sure if `blocking` is too generic of a name for this option - we could get more specific).

cibernox:
@dgeb Don't you feel that pushing this `blocking` (or whatever name we choose) to the routes would push that configuration too far away from the entity it applies to?

dgeb:
> Don't you feel that putting this blocking (or whatever name we choose) to the router.js would push that configuration too far away from the entity it applies to?

@cibernox Just to be clear (and I think we're on pretty much the same page), I see two types of blocking scenarios:

I think the most common blocking scenario will be the one in which data dependencies between routes block the loading of child routes. This configuration pretty clearly belongs in child routes. If we have a `fetchModelFor` (or `asyncModelFor`) method at our disposal, then we can use it in a child route's model hooks to ensure that a dependency upon a parent route's data (or multiple route's data) is resolved before proceeding to fetch the child route's own data.

The less common blocking scenario will be the one in which child routes are fully blocked from starting to load until a parent route has fully resolved. This is not obviously a concern of either the parent or child route - in any given app it seems that it could be either, or perhaps result from a combination of the two. That's why I suggested that this type of blocking be specified in the routing map which describes route relationships. Maybe we'll need to feature flag it and experiment to gather more use cases and discover best practices? Regardless of where we put it, we all seem to agree about the need for some sort of flag or overridable hook to maintain the current full blocking behavior.

nickiaconis:
Hi @cibernox & @dgeb, I've built and tested both versions: making the existing model hooks run in parallel and adding a new `prefetch` hook. (You will find the previous version of this RFC described the former.) Having a separate `prefetch` hook has the advantage of being an additive API change as well as making more semantic sense for this sort of behavior.

Can you see if [ember-prefetch](https://www.npmjs.com/package/ember-prefetch) fulfills your use cases? A route with no data dependencies should be able to opt-in simply by renaming its `model` hook to `prefetch`.

A method like `fetchModelFor`/`asyncModelFor` is not provided because its use cancels out the performance benefit of running the hooks upfront. Routes with data dependencies should just use the existing `model` hook wherein `modelFor` fills this role. The `model` hook abstracts away the promise waiting required to depend on a parent route's data, like

```javascript
model() {
  var fooModel = this.modelFor('foo');
  // make requests based on fooModel
}
```

which seems more straightforward than

```javascript
model() {
  this.fetchModelFor('foo').then((fooModel) => {
    // make requests based on fooModel
  });
}
```

or even

```javascript
async model() {
  var fooModel = await this.fetchModelFor('foo');
  // make requests based on fooModel
}
```

nickiaconis:
Two quick comments:

1. I've updated the RFC to reflect `prefetched` being a property on the route, rather than a method.
2. I've updated [ember-prefetch](https://github.com/nickiaconis/ember-prefetch) to remove the `transition.abort()` on `Router#willTransition`. It was causing issues, and there was an easier way to achieve the same result. It still changes when `Router#willTransition` is triggered, but does so in a less disruptive manner.

text/0000-prefetch-hook.md
```diff
@@ -0,0 +1,79 @@
+- Start Date: 2015-10-01
+- RFC PR: (leave this empty)
+- Ember Issue: (leave this empty)
+
+# Summary
+
+The waterfall way in which Ember resolves a transition's model hooks can result in unnecessary blocking.
```
stefanpenner: some quick thoughts as i read.

Embers typical inter-route data-fetching is sequential, although this has upsides (simpler programming model) the resulting latency of each request is compounded and can result in a degraded experience. The general solution to the "waterfall" problem of sequential requests, is to execute independent requests in parallel.

Automatic inference of which requests can be or can't be done in parallel is tricky, and likely impossible while maintaining a semver compatible public API. Although it may not be automatic, a developer has sufficient context to tune the parallel/sequential steps. 

This RFC proposes an approach allowing a pro-user to tune the sequential vs parallel....

text/0000-prefetch-hook.md
```diff
@@ -0,0 +1,79 @@
...
+});
+
+App.PostCommentsRoute = Ember.Route.extend({
+  prefetch(params, transition) {
+    return Ember.$.get(`/api/posts/${transition.params.post.id}/comments`);
+  }
+});
+```
+
+The default functionality of the `model` hook is modified to return the prefetched data if it exists.
```
stefanpenner: I'm unsure if we should modify the existing/default `model`. It is unclear to me that the relationship between prefetch and model is actually 1:1

given the following scenario:

```js
// routes/user.js
export default Route.extend({
  prefetch(params) {
    return RSVP.hash([
       user: this.store.find('user', params.user_id),
       friends: this.store.find('friends', { user_id: params.user_id }),
       somethingElse: ajax(...)
    })
  },

  // the default model functionality of `this.store.find('user', params.user_id)` will "just work" and the ED identify map would benefit from the existing fetch.
})
```

I would imagine, adding / removing prefetch request would be common, as the the UI changes or the data topology / size changes, tuning will occur.

In addition, query-params may or may not go through the pre-fetch if model refreshes.

If the model wants access to the prefetched data, it should likely be explicit.

```js
export default Route.extend({
  prefetch()  { ... }
  model() {
    return this.prefetched();
  }
})
``

```js
export default Route.extend({
  prefetch()  { ... }
  model() {
    return this.prefetched('other-route');
  }
})
````

It almost seems like `prefetched` concept could be symmetric to modelFor, but always returning a promise.

This would also allow an advanced user to create additional async dependencies while `prefetching`

maybe for example:

```js
export default Route.extend({
  prefetch() {
    // although this now creates a sequence, prefetch firing concurrently at the start of routing,
    // enables these to speed ahead, but truthfully support inter-route prefetching dependencies.
    return this.prefetched('otherRoute').then(other => {
      return this.store.find('user', other.id);
    })
  }
})
```

text/0000-prefetch-hook.md
```diff
@@ -0,0 +1,79 @@
...
+As such, a route that defines a `prefetch` hook is not required to define a `model` hook.
+
+A `prefetched` property is added to `Route`.
```
stefanpenner: based on my above example, I am inclined to recommend this be a function (or a sibling function symetric to modelFor but always async be added)

text/0000-prefetch-hook.md
```diff
@@ -0,0 +1,79 @@
...
+`prefetched` can also be used to get a route's own data.
+This is useful for expanding upon the prefetched data in the route's `model` hook.
+
+```javascript
+App.PostCommentsRoute = Ember.Route.extend({
+  prefetch(params, transition) {
+    return Ember.$.get(`/api/posts/${transition.params.post.id}/comments`);
+  },
+  
+  async model() {
```
stefanpenner: ^

text/0000-prefetch-hook.md
```diff
@@ -0,0 +1,79 @@
...
+
+- Ember's API becomes larger.
+
+# Alternatives
+
+- Provide something like `Router#willTransition`, but that is triggered on redirects, so this can be built as an addon. It could be triggered on either all transitions (`Router#willChangeURL`?) or only redirects (`Router#willRedirect`? to supplement `willTransition`). Otherwise, building this as an addon must change the functionality of `Router#willTransition` in order to function properly.
+
+# Unresolved questions
+
+- Is there a use case for the API to include a method like `modelFor`, but for `prefetch`? (`prefetchedFor`/`asyncModelFor`)
```
stefanpenner: yes, I like `prefetchedFor` (my above comments may hint that this fact)

This could allow for `this.prefetched` property as you describe above to co-exit.

`asyncModelFor` may also be legit addition, but it feels maybe orthogonal to the prefetch world, but I could be convinced.  I am still sad the `modelFor` itself isn't already the async variant.

stefanpenner:
I suppose with transitions, we may cause many pre-fetches to fire that may or may not be needed. I suspect this is ok, one cancellation is added to promises we could explore auto-canceling pending and unrelated pre-fetches..

stefanpenner:
@nickiaconis was my proposal sufficient in handling the original use-cases? It would be great if it was, and if not I would like to know what (if any) issues remained.

stefanpenner:
One final note, i believe the parallel escape valve is a big missing piece. I believe with some investigating it could be used to better integration with solution such as graphQL etc.

text/0000-prefetch-hook.md
cibernox: > It almost seems like prefetched concept could be symmetric to modelFor, but always returning a promise.

I think that Yehuda has the one that first suggested this, proposing something alike to `modelFor` that returns an promise. I did like this idea because it preserves the beforeModel/model/afterModel methods we already know.

Couldn't just adding `async` before the `model` hook be enough to make the route non-blocking? And if in child routes you need access to this model you can access it with a `asyncModelFor('parent.route').then(model => ...)`?

I think that this would add the very minimum cognitive syntax and cognitive overhead to the existing model.

nickiaconis: @cibernox The way this is written right now, execution would go like: prefetch1, prefetch2, beforeModel1, model1, afterModel1, beforeModel2, model2, afterModel2.

The async function syntax essentially turns a function into a generator, which yields a promise each place there is an await. So adding async in front of the model hook will basically make it return a promise, which  Ember already assumes it does. It will not cause it to execute outside of the promise resolution chain that is a part of `TransitionState#resolve` in tildeio/router.js

text/0000-prefetch-hook.md
nickiaconis: I'm beginning to come around again on the idea of `this.prefetched('otherRoute')`. Previously, all examples I had seen or devised had involved a route that was a direct descendent of the parent whose data it needed, in which case using the `model` hook is mostly synonymous with using the `prefetch` hook and awaiting a call to `prefetched`. In the case when the child route is the second descendent (or further removed), it makes sense to be able to do this in the `prefetch` hook.

I worry that a `Route#prefetched` method will be (ab)used to get values that are already found in the parameters of parent routes. The correct way of doing that is to use `Route#paramsFor`, but I'm not sure it's ubiquitously known as evidenced by this [SO question](http://stackoverflow.com/questions/22376994/ember-deeply-nested-routes-do-not-keep-parent-dynamic-parameter/27871764), two of the answers for which suggest using the parent route's model. And then there's [this tutorial](https://www.airpair.com/javascript/emberjs-using-ember-cli#6-direct-access-to-router-states) that instead uses the private API of the transition.

However, someone who would to use the `prefetched` method to get a parameter value would probably also use the `model` hook with `modelFor` instead of the `prefetch` hook if the `prefetched` method is not provided.

nickiaconis:
@stefanpenner the variation of your proposal that I've implemented in ember-prefetch has handled our original-use cases well. However, there are unforeseen implications to redirection that we've discovered along the way. For instance, some routes are "non-terminal" in that a transition to them will always cause a redirect to some other "terminal" route (possibly via other "non-terminal" routes). This is troublesome for prefetch since the pivot handler is not changed when the first transition is aborted, and the new transition causes prefetch hooks to be run again. (Writing this now has given me an idea of how to work around it, which I will try in short order.)

text/0000-prefetch-hook.md
stefanpenner: I still think the this.prefetched(x) or similar is an important part of this pattern, would be open to persuasion .. maybe thoughts have changed since these comments were made

text/0000-prefetch-hook.md
```diff
@@ -0,0 +1,85 @@
...
+```javascript
+App.PostCommentsRoute = Ember.Route.extend({
+  prefetch(params, transition) {
+    return Ember.$.get(`/api/posts/${transition.params.post.id}/comments`);
+  },
+
+  async model() {
+    return {
+      OP: this.modelFor('post')).author,
+      comments: await this.prefetched
```
stefanpenner: are we not already guarenteed that the prefetch has completed before the model hook is invoked?

also, i do feel this was to be symmetric to model-for

text/0000-prefetch-hook.md
nickiaconis: Yes, thoughts have changed. The ember-prefetch add-on even implements this.prefetched(x) now, but it would appear I never updated the RFC. Will fix.

stefanpenner:
With the recent changes, and the addon in the wild. I would like to explore up streaming this. @nickiaconis feel free to pester me.

Il start by bringing this up at this Friday's meeting and we can go from their.

nickiaconis:
Edited again to change `prefetched` to a method and to update the information about the alternative implementation as an addon.

text/0000-prefetch-hook.md
```diff
@@ -0,0 +1,82 @@
...
+      OP: this.modelFor('post')).author,
+      comments: await this.prefetched()
+    };
+  }
+});
+```
+
+# Drawbacks
+
+- Ember's API becomes larger.
```
stefanpenner: it may be confusing to the programming model

text/0000-prefetch-hook.md
```diff
@@ -0,0 +1,82 @@
...
+
+# Alternatives
+
+- Implement the functionality as an addon.
+  - Would require Ember to provide something like `Router#willTransition` that is triggered on redirects. It could be triggered on either all transitions (`Router#willChangeURL`?) or only redirects (`Router#changeTransition`? to supplement `willTransition`).
```
stefanpenner: it seems like these hooks are required regardless, even if this where part of ember no?

text/0000-prefetch-hook.md
```diff
@@ -0,0 +1,82 @@
...
+
+Currently, the `model` hook is Ember's way for routes to request data.
+However, a route's `model` hook is not entered until it's parent route has resolved.
+This is suboptimal for nested application structures.
+Child routes are forced to wait for their parents to receive data before they may make requests for their own data.
+This occurs even when the child doesn't make use of its parents' data.
+
+This RFC proposes adding a new `Route#prefetch` hook, which is used to allow routes to make requests in parallel.
+Child routes that make use of `prefetch` will resolve faster since their data can resolve at the same time as (or before) their parents'.
+The semantics and ordering of `Route`'s existing model hooks (`beforeModel`, `model`, `afterModel`) are preserved.
```
stefanpenner: I wonder if prefetch cause become part of the chain, essentially invoked eagerly, but its completion blocks beforeModel on that route. This may actually be least surprising. (maybe, needs thought)

stefanpenner:
Im not quite sure (from reading this) what happens on refresh of a route, like when a QP changes and the model hook is intended to be run. 

Or what happens on a link-to (with model, not id provided) pivoting on the route which has a prefetch.

text/0000-prefetch-hook.md
```diff
@@ -0,0 +1,82 @@
...
+
+```javascript
+App.PostCommentsRoute = Ember.Route.extend({
+  prefetch(params, transition) {
+    return Ember.$.get(`/api/posts/${transition.params.post.id}/comments`);
+  },
+
+  async model() {
+    return {
+      OP: this.modelFor('post')).author,
```
stefanpenner: extra parenthesis

tomdale:
@nickiaconis We discussed this in the core team meeting today and we had some concerns with the RFC as is. @stefanpenner wanted to make sure this didn't languish, so he asked me to write the concerns down. I speak only for myself but have tried to incorporate the problems other core team members identified. I'll start by laying the philosophical groundwork then get down to the merits of this specific RFC after.

For me, the high-level concern is this: as software libraries grow, there's a tendency to add additional API as new use cases (and deficiencies in the original design) are discovered. It is very tempting to keep merging small, targeted new APIs to solve pain points that developers are facing in the real world.

This is great for advanced users; learning a series of small changes is easy because it's amortized over months or years. For new users, though, it make the framework feel like an incoherent soup of complexity. They have a hard time creating a strong mental model and they may abandon the framework altogether.

Angular 1 is a good example of this: I see new users frequently confused by services, factories and providers. They solve similar problems and you have to understand their subtleties to know which one to use. And directives have so many knobs to turn (transclusion, scope, `restrict`, etc.) that many people consider them to be the domain of only power users.

The way to fix this is to periodically gather up all of the constraints, including the new ones discovered since the original feature was developed, and see if you can find a new Grand Unified Feature that solves all of the constraints with one concept instead of several. Many of us on the core team think that it is probably time for that to happen for model loading in routes.

**Rather than adding a parallel `prefetch()` hook to supplement the serial `model()`, we would like to investigate if there is a new API that can solve the use cases for both with one concept/hook.** Then we can stop telling people to use `model()` and use a new API that does the right thing when their models are serial and the right thing when they're parallel.

For example, perhaps we replace the `model()` hook with one called `fetchModel()`. It works like `model()` now except that they are called in parallel. In addition, we add a `waitForModel()` method that returns a promise for a parent route's model for the cases where you need it to be serial. That way, a serial route might look like this:

```js
// app/routes/post/comments.js
export Route.extend({
  fetchModel() {
    return this.waitForModel('post').then(post => {
      return $.getJSON(`/posts/${post.get('id')}/comments.json`);
    });
});
```

This proposal almost surely has flaws that need to be ironed out, but it's an example of what I mean: rather than having a serial model pipeline and a parallel model pipeline, we try to find _one_ pipeline that can do the right thing based on what the developer expresses inside.

I think the best way to move forward is to create a document that enumerates use cases from the real world. Offhand, that includes at least a model that depends on a parent, two models that don't have any dependencies, and a GraphQL example. Once we have a list of the use cases, we can see how the proposed API looks in context. As it is, though, I think the current RFC has way too little detail and consideration of all use cases for something that touches such a fundamental part of the programming model.

text/0000-prefetch-hook.md
nickiaconis: It's possible that it _could_ be implemented without these hooks, but it's doubtful that it would be preferred. Doing so would likely involve tight coupling with the router, which is no bueno.

Since these hooks are useful for other purposes too (accessibility, advanced tracking, etc), I'd say they're necessary regardless.

text/0000-prefetch-hook.md
nickiaconis: Need linting for code blocks in markdown. :stuck_out_tongue:

text/0000-prefetch-hook.md
nickiaconis: Sounds kinda like what @tomdale has described [below](https://github.com/emberjs/rfcs/pull/97#issuecomment-180560394), though he suggests it replace `model()` outright. Not sure how he imagines it would interact with `beforeModel()` and `afterModel()` or if these would even still exist in such an API.

stefanpenner:
> The way to fix this is to periodically gather up all of the constraints, including the new ones discovered since the original feature was developed, and see if you can find a new Grand Unified Feature that solves all of the constraints with one concept instead of several.

@tomdale If on every new discovery we have to re-invent the world, we end up with mega churn, as the paradigm shifts (rather then simply evolving it). As what happens (as evident in prior such re-works) the grass isn't actually any greener on the other side, instead a new fancy paradigm appears, shifting the problem space rather then reducing it.

We should first expand (carefully), before re-inventing the whole model.

I do not believe supporting this in core requires a rework of the entire programming model, in-fact with some more massaging of what is proposed here complements the existing model nicely.

My suggestion is that we massage this RFC, let it ride the release trains, get feedback. A year from now, when we have time/energy/motivation to improve the routing DSL we can use additional knowledge it provides to make better and more informed choices in a redesign (if that is even required).

----

It seems like the mood is surprisingly negative, if that is the case. Maybe we should explore codifying the exact required missing events/API hooks to implement this well in user-space and call it a day. This would be unfortunate.

nathanhammond:
I'm a bit surprised at the degree to which you have concerns, @tomdale. We have a [working implementation of this in production for LinkedIn](https://www.linkedin.com/m/), a [popular addon implementing this RFC](http://emberobserver.com/addons/ember-prefetch), and a [pretty compelling demo](http://nickiaconis.github.io/ember-parallel-model-demo/). In addition, internally, @nickiaconis built a dynamically constructed request multiplexer on top of this simple `prefetch` primitive which will play nicely with our upcoming GraphQL-like tool we call Deco. All told, this little feature enabled multiple seconds worth of performance improvement for our application, and will go a long way toward dispelling the notion that Ember is slow.

Philosophically I disagree with you as I'm concerned that a 2 => 3 transition where we move to something like `fetchModel` will add just as much complexity with the added cost of breaking backwards compatibility. Or, if we don't break backwards compatibility, is identically as complex as this `prefetch` proposal. As this stands it simply provides an escape valve for parallel behavior, is entirely backwards compatible, and achieves the "stability without stagnation" goal. We've spent months honing this solution internally and no matter what, to solve this problem, we felt like we needed to expose parallel and sequential primitives to the framework user.

On the topic of "stability without stagnation," a non-blocking model hook has been a feature that people have been asking for since day one. So far the answer has been "just return a plain object and set properties on it later" which is pretty clearly an inferior version of this solution. Not providing a primitive for this behavior (which has been desired since before 1.0!) until the 3.0 timeline feels like a disservice to users of Ember. I truly feel like we should address this now.

Regardless, @tomdale, do you hold the same reservations about adding the underlying primitives necessary to build this in user space? Something like `willChangeTransition` is the only thing which has to be monkey-patched in, the rest just relies on private (but accessible) API surface area.

mixonic:
FWIW I'd like to see what a hook supporting the addon and use-case would look like. If it really is trivial, it could go straight to a PR and feature flag and likely be discussed/merged from that. Small win, I know, but potentially a win.

tomdale:
@nathanhammond Trying to amp up the sense of urgency is not going to be a winning strategy. The pain is real and I want to fix it. I am happy that you have a proof-of-concept that is working well for you. However, I've been doing this long enough now that I know that rushing in features to solve pain (no matter how acute) ends up being worse in the end, even if I can't yet guess how.

As @mixonic identified, the best path going forward is to figure out the minimum set of hooks we need to expose via public API to let an addon implement this in a forward-compatible way. That will help us eliminate the pain as rapidly possible while buying us time to have a thoughtful discussion about the programming model.

tomdale:
@stefanpenner 

> If on every new discovery we have to re-invent the world, we end up with a different but also high cost learning hazard, as the paradigm shifts (rather then simply evolving it). As what happens (as evident in prior such re-works) the grass isn't actually any greener on the other side, instead a new fancy paradigm appears, shifting the problem space rather then reducing it.

Maybe this is semantics, but I feel like my proposal is much more of an evolution of the current model than the `prefetch` proposal. It literally is creating a new hook that is semantically identical to `model()` except that it gets invoked in parallel instead of serially, and provides a helper to create a promise to wait for another route's models if you need that behavior. I'm not sure how this is "reinventing the world" and prefetch is "simply evolving it." To me, `prefetch` reinvents the world by providing an alternate universe of promise chains that is going to be hard for developers to reason about.

tomdale:
To be clear, the transition strategy for my proposal is:

1. Go through your app and rename `model()` to `fetchModel()`. In 90% of cases, that is the only change needed.
2. For routes that depend on other route's models, they refactor _just that part_ to use async model fetching. It's "Just Promises™" so it uses a programming model they're familiar with already.

`model()` eventually goes away. There is no reasoning about whether you should use `model()` or `prefetch()`. There is no manual composing of promise chains required. There's no thinking about how `model()` and `prefetch()` may or may not interact. It's a "`let` is the new `var`" strategy—it's not reinventing the world, it's making it work the way it should have worked from the beginning with a small adjustment.

One thing I'm unclear on is if my proposal works well with GraphQL, which I think is going to end up being very important. That touches on a whole bunch of stuff like when promises are resolved, how loading routes work, etc. All of those things, by the way, are punted on by the `prefetch` proposal.

So I agree with @stefanpenner that rushing things in always leads us to realize later that the grass isn't greener on the other side. That applies just as much to `prefetch` as it does to my proposal, which is why I'm explicitly trying to apply the brakes to this and push back on everyone who wants to jam something in that clearly needs more bake time. Let's expose a hook, make an addon, and let it play out from there.

stefanpenner:
> It literally is creating a new hook that is semantically identical to model() except that it gets invoked in parallel instead of serially, and provides a helper to create a promise to wait for another route's models if you need that behavior. I'm not sure how this is "reinventing the world" and prefetch is "simply evolving it." To me, prefetch reinvents the world by providing an alternate universe of promise chains that is going to be hard for developers to reason about.

I am surprised that you think this is semantically identical to model.

The current flow is as follows

#### first time we discover that this route will likely be required
* init()

#### on each entering of a route
* activate()
* beforeModel()
* model()
* afterModel()
* setupController()

### finally on exit of the route.
* deactivate();

To this prefetch adds the notion of "we will likely activate this route, you now have the opportunity to preemptively prepare yourselves" which enables pipelining.

---

In a perfect world, `modelFor` is made to return a promise, and all hooks are invoked concurrently. The promise chain the user describes is all the enforces any linearization. Pipelining happens by default, most people become happy. Unfortunately, that would be a drastic departure from the current model (as we discussed) so an alternative approach that wasn't as dramatic was requested, this is that approach. I am surprised that now you are saying, we should instead take a more dramatic approach and re-work the model.

rwjblue:
> Should we instead propose that model?

That is what @tomdale proposed I believe.

stefanpenner:
> Should we instead propose that model?
> That is what @tomdale proposed I believe.

In a meeting, we actually (when the PR was first introduced) decided it was to large a departure, and something less drastic would be required.

nickiaconis:
> > > Should we instead propose that model?
> That is what @tomdale proposed I believe.
> In a meeting, we actually (when the PR was first introduced) decided it was to large a departure, and something less drastic would be required.

I'm at a loss for what is desired considering this already exists and was (rightly, I think) met with opposition: emberjs/ember.js#12415

I feel like this may, in large part, be an issue of naming.

> To be clear, the transition strategy for my proposal is:
> 1. Go through your app and rename model() to fetchModel(). In 90% of cases, that is the only change needed.
> 2. For routes that depend on other route's models, they refactor just that part to use async model fetching. It's "Just Promises™" so it uses a programming model they're familiar with already.

This is the same transition strategy for the proposed `prefetch` hook. [This presentation](https://docs.google.com/presentation/d/17v-tKlaDcu-uTh54oiVPEbKXT4za2NMaoXjPnh2vz0s/edit?usp=docslist_api) might give a clearer picture of that, and I apologize if I've overlooked including some pieces of that picture here due to my familiarity with the implementation. Please let me know what information is missing from this RFC, so I may include it.

stefanpenner:
@nickiaconis ya unsure.. Regardless, It is likely a good idea to get the required events/hooks in, this will aide your add-on and also aide this if it can be moved forward.

nathanhammond:
One aside, from random conversation: implicit `prefetch`ing or parallelization of any kind could accidentally trigger something ​_super_​ expensive on the backend which you only want to do it when absolutely required. (A thought in favor of explicitness, ignoring API.)

stefanpenner:
@nathanhammond yes. Cancellation may also help (at some point).

Due to the apprehension towards this... I believe may the best bet is to push the public API's needed to implement this safely as an add-on.  Regardless those API's would be a big win, and if this does land in the future it would be aided by those hooks..

nickiaconis:
@stefanpenner That works for me. Do we need a separate RFC to suggest those public APIs or can we (I?) get started with an implementation?

nathanhammond:
@nickiaconis can you pull together a PR for `willChangeTransition`and other public API requirements?

/cc @mixonic

stefanpenner:
I believe superseded by https://github.com/emberjs/rfcs/pull/126

cibernox:
@stefanpenner I fail to see how #126 cover this functionality.

nathanhammond:
@cibernox #126 creates the primitive necessary to implement `prefetch` in user space (as demonstrated by where `ember-prefetch` needs to monkeypatch Ember internals).

erichonkanen:
late to the party, what is the general consensus of using prefetch in a new app (ember-prefetch)? one of our devs implemented it but then it broke the app because we have some control-flow logic that redirects around our auth wall in our `ApplicationRoute.beforeModel`, and now that logic is severed as prefetch tries to grab data but makes the http call without the auth information from `ember-simple-auth` (that package includes a mixin that adds the auth info via ApplicationRoute.beforeModel).

So I'm wondering if it's a good idea to use the prefetch package right now or wait until something comes that is part of ember core?

PR #98, opened 2015-10-08
Service methods and method injection

tim-evans:
[Rendered](https://github.com/tim-evans/rfcs/blob/master/text/0000-service-methods.md)

This RFC is an evolution of https://github.com/emberjs/rfcs/pull/71 that piggy backs on Ember Services.
This is implemented as an addon: https://github.com/tim-evans/ember-service-methods

### DIFF
--- a/text/0000-service-methods.md
+++ b/text/0000-service-methods.md
@@ -1,0 +1,173 @@
+- Start Date: 2015-10-08
+- RFC PR: (leave this empty)
+- Ember Issue: (leave this empty)
+
+# Summary
+
+Service methods provide a way to isolate business logic from your app and test it in isolation from the rest of your app.
+
+# Motivation
+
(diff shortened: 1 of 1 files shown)


tim-evans:
I think I'm going to close this due to inactivity.

DingoEatingFuzz:
As far as code management, we use modules for these types of functions. You don't get DI that way, but we have gotten away with mocking services and using Pretender for testing.

tim-evans:
@DingoEatingFuzz yeah, I was using those as well for a bit of code management as well, but it became a mess when the store was suddenly required.

I have a bunch of examples in this folder for a variety of service methods:

https://github.com/tim-evans/queertangoclub/tree/master/app/services

tim-evans:
Closing due to inactivity :)

PR #99, opened 2015-10-11
Allow query parameters to be specified when finding records via Ember Data's Store

HeroicEric:
[Rendered](https://github.com/HeroicEric/rfcs/blob/add-query-to-finders/text/0000-my-feature.md)

### DIFF
--- a/text/0000-my-feature.md
+++ b/text/0000-my-feature.md
@@ -28,10 +28,10 @@ accurate results but that is not always the case.
 # Detailed design
 
 In short, I think it would be useful to allow a `query` object to be provided to
-both `store.findRecord()` and `store.findAll()` inside the already existing
-`options` object. The `query` object would be passed to your adapter and
-serialized into the URL of the request that is made to fetch the resource from a
-server.
+both `store.findRecord()` and `store.findAll()`. The `query` would be set inside
+of `adapterOptions`, which is already available inside the adapter find methods
+as part of the `snapshot` that they receive from the store. This query will be
+serialized into the URL of the request that is made to fetch the resource.
 
 ## Inclusion of Related Resources
 
@@ -45,8 +45,10 @@ Including related resources using `store.findRecord()`:
 // GET /articles/1?include=comments
 
 const article = this.store.findRecord('article', 1, {
-  query: {
-    include: 'comments'
+  adapterOptions: {
+    query: {
+      include: 'comments'
+    }
   }
 });
 ```
@@ -57,8 +59,10 @@ Including related resources using `store.findAll()`:
 // GET /articles?include=comments
 
 const articles = this.store.findAll('article', {
-  query: {
-    include: 'comments'
+  adapterOptions: {
+    query: {
+      include: 'comments'
+    }
   }
 });
 ```
@@ -77,11 +81,13 @@ the name of its author.
 // GET /articles/1?include=author&fields[articles]=title,body&fields[users]=name
 
 const article = this.store.findRecord('article', 1, {
-  query: {
-    include: 'author',
-    fields: {
-      articles: 'title,body',
-      users: 'name'
+  adapterOptions: {
+    query: {
+      include: 'author',
+      fields: {
+        articles: 'title,body',
+        users: 'name'
+      }
     }
   }
 });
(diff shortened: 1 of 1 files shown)


text/0000-my-feature.md
```diff
@@ -0,0 +1,178 @@
...
+might need for features of their own API.
+
+## Implementation
+
+Since `store.findRecord()` and `store.findAll` already accept an `options`
+object, their public APIs will not need to be changed. The private finders that
+they use, however, will need to be modified to pass the `query` (or all of
+`options`) through to their corresponding adapter methods. An example of one of
+the places that would require change is
+[here](https://github.com/emberjs/data/blob/8f83fbf27acdd642cb824e3286832acebdac0da1/packages/ember-data/lib/system/store/finders.js#L19).
```
bmac: before `adapterOptions` were attached to the `snapshot` we considered passing `adapterOptions` into `findRecord` as an extra argument. In the end we decided to attache it to the snapshot in order to make it work with colaced requests. 

I think we should follow the same pattern here and copy `query` into `adapterOptions.query`. This means 

```js
const articles = this.store.findAll('article', {
  query: {
    include: 'comments'
  }
});
```

would be equivalent to 

```js
const articles = this.store.findAll('article', {
  adapterOptions: {
    query: {
      include: 'comments'
    }
  }
});
```

text/0000-my-feature.md
```diff
@@ -0,0 +1,178 @@
...
+# Unresolved questions
+
+## Knowing When to Fetch
+
+It's not totally obvious to me what impact this should have (if any) in terms of
+when to fetch data from the server. We could always fetch data from the server
+if any query parameters are passed but I think the better default is to rely on
+loading the data in the background to make sure that we eventually end up with
+all of the data that we need. If a user wants to make sure that they always have
+fresh data, they can do so by setting the `reload` option to `true`.
```
bmac: cc @wycats. I believe you had some thoughts on how this case should be handled.

text/0000-my-feature.md
HeroicEric: Nice! I wasn't aware that those were already being passed via the snapshot. I'll update the RFC.

HeroicEric: @bmac I just updated this to nest `query` under `adapterOptions` in the examples but I think I just realized that you were suggesting to keep `query` as the as option to `findRecord` but inside `findRecord` copy the query over to `adapterOptions`. 

So the api would still be something like:

```javascript
this.store.findRecord('post', 1, {
  query: {
    include: 'author'
  }
});
```

but that `query` would be copied over into `adapterOptions` before the snapshot is created? Is that what you meant?

mwpastore:
It would also be useful to be able to supply `adapterOptions` when fetching associated records, e.g.:

```javascript
this.store.peekRecord('foo', 1).get('bars', {
  adapterOptions: {
    include: 'quuxes'
  }
});
```

bmac:
Hi @HeroicEric I'm a little weary about the potential for the query parameter to load a record in an incomplete state. My fear is that a record may be loaded into the store with fields missing when visiting one route then retrieved from the cache when the user navigates to a new route. 

I would like to start with a more conservative API that only exposes the ability to specify the `include` parameter when making a find request.

```js
const article = this.store.findRecord('article', 1, { include: 'comments' });
const articles = this.store.findAll('article', { include: 'comments' });
article.save({ include: 'comments' });
article.destroyRecord({ include: 'comments' });
```

This still has the potential for the same problem of expected relationship data not being loaded when loading a record from the cache, however, Ember Data has better tools (namely async relationships) for dealing with this case then it has for partially loaded fields.

mwpastore:
That sounds great to me, @bmac. Thoughts on extending that functionality to relationship fetching? Have you seen the ember-data-has-many-query add-on?

luxferresum:
I would love a official ember way to query data with a cache.
At the moment I've written my own hack [here](https://gitlab.com/ddag/ember-query-cache).

It hacks into the store and asks `shouldReloadQuery` on the adapter if the query result is cachable.

Something to notice: Think about how to handle with pagination and infinity-scroll.
If you query the first page, then load the second, you want to access both if you access the cache.

HeroicEric:
This was added in https://github.com/emberjs/data/pull/3976

PR #101, opened 2015-10-23
Friendly Errors

nikz:
Hopefully I filled out everything correctly.

Thanks for looking at this! :sunglasses: 

Here's a wee "friendly error" preview:

### DIFF
--- a/text/0000-ember-data-friendly-errors.md
+++ b/text/0000-ember-data-friendly-errors.md
@@ -1,0 +1,90 @@
+- Start Date: 2015-10-23
+- RFC PR: (leave this empty)
+- Ember Issue: (leave this empty)
+
+# Summary
+
+Add more illustrative detail to the default Ember Data Adapter Errors.
+
+# Motivation
+
+With a production Ember project, it's common to have many errors of the form "Adapter Error",
+originating from deep in the Ember Data stack and carrying little context about what the
+original error cause was.
+
+The intent is to add the original request URL, the response code, and some payload information
+to the default Error message for `DS.AdapterError`s. From there Errors can be handled or
+tracked as normal.
+
+# Detailed design
+
+I've been using something similar to the following Adapter (`friendly-error-adapter.js`):
+
+```js
+import ActiveModelAdapter from 'active-model-adapter';
+
+import DS from 'ember-data';
+
+export default ActiveModelAdapter.extend({
+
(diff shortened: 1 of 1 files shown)


alexspeller:
:+1:

tim-evans:
It would be neat to provide something like Rail's `filter_parameters` if requests are shown.

jcope2013:
a start https://github.com/emberjs/data/pull/3862

bmac:
Hi @nikz. I like this rfc a lot. Would you like to open a pr on the Ember Data repo where we can iterate on the implementation of the `friendly-error-adapter` code?

text/0000-ember-data-friendly-errors.md
```diff
@@ -0,0 +1,90 @@
...
+
+```js
+import ActiveModelAdapter from 'active-model-adapter';
+
+import DS from 'ember-data';
+
+export default ActiveModelAdapter.extend({
+
+  ajax(url, method)  {
+    this.lastRequest = {
```
bmac: This technique may not work with multiple concurrent ajax requests. Instead I suspect we will need to pass the request or errorMessage as an argument to `handleResponse`.

nikz:
@bmac awesome! I will do so over the weekend and cc you :)

text/0000-ember-data-friendly-errors.md
nikz: Yeah, mostly a hack to avoid trying to drop too deeply into the `ember-data` code. Definitely better ways to do it!

fivetanley:
Assigned an RFC number and merged in 61576858687fc61771f5200854b8c41009eedb51

@nikz's PR in emberjs/data#3930 was the implementation. Thank you so much!

grapho:
I really like the idea of this RFC.  However I also hold out hope that an Adapter error could hold an actual data payload.. for those of us that might receive an array of errors or a hash of error parameters? rather then just a serialized string of info?

nikz:
@grapho Hiya! Remember this is just the default handler if you haven't specified anything else, mostly for debugging and error tracking.

If you use a blob of JSON then that'll be spit out, or if you're looking for more model validation things then a JSON API "errors" payload would work best.

grapho:
@nikz  yeah oops, i momentarily forgot that error payload can already get passed in... just ignore me ;)

PR #110, opened 2016-01-17
promise-aware templates

fivetanley:
Rendered: https://github.com/fivetanley/rfcs/blob/promise-aware-templates/text/0000-promise-aware-templates.md

### DIFF
--- a/text/0000-promise-aware-templates.md
+++ b/text/0000-promise-aware-templates.md
@@ -1,0 +1,74 @@
+- Start Date: 2016-01-16
+- RFC PR: (leave this empty)
+- Ember Issue: (leave this empty)
+
+# Summary
+
+A Promise-aware rendering engine allows users to declaratively write synchronous looking code in their templates and have the template re-render when the promise is resolved.
+
+# Motivation
+
+A common point of frustration when using Ember Data relationships is the fact that they don't often re-render as expected.
+
+A common example is the following. A `Book` belongs to an `Author`, but the `Author` may be null. This relationship is async:
+
+```javascript
+// app/models/book.js
+import DS from 'ember-data';
+
+export default DS.Model.extend({
+  name: DS.attr(),
+  author: DS.belongsTo('author'. {async: true})
+});
+```
+
+Assuming that `book` is a `Book` model in the following code:
+
+```handlebars
+{{#if book.author}}
+  <span>Yay, there is an author: {{book.author.name}}!</span>
+{{else}}
+  <strong>There's no author! Please input an author name and we'll figure it out.</strong>
+  {{input value=form.author}}
+{{/if}}
+```
+
+Even if the `book.author` promise resolves with `null`,  the rendered output will state that there is an author every time, which is innacurate. In addition, there's no way to know if the promise is resolving. This is also accurate for any promise-like object (jQuery, RSVP, native Promise, etc).
+
+In the past, we have used `PromiseProxy`, which wraps a promise and proxies property look-ups to its `content`, which is a promise. This doesn't work because (fill this in when you can remember), and the implementation in Ember itself has known to be error-prone. This would also require the user to wrap all their promise objects in PromiseProxies, which feels unintuitive and repetitive.
+
+# Detailed design
+
+When Glimmer encounters a promise-like object, it should call `then`, and re-render the node when the promise is resolved or rejected. 
+
+
+# Drawbacks
+
+* Uses duck-typing to determine if something is a promise by calling `then`
+* Could potentially be slow
+* Potentially backwards incompatible.
+* No way to inspect promise state unless helpers are provided (see below).
+
+# Alternatives
+
+An alternative without changing glimmer itself is to provide a set of helpers that allow the state of a promise to be handled by the user.
+
+```handlebars
+{{#await model.author as |author, error, isPending|
+  {{#if isPending}}
+    Loading...
+  {{else if error}}
+	Error! {{error}}
+  {{else}}
+	{{#if author}}
+	  Yay! The author is: {{author.name}}
+	{{else}}
+	  Boo! There is no author!
+	{{/if}}
+  {{/if}}
+{{/await}}
+```
+
+The syntax is a bit verbose, and doesn't feel as intuitive.
+
+# Unresolved questions


fivetanley:
Hoping to cycle in some core team members during the face to face this weekend, so i apologize for holes in the RFC. This is what I can remember from our Ember Data discussions on the problem.

workmanw:
I agree the `{{await}}` option is overly verbose, but one thing that is appealing is an explicit `isPending` option. Our UX is frequently to show a pending state first, then show result state. Currently we use `isFulfilled` eg (`{{#if book.author.isFulfilled}}`). For that reason, I do favor a helper based solution that makes this easier and more explicit. 

Also, this scenerio is related: http://discuss.emberjs.com/t/unless-helper-dilemma/9361

tchak:
I also find it unclear what's happening if the promise is rejected? There is no way to handle it directly in the template?

stefanpenner:
@fivetanley gave this a read, i have some ideas but will save them for in-person. We can then write changes/ideas down after.

tim-evans:
This has definitely been a pain point for me in the past, especially when interacting with helpers. Passing things that may / may not be promises to a helper may be a good exercise to reveal what the optimal solution for this is. (In the end, I just want promises to work transparently in templates.)

For an example with helpers, I've written one that formats a location:

```hbs
import Ember from 'ember';

const { get } = Ember;

export default Ember.Helper.extend({
  compute(location) {
    if (location) {
      this.location = null;
      Ember.RSVP.cast(location).then((loc) => {
        this.location = loc;
        this.recompute();
      });
    }

    if (this.location) {
      let location = this.location;
      return get(location, 'addressLine') + get(location, 'extendedAddress') + '<br>' +
                `${get(location, 'city')}, ${get(location, 'regionCode')}`;
    }
  }
});
```

stefanpenner:
I *believe* a lot of this could be implemented as a helper/subexpression today (not to say that would superseed an RFC) but a reference implementation to play with would allow us to explore this RFC in more detail.

fivetanley:
@stefanpenner I started working on it as an addon last night, but forgot to update. For those interested in following along: https://github.com/thefrontside/ember-promise-helpers

Only the README is written right now.

tim-evans:
Psst... Here's `await` implemented as a `Ember.Helper`:

```handlebars
/**
  `await` unwraps promises for use in templates. Traditionally,
  passing promises to helpers or trying to access properties on
  a promise requires unwrapping promises in a controller or
  component. Rejected promises return `null`; resolved promises
  return the value.

  Example:

  ```handlebars
    {{#with (await currentUser) as |user|}}
      {{user.name}}
    {{else}}
      No User
    {{/with}}
  ```

  @public
  @method await
 */
export default Ember.Helper.extend({
  compute([promise]) {
    if (promise !== this.promise) {
      this.promise = promise;
      this.value = null;
      Ember.RSVP.cast(promise).then((value) => {
        this.value = value;
        this.recompute();
      }, function () {});
    }

    return this.value;
  }
});
```

stefanpenner:
`Ember.RSVP.Promise.resolve(promise).then((value) => {`

pixelhandler:
@fivetanley I like the idea of using a promise proxy object/array with the promise proxy mixin so that properties like you mentioned `isFullfilled` can be used.

I'm not sure how the template helpers e.g. `each` would become promise aware. Are they not currently? The guide page for rendering a list is not clear whether or not the value passed to the `each` helper is a promise or not. see: https://guides.emberjs.com/v2.3.0/templates/displaying-a-list-of-items/

```hbs
<ul>
  {{#each people as |person|}}
    <li>Hello, {{person.name}}!</li>
  {{/each}}
</ul>
```

So if `people` is a property that depends on a request to the `store` (that is a promise) would the list items never render?

In your original issue, you mentioned that developers are led to solving this creatively, e.g. own book-keeping. My solution was in a data persistence library rather then in my application code. For the sake of the sharing use cases on how people address this, below is an example…

Here is an example of how I use a computed property for a relation that resolves to the data returned from an API request. Below `model` is the `post.comments` computed property. It's a promise proxy using an array, so it starts as a promise and when resolved the content is set to the array of data. the change to the computed property for `post.comments` does re-render the template.

```hbs
{{#each model key="@identity" as |comment|}}
  <li>
    {{#with comment.commenter as |commenter|}}
      <span>by <strong>{{commenter.name}}</strong>…</span>
    {{/with}}
    on <span>{{comment.date}}</span>
    <p>{{comment.body}}</p>
  </li>
{{/each}}
```

- Post has-many "comments"
- Comment has-one "commenter"
- See the models for post, comment, commenter here: https://github.com/pixelhandler/jr-test/tree/master/app/models

These "relationships" use a related proxy object, `ArrayProxy` (has-many) or `ObjectProxy` (has-one) combined with the `PromiseProxyMixin`, see: https://github.com/pixelhandler/ember-jsonapi-resources/blob/master/addon/utils/related-proxy.js#L90-L92

So, the computed property of the model for has-one and has-many relations use a proxy for the data.

When the comments template (above, used in an outlet of the post template) has a resolved `model` the array of comments are rendered, likewise when the comments' related commenter is resolved the `comment.commenter` renders.

My template's are not "promise aware" instead the model relations are proxy objects, but the template does re-render when the promises resolve.

elucid:
@fivetanley In my experience this started becoming an issue after ember-data 1.0.0.beta.11 was released. Prior to this the promise proxies wrapping related records would have `content: null` until the promise resolved. However, https://github.com/emberjs/data/commit/12301b9405c4ffde31d9dd027c8b266c1231d0f8 introduced different behavior whereby `content` was pre-set to a record that would have its attributes filled in once the promise resolved.

Consequently, certain patterns that previously worked within templates broke. When the promise proxies previously had `content: null`, the bindings would fire on promise resolution when `content` was set to the returned record.

I had started work on a pull-request to fix this (https://github.com/emberjs/data/pull/3812). Initially I misunderstood what the pre-set content change implied and hadn't run the tests properly. Once I got the tests to run correctly and started looking into the failures it became clear that reverting to having `content: null` promise proxies before resolution would not be trivial.

/cc @ghedamat

fivetanley:
@elucid `content` is marked as private API but frequently used anyway. I'm not sure restoring the private behavior is what we would like to do long term.

elucid:
I don't mean using `content` directly. Rather, let's say I have `Post` belongsTo `Author`, and I do

```
  {{post.author.someComputedProperty}}
```

in a template. With the old behavior, `someComputedProperty` would not be evaluated before the promise resolves because when the proxy falls through to look up `someComputedProperty` on `content` and `content` is `null`, the lookup does not proceed.

However, if content is pre-set to an incomplete record, the lookup of `someComputedProperty` will actually result in the property being computed before the data has been populated in the record.

fivetanley:
I've added some failing tests here if anyone wants to give feedback. https://github.com/fivetanley/ember-promise-helpers/pull/1

fivetanley:
I also don't think this belongs in ember core (at least not until it gets battle tested in userland for a bit), so I'm closing this for now.

elucid:
@fivetanley is it worth having a discussion about revisiting the way that promise proxies are used in ember data? That is, reverting to having content null until the promise resolves?

This doesn't address the general issue but in my experience comprises the majority of instances where I see this.

fivetanley:
That doesn't feel like an RFC to me, but would be welcome on the Ember Data issue tracker.

elucid:
Okay thanks, will do.

fivetanley:
I have a 1.0 for those following along at https://github.com/fivetanley/ember-promise-helpers. Please kick the tires and open issues there, thanks!

PR #112, opened 2016-01-24
Element Modifiers

mixonic:
[Rendered](https://github.com/mixonic/rfcs/blob/element-modifier/text/0111-element-modifier.md)

### DIFF
--- a/text/0111-element-modifier.md
+++ b/text/0111-element-modifier.md
@@ -1,0 +1,226 @@
+- Start Date: 2016-01-23
+- RFC PR: https://github.com/emberjs/rfcs/pull/111
+- Ember Issue:
+
+# Summary
+
+As an outcome of [RFC 100](https://github.com/emberjs/rfcs/pull/100#issuecomment-172427565)
+it was decided that element-space helpers should be added to Ember as a public
+API.
+
+In this example, `add-event-listener` fits the syntax of an element modifier:
+
+```hbs
+<span {{add-event-listener 'click' (action 'save')}}>Save</span>
+```
+
+The implementation of `add-event-listener` would have several hooks called during
+rendering, similar to the rendering hooks fired on a component. Unlike a
+component, there is no template/layout for an element modifier. Unlike a
+helper, an element modifier does not return a value.
+
+# Motivation
+
+Element modifiers allow for user-space implementation of event listeners,
+style, and animation tooling that does not exist in Ember today. It brings
+back some functionality removed from the framework in Ember 2.0 (when
+various intimate helper APIs were removed).
+
+The introduction of this API will likely result in the proliferation of
+one or several popular addons for managing element event listeners, style
+and animation.
+
+# Detailed design
+
+### Invocation
+
+An element modifier is invoked in "element space". This is the space between
+`<` and `>` opening an HTML tag. For example:
+
+```hbs
+<button {{flummux}}></button>
+<span {{whipperwill 'carrot'}}><i>Some DOM</i></span>
+<b {{crum bing='whoop'}} zip="bango">Hm...</b>
+```
+
+Element modifiers may be invoked with params or hash arguments.
+
+### Definition and lookup
+
+A basic element modifier is defined with the type of `element-modifier`. For
+example these paths would be global element modifiers in an application:
+
+```
+Classic paths:
+
+  app/element-modifiers/flummux.js
+  app/element-modifiers/whipperwill.js
+
+Pods paths:
+
+  app/flummux/element-modifier.js
+  app/whipperwill/element-modifier.js
+```
+
+Element modifiers, like component and helpers, are eligible for local lookup.
+For example:
+
+```
+Pods paths:
+
+  app/posts/index/element-modifiers/flummux.js
+```
+
+The element modifier class is a default export from these files. For example:
+
+```js
+import Ember from 'ember';
+
+export default Ember.ElementModifier.extend({});
+```
+
+### Hooks
+
+During rendering and teardown of a target element, any attached element
+modifiers will execute a series of hooks. These hooks are:
+
+* `willInsertElement` (only upon initial render)
+* `didUpdateAttrs` (only upon subsequent render)
+* `willRender`
+* `didRender`
+* `didInsertElement` (only upon initial render)
+* `willDestroyElement` (only upon teardown)
+
+These lifecycle hooks are similar to those of the component lifecycle. Similarly
+the properties of `this.element` and `this.attrs` may be present during the
+execution of a hook.
+
+* during `willInsertElement`, initial `this.attrs` will be present, but no `this.element`
+* during `didUpdateAttrs` both `this.attrs` and `this.element` are present. Additionally
+  the hook is pass `oldAttrs, newAttrs` as arguments.
+* during `willRender` both `this.attrs` and `this.element` are present
+* during `didRender` both `this.attrs` and `this.element` are present
+* during `didInsertElement` both `this.attrs` and `this.element` are present
+* during `willDestroyElement` both `this.attrs` and `this.element` are present
+
+An example of a simple element modifier definition:
+
+```hbs
+<button {{noodle logEvent='mouseover'}}></button>
+```
+
+```js
+// app/element-modifiers/noodle.js
+import Ember from 'ember';
+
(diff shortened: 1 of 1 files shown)


rwjblue:
This looks good, thanks for taking the time to put it together.

I have a few thoughts/questions:

* From the examples, I assume that `this.attrs` is roughly the same as in `Ember.Component`'s today. In "curly components" the mutable object confusion in `this.attrs.` is very annoying. Why introduce yet another place where `this.attrs` bizarro land still exists?  There also does not seem to be a path forward away from these as two way bound objects, since you wouldn't be able to lean on angle bracket syntax inside of an element.  I would prefer to see either no `this.attrs` or have `this.attrs` not be two-way bound.
* Can you detail which hooks fire in server side rendering (aka Fastboot)?
* We have generally regretted passing `attrs`, `newAttrs`, `oldAttrs`, etc to the various hooks. Should we avoid that mistake here?
* Will element modifiers only be looked for while inside an element without assignment? Given `<button {{foo}} data-bar={{foo}}></button>`, `element-modifier:foo` would be used for the invocation inside the element but `helper:foo` would be used for the `data-bar` usage.  Is this correct?
* Similar to above, will these two usages use different `foo` objects: `<button {{foo}}>{{foo}}</button>`. One for the element space invocation (`element-modifier:foo`) and another for the usage in "normal" space (`helper:foo`)?

Again, great job on this, and thank you for pushing this forward!

rwjblue:
We should also detail how this interacts with angle bracket components.

* Will `<my-button {{add-event-listener 'click' (action 'click')}}></my-button>` "work" in some way?
* Would element space modifiers be usable from both the invoking side and the components own layout?

chadhietala:
I'm wondering if there is value in giving my well defined and dependent hooks. My worry is around the higher potential of memory leaks being introduced into a project. If you had hooks like `setupListeners` and `teardownListeners` it is clear what you should implement there. Furthermore from a framework point of view, a development version of Ember could do something like:

```js
let hasListener = !!component.setupListeners;
let hasRemoveListener = !!component.teardownListeners;

assert(`You setup an event listener on ${component.toString()}, but did not tear it down in `teardownListeners`. If you are not setting up a listener please use another hook.`, component.setupListeners && !hasRemoveListener)
```

mmun:
Totally an aside, but I hope if we ever have a helper like `add-event-listener` we call it `on` instead.

cibernox:
Mixing existing ideas (the global dispatcher) with this approach, this would solve the problem that `onclick={{action 'foo'}}` makes extremely complicated to use event delegation.

Something like 
```
<ul>
  {{#each option as |opt|}}
    <li {{dispatch-event 'mousedown' (action "choose" opt)}}></li>
  {{/each}}
</ul>
```
could use Ember's global dispatcher and avoid adding/removing loads of event handlers if the collection changes?

tchak:
In pre 2.0 lots of confusion was coming from the fact that Ember had way too much "primitives". I was hoping we were on the right path with components only in 2.0. But we introduced state full helpers and now this. This is not a complaint out of nowhere - I already found myself in situations where I had to explain to some people how a state full helper is different from a component... Feels like explaining how a view is different form a component in 1.0... I understand that there is a problem to be solved here, but I feel that this will extend api surface in the area where it is already pretty large.

cibernox:
To some extent I feel the same way than @tchak about this, and I wonder if there is any way of unify helpers with this. 
Some sort of helpers with access to the DOM of the element they are invoked in. Perhaps just a stateful helper that happens to have lifecycle hooks other than `compute` and those hooks are invoked with the DOM element as the first argument.

I don't know if can be considered having less primitives, because it's basically adding more capabilities to one existing primitive.

jgwhite:
Thanks @mixonic, this is excellent.

From an addon-author’s point of view, I would love to move the API for ember-sortable towards something like:

```hbs
{{! my-cool-template.hbs}}
<ul>
  {{#sortable-group as |group|}}
    {{#each items as |item|}}
      <li {{group.sortable-item model=item}}>{{item.name}}</li>
    {{/each}}
  {{/sortable-group}}
</ul>
```

```hbs
{{! sortable-group.hbs}}
{{yield (hash
  sortable-item=(element-modifier "sortable-item" group=this)
)}}
```

Note the assumed availability of yielding contextual-element-modifiers. Did you see that as being a possibility?

text/0111-element-modifier.md
```diff
@@ -0,0 +1,226 @@
...
+
+export default Ember.ElementModifier.extend({});
+```
+
+### Hooks
+
+During rendering and teardown of a target element, any attached element
+modifiers will execute a series of hooks. These hooks are:
+
+* `willInsertElement` (only upon initial render)
```
krisselden: what is the use case for this with an element modifier?  I was thinking these would be scope to didInsertElement and willDestroyElement

text/0111-element-modifier.md
```diff
@@ -0,0 +1,226 @@
...
+
+export default Ember.ElementModifier.extend({
+
+  init() {
+    this._super(...arguments);
+    this._logMouseover = () => console.log('mouseover!');
+  },
+
+  didRender() {
+    document.addEventListener(this.attrs.mouseover, this._logMouseover);
```
krisselden: shouldn't this be this.attrs.logEvent?

stefanpenner:
Am I correct in assuming these are isomorphic to decorators, maybe a class decorator if one assumes each element is a class (for this discussion)

krisselden:
Should this have a `this.component` in the case it is an Ember component?

mixonic:
Thanks for the early feedback all. I'm still digesting but will take a pass at improvements soon.

From @rwjblue:

> From the examples, I assume that this.attrs is roughly the same as in Ember.Component's today. In "curly components" the mutable object confusion in this.attrs. is very annoying. Why introduce yet another place where this.attrs bizarro land still exists? There also does not seem to be a path forward away from these as two way bound objects, since you wouldn't be able to lean on angle bracket syntax inside of an element. I would prefer to see either no this.attrs or have this.attrs not be two-way bound.

To me, this is the most concerning comment. The proposal as it stands leans heavily on the API of components, with the goal not introducing a new set of hook names when they already largely match components. Using `attrs` was intended to keep with that goal, but @rwjblue is right that `attrs` is not mature enough to consider a starting point. Unfortunately that leaves me going back to the drawing for for wholly new hooks, at least one of which would need the `params, hash` passed like it is for helpers.

taras:
@mixonic @rwjblue do you have any thoughts on what the story with this RCF is going to be given work that's happening with Glimmer 2 and everything that's going on?

mixonic:
@taras I'm still very much in favor of completing an RFC for this feature, though the implementation is indeed very likely blocked on Glimmer 2.

The API design is somewhat intertwined with that of Glimmer components, since I would like the way attrs are passed to match the Glimmer component API. For example this seems plausible:

```hbs
<div
  class="red button"
  {{on-event 'click' (action 'save' model) @bubble=false}}
>Save</div>
```

Roughly:

* `@ember-prop` passes to `this.args['ember-prop']`
* element modifier gets a `this.element`
* if there is a glimmer component, element modifier gets `this.component`
* lifecycle hooks match those of Glimmer components

Again some of this is directly related to what APIs we design for Glimmer components, so I'll need to coordinate w/ @wycats and others to get consensus. I'm eager to take another pass :-)

text/0111-element-modifier.md
knownasilya: I submitted a PR against the rfc too fix this.

knownasilya:
Looking forward to some of this picking up now that Glimmer2 has basically landed.

cibernox:
This is also something I find interesting and it's a shame it has been staled for a year. Was it abandoned for specific reasons?

knownasilya:
Probably the priorities have been diverted, but this seems like an important topic, especially for glimmer.js and it's custom elements story. It should work it's way upstream I think.

mixonic:
I think progress on this is again tractable. The biggest blocker is the API for how to access positional and named arguments inside the modifier. The hooks should be normalized to match those on Glimmer components.

I'm wary of trying to push it forward before closing up the namespaces/addon support for module unification, re: my own time management 😬

mixonic:
Long belated: Closing this in favor of https://github.com/emberjs/rfcs/pull/353

PR #113, opened 2016-01-27
Expose query param defaults

kellyselden:
[Rendered](https://github.com/kellyselden/rfcs/blob/expose-query-param-defaults/text/0000-expose-query-param-defaults.md)

### DIFF
--- a/text/0000-expose-query-param-defaults.md
+++ b/text/0000-expose-query-param-defaults.md
@@ -1,0 +1,62 @@
+- Start Date: 2016-01-27
+- RFC PR: (leave this empty)
+- Ember Issue: (leave this empty)
+
+# Summary
+
+Expose the default starting value of queryParams so they can be reset in an automatic way.
+
+# Motivation
+
+If you want to reset query params for a route now, you have to copy all the default values to reset them in the route's `resetController`, or you have to use internal route data to find the default values. If we expose the defaults, a general case mixin can be created that will reset any arbitrary route's query params.
+
+# Detailed design
+
+Here is the current recommended way to reset query params:
+
+```js
+resetController(controller, isExiting) {
+  if (isExiting) {
+    controller.set('firstParam', 'myDefaultValue');
+    controller.set('secondParam', 'myOtherDefaultValue');
+  }
+}
+```
+
+This can't be applied in a general way because it has to know specifics about each route.
+
+Here is a sample from my addon [ember-query-params-reset](https://github.com/kellyselden/ember-query-params-reset/blob/master/addon/mixins/query-params-reset-route.js). This sniffs for the appropriate internals for your ember version:
+
+```js
+resetController(controller, isExiting) {
+  if (isExiting) {
+    this.get('_qp.qps').forEach(qp => {
+      let defaultValue;
+
+      if (qp.hasOwnProperty('def')) {
+        // < v2.0
+        defaultValue = qp.def;
+      } else {
+        // >= v2.0
+        defaultValue = qp.defaultValue;
+      }
+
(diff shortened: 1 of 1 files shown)


workmanw:
We have this exact use case all over the place. However, it's not just limited to `resetController`, we often use `this.transitionTo` or `{{link-to}}` with query-params throughout the app. When query-params are sticky we have cases where we need to reset back to the default value as part of the transition. 

So I suggest that we consider expanding the scope of the RFC to provide a mechanism for other routes / components to determine the default values.

We ended up making constants for these that could be exported from a single global file. Eg:

```
// app/constants.js
export const ROOM_QP_SORT = 'newest';
export const ROOM_QP_FILTER = 'all';
```

```
// app/controllers/room.js
import { 
  ROOM_QP_SORT, 
  ROOM_QP_FILTER, 
  ROOM_QP_SEARCH
} from 'app/constants';

export default Ember.Controller.extend({
  queryParams: ['sort', 'filter', 'search'],

  sort: ROOM_QP_SORT,
  filter: ROOM_QP_FILTER,
  search: ROOM_QP_SEARCH
});
```

```
// app/routes/other-route.js
  actions: {
   deleteThingAndGoToList() {
     this.currentModel.destroyRecord().then(() => {
       this.transitionTo('room', { 
         queryParams: {
           sort: ROOM_QP_SORT,
           filter: ROOM_QP_FILTER,
           search: ROOM_QP_SEARCH
         }
       });
     });
    }
  }
});
```

```
{{!-- x-routable/other-route.hbs }}
{{#link-to 'room' (query-params 
    sort=(c "ROOM_QP_SORT") 
    filter=(c "ROOM_QP_FILTER") 
    search=(c "ROOM_QP_SEARCH"))}}
  Show All
{{/link-to}}
```

trentmwillis:
:+1: I'm onboard with this since I think, if designed properly, the API would be beneficial to asynchronous routing. Since [default QP values are serialized in URLs](https://guides.emberjs.com/v2.3.0/routing/query-params/#toc_default-values-and-deserialization), having some way of exposing them (ideally, separated from the controller) could enable pre-emptive loading ahead of the actual controller.

cc @nathanhammond

kellyselden:
> Since default QP values are serialized in URLs

@trentmwillis You mean "aren't" right?

From your link: "When a controller's query param property is currently set to its default value, this value won't be serialized into the URL"

trentmwillis:
@kellyselden yeah, what I really meant was more along the lines of "are taken into consideration when serializing URLs".

nathanhammond:
So today is the day where I finally learn query params. I've made it four years without doing anything more than trivial usage. The goal @trentmwillis mentions here is basically regarding this: https://github.com/dgeb/ember-engines/issues/6#issuecomment-180823082

Right now if you don't expose query param defaults for route targets in a global manner when constructing links in engines land you end up having the defaults appear in URLs (if they're passed). Suboptimal, but not the end of the world.

Because of this RFC and related issues it might make sense to reconsider how we do that, solve the engines problem, and solve this ergonomics problem all at once.

wagenet:
I believe there is some router-related spiking happening around Polaris. That may eliminate the need for this RFC.

wagenet:
Thanks for spending the time writing this RFC! Given the significant number of concerns around routing, we’re going to be completely revamping the router as part of Polaris. This RFC won’t directly apply to the new router so we’re going to move this PR to FCP to Close. But don’t worry, your work here was not in vain! We will be keeping track of this RFC and others like it to make sure that the concerns it aims to address will be covered in the new router. Your work here does a significant service to us in this regard!

PR #115, opened 2016-01-29
Add Ember.run.callback()

davewasmer:
[Rendered](https://github.com/davewasmer/ember-rfcs/blob/run-callback/text/0000-run-callback.md)

### DIFF
--- a/text/0000-run-callback.md
+++ b/text/0000-run-callback.md
@@ -37,10 +37,36 @@ The implementation has been spiked out in an addon: [ember-run-callback](https:/
 
 # Alternatives
 
-* Users could continue to use existing solutions, typically custom test helpers which either:
-  * register waiters from the test side which check for some result of the async action, or
-  * reach into the app's running state via private APIs like `__container__` to grab a direct reference to the actual async activity (i.e. a promise)
-* This could stay a userland addon rather than be adopted into the core framework
+This could always stay a userland addon rather than be adopted into the core framework.
+
+## Current solutions
+
+There are two main approaches to dealing with this problem currently:
+
+### Manual waiters
+
+```js
+click('.start-something-async');
+Ember.Test.registerWaiter(function myCustomWaiter() {
+  return find('.some-async-result').text().trim() === 'All done!';
+});
+andThen(function() {
+  Ember.Test.unregisterWaiter(myCustomWaiter);
+  expect(find('.some-async-result').text().trim()).to.equal('All done!');
+});
+```
+
+### Direct referencing
+
+```js
+click('.start-something-async');
+andThen(function() {
+  let promise = app.__container__.lookupFactory('controller:foo').get('promise');
+  promise.then(function() {
+    expect(find('.some-async-result').text().trim()).to.equal('All done!');
+  });
+});
+```
 
 # Unresolved questions
 
--- a/text/0000-run-callback.md
+++ b/text/0000-run-callback.md
@@ -1,0 +1,48 @@
+- Start Date: 2016-1-28
+- RFC PR:
+- Ember Issue:
+
+# Summary
+
+Add `Ember.run.callback()` to support running asynchronous code which tests (specifically `andThen()`) will wait for before considering the app "settled".
+
+# Motivation
+
+Ember's asynchronous test helpers (i.e. `visit()`, `fillIn()`, `andThen()`) provide a robust approach to acceptance testing in the face of asynchronous user interactions and application code. The `andThen()` helper is the backbone of this approach - it waits until the app has "settled" (i.e. async activity has competed) before asserting the desired state.
+
+For many, if not most, ambitious web applications, integration with third party libraries that require asynchronous interaction is a requirement. But unfortunately, there is currently no elegant solution for informing Ember of this asynchronous activity without resorting to lower level APIs like `registerWaiter()`.
+
+For such a common problem, Ember's opinionated approach needs an opinionated answer that matches the elegance of the rest of the test API.
+
+# Detailed design
+
+This RFC adds a single method to the `Ember.run` namespace: `Ember.run.callback()` (better naming suggestions are welcome!).
+
+The method takes a single function as an argument, and returns a wrapper function that can be used in place of the originally supplied one. This wrapper does two things:
+
+1. Ensures the callback is wrapped in an `Ember.run()` call, mostly for convenience.
+2. More importantly, it wraps the user-supplied function with a `registerWaiter` method (and the associated state management / cleanup) which resolves once the
+callback is invoked.
+
+By having the application register the waiters itself, we can avoid compromising the "purity" of the acceptance tests which would otherwise need to reach into the application's running state.
+
+The return value of of `Ember.run.callback()` would be the wrapped function, but the waiter would be registered immediately. If the callback is no longer needed, the user could cancel the waiter by passing the callback to `Ember.run.cancel()` (matching the other async `Ember.run.*` methods).
+
+The implementation has been spiked out in an addon: [ember-run-callback](https://github.com/davewasmer/ember-run-callback).
+
+# Drawbacks
+
+* This would increase the `Ember.run` API surface and complexity, which (anecdotally) seems to be one of the less understood APIs in Ember.
+* There could be some potentially tricky edge cases if a user creates a test-aware callback via this method, but then fails to invoke it. Because invoking `Ember.run.callback()` _immediately_ starts to block the app from settling, if the callback is never invoked, it would hang the tests.
+
+# Alternatives
+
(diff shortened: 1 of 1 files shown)


text/0000-run-callback.md
```diff
@@ -0,0 +1,48 @@
...
+For such a common problem, Ember's opinionated approach needs an opinionated answer that matches the elegance of the rest of the test API.
+
+# Detailed design
+
+This RFC adds a single method to the `Ember.run` namespace: `Ember.run.callback()` (better naming suggestions are welcome!).
+
+The method takes a single function as an argument, and returns a wrapper function that can be used in place of the originally supplied one. This wrapper does two things:
+
+1. Ensures the callback is wrapped in an `Ember.run()` call, mostly for convenience.
+2. More importantly, it wraps the user-supplied function with a `registerWaiter` method (and the associated state management / cleanup) which resolves once the
```
stefanpenner: Run is part of metal, and register waiter is part of test, i don't know if this should be part of run namespace.

stefanpenner:
can side-by side examples be provided comparing this to registerWaiter?

text/0000-run-callback.md
davewasmer: That's fair. My thought was that the behavior here matches similar behavior with the other Ember.run.* methods (i.e. want an Ember aware setTimeout? `run.later`. Want an Ember aware callback? `run.callback`). It mixes concerns under the hood, but from my perspective when trying to solve this problem, `Ember.run` was my gut reaction first place to check.

Certainly open to suggestions of a better spot for it to live.

davewasmer:
@stefanpenner added example code for both current approaches in the Alternatives section (manual waiters and direct state manipulation).

I took a stab at the example code, but if either one seems uncharitable to the current alternatives, feedback is welcome.

text/0000-run-callback.md
```diff
@@ -0,0 +1,74 @@
...
+callback is invoked.
+
+By having the application register the waiters itself, we can avoid compromising the "purity" of the acceptance tests which would otherwise need to reach into the application's running state.
+
+The return value of of `Ember.run.callback()` would be the wrapped function, but the waiter would be registered immediately. If the callback is no longer needed, the user could cancel the waiter by passing the callback to `Ember.run.cancel()` (matching the other async `Ember.run.*` methods).
```
courajs: The return value __of of__

stefanpenner:
The manual registration example is somewhat strange. Wouldn't the waiter be in the app code, so that all tests interacting with that entity would work the same?


Ember.run.callback itself has a fairly vague name, I could users using it in place of run.bind and being confused by the outcome.any ideas?

Another thing, which types of callbacks should this be passed to, all/any/some. How can we make users fall into a pit of success.

davewasmer:
> The manual registration example is somewhat strange. Wouldn't the waiter be in the app code, so that all tests interacting with that entity would work the same?

Fair point. I'll add an example demonstrating that.

> Ember.run.callback itself has a fairly vague name, I could users using it in place of run.bind and being confused by the outcome.any ideas?

Totally agree, I don't really like the name either, I just couldn't think of a better one :wink: 

It does bear some similarity to `run.bind()`, since you return a function rather than immediately invoking something (i.e. like `schedule()`), and the return function should be used as a callback function itself, so that's what lead me to `callback()`, but suggestions are certainly welcome.

> Another thing, which types of callbacks should this be passed to, all/any/some. How can we make users fall into a pit of success.

It should be used for any async callbacks that aren't already tracked by Ember.

I can see a few places where users might run awry:

1. Using `callback()` for already-tracked asynchrony. I.e. `new RSVP.Promise(Ember.run.callback(...))`. This could be straightforward to guard against; Promise, ajax, other run methods, etc could all check to see if the function they received was produced by `run.callback()`, and warn if so.
2. Using `callback()` for synchronous callbacks. I don't see any bugs this would produce, but it's not strictly necessary. Perhaps we could try to detect if the callback is invoke synchronously and warn the user?
3. Assuming that it behaves more like `schedule()` than `bind()`, i.e. assuming immediate invocation. This one is trickier to guard against (not sure how we could detect that the function _wasn't_ invoked). The ugliest form of this is if it occurs during tests, which would simply cause them to hang with no explanation. Perhaps some kind of debugging tool to inspect what waiters the app is currently waiting on? Although that's probably outside the scope of this RFC.

@stefanpenner do these points belong in the RFC somewhere (Unresolved Questions)? Or should we let the discussion play out first?

text/0000-run-callback.md
davewasmer: Good catch :+1:

stefanpenner:
> Using callback() for already-tracked asynchrony. I.e. new RSVP.Promise(Ember.run.callback(...)). This could be straightforward to guard against; Promise, ajax, other run methods, etc could all check to see if the function they received was produced by run.callback(), and warn if so.

This isn't an async method, and we don't really have control checking this from the callee side. Literally any-place that we would want to use this, is more or less by definition out of our control. `$(..).click` `setTimeout` etc.etc.

> Perhaps some kind of debugging tool to inspect what waiters the app is currently waiting on? Although that's probably outside the scope of this RFC.

Maybe, but this would be valuable as well (maybe a different RFC) aka, named waiters that track their callstack that invoked them during testing for easy inspection.

Introducing a high-level API does usually mean we need to investigate improved debug-ability.

> Assuming that it behaves more like schedule() than bind(), i.e. assuming immediate invocation. This one is trickier to guard against (not sure how we could detect that the function wasn't invoked).

This isn't the hazard I'm concerned with, let me demonstrate, given:

```
this.$().click(Ember.run.bind(this, this.method));
```

If someone was like, oh a callback, I should use `Ember.run.callback` in some cases this will be desired and in others it won't be. It may even cause unexpected, or unexplained test hangs.

won't be:
```js
didInsertElement() {
  this.$().click(Ember.run.callback(this, this.method));
  // testing would just hang when the component was rendered, regardless of intent.
}
```

will be:

```js
didInsertElement() {
  this.$().click(_ => {
    $(this).one('didLoad', Ember.run.callback(this, this.method)); // likely works as expected
  });
}
```

----

I do think we need to improve this, (i literally spent 2 days last week tracking down related issues in test suites). But maybe first several common scenarios that don't pause tests but should should be enumerated. That way we can look at each case, and decide what the appropriate solution is (I would venture to guess, we could do more automatically then we do today).

Another thing to consider, is testWaiters can be aborted but callback in this RFC cannot. Their is no way  (currently) to say, oops NVM no longer block tests. Here is a quick example (inspired by one of the fixes I made last week);

```js
let complete = false;
Ember.Test.registerWaiter( _ => complete);

this.one('didLoad', _=> complete = true); // assume cleanup of the error listener
this.one('error', _=> complete = true);     // assume cleanup of the error listener
```

davewasmer:
@stefanpenner:

> This isn't an async method, and we don't really have control checking this from the callee side. Literally any-place that we would want to use this, is more or less by definition out of our control. $(..).click setTimeout etc.etc.

My point was the opposite - that we might want to warn if someplace _inside_ our control receives a `callback()` wrapped function, since like you pointed out, by definition that's likely a mistake. Or perhaps I'm misunderstanding your comments ...

> If someone was like, oh a callback, I should use Ember.run.callback in some cases this will be desired and in others it won't be. It may even cause unexpected, or unexplained test hangs.

Ah, okay. Your examples make sense. Yea, that could be tricky, and I don't see a good way of preventing that apart from choosing a better name and good docs, since there isn't a great way to tell those use cases apart in an automated fashion (AFAICT).

> But maybe first several common scenarios that don't pause tests but should should be enumerated.

Sure. Should I add these additional examples to the RFC doc? Suggestions of scenarios are welcome.

> Another thing to consider, is testWaiters can be aborted but callback in this RFC cannot.

The callback can be canceled. From the RFC:

> If the callback is no longer needed, the user could cancel the waiter by passing the callback to Ember.run.cancel() (matching the other async Ember.run.* methods).

This is implemented [in the addon spike](https://github.com/davewasmer/ember-run-callback/blob/master/addon/index.js#L43). Are you looking for some other kind of "abort" behavior?

stefanpenner:
Ah. I must have missed the cancellation aspect. 

Yes I think enumerating the use-cases explicitly would be good. It will help this discussion, and maybe spark complementary solutions.

lolmaus:
You have two *Current solutions* in the RFC which illustrate quite vividly the pain this RFC is aiming to resolve. 👍 

But can you please add a code sample to the RFC demonstrating what those *Current solutions* could be replaced with?

lolmaus:
Oops, didn't notice this RFC is old!

Is it still relevant?

wagenet:
I'm going to assume this is no longer relevant. If I'm incorrect, let me know and I'll see what I can do to get it moving forwards again.

PR #116, opened 2016-02-01
Define routes without `this`

workmanw:
[Rendered](https://github.com/workmanw/rfcs/blob/define-routes-without-this/text/0000-define-routes-without-this.md).

I've always felt that using `this` when defining the route was awkward and deceptive, but the source of this RFC was this ember.js issue: emberjs/ember.js#12884 .

### DIFF
--- a/text/0000-define-routes-without-this.md
+++ b/text/0000-define-routes-without-this.md
@@ -1,0 +1,78 @@
+- Start Date: 2016-02-01
+- RFC PR: (leave this empty)
+- Ember Issue: (leave this empty)
+
+
+# Summary
+
+Improve the route definition process by migrating away from using `this.route(...)` to using
+`r.route(...)` where `r` is the argument provided to the function callback at each level.
+Using `this.route(...)` will continue to work for backwards compatibility.
+
+
+# Motivation
+
+The primary purpose of this change is to eliminate confusion about what `this` actually is
+when defining routes. Using a callback argument helps to clarify and enforce that it's changing at
+each level, as opposed to `this` which is seemingly the same throughout the hierarchy.
+
+A secondary purpose of this change is to enable the use of ES6 arrow functions, which are viewed by many
+as more elegant.
+
+
+# Detailed design
+
+The idea for this came from the way that [ember-test-helpers](https://github.com/switchfly/ember-test-helpers)
+passes it's `assert` helper into each test as the function argument.
+
+When declaring an application's routes instead of using `this.route(...)` you could use
+`r.route(...)`. Given this example:
+
+```js
+Router.map(function() {
+  this.route('search', function() {
+    this.route('index');
+    this.route('advanced');
+  });
+});
+```
+
+It would instead be written as:
+
+```js
+Router.map(r => {
+  r.route('search', r => {
+    r.route('index');
+    r.route('advanced');
+  });
+});
+```
+
+*Note: Using `r` as the variable could be documented more verbosely as `router` or anything else.*
+
+I believe this change only involves touching [Ember](https://github.com/emberjs/ember.js).
+On first look, it does not appear that [tildeio/route-recognizer](https://github.com/tildeio/route-recognizer)
+or [tildeio/router.js](https://github.com/tildeio/router.js) are affected.
+
+These proposed changes are backwards compatible with existing API for defining routes. Ideally
+using `this.route` would be deprecated in a future version and removed in Ember 3.0. But I think
+that is outside the scope of this RFC.
+
+
+# Drawbacks
+
+It's possible this could introduce some confusion about the support of `this` vs. the argument,
+as well as confusion about using arrow functions with older versions of Ember/Router.
+Mixing arrow functions and using `this` can never be supported, so it must be carefully
+documented and messaged to the user.
+
+# Alternatives
+
+Since the primary goal of this is to eliminate confusion and prevent new users from tripping up,
+it's possible a linting solution such as ember-suave could be applied to ensure the arrow function
+is not used for route definitions. However, this would require that the user is actually using the linter.
+
+
+# Unresolved questions
+
+There are no unresolved questions that I am aware of.


locks:
I believe the API proposed is very prone to errors by typos and misuse. For example:

```javascript
Router.map(r => {
  r.route('search', s => {
    r.route('index');
    s.route('advanced');
  });
});
```

At the suggestion of another embereño I have started implementing an [alternative API](https://github.com/locks/ember-router-dsl/blob/61a591e3fb562dc5db64909c9d6aedfafad1de88/tests/dummy/app/router.js#L10-L19):

```javascript
Router.map(() => {
  route('search', () => {
    route('index');
    route('advanced');
  });
});
```

workmanw:
@locks So I'm torn on that. 

I certainly like the syntax of your alternative much better. It's even easier to read. But my conflict is that it's also kind of "magically" managing the scope hierarchy. While there isn't anything technically wrong that, I had to stop and think about how it's working to feel comfortable using it. I personally favor more explicit solutions (IE my original proposal).

¯\\\_(ツ)_/¯ Hopefully a few more folk will weigh in. If the consensus is around your proposal I'd be more than happy to update this RFC.

rwjblue:
I haven't looked into the linked commit yet, but generally share @workmanw's concern. Much work lately has gone into making things feel like "less magic", so we should be careful to explain how this would be taught (and perceived by new users). Generally, I think these issues are solvable, but they need thought...

nathanhammond:
What about MOAR MAGIC? :tophat:

The Router DSL as it stands is slow, not to say we couldn't speed it up, just describing present state. One thing that we've been discussing is preparsing it into a primitive result that we could feed into route-recognizer which has no bootup cost beyond parsing. As a part of that process there is theoretically nothing stopping us from designing this with literally any API. (It's likely already going to walk and statically analyze an AST.) Get @mmun up in here with a few transforms and _poof_ Bob's your uncle. :smile:

On a more serious note, in general I'm opposed to a pure syntactic sugar change without extremely compelling benefits. The vast majority of people are using `ember g route foo` to get their routes, and that adds it into router.js for them. More generally I'm hesitant to fiddle with this while we're still working on figuring out how async engines is going to work unless we spend some time exploring those consequences in the RFC (@trentmwillis has an incoming RFC related to this).

workmanw:
> On a more serious note, in general I'm opposed to a pure syntactic sugar change without extremely compelling benefits.

I generally agree. And in this case I do see it as mostly an aesthetic change. But to be clear I think the fact that it uses `this.route...` is wonky. In my view, this RFC was more about bring clarity to the API and resolving the potential confusion of `this`, than it is about using arrow functions. Arrow functions are a secondary benefit to me.

> More generally I'm hesitant to fiddle with this while we're still working on figuring out how async engines is going to work unless we spend some time exploring those consequences in the RFC (@trentmwillis has an incoming RFC related to this).

Yea I hear that. Though I think it's safe to say the past year of ember has been a lot of "big features". I think the argument that we shouldn't make a small change because of a coming big change should be used sparingly. It can be a little off putting for someone who is just trying to pitch in where they can.

To give you an idea of how small of a code change this is I made this commit:
[1e4f7c8b](https://github.com/workmanw/ember.js/commit/1e4f7c8ba4c89f633536309b47ec04cee8aee90c). Disclaimer, I didn't fully test that commit so there might be slightly more to it, but that's the general idea.

Obviously the code commit is the easy part with this RFC, documenting it and support it is the hard part.

text/0000-define-routes-without-this.md
```diff
@@ -0,0 +1,78 @@
...
+or [tildeio/router.js](https://github.com/tildeio/router.js) are affected.
+
+These proposed changes are backwards compatible with existing API for defining routes. Ideally
+using `this.route` would be deprecated in a future version and removed in Ember 3.0. But I think
+that is outside the scope of this RFC.
+
+
+# Drawbacks
+
+It's possible this could introduce some confusion about the support of `this` vs. the argument,
```
stefanpenner: this will also need to update router generation tools

GavinJoyce:
> One thing that we've been discussing is preparsing it into a primitive result that we could feed into route-recognizer which has no bootup cost beyond parsing.

I'd very much like this to happen. Has there been discussions on a potential API? It seems like this could easily be implemented as an addon first and then considered for inclusion in Ember. I'd be happy to build or contribute to the addon

jmurphyau:
One thing I've put some time thinking into - with the widely used ember-cli-deploy in mind and bringing the app I was building come closer to deployment phase - is how to extract a list of URLs based on the ember router config.. I found there was something about the way defining routes worked and the way I was putting together my nginx config file to direct those routes to the static ember page that made it difficult to do that particular step programmatically.. I'm not sure how much energy I put into it now it was a while ago and to be honest..


I guess the reason for mentioning it here is to express that if routes were defined in a way that allowed for easier extraction to a simple list - one line per route:
/blog/*
/blog/*/comments
/blog/*/comment/*
Then that would open up the ability to use routes in bash scripts that are used to create things - in my case an nginx config file

My comments probably go beyond the scope of removing 'this' in route definitions - but I've now typed all of that on my phone on the train home - so I'm going to press comment anyway :) haha


To be honest it was a while ago

workmanw:
There has been little action here and this was just a very trivial thought from the get go. It could easily be pursued at a later date with little effort. I'm closing this. Thanks to those who participated in the discussion.

nathanhammond:
@workmanw There is lots of movement in this area, actually.

See:
- https://github.com/emberjs/rfcs/pull/154
- http://www.nathanhammond.com/route-recognizer

I've just about completed a full rewrite of route-recognizer and it's closing in on a perfect time to look at this in more depth. I have the backing data structure that everything needs to be parsed into defined which means we can now entertain making any changes that we want inside of Ember's `RouterDSL`. I'd encourage you to leave this open, though making it a bit more general to discuss an API proposal instead of just "dropping `this`."

locks:
> I'd encourage you to leave this open, though making it a bit more general to discuss an API proposal instead of just "dropping `this`."

I would rather have a new RFC be drafted and start the discussion anew. Furthermore, I think we should explore the problem space and see what solutions we can come up with so the proposal has some direction to it.

PR #120, opened 2016-02-11
Introduce Route Serializers

trentmwillis:
[Rendered](https://github.com/trentmwillis/rfcs/blob/route-serializers/text/0000-route-serializers.md).

### DIFF
--- a/text/0000-route-serializers.md
+++ b/text/0000-route-serializers.md
@@ -1,5 +1,5 @@
 - Start Date: 2016-02-11
-- RFC PR: (leave this empty)
+- RFC PR: [emberjs/rfcs#120](https://github.com/emberjs/rfcs/pull/120)
 - Ember Issue: (leave this empty)
 
 # Summary
--- a/text/0000-route-serializers.md
+++ b/text/0000-route-serializers.md
@@ -1,0 +1,98 @@
+- Start Date: 2016-02-11
+- RFC PR: (leave this empty)
+- Ember Issue: (leave this empty)
+
+# Summary
+
+This RFC proposes introducing a new construct called a `RouteSerializer` to replace the existing `Route#serialize` method.
+
+# Motivation
+
+As we move towards an increasingly asynchronous world, we need to separate knowledge about how to _**link** to a route_ and how to _**enter** a route_. Linking to a route should be able to happen _before_ a route object is instantiated, but entering a route can only occur after a route object is instantiated. However, in our current reality, a route object must be instantiated before being able to link to _or_ enter a route.
+
+By separating these concerns, we can preemptively load the information on how to link to a route without also requiring all the knowledge of how to enter that route. This would be beneficial in both the asynchronous and synchronous worlds by allowing us to defer work.
+
+Since the `serialize` method is the only method currently used by the `Route` class to define how to link to a route, the proposal is to extract this into a standalone construct.
+
+_**Note:** there are also changes that need to be made in router.js for the preemptive loading proposed here to actually work, but we can prepare for that future world by creating a separation of concerns now._
+
+# Detailed Design
+
+Since the current API is a simple function, the new API will also be a simple function that mirrors the signature of the original. However, in order to make sure future compatibility and API changes can be done smoothly, it will require being wrapped in a helper function, `Ember#routeSerializer`. Here's an example:
+
+```js
+// route-serializers/foo.js
+import Ember from 'ember';
+const { routeSerializer } = Ember;
+
+export default routeSerializer(function(model, params) {
+   // do serialization stuff
+});
+```
+
+This means that refactoring existing code should be super simple. Here's the example currently given in the Ember docs (updated to reflect Ember-CLI):
+
+```js
+// app/routes/post.js
+import Ember from 'ember';
+export default Ember.Route.extend({
+  model: function(params) {
+    // the server returns `{ id: 12 }`
+    return Ember.$.getJSON('/posts/' + params.post_id);
+  },
+
+  serialize: function(model) {
+    // this will make the URL `/posts/12`
+    return { post_id: model.id };
+  }
+});
+```
+
+Here is that same code refactored for the proposal:
+
+```js
+// app/routes/post.js
+import Ember from 'ember';
+export default Ember.Route.extend({
+  model: function(params) {
+    // the server returns `{ id: 12 }`
+    return Ember.$.getJSON('/posts/' + params.post_id);
+  }
+});
+
+// app/route-serializers/post.js
+import Ember from 'ember';
+export default Ember.routeSerializer(function(model) {
+  // this will make the URL `/posts/12`
+  return { post_id: model.id };
+});
+```
+
+# Migration Plan
+
+Even though the refactoring needed here is easy, we still need a clear (though simple) migration plan.
+
+The first step will be to introduce the `RouteSerializer` concept into both Ember and Ember-CLI. Once that is done, we can deprecate `Route#serialize` over the remainder of the 2.x series to give developers the time to update their code base. We can then remove support in 3.x.
+
+As noted in the "Motivation" section, there is still work to be done in router.js in order to support this separation of concerns. Due to this, the initial implementation of `RouteSerializer` will essentially be a polyfill that proxies the corresponding `Route#serialize` property. This will set us up for an internal migration at a later point to actually separate the two; this, however, should not affect developers as it will be internal.
+
+# Pedagogy (How We Teach This)
+
+Once the `RouteSerializer` concept is introduced, the Ember guides will need to be updated to reflect this new construct. Those changes should be relatively straightforward as shown in the example above. This will help introduce the feature to new users and those users that haven't used `Route#serialize` before.
+
+For existing users, we can introduce this feature through deprecation warnings (as mentioned above). The deprecations should briefly introduce `RouteSerializers` and point to an appropriate deprecation guide that explains how to migrate.
+
+# Drawbacks
+
+- Adds another function to the API.
+- Adds another type of construct to manage.
+
+# Alternatives
+
+- Introduce a more holistic construct. Instead of keeping the API as a function, we could introduce a class that would represent all the information needed to link to a route. Since there is not currently any other information needed, this seems overkill.
+- Don't do this and continue loading and instantiating all route information upfront. This is bad for performance and keeps concerns coupled.
+
+# Unresolved Questions
+
+- Do we wish to apply a similar approach for default query params? And if so, do we wish to incorporate that approach into this new construct?
+- Where should the new `routeSerializer` method live in a modules world? `ember-route-serializer`?
--- a/text/0000-route-serializers.md
+++ b/text/0000-route-serializers.md
@@ -22,13 +22,12 @@ Since the current API is a simple function, the new hash option will also be a s
 
 ```js
 // app/router.js
+function serializePostRoute(model, params) {
+  // serialize the model into the dynamic paths
+}
+
 export default Router.map(function() {
-  this.route('post', {
-    path: '/post/:id',
-    serialize(model, params) {
-      // serialize the model into the dynamic paths
-    }
-  });
+  this.route('post', { path: '/post/:id', serialize: serializePostRoute });
 });
 ```
 
@@ -70,14 +69,13 @@ export default Ember.Route.extend({
 });
 
 // app/router.js
+function serializePostRoute(model) {
+  // this will make the URL `/posts/12`
+  return { post_id: model.id };
+}
+
 export default Router.map(function() {
-  this.route('post', {
-    path: '/post/:id',
-    serialize(model) {
-      // this will make the URL `/posts/12`
-      return { post_id: model.id };
-    }
-  });
+  this.route('post', { path: '/post/:id', serialize: serializePostRoute });
 });
 ```
 
--- a/text/0000-route-serializers.md
+++ b/text/0000-route-serializers.md
@@ -4,47 +4,56 @@
 
 # Summary
 
-This RFC proposes introducing a new construct called a `RouteSerializer` to replace the existing `Route#serialize` method.
+This RFC proposes replacing the existing [`Route#serialize`](http://emberjs.com/api/classes/Ember.Route.html#method_serialize) method with an equivalent method on the options hash passed into [`this.route` within `Router#map`](http://emberjs.com/api/classes/Ember.Router.html#method_map). The primary goal here is to enable asynchronous Engines by decoupling information about how to link to a route from the actual route object.
 
 # Motivation
 
-As we move towards an increasingly asynchronous world, we need to separate knowledge about how to _**link** to a route_ and how to _**enter** a route_. Linking to a route should be able to happen _before_ a route object is instantiated, but entering a route can only occur after a route object is instantiated. However, in our current reality, a route object must be instantiated before being able to link to _or_ enter a route.
+As we move towards an increasingly asynchronous world with Engines, we need to separate knowledge about how to _**link** to a route_ and how to _**enter** a route_. Linking to a route should be able to happen _before_ a route object is instantiated, which is the behavior needed to asynchronously load Engines. However, in our current reality, these concerns are coupled and a route object must be instantiated before being able to link to _or_ enter a route.
 
 By separating these concerns, we can preemptively load the information on how to link to a route without also requiring all the knowledge of how to enter that route. This would be beneficial in both the asynchronous and synchronous worlds by allowing us to defer work.
 
-Since the `serialize` method is the only method currently used by the `Route` class to define how to link to a route, the proposal is to extract this into a standalone construct.
+Since the `serialize` method is the only method currently used by the `Route` class to define how to link to a route, the proposal is to extract this method into the space which currently contains the other linking information (e.g., the Router's map).
 
-_**Note:** there are also changes that need to be made in router.js for the preemptive loading proposed here to actually work, but we can prepare for that future world by creating a separation of concerns now._
+_**Note:** this separation of concerns will also need to be implemented in router.js for the preemptive loading proposed here to actually work, but we can prepare for that future world by creating a separation of concerns within application code now._
 
 # Detailed Design
 
-Since the current API is a simple function, the new API will also be a simple function that mirrors the signature of the original. However, in order to make sure future compatibility and API changes can be done smoothly, it will require being wrapped in a helper function, `Ember#routeSerializer`. Here's an example:
+Since the current API is a simple function, the new hash option will also be a simple function that mirrors the signature of the original. Here's an example:
 
 ```js
-// route-serializers/foo.js
-import Ember from 'ember';
-const { routeSerializer } = Ember;
-
-export default routeSerializer(function(model, params) {
-   // do serialization stuff
+// app/router.js
+export default Router.map(function() {
+  this.route('post', {
+    path: '/post/:id',
+    serialize(model, params) {
+      // serialize the model into the dynamic paths
+    }
+  });
 });
 ```
 
-This means that refactoring existing code should be super simple. Here's the example currently given in the Ember docs (updated to reflect Ember-CLI):
+Preserving the current function signature means that refactoring existing code should be simple in most cases. Here's the example currently given in the Ember docs (updated to reflect Ember-CLI):
 
 ```js
 // app/routes/post.js
 import Ember from 'ember';
 export default Ember.Route.extend({
-  model: function(params) {
+  model(params) {
     // the server returns `{ id: 12 }`
     return Ember.$.getJSON('/posts/' + params.post_id);
   },
 
-  serialize: function(model) {
+  serialize(model) {
     // this will make the URL `/posts/12`
     return { post_id: model.id };
   }
+});
+
+// app/router.js
+export default Router.map(function() {
+  this.route('post', {
+    path: '/post/:id'
+  });
 });
 ```
 
@@ -54,17 +63,21 @@ Here is that same code refactored for the proposal:
 // app/routes/post.js
 import Ember from 'ember';
 export default Ember.Route.extend({
-  model: function(params) {
+  model(params) {
     // the server returns `{ id: 12 }`
     return Ember.$.getJSON('/posts/' + params.post_id);
   }
 });
 
-// app/route-serializers/post.js
-import Ember from 'ember';
-export default Ember.routeSerializer(function(model) {
-  // this will make the URL `/posts/12`
-  return { post_id: model.id };
+// app/router.js
+export default Router.map(function() {
+  this.route('post', {
+    path: '/post/:id',
+    serialize(model) {
+      // this will make the URL `/posts/12`
+      return { post_id: model.id };
+    }
+  });
 });
 ```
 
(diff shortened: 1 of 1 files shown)


nathanhammond:
All of my feedback was addressed prior to this being posted. :smile:

@rwjblue @dgeb @stefanpenner and everybody else following along at home we'd love to hear your thoughts.

raytiley:
This is pretty similar to a discussion @rwjblue and I had a while back.

I think the `RouteSeralizer` should probably be a full blown object that can serialize the route url as well as any query-params defined for that route. Being able to link to Routes that use query params without needing to load the full route / controller etc is super important.

trentmwillis:
@raytiley thanks for the input. Whether or not to roll query param support into this is definitely an open question. However, it looks like the only relevant QP info for linking to a route are [default values](https://guides.emberjs.com/v2.3.0/routing/query-params/#toc_default-values-and-deserialization) (which get removed from URLs), but [there has been discussion over whether that behavior is even correct/needed](https://github.com/dgeb/ember-engines/issues/6#issuecomment-167356262).

Are you aware of any other QP-related behavior that would be need before attempting to enter a route?

davewasmer:
Just two cents from one of the "folks at home" :wink: ...

As someone who has worked daily on a decently complex production Ember app for 1.5+ years: I'm not entirely sure I understand the need or use case for this. To be clear, I'm sure it exists, but from a pedagogy perspective, the provided examples look identical to me, except now the serialize function is separate. I get that it will improve the handling of asynchrony in routes, but it's not clear _why_.

In my experience training the junior devs on my team, a common struggle learning Ember is the number of concepts to grapple with. Now, there's a decent chance I'm just being dense at the moment and the use case is obvious. But if that's not the case, I'd be concerned about introducing a new "top-level" concept. Especially if I can't grok it quickly, it will be even tougher and seem more arbitrary for the less experienced devs on the team.

Like I said above, I'm sure something like this change is necessary. I'd just caution that the reason _why_ it exists should be articulated clearly to even beginners. Certainly the RFC language doesn't need to be written at that level, but just a friendly heads up from the audience.

raytiley:
> but it's not clear why.

@davewasmer imagine you have a BIG app with a user facing section and an admin section. You probably want to link to different admin sections from the main app, but you want to reduce your payload size by not delivering the admin section to the client unless they enter it.

This RFC allows you to link to those sections of the app, without them actually being downloaded and evaluated by the client. It's almost entirely for performance, although it can make dealing with larger code bases more maintainable to separate them.

raytiley:
@trentmwillis 

It's been a long while since I worked on the query params stuff, but basically the Route object as two hooks for serializing  / deserializing the query-parsms. I *thought* they were public, and not sure if they were supposed to be but get turned private in the great private / public change over a few months ago? @machty might know.

The hooks: https://github.com/emberjs/ember.js/blob/76e9c595809957c14d6616d45be35ae2fe94fd3b/packages/ember-routing/lib/system/route.js#L361-L395

The idea for overriding them was that you might want to take a a complex object and control the serialization so you could make it prettier than calling `JSON.stringify` on it and sticking that in the URL.

workmanw:
I would ditto what @davewasmer said. Our app has been around for a while. We almost never have to implement `serialize` because the default implementation will ulimate do `get(model, 'id')`.

We do have a couple of cases where we have "base route definitions" that similar routes extend, eg:

```
import BaseFileRoute from 'app/routes/base-file';

export default BaseFileRoute.extend({
  /* ... */
});
```

Alternatively, you could create a reusable mixin. Also, you could just create a reusable pure functions for serialization/deserialization based on type.

So with all of those existing options in place. I personally think this adds complexity without providing much benefit to most users. And advanced users have some options as mentioned.

stefanpenner:
> will improve the handling of asynchrony in routes, but it's not clear why.

I'm not sure what the solution should be (this may be it), but the problem is as follows:

1. many want to be able to lazy load parts of their app
2. many want to link to a portion of the app that is lazy loaded
3. url construction is  (do to serialization) a combined effort of link-to, the router  and the target route.
4. if the target route does not yet exists, how does one calculate the URL?

It's a bit of chicken + egg problem, one approach is to separate out the part of a route responsible for serialization, so that it can be part part of the eagerly loaded bundle.

In-practice, although some users rely on this feature (it serves a real purpose), many more users are totally unaware of its existence and get by just fine. I would say adding an extra *thing* sure is unfortunate (and if it can be avoided it should), but in general this *thing* isn't on that many critical paths... 

I did a quick survey of several million lines ember related code on my machine, and this is used 7 times, but those use-cases are legitimate. 

What does this mean? Adding this concept (if that is the approach taken) has very small downside (as it affects very few users), and an upside of enabling asynchronous loading of engines.

stefanpenner:
> We almost never have to implement serialize 

Then this RFC isn't applicable to you. This RFC exists to explore/solve a real problem blocking async engines, this feature (which you don't use) needs an alternative.

raytiley:
> and if it can be avoided it should

Could it be solved by tooling? Since we are just extracting a function this is currently on the route to the eagerly loaded bundle, couldn't the ember-engines addon do this at build time?

stefanpenner:
> Since we are just extracting a function this is currently on the route to the eagerly loaded bundle, couldn't the ember-engines addon do this at build time?

This is an option, and was discussed earlier, it wasn't ruled out, but some concerns arose let me share:. The transformation is pretty surprising, as basically this one function can only support some limited subset of the language. Enforcing this by creating a different "thing"

Panman8201:
> This RFC exists to explore/solve a real problem blocking async engines

I think that needs to be mentioned in the RFC. When I seen Ray's response I was going to reply back "ah, engines?" not realizing this is _for_ async engines.

I agree with others that it seems a bit overkill but understand if it's needed for async engines, yet another type but what do you do...

mmun:
How does putting it in a separate module help? Why not just ship the entire route.js files themselves? I'm not opposed to this RFC but it should be spelled out.

mmun:
I guess there are people who heavily put actions in routes who don't like the idea of shipping all the route.js files. I keep the vast majority of my actions on controllers so my routes end up being lean (just a model hook and redirection logic typically).

trentmwillis:
> How does putting it in a separate module help? Why not just ship the entire route.js files themselves?

A separate module which is just a function means that there is no backing object that would need to be instantiated. This is tremendously beneficial especially in apps where all routes extend from some other base class. It also means any miscellaneous dependencies used by the route won't need to be loaded just for this one function.

> I don't want to have yet another top-level folder in my project directory. 

I agree this is a strong downside, but as others have mentioned, the majority of projects will never even see this.

> If you had the route.js files in advance, could you preload the models while you were loading the async engine? Are we throwing the baby out with the bath water?

Potentially, but going back to my first response what happens when your route depends on a parent class and that class imports utilities from other addons (this occurs in an application I work on)? We'll have to draw a line somewhere. Maybe this isn't the right line, but we would have to establish some criteria for it.

trentmwillis:
@raytiley if there is any sort of public way to serialize query params then I think we'd definitely want to roll that into this construct.

> Could it be solved by tooling?

I imagine so, but aside from @stefanpenner's concern, it would also add more "magic" than I feel should occur. I would definitely think an ember-watson addition should be created to help migration though as I think this is a perfect candidate for it.

mmun:
> this occurs in an application I work on

Mine too (using ember-simple-auth). And for the reason I'm ok with this RFC with the current state of things. I do think the next logical thing will be wanting to load the models in parallel with the engine whenever possible. This is similar in spirit to the prefetch RFC.

trentmwillis:
> This is similar in spirit to the prefetch RFC.

We use prefetch currently as well and would definitely like a solution that accommodates it and async engines, but we haven't been able to think of a solution within the current system.

stefanpenner:
> This is similar in spirit to the prefetch RFC.

seems orthogonal

courajs:
> Once that is done, we can deprecate Route#serialize over the remainder of the 2.x series to give developers the time to update their code base. We can then remove support in 3.x.

Requiring my tiny route serialize function to be in its own file seems like pointless overhead for most apps. By not deprecating here, can we keep the common case simpler? It seems to me the existence of the route serializer module could just override a `Route#serialize` if it exists.

trentmwillis:
> It seems to me the existence of the route serializer module could just override a `Route#serialize` if it exists.

This would be fine if the plan was to continue using the entirety of `Route` for all of routing, but the goal is to "separate knowledge about how to link to a route and how to enter a route".

So instead of overriding `Route#serialize` with a route-serializer, we would need to do the reverse and extract `Route#serialize` to use as a route-serializer so that we separate out knowledge about how to link to a route. That would require instantiating the entire route which winds us up back where we started.

mmun:
@stefanpenner Wat? The spirit of prefetch is to load things faster. The spirit of loading the model and engine together is to load things faster. It's literally the same purpose.

Panman8201:
Instead of yet another file / folder structure, what if the `route.js` file `export`s the function.?. It would be clear that function is separate from the actual `Route` object, and tooling could `import` just that function. I suppose that would require all `route.js` files to export a function, or is there a way to handle a missing export??

stefanpenner:
maybe we are talking about the same thing but words are different.

I'm going to describe what I believe is motivating the RFC, just to be sure the motivations of the RFC are correctly understood.

* prefetch is improving the ability (during transitions) to pipeline asynchrony. It does so, by allowing every route, that is likely to be entered to participate early.
* async engines (which this RFC aims to unblock, by allowing one to generate link-to href's for code not yet loaded) enable the on-demand loading of segments of ones application.

Improved concurrency seems like a cool emergent property of async engines, as one can imagine a slim shell that lazy-loads all large slabs of work, but I would say that is an emergent property, rather then a motivating factor. Instead I would say the motivating factors is primarily to reduce up-front load times, by making different aspects of the application pay as you go. So rather then incurrent byte size the penalty of the entire app, one only incurs the penalty of that segment of the application.

stefanpenner:
> Instead of yet another file / folder structure, what if the route.js file exports the function.?. It would be clear that function is separate from the actual Route object, and tooling could import just that function. I suppose that would require all route.js files to export a function, or is there a way to handle a missing export??

some responses: 

* It would be rare for users to add this file, as in-practice it seems very infrequently used (people don't need it, they wont notice it)
* merging serialize with its route.js, brings us back to non-obvious separation which requires more magic to accomplish.

stefanpenner:
people about to comment, please be sure to read [the first line](https://github.com/emberjs/rfcs/pull/120/files#diff-2407e9d4ceec7bdb9eb20764895cd4feR7) as this RFC is providing an proposed alternative to an existing feature, a solution is required to decouple an engine's routes from URL generation.

Without a solution, one cannot link to engines which have not yet been loaded.

Creating an optional entity, although quite unfortunate, is a solution which provides very clear / easily verifiable separation.

mmun:
@stefanpenner I agree with everything you wrote. :+1: I was discussing something else entirely (if you bundle all the route.js files in the engine you can do much more interesting things, like preloading engine route's models while loading the engine bundle).

I especially agree with https://github.com/emberjs/rfcs/pull/120#issuecomment-183168614. In my experience Route#serialize is very rare. I've seen one among all the apps I've worked on.

Panman8201:
> merging serialize with its route.js, brings us back to non-obvious separation which requires more magic to accomplish.

It's still a route related thing...

I suppose we should also talk about making Transforms, and possibly an Adapter layer. :trollface:

dgeb:
Since the need for custom serializers is rare (as others have said), most developers can ignore the entire concept until an app actually needs it.

I'm also not in favor of magic extractions of this method from routes. It's better off as a pure function in a standalone module with clear dependencies.

I am :+1: to this RFC. Thanks for writing this up @trentmwillis. It's just what @rwjblue and I had in mind for enabling async engines.

stefanpenner:
> (if you bundle all the route.js files that the engine defines you can do much more interesting things at the cost of those bytes, like preloading engine routes' models while loading the engine bundle, redirecting, etc. It seems complicated and I don't feel strongly about it so I'll drop it).

Please note, engines have imports + injections etc, they may end up easily retaining large chunks of the engine. Which I believe to be some we should avoid as :footprints: :gun:'s, instead clear cut easily static analyzable engines

nathanhammond:
### Regarding `prefetch`

@mmun: @trentmwillis and I completely agree that this is an incredibly interesting use case. We had a back-channel conversation about how it could be possible to make `prefetch` work with this idea of `route-serializers`. Trent's premise was "fastboot the engine" but in approximately one minute we came up with _way_ too many possible side effects to make that sort of thing work out. You'd have to manage a complete serialized state, ship that to the server, have the server understand that, and then have it emit some bootstrapping file in addition to the engine package, and who knows what else we didn't account for.

Here's the simplest failure demo which would prevent a "fastbooted engine" approach of offloading that to the server:
```
Ember.Route.extend({
  redirect() { 
    if (this.get('someService.prop')) {
      this.transitionTo('game.over');
    }
  }
});
```

Static analysis to identify it is is possible to do this/how much we would import could easily become one of @stefanpenner's foot shotguns. The advantage of an `engine` as proposed is that it is an opt-in to a single-edge, single vertex dependency graph. It's okay for that to have tradeoffs, and we can identify other optimizations to this story as emergent properties, likely in user space.

Also, having spent some time looking into it, anything we do that breaks down the engine firewall is likely better implemented as something that falls out of linker/packager work instead of being accounted for as part of the engines world. @stefanpenner is right that we could accidentally end up unraveling most of the engine by the time we grabbed all of the imports for the now-included routes.

### Magic Extractions
All it takes is one `this` in `Route#serialize` for game over. If included in `route.js` as a separate import and extracted then there are still many other things that could possibly be in scope which aren't required for the serializer. Strongly opposed as I see this as a foot shotgun. I'm with @trentmwillis that we can just `ember-watson` people to move out of the existing world.

stefanpenner:
@nathanhammond yes a this, or any free variable. Meaning, it may live physically within a routes module, but it cannot actually interact with the module. (Not without sufficient static analysis).

Which, I mean isn't off the table. But has sufficiently more unknown unknowns to make me nervous.

webark:
Just an idea, but since this is so rare, and it's just a function, could it be added into the routes definition? It's a little ugly, but due to its rarity, that might not be a bad thing, and the route definitions I assume would already be extracted out and included globally, plus, it wouldn't add another thing.

webark:
Something like..
```javascript
this.route('post', {
  path: '/post/:post_id',
  serialize: function(model) {
    // this will make the URL `/posts/12`
    return { 
      post_id: model.id
    };
  },
}, function() {
  this.route('comments');
});
```
I think is how it could/would work.

trentmwillis:
@webark I think I like that idea. It has the benefits you mentioned, plus it also keeps the information for linking to a route in a single location. I also believe we could ditch the `routeSerializer` method...

@rwjblue or @dgeb thoughts?

dgeb:
You know, I like the simplicity of @webark's idea as well. It means that an engine's `routes.js` would contain all that's needed to form links into the engine.

The aesthetics could even be improved by defining the actual serialize function above the route map and dropping it into the route's options (as opposed to defining it inline).

:+1:

trentmwillis:
> The aesthetics could even be improved by defining the actual serialize function above the route map and dropping it into the route's options

I was thinking the same thing. This would also make potential reuse easier than standalone modules.

webark:
oh totally.. Or imported if you needed to.. I just did it online for explicitness.  And glad I could help! :)

stefanpenner:
I'm dubious of this being part of the route DSL

trentmwillis:
> I'm dubious of this being part of the route DSL

@stefanpenner care to elaborate? While a bit more verbose than other aspects of it, I feel it makes sense given it describes how to construct portions of a route's url.

webark:
I would say it's fair to be dubious. But the the choice was made to have the routes be defined as methods rather then just a json blob, which allows for these types of choices to be even considered. And even though it is diverging away from the standard route DSL, it has already been mentioned that it is a very rare occurrence where this would need to be actually overwritten. It also bodes well for not having to do directory crawls for these files, and not introducing a concept that is unlike any of the others. All things currently are first and foremost objects (at least everything I can think of off the top of my head). Having a new type, albeit a seldom used feature, would cause typical roadblocks and misconceptions. You could not include a mixin, it wouldn't be something you would extend, inheritance would look different.

stefanpenner:
literally just feels bad, I will have to think about it more.

stefanpenner:
I think  it is totally related to `path`, but a really ugly relative :P

trentmwillis:
Updated the RFC to reflect @webark's approach and provide more clarity around motivation (enabling async engines). Still want to encourage discussion around it, but I think that the new approach meets our goals while addressing the concerns of introducing a totally new construct into the programming model.

stefanpenner:
@trentmwillis although not a firm :+1:, treating serialize as a stateless function, used as part of router.map was surprisingly well received.

knownasilya:
This is nice, because you define the dynamic segments in the router, so the serialize makes sense there. If you need a big process, just import a utility..

trentmwillis:
@stefanpenner @dgeb what should be the next step here? I've [submitted a POC](https://github.com/emberjs/ember.js/pull/13016) to get a feel for potential implementation and it seems mostly straightforward.

Only unresolved question (that I can think of) would be if there is a use-case for needing state in the serialize function, but I haven't seen any and can't think of any.

dgeb:
@trentmwillis yes, thanks very much for the POC! I'll review your PR soon and move this along.

> Only unresolved question (that I can think of) would be if there is a use-case for needing state in the serialize function, but I haven't seen any and can't think of any.

Same. I'm pretty sure there's consensus that state should not be needed in the `serialize` function. This assumption is critical for lazy loading routable engines, regardless of how engines export their route serializers.

trentmwillis:
Awesome! Thanks @dgeb.

> I'm pretty sure there's consensus that state should not be needed in the `serialize` function. This assumption is critical for lazy loading routable engines

Makes sense.

mixonic:
@trentmwillis thanks for the POC, looks really great. I've added this to Friday's agenda

MiguelMadero:
*tl;dr;* Why don't we just simply use a different helper when we need to use the new serialize function and still have `{{link-to}}` with the default behavior for everyone else? Anyway, given that everyone `rarely` uses serialize, moving to it even without `ember-watson` it should be trivial. Those not using engines won't need to worry until 3.0, so a big :+1: to this RFC. 


I get the importance of this for async. I had to solve this in a couple of projects where we were lazy loading packages. While having a separate Route#serialize would've helped us, we worked around it just fine in two ways:

1. Avoid using the link-to helper for routes that might not be available. Instead, we simply added a link with an `href="#routeName/param1"`, not great but it's simple and did the trick for 99% of the cases we had links across packages. The problems is that it assumes not only the serialization logic but also the path to a route, additionally, it would call the `model` hook since we don't pass a model, in most cases that's not an issue since we can rely on caching to avoid that op from being expensive, but it's something to consider.
2. Have a new helper to replace `link-to` that uses the routingService so it can transition, check the active class, etc, but relies on a default implementation of serialize.

  ```
  {{link-to-external-package 'routeName' model}}
  ```

This way, everyone can keep using the current Route#serialize in the rare cases when it's needed and only those using engines pay the cost when they have links across packages/engines.

The drawback of this is that we're creating some coupling now because the consumer needs to be aware of whether the route is fine with the default implementation for serialize, but most of the time the default implementation works just fine based on Stefan's sample of millions of line of JS code and a quick grep I did on a few large projects, that's rarely even used. 


An additional workaround, that I never used, but thought of it for cases where we wanted to override serialize was to rely on a loose function that we can pass to the `link-to-external-package` helper.

```
{{link-to-external-package 'routeName' model serializer=routeNameSerializer}}
```

Then routeNameSerializer is simply a prop in your component to a pure function that you can import.

```
// components/consumer.js
import routeNameSerializer from 'shared/utils/route-name-serializer'

  routeNameSerializer: routeNameSerializer
});
```

The problem with that is that now your component needs to know about everything that is being linked by its template and we still need to know of cases where we actually override the serializer, which ideally the component should be unaware of. This can be solved by simply relying on the serialze function proposed on this RFC defined as part of `Router.map`, but still using a new helper:

```
{{link-to-external-package 'routeName' model}}
```

text/0000-route-serializers.md
```diff
@@ -0,0 +1,111 @@
...
+export default Router.map(function() {
+  this.route('post', {
+    path: '/post/:id',
+    serialize(model) {
+      // this will make the URL `/posts/12`
+      return { post_id: model.id };
+    }
+  });
+});
+```
```
dgeb: The core team discussed the pedagogy and strongly favors separating the definition of the `serialize` function from the main route map. This is analogous to how we separate the declaration of `initialize` functions in initializers. 

@trentmwillis could you please change the example above to something like:

```
// app/router.js

function serializePostRoute(model) {
  // this will make the URL `/posts/12`
  return { post_id: model.id };
}

export default Router.map(function() {
  this.route('post', { path: '/post/:id', serialize: serializePostRoute });
});
```

dgeb:
> tl;dr; Why don't we just simply use a different helper when we need to use the new serialize function and still have {{link-to}} with the default behavior for everyone else? Anyway, given that everyone rarely uses serialize, moving to it even without ember-watson it should be trivial. Those not using engines won't need to worry until 3.0, so a big :+1: to this RFC.

@MiguelMadero if `Route#serialize` is so rarely used, there should be a very low cost to deprecating it and moving it to the route map as proposed here. I'm more concerned about the cost of remembering to use a special helper instead of `{{link-to}}`.

text/0000-route-serializers.md
trentmwillis: Sounds good. Will update this and also update the proof-of-concept PR to reflect this pattern as well.

MiguelMadero:
Agreed!

trentmwillis:
Updated the examples given to reflect separating the serialize function declaration and passing it into the router map.

dgeb:
@trentmwillis thanks for updating. I think that was the only blocker to merging this RFC as well as your POC. I'll try to keep this moving this along ...

rwjblue:
I agree RE: landing the RFC, but I think the POC needs a bit of work (some compat tests, deprecations, etc), though it is possible that has been added since I last reviewed it...

trentmwillis:
@rwjblue I've added a deprecation and tests for the cases I could think of. Just let me know of anything else that is needed.

mixonic:
Hey folks, @trentmwillis. At the last meeting I brought this up, but I didn't follow through here. Sorry :-/ I'll publish the relevant notes and link.

Updating the examples to move the serialization declaration out of that map was indeed the last blocker on merging this RFC. It would have been nice if "Pedagogy" was also updated referencing that change, as that style is related to how we present this API to end-users. I'll make a followup commit to address that.

Mergin'! Thanks for your thoughtfulness and effort as we work toward consensus all :tada:

grapho:
Sorry to be late to the party.. as this is already merged i just wondered if anyone would know if this use case was considered or not?

```
model(params) {
  return RSVP.hash({
    person: this.store.findRecord('person', params.person_id),
    miscData:  this.store.query('misc-data', { filter: params.person_id })
  });
},

serialize(model) {
  return { person_id: model.person.id };
}
```
In this case, it can easily be extrapolated by looking just at the route file, why some extra serialization is needed to perform.  When the serialization is moved over to the router.js, it will not be clear why a serialized id must exist at `model.person.id`.. one would need to examine both the router and the route in question before considering any refactor.

So, since this is already merged and seems to be the agreed path forward, I will ask.

1. is my above example uncommon, or anti-pattern?

1. is there a nicer way to do what i am doing, with the new proposed route serialization?

knownasilya:
I think it's a non-issue because how you get the id is really the same. How many levels you have to go down will just depend on the route.

This refactor in general is a loss in clarity, since the serializer is related to the model and the router has no visual context to that, but it was either this, or have it in it's own file, which is even less clear, because you don't get the `path: ':person_id'` context from the router.

PR #130, opened 2016-04-03
RFC doc for adding uniqBy to enumerable

lorrocha:


rwjblue:
Was this proposed to support the efforts in https://github.com/emberjs/ember.js/pull/12875?

lorrocha:
Ah, it is now! I snooped around for an RFC for it and didn't find any either pending or accepted, so I thought I would submit my own. I didn't think to look at the open PRs as I thought, as an api change, it would require an RFC before implementation. Sorry about that!

rwjblue:
No apologies, you are absolutely right! I think this RFC applies to that PR very well.

lorrocha:
I'll update the rfc to remove the specifics to the enumerable mixin!

text/0000-implement-uniqby.md
```diff
@@ -0,0 +1,63 @@
...
+  var uniqKeys = arguments;
+  var ret = emberA([]);
+
+  this.forEach((item) => {
+    var options = ret;
+
+    for (var i = 0; i < uniqKeys.length; i++) {
+      var key = uniqKeys[i];
+
+      options = options.filter(iter.apply(options, [key, item[key]]));
```
stefanpenner: is this supposed to be `get(item, key)`?

stefanpenner:
This seems reasonable, we should discuss at the next meeting.

text/0000-implement-uniqby.md
lorrocha: Yes, it is! I'll update it in a bit.

stefanpenner:
This is already in ember, we should be good.

PR #136, opened 2016-04-16
Add contains to includes RFC

alexspeller:
[`contains`](http://emberjs.com/api/classes/Ember.Array.html#method_contains) is implemented on `Ember.Array`, but [contains was renamed to includes in 2014](https://github.com/tc39/Array.prototype.includes/commit/4b6b9534582cb7991daea3980c26a34af0e76c6c) - should `contains` be deprecated in favour of an `includes` method in ember?

[Rendered](https://github.com/alexspeller/rfcs/blob/contains-to-includes/text/0000-contains-to-includes.md)

### DIFF
--- a/text/0000-contains-to-includes.md
+++ b/text/0000-contains-to-includes.md
@@ -1,0 +1,74 @@
+- Start Date: 2016-04-16
+- RFC PR: (leave this empty)
+- Ember Issue: (leave this empty)
+
+# Summary
+
+[`contains`](http://emberjs.com/api/classes/Ember.Array.html#method_contains) is
+implemented on `Ember.Array`, but [contains was renamed to includes in 2014]
+(https://github.com/tc39/Array.prototype.includes/commit/4b6b9534582cb7991daea3980c26a34af0e76c6c)
+- this proposal is for `contains` to be deprecated in favour of an `includes`
+method on `Ember.Array`
+
+# Motivation
+
+Motivation is to stay in line with web standards
+
+# Detailed design
+
+First, implement `includes` polyfill in compliance with `includes` spec. Polyfill
+sample from MDN is:
+
+```js
+if (!Array.prototype.includes) {
+  Array.prototype.includes = function(searchElement /*, fromIndex*/ ) {
+    'use strict';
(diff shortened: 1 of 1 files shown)


mixonic:
@alexspeller `includes` would be an extension of the native array prototype, and thus should be 100% API compatible with the native implementation that will replace it when present. For example `includes` ([MDN docs here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes)) takes an optional `fromIndex`.

Basically, I suggest that `includes` be a polyfill for the spec'd behavior. Then `contains` can be deprecated. However just renaming `contains` -> `includes` is not sufficient.

alexspeller:
@mixonic updated RFC to reflect that

text/0000-contains-to-includes.md
```diff
@@ -0,0 +1,72 @@
...
+}
+```
+
+Then, alias `contains` to `includes` with deprecation warning, deprecate in line with standard
+deprecation process. I don't believe that adding the additional parameter will
+have any affect on existing usage of `contains`.
+
+# How We Teach This
+
+Update any references in docs and guides to `includes`
```
locks: A deprecation guide also needs to get written, and since `includes` and `contains` seem to have slightly different semantics, the guide should address that with examples of how to migrate existing code.

locks: We might also want to indicate in the API documentation that this is a polyfill.

alexspeller:
@locks updated based on your comments

rwjblue:
We discussed in the core team meeting yesterday, we are 👍 on this.

mixonic:
:-D

bmeurant:
I think I can try to handle this if still available

One thing to confirm, it has to be implemented in Enumerable, right ? Not only in Array ?

rwjblue:
FYI - I created https://github.com/rwjblue/ember-runtime-enumerable-includes-polyfill to make this a bit easier for addons to avoid deprecations, use the newer syntax, and continue to support older versions of Ember.

PR #139, opened 2016-04-18
Ember.String.isHtmlSafe RFC

workmanw:
[Rendered](https://github.com/workmanw/rfcs/blob/isHtmlSafe/text/0000-isHtmlSafe.md).

More details about this RFC can be found: https://github.com/emberjs/ember.js/issues/13318

Polyfill implementation: https://github.com/workmanw/ember-string-ishtmlsafe-polyfill

### DIFF
--- a/text/0000-isHtmlSafe.md
+++ b/text/0000-isHtmlSafe.md
@@ -1,0 +1,89 @@
+- Start Date: 2016-04-18
+- RFC PR: (leave this empty)
+- Ember Issue: (leave this empty)
+
+# Summary
+
+Introduce `Ember.String.isHtmlSafe()` to provide a reliable way to determine if an object is an "html safe string", i.e. was it created with `Ember.String.htmlSafe()`.
+
+
+# Motivation
+
+Using `new Ember.Handlebars.SafeString()` is slated for deprecation. Many people are currently using the following snippet as
+a mechanism of type checking: `value instanceof Ember.Handlebars.SafeString`. Providing `isHtmlSafe` offers a
+cleaner method of detection. Beyond that, the aforementioned test is a bit leaky. It requires the developer to understand
+`htmlSafe` returns a `Ember.Handlerbars.SafeString` instance and thus limits Ember's ability to change
+`htmlSafe` without further breaking it's API.
+
+Based on our app at Batterii and some research on Github, I see two valid use cases for introducing this API.
+
+First, and most commonly, is to make it possible to test addon helpers that are expected to return a safe string. I believe this test on ember-i18n says it all: ["returns HTML-safe string"](https://github.com/jamesarosen/ember-i18n/blob/master/tests/unit/utils/i18n/default-compiler-test.js#L56-L59).
+
+The second use case is to do type checking. In our app, we have an `isString` utility that is effectively:
+
+```javascript
+import Ember from 'ember';
+
+export default function(value) {
+  return typeof value === 'string' || value instanceof Ember.Handlebars.SafeString;
+}
+```
+
+Newer versions of ember-i18n, doing `this.get('i18n').t('someTranslatedValue')` will return a safe string. Thus our `isString` utility has to consider that.
+
+
+# Detailed design
+
+`isHtmlSafe` will be added to the `Ember.String` module. The implementation will basically be:
+
+```javascript
+function isHtmlSafe(str) {
+  return str && typeof str.toHTML === 'function';
+}
+```
+
+It will be used as follows:
+
+```javascript
+if (Ember.String.isHtmlSafe(str)) {
+  str = str.toString();
+}
+```
+
+
+# Transition Path
+
+As part of landing `isHtmlSafe` we will simultaneously re-deprecate `Ember.Handlebars.SafeString`. This deprecation will
+take care to ensure that `str instanceof Ember.Handlebars.SafeString` still passes so that we can continue to
+maintain backwards compatibility.
+
+Additionally, a polyfill will be implemented to help provide forward compatibility for addon maintainers and others
+looking to get a head while still on older versions of Ember. Similar to [ember-getowner-polyfill](https://github.com/rwjblue/ember-getowner-polyfill).
+
+
+# How We Teach This
+
+I think we'll continue to refer to these strings as "html safe strings". This RFC does not
+introduce any new concepts, rather it builds on an existing concept.
+
+I don't believe this feature will require guide discussion. I think API Docs will suffice.
+
+The concept of type checking is a pretty common programming idiom. It should be relatively self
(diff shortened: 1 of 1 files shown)


stefanpenner:
I don't believe instanceof is the right thing here. Rather if toHtml is a function

stefanpenner:
I don't believe instanceof is the right thing here. Rather if toHtml is a function

text/0000-isHtmlSafe.md
```diff
@@ -0,0 +1,79 @@
+- Start Date: 2016-04-18
+- RFC PR: (leave this empty)
+- Ember Issue: (leave this empty)
+
+# Summary
+
+Introduce `Ember.String.isHtmlSafe()` to provide a reliable way to determine if an object is an "html safe string", IE was it created with `Ember.String.htmlSafe()`.
```
mixonic: `i.e.` I think, thought you were talking about the browser :-)

text/0000-isHtmlSafe.md
```diff
@@ -0,0 +1,79 @@
...
+  return typeof value === 'string' || value instanceof Ember.Handlebars.SafeString;
+}
+```
+
+Newer versions of ember-i18n, doing `this.get('i18n').t('someTranslatedValue')` will return a safe string. Thus our `isString` utility has to consider that.
+
+
+# Detailed design
+
+`htmlSafe` will be added to the `Ember.String` module. The implementation will basically be:
```
rwjblue: I believe you mean `isHtmlSafe` here (since `Ember.String.htmlSafe` is already a thing).

text/0000-isHtmlSafe.md
```diff
@@ -0,0 +1,79 @@
...
+
+```javascript
+function isHtmlSafe(str) {
+  return str instanceof SafeString;
```
rwjblue: This should do what the rest of the internals do, and check `str && typeof str.toHTML === 'function'`.

rwjblue:
@workmanw - This looks good!  Since we will be removing the deprecation that was added in 2.6.0-beta.1 can you add that deprecation as part of this RFC? When we add it back, we will make sure that it still passes a `str instanceof Ember.Handlebars.SafeString` check (to prevent the original issue).

text/0000-isHtmlSafe.md
workmanw: Hehe. Yea I should clarify that. Thanks!

text/0000-isHtmlSafe.md
workmanw: @rwjblue Excellent. Thank you for calling this out. CC: @stefanpenner 

Would you guys like to see the internal checks, `str && typeof str.toHTML === 'function'`, be replaced with the newly added `isHtmlSafe`? Or would you prefer see that work take place outside of the scope of this RFC? Possibly even at a later date.

workmanw:
@rwjblue Yea, absolutely.

text/0000-isHtmlSafe.md
rwjblue: @workmanw - *How* `Ember.String.isHtmlSafe` works internally isn't a huge part of this RFC.  If we can move people to using it, that allows us to (in the future) change how we detect `SafeString`'s in the future, but changing that (IMHO) is out of scope of this proposal. 

> Would you guys like to see the internal checks be replaced with the newly added isHtmlSafe?

In the long run yes,  but it is unlikely to start that way...

stefanpenner:
I would be fine with isHtmlSafe being a function that returns true. It is likely more ergonomic then the typeof check.

alternatively and likely a better choice would be to have SafeString.isHtmlSafe(any) that way the user doesn't need any guards or anything, also it decouples the implementation of branding from the identification of the branding. Which seems quite future proof

rwjblue:
> I would be fine with isHtmlSafe being a function that returns true. It is likely more ergonomic then the typeof check.

Yes, that is what is being proposed here.

> alternatively and likely a better choice would be to have SafeString.isHtmlSafe(any) that way the user doesn't need any guards or anything

How would this work?  The proposal here suggests adding `Ember.String.isHtmlSafe(<any>)` that returns `true` / `false`.  We would ultimately remove `Ember.Handlebars.SafeString` (adding a deprecation to `Ember.Handlebars.SafeString` is what drummed up the need for a public API to detect if a given object is an HTML safe string).

> it decouples the implementation of branding from the identification of the branding

Hmm, the go forward path for operating with HTML safe strings is two methods on the same object:

* `Ember.String.htmlSafe(someString)` -- This generates an HTML safe string.
* `Ember.String.isHtmlSafe(<any>)` -- This detects if a given thing is a safe string (and is what this RFC is proposing).

This seems exactly like you are describing, what am I missing?

stefanpenner:
@rwjblue the implementation

rather then isHtmlSafe doing the instanceof check, it checks the branding.

rwjblue:
@stefanpenner - Yep, I commented similarly inline [here](https://github.com/emberjs/rfcs/pull/139#discussion_r60073914) (we should definitely not be doing `instanceof` to check this).  IMO, the implementation of *how* to determine if a given thing is an HTML safe string is not part of this RFC (though obviously it will need to be done at implementation time), and we should definitely be doing the `function` check vs `instanceof`.

stefanpenner:
@rwjblue sounds like we are in agreement, my mobile client doesn't do a good job of letting me do line comments.

rwjblue:
@stefanpenner - Gotcha, sorry for talking past each other there!  Yes, I believe we are both on the same page.  This RFC has been updated to mention the `typeof obj === 'function'` check (instead of `instanceof`), and now talks about the deprecation path.  I think it is ready for a re-read...

workmanw:
Updated the RFC with the proposed implementation change. Also put together a polyfill that uses the "branding" technique that Stefan described. [ember-string-ishtmlsafe-polyfill](https://github.com/workmanw/ember-string-ishtmlsafe-polyfill).

mixonic:
@workmanw thanks for this, and sorry this has hung here for so long. I'll push for some approval this weekend or this Friday and hopefully we can get it rolling :-D

workmanw:
Thanks! Happy to submit a PR for implementation given how trivial this is it shouldn't take much time at all.

rwjblue:
Discussed during core team F2F, and we believe this is good for implementation.

@workmanw - Would you mind working on implementation in Ember and creation of a polyfill once the Ember implementation PR is landed?

workmanw:
@rwjblue Happy to work on the implementation. A polyfill already exists :). https://github.com/workmanw/ember-string-ishtmlsafe-polyfill . But I will improve this polyfill once the implementation lands.

rwjblue:
👍

PR #141, opened 2016-04-26
Adds lazy-load-overview

MiguelMadero:
- [ ] Create [asset-service](https://github.com/MiguelMadero/ember-asset-service)
- [ ] Create asset-service RFC
- [ ] Remove low-level details to keep this only as overview and link to other RFC's or other work for additional info. 
- [x] Add info about LOG_RESOLVER flag to the list of unresolved questions. It seems like all the route related objects are eagerly resolved if the flag is on.

cc: @stefanpenner @chadhietela @rwjblue @ebryn @dgeb

### DIFF
--- a/active/0000-lazy-load-overview.md
+++ b/active/0000-lazy-load-overview.md
@@ -1,0 +1,185 @@
+- Start Date: 2016-04-13
+- RFC PR: (leave this empty)
+- ember-cli Issue: (leave this empty)
+
+# Summary
+
+Provide an overview of the parts required to enable different lazy-loading scenarios for Ember Applications.
+
+# Motivation
+
+Lazy loading helps us optimize boot time, by minimizing the amount of code that is loaded and evaluated during boot.
+This allows apps to grow horizontally, support different user roles or take more dependencies without
+necessarily increasing the app's payload or the costs associated with it.
+
+
+# Detailed design
+
+Lazy loading is complex and applies to multiple scenarios. In order to simplify its implementation I'm breaking it
+in different, independent parts, all of those could be delivered incrementally and still provide value. We need to
+build different assets, provide a standard way to load and track assets and support hooks to lazy load assets from routes
+and components. It is also important to consider asset dependencies.
+
+## Building
+
+We need to change the way we build our assets. By default Ember CLI bundles everything in two app and vendor assets for
+CSS and JS. In order to be able to lazy-load, we need to allow for customizations to how we build. This is partially
+possible today, but further enhancements are required.
+
+### Building multiple CSS:
+
+Today we have two ways of building multiple CSS assets. For vendor static assets (from `vendor/` or `bower_components`)
+we can import them to a different `outputFile` as specified on
+[RFC#28](https://github.com/ember-cli/rfcs/pull/28) and supported since ember-cli 2.4.
+
+```
+app.import(app.bowerDirectory + '/bootstrap/dist/css/bootstrap.css', {outputFile: 'vendor2.css'});
+```
+
+For files that need a pre-processor we can use `outputPaths`. In the example below, `styles/deferred-styles.scss` is
+simply a file with references to other CSS or SASS files. This works for `vendor` and `app` styles.
+
+```
+let app = new EmberApp(defaults, {
+  outputPaths: {
+    app: {
+      css: {
+        'app': '/assets/css/app.css',
+        'deferred-styles': '/assets/css/deferred-styles.css',
+      }
+    }
+  }
+});
+```
+
+### Building multiple vendor JS assets from static files (e.g. bower/vendor):
+
+For vendor static assets (from `vendor/` or `bower_components`) we can import them to a different `outputFile`
+as specified on [RFC#28](https://github.com/ember-cli/rfcs/pull/28) and supported since ember-cli 2.4.
+
+
+```
+app.import('vendor/dependency-1.js', { outputFile: 'assets/alternate-vendor.js'});
+```
+
+### Building multiple vendor JS assets from addons
+
+There is an [RFC](https://github.com/ember-cli/rfcs/pull/52) for this, but the concept is similar to the above, but
+extended for addons.
+
+### Building multiple JS assets for app code
+
+This isn't supported out of the box today, but it's possible to do. There are two example,
+[ember-cli-bundle-loader](https://github.com/MiguelMadero/ember-cli-bundle-loader) and
+[ember-lazy-loader](https://github.com/duizendnegen/ember-cli-lazy-load). The first one relies on a different file structure 
+and uses multiple ember-apps to build each asset. The latter relies on configuration to specify which app files will go to each
+asset.
+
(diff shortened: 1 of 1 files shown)


stefanpenner:
this seems like an ember-cli roc?

active/0000-lazy-load-overview.md
```diff
@@ -0,0 +1,185 @@
...
+build different assets, provide a standard way to load and track assets and support hooks to lazy load assets from routes
+and components. It is also important to consider asset dependencies.
+
+## Building
+
+We need to change the way we build our assets. By default Ember CLI bundles everything in two app and vendor assets for
+CSS and JS. In order to be able to lazy-load, we need to allow for customizations to how we build. This is partially
+possible today, but further enhancements are required.
+
+### Building multiple CSS:
```
stefanpenner: this section describes static assets

active/0000-lazy-load-overview.md
```diff
@@ -0,0 +1,185 @@
...
+asset.
+
+This RFC is limited to provide an overview of what is required, so this topic requires a bit more discusion on a
+separate RFC. TODO: create the separate RFC. Engines can provide a nice default, where we can have a convention of
+one asset per engine when lazyLoading is enabled, which can be overriden to allow for one or more engines per file.
+
+## Loading assets
+
+[ember-cli-bundle-loader](https://github.com/MiguelMadero/ember-cli-bundle-loader) provides a
+[service](https://github.com/MiguelMadero/ember-cli-bundle-loader/blob/master/addon/services/lazy-loader.js)
```
stefanpenner: i like the asset loading service

stefanpenner:
This RFC seems to go a bit deep on brand new ideas, I think it should either focus entirely on 1 of those ideas (the asset-service) or it should stick to entirely high level constructs and how they interact.

active/0000-lazy-load-overview.md
```diff
@@ -0,0 +1,185 @@
...
+let app = new EmberApp(defaults, {});
+app.import('bower_components/hands-on-table/dist/hands-on-table.js', {outputFile: 'hands-on-table.js');
+```
+
+
+## Lazy Loading from routes
+
+For app code, we likely want an integration with the router, which can rely on this service to do the actual loading.
+
+When splitting the app's JS and CSS into multiple assets related to different sectionsof an application, we want to load the code dynamically. This today poses additional challenges because the handler (route) needs to exist to initiate a
```
stefanpenner: a separate RFC for async route handlers seems appropriate.

cc @dgeb / @rwjblue i think you guys have WIP or atleast ideas here.

active/0000-lazy-load-overview.md
```diff
@@ -0,0 +1,185 @@
...
+outside of the scope of this RFC. For the purpose of this overview we simply assume that we have a way to map between routeNames and assetNames.
+
+## Engines
+
+Engines can simply rely on all of the parts above and provide nice conventions.
+
+## Dependencies and concatenation
+
+Sometimes an asset has a dependency on another asset. For example,
+[ember-cli-bundle-loader](https://github.com/MiguelMadero/ember-cli-bundle-loader/blob/master/tests/dummy/config/bundles.js)
```
stefanpenner: Our bundles will most likely need to be engines, I don't know if we can offer a safe/reliable solution with smaller non isolated slabs of user code. Race conditions and zalgo bugs will be hard mitigate without, although I would be open to other ideas.

MiguelMadero: I'll update the RFC to make that clear and remove the references to that project. 

What I'm referring to here is concatenating two assets. While it might make sense to logically think of, for example, insurance and  commute as two separate "engines", we might want to concatenate them into a "benefits bundle" if we know that they will always be used together or they have dependencies. Do you see any risk in doing this?

active/0000-lazy-load-overview.md
MiguelMadero: @trentmwillis has been working on this. For what I understand, he's making `getHandler` function promise-aware.

active/0000-lazy-load-overview.md
MiguelMadero: I'll look into it. 

>  it requires tight coupling to the build-pipeline 

I think it makes more sense to decouple them, but let me think it through a bit more. There are many ways in which we can build assets today from CLI, engines will add one more and we have cases where we might get assets from CDNs or generated by a different asset-pipeline. I think we need a manifest (for all CLI assets, similar to what we get from AssetRev) an extension point, so users can extend it for external assets, but the service can focus on loading JS/CSS in a reliable promise-aware way that can be used by components and the router. 

- [ ] Extract AssetService from bundle-loader and write RFC.

active/0000-lazy-load-overview.md
trentmwillis: That's more or less the gist of it. If you return a `Promise` from `Router#getHandler`, any `Transition` should wait until it resolves with a handler. I picked up the work from @dgeb and with guidance from @rwjblue. Didn't see much reason for an RFC since this change should be totally transparent to Ember/Ember-CLI consumers.

MiguelMadero:
> this seems like an ember-cli rfc?

@stefanpenner I can move it there, but I thought that it was a bit broader, specially considering the integration points in router.js and ember's router. 

> This RFC seems to go a bit deep on brand new ideas, I think it should either focus entirely on 1 of those new ideas (like the asset-service) or it should stick to the high-level narrative between the solutions.

I'll have another pass and keep it high-level and create separate ones as needed. There's already one for [addon's outputFiles](https://github.com/ember-cli/rfcs/pull/52) and I can create another one for the AssetService.

active/0000-lazy-load-overview.md
```diff
@@ -0,0 +1,187 @@
...
+# Alternatives
+
+This section needs further discussion. 
+
+TODO: fill this in
+
+# Unresolved questions
+
+* Ember Inspector breaks the apps the lazy-load routes because it eagerly resolves them, which results in a default route being generated, taking precedence over the lazy loaded one. We can fix Ember Inspector and even provide information about assets related to each route once we generate the route info at compile time. However, this will require users taking a new version of the extension. Not sure if we can force an update to avoid breakign the apps or if we can avoid that behavior on the app itself. 
+
```
MiguelMadero: [ ] Add info about LOG_RESOLVER flag to the list of `unresolved questions`. It seems like all the route related objects are eagerly resolved if the flag is on.

rwjblue:
@MiguelMadero - Was this superseded by https://github.com/emberjs/rfcs/pull/158 and https://github.com/emberjs/rfcs/pull/153?

MiguelMadero:
Hi @rwjblue sorry for the late response. Yes, this was superseded by those two. There is still useful info about other use cases that are still not addressed like build multiple CSS assets, building multiple vendor-JS assets and lazy-loading from components. This was originally intended as an overview, but at this point, it probably makes more sense to close this and open others if/as needed.

Issue #125, opened 2016-03-08
Can we make `lookupFactory` public on `ContainerProxyMixin`?


workmanw:
With [ember-data-model-fragments](https://github.com/lytics/ember-data-model-fragments/blob/557f4a288a6966fb4e8df48a80871ae11a17eea0/packages/model-fragments/lib/fragments/ext.js#L319-L327) we have to rely on the private variation of `_lookupFactory` on the `owner` to lookup fragment classes, see: [ember-data-model-fragments/.../ext.js](https://github.com/lytics/ember-data-model-fragments/blob/557f4a288a6966fb4e8df48a80871ae11a17eea0/packages/model-fragments/lib/fragments/ext.js#L319-L327).

This was part of the container API originally. And I know that was private and also we don't want to encourage users to be instantiating their own instances, but this is something we could really use for addons.

*Originally filed as https://github.com/emberjs/ember.js/issues/13065*

fivetanley:
I tend to agree, it seems like indirection to add a lot of `controllerFor` type hooks when that just delegates to the container anyway.

rwjblue:
The main reason that we didn't make `lookupFactory` public along with `lookup` was that we intend to change the way things work internally with regards to factory injections and while we realize that we will still have to support folks using the current `instance._lookupFactory`  and `container.lookupFactory` we wanted to provide a better public API going forward.

The goal of that refactor is to remove the current requirement that we `.extend` the factory being looked up just to perform injections. Instead, we would like to make the system return an object that itself has access to the original factory (non-extended) and all the hash arguments passed to its `.extend` so that we can merge those hash args later when we actually call `.create`.  (FWIW, this is roughly similar to what we do with contextual components).

Some rough pseudo code:

```js
class FactoryWrapper {
  constructor(Factory) {
    this.Factory = Factory;
    this._extendValues = [];
  }

  extend() {
    this._extendValues.push(...arguments);
  }

  create(options) {
    return Factory.create(...this._extendValues, options);
  }
}

// ...snip...

  factoryFor(name) {
    let Factory = normalResolverStuffHere(name);
    let wrapper = new FactoryWrapper(Factory);

    wrapper.extend(injectionsFor(name));
    return wrapper;
  }
```

Obviously, there are a number of things missing in my example above, but that should help understand why things are the way they are today...

bcardarella:
[...]

PR #149, opened 2016-06-23
Introduce "final comment period" into process description

dgeb:
This PR introduces the concept of a "final comment period" into the RFC process. This period is intended to focus the community's attention for a final round of consideration and feedback before an RFC is merged.

This concept was proposed and agreed upon by the core team at the June F2F meeting.

PR #150, opened 2016-07-01
`factoryFor` as a public API replacing `_lookupFactory`

mixonic:
[Rendered](https://github.com/mixonic/rfcs/blob/factoryFor/text/0000-factoryFor.md)

Based on [conversation from the F2F](https://github.com/emberjs/core-notes/blob/master/ember.js/2016-06/june-11-12_F2F.md#removing-double-extend) and with preliminary review and modifications by @stefanpenner, @rwjblue.

Closes https://github.com/emberjs/rfcs/issues/125

### DIFF
--- a/text/0000-factoryFor.md
+++ b/text/0000-factoryFor.md
@@ -12,7 +12,7 @@ public API to support use cases long-served by a private API, a new API of
 
 Ember's dependency injection container has long supported fetching a factory
 that will be created with any injections present. Using the private API that
-provided this support would allow an instance of the factory to be created
+provided this support allows an instance of the factory to be created
 with initial values passed via `create`. For example:
 
 ```js
@@ -102,16 +102,28 @@ better performance characteristics than `_lookupFactory`.
 
 # Detailed design
 
-This feature will be added in three steps.
+Throughout this document I reference Ember 2.12 as it is the next LTS at writing. This
+proposal may ship for 2.12-LTS or be bumped to the next LTS.
 
-1. Introduce `ApplicationInstance#factoryFor` in Ember 2.8
-2. Deprecate `ApplicationInstance#_lookupFactory` in Ember 2.8, remove in 2.9
-3. Release a polyfill addon for this feature
+This feature will be added in these steps.
 
-#### Step #1: Introduce `ApplicationInstance#factoryFor` in Ember 2.8
+1. In Ember introduce a `ApplicationInstance#factoryFor` based on
+   `_lookupFactory`. It should be documented that certain behaviors
+   inherent to "double extend" are not supported. In development builds
+   and supporting browsers, wrap return values in a Proxy. The proxy should
+   throw an error when any property besides `create` or `class` is accessed.
+2. In the same release add a deprecation message to usage of `_lookupFactory`.
+   As this API is intimate it must be maintained through at least one LTS
+   release (2.12 at this writing).
+3. In 2.13 drop `_lookupFactory` and migrate the `factoryFor` implementation to avoid
+   "double-extend" entirely.
+
+Additionally, a polyfill will be released for this feature supporting prior version of Ember.
+
+#### Design of `ApplicationInstance#factoryFor`
 
 A new API will be introduced. This API will return both the original base
-class registed into or resolved by the container, and will also return a function
+class registered into or resolved by the container, and will also return a function
 to generate a dependency-injected instance. For example:
 
 ```js
@@ -188,15 +200,13 @@ the following:
 +--------------------+ +--------------------+
 ```
 
-This means, between test runs 2 instances of `model:foo` will have a common
+Between test runs 2 instances of `model:foo` will have a common
 shared ancestor the grandparent `Class[model/Foo]`.
 
 This implementation of `factoryFor` proposes to remove the intermediate
 subclass and instead have a generic
 factory object which holds the injections and allows for injected instances
 to be created. The resulting object graph would look something like this:
-(between test runs the 2 instances of `model:foo` will have a shared ancestor
-(parent) `Class[model/Foo]`.
 
 ```
                   Proposed:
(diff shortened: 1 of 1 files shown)


krisselden:
Let's please use constructor instead and say it is optional, I think it is ok to register a factory to create an object literal or use Object.create().

stefanpenner:
> Let's please use constructor instead and say it is optional, I think it is ok to register a factory to create an object literal or use Object.create().

constructor may have meaning in that position, which is why class was choosen.

mixonic:
@krisselden regarding registration- You're looking for language that it is acceptable to call `owner.register('foo:baz', {create(opts) { return whatever }})`, and that would work with both `lookup` and `factoryFor`? Anything with a `create` method is considered a valid factory?

rwjblue:
@mixonic - That protocol (having a `.create` and possibly an `.extend` method on an object) generally works today, and I think it should be documented (though I'm unsure that documentation effort is specifically related to this RFC).

mixonic:
Seem closely enough related to fold in. Today `extend` is required for DI to be supported, and it is indeed a meaningful change that only `create` will be required when this lands.

martndemus:
I think the how we teach this section should also mention a migration guide for most of the cases that will break.

@rwjblue's example implementation in #125 also seems like it could be added to the RFC. His comment tells the story much better than this 280 line RFC.

Could you also add some reduced examples to the drawbacks section? The links towards projects do not always paint the picture of what is meant very clearly.

mixonic:
@martndemus a migration guide draft seems appropriate here.

> His comment tells the story much better than this 280 line RFC.

😭 which you've asked me to add further details to :trollface: I'll add a link to his comment somewhere relevant.

> Could you also add some reduced examples to the drawbacks section?

There are really not any good use-cases where you need this behavior, so any reduced example would also be pretty contrived. I will attempt to explain what the examples I've linked to are.

martndemus:
The more ELI5 this RFC is, the more easier it is to crowdsource docs/guides updates. 😄

mixonic:
I've pushed a round of edits here, much improved. Please give it another look!

text/0000-factoryFor.md
```diff
@@ -0,0 +1,452 @@
...
+Any state stored on the constructor would of course leak between the tests.
+
+An example implementation of `factoryFor` can be reviewed [on this GitHub
+comment](https://github.com/emberjs/rfcs/issues/125#issuecomment-193827658).
+
+##### Implications for `owner.register`
+
+Currently, factories registered into Ember's DI system are required to
+provide an `extend` method. Removing support for extend-based DI in `_lookupFactory`
+will permit factories without `extend` to be registered. Instead factories
```
stefanpenner: I agree with this change, but it is likely the second "breaking change" in this RFC. Is this a reasonable concern, or is it ok, and if so how do we migrate forwards + minimize any potential fallout.

stefanpenner:
I am slightly confused, what part of the rollout do we actually diverge the inheritance structure to the new simpler form?

mixonic:
@stefanpenner this proposal is fairly aggressive in that it suggests removal of the `_lookupFactory` API in Ember 2.9. See: https://github.com/mixonic/rfcs/blob/eccbd06ef6dfd4b50d1c8d4bb45ad58d110334f4/text/0000-factoryFor.md#step-2-deprecate-applicationinstance_lookupfactory-in-ember-28-remove-in-29 Ember 2.8-LTS would provide both APIs, giving LTS supporting addons/apps until Ember 2.12-LTS to make the migration.

At the same time, this RFC also proposes releasing a polyfill for `factoryFor` going back to at least Ember 2.4. An application (or addon) migration strategy would be to migrate to the new API via the polyfill, then remove that polyfill when support for Ember pre-2.9 is no longer a requirement.

stefanpenner:
@mixonic this RFC describe the API, but not when we deprecate the internal behavior.

stefanpenner:
In the meeting we also discussed wrapping `.class` in a proxy for development, a proxy which:

* disallows all sets & obvious mutations.
* disallows `reopen` `reopenClass`
* allows all reads
* allows all other operations

stefanpenner:
@mixonic  / @rwjblue I (well actually @krisselden) have found a legit in-the-wild example of how we will end up breaking real code by no longer calling `extend`: https://github.com/emberjs/ember.js/blob/master/packages/ember-glimmer/lib/views/outlet.js#L49

One work-around would be to only not call `extend` if it is the `extend` we know to be the one on an ember object...

krisselden:
I would not be surprised if there is not addons that use extend to return a custom factory

stefanpenner:
> I would not be surprised if there is not addons that use extend to return a custom factory

Yes, i believe we can detect those cases and do the old thing.

sukima:
To help this RFC along, our team is desperately trying to migrate from 1.13. One of the hardest blockers of this is our app design relied heavily on JS only `imports` and window/Ember globals. Newer implementations of these features use services now which are unavailable without a container. To help this migration we would love to get our dependency injections set for our `Ember.Object` based models so that when we do eventually convert to ember-data the services we use would already be available.

The downsides are:

* The docs show `_lookupFactory` as **private**. This makes our team members nervous.
* The method starts with an underscore which makes our team doubly nervous.

Although `_lookupFactory` is a de facto standard and used among addon authors in the wild it is still an up hill battle in some teams to convince them that a private method that starts with an underscore is technically *safe*.

This RFC would offer a sanctioned way to create objects via a factory with dependency injection where the usual `Ember.Object.create` would not otherwise. The examples in this RFC are an exact match for what we need in our app to move it forward. For now we will use `_lookupFactory` but would really love to see this RFC move into a recognized implementation and a polyfill to get older apps there.

ember-data is great that it can do this stuff automagically but it sure is nice to be able to do these tricks on our own. As an addon author it opens up a large opportunity for some very powerful features. I hope this testimony of one possible use case in a production app helps adds some context to this RFC.

sukima:
For what it's worth here is how we worked with this (with out the ember-getowner-polyfill):

```javascript
import Ember from 'ember';

const { isPresent, Service, deprecate } = Ember;

function getOwner(ctx) {
  deprecate(
    'Ember.getOwner() is available now; please cleanup references to this.container',
    isPresent(Ember.getOwner),
    {id: 'container-lookup'}
  );
  return isPresent(Ember.getOwner) ? Ember.getOwner(ctx) : ctx.container;
}

export default Service.extend({
  createRecord(modelName, payload) {
    const owner = getOwner(this);
    const lookupFactory = owner._lookupFactory || owner.lookupFactory;
    const ModelFactory = lookupFactory.call(owner, `model:${modelName}`);
    return ModelFactory.create(payload);
  }
});
```

mixonic:
This RFC is basically in FCP pending the process item of a tweet from https://twitter.com/emberjs and tag. If you've any remaining concerns you believe are missed here, please let me know!

pixelhandler:
@mixonic I use  `_lookupFactory` in my data persistence addon. Today I read over the RFC and am very pleased with the thought and depth put into this RFC.

As an addon developer I think it will be straight forward to make the upgrade to `factoryFor` and at the same time continue support `_lookupFactory` as well. (the build for the addon tests v1.13 though canary).

👍

chadhietala:
I would like to propose we don't inject the deprecated container into the instances returned from `factoryFor` and we mitigate that to the deprecated `_lookupFactory` path.

mixonic:
@locks thats for the tagging!

@chadhietala Can you say more? There is no "deprecated container" described here.

dgeb:
*Note: To make the Final Comment Period official, we need to announce it here and on Twitter, so we're restarting the period now.*

---

This RFC is now entering a ["final comment period"](https://github.com/emberjs/rfcs/pull/149). This period is intended to focus the community's attention for a final round of consideration and feedback before an RFC is merged.

If no significant blockers are raised by **October 7**, we plan to merge this RFC and make plans for its implementation. :shipit:

rwjblue:
Looks like no additional comments came through during the FCP process.  Lets do it!!!

pixelhandler:
@rwjblue should the "Final Comment Period" label be removed?

dgeb:
@pixelhandler good call - done.

chadhietala:
See https://github.com/emberjs/ember.js/pull/14360 for the WIP implementation.

PR #153, opened 2016-07-17
RFC: Asset Manifest

trentmwillis:
[Rendered](https://github.com/trentmwillis/rfcs/blob/asset-manifest/text/0000-asset-manifest.md).

Additional context:
- Original motivation from Engines RFC: [Lazy loading manifests](https://github.com/emberjs/rfcs/blob/master/text/0010-engines.md#lazy-loading-manifests)
- [Lazy Loading Engines Attack Plan](https://github.com/dgeb/ember-engines/issues/154)


dgeb:
Since we are only concerned with the subset of URIs that are URLs, we should consider using `url` or perhaps `src` instead of `uri` to specify the location of each bundle.

trentmwillis:
@dgeb we could potentially switch to `url`, the primary reason I went with `uri` is that depending on the nature of your resource delivery, you could just have an ID or some path that is not necessarily a resolved URL.

I'd also like to avoid `src` as it feels specific to JS assets.

dgeb:
> depending on the nature of your resource delivery, you could just have an ID or some path that is not necessarily a resolved URL.

Good point - let's stick with `uri` 👍

nathanhammond:
Capturing a note from a Slack conversation I believe originally brought up by @dgeb:
The asset manifest in this case would be a product generated at the very tail end of the process (sometimes even after deploy-specific post-processing) which will then need to be inserted into `index.html` which will already have a serialized config meta tag.

This isn't particularly elegant and is something we should think more about.

nathanhammond:
Note from @rwjblue: should address the fact that the `assetmanifest.json` is not a complete enumeration of all engine assets, it excludes: 
- Fill...
- ...in...
- ...list.

trentmwillis:
Updated. Changes discuss generating an actual file (`asset-manifest.json`), handling of on-demand static assets (e.g., anything in `public`), and overrides/customization.

nathanhammond:
Changes look good. Do we want to mention asset-rev rewriting of the `asset-manifest.json` in this or leave that as unspecified behavior per this RFC?

trentmwillis:
@nathanhammond it mentions accommodating asset-rev twice, thought doesn't go into detail on what needs to be done as I feel that is outside the scope of this RFC

rwjblue:
Where are we at with this?  Discussion seems to be mostly quieted down.

@dgeb / @mamadero / @nathanhammond / @trentmwillis - Are more changes needed here or are we ready to try to advance to FCP?

trentmwillis:
We have a working implementation in the [ember-asset-loader](https://github.com/trentmwillis/ember-asset-loader/) addon. Some changes have occurred in how we generate the manifest, so I'll need to go and see if that requires any updates to the RFC as it stands here. Will try to do that by end of week.

text/0000-asset-manifest.md
```diff
@@ -0,0 +1,181 @@
...
+
+Since we can access the package name of an Engine at runtime, regardless of aliasing, we should always load an Engine's assets according to its package name and not its mounted path. This will make it easier to avoid duplicate loading of assets and ensures we can support multiple inclusion of the same engine. Additionally, referencing bundles via a name rather than route path decouples asset loading from the routing behavior.
+
+Each bundle will have two primary fields: `assets` and `dependencies`.
+
+`assets` is an array of asset objects, which specify a `uri` and `type` for an asset to be loaded. The asset objects are sorted according to load order. This is important for assets such as `vendor.js` which need to be loaded before other assets like `app.js`.
```
MiguelMadero: > The asset objects are sorted according to load order

I didn't see anything about order on the [AssetLoader RFC](https://github.com/trentmwillis/rfcs/blob/asset-loader-service/text/0000-asset-loader-service.md). We had problems with this before. Not all browsers handle this the same and we had to explicitly set `async=false` to make sure that the scripts are really executed in the order they are added. 

I think we need to do it on https://github.com/trentmwillis/ember-asset-loader/blob/master/addon/loaders/js.js#L20

MiguelMadero: It's easier if I create a PR to follow-up the discussion there.

MiguelMadero:
Hi @trentmwillis and all, I've been quietly following this along. I didn't have much to add and mostly agreed with this. It's really similar to what we used already with some minor improvements. I'm in general pretty happy with this as it's. 

>  I'll need to go and see if that requires any updates

Does this still require an update? It looks ready for FCP.

text/0000-asset-manifest.md
MiguelMadero: Created https://github.com/trentmwillis/ember-asset-loader/pull/38

trentmwillis: I added some additional language about ordering and the relationship between the manifest ordering and behavior of the loader.

trentmwillis:
@rwjblue I've updated with some clarifications. At this point, I think we're ready to proceed to FCP if no objections are raised.

rwjblue:

tomdale:
I was at the core team meeting but I had some questions post-meeting. 😁 

1. It seems like there is some overlap between this and [dynamic imports/`import().then()`](https://github.com/tc39/proposal-dynamic-import). Is it possible to align this proposal with that proposal, instead of creating Ember-proprietary APIs?
2. I suspect Webpack and others are generating similar manifests (e.g. to support dynamic imports). Is this something we can align with?
3. FastBoot also generates a manifest file that the FastBoot server uses to know which files to load. Does it make sense to move that manifest out of something FastBoot-specific and into this?

blittle:
@joeldenning and I built a similar tool on top of System.js called [sofe](https://github.com/CanopyTax/sofe). We have really liked using the new dynamic imports syntax. Utilizing System.js's module registry makes sharing common dependencies easy as well. As far as webpack, we've had multiple discussions with @TheLarkInn on [Twitter](https://twitter.com/TheLarkInn/status/789968589419745280) about how webpack doesn't have plans to do this type of dynamic loading.

One major suggestion I'd have for your implementation is to build local overrides into your implementation. For example, in sofe I could load a module:

```js
import('my-app!sofe').then(...)
```

If there is a local storage override:
```js
localStorage.setItem('my-app:sofe', 'https://localhost:3333/my-app.js')
```
Then in the context of production, I can override the asset manifest location of a specific bundle to point to my localhost. This makes testing and development in production really easy. I can override the specific bundle that I care about in the exact context it is in prod!

rwjblue:

PR #154, opened 2016-07-19
RFC for router enhancements

cibernox:
Expand Ember's router with support for optional segments and constraints.

[Rendered](https://github.com/cibernox/rfcs/blob/router_enhancements/text/router-enhancements.md)

### DIFF
--- a/text/router-enhancements.md
+++ b/text/router-enhancements.md
@@ -1,0 +1,175 @@
+- Start Date: 2016-07-18
+- RFC PR: (leave this empty)
+- Ember Issue: (leave this empty)
+
+# Summary
+
+The features of Ember's router have been unchanged since pretty much the initial
+release.
+
+This RFC focuses in twow features:
+
+- Optional segments
+- Dynamic segment constraints
+
+These features can enable some usages that right now are cumbersome or require code duplication.
+
+# Motivation
+
+
+### Dynamic segments
+
+Consider an e-commerce or marketing site nuilt in Ember that wants to leverage server-side rendering
+with fastboot to be indexed in search engines.
+
+Search engines are smart enough to index information based on the language, but it's common practice
+to namespace content by language and/or country. Sometimes even the items displayed depeds on
+the region, but the page itself is identical.
+
+One real-world example of this is Apple's page.
+
+The page version for the UK is `https://www.apple.com/uk/mac`, while the spanish version is in
+`https://www.apple.com/es/mac`.
+
+So far, this route can be expressed in the current's version of the router as `/:regionCode/mac`.
+
+However, the US/International/Default version of the page is also available in `https://www.apple.com/mac`.
+The content and layout of the page is the same in the three routes and yet as of today, implementing
+this kind of pattern requires to duplicate the entire routes/controllers under two different
+places. The template could be reused by overriding the `render` hook in the route, but still this
+is awkward.
+
+Ideally the user should be able to define the initial segment as optional so the router maps all those
+urls to the same route giving.
+
+Regional/Locale is just one usage of optional segments that is particularly common but not the only.
+
+### Dynamic segments' constraints
+
+Another feature that can ease the development is determine if a pattern matches based on some
+condition that cannot be expressed solely with the microsyntax that the router provides.
+
+Consider a weather app that displays the forecast for tomorrow with temperatures, and the user gets
+to choose among a Celcius, Fahrenheit and Kelvin degrees, and that lives in the url.
+
+This can be encoded on this string: `/forecast/:scale`. That entry would match `/forecast/celsius`,
+`/forecast/fahrenheit` but also URLs like `/forecast/rankine` which is a unsupported scale.
+
+Currently there is two solution for this, either codify these three routes as static routes
+without dynamic segments, which would leave us with a similar duplication problem as the one
+described for optional segments, or guard agains invalid scales in the route and perform a redirect
+via `replaceWith` to the not-found page. However, taking lazy-loading engines into consideration,
+handling constrains manually in the routes requires the app to load and boot the engine in order to
+determine something that could be done much before and save all this expensive work.
+
+Backend fragments like Ruby on Rails get a DRY solution to this allowing to receive an executable
+piece of code that the dynamic segment has to validate in order of that route to match.
+
+In the previous example the constraint would be that the scale is one of our whitelisted values, and
+any other would cause the route to miss.
+
+
+# Detailed design
+
+
+### Optional segments.
+
+Mimicking rails microsintax, I propose optional segments to be enclosed between parenthesis. The
+example of the Apple store would be encoded
+
+```
+Router.map(function() {
+  this.route('product', { path: '(/:region)/:productName' });
+});
+
+
+router.serialize('/mac'); // => routeName='products', params: { productName: 'mac' }
+router.serialize('/uk/mac'); // => routeName='products', params: { productName: 'mac', region: 'uk' }
+```
+
+
+### Dynamic segment's constrains
+
+Also stealing Ruby on Rails' syntax, the user can provide a constraint using the
+config object passed as second argument.
+
+
+```js
+const allowedScales = ['celsius', 'fahrenheit', 'kelvin'];
+
+function isAllowedScale(scaleName) {
+  return allowedScales.includes(scaleName);
+}
+
+Router.map(function() {
+  this.route('forecast', { path: 'forecast/:scale', constraints: { scale: isAllowedScale } });
+});
+
+
+router.serialize('/forecast/kelvin'); // => routeName='forecast', params: { scale: 'kelvin' }
+router.serialize('/forecast/newton'); // => not recognized route.
+```
+
+Although the generic case of receiving a function that returns a boolean is the most generic one,
+it might worth to add a RegExp support as a convenience.
+
+```js
+Router.map(function() {
+  this.route('forecast', { path: 'forecast/:scale', constraints: { scale: /(celsius|fahrenheit|kelvin)/ } });
+
+  # Equivalent to:
+  # this.route('forecast', {
+  #   path: 'forecast/:scale',
+  #   constraints: {
+  #     scale: function(scale) {
+  #       return /(celsius|fahrenheit|kelvin)/.test(scale);
+  #     }
+  #   }
+  # });
+});
+```
+
+# How We Teach This
+
+Expand the guides on the routing topic to cover this use cases.
+
+# Drawbacks
+
+It adds some extra complexity to the routes for a feature that we've survived without for years.
+
+Particularly the dynamic segment constraints, without considering the case of lazy-loaded engines,
+can be emulated without too much effort by manually by checking the validity of the params in the
+routes, and adds some vector of complexity since it's not clear which approach to use as they
+admitely overlap.
+
+# Alternatives
+
+What other designs have been considered? What is the impact of not doing this?
+
+# Unresolved questions
+
+- Do optional segments have a default value?
+- What arguments should the constraint function receive apart from the dynamic segment itself?
+- In routes with more than one dynamic segment, can constraints perform checkings on all segments
+at once? Per example:
+
+```js
+const allowedScales = ['celsius', 'fahrenheit', 'kelvin'];
+
+function isAllowedScale(scaleName) {
+  return allowedScales.includes(scaleName);
+}
+
+Router.map(function() {
+  this.route('forecast', {
+    path: 'activities/from/:from/to/:to',
+    constraints: function(from, to) {
+      return moment(from).isBefore(moment(to));
+    }
+  });
+});
+
+
+router.serialize('/activitis/from/2016-04-11/to/2016-05-20'); // => matches
+router.serialize('/activitis/from/2016-06-11/to/2016-05-20'); // => fails to match
+```
--- a/text/router-enhancements.md
+++ b/text/router-enhancements.md
@@ -102,8 +102,10 @@ Router.map(function() {
   this.route('sample1', { path: '/foo/bar' });
   this.route('sample2', { path: '/foo/(:param)' });
   this.route('sample3', { path: '/foo/*path' });
-  this.route('sample4', { path: '/foo' } ,function() {
-    this.route('opt', { path: '/(:param)' });
+
+  this.route('sample4' { path: '/bar/(:optional)' })
+  this.route('sample5', { path: '/bar' } ,function() {
+    this.route('opt', { path: '/(:optional)' });
   });
 });
 ```
@@ -119,10 +121,12 @@ The resolution order stays as it is currently. That is
 Examples:
 
 ```
-/foo/bar            => 'sample1' (two segments, both static)
-/foo/dynamic        => 'sample2' (two segments, one static, one dynamic)
-/foo/something/else => 'sample3' (two segments, one static, one glob)
-/foo                => 'sample4' (one static segment, but wins over sample 2 because has more handlers)
+/foo/bar            => 'sample1'      (two segments, both static)
+/foo/dynamic        => 'sample2'      (two segments, one static, one dynamic)
+/foo/something/else => 'sample3'      (two segments, one static, one glob)
+/foo                => 'sample2'      (one static segment)
+/bar/baz            => 'sample5.opt'  (two segments, one static, one dynamic, but wins over `sample4` because has more handlers)
+/bar                => 'sample4'      (one static segment, and wins over `sample5` because it's defined before)
 ```
 
 ### Dynamic Segment Constraints
--- a/text/router-enhancements.md
+++ b/text/router-enhancements.md
@@ -17,7 +17,7 @@ These features can enable some usages that right now are cumbersome or require c
 # Motivation
 
 
-### Dynamic segments
+### Optional segments
 
 Consider an e-commerce or marketing site built in Ember that wants to leverage server-side rendering
 with Fastboot to be indexed in search engines.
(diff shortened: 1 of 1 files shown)


text/router-enhancements.md
```diff
@@ -0,0 +1,175 @@
...
+routes, and adds some vector of complexity since it's not clear which approach to use as they
+admittedly overlap.
+
+# Alternatives
+
+What other designs have been considered? What is the impact of not doing this?
+
+# Unresolved questions
+
+- Do optional segments have a default value?
```
nathanhammond: No, they do not. They're literally skipped in the route parsing pass. The receiving handler should be capable of surviving `params.optionalDynamicSegmentValue === undefined`.

text/router-enhancements.md
```diff
@@ -0,0 +1,175 @@
...
+});
+```
+
+# How We Teach This
+
+Expand the guides on the routing topic to cover this use cases.
+
+# Drawbacks
+
+It adds some extra complexity to the routes for a feature that we've survived without for years.
```
nathanhammond: The complexity has been reduced tremendously by matching segment-at-a-time.

text/router-enhancements.md
```diff
@@ -0,0 +1,175 @@
...
+
+Particularly the dynamic segment constraints, without considering the case of lazy-loaded engines,
```
nathanhammond: The `routes.js` file from engines are hoisted into the host application immediately. This would allow the constraint to be consistently applied even without attempting the transition and triggering engine load.

text/router-enhancements.md
```diff
@@ -0,0 +1,175 @@
...
+
+
+### Optional segments.
+
+Mimicking rails micro syntax, I propose optional segments to be enclosed between parenthesis. The
+example of the Apple store would be encoded
+
+```
+Router.map(function() {
+  this.route('product', { path: '(/:region)/:productName' });
```
nathanhammond: The paren would be inside of the segment: `/(:region)/:productName`

text/router-enhancements.md
```diff
@@ -0,0 +1,175 @@
...
+
+```js
+const allowedScales = ['celsius', 'fahrenheit', 'kelvin'];
+
+function isAllowedScale(scaleName) {
+  return allowedScales.includes(scaleName);
+}
+
+Router.map(function() {
+  this.route('forecast', { path: 'forecast/:scale', constraints: { scale: isAllowedScale } });
```
nathanhammond: Should include an example of using a constraint on an optional dynamic segment.

text/router-enhancements.md
```diff
@@ -0,0 +1,175 @@
...
+piece of code that the dynamic segment has to validate in order of that route to match.
+
+In the previous example the constraint would be that the scale is one of our white-listed values, and
+any other would cause the route to miss.
+
+
+# Detailed design
+
+
+### Optional segments.
```
nathanhammond: Should mention that it should support all three primary segment types:
- Static
- Dynamic
- Glob

text/router-enhancements.md
cibernox: I mentioned that in the **motivation** section. What I wanted to state is that if it wasn't for lazy-loaded engines, I wouldn't be too upset if constrains are not implemented, but they can be valuable to avoid loading engines if the route won't match a check that could have been easily performed eagerly.

text/router-enhancements.md
```diff
@@ -0,0 +1,175 @@
...
+Expand the guides on the routing topic to cover this use cases.
+
+# Drawbacks
+
+It adds some extra complexity to the routes for a feature that we've survived without for years.
+
+Particularly the dynamic segment constraints, without considering the case of lazy-loaded engines,
+can be emulated without too much effort by manually by checking the validity of the params in the
+routes, and adds some vector of complexity since it's not clear which approach to use as they
+admittedly overlap.
```
nathanhammond: One throws `UnrecognizedURLError` and triggers the error route. The other has to manually handle the behavior. I prefer constraints for obviously bad data.

text/router-enhancements.md
```diff
@@ -0,0 +1,175 @@
...
+
+# Alternatives
+
+What other designs have been considered? What is the impact of not doing this?
+
+# Unresolved questions
+
+- Do optional segments have a default value?
+- What arguments should the constraint function receive apart from the dynamic segment itself?
```
nathanhammond: Things it could receive: the node in the trie representing the route, the handlers, the segment value, and the full set of parsed params. Since I want this to be primarily used for simplistic reduction I'd propose nothing more than the segment value itself.

text/router-enhancements.md
```diff
@@ -0,0 +1,175 @@
...
+- In routes with more than one dynamic segment, can constraints perform checks on all segments
+at once? Per example:
```
nathanhammond: That would require backtracking on the union set at the very end, but is possible.

text/router-enhancements.md
```diff
@@ -0,0 +1,175 @@
...
+
+This can be encoded on this string: `/forecast/:scale`. That entry would match `/forecast/celsius`,
+`/forecast/fahrenheit` but also URLs like `/forecast/rankine` which is an unsupported scale.
+
+Currently there is two solutions for this. Either to codify these three routes as static routes
+without dynamic segments, which would leave us with a similar duplication problem as the one
+described for optional segments, or guard against invalid scales in the route and perform a redirect
+via `replaceWith` to the not-found page. However, taking lazy-loading engines into consideration,
+handling constrains manually in the routes requires the app to load and boot the engine in order to
+determine something that could be done much before and save all this expensive work.
```
nathanhammond: 👍

nathanhammond:
Hey everybody! This isn't some greenfield proposal–I know exactly how to implement 99% of this and will have it working in an addon you can try in your applications in the next few days (which will also do build-time serialization as a side-effect). @cibernox has been really itching to have these features and wrote this RFC to get the discussion started.

text/router-enhancements.md
cibernox: In Rails the parenthesis enclose the slash too. 
I don't have a strong opinion on this and if it's technically simpler so be it, but I think that it reads better that way because if you delete everything between parents you don't end up with a double slash.

RuslanZavacky: having (/:region) `/` inside also helps if your segment is the last one. In cases when you don't want your URL to be accessible by two different routes, eg. `/my-cool-seo-url/` and `/my-cool-seo-url`. So you just specify, that additional segment only works, if there is a `/` in the end. Also, you might want to have the last segment as empty. And its possible to distinguish when you explicitly say, if there is `/` in the end, than segment should appear.

RuslanZavacky: And one more use-case is, that I actually want a lot :)
I have 3 languages -> `en`, `es`, `fr`
Default language is `en`. I should be able to have access to default route, without language prefix. Eg.

`(/:lang)/welcome`

Should be translated into:
`/welcome`
`/es/welcome`
`/fr/welcome`

cibernox: I think that the example I used is identical.

text/router-enhancements.md
cibernox: Added

text/router-enhancements.md
RuslanZavacky: Yeap, true :)

text/router-enhancements.md
nathanhammond: This consideration actually means that the constraint function should either be a pure function which has no access to outside scope or even just a regular expression. Otherwise somebody can import something into `router.js` and make a mess of things.

cibernox: Initially I'd consider that the constraint function should be pure. Eventually someone will want to have access to the session service to do back magic, but if that's ever allower I'd consider that a phase two.

nathanhammond:
@RuslanZavacky @cibernox Y'all's thoughts on parentheses "including the slash" come from the (reasonable!) assumption that `/` is a meaningful character inside of the Router DSL. It isn't. Try it out!

```javascript
Router.map(function() { 
  this.route('something', { path: '////something////'} );
});
```

```handlebars
{{#link-to 'something'}}Something{{/link-to}}
```

Output:
```html
<a href="/something">Something</a>
```

But, if you move the parentheses _outside_ of the segment definition you've now made it _tremendously_ more difficult to parse. I'm standing firm on this one: the `()` are a property of the segment and belong between the slashes.

RuslanZavacky:
@nathanhammond ah, so its not actually required to place it inside. If its easier to parse it, when its outside, I am quite okey with that, @cibernox thoughts? :)

nathanhammond:
Regarding constraint `function`s: having thought about it more, I'm pretty sure that I want to avoid the situation where the constraint function has references to all sorts of interesting things that we may change to support tremendous performances wins (router precompilation). If we allow a user to define a function it's in a really dangerous scope for abuse. (See what I did inside of [ember-route-alias](https://github.com/nathanhammond/ember-route-alias/blob/master/addon/initializers/route-alias.js). Hack and a half.)

I believe that for V1 we should only support regular expressions. Advanced users can *still* abuse that by throwing in an IIFE that returns a regular expression but we can throw an error at build time (via an AST analysis) to prevent that.

text/router-enhancements.md
```diff
@@ -0,0 +1,211 @@
...
+is awkward.
+
+Ideally the user should be able to define the initial segment as optional so the router maps all those
+URLs to the same route.
+
+Region/Locale is just one usage of optional segments that is particularly common but not the only.
+
+All segments can be optional, no matter globs, static or dynamic.
+
+### Dynamic segments' constraints
```
nathanhammond: This can also apply to glob segments.

text/router-enhancements.md
```diff
@@ -0,0 +1,211 @@
...
+  this.route('forecast', { path: 'forecast/:scale', constraints: { scale: isAllowedScale } });
+});
+
+
+router.serialize('/forecast/kelvin'); // => routeName='forecast', params: { scale: 'kelvin' }
+router.serialize('/forecast/newton'); // => not recognized route.
+```
+
+Although the generic case of receiving a function that returns a boolean is the most generic one,
+it might worth to add a RegExp support as a convenience.
```
nathanhammond: Having two invocation patterns puts a branch in a hot loop. I'd prefer only one.

text/router-enhancements.md
```diff
@@ -0,0 +1,211 @@
...
+});
+
+
+router.serialize('/forecast/celsius'); // => routeName='forecast', params: { scale: 'celsius' }
+router.serialize('/forecast/newton'); // => not recognized route.
+router.serialize('/forecast'); // => routeName='forecast', params: { }
+```
+
+On this example `forecast` route is invoked even if the scale is ommited, but if it's not ommited,
+the constraint ensures that it's among a white-list of supported scales.
```
nathanhammond: Further, the `scale` function is invoked only if there is a segment following `/forecast` to identify it could possibly match the optional dynamic path segment.

text/router-enhancements.md
```diff
@@ -0,0 +1,211 @@
...
+router.serialize('/activities/from/2016-04-11/to/2016-05-20'); // => matches to `activities`
+router.serialize('/activities/from/2016-06-11/to/2016-05-20'); // => fails to match
+```
+
+- How complex can constraints be? Just pure functions or can they have access to services? I think
+that for simplicity they should start as simple as possible, and they could grow to support more
+complex usages.
+- When an optional dynamic segments are combined with constraints on that segment, and the segment
+is not present, does the constraint function run with `undefined` as value or it doesn't run at all?
+I lean towards the second.
```
nathanhammond: It'll be invoked by the transition function for any segment which appears following a matching static segment. [It'd do a `filter` here.](https://github.com/nathanhammond/ember-route-recognizer/blob/a0ab4117d7f6596c107817806c90c6a0721bcdcb/addon/-private/nfa-transition.js#L32)

cibernox: So, to clarify, no segment, no invocation, right?

cibernox:
I'm ok with constrain constraints (pun intended) to regexes initially if that simplifies the implementation. Support for functions could be added in the future if people really wants it. I'll remove references to function constrains from the RFC.

cibernox:
About the parentheses wrapping the slash or not, I didn't suggested because I thought that slashes have a meaning in the routes, but for familiarity with rails. But if the parsing is significantly more complicated that way, let's make it simple.

text/router-enhancements.md
nathanhammond: Kinda weird:

Given these two paths:
```javascript
"/foo/:bar"
"/foo/thing"
"/foo/thing/else"
```

These are the results:
```javascript
deserialize('/foo'); // no constraint check.
deserialize('/foo/thing'); // constraint check.
deserialize('/foo/thing/else'); // constraint check is an implementation detail and may be optimized out.
```

In other words one should never rely upon side effects from a constraint check function (which we're not implementing at this time) as they're only guaranteed to be called if you end up in the route with the constraint check or have multiple dynamic segments at a particular segment count.

mmun:
This RFC doesn't describe how specificity works with optional segments, e.g.

```
/foo/bar
/foo/(:param)
/foo/*path
```

nathanhammond:
@mmun Underneath the hood in a naïve implementation it will actually just append that route subtree to two separate parent nodes in the radix NFA. In a more sophisticated implementation during capture it'll annotate which `parent` path to take in the reverse traversal during processing to prevent doubling subtree size.

Early implementation of the process behavior: https://github.com/nathanhammond/ember-route-recognizer/blob/master/addon/-private/route-recognizer.js#L38-L87

To answer your question, given this path map:
```
/foo/bar
/foo/(:param)
/foo/*path
```

Ranked in order of specificity when matching:
```
/foo/bar            => (two static)
/foo/dynamic        => (one static, one dynamic)
/foo/something/else => (one static, one glob)
/foo                => (one static)
```

If an optional segment is present it counts for the specificity. If it isn't matched it doesn't count.

text/router-enhancements.md
```diff
@@ -110,15 +113,16 @@ The resolution order stays as it is currently. That is
...
   than globs.
-- In case of tie, the first defined route wins
+- In case of tie in the number and type of segments, the route with more handlers wins.
+- If all the previous fail, the first defined route wins
 
 Examples:
 
 ```
 /foo/bar            => 'sample1' (two segments, both static)
 /foo/dynamic        => 'sample2' (two segments, one static, one dynamic)
```
nathanhammond: This would be sample4 because it has more handlers.

text/router-enhancements.md
```diff
@@ -110,15 +113,16 @@ The resolution order stays as it is currently. That is
...
 /foo/something/else => 'sample3' (two segments, one static, one glob)
-/foo                => 'sample2' (one static segment)
+/foo                => 'sample4' (one static segment, but wins over sample 2 because has more handlers)
```
nathanhammond: This would be `sample2` because that definition creates an accepting state for `/foo` and is defined before `sample4`.

text/router-enhancements.md
```diff
@@ -122,7 +124,8 @@ Examples:
 /foo/bar            => 'sample1' (two segments, both static)
 /foo/dynamic        => 'sample2' (two segments, one static, one dynamic)
 /foo/something/else => 'sample3' (two segments, one static, one glob)
-/foo                => 'sample4' (one static segment, but wins over sample 2 because has more handlers)
+/foo                => 'sample2' (one static segment)
+/bar                => 'sample5' (one static segment, but wins over `sample4` because has more handlers)
```
nathanhammond: This would still be `sample4`.

This definition:
```
this.route('sample4' { path: '/bar/(:optional)' });
```

Should be thought of as:
```
this.route('sample4' { path: '/bar' }) 
this.route('sample4' { path: '/bar/:optional' }) 
```

nathanhammond:
```
/bar/something       => 'sample5.opt'
```

...is a good inclusion and appears to be the example you're going for.

IAmJulianAcosta:
@nathanhammond Hi! I just have a small question.

Currently I'm using dynamic segments like this:

```
/foo/xyz
/bar/xyz
/foobar/xyz
```

my router is:

```js
this.route (
			'xyz',
			{
				path: ':param/xyz'
			}
		);
```

But I need param to be optional, so I can have a route `/xyz`. Did you think about cases like this one?

nathanhammond:
@IAmJulianAcosta Order inside of the specified path doesn't matter, so this will work just fine. The microsyntax would be `(:param)/xyz`.

IAmJulianAcosta:
@nathanhammond Thanks! I want to know also if the addon is ready to use or I'll have to wait a little bit for it.  Thanks!

jeffrey008:
Great to see this RFC. I have been using optional dynamic segments a lot in Rails, and I think we need one in Ember. 
@nathanhammond I checked out your addon `ember-route-alias`, is your addon capable of doing optional dynamic segments at the moment?
```
Router.map(function() {
  this.route('sample', { path: '/foo/(:param)' });
});
```

Hope this RFC can be merged ASAP. Thanks the great work!

simonihmig:
I have not seen a discussion about default values for optional segments here. Is that not under consideration?

Given the one example from the RFC wouldn't it be nice to do something like that:
```js
Router.map(function() {
  this.route('forecast', {
    path: 'forecast/(:scale)',
    constraints: { scale: /(celsius|fahrenheit|kelvin)/ },
    defaults: { scale: 'celsius' }
  });
});


router.serialize('/forecast/celsius'); // => routeName='forecast', params: { scale: 'celsius' }
router.serialize('/forecast/newton'); // => not recognized route.
router.serialize('/forecast'); // => routeName='forecast', params: { scale: 'celsius' }
```

This way the `forecast` route would not have to explicitly handle the case where the 'scale' param would be `undefined` (see the example's last line).

humanchimp:
@nathanhammond Did you end up making an add-on for this?

simonihmig:
Just came back after needing this... Could we get a message (from core?) what the status is for this? I see a lot of support, no serious objections, is there something which prevents this from going into FCP?

knownasilya:
Ping on the above :)

text/router-enhancements.md
```diff
@@ -0,0 +1,196 @@
...
+
+Also borrowing from Rails' microsyntax, the user can provide a constraint using the
+config object passed as second argument. For this proposal constraints may only be
+regular expressions.
+
+```js
+Router.map(function() {
+  this.route('forecast', {
+    path: 'forecast/:scale',
+    constraints: { scale: /(celsius|fahrenheit|kelvin)/ }
```
locks: Why isn't `constraints.scale` an array of values?

locks:
@simonihmig turning the question around, is there something preventing the community from implementing this via an addon?

simonihmig:
@locks tbh, I don't know at all. Regarding low-level APIs it seems @nathanhammond can probably answer this? I understood the purpose of this RFC to get this into core, but when there is a way to implement that as an addon without monkey patching things, I am fine with that.

Generally speaking, I would expect these RFCs to enter some defined state after discussion has settled. Like according to https://github.com/emberjs/rfcs#what-the-process-is to be either accepted and merged, or rejected and closed. But this seemed to be kind of left in a limbo, so I just wanted to ask what is preventing this to proceed in whatever way! :)

mazondo:
Would this also allow define routes based on partial segments as well?

Specifically, looking to match
`segment1/segment2-listing` and `segment1/segment2` to different routes

erkie:
So much effort put into this RFC and now it's dormant... Or what's the status on this?

wagenet:
Given the inactivity here I'm going to go ahead and close this. I apologize for the fact that this stagnated and if you would still like to try to move this forward, I'll ensure that it gets a proper review.

It's also worth noting that there is some exploratory router work being done as part of Polaris. That may ultimately address these concerns.

PR #155, opened 2016-07-19
RFC: Ancillary Package Module Standardization

tomdale:
This (first draft) RFC proposes a unified build system for Ember's ancillary packages, like `fastboot`, `simple-dom`, `glimmer`, etc.

[Rendered](https://github.com/monegraph/rfcs/blob/ancillary-package-module-standardization/text/0000-ancillary-package-module-standardization.md)

### DIFF
--- a/text/0000-ancillary-package-module-standardization.md
+++ b/text/0000-ancillary-package-module-standardization.md
@@ -1,4 +1,4 @@
-- Start Date: (fill me in with today's date, YYYY-MM-DD)
+- Start Date: 2016-07-19
 - RFC PR: (leave this empty)
 - Ember Issue: (leave this empty)
 
--- a/text/0000-ancillary-package-module-standardization.md
+++ b/text/0000-ancillary-package-module-standardization.md
@@ -181,6 +181,12 @@ We use two tools to produce the desired output formats:
 2. Rollup, to package the ES5 source into legacy bundle formats like
    UMD.
 
+## Examples
+
+| Repo                 | Description                                                                                               | PR                                                 |
+|----------------------|-----------------------------------------------------------------------------------------------------------|----------------------------------------------------|
+| [fastboot](https://github.com/ember-fastboot/fastboot) (Node-only) | Uses broccoli-module-alchemist to utilize Node 4's ES2015 syntax features while remaining 0.12 compatible | https://github.com/ember-fastboot/fastboot/pull/74 |
+
 # How We Teach This
 
 Because this is an internal infrastructure concern, it does not have
--- a/text/0000-ancillary-package-module-standardization.md
+++ b/text/0000-ancillary-package-module-standardization.md
@@ -1,0 +1,234 @@
+- Start Date: (fill me in with today's date, YYYY-MM-DD)
+- RFC PR: (leave this empty)
+- Ember Issue: (leave this empty)
+
+# Summary
+
+Reduce complexity in Ember and its dependencies' build pipelines by
+standardizing on the JavaScript module syntax.
+
+# Motivation
+
+Though many users think of Ember as a single entity, under the hood it
+is composed of many different small packages.  Due to fragmentation in
+the JavaScript ecosystem, these packages use different build pipelines,
+module formats, output formats, and naming conventions.  This
+fragmentation creates significant overhead for contributors, as they
+must familiarize themselves with the quirks of each package's build
+system and module format. Typically, this information is severely
+underdocumented if it's documented at all.
+
+Additionally, this complexity leaks out to anyone who wishes to consume
+these packages. Rather than adhering to current conventions and best
+practices for distributing code on npm, users must learn the vagaries of
+how to consume each package; whether that's by consuming a global,
+consuming anonymous AMD, named AMD, or something else. Ember's build
+system accrues needless complexity because it has to handle all of these
+different formats distributed in different ways.
+
+Lastly, due to efforts like FastBoot, users expect to be able to run
+their JavaScript code both in the browser and in Node.js.  We need a
+system that makes our packages portable, so that they can be tested and
+run in any environment.
+
+# Detailed Design
+
+This RFC proposes that all packages under the Ember.js umbrella be
+incrementally migrated to standard JavaScript (ES6) modules.  This
+format is trivially transpilable into whatever ultimate module system
+the end user desires, whether that's {named, anonymous} AMD, CommonJS,
+or
+no module system at all using something like Rollup's bundle builds.
+
+JavaScript modules have the benefit of being statically analyzable, so
+tools can determine the dependency graph at build time.  By normalizing
+Ember's dependencies into JavaScript modules, we make it much easier to
+eliminate code that is not actually needed in final builds.
+
+To facilitate a backwards-compatible migration, we will standardize on a
+directory layout for projects, as well as a directory layout for the
+final built output. All packages will be distributed via npm.
+
+To ensure standardized behavior across repositories, a reusable npm
+package `broccoli-module-alchemist` will be created to encapsulate the
+decisions of this RFC.
+
+The source code for broccoli-module-alchemist is available at
+<https://github.com/monegraph/broccoli-module-alchemist>.
+
+## Supported Authoring Formats
+
+Authors may write package source code in TypeScript, ES2015, or
+a mix of both. Because transpilation is done using the TypeScript
+compiler, which supports both, projects can incrementally migrate
+towards TypeScript, or use it only in portions where it makes sense.
+
+Whether using TypeScript or JavaScript, authors should always use
+standard ES2015 module syntax for both importing and exporting modules.
+
+## Target Formats
+
+The original source code is compiled from TypeScript or ES2015+ to ES5
+to ensure maximum compatibility across browser and Node.js versions.
+
+Out of the box, several different module formats are emitted. Output
+targets can be configured on a package-by-package basis. For example, a
+package that only targets Node.js may want to author in TypeScript and
+only emit CommonJS-compatible modules.
+
+### CommonJS
+
+For use by Node.js and tools like browserify/webpack.
+
+### JavaScript (ES2015/ES6) Modules
+
+The standard JavaScript module syntax specified in ES2015, this
+statically-analyzable format is intended to be used by tools like Rollup
+or Ember CLI, which can analyze an entire app and its dependencies and
+strip out unused code.
+
+For Ember's packages, this is the output target that Ember will consume.
+
+Note that only the import/export syntax remains in the output; 
+all other ES2015 syntax is transpiled into ES5. This makes it easy for
+tools like Rollup to consume without having to add a transpiler like
+Babel.
+
+### UMD
+
+UMD modules can be consumed as CommonJS, AMD, or by exporting a global.
+
+#### Using Default Export as Namespace (e.g. `$()` and `$.getJSON()`)
+
+Some libraries designed before standard JavaScript modules use a pattern
+where the default export is a function or other value, and additional
+functionality is added via methods on that value.
+
+For example, in jQuery, you can create a jQuery object using
+`$('.some-selector')`; in other words, the global `$` is a function you
+invoke to access jQuery functionality. But there are additional methods
+on that function, such as `$.getJSON('/users.json')`, to access
+secondary functionality.
+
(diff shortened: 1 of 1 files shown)


text/0000-ancillary-package-module-standardization.md
```diff
@@ -0,0 +1,234 @@
+- Start Date: (fill me in with today's date, YYYY-MM-DD)
```
martndemus: You forgot the start date ;)

text/0000-ancillary-package-module-standardization.md
```diff
@@ -0,0 +1,234 @@
...
+compiler, which supports both, projects can incrementally migrate
+towards TypeScript, or use it only in portions where it makes sense.
+
+Whether using TypeScript or JavaScript, authors should always use
+standard ES2015 module syntax for both importing and exporting modules.
+
+## Target Formats
+
+The original source code is compiled from TypeScript or ES2015+ to ES5
+to ensure maximum compatibility across browser and Node.js versions.
```
martndemus: Would there be room for an opt in to an ES2015+ target? Or are there more reasons to only compile to ES5 other than compat? Something tells me that compiled ES5 might not necessarily be worse off from ES2015 code.

tomdale: @martndemus I was just discussing this with @dgeb. My feeling is that consumers who wanted latest JavaScript would just consume `src/`, but that would mean all consumers have to understand TypeScript, for example.

I'm just not sure how to do this without creating a new directory every year with different versions. I was thinking that, once more target environments support ES2015, a package can make the decision to drop ES5 in a major version rev and start shipping ES2015.

text/0000-ancillary-package-module-standardization.md
tomdale: Fixed :D

text/0000-ancillary-package-module-standardization.md
martndemus: Oh, so the packages itself deliver the compiled output, not the consumer? I read into it as if the whole tree of packages got built with this single one-size-fits-all strategy during build time.

text/0000-ancillary-package-module-standardization.md
```diff
@@ -0,0 +1,234 @@
...
+For Ember's packages, this is the output target that Ember will consume.
+
+Note that only the import/export syntax remains in the output; 
+all other ES2015 syntax is transpiled into ES5. This makes it easy for
+tools like Rollup to consume without having to add a transpiler like
+Babel.
+
+### UMD
+
+UMD modules can be consumed as CommonJS, AMD, or by exporting a global.
```
chadhietala: Historically all UMD stuff in an Ember app does not work as one would expect. This is largely due to the fact that loader.js is not AMD compliant (does not implement `define.amd`) so it always falls into globals mode.

tomdale: @chadhietala The UMD stuff here is for other, non-Ember consumers. My hope is that Ember can jump to consuming the ES6 modules version, and the responsibility for compiling into something loader.js-compatible happens there. If not, it would be straightforward to add an AMD build.

text/0000-ancillary-package-module-standardization.md
tomdale: For maximum compatibility with AMD, older Nodes, webpack, etc., I think we have no choice but to provide compiled output. _That said_, if we're able to standardize across the ecosystem, it would be nice if the ultimate package (Ember CLI in this case) could just consume `src/` across all of its dependencies. I don't think that's that far-fetched. The only wrinkle might be TypeScript, which can have per-package configuration details that change e.g. module lookup. I'm not sure that would survive the put-it-all-in-one-directory-and-compile approach.

tomdale:
Added a link to [this FastBoot PR](https://github.com/ember-fastboot/fastboot/pull/74).

text/0000-ancillary-package-module-standardization.md
stefanpenner: > @martndemus I was just discussing this with @dgeb. My feeling is that consumers who wanted latest JavaScript would just consume src/, but that would mean all consumers have to understand TypeScript, for example.

I'm not sure this is sufficient, as the `src/` will not just be es2015 CODE but could be anything. TS/Flow/Dart and can be decorated with dev instrumentation and asserts.

I suspect what consumers of ES2015 will want, is actually (most likely) also the result of some build step. Which leads me to believe, we should consider the *baseline target* being configurable, and by default we can guarantee ES5 and ES2015, but over time add ES2016 and so an so forth.

So What if we lifted the constraint slightly, forcing all our micro-libs to support something like:

```js
function buildMyMicroLib(options) {
  options.target // => 2015 || ES5 || ...

  return produceForTarget(tree, target);
}
```

That way:
* for our build tools, we can request the baseline target and the tool will provide.
* For others not using our build tools, we can pre-compute the various outputs they need (maybe one per baseline)

--

I guess my above thoughts are inspire by things like typings, asserts, instrumentation and likely other build time concerns source, which leave our source unconsummable without a build step.

text/0000-ancillary-package-module-standardization.md
stefanpenner: The loader is AMD compat, it just supports a minimal subset (something requiring the smallest number of moving parts and a build pipeline)

text/0000-ancillary-package-module-standardization.md
```diff
@@ -0,0 +1,240 @@
...
+started.
+
+Because users who need this information are intermediate to advanced,
+this RFC itself should serve as a good starting point. Additionally,
+configuration information and how-tos will be incorporated into
+`broccoli-module-alchemist`'s README.
+
+# Drawbacks
+
+The biggest drawback is that this change may introduce temporary
```
stefanpenner: seems like we could bump our micro-libs major version to overcome this (slightly)

text/0000-ancillary-package-module-standardization.md
```diff
@@ -0,0 +1,240 @@
...
+instability into the ecosystem. Many packages may rely on non-standard
+semantics or syntax, and updating them may divert work that could go
+elsewhere.
+
+Additionally, this RFC does not go so far as attempting to standardize
```
stefanpenner: We should most likely utilize ember-cli's existing infrastructure with supports any (but mocha/qunit easily).
If changes are needed for the default support of those to cater to these microlibs, we should do that.

text/0000-ancillary-package-module-standardization.md
```diff
@@ -0,0 +1,240 @@
...
+the testing infrastructure. Some packages use mocha, while others use
+QUnit. Some run their tests in Node.js, others in PhantomJS. Ideally, a
+future RFC could attempt this unification; although rewriting entire
+test suites just to use a different test harness does not seem like the
+best use of precious time and energy.
```
stefanpenner: Ya, ideally we will want to share some of this infrastructure with our ember-app's, as they also would like to be tested in node/phantom/browser/etc.

text/0000-ancillary-package-module-standardization.md
```diff
@@ -0,0 +1,240 @@
...
+# Unresolved questions
+
+- What is the exact file system layout?
+- How do we handle "multi-package" packages, like Glimmer, which include
+  multiple logical packages in a `packages` directory? Should each
+  package have its own `package.json`?
+- Does it make sense to TypeScript-enable the infrastructure, given that
+  many new ecosystem packages rely on it?
+- Is it possible to conventionalize testing? If the only two systems in
+  common use are Mocha and QUnit, can we support both?
```
stefanpenner: Do we want to explore linker/packager esq work to allow resolving / building nested dependencies as part of this RFC. Or should we explore this at a later time, considering it an enhancement tot his baseline.

(i prefer the later)

stefanpenner:
glad to see movement on this 👍

text/0000-ancillary-package-module-standardization.md
tomdale: @stefanpenner The idea of a configurable baseline with es5 as the default is the right approach here I think.

tomdale: @stefanpenner Also right now `broccoli-module-alchemist` assumes the `src/` directory as input, but you point out correctly that people may want to strip assertions, etc. I will change that to have it accept a path or Broccoli tree as the input directory to allow for people to do their own Broccoli-based transforms before passing the end result to `broccoli-module-alchemist` for module compilation.

text/0000-ancillary-package-module-standardization.md
tomdale: @stefanpenner No doubt, one of the biggest problems with AMD is how underspecified it is. It's not that the loader is AMD-compatible, but rather that the UMD format relies on heuristics that are not part of the AMD spec. Because of that, even though we are AMD compatible, UMD modules do not reliably determine that they are being consumed by an AMD loader. Hopefully soon this dumpster fire will be behind us.

text/0000-ancillary-package-module-standardization.md
tomdale: @stefanpenner The big question in my mind is: if we bump the major version for our microlibs, what module formats should we support? Do we continue to provide globals/AMD builds? I'm half tempted to switch over entirely to ES6 modules, since most (maybe all?) people using these libraries use a build tool that can be easily augmented to support ES6 modules, if they're not using them already.

text/0000-ancillary-package-module-standardization.md
tomdale: @stefanpenner Not a bad idea, but does that mean we add ember-cli as a dev dependency to each microlib? Not totally opposed to this, but it is a significant dependency in terms of file size. If we can get away with just broccoli+broccoli-cli, that's much less code to install over airplane wi-fi. 😁

text/0000-ancillary-package-module-standardization.md
tomdale: I think there is value in this standardization before the linker/packager stuff is figured out, so I don't think it blocks, but I want to be absolutely sure it's incorporated into the design and we don't do something that would make that work harder.

text/0000-ancillary-package-module-standardization.md
stefanpenner: to make this very general, maybe that path should be read from `package.json`

text/0000-ancillary-package-module-standardization.md
stefanpenner: @tomdale I think we should use ember-cli, and continue the effort to reduce "weight" of it, as that will help our whole ecosystem. For example, making bower an optional dependency will cut the filesize and filecount by about 30%.

My motivation for this is as follows, rather then re-inventing the bits that are good in ember-cli (there is a very large number of these things) outside of it, we should take that energy in removing the heaviness of ember-cli's default install. Fixing the heaviness, will also provide us larger leverage. 

On this note, the effort is actually on-going @kellyselden is working on it, I am as well, as are others.

So in the interim, lets take incur this cost.

text/0000-ancillary-package-module-standardization.md
stefanpenner: @chadhietala you should jump in, and just make sure what happens here doesn't prevent or hurt the future of linker/ etc.

rtorr:
The title/RFC only suggests _ancillary packages_, will this be the eventual goal of ember-cli to use typescript over babel to compile ember projects? We are making choices right now that rely on babel, and knowing this would be helpful for our team.

tomdale:
@rtorr No immediate plans for that. We certainly do want to eventually support authoring Ember apps in TypeScript, but never at the expense of the JavaScript experience. Given how entrenched Babel is in the Ember ecosystem, I think going in that direction is a very safe choice.

For us, the desire to write libraries in TypeScript means we need to support the TypeScript compiler anyway, and no point using two tools if one tool will do the job.

rtorr:
@tomdale Good to hear and thanks for clearing that up!

tomdale:
We need to make sure that we account for potential changes in the language, particularly if we want to take advantage of experimental features. Probably this means we will need to support per-project overrides of how transpilation happens. Because TypeScript is less configurable along these lines, it means we may have to revisit the decision to use TypeScript as the transpiler and use Babel for JavaScript transpilation.

tomdale:
Closing this RFC for now because it's gone stale and been superseded by some work happening in Ember CLI itself. We can revisit this in the future if needed.

PR #158, opened 2016-07-22
RFC: Asset Loader Service

trentmwillis:
Pre-reading/context: [Asset Manifest RFC](https://github.com/emberjs/rfcs/pull/153)

[Rendered](https://github.com/trentmwillis/rfcs/blob/asset-loader-service/text/0000-asset-loader-service.md).

### DIFF
--- a/text/0000-asset-loader-service.md
+++ b/text/0000-asset-loader-service.md
@@ -1,0 +1,205 @@
+- Start Date: 2016-07-22
+- RFC PR: (leave this empty)
+- Ember Issue: (leave this empty)
+
+# Summary
+
+The Asset Loader service is an [Ember Service](http://emberjs.com/api/classes/Ember.Service.html) that is responsible for loading assets specified in an [Asset Manifest](https://github.com/emberjs/rfcs/pull/153). The API for loading these assets is Promise-based for integration with [the `getHandler` API from Router.js](https://github.com/tildeio/router.js/pull/183).
+
+Note: It is recommended that you read the [Asset Manifest RFC](https://github.com/emberjs/rfcs/pull/153) before reading this RFC.
+
+# Motivation
+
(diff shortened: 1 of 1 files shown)


rwjblue:
Where are we at with this? 

@dgeb / @MiguelMadero / @nathanhammond / @trentmwillis - Have y'all reviewed?

trentmwillis:
We have a working implementation in [ember-asset-loader](https://github.com/trentmwillis/ember-asset-loader/) that has been used successfully for both lazy loading Engines and generic use cases (e.g., loading a charting lib).

Primary open question to me is where does this Service live in the future? For many use cases, having a standalone, core-maintained addon seems sufficient. But, for the lazy Engines use case, it needs to integrate with some of the internals of Ember's Router, which makes it seem like this could potentially be part of Ember itself. At the very least, if it continues as a separate addon, we likely need a solid API for hooking into the Router.

nathanhammond:
Have reviewed the RFC and many of the PRs that went into the implementation. This is set by my estimation.

trentmwillis:
@rwjblue clarified some wording. I think this is ready to move to FCP.

rwjblue:

text/0000-asset-loader-service.md
```diff
@@ -0,0 +1,208 @@
...
+};
+```
+
+### `loadAsset`
+
+The `loadAsset` method loads a single asset into the application. It accepts an `asset` object which contains `uri` and `type` fields, both of which are strings. The `type` determines which loading method will be called internally according to the type of asset being loaded and the `uri` will be passed as an argument to said loading method to specify the actual asset to load.
+
+`loadAsset` will return an `AssetPromise` that will resolve when the asset is successfully loaded or reject when the asset fails to load.
+
+Subsequent calls to `loadAsset` with the same arguments will return the same `AssetPromise` object to ensure a single source of truth for determining asset state as well as to prevent duplicate requests being made for the same asset.
```
courajs: Can we clarify here that this is true unless/until `retryLoad` is called, and link down to the `Provide A Way To Attempt Recovery` section?

rwjblue:

PR #143, opened 2016-05-09
Module Unification

dgeb:
[Rendered](https://github.com/dgeb/rfcs/blob/module-unification/text/0000-module-unification.md)

PR #159, opened 2016-07-26
Add base model

pangratz:
[Rendered](https://github.com/pangratz/rfcs/blob/ember-data-base-model/text/0000-ember-data-base-model.md)

### DIFF
--- a/text/0000-ember-data-base-model.md
+++ b/text/0000-ember-data-base-model.md
@@ -13,7 +13,7 @@ available for models as well.
 
 # Motivation
 
-Currently if the models of your app have common functionality, a commong `attr`
+Currently if the models of your app have common functionality, a common `attr`
 or something similar, you need to create custom base model and make sure you
 extend all existing models from it and update all models generated via the
 `ember generate model` blueprint, so they extend from the same base model.
--- a/text/0000-ember-data-base-model.md
+++ b/text/0000-ember-data-base-model.md
@@ -1,0 +1,60 @@
+- Start Date: 2016-07-26
+- RFC PR: (leave this empty)
+- Ember Issue: (leave this empty)
+
+# Summary
+
+Add a base model at `app/models/-base.js` which is used to extend all the
+applications' models from.
+
+There is already the concept of an application wide adapter and serializer, so
+this concept of a base class which all sub-classes extend from is made
+available for models as well.
+
+# Motivation
+
+Currently if the models of your app have common functionality, a commong `attr`
+or something similar, you need to create custom base model and make sure you
+extend all existing models from it and update all models generated via the
+`ember generate model` blueprint, so they extend from the same base model.
+
+This is cumbersome, error prone and should be handled by Ember Data out of the
+box.
+
+Examples are `attr`s like `createdAt` and `updatedAt` available on all models;
+or `revision: attr('string')` used in combination with a CouchDB backend.
+
+# Detailed design
+
+Create a base model at `app/models/-base.js` and update the model blueprint so
+it doesn't extend from `DS.Model` but from the new base model instead.
+
+# How We Teach This
+
+Since this is a similar to the currently existing concept of an application
+wide adapter and serializer, this doesn't introduce a new concept per se and
+we can extend the documentation by the new - model - facet.
+
+# Drawbacks
+
+`¯\_(ツ)_/¯`
+
+We can't be consistent with the adapter and serializer naming, as a model named
+`application` might exist in the app and this would mean that all other models
+would extend from this, which might lead to unexpected behavior.
+
+# Alternatives
+
+- create a mixin which is used instead of a base model and manually mix it into
+  all your models
+- create a base model yourself, and have the advantage of naming it as you like
+  (`-base` might not be the best name)
+
+# Unresolved questions
+
+- what should the name of the base model be?
+  - is `app/models/-model.js` better?
+  - should the base adapter and serializer be renamed to
+    `app/{adapter,serializer}/-base.js` too?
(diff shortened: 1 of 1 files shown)


text/0000-ember-data-base-model.md
```diff
@@ -0,0 +1,60 @@
...
+Add a base model at `app/models/-base.js` which is used to extend all the
+applications' models from.
+
+There is already the concept of an application wide adapter and serializer, so
+this concept of a base class which all sub-classes extend from is made
+available for models as well.
+
+# Motivation
+
+Currently if the models of your app have common functionality, a commong `attr`
```
rwjblue: s/commong/common/

pangratz: Mille grazie

pangratz:
I haven't followed #143 too closely, but this RFC might need to be updated once that one is accepetd...

nathanhammond:
Notes:
- Should specify behavior for crossing engine boundaries since the store is often reachable in child engines.

Concerns:
- It starts to be a place where more and more properties get collected and you end up with the one model to rule them all?
- It's a dynamic runtime injection of a dependency instead of declarative. It makes bundling more complicated.
- Model.extend().extend().extend() starts to weigh on performance.
- We can already implement this relatively gracefully in user space and we're introducing more magic into the default set of magic you have to learn to use Ember.

fivetanley:
As somebody who refactored the logic we use for "base" serializers and adapters recently, I'd rather not carry forward any more of this kind of default magic. I agree with @nathanhammond that the amount to learn, performance, and bundling issues make this feature seem like a lot of maintenance burden to avoid a little boilerplate.

If we look at other ORM implementations like Rails's ActiveRecord, Django, Sails, Kinex, etc., none of them seem to use this pattern and I haven't seen it become very popular in userland for the reasons listed above either.

Maybe some alternatives:

For the `revision` use case, maybe use `DS.Model.reopen`, but not a huge fan of this either. (Could we store `revision` in `meta` and translate at the serializer/adapter layer?). Maybe ember data needs some place other than meta for extra attributes.

hjdivad:
> Currently if the models of your app have common functionality, a common attr or something similar, you need to create custom base model and make sure you extend all existing models from it and update all models generated via the ember generate model blueprint, so they extend from the same base model.

> This is cumbersome, error prone and should be handled by Ember Data out of the box.

@pangratz what do you think about instead making it easier to create an app-specific model blueprint?  This gives users the opportunity to codify this kind of app convention (our models extend from our own base class) without committing ember-cli to making apps with a preemptive app-specific base class.

pangratz:
@nathanhammond @fivetanley thanks for the great feedback. Adding another magical component might not be the best solution; I haven't thought of that...

---

@hjdivad I like the idea of solving this on the blueprint level!

We could extend the existing model blueprint by a `--base` option, as it is already possible with `adapter` and `serializer`. This might be the easiest solution, but it would need the user to always specifiy the `--base` option. We could make this smart to always use the `-base` model, if `app/models/-base.js` exists - or we could make it possible to save the default value for this option in `.ember-cli` for example, though I am not sure if this is encouraged...

I am not sure how an app-specific model blueprint would be implemented, because [at the moment it contains template variables](https://github.com/emberjs/data/blob/7e27ecdd7ffb2f8a51cdb8897c121733a560952f/blueprints/model/files/__root__/__path__/__name__.js#L4) which would be something the user would need to consider when the app-specific blueprint would be adapted...

nathanhammond:
I really would like people to get into the habit of creating an in-repo-addon for their custom blueprints. If people knew this tool were available to them I think it'd solve this whole problem.

```
ember g in-repo-addon custom-blueprints
mkdir -p lib/custom-blueprints/blueprints/model/files/__root__/__path__/
touch lib/custom-blueprints/blueprints/model/files/__root__/__path__/__name__.js
```

hjdivad:
@nathanhammond keeping them in an in-repo addon sounds like a good convention.  

I still think there's some value to being able to override base bluerprints for a couple of reasons

1. Less cognitive overhead to keep doing standard `ember g model Thing` instead of `ember g my-app-model Thing`
2. The natural thing for the generator to do is copy the standard blueprint instead of creating a fresh one.  This makes it much easier for users to make small changes like changing the base class without having to deeply understand the blueprint API


@pangratz seems reasonable to me that `ember g model NAME` would accept `--base`.  It feels weirdly specific for serializer and adapter to take it but not model.

> I am not sure how an app-specific model blueprint would be implemented, because at the moment it contains template variables which would be something the user would need to consider when the app-specific blueprint would be adapted...

Good point.  Maybe we would create an `index.js` with a `locals` that already computed the base variables, or, if we're nicer, pass in the base variables to `locals`.

---

I realize that copying the base blueprint and editing it creates upgrade work, but i suspect people would find that reasonable when they're overriding base blueprints.

nathanhammond:
@hjdivad Blueprints are overrideable and the in-repo-addon can either add a new blueprint type or clobber the default blueprint type as long as it specifies that it comes after ember-data and ember-cli-legacy-blueprints. An exceptionally clever person would even be able to `require` the upstream `index` file and not have to reimplement that themselves.

As to `ember g model NAME` not accepting `--base` that is controlled completely inside of the Ember Data blueprints at this point.

(Note: things that seem reasonable to the people currently chatting on this thread aren't necessarily reasonable for everybody else. Maybe the real feature here is making `ember g blueprint model --based-on=model` do the right thing.)

hjdivad:
@runspired `--extends` sounds like a good way of achieving the main goal while also narrowing the scope a bit.

PR #164, opened 2016-08-19
RFC for public ApplicationInstance API

rynam0:
[Rendered](https://github.com/rynam0/rfcs/blob/master/text/0000-public-appInstance-api.md)

### DIFF
--- a/text/0000-public-appInstance-api.md
+++ b/text/0000-public-appInstance-api.md
@@ -1,0 +1,52 @@
+- Start Date: 2016-08-19
+- RFC PR: (leave this empty)
+- Ember Issue: (leave this empty)
+
+# Summary
+
+Expose a public API that will give access to an Ember ApplicationInstance. This would allow external functions to lookup and invoke actions defined in an Ember Application.
+
+# Motivation
+
+Embedding Ember applications in existing apps works wonderfully and is a huge asset to those of us who wish to incrementally replace an existing UI with standalone Ember Apps with the eventual goal of using Ember Engines as a full replacement of the existing app.  However, simply embedding Ember Applications does not address the need of integrating these apps with the existing application.
+
+To describe the motivation behind this request, please picture a large enterprise application written in some other framework.  In my case, these apps are written in GWT (Google Web Toolkit). Some of the existing features are easily replaced with standalone embedded Ember Applications.  There are other parts of the application that are not so simple to replace as the existing application must be able to invoke actions on an embedded Ember app.
+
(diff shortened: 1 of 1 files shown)


rwjblue:
Hmm.  It seems that you are talking about the `Ember.getOwner` API here.  How does this proposal differ from that?

rynam0:
@rwjblue I'm not entirely positive TBH.  I don't think we can get the default instance from an Application currently, can we?  Is there something in the `Ember.getOwner` API I'm not aware of that would be similar to the example given in the RFC?

rynam0:
To clarify: From what I can gather, `Ember.getOwner` is not an API for accessing the `ApplicationInstance` from a booted Ember `Application`.  This request is to provide public access to the current `__deprecatedInstance__`.  Perhaps I am missing something very obvious in which case I apologize and hope you can point me in the right direction. Please let me know if I can provide more information to facilitate further discussion.  Thanks!

rynam0:
@rwjblue: Should I have opened an issue for this instead?

rynam0:
Closing as I was under the impression that one had to go through the '__deprecatedInstance__' to access the container.  The following appears to be a public API that I can rely on for this purpose.

    myapp.__container__.lookup('route:index').send('myAction');

locks:
@rynam0 something with a double underscore is certainly not a public API :P

rynam0:
Hmm, ok.  Thanks, @locks. So my initial impression on naming was correct after all.  I did try to locate docs on variable naming conventions but came up empty-handed.  After some pondering I thought maybe the convention could be:  a single underscore prefix for private and double underscores for public (named this way to avoid collisions).  Are there conventions documented somewhere you could refer me to?  Is this request indeed the `Ember.getOwner` API as @rwjblue  suggests?  Any information the team could supply here would be awesome as I am now using this in a production app and will be in some trouble if I don't a) move to engines very soon or b) figure out how to keep this from breaking in the near future.  Thanks!

nathanhammond:
You should probably accomplish your goals by setting up an `initializer` or `instance-initializer` (depending upon your needs) in your Ember application and pass off a reference to the first argument to your external script.

Extra reading:
- http://emberjs.com/blog/2015/05/13/ember-1-12-released.html#toc_instance-initializers
- http://emberjs.com/deprecations/v2.x/#toc_initializer-arity
- http://emberjs.com/deprecations/v2.x/#toc_ember-applicationinstance-container

Given those tools I suspect you're able to meet your needs. As it stands the RFC you have proposed doesn't include enough detail to be actionable. It feels more like a question on "how do I do this?" than it does a feature proposal.

Please drop by [#dev-ember](https://embercommunity.slack.com/messages/dev-ember/) in the [Ember Community Slack](https://ember-community-slackin.herokuapp.com/) for discussion of feature development on Ember itself.

wagenet:
Given the six years of inactivity here I think we can close it. If there is something here still worth considering, please let me know.

PR #171, opened 2016-10-11
Improved ajax

pangratz:
[Rendered](https://github.com/pangratz/rfcs/blob/ember-data-improved-ajax/text/0000-ember-data-improved-ajax.md)

### DIFF
--- a/text/0000-ember-data-improved-ajax.md
+++ b/text/0000-ember-data-improved-ajax.md
@@ -7,6 +7,19 @@
 Add public hooks on `rest`-adapter which can be used to customize the
 properties of an AJAX request being made for an adapter operation.
 
+1) hooks to get properties for an request which should be made
+  - `methodForRequest`
+  - `headersForRequest`
+  - `urlForRequest`
+    - `host`
+    - `namespace`
+    - `pathForRequest`
+    - `queryForRequest`
+  - `bodyForRequest`
+
+2) hook for making the request
+  - `makeRequest`
+
 # Motivation
 
 Currently `host`, `namespace`, `urlForXXX()` and `headers` is the only public
@@ -44,6 +57,7 @@ findRecord(store, type, id, snapshot) {
   return RSVP.resolve(requestOptions).then((options) => this.makeRequest(options) );
 }
 ```
+
 ### `_requestFor`
 
 Each hook for step 1) is invoked with all the arguments which are passed to the
@@ -176,10 +190,12 @@ headersForRequest(options) {
 
 # Unresolved questions
 
-- the currently propose API doesn't allow to only use public API to
+- the currently proposed API doesn't allow to only use public API to
   [customize](https://github.com/emberjs/data/blob/2926c47453d50d6b75590d2ff447a4d0da66833a/addon/adapters/json-api.js#L226-L234)
   the `json-api`-adapter, though the current implementation could be changed to
-  set the `Content-Type` header instead (which would be more agnostic anyway)
+  set the `Content-Type` header instead (which would be more agnostic anyway).
+  This might be an indicator that the proposed API still needs some
+  adjustments so `json-api` adapter uses only public API.
 - currently `requestType` is used to indicate the specific adapter operation,
   but that name might not be to expressive. Should this be renamed to the more
   verbose, like `adapterOperation`?
--- a/text/0000-ember-data-improved-ajax.md
+++ b/text/0000-ember-data-improved-ajax.md
@@ -30,7 +30,7 @@ customize a request with only public API.
 An adapter operation is split up into 2 steps
 
 - 1) collecting the properties for the request, i.e. `_requestFor`
-- 2) executing the request, i.e. `_makeRequest`
+- 2) executing the request, i.e. `makeRequest`
 
 So for a `findRecord` it would look like this:
 
@@ -41,7 +41,7 @@ findRecord(store, type, id, snapshot) {
     requestType: "findRecord"
   });
 
-  return RSVP.resolve(requestOptions).then(this._makeRequest);
+  return RSVP.resolve(requestOptions).then((options) => this.makeRequest(options) );
 }
 ```
 ### `_requestFor`
@@ -51,7 +51,8 @@ corresponding adapter method. Additionally `requestType` is set so the hooks
 know for which specific adapter operation they are invoked. To allow for
 asynchronous configurations (e.g. pre-flight request to refresh an
 authentication token), `_requestFor` returns a promise which resolves with an
-object describing the request to make.
+object describing the request to make. The `_requestFor` method is private and
+calls the new public hooks.
 
 ``` js
 _requestFor(params) {
@@ -73,7 +74,7 @@ Invoked to get the HTTP method for the request. Should return one of `"GET"`,
 Get the URL for the request. This hook returns the URL for the request which
 should be made, e.g. `/users/1` or `api.example.com:4321/users?name=abc`
 
-This hook uses 2 new sub-hooks to construct the URL:
+This hook uses 2 new public sub-hooks to construct the URL:
 
 - `pathForRequest` - re-uses the existing `urlForXXX` hooks
 - `queryForRequest` - get an object for all query params for the request
@@ -112,6 +113,7 @@ queryForRequest({ requestType, query }) {
   }
 }
 ```
+
 ##### `bodyForRequest` (currently `dataForRequest` in `ds-improved-ajax`)
 
 Body for the HTTP request which should be made.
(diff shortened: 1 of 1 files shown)


nathanhammond:
If we're going through this effort is it worth trying to build something that decouples us from the underlying jQuery implementation? There is hopefully a future in which we don't ship that by default and it would be sad to require jQuery just for `$.ajax`.

See: https://github.com/ember-cli/ember-ajax

pangratz:
@nathanhammond yep, I definitely agree. The hooks should be designed to be agnostic of the specific library/implementation which does the actual request. I forgot to mention that explicitly in the RFC...

mkay581:
@nathanhammond [fetch](https://fetch.spec.whatwg.org/) maybe?

tchak:
I think the hooks as described in this RFC have no dependency on `$.ajax`. They are some examples based on `$.ajax` usage, men we could remove them, or add more examples with `fetch` for example.

One suggestion I have is that we are passing to `makeRequest()` a hash of options. The hash looks like this:
```
{
  url: 'https://api.com/posts',
  method: 'POST',
  headers: {},
  body: '{ post: {} }'
}
```

There is a spec for a [`Request` object](https://developer.mozilla.org/en-US/docs/Web/API/Request) (as part of `fetch`). I am not saying we should use it directly, but, I think we should model a `Request` object in `ember-data` after it. `DS.Request`? I would also expect both be compatible and so once we support fetch, it would be possible to pass in native `Request` object.

text/0000-ember-data-improved-ajax.md
```diff
@@ -0,0 +1,207 @@
...
+calls the new public hooks.
+
+``` js
+_requestFor(params) {
+  let method  = this.methodForRequest(params);
+  let url     = this.urlForRequest(params);
+  let headers = this.headersForRequest(params);
+  let body    = this.bodyForRequest(params);
+
+  return RSVP.hash({ method, url, headers, body });
```
bmac: Should `_requestFor` be made a public method in case the user wants to add some form of config other then method, url, headers or body?

text/0000-ember-data-improved-ajax.md
```diff
@@ -0,0 +1,207 @@
...
+describe the request which should be made.
+
+``` js
+/**
+  @method
+  @public
+  @param request Object containing all properties for the request which should be made
+  @return Promise
+ */
+makeRequest(request) {
```
bmac: I think we should enumerate all of the properties Ember Data expects/supports on the `request` object.

pangratz: Good point!

bmac: Based on my reading of this rfc it looks like it supports  `url`, `method`, `headers` and `body`.  With the exception of `url` all of these properties are shared by the 2nd argument to `fetch` (See: https://fetch.spec.whatwg.org/#requestinit). Which may lead a user to believe it supports all the properties that `fetch` supports.

This leads me to ask 2 questions.
1. How much of the RequestInit api should Ember Data implement?
2. Should the `url` property be pulled out of the request object and `makeRequest` API be change to align more closely with the `fetch` api? `makeRequest(url, requestInit)`

bmac:
@tchak I'm not sold on a `DS.Request` class. What is the advantage of that wrapping the request state in a class instance vs just having users return a plain object with the same properties?

bmac:
@pangratz in response to you unresolved questions section at the end of the rfc.
> currently requestType is used to indicate the specific adapter operation, but that name might not be to expressive. Should this be renamed to the more verbose, like adapterOperation?

I'm in favor of renaming `requestType` to `adapterOperation`.

> should queryForRequest be named queryParamsForRequest?

Again I like this change but I don't feel very strongly about it.

> should there be more hooks within urlForRequest, as outlined in the [Syntax](https://en.wikipedia.org/wiki/Uniform_Resource_Identifier#Syntax) section on the URLs wiki page?

I do not think we need any more hooks inside `urlForRequest` other then what is described in this rfc. I believe the current hooks satisfy most of not all of the requests for customization points in the url that we've seen in the Ember Data repo over the years. I suspect if we added more hooks they would be unused in 99% of cases and the remaining cases users can still override `urlForRequest` to generate the url they need.

> - the currently proposed API doesn't allow to only use public API to customize the json-api-adapter, though the current implementation could be changed to set the Content-Type header instead (which would be more agnostic anyway). This might be an indicator that the proposed API still needs some adjustments so json-api adapter uses only public API.
> - what hooks are missing?
>    - setting the dataType [#4357](https://github.com/emberjs/data/pull/4357)?

I was originally thinking `_requestFor` should be made public so users can return other config options that the [fetch api supports](https://fetch.spec.whatwg.org/#requestinit) besides `headers`, `method` and `body`. But that doesn't solve the case of special jquery specific configuration. 

2 possible approches:
1. Limit the ajax adapter to only allow configuring of jQuery ajax features that can be described using a fetch api. (For example contentType would use the Content-Type header but configuring `dataType` would be unsupported.) The rational being this would make it easier to move the rest adapter to a service worker some day since xhr isn't supported inside a service worker but fetch is.
2. Add a `jqueryAjaxOptions()` method and document that it will only be called when jQuery is being used for `makeRequest()` (Possibly difficult to detect).

text/0000-ember-data-improved-ajax.md
```diff
@@ -0,0 +1,207 @@
...
+``` js
+urlForRequest(options) {
+  let path = this.pathForRequest(options);
+  let query = this.queryForRequest(options);
+
+  if (path.indexOf('?') !== -1) {
+    warning("use queryForRequest to specify query params");
+  } else if (query) {
+    let sortedQueryParams = this.sortQueryParams(query);
+    let queryParams = Ember.$.param(sortedQueryParams);
```
pangratz: You mean additionally to `queryForRequest`?

pangratz: Hmm, I see. Maybe `queryStringForRequest()`?

But then we would have 5 hooks which are involved getting the URL for a request:

- `urlForRequest`
  - `pathForRequest`
  - `queryForRequest`
  - `sortQueryParams`
  - `queryStringForRequest`

For now I am not convinced that an additional hook is necessary. Can you think of an use case where such a hook would come in handy?

bmac: I've seen a few comments on slack recently were people want arrays or nested query params to be serialized differently from how `$.param` does it.

I wonder how widely used `sortQueryParams` is and I wonder if we could deprecate that and let users do that work in `queryForRequest` or `queryStringForRequest` if they need it.

pangratz: Ah, arrays and nested queries are definitely a valid use case 👍 

Deprecating `sortQueryParams` in favor of `queryStringForRequest` sounds reasonable to me!

text/0000-ember-data-improved-ajax.md
```diff
@@ -0,0 +1,207 @@
...
+  let headers = this.headersForRequest(params);
+  let body    = this.bodyForRequest(params);
+
+  return RSVP.hash({ method, url, headers, body });
+}
+```
+##### `methodForRequest`
+
+Invoked to get the HTTP method for the request. Should return one of `"GET"`,
+`"PUT"`, `"POST"` or `"DELETE"`.
```
woprandi: Why not "PATCH" ?

amiel:
This will help resolve some issues we've had with ember-data-url-templates :)

cibernox:
This seems to have fallen idle. This would be much needed in order the build a reasonable compatibility layer to be able to swap jQuery by just raw fetch. 
At the moment adapters setup headers using the `beforeSend` function on ajax, which has no fetch counterpart and it is hard, if not plain impossible, to introspect.

Is there any open question or significant pushback from anyone that prevents this from advancing?

denzo:
Same as above comment. It has been a while, it would be great to know where it's at or where it is on the timeline.

Will be happy to contribute to speed it up, if anything at all is needed.

The fact that https://emberjs.com/api/ember-data/2.16/classes/DS.RESTAdapter/methods/headersForRequest?anchor=headersForRequest is still in the docs but can not be used isn't ideal.

@pangratz @bmac @tchak

tchak:
Her is the list of remarks and concerns so far on this RFC:
- make `requestFor` public
- the RFC dose not provide a way to customise `json-api` adapter with public methods
- `queryForRequest ` -> `queryParamsForRequest `
- deprecating `sortQueryParams` in favour of `queryStringForRequest`

"make things more standard":
- return `Request` object from `requestFor` (?)
- `makeRequest` to take `Request` (?)
- return a `URLSearchParams` from `queryParamsForRequest ` (?)

wycats:
@emberjs/ember-data-contributors @igorT what do we need to do to make progress on these questions?

sandstrom:
Regarding `headersForRequest`, I think it would be good if it returned a promise that resolved to a hash of headers (i.e. any consumer must expect either an object or a promise).

It's common to append authentication/session tokens in headers, and they may be stored asynchronously (for example in IndexedDB).

fivetanley:
> return a URLSearchParams from queryParamsForRequest

@tchak is this polyfillable? This isn't supported by Internet Explorer natively.

rwjblue:
Where are we on this? Is this still a direction we would like to go?

tchak:
@rwjblue I think we want to finish migration to `fetch` in `ember-data` before adding anything else to the network layer. For what it worth, I implemented this RFC (and other things) in an addon https://github.com/tchak/ember-fetch-adapter

runspired:
Closing as we are definitely moving in a very different direction. Keep your eyes open for a new RFC :)

PR #173, opened 2016-10-18
Add support for nested saves

hjdivad:
[Rendered](https://github.com/hjdivad/rfcs/blob/nested-save/text/0000-nested-save.md)

### DIFF
--- a/text/0000-nested-save.md
+++ b/text/0000-nested-save.md
@@ -1,4 +1,4 @@
-- Start Date: 2016-10-02
+- Start Date: 2016-10-18
 - RFC PR:
 - Ember Issue:
 
@@ -15,9 +15,9 @@ This rfc proposes two main API changes:
     let promise = parentRecord.save();
     childRecord.savedWith(promise):
   ```
-  where the childRecord will now follow the state transitions of the parent record.
+  where `childRecord` will now follow the state transitions of the parent record.
 
-  2. Allow the `store.push` method to receive an optional client id, which it will
+  2. Allow `store.push` to receive an optional client id, which it will
   use as a fallback to avoid creating duplicate records.
 
 # Motivation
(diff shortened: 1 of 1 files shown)


sandstrom:
@hjdivad I was mainly throwing the idea out there (id-naming). If you prefer `clientID` lets go with that, I'm very happy whichever name it gets 😃 

Moving `savedWith` into an adapter-concern would be awesome! 🚀

mspisars:
Is there a timeline for this? Can this be brought up for discussion on being added to the timeline?

robneville73:
is anyone working on this? I might be interested in tackling some of this.

hjdivad:
@robneville73 nobody is working on it at the moment, but everyone's pretty positive on the idea.

The first step is  to rework the rfc (slightly) to make the mechanics entirely an adapter concern.
- move `savedWith` to `snapshot` (off `DS.Model`)
- entangle in snapshot via adapter
- move `clientId` to the `meta` portion of a resource to be consistent with jsonapi

robneville73:
I can fork your rfc and try to do that if you'd like though I'm a little fuzzy on what you mean by entangling in snaphot via adapter.

hjdivad:
@robneville73 
> I can fork your rfc and try to do that if you'd like though I'm a little fuzzy on what you mean by entangling in snaphot via adapter.

Basically think through how to change the [example](https://github.com/hjdivad/rfcs/blob/nested-save/text/0000-nested-save.md#example-usage-with-api-server-echoing-clientid) when you don't have `DS.Model#savedWith`.

I imagine the way this would work is that the parent object will be `save`d and then in its adapter the relevant hook (`createRecord`, `updateRecord`) would call `belongsTo` / `hasMany` to get the related snapshots to call `savedWith`.

Something like

```js
// models/order.js
DS.Model.extend({
  items: DS.hasMany('items'),
});

// routes/order.js
Ember.Route.extend({
  actions: {
    saveOrder(order) {
      order.save();
    }
  }
});

// adapters/order.js
DS.Adapter.extend({
  createRecord(store, modelClass, snapshot) {
    snapshot.belongsTo('items').forEach(itemSnapshot => {
      // this would transition `itemSnaphot.record` and entangle `itemSnapshot` with 
      //`snapshot` such that when the order's record transitions away from in flight, 
      // so does the item's record
      itemSnapshot.savedWith(snapshot);
    });
  }
});
```

Does that make sense?

hjdivad:
Closing.

Of the three items this RFC is concerned with, two are handled by the merged [identifiers rfc](https://github.com/emberjs/rfcs/pull/403).

The remaining item is an API for co-mingling record lifecycles for multiple records with one promise (whether nested saves, or optimistic changes based on knowing the semantics of a particular API) which we'd prefer to do as a separate rfc.

PR #176, opened 2016-11-06
RFC: JavaScript Modules API

tomdale:
#### 12/13/2016

* [Conditionally enabling prototype extensions](https://github.com/emberjs/rfcs/pull/176#discussion_r92295467)

#### [11/16/2016](https://github.com/emberjs/rfcs/pull/176#issuecomment-261100221)

* Distribution & Versioning
* File Size
* Scoped Packages
* Templates
* Computed Properties
* Owner API
* Testing
* Formalization
* Generators
* Ember Data
* Churn / Complexity

#### 1/10/2017

- [Add Ember.LinkComponent](https://github.com/emberjs/rfcs/pull/176/commits/f25c7c3b90e0889db9f2c7963dcbe3a6fbfa66b3)

#### 12/13/2016

- [Note that this RFC doesn't preclude future additions, at @rwjblue's request](https://github.com/emberjs/rfcs/pull/176/commits/ed84cb755962b21dbd97990bcea240cbc22f8014)
- [Add more description on effectful/prototype extension modules](https://github.com/emberjs/rfcs/pull/176/commits/c021a91a562e21c75632802ac020ea0419d27588)

#### 11/16/2016

- [Add missing AutoLocation](https://github.com/emberjs/rfcs/pull/176/commits/13698de3344ed8498ec8349e58e9010cb1cdbb17)
- [Move `getOwner` and `setOwner` to `@ember/application` from `@ember/object/internal`](https://github.com/emberjs/rfcs/pull/176/commits/dc4fc8706a16816391518a4cbb856f09a32bf6a5), per [@tchak's suggestion](https://github.com/emberjs/rfcs/pull/176#issuecomment-258678662)
- [Add @zeppelin's previous module RFC, #68, to Prior Art section](https://github.com/emberjs/rfcs/pull/176/commits/c8283adfad47a308b10b42d1777b2d45aa9a7133)
- [Note that we need to update generators to the How We Teach This section](https://github.com/emberjs/rfcs/pull/176/commits/cea5cc30b4d2d06d62db9f5ffe77889cc7b4fe2a), per [@HeroicEric's suggestion](https://github.com/emberjs/rfcs/pull/176#issuecomment-259201703)
- [Add constraint around not deeply nesting modules](https://github.com/emberjs/rfcs/pull/176/commits/90cf471943c9a3d63de708ea9da1453a80bdd6d9), per [@nathanhammond's suggestion](https://github.com/emberjs/rfcs/pull/176#discussion_r87094024).
[...]

PR #178, opened 2016-11-18
Deprecate Ember.K

cibernox:
[Rendered](https://github.com/cibernox/rfcs/blob/deprecate-ember-k/text/0000-deprecate-ember-k.md)

### DIFF
--- a/text/0000-deprecate-ember-k.md
+++ b/text/0000-deprecate-ember-k.md
@@ -41,7 +41,7 @@ The second downside of reusing the same instance in many places is that if for
 some reason the VM deoptimizes that function, that deoptimization is spreaded
 across all the usages of `Ember.K`.
 
-Lastly, the chainable nature of `Ember.K` tends to surprise the users:
+Third, the chainable nature of `Ember.K` tends to surprise the users:
 
 ```js
 let derp = {
@@ -53,6 +53,27 @@ let derp = {
 derp.foo().bar().baz(); // O_o
 ```
 
+And lastly but more importantly for simplicity. Consider the following code:
+
+```js
+export default Component.extend({
+  onSubmit() {}
+});
+```
+
+Any JS developer will understand that this is an empty function and will probably understand
+that is a placeholder to provide your own function instead. However, JS developers that come
+across `Ember.K` for the first time will se this:
+
+
+```js
+export default Component.extend({
+  onSubmit: Ember.K
+});
+```
+
+and will think that it is some cryptic Ember magic that they have to learn.
+
 # Transition Path
 
 The necessary first step is to make sure Ember, Ember Data and other pieces of the
@@ -61,8 +82,7 @@ ecosystem don't use `Ember.K` internally.
 Phased approach:
 * Deprecate `Ember.K`: Use the deprecation API to signal the deprecation, and deprecation guide entry.
 * Add rule to ember-watson
-* Extract to addon. Precedence: `active-model-adapter`, `store.filter`
-* Do not include in #176.
+* Do not include export path in #176.
 
 # How We Teach This
 
--- a/text/0000-deprecate-ember-k.md
+++ b/text/0000-deprecate-ember-k.md
@@ -4,13 +4,14 @@
 
 # Summary
 
-The `Ember.K` utility function was a low level utility that has lost most of its value today.
+The `Ember.K` utility function is a low level utility that has lost most of its value today,
+like the wisdom tooth of the frameworks were we only think of it when it gives us problems.
 
 # Motivation
 
-Let's start with what is `Ember.K`.
+Let's start explaining what is `Ember.K`.
 
-It is this:
+It is this function:
 
 ```js
 Ember.K = function() {
@@ -18,7 +19,7 @@ Ember.K = function() {
 }
 ```
 
-The purpose of this function was to avoid some amount of boilerplate code
+The purpose of this utility was to avoid some amount of boilerplate code
 and limit the creation of function instances in Ember's internals.
 
 In a world of globals, effectively writing `somefn: Ember.K` was shorter
@@ -29,11 +30,12 @@ someFn: function() {
   return this;
 }
 ```
+and generated less function alocations.
 
 However with the introduction of ES6 modules and the modularization of Ember
-in process, keeping this feature would require to design an import path for it.
+in process (#176), keeping this feature would require to design an import path for it.
 
-While doable, any possible benefit in size or allocations obtained by reusing
+While doable, any possible benefit in code size or allocations obtained by reusing
 the exact same Function instance everywhere would be greatly smashed by the
 overhead both in space and CPU cycles of importing it in AMD-transpiled modules.
 It's worth noting that ES6 shorthand method syntax has made writing functions
@@ -53,16 +55,16 @@ let derp = {
   baz: Ember.K
 }
 
-derp.foo().bar().baz();
+derp.foo().bar().baz(); // O_o
 ```
 
 # Transition Path
 
 The obvious first step is to make sure Ember, Ember-data and other pieces of the
-ecosystem don't use `Ember.K.`.
+ecosystem don't use `Ember.K.` internally.
 
-The suggested path is intentionally not give `Ember.K` an import path in the new JS modules
-being discussed in #176, or give it a import path that is clearly privated.
+The suggested transition is to *intentionally* not give `Ember.K` an import path in the new JS modules
+being discussed in #176 or give it a import path that is clearly private.
 
 Part of that RFC states that some sort of shim mode will allow people to keep using
 `import Ember from 'ember';` for a while. Is in that shim where we could put a deprecation:
@@ -75,7 +77,7 @@ Ember.K = deprecatedVersionOfEmberK.
 ```
 
 As people migrate to the ES6 modules they will have to update their code since `Ember.K`
-cannot be imported without using the shim.
+cannot be used without using the shim.
 
 Perhaps the codemod that will perform this transition should be aware of this case so usages of `Ember.K`
 are replaced by the import from a private path or maybe go one step beyond and replace
--- a/text/0000-deprecate-ember-k.md
+++ b/text/0000-deprecate-ember-k.md
@@ -132,6 +132,17 @@ Example usage:
 npm install -g ember-k-codemod && ember-k-codemod --empty
 ```
 
+Versions of [ember-watson](https://github.com/abuiles/ember-watson) starting in `0.8.5` wrap this
+codemod so you can achieve the same transformation with it:
+
+```
+ember-watson remove-ember-k --empty
+
+// or if installed as an addon
+
+ember watson:remove-ember-k --empty
+```
+
 ## Addendum 2 - `Ember.K` usage across published addons
 
 ```
--- a/text/0000-deprecate-ember-k.md
+++ b/text/0000-deprecate-ember-k.md
@@ -115,8 +115,21 @@ https://github.com/cibernox/ember-k-codemod
 
 To use it you can install it globally and invoke the command on any app or addon.
 
+The commands **requires** the user to decide the approach to replace occurenced of `Ember.K`. The
+possible flags are `--empty` and `--return-this`.
+
+- `--empty` replaces `Ember.K` with an empty function. This leads to the most idiomatic and
+  intention-revealing code, but does not allow chaining, like the original `Ember.K` did.
+  Despite of that, chaining `Ember.K` was such an uncommon patterns that we thing virtually
+  everybody can use this option.
+- `--return-this` replaces `Ember.K` with a function that just returns `this`. This allows chaining
+  like the original one.
+
+
+Example usage:
+
 ```
-npm install -g ember-k-codemod && ember-k-codemod
+npm install -g ember-k-codemod && ember-k-codemod --empty
 ```
 
 ## Addendum 2 - `Ember.K` usage across published addons
--- a/text/0000-deprecate-ember-k.md
+++ b/text/0000-deprecate-ember-k.md
@@ -81,8 +81,9 @@ ecosystem don't use `Ember.K` internally.
 
 Phased approach:
 * Deprecate `Ember.K`: Use the deprecation API to signal the deprecation, and deprecation guide entry.
+  Target version will be 3.0, as usual.
 * Add rule to ember-watson
-* Do not include export path in #176.
+* Do not include export path in #176, but include it until 3.0 in the "globals" build.
 
 # How We Teach This
 
--- a/text/0000-deprecate-ember-k.md
+++ b/text/0000-deprecate-ember-k.md
@@ -4,8 +4,7 @@
 
 # Summary
 
-The `Ember.K` utility function is a low level utility that has lost most of its value today,
-like the wisdom tooth of the frameworks were we only think of it when it gives us problems.
+The `Ember.K` utility function is a low level utility that has lost most of its value today.
 
 # Motivation
 
@@ -30,17 +29,14 @@ someFn: function() {
   return this;
 }
 ```
-and generated less function alocations.
+and generated less function allocations.
 
 However with the introduction of ES6 modules and the modularization of Ember
 in process (#176), keeping this feature would require to design an import path for it.
 
-While doable, any possible benefit in code size or allocations obtained by reusing
-the exact same Function instance everywhere would be greatly smashed by the
-overhead both in space and CPU cycles of importing it in AMD-transpiled modules.
-It's worth noting that ES6 shorthand method syntax has made writing functions
-less verbose and those functions gain an internal name property that shows in
-stack traces making debugging easier.
+While doable, the transpiled output is actually bigger then defining the functions
+inline, specially with the ES6 shorthand method syntax, and the perf difference
+of saving a few function allocations is despicable.
 
 The second downside of reusing the same instance in many places is that if for
 some reason the VM deoptimizes that function, that deoptimization is spreaded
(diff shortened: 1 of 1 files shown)


vvscode:
Never use chaining with `Ember.K`. In my experience - it's  an alternative to non-existing `nop` ( or I should "implement" it in each project? ) at places where function should be, but I'm not sure about implementation in children, and won't check `typeof` / use `tryRun` - so didn't see any pros ( or lets add `nop` operation instead of )

kimroen:
I remember coming over this function the first time in this blog post, which was published in July, 2014: ["What the hell is Ember.K?"](http://mcdowall.info/what-the-hell-is-ember-k/).

The description of what it is useful for in that blog post is summarized in these bullets:

> - Making methods safe to call in the event that they might be undefined at the time of calling.
> - Providing hooks for subclasses to implement in a safe way.
> - Creating stand in action event handlers until you are ready to implement the necessary logic, and avoid Ember.js throwing an error.

So essentially, temorary code or hooks for subclasses.

Is this still useful? Is the suggested change in this RFC an acceptable replacement for `Ember.K`? I think it is. There are other ways of accomplishing the same goals.

text/0000-deprecate-ember-k.md
```diff
@@ -0,0 +1,105 @@
...
+While doable, any possible benefit in code size or allocations obtained by reusing
+the exact same Function instance everywhere would be greatly smashed by the
+overhead both in space and CPU cycles of importing it in AMD-transpiled modules.
+It's worth noting that ES6 shorthand method syntax has made writing functions
+less verbose and those functions gain an internal name property that shows in
+stack traces making debugging easier.
+
+The second downside of reusing the same instance in many places is that if for
+some reason the VM deoptimizes that function, that deoptimization is spreaded
+across all the usages of `Ember.K`.
```
mixonic: I've never seen this when profiling. Is it a real-world issue you've seen?

cibernox: @rwjblue mentioned it in slack and [here](https://github.com/emberjs/rfcs/pull/176#discussion_r88422065) an some smart fellows 👍 'd it

mixonic: ...but is it a real-world thing? Or just an idea that it might sometimes be an issue?

I'd like to keep RFCs rigorous by including, as justification for a decision, things which we know to be true when possible. There are some constraints that are based on intuition (does this API feel good) and some that are analytic. Performance is almost always an analytic constraint we can understand and test.

I'm wary of an off comment by @rwjblue being cargo-culted into a learned fact. We should be careful in RFCs we don't perpetuate assumption.

text/0000-deprecate-ember-k.md
```diff
@@ -0,0 +1,105 @@
...
+}
+```
+and generated less function alocations.
+
+However with the introduction of ES6 modules and the modularization of Ember
+in process (#176), keeping this feature would require to design an import path for it.
+
+While doable, any possible benefit in code size or allocations obtained by reusing
+the exact same Function instance everywhere would be greatly smashed by the
+overhead both in space and CPU cycles of importing it in AMD-transpiled modules.
```
cibernox: Code wise the transpiled output is bigger. 

Perhaps I was overconfident  that the overhead of adding one extra import (the more imports the more work `this.reify()` has to do to create the `this.reified`) will overcome function allocations without a proof, but I can run some benchmarks to check it.

mmun:
I haven't used it in 3 years. :+1: for removal.

martndemus:
I  am 💯% fine with killing this.

Let's also celebrate that we're having an RFC for deprecating a three line 'no-op' function 🎉

simonihmig:
I have actually used it and still using it in a number of apps. Almost exclusively for noop default actions in components, in case the component consumer does not provide its own (closure) action, so I can safely do `this.get('onSomething')();` without having to put this in a guard (so to not get an `undefined is not a function` exception).

I can see different styles of guards against this, like inline (https://github.com/cibernox/ember-power-select/blob/master/addon/components/power-select.js#L324-L327) or using a utility function for that (https://github.com/offirgolan/ember-light-table/blob/master/addon/utils/call-action.js). 

I am happy to learn if that is not considered a good style for whatever reason, but until then just declaring the action as `noop` (when importing `Ember.K` as `noop`) and simply calling it straight away seemed to me the style with the least boilerplate code and good readability. 

Redefining `noop` with `() => {}` (never used it for chaining) is certainly no big deal, but just keeping `Ember.K` is not either. And until now I thought reducing function allocations would be a good thing. 

On that matters, is the function allocations vs. import overhead argument still valid if rollup.js style module inlining should eventually land in ember-cli?

cibernox:
@simonihmig I ran some benchmarks with 100K imports/allocations and performance difference is negligible either way.
Transpiled code is mover vermose unless you use `Ember.K` a quite few times on the file, although in that case doing `const noop = ()=>{};` and reusing it is still shorter.

I'm going to shift the tone of the RFC away of the performance topic because it's not a big deal either way, and focusing it on removing from the public API low level utility that is trivial to replace.

workmanw:
> I'm going to shift the tone of the RFC away of the performance topic because it's not a big deal either way, and focusing it on removing from the public API low level utility that is trivial to replace.

I too use `Ember.K` a lot for base and interface classes. And while it would be simple for me to replace in my app with my own helper or just use inlined functions, if the purpose of this RFC is not performance, is it really necessary at all to deprecate it? To be perfectly honest, this is the kind of churn that so many people complain about. It's not a huge deal and it really doesn't effect anything, but now they have to learn to do something a little bit differently. They have to replace an old pattern with a new and things are exactly the same.

locks:
@workmanw 
> is it really necessary at all to deprecate it? To be perfectly honest, this is the kind of churn that so many people complain about.

The performance argument isn't the only one that can be done for a deprecation. Ember has been trying to slowly prune unnecessary APIs from the core library, and this is a good candidate due to being a very small, easily replaceable API. You might have noticed how FastBoot, Engines and more are mostly being done outside the core library :D I'm currently trying to gather some data so we have an idea of its usage, and will post an update when possible.

cibernox:
@workmanw 
> is it really necessary at all to deprecate it?

Obviously no, it's not the end of the world if we keep `Ember.K` around. However with ember deciding as we speak its future module API, it is the perfect time to remove or make private APIs in ember that are sort-of outdated. If there is a lot of pushback we can keep it. 

If this helps you feel better about the churn, there will be an ember-watson task to automatically remove usages of `Ember.K`, the very deprecation will point to the entry in the guides about how to do it and we can automatically create PRs for all ember-addons removing it.

If this goes ahead, it will be a very smooth process.

workmanw:
@cibernox Just to be clear, it's not only about modifying the existing code, it's also about learning a new way to do things. And again, I completely acknowledge this is a trivial change technically. IMO it's more about the casual user who doesn't follow all these RFCs, read every blog post, and isn't a full time ember dev. Sometimes they're project contracting folks, or backend guys floating to the frontend for a sprint. They try to use `Ember.K`, get a deprecation, then the go read about the change and figure out what the new pattern is. At the end of it all, if you can't demonstrate an upside to a change, it just looks like churn to a user.

locks:
> IMO it's more about the casual user who doesn't follow all these RFCs, read every blog post, and isn't a full time ember dev.

How do they come upon `Ember.K`?

> They try to use Ember.K, get a deprecation, then the go read about the change and figure out what the new pattern is.

That seems fine? There are still some improvements we could do to get the deprecation guides closer to the user.

workmanw:
> How do they come upon `Ember.K`?

Just something that has already been learned. Either by looking at other parts of the application, or by looking at an addon's implementation or even seeing it via debugging through Ember. 

I'm just trying to highlight that when something new replaces something old, which is often the case with an evergreen framework like Ember, if there isn't a clear benefit it can feel it can just feel like churn. Additionally, ember-watson is great for refactoring, but that is only one consequence of change. Relearning / retraining is another important consideration.

cibernox:
We'll run a search to see how widespread is the usage of `Ember.K`.

stefanpenner:
> if there isn't a clear benefit it can feel it can just feel like churn

Some quick thoughts:

* if removed we should be sure to provide a 1 step tool to remove this from code (the churn concern is real, we left this is the framework for so long, users have grown to use it).
* by not offering it in the framework, it is a small win in curriculum reduction.
* it is unlikely to pass the litmus test for a new feature today.
* alternative `foo: Ember.K`  -> `foo() { }` in classes or `() => ()`
* the chaining aspect of the K combinator has been a source of confusion, and unexpected bugs. For example someone uses `foo: Ember.K` the caller of `foo` can now chain, `foo` must then continue to always support chaining, even though it may not have been expected.
* quick note: as we work hard to make ES6 classes first class, this will continue to have less value.

text/0000-deprecate-ember-k.md
```diff
@@ -0,0 +1,102 @@
...
+However with the introduction of ES6 modules and the modularization of Ember
+in process (#176), keeping this feature would require to design an import path for it.
+
+While doable, the transpiled output is actually bigger then defining the functions
+inline, specially with the ES6 shorthand method syntax, and the perf difference
+of saving a few function allocations is despicable.
+
+The second downside of reusing the same instance in many places is that if for
+some reason the VM deoptimizes that function, that deoptimization is spreaded
+across all the usages of `Ember.K`.
```
stefanpenner: ya, if this happens. Large chunks of inlined code could be evicted, making many things sad. Although this risk does seem to continue to reduce as JITs get better, if we can avoid it I would recommend we do.

text/0000-deprecate-ember-k.md
```diff
@@ -0,0 +1,102 @@
...
+  bar: Ember.K,
+  baz: Ember.K
+}
+
+derp.foo().bar().baz(); // O_o
+```
+
+# Transition Path
+
+The obvious first step is to make sure Ember, Ember-data and other pieces of the
```
stefanpenner: ember-data I believe has some use-cases.

text/0000-deprecate-ember-k.md
```diff
@@ -0,0 +1,102 @@
...
+let derp = {
+  foo: Ember.K,
+  bar: Ember.K,
+  baz: Ember.K
+}
+
+derp.foo().bar().baz(); // O_o
+```
+
+# Transition Path
```
stefanpenner: When do we plan to drop? 3.0 or..

text/0000-deprecate-ember-k.md
locks: I have a [PR pending](https://github.com/emberjs/data/pull/4670) which might not even be necessary, as @runspired goes through the codebase.

stefanpenner: A wild @runspired unleashed upon a codebase you say?

stefanpenner: merged

text/0000-deprecate-ember-k.md
cibernox: Yes, as any regular feature. Although it will not have a export path in ember-core/source, but it will be present in the bundled version until 3.0

text/0000-deprecate-ember-k.md
```diff
@@ -0,0 +1,395 @@
...
+plaid/addon/mixins/dimensions.js:  didMeasureDimensions: Ember.K,
+plaid/addon/mixins/global-resize.js:  didResize: Ember.K
+spree-ember-paypal-express/addon/services/paypal-express.js:  spree: Ember.K,
+torii/addon/services/torii-session.js:  setUnknownProperty: Ember.K,
+torii/tests/unit/redirect-handler-test.js:    close: Ember.K
+torii/tests/unit/services/popup-test.js:    focus: Ember.K,
+torii/tests/unit/services/popup-test.js:    close: Ember.K
+```
+
+## Addendum 2 - `Ember.K` usage via destructuring across published addons
```
locks: Forgot to change ;P

text/0000-deprecate-ember-k.md
```diff
@@ -0,0 +1,395 @@
...
+```
+
+and generated fewer function allocations.
+
+However with the introduction of ES6 modules and the modularization of Ember
+in process (#176), keeping this feature would require to design an import path for it.
+
+While doable, the transpiled output is actually bigger then defining the functions
+inline, specially with the ES6 shorthand method syntax, and the perf difference
+of saving a few function allocations is despicable.
```
cibernox: It's a false-friend word I suppose, I meant negligible. Fixing it

mixonic:
@locks I think you're requested changes are still blocking.

@workmanw I'm surprised that you aren't convinced by the idea that `Ember.K` is a hurdle for new developers learning the framework. I've worked with a number who found this to be a massive gotcha, or even worse something they cargo culted through their codebases without understanding it at all. If you saw it in another framework you went to learn, I expect you would be befuddled at what the authors were thinking, right?

We're always trying to balance support for current users with making the framework as attractive as possible for new users. 💯. Part of this is to have a multi-pronged approach for existing users:

* Communicate changes about recommended patterns in the blog posts as they change.
* Permit users to upgrade at their own pace by providing LTS releases (upgrades only twice a year).
* Allow users to address changes at their own pace by permitting them to [ignore whatever deprecations they want](https://github.com/mixonic/ember-cli-deprecation-workflow) and ensuring most releases (all minor releases) are backwards compatible.
* Publish a [deprecation guide](http://emberjs.com/deprecations/) explaining the motivation for a deprecation and the way to refactor around it.
* Attempt to automate as much of the upgrade process as we can via [ember-watson](https://github.com/abuiles/ember-watson) (and this is something we can get much better at).

You're definitely a contributor who surfs the cutting edge of Ember, and as such you likely perceive change to happen more rapidly than other users. I've worked as plenty of shops that have deprecations from Ember 2.2 they still haven't bothered to look at even as they upgrade to 2.10.

Change is inevitable in software. If Ember doesn't improve, we all end up rewriting in something else anyway. The tradeoff our community strives for is constant improvement with as much process and tooling as possible to pick-your-own-pace.

I want to address your concerns, but they also seem pretty vague. Perhaps there is another discussion we can have some time about ways to better the communication for deprecations, especially around their motivation. I think for sure we should expect the deprecation guide PR to explain motivation.

In general I'd like to advance this RFC and go into FCP sometime this week. I've added it to the core team agenda for Friday, so it will be discussed then if no sooner.

workmanw:
@mixonic I was just trying to add a little friction to this discussion because I think `Ember.K` is unique situation in that it's really does nothing. Unlike virtually every feature ever, there is some variable amount of on going maintenance and that is often valid justification for removal of cruft. This case is very unique in that it doesn't rely on any other code in the framework at all. Nor can its behavior be debatable. 

Following through this RFC, before my initial comment, it seemed to be about performance. Which in general is a great reason to remove cruft. But then it kind of shifted from being about performance to being about removing something that was a no-op. That's when I wanted to raise my hand and ask if it was even beneficial to do this given that it will create deprecation warnings throughout the ecosystem and could seem like churn to some. 

> ... even worse something they cargo culted through their codebases without understanding it at all. 

Bahauaha. Too funny. I think it was more about the fact that it is already there, and it has been for a really long time, than it was about the value of it overall. My comments were not meant to be a last stand to save `Ember.K`. My sole purpose of piping up was to just to call out that many casual users I know and talk to (folks who don't read the RFCs, don't write ember full time and don't hit every upgrade) often express some frustration over little changes that seem to offer no benefits to them. It can make things feel unstable when lots of little things change and there is no perceived benefit.

mixonic:
@cibernox Thanks for the final edits regarding `--empty` and `--return-this`. That change addresses the last open issue I know of here, so I'm going to move this RFC into FCP today 🎉 

@workmanw I think we're pretty aligned in our goals. Removing `K` is definitely a simplification, and though simplification changes are a win for new users, they have a higher cognitive cost to existing users. This RFC and the codemod provide tools to help lower that cost. Good deprecation guide content and messaging can further ease the impact. Thanks for raising the feedback! 👏

john-kurkowski:
> though simplification changes are a win for new users, they have a higher cognitive cost to existing users

There are new users to Ember.js **and** new users to `Ember.K`, i.e. they develop in Ember.js but never bothered to learn this API morsel. I think this identifies even more who would benefit from the deprecation.

Our team has Ember.js code going back almost 4 years. Somewhere in history, there's sure to be sprinkles of `Ember.K` that seemed like a good fit at the time. More recently, we prefer native syntax. If you know JavaScript, the native version is immediately grokkable. No RTFS. Those un/familiar with `Ember.K` both benefit.

If users new to `Ember.K` go back in our code history and encounter its deprecation, that's a relief. It's apparent at a glance that it was "a massive gotcha." It generally remains one less thing they have to keep in their heads. They follow the one-time instructions how to replace `Ember.K`, and move on with their lives.

mixonic:
As there have been no issues raised in the FCP period that this RFC does not address, we will consider it accepted. I've rebased the commits down and merged into master: [rfc commit](https://github.com/emberjs/rfcs/commit/322499eed4ade5013122a9127355ee1c485be097), [merge commit](https://github.com/emberjs/rfcs/commit/76fa822568c87be19b914c2e49ae20d86b2c8137), [commit renaming the file](https://github.com/emberjs/rfcs/commit/9319517daea5ea85e98a71c00d459ff05e8b48e3).

🎉  Thank you for your input and discussion! 🎉

bstro:
Would love to read the RFC but it looks like the [link is broken](https://github.com/cibernox/rfcs/blob/deprecate-ember-k/text/0000-deprecate-ember-k.md)

locks:
@bstro the merged RFC is at https://github.com/emberjs/rfcs/blob/master/text/0178-deprecate-ember-k.md.

PR #186, opened 2016-12-05
RFC: Track unique history location state

bcardarella:
[rendered](https://github.com/bcardarella/rfcs/blob/0ad4c1360f77b6db637417c1f97fa79239e29ecb/text/0000-track-unique-history-location-state.md)

### DIFF
--- a/text/0000-track-unique-history-location-state.md
+++ b/text/0000-track-unique-history-location-state.md
@@ -12,7 +12,7 @@ The path alone does not provide enough information. For example, if you
 visit page A, scroll down, then click on a link to page B, then click on
 a link back to page A. Your actual browser history stack is [A, B, A].
 Each of those nodes in the history should have their own unique scroll
-position. In order to record this position we need an ID that is unique
+position. In order to record this position we need a UUID 
 for each node in the history.
 
 This API will allow other libraries to reflect upon each location to
@@ -29,8 +29,11 @@ properly based upon where you are in the history stack, as described in
 
 Code: [PR#14011](https://github.com/emberjs/ember.js/pull/14011)
 
-We simply add a `stateCounter` so we can track uniqueness on two
-dimensions. Both `path` and the generated `id`.
+We simply unique identifier (UUID) so we can track uniqueness on two
+dimensions. Both `path` and the generated `uuid`. A simple UUID
+generator such as
+https://gist.github.com/lukemelia/9daf074b1b2dfebc0bd87552d0f6a537
+should suffice.
 
 # How We Teach This
 
@@ -41,29 +44,29 @@ history stack. For example, it could look like:
 // visit page A
 
 [
-  { path: '/', id: 1 }
+  { path: '/', uuid: 1 }
 ]
 
 // visit page B
 
 [
-  { path: '/about', id: 2 },
-  { path: '/', id: 1 }
+  { path: '/about', uuid: 2 },
+  { path: '/', uuid: 1 }
 ]
 
 // visit page A
 
 [
-  { path: '/', id: 3 },
-  { path: '/about', id: 2 },
-  { path: '/', id: 1 }
+  { path: '/', uuid: 3 },
+  { path: '/about', uuid: 2 },
+  { path: '/', uuid: 1 }
 ]
 
 // click back button
 
 [
-  { path: '/about', id: 2 },
-  { path: '/', id: 1 }
+  { path: '/about', uuid: 2 },
+  { path: '/', uuid: 1 }
 ]
 ```
 
--- a/text/0000-track-unique-history-location-state.md
+++ b/text/0000-track-unique-history-location-state.md
@@ -1,0 +1,93 @@
+- Start Date: 2016/12/05
+- RFC PR: (leave this empty)
+- Ember Issue: (leave this empty)
+
+# Summary
+
+Track unique history location states
+
+# Motivation
+
+The path alone does not provide enough information. For example, if you
+visit page A, scroll down, then click on a link to page B, then click on
+a link back to page A. Your actual browser history stack is [A, B, A].
+Each of those nodes in the history should have their own unique scroll
+position. In order to record this position we need an ID that is unique
+for each node in the history.
+
+# Detailed design
+
+Code: [PR#14011](https://github.com/emberjs/ember.js/pull/14011)
+
+We simply add a `stateCounter` so we can track uniqueness on two
+dimesions. Both `path` and the generated `id`.
+
+This API will allow other libraries to reflect upon each location to
+determine unique state. For example,
+[ember-router-scroll](https://github.com/dollarshaveclub/ember-router-scroll)
+is making use of a [modified `Ember.HistoryLocation` object to get this
+behavior](https://github.com/dollarshaveclub/ember-router-scroll/blob/master/addon/locations/router-scroll.js).
+
+Tracking unique state is required when setting the scroll position
+properly based upon where you are in the history stack, as described in
+[Motivation](#motivation)
+
+# How We Teach This
+
+We could describe what meta data is generated for each location in the
+history stack. For example, it could look like:
+
+```js
+// visit page A
+
+[
+  { path: '/', id: 1 }
+]
+
+// visit page B
+
+[
+  { path: '/about', id: 2 },
+  { path: '/', id: 1 }
+]
+
+// visit page A
+
+[
+  { path: '/', id: 3 },
+  { path: '/about', id: 2 },
+  { path: '/', id: 1 }
+]
+
+// click back button
+
+[
+  { path: '/about', id: 2 },
+  { path: '/', id: 1 }
+]
+```
+
+# Drawbacks
+
+None that I'm aware of
+
+# Alternatives
+
+The purpose for this behavior is to enable scroll position libraries.
+There are two other solutions in the wild. One is in the guides that
+suggests resetting the scroll position to `(0, 0)` on each new route
+entry. The other is
+[ember-memory-scroll](https://github.com/ef4/memory-scroll) which I
+believe is better suited for tracking scroll positions for components
+rather than the current page.
+
+However, in both cases neither solution provides the experience that
+users have come to expect from server-rendered pages. The browser tracks
+scroll position and restores it when you revisit the page in the history
+stack. The scroll position is unique even if you have multiple instances
+of the same page in the stack.
+
+# Unresolved questions
+
+Is this public API? If so, would this be considered additive or
+breaking?
--- a/text/0000-track-unique-history-location-state.md
+++ b/text/0000-track-unique-history-location-state.md
@@ -15,13 +15,6 @@ Each of those nodes in the history should have their own unique scroll
 position. In order to record this position we need an ID that is unique
 for each node in the history.
 
-# Detailed design
-
-Code: [PR#14011](https://github.com/emberjs/ember.js/pull/14011)
-
-We simply add a `stateCounter` so we can track uniqueness on two
-dimesions. Both `path` and the generated `id`.
-
 This API will allow other libraries to reflect upon each location to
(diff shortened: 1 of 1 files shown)


text/0000-track-unique-history-location-state.md
```diff
@@ -0,0 +1,93 @@
...
+Each of those nodes in the history should have their own unique scroll
+position. In order to record this position we need an ID that is unique
+for each node in the history.
+
+# Detailed design
+
+Code: [PR#14011](https://github.com/emberjs/ember.js/pull/14011)
+
+We simply add a `stateCounter` so we can track uniqueness on two
+dimesions. Both `path` and the generated `id`.
```
stefanpenner: dimesions   -> dimensions

text/0000-track-unique-history-location-state.md
```diff
@@ -0,0 +1,93 @@
+- Start Date: 2016/12/05
+- RFC PR: (leave this empty)
+- Ember Issue: (leave this empty)
+
+# Summary
+
+Track unique history location states
```
stefanpenner: We should likely outline the public/intimate API that plugins (or consumers) will use to utilize this feature. This contract/protocol should be clearly defined, as it will be sticking around for some time.

text/0000-track-unique-history-location-state.md
```diff
@@ -0,0 +1,93 @@
...
+a link back to page A. Your actual browser history stack is [A, B, A].
+Each of those nodes in the history should have their own unique scroll
+position. In order to record this position we need an ID that is unique
+for each node in the history.
+
+# Detailed design
+
+Code: [PR#14011](https://github.com/emberjs/ember.js/pull/14011)
+
+We simply add a `stateCounter` so we can track uniqueness on two
```
stefanpenner: * the pr seems to add `_stateCounter`not `stateCounter`
* `_stateCounter` seems abit strange, to provide an intentional private looking property as public API. What is the reasoning behind this?
* where does this property live?
* is it readOnly access or read/write property

text/0000-track-unique-history-location-state.md
```diff
@@ -0,0 +1,93 @@
...
+dimesions. Both `path` and the generated `id`.
+
+This API will allow other libraries to reflect upon each location to
```
stefanpenner: this seem motivation, not detailed design

text/0000-track-unique-history-location-state.md
```diff
@@ -0,0 +1,93 @@
...
+determine unique state. For example,
+[ember-router-scroll](https://github.com/dollarshaveclub/ember-router-scroll)
+is making use of a [modified `Ember.HistoryLocation` object to get this
+behavior](https://github.com/dollarshaveclub/ember-router-scroll/blob/master/addon/locations/router-scroll.js).
+
+Tracking unique state is required when setting the scroll position
```
stefanpenner: this seem motivation, not detailed design

text/0000-track-unique-history-location-state.md
```diff
@@ -0,0 +1,93 @@
...
+
+[
+  { path: '/about', id: 2 },
+  { path: '/', id: 1 }
+]
+```
+
+# Drawbacks
+
+None that I'm aware of
```
stefanpenner: * the property access is writable
* the `_` property queues private, but we want to enable add-ons to publicly use this, this seems like a teaching hazard.
* no scroll position between tabs/sessions (likely fine though)

text/0000-track-unique-history-location-state.md
```diff
@@ -0,0 +1,93 @@
...
+
+However, in both cases neither solution provides the experience that
+users have come to expect from server-rendered pages. The browser tracks
+scroll position and restores it when you revisit the page in the history
+stack. The scroll position is unique even if you have multiple instances
+of the same page in the stack.
+
+# Unresolved questions
+
+Is this public API? If so, would this be considered additive or
```
stefanpenner: If it is intended for an addon to use, i believe it is public API.

text/0000-track-unique-history-location-state.md
```diff
@@ -0,0 +1,93 @@
...
+  { path: '/', id: 1 }
+]
+
+// click back button
+
+[
+  { path: '/about', id: 2 },
+  { path: '/', id: 1 }
+]
+```
```
stefanpenner: if this is intended for add-on authors or custom location authors. We should likely describe in more detail how to teach them to use the feature.

text/0000-track-unique-history-location-state.md
stefanpenner: `stateCounter` i assume returns a number, do we wan to also guarantee that `stateCounter` is always incremented, such as `counterFor(state1) < counterFor(state2) < counterFor(stateN)`?

If it is guaranteed to be incrementing, is it incrementing for the life of the page, for the life of the app or for the life of the location.  (this is more something i thought of, when thinking about how acceptances tests might interact with one another.

Alternatively, we can simply say it is a session unique id for a given application instance, and in practice it could be represented by a number, or a guid or....

text/0000-track-unique-history-location-state.md
```diff
@@ -0,0 +1,93 @@
...
+
+# Drawbacks
+
+None that I'm aware of
+
+# Alternatives
+
+The purpose for this behavior is to enable scroll position libraries.
```
stefanpenner: If all we need is each state to have a unique identifier, could we just use a Map or WeakMap? I believe this may result in no public API change.

```js
const scrollPositions = new Map()

scrollPositions.set(state, [10, 10]);
scrollPositions.set(state2, [100, 100]);

scrollPositions.get(state) // => [10, 10]
scrollPositions.get(state2) // => [100, 100]
```

If we are worried about a memory leak, a WeakMap could be used instead. Which ember actually has a polyfill that would work for this use case kicking around: https://github.com/emberjs/ember.js/blob/master/features.json#L6

text/0000-track-unique-history-location-state.md
bcardarella: @stefanpenner the private name was requested by @rwjblue

bcardarella: > or a guid

an early implementation of the code used `Ember.guidFor` however the problem is the only value available to generate the guid was the path and if you visit the same path but as different parts of the stack it will generate similar guids, we want unique values for each new visit.

text/0000-track-unique-history-location-state.md
bcardarella: @stefanpenner the `_stateCounter` isn't meant to be accessed directly. The only thing that add authors will have access to and should use is the `{ path: '/foo', id: 3 }` object, in this case the `id` property being the new part of the object being proposed.

text/0000-track-unique-history-location-state.md
bcardarella: I'm all for that, beyond the the `id` property being accessible on the history object there isn't anything else being proposed. IMO the easiest example would be to point towards ember-router-scroll's implementation

text/0000-track-unique-history-location-state.md
stefanpenner: GuidFor of the state object would assign a hidden ID to it, then forever that object would return that hidden id. Which seems like this would work, as all we are doing is assigning our own custom ID to state

text/0000-track-unique-history-location-state.md
stefanpenner: I am not recommending we change the state object itself, But that we associated data (such as scroll position) with any given state pojo via map or weakmap.

This RFC seems to want to give unique identity to a given state pojo, so that other parts of the system can differentiate between them. As currently described, it suggests branding each pojo with an ID, but why brand at all if we can just use the objects own identity as a way to differentiate. A way to use an objects identity in such a way is a map, where one would use the state pojo as the key in the map. But that may make memory management tricky, hence recommending a weakmap instead.

The only thing I can think of that would prevent this from working, is if the state pojo we push onto the stack, isn't the same pojo (but still contains the same properties/values) we get when we pop off the stack.

stefanpenner: Actually, i take that back. WeakMap wont work, because `state` is serialized/deserialized which means when popping it, we get a copy of the original state back. This means, object identity can't be used as key. So branding with something like `__id__` or `id` as you proposed, seems like the correct path.

Sorry for taking you on an adventure, I did not realize that state object was serialized this way. In retro spec this makes lots of sense.

text/0000-track-unique-history-location-state.md
bcardarella: @krisselden this would be ideal, do you have any resources I can read upon for this?

bcardarella: @krisselden reading https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage Its not clear to me on how to track unique page visits. If I visit A, then B, then A with a history stack of [A, B, A] how would sessionStorage API be used? Would the page be the key? If yes, then that only have a single dimension of uniqueness and doesn't account for multiple page visits. (which this RFC addresses)

If the entire history object is to be stored in sessionState, is this something that should be restored at app launch? That would seem like an odd user experience.

bcardarella:
@stefanpenner thanks! I do think there might be an outstanding issue however, one that I haven't tested. I think the use case of navigating out of the app then back-button'ing to the app. Considering that history state is being pushed on the navigator level I suspect the data and this RFC should be OK. I suspect the implementation in https://github.com/dollarshaveclub/ember-router-scroll needs to be updated to catch this when instantiating

stefanpenner:
@bcardarella sorry for letting this linger. It slipped my mind.

Based on how you are serializing the state, I suspect that scenario should be fine. Although someone confirming it would be good.

stefanpenner:
I'll be championing this RFC at this Fridays core team meeting.

bcardarella:
🙌

nathanhammond:
[Can confirm that this works correctly with the `history.state` approach.](https://github.com/emberjs/rfcs/pull/186#issuecomment-271416805) (Was involved in this original spec.)

stefanpenner:
@bcardarella one last thing, @krisselden pointed this one.  A monotonically increasing identifier wont work correctly, as it will result in duplicate ID's being serialized as the counter will be reset every time we reload the page.

Instead of a monotonically incrementing counter, we could use the following `uuid` https://gist.github.com/lukemelia/9daf074b1b2dfebc0bd87552d0f6a537 snippet. Which will have sufficient entropy for the history state objects, and should mitigate the above concern.

With this change, we are comfortable moving this into FCP.

krisselden:
Also, if you close and reopen, the tab is restored with its deserialized session.  The counter will reset in this case even though the history states have been restored.

bcardarella:
@stefanpenner @krisselden thank you for the feedback. Should I reference Luke's snippet directly as the suggested `uuid` solution or leave it ambiguous?

stefanpenner:
> leave it ambiguous?

Specifically, calling out that it requires a sufficient unique identifier to be resilient to the points @krisselden raised.

nathanhammond:
We don't need a random ID for each time. One monotonically increasing identifier, and one generated GUID per session. Store both in the history state object.

stefanpenner:
> We don't need a random ID for each time. One monotonically increasing identifier, and one generated GUID per runtime session. Store both in the history state object.

Both approaches would work, but its not clear to me what the benefit of having two separate values would be.

This RFC aims to provide a way to disambiguate between serialized / deserialized history state items, a UUID accomplishes this without implying ordering (I fear the monotonically increasing value my imply something we do not wish to guarantee). We should defer ordering to the history implementation.

rwjblue:
Discussed again at today's Ember core team meeting. There have been no new issues brought up in the final comment period.

Landing....

nathanhammond:
The only reason I would propose an incremental ID + GUID pair is for reasons of performance and "sufficient" randomness. Either works.

PR #192, opened 2016-12-14
Query-Params Lifecycle

jamesarosen:
[Rendered](https://github.com/jamesarosen/rfcs/blob/qp-lifecycle/active/0000-qp-lifecycle.md)

### DIFF
--- a/active/0000-qp-lifecycle.md
+++ b/active/0000-qp-lifecycle.md
@@ -1,0 +1,88 @@
+- Start Date: 2016-12-14
+- RFC PR: (leave this empty)
+- Ember Issue: (leave this empty)
+
+# Summary
+
+Formalize the lifecycle of query-params.
+
+# Motivation
+
+The [guides for query-params](https://guides.emberjs.com/v2.10.0/routing/query-params/)
+discuss at length how to set query-params, including customizations. They do
+not, however, formally specify the _lifecycle_ of query-params. Specifically,
+the following are left unspecified:
+
+ * at what point in a transition are query-params synced from the Location
+   to the Controller?
+ * at what point in a transition are query-params synced from the Controller
+   to the Location?
+
+[This ember-twiddle](https://ember-twiddle.com/fee6f09acd62b0b567378b81dc76d4fc)
+demonstrates some oddities:
+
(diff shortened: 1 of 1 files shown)


jamesarosen:
Related: emberjs/ember.js#9819, emberjs/ember.js#11592. 

See also http://stackoverflow.com/q/35192211/1190 (which works in an action handler, but not in a route hook like `afterModel`)

jamesarosen:
I've done some monkey-patching to fix some of these things. In Ember 1.13, I had the following to ensure that all query-params were synced 
```js
// route:application
serializeQueryParam(value, key, ...args) {
  value = value || this.controllerFor('application').get(key);
  return this._super(value, key, ...args);
},

model() {
  const controller = this.controllerFor(this.routeName);
  controller.getWithDefault('queryParams', []).forEach(function(key) {
    if (controller.get(key) != null) return;
    controller.set(key, transition.queryParams[key]);
  });
}
```
But that broke on Ember 2.x due to emberjs/ember.js#11592. I changed the `model` hook to
```js
model(params, { queryParams }) {
  return queryParams;
},

afterModel(model, transition) {
  transition.send('fixQueryParams', transition.queryParams);
},

setupController(controller, model) {
  model = _.pick(model, [ 'valid', 'queryParams' ]);
  return this._super(controller, model);
},

actions: {
  fixQueryParams(queryParams) {
    let needsTransition = false;
    // do some munging of queryParams; if any munging, set needsTransition
    if (needsTransition) {
      Ember.run.next(this, 'replaceWith', { queryParams });
    }
  }
}
```
I couldn't do the `replaceWith({ queryParams })` in `afterModel` because it blew up with an error about `handlerInfos[handlerInfos.length - 1]` being `null`.

knownasilya:
cc @nathanhammond

rwjblue:
I think it is great to have this use case in mind and to start fleshing out what the constraints are, but this seems more like it belongs as an issue while we flesh it out into an actual concrete proposal.

Thoughts?

jamesarosen:
> this seems more like it belongs as an issue

No objections. I'll reframe it as an issue on emberjs/ember.js and reference this RFC.

jamesarosen:
Replacing with emberjs/ember.js#14785 for now.

PR #194, opened 2016-12-26
Deprecate component `eventManager`

GavinJoyce:
[Rendered](https://github.com/GavinJoyce/rfcs/blob/c0106e2051afa06e97bf86cfb8a6b9ba94ae4f36/text/0000-deprecate-custom-event-manager.md)

---

Part of https://github.com/emberjs/ember.js/issues/14754

### DIFF
--- a/text/0000-deprecate-custom-event-manager.md
+++ b/text/0000-deprecate-custom-event-manager.md
@@ -1,0 +1,38 @@
+- Start Date: (fill me in with today's date, YYYY-MM-DD)
+- RFC PR: (leave this empty)
+- Ember Issue: (leave this empty)
+
+# Summary
+
+Support for component `eventManger`s is a seldom used feature and should
+be deprecated.
+
+# Motivation
+
+We should strive to simplify the Ember API and source code where possible. As
+the custom `eventManager` feature is rarely used in apps, we should deprecate
+it.
+
+# Detailed design
+
+We'll introduce a deprecation warning which will be displayed when a component
+defines an `eventManager` property or when `canDispatchToEventManager` is set to
+true on `EventDispatcher`. The warning will have a target version of `3.0`.
+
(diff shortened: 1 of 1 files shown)


rwjblue:
Awesome, thank you for putting this together! 

We should also have this deprecation cover the `canDispatchToEventManager` property of `EventDispatcher` (mostly because the option does nothing at that point and is misleading) so that when this lands defining an `eventManager` on any component **or** specifying `canDispatchToEventManager` trigger a deprecation.

kellyselden:
I haven't used this feature before, but if I did, what would be the alternative to using it? (how do I un-deprecate my code)

GavinJoyce:
@kellyselden you could register an extended `EventDispatcher` which implements the soon to be removed `_findNearestEventManager` in event handlers. I'm happy to create an addon for this if anyone requests it so that it's a simple opt-in.

text/0000-deprecate-custom-event-manager.md
```diff
@@ -0,0 +1,38 @@
...
+it.
+
+# Detailed design
+
+We'll introduce a deprecation warning which will be displayed when a component
+defines an `eventManager` property or when `canDispatchToEventManager` is set to
+true on `EventDispatcher`. The warning will have a target version of `3.0`.
+
+If required, we can create an addon which extends the `EventDispatcher` allowing
+for opt-in custom `eventManager`s in Ember apps.
```
GavinJoyce: if anyone has a need for this, let me know and I'll create the addon

kellyselden:
@GavinJoyce Addon route seems like a good idea to me.

rwjblue:
Sorry for the delay here. We discussed in the core team meeting on 2017-01-20, and are 👍 on moving this to final comment period.

mixonic:
I'm merging this, as we are certainly past the one-week window and I see no new comments 😗 

Thank you!

PR #204, opened 2017-01-24
RFC: Template import directives

martndemus:
[Rendered](https://github.com/martndemus/rfcs/blob/102edbec6d67be8479375a8682c3c258f034df61/text/0000-template-import-directives.md)

This RFC might need some editing + expanding. But I'm positive it's
good enough to get some good discussion going and then go from there.

The TL;DR version:
- Standardize on a comment format as hinting to static analysis tools.
- This will allow tools to tree shake your components/helpers + templates

### DIFF
--- a/text/0000-template-import-directives.md
+++ b/text/0000-template-import-directives.md
@@ -1,0 +1,85 @@
+- Start Date: (fill me in with today's date, YYYY-MM-DD)
+- RFC PR: (leave this empty)
+- Ember Issue: (leave this empty)
+
+# Summary
+
+To make static resolution of a dependency graph of components and helpers in
+Handlebars templates possible at build time.
+
+# Motivation
+
+RFC #176 proposes a version of Ember with a "first-class system for importing
+just the parts of the framework you need". Landing that RFC would make almost
+all modularized code in the Ember ecosystem statically resolvable. The exception
+being Handlebars templates where the `{{component}}` helper has been used.
+
+This RFC is seeking to provide a system that can resolve usage of any component
+and or helper for any given Ember application, by the means of the app author
+declaring each component used in a template up front.
+
+When implemented this RFC should unlock being able to tree-shake a full Ember
+application.
+
+As a secondary motivation, it should make it easier for an app author to trace
+where a component and/or helper originates from.
+
+# Detailed design
+
+The implementation will rely on a syntax analogue to the ECMAScript 2015 syntax
+for importing values from modules. Were in a JavaScript module importing another
+module looks like this:
+
+```js
+import Ember from "ember";
+import Analytics from "../mixins/analytics";
+
+export default Ember.Component.extend(Analytics, {
+  // ...
+});
+```
+
+The proposed Handlebars analogue to this is:
+
+```hbs
+{{!import ./post-editor}}
+
+<div>
+  <h1>Editing: {{post.title}}</h1>
+  {{post-editor post=post}}
+</div>
+```
+
+The important bit is `{{!import ./post-editor}}`. This is seen by the current
+Handlebars compiler as a comment and thus is ignored. The proposed
+implementation would add meaning to this comment as being a import directive.
+The choice for the import directives to be comments is a deliberate choice to
+make the implementation backwards compatible with older implementations of the
+Handlebars compiler.
+
+This new directive does not have any output in the compiled result of the
+template. It is only meant to be read by static-analysis tools.
+
+Breaking down the import directive, it should start with the word `import`
+followed by a path. This path can either be relative to the current
+component/helper or absolute.
+
+# How We Teach This
+
+This system should be an opt-in feature to the Ember ecosystem. Proper guides in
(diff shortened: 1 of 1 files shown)


Turbo87:
@martndemus is that stuff actually needed? I think we could extract a lot of the same information from the existing AST

martndemus:
@Turbo87 Yes, but not 100%. This is to make the 100%.

buschtoens:
Two questions instantly come to mind:

1. Why can't ember-cli detect `{{post-editor}}` as a component and thereby infer that this component is being used? Why the extra `import` statement.
2. What about dynamic components, i.e. `{{component "post-editor"}}` where `"post-editor"` could also be a variable string that is only known at runtime?

Clearly I must be missing something.

simonihmig:
Just wondering, if analysis is happening at build-time, and import statements implemented as comments, thus no effect on compilation, is there actually any change required to Ember itself? If not, no need for an RFC, right?

Turbo87:
> Yes, but not 100%

@martndemus can you explain what the edge cases are that can not be handled from the current AST? I know about `{{component foobar}}`, but I don't see how the current proposal would solve that case. I think it was discussed to used a fixed prefix for such cases and always include all the components starting with the prefix (e.g. a components subfolder).

cibernox:
Just one question. This kind of comments is only needed in templates that use `(component boundValue)`, correct?

locks:
@simonihmig yes, this has many implications to the way Ember works in multiple levels like the resolver, build pipeline, template parser, and it's definitely a significant new API that needs to be discussed by the community.

martndemus:
@cibernox In the case of a addon, I sometimes don't know beforehand which component I'm getting passed in, how would the restrict helper work in those situations?

Turbo87:
> In the case of a addon, I sometimes don't know beforehand which component I'm getting passed in, how would the restrict helper work in those situations?

@martndemus I might be missing something obvious, but how would the comments approach solve this?

martndemus:
@Turbo87 By importing what you know that will get used. I.E. if you know this component will cause a component from an addon to render a component you know of, then import it there.

Turbo87:
@martndemus isn't it your responsibility to pass that component in via `useThisComponent=(component "foobar")`? in that case it would already be known to the AST as well.

simonihmig:
@locks Thanks for trying to explain this, but I am still not sure I can follow. Maybe the proposal should make clear what implications this has for the public API of Ember? So what has to be unlocked in Ember before an addon could implement this? 

My understanding was that this has no effect on runtime behaviour, just enable build-time tree-shaking. So resolver does not have to know anything about this, or does it? Sure handlebars AST has to understand this, but this does not affect Ember as well?

martndemus:
@simonihmig It affects neither Handlebars AST nor Ember. The current proposal as-is only proposes to standardize tooling along a common format.

simonihmig:
@martndemus Thanks for clarification! This aligns with how I did understand this. So I am fine with everybody continuing that discussion, here or somewhere else, but technically an Ember RFC would not be required, right?

martndemus:
@simonihmig I have no clue what does and doesn't warrant an RFC anymore. AFAIK anything only slightly impactful to how someone would be writing Ember apps should be an RFC.

miguelcobain:
@Turbo87 and others, I think the host extreme example is `{{component componentName`. And imagine that `componentName` comes from a remote API.

There is nothing a static analysis can do to know which components will be rendered here. So it must assume that *any* component can be rendered. So, all the components accessible in the entire APP would need to be included.

This RFC proposes a way to keep these situations under control. I agree we might need something like this.

Turbo87:
> This RFC proposes a way to keep these situations under control. I agree we might need something like this.

where? I don't see anything that would solve this issue. maybe I'm blind... 🤔

miguelcobain:
@Turbo87 so, I think the idea is that the import directives give the compiler the missing information. Instead of considering the entire set of components available, the build process now just considers the "imported" components and not everything.

Turbo87:
how is that different from what @cibernox proposed in https://github.com/emberjs/rfcs/pull/204#issuecomment-275079266? I guess I just don't like using comments to actually have semantic meaning if we can use pretty much the same within the current AST constructs

locks:
@simonihmig the import statement is a thing you would put in an Ember application template. It's a new API, you're introducing a keyword to comments that has special semantics, if nothing else. "Build-time" is still Ember.

miguelcobain:
@Turbo87 both approaches would give us the same result, as far as I can tell. I'm not in favor of any approach yet, I'm just saying what I understood from the RFC and that I think something to solve the problem might be needed.

cibernox:
I think that my idea and this one would achieve the same goal, functionally speaking. It's just a matter of syntactic preference. Not that I've given a deep though to this tho, just came it naturally to mind.

rwjblue:
Yes, this RFC will be replaced by https://github.com/emberjs/rfcs/pull/454.

rwjblue:
Closing in favor of #454

PR #205, opened 2017-01-26
Allow for specifying epsilon segments in routes

adye20:
Hi. This is to allow for specifying of epsilon segments in the Ember Router DSL. This is for the ability to have a nameless route.

Reference constraint 2.1 in http://www.nathanhammond.com/route-recognizer for
more information about epsilon segments.


wagenet:
Given the inactivity here I'm going to go ahead and close this. If you would still like to try to move this forward, please let me know and I'll ensure that it gets a proper review.

PR #208, opened 2017-02-07
Contextual helpers

cibernox:
[rendered](https://github.com/cibernox/rfcs/blob/contextual-helpers/text/0000-contextual-helpers.md)

### DIFF
--- a/text/0000-contextual-helpers.md
+++ b/text/0000-contextual-helpers.md
@@ -1,16 +1,16 @@
-- Start Date: 2016-10-14
+- Start Date: 2016-02-7
 - RFC PR: (leave this empty)
 - Ember Issue: (leave this empty)
 
 # Summary
 
-The goal of this RFC is to allow for better helper composition by allowing helpers
+The goal of this RFC is to increase composability oportunities in ember by allowing helpers
 to be invoked and curried in a similar way as how actions and components can be dynamically
 be invoked now.
 
 # Motivation
 
-Since Ember 2.3, released in January of 2016, the framework has enjoyed a feature commonly
+Since Ember 2.3, released over a year ago, the framework has enjoyed a feature commonly
 referred a **contextual components** that allows components to be dynamically invoked by name
 and also to be wrapped along with arguments and passed as regular values to later on be invoked.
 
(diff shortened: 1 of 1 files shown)


text/0000-contextual-helpers.md
```diff
@@ -0,0 +1,122 @@
...
+Examples:
+
+```hbs
+<strong>My name is {{invoke-helper 'capitalize-str' user.name}}</strong>
+<button title={{invoke-helper 'capitalize-str' user.name}}></button>
+<strong>My name is {{helper 'capitalize-str' user.name}}</strong> <!-- Invalid invocation -->
+<button title={{helper 'capitalize-str' user.name}}></button> <!-- Invalid invocation -->
+
+{{my-component format-date=(helper "moment-format")}} <!-- Creates closure helper -->
+{{my-component format-date=(invoke-helper "moment-format")}} <!-- Invokes helper -->
```
NLincoln: What happens to the existing syntax for invoking helpers, which for this example would be:

```hbs
{{my-component format-date=(moment-format)}}
```

cibernox: that would invoke the helper which generates a string that is passed to the component, like today.

knownasilya:
Yes please, and then deprecate `{{my-helper}}` so we don't have that ambiguity.. is it a value, component, or helper?

miguelcobain:
It can be any of those, but there are priorities in its resolution. value, helper and then component, I think.

knownasilya:
Yes, but from a developer point of view, it's ambiguous. Especially for developers new to Ember.

miguelcobain:
A bit unrelated to the RFC, I guess, but I'm not sure how to solve that.

machty:
There's an inconsistency in helper lifespans that I've noticed in a few other cases that we might want to consider as part of this RFC: https://ember-twiddle.com/d869daef1f0c2c8e2808a8e60dc06ca3?numColumns=2&openFiles=templates.application.hbs%2Chelpers.count-up.js

ming-codes:
An helper is mostly just a function. Can the problem be solved by simply yielding a function (or action)?

```hbs
{{yield
  (action "capitalize-str")
}}
```

```hbs
{{#parent-component as |capitalize-str|}}
  {{child-component helper=capitalize-str}}
{{/parent-component}}
```

Only down side I see is that you can't create a helper that maintain it's own state.

cibernox:
@ming-codes It could, and [it's what I'm doing for now](https://github.com/cibernox/ember-power-datepicker/blob/master/addon/templates/components/power-datepicker.hbs#L39), but there is no built in way of invoking that helper in the template, so I have to use ember-composable-helper's [compute](https://github.com/DockYard/ember-composable-helpers#compute) utility.

This proves that it's technically possible, but:

- Syntax is 👎 
- It doesn't support stateful helpers
- It doesn't support, like in contextual components, to override passed default values

Hence, I think there is room for a feature here.

knownasilya:
Does https://github.com/emberjs/rfcs/pull/432 supersede this RFC?

cibernox:
@knownasilya maybe? I haven't seen that until now.

chancancode:
As others have pointed out, I think the design and exploration here have been absorbed into #432. Let's move the discussion there. Thanks again @cibernox for pioneering the work in this area!

PR #213, opened 2017-03-13
Custom Components API

chancancode:
[rendered](https://github.com/emberjs/rfcs/blob/custom-components/text/0000-custom-components.md)

### DIFF
--- a/text/0000-custom-components.md
+++ b/text/0000-custom-components.md
@@ -5,22 +5,20 @@
 # Summary
 
 This RFC aims to expose a _low-level primitive_ for defining _custom
-components_. This API will allow addon authors to implement special-purpose
-component-based APIs (such as LiquidFire's animation helpers or low-overhead
-components for performance hotspots).
+components_.
 
-In the medium term, this API and expected future enhancements will enable the
-Ember community to experiment with alternative component APIs outside of the
-core framework, for example enabling the community to prototype "angle bracket
-components" using public APIs outside of the core framework.
+This API will allow addon authors to provide special-purpose component base
+classes that their users can subclass from in the apps. These components are
+invokable in templates just like any other Ember components (decendents of
+`Ember.Component`) today.
 
 # Motivation
 
 The ability to author reusable, composable components is a core features of
-the Ember.js framework. Despite being a [last-minute addition](http://emberjs.com/blog/2013/06/23/ember-1-0-rc6.html)
-to Ember.js 1.0, the `Ember.Component` API has proven itself to be an extremely
-powerful programming model and has aged well over time into the primary unit of
-composition in the Ember view layer.
+Ember.js. Despite being a [last-minute addition](http://emberjs.com/blog/2013/06/23/ember-1-0-rc6.html)
+to Ember 1.0, the `Ember.Component` API and programming mode has proven itself
+to be an extremely versatile tool and has aged well over time into the primary
+unit of composition in Ember's view layer.
 
 That being said, the current component API (hereinafter "classic components")
 does have some noticeable shortcomings. Over time, classic components have also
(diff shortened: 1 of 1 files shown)


text/0000-custom-components.md
```diff
@@ -0,0 +1,1078 @@
...
+Based on the information in the definition and the *component arguments*, the
+manager should return a *component instance* from the *create* method. From
+Ember's perspective, this could be any arbitrary value – it is only used for
+the component manager's internal book-keeping. In practice, this value should
+store enough information to represent the internal state of the component. For
+these reasons, it is often referred to as the "opaque state bucket".
+
+At a later time, before Ember is ready to render the template, the component
+manager's *getContext* method will be called. It will receive the *component
+instance* and return the context object for the template. The context binds
```
knownasilya: What can the component instance be? Just `Ember.Component` or can it be a web component/dom node?

text/0000-custom-components.md
```diff
@@ -0,0 +1,1078 @@
...
+the template `template:components/foo-bar` from the resolver.
+
+> **Open Question**: How does this interact with local lookup?
+
+The *manager* property is a string key specifying the `ComponentManager` to use
+with this component. For example, if "foo" is specified here, Ember will lookup
+the component manager `component-manager:foo`. This will be described in more
+detail in the sections below.
+
+> **Note**: Specifying the manager as a lookup key allows the component manger
```
tragiclifestories: Typo ("manger" -> "manager")

text/0000-custom-components.md
chancancode: Do you mean the *component instance* or the *comtext object*?

The *component instance* is for your own internal use, so it could be anything.

The *comtext object* is the template context (what `{{this}}` resolves to). It could also be any object. (Technically it can be a dom node but it won't do what you want.)

chancancode: You can already render a DOM node today. e.g. if a helper or property lookup returns a DOM node, it already works

knownasilya: > The component instance is for your own internal use, so it could be anything. It's whatever you returned from create.

Not sure I understand the reasoning behind returning _anything_.

text/0000-custom-components.md
chancancode: Fixed!

text/0000-custom-components.md
```diff
@@ -0,0 +1,1078 @@
...
+```
+
+Finally, when any of the *component arguments* have changed, Ember will call
+the *update* method on the manager, passing the *component instance* and a
+snapshot of the current *component arguments* value. This happens before the
+template is revalidated, therefore any updates to the context object performed
+here will be reflected in the template.
+
+Ember does not provide a fine-grained change tracking system, so there is no
+guarantee on how often *update* will be called. More precisely, if on of the
```
GavinJoyce: typo: if ~~on~~ _one_ of

text/0000-custom-components.md
tragiclifestories: There's a couple more of those mangers lying around. Festive :christmas_tree:

text/0000-custom-components.md
```diff
@@ -0,0 +1,1078 @@
...
+in flexi). In this case, it would make sense to store the components pool on
+the manager:
+
+```typescript
+class PooledManager implements ComponentManager<Object> {
+
+  private pools = {};
+
+  create({ metadata }: ComponentDefinition, args: ComponentArguments): Object {
+    let pool = this.pools[meatadata.name];
```
tim-evans: Typo; `meatadata` 🍖

text/0000-custom-components.md
```diff
@@ -0,0 +1,1078 @@
...
+
+    if (isRecycled) {
+      instance.didUpdateAttrs();
+    } else {
+      instance.didInitAttrs();
+    }
+
+    instance.didReceiveAttrs();
+
+    return instance;
```
MiguelMadero: Shouldn't this return `entry` instead of instance?

joukevandermaas:
I have a question on the scope of this RFC. Given the announcements at EmberConf, I sort of assume that the Glimmer/Ember integration can use this RFC as a building block. However, from the text in this RFC it appears that the layout/template of a component is an opaque thing, while Glimmer components also introduce new template syntax/semantics:

- using `@` to mark component parameters vs. DOM attributes
- using angle-bracket syntax to invoke components
- maybe more(?)

Is supporting that kind of semantics change in scope of this RFC? If so, how would I do something like that?

locks:
> I have a question on the scope of this RFC. Given the announcements at EmberConf, I sort of assume that the Glimmer/Ember integration can use this RFC as a building block.

That's exactly what this is ;) I think Tom and Yehuda mentioned in the keynote that the more successful experimentations in the Ember ecosystem came by providing primitives from which third-parties can build on, namely fastboot and engines.
This RFC is the primitive that will unlock the possibility of using different kinds of components, one of them angle bracket components that will be discussed in a follow-up RFC. You will notice that Glimmer.js components are very barebones at the moment, partially due to this.

NLincoln:
Would it be out of scope for this RFC to support changes to how components are rendered?

For example, say I have a google maps component that works like this:

```hbs
{{#google-map as |map|}}
  {{#map.marker as |marker|}}
    {{#marker.info-window}}
      This is rendered in the markers info window
    {{/marker.info-window}}
  {{/map.marker}}
{{/google-map}}
```

Currently the info window content is rendered into the DOM, and then google maps moves the element into its own internal structure. elements for markers, etc are also rendered. It would be amazing if there was a way to control how the element is inserted into the DOM, possibly like this?

```js
class MapsComponentManager {
  render(/* dom, element: HTMLElement */) {
    // no-op since we don't want these rendered
  }
}

// components/google-map/info-window.js
Ember.Component.extend({
  didInsertElement() {
    this._super(...arguments);
    // despite the component not being directly rendered, it's element 
    // is still available as a property on the component itself as usual
    assert(this.element);
  }
});
```

The `render` api could also be exposed on components instead of component managers instead.

MiguelMadero:
@chancancode this looks super useful and I like the idea of providing the primitives and to allow experimentation outside of ember. However, while in this case, I think the primitives look right, is only once we start building on top of those that we realize if they're right or now. I think we may need some time to experiment with this a throw a few different use cases to them before solidifying this or "committing" to an API. I'm mainly thinking of my use case for ember-cli-hot-loader, but I think the same would apply to liquid-fire and angle brackets... 
What do you think?

twokul:
First pass on registering custom component manager landed in Ember (https://github.com/emberjs/ember.js/pull/15254) and allows one to do the following:

```hbs
{{use-component-manager "nyan-cat"}}
```

knownasilya:
Is there a glimmer component manager yet?

rwjblue:
[Yes](https://github.com/glimmerjs/glimmer-component/blob/master/src/component-manager.ts)

knownasilya:
I'm assuming this has to be imported manually to work,  and isn't included in ember yet.

rwjblue:
Yep, the only thing landed in ember (behind a feature flag that is disabled by default) is the ability to specify a custom manager.

knownasilya:
Great 👍 So you can also use curly components in a glimmerjs app?

auvipy:
sorry for noise, just curious if there is any further development on this?

twokul:
@auvipy yeah. Here's a [PR](https://github.com/emberjs/ember.js/pull/15628) that aligns component manager's methods with RFC. The next step would be to work on capabilities (this will likely require introducing that functionality in Ember.js itself and making sure Ember.js itself is compatible with the latest Glimmer version)

runspired:
There are a few things this RFC does not address but which are required for flexi's `sustain` and `vertical-collection`'s recycling behavior, so I'm documenting them here for further clarification. I suspect they would affect `liquid-fire` as well and @ef4 can c/d. Some will likely require hands on exploration of additional needed APIs for `component-manager` and `glimmer`.

*1. Are lifecycle hooks only manageable for the top-most component instance managed?*

Example: `<managed-component>`, a component managed by a component manger, has the following template.

```hbs
<another-component>
  <component-c>
    <component-d/>
  </component-c>
</another-component>
```

While lifecycle-hooks for `managed-component` are under the component-manager's control, what about the corresponding lifecycle hooks for `another-component`, `component-c`, and `component-d`?

*2. The "re-init" problem* and ES Class `constructor` functions

`component.init` is often used to set initial state from the perspective that an instance is only ever used within a single context, buy with recycling this does not remain true. For instance, a component may set some "stable" public or private state in the following way:

```js
Component.extend({
  init() {
    this._super();
    this.createdTime = Date.now();
    this._updateCount = 0;
  }
});
```

A component-manager could choose to re-run `init`, but this could create issues with init's `_super` in the current implementation.  Were `init` to be removed in favor of class constructors, there would be no way to re-run this initialization on a scrubbed and recycled component instance.

*2.b {{unbound}} does not update for recycled component instances*

A related issue is the use of `unbound` in templates. Unless templates are completely rebuilt, currently these values would become static.  When the component is recycled and a new context applied, the previous unbound value would remain and fail to be overwritten.

*3. The template recycle problem*

When removing a recycle-able component instance from the render environment and returning it to the pool, we might not want to destroy it's DOM, it's child-component instances, or it's child-component instances' DOM.

When reusing a recycle-able component instance from the pool, we might want to apply fresh initial state to the pre-existing DOM from it's last usage.

In order to support both of these use cases, additional hooks and potentially glimmer behaviors would be needed.

ef4:
@runspired is correct that component recycling is potentially interesting for animation. However, I am ambivalent about whether it's worth the complexity it would introduce to the component mental model.

We have one place where we already deal with component identity and reuse: `{{#each}}`. It takes a `key` option because there's not one right answer for "what really counts as *the same* component". I have adopted `each`'s semantics in ember-animated for deciding when an element that's leaving one place in DOM should match with an element that's being inserted elsewhere, though I stop short of trying to move the actual components. The continuity is handled visually instead. I think it's important that animation is a presentation-level concern that won't alter the correctness of your program, and accidentally leaking state sideways from one component to another violates that.

The one place where I would feel comfortable stealing components and preserving their state is when I'm moving them into a temporary purgatory just to animate them before their destruction. But for that case, cloning DOM is basically good enough.

If we were talking about referentially transparent components that don't have internal state, recycling would be a no brainer. But that's not what we have. If you choose to recycle a component vs create a new one, you can't really be sure you'll get the same results. It may be possible to allow recycling of *blocks* rather than components, because blocks are just handlebars and their entire data flow is visible to glimmer, so glimmer can offer stronger correctness guarantees.

As for recycling as a performance optimization: first, the tradeoffs are likely to keep changing. It's not a given that recycling will stay faster than recreating. Second, if we do end up recycling I think the user-visible component layer is the wrong place for it. It should be below that, so that it's not observable by app code.

dbbk:
Is any progress being made on this?

twokul:
@dbbk yes, currently blocked by https://github.com/emberjs/ember.js/pull/15828. [This](https://github.com/emberjs/ember.js/pull/15783) and [this](https://github.com/emberjs/ember.js/pull/15628) will be the last steps to land this functionality behind a feature flag.

simonihmig:
@twokul your references seem to address the *implementation* of this RFC, but shouldn't the RFC itself enter FCP/be adopted, at some point? :)

twokul:
@simonihmig on that, Good Sir, I have no intel :P

rwjblue:
We are actively working through a few prerequisites to this RFC during the core team face to face meeting this weekend. There will be some updates here to cleanup verbiage, address unanswered questions, and add additional required capabilities in the next couple of weeks.

chancancode:
Hello everyone! Sorry for the delay. We had some new breakthroughs at the December core team face-to-face meeting. I updated the RFC to reflect the new development and this should be ready for another round of review, and hopefully FCP soon.

CHANGELOG:

✂️  typescript
✂️  component definitions
✂️  other overly abstract stuff
✂️  ember inspector stuff (seems like we can just do it automatically)
✂️  open/unresolved questions
➕ added nwe typos

The previous version is still available here, for reference: https://github.com/emberjs/rfcs/blob/e86b38d^/text/0000-custom-components.md

text/0000-custom-components.md
```diff
@@ -0,0 +1,1113 @@
...
+is intended for performing object-model level cleanup.
+
+Because this RFC does not provide ways to access or observe the component's DOM
+tree, the timing relative to DOM teardown is undefined (i.e. whether this is
+called before or after the component's DOM tree is removed from the document).
+
+Therefore, this hook is not suitable for invoking user callbacks intended for
+performing DOM cleanup, such as `willDestroyElement` in the classic components
+API. We expect a subsequent RFC addressing DOM-related functionalities to
+clearify this issues or provide another specialized method for that purpose.
```
rwjblue: s/clearify/clarify/

text/0000-custom-components.md
```diff
@@ -0,0 +1,1113 @@
...
+```
+
+```hbs
+Hello world! I have no backing class! {{this}} would be <code>null</code>.
+```
+
+## Recycling Components
+
+This example implements an API which maintain a pool of recycled component
+instances to avoid allocation costs, similar to Flex's "sustain" feature.
```
rwjblue: s/Flex's "sustain"/[flexi-sustain](https://github.com/html-next/flexi-sustain)/

text/0000-custom-components.md
```diff
@@ -0,0 +1,1113 @@
...
+users to interact with this layer directly. Instead, most users will simply
+benefit from this feature by subclassing these special components provided by
+addons.
+
+At present, the classic components APIs is still the primary, recommended path
+for almost all use cases. This is the API that we should teach new users, so we
+do not expect the guides need to be updated for this feature (at least not the
+components section).
+
+For documentation purposes, each Ember.js release will only document the latest
```
rwjblue: I don't think this is good enough. We need to maintain a canonical list of capabilities that are supported (and any options that were allowed with each).

chancancode:
I think I have addressed all the feedback so far!

chancancode:
I updated the RFC to add some details about the support policy of component manager APIs

rwjblue:
This was discussed in todays core team call, and with the last round of updates we are happy to get this moved into final comment period.

Please take a 🆕 👀 .....

chancancode:
I would like to point out that this is not intended to be the ultimate API that solves every use case. Rather, this is the MVP of the _framework_ that allows adding extensions that would solve those use cases. So I would appreciate feedback to focus on whether that framework is flexible enough to eventually support the use case you have in mind, rather than whether it is sufficient to support that today as proposed.

runspired:
I think to avoid confusion we should remove the theoretical `flexi-sustain` implementation from the RFC as its not an adequate implementation.

I chatted with @chancancode a bit this morning about the current state of this RFC and have no objections to it; however, I do want to document that while it hopes to be a primitive upon which things such as `flexi-sustain` could be built, additional RFCs would be needed on top of this MVP.

chancancode:
I updated it to remove the mention

rwwagner90:
I'd love to see some support for flexi-sustain and similiar things, but I realize it might be out of scope here.

text/0000-custom-components.md
```diff
@@ -0,0 +1,1166 @@
...
+- Ember Issue: https://github.com/emberjs/ember.js/issues/16301
+
+# Summary
+
+This RFC aims to expose a _low-level primitive_ for defining _custom
+components_.
+
+This API will allow addon authors to provide special-purpose component base
+classes that their users can subclass from in apps. These components are
+invokable in templates just like any other Ember components (decendents of
```
rtm: "descendants"

wycats:
Since 12 days have passed with no additional unaddressed comments, I'm going to merge this RFC.

PR #216, opened 2017-03-20
Add dasherized-test-name-generator-option RFC

GabrielCW:
See emberjs/ember.js#14915


0000-dasherized-test-name-generator-option.md
```diff
@@ -0,0 +1,92 @@
...
+(1) Add an option to the test generating commands to allow for the test name to be dasherized instead of being humanized (as it is [at the moment](https://github.com/ember-cli/ember-cli-test-info/blob/master/index.js#L27)), for instance `--dasherize`
+
+OR
+
+(2) Dasherize the test name by default and provide an option to humanize it, to get today's output, for instance `--humanize`
+
+Test name examples for `some-multi-word-service-test.js` :
+
+* With `--dasherize`: `Unit | Service | some-multi-word-service`
+* With `--humanize` (default today): `Unit | Service | some multi word service`
```
Turbo87: IMHO this RFC should just set the new default to `dasherize`, but without adding any new options that would increase the complexity

GabrielCW: Best case scenario for me ;)

cibernox:
I also agree that humanized names make no sense. I'm all in to change it.

rwjblue:
Heh, I actually think we should remove the description completely...

rwjblue:
But if we can't do that, we should use dasherized version.

0000-dasherized-test-name-generator-option.md
rwjblue: Agreed, this RFC should only be about changing the defaults.  I am generally 👎 on adding new options...

GabrielCW:
According to your feedback, I updated the RFC by removing the options.

The fix is also simpler, although I don't know the uses of `ember-cli-test-info` well enough to know if the `humanize()` function is called elsewhere, as it is public.

Turbo87:
> Heh, I actually think we should remove the description completely... But if we can't do that, we should use dasherized version.

@rwjblue if we pull in https://github.com/emberjs/ember-qunit/pull/258 then removing the test description is not an option

can we do anything to move this forward? I'm constantly being trolled by this when trying to test only single components...

rwjblue:
> @rwjblue if we pull in emberjs/ember-qunit#258 then removing the test description is not an option

I think you misunderstand me, I am saying we do not need **both** a _name_ **and** a _description_.  We should only have one, and actually it matches better with https://github.com/emberjs/ember-qunit/pull/258!!

Turbo87:
@rwjblue indeed, I still don't understand what you mean. what are your definitions of "name" and "description"?

rwjblue:
@GabrielCW - Thank you very much for taking the time to write up this RFC!  

I think it's very clearly correct, and in our opinion points out some pretty bad developer ergonomics in the current default generation system. It has been the policy for quite some time in Ember-land to consider bad developer ergonomics to be bugs.  

I am going to close this RFC as unneeded and advocate for landing the changes proposed in this RFC as a straight up bug fix.

Thank you again for helping push this forward!

PR #220, opened 2017-04-10
Introduce external redirects

kvas-damian:
[Rendered](https://github.com/Wikia/rfcs/blob/rfc-external-redirects/text/0000-external-redirects.md)

### DIFF
--- a/text/0000-external-redirects.md
+++ b/text/0000-external-redirects.md
@@ -60,8 +60,8 @@ Extending functionality of existing `transitionTo()` and `replaceWith()` methods
 developers from making a bug using an url which isn't handled by th application.
 Although it still can be a valid approach as Guides suggest using `routeName` 
 instead of URL.
-1. Add `allowExternal` flag as a new parameter in both methods. This solution breaks
-backwards compatibility - `transitionTo()` method accepts multiple model objects.
+1. Add `allowExternal` flag in `options` parameter for existing `transitionTo()` 
+and `replaceWith()` methods. 
 
 # Unresolved questions
 
--- a/text/0000-external-redirects.md
+++ b/text/0000-external-redirects.md
@@ -9,10 +9,10 @@ by current Ember application.
 
 # Motivation
 
-It's pretty common that Ember application is connected with another service 
-or module that can't be easily integrated into Ember application. For a basic 
-use-case we can bring an Ember application with an external login page. User 
-should be redirected to that page when he enters a page with restricted access.
+Our use-case is an Ember application is connected with another service 
+or module that can't be easily integrated into Ember application. Basic example 
+use-case is an Ember application with an external login page. User should be 
+redirected to that page when he enters a page with restricted access.
 There can be other applications using the same login page so integrating it into 
 Ember app can be complex. We cannot use `this.trasitionTo()` as it 
 asserts if `routeName` is handled by Ember application: `The route ${targetRouteName} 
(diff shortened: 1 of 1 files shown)


joukevandermaas:
I wonder what is the reason why this has to live on the route instead of as a separate package? Example:

```js
import { externalTransitionTo } from 'ember-external-transition';

externalTransitionTo('someplace');
```

kvas-damian:
@joukevandermaas Currently routes are responsible for redirecting to another route. Extracting `externalTransitionTo` to a separate package makes it less intuitive to find.

Can you share your reasons why we shouldn't put it in the Route?

joukevandermaas:
@kvas-damian I don't necessarily have strong feelings one way or the other, but it is an alternative. Since the entirety of FastBoot is a package (the RFC mentions FastBoot as one of the main reasons for the necessity of this new API), it might make sense to keep this separate also.

kvas-damian:
@joukevandermaas In our use-case FastBoot isn't the only consumer of this API. Our application requires this method on frontend as well. 

I'm also ok with implementing it as separate package when we agree that it's better way.

courajs:
I don't really think reducing your six-line snippet down to one line justifies adding new core API, and making it harder to teach how routing works in Ember. This is an issue that pretty much _only_ will come up when doing FastBoot/server rendering, as otherwise the browser API works perfectly well.
I'm sure some people would appreciate if a wrapper around the two approaches was published as an addon.

cibernox:
I also see this as a 5 lines utility rather than something to be added to the code. In the frontend, `window.location = 'https://new-location.com'` is enough. 

Only in fastboot it's a bit trickier, so could be a method in the fastboot service tho.

courajs:
This also conflicts to some extent with the external transitions of engines: http://ember-engines.com/guide/linking-and-external-links

les2:
at work we created a service called `navigation` for routing to external applications.

we define the external routes in config and use an api similar to `transitionTo`.

before engines was a thing, we developed a`{{link-to-external` helper for usage in templates. haven't changed the name yet ...

this architecture made it very easy to test by mocking out the navigation service, preventing tests from navigating the test runner to external sites, verify in tests that external links are functioning properly. it also neatly defined the external routes in one place.

i wish ember came with all that. :)

the point of all that was to say that 1) i think this api proposal would be useful, 2) i think the use cases could be taken a little further, e g., allow configuring a "map' for external urls when possible, 3) should be possible to implement in an add-on without modifying ember core.

the alternative of having if statements all over the place that bifurcate code base on fastboot availability is in my view just horrible. :sadpanda:

locks:
@kvas-damian thank you for your proposal, it's always interesting to hear about pain points in the practice of developing Ember applications.
Like some of the commenters, I also want to suggest that you create an addon to solve this use case, and then share it with the community. This is usually the recommended that for several reasons:
- It enables you to explore the problem space and validate the approach
- It strengthens the proposal by virtue of being able to say "install this addon to try it"
- You can argue for including the addon in the default blueprint, instead of adding to Ember's API

In this case, I believe that between the [new router service](https://github.com/emberjs/ember.js/blob/master/features.json#L8) and FastBoot that you have adequate APIs to implement this via addon.

mehulkar:
@villander didn't you just do some work on linking to ember engines? Would you mind linking that here (if i remember correctly)

buschtoens:
- https://github.com/buschtoens/ember-link/issues/290: external links
- [`ember-engines-router-service`](https://www.npmjs.com/package/ember-engines-router-service) based on https://github.com/ember-engines/ember-engines/issues/587: bringing the `RouterService` to engines
- https://github.com/ember-engines/ember-engines/issues/691: merging internal and external route namespaces in engines

villander:
@mehulkar we have also [Linking Between Engines RFC](https://github.com/emberjs/rfcs/pull/122/files) that based our external API for external routes.

Everything mentioned by @buschtoens follows this RFC and current API already implemented in Ember Engines addon.

wagenet:
I'm going to agree with @locks's suggestion and, since this has been relatively inactive, go ahead close this. If someone still wants to advocate this, let me know and I'll see that it gets a more comprehensive review.

PR #225, opened 2017-04-26
Allow {{mount}} syntax to accept parameters

mike183:
[Rendered](https://github.com/status200/rfcs/blob/ember-engines-mount-params/text/0000-ember-engines-mount-params.md)

### DIFF
--- a/text/0000-ember-engines-mount-params.md
+++ b/text/0000-ember-engines-mount-params.md
@@ -1,0 +1,126 @@
+- Start Date: 2017-04-26
+- RFC PR: [#15174](https://github.com/emberjs/ember.js/pull/15174)
+- Ember Issue: (leave this empty)
+
+# Summary
+
+This RFC proposes allowing parameters to be passed to the `{{mount}}` syntax.
+
+# Motivation
+
+This will enable developers to pass contextual data into routeless engines at
+runtime, allowing individual engines to be used multiple times through a single
+application under different contexts.
+
+An example could be a dashboard of charts where each chart is a routeless engine.
+Each chart could be of a different type and would require different data. This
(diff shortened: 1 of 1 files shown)


rwjblue:
👍 - Thanks for working on this (and the implementation PR https://github.com/emberjs/ember.js/pull/15174). The proposal seems good to me, but we should get @dgeb and @trentmwillis to weigh in on it from the Engines angle.

I'll add this to the Ember core team meeting agenda for this week also, to gather feedback from the group...

dgeb:
@mike183 Thanks very much for working on this. I think this proposal aligns nicely with Ember as it stands today. This feature does not need to wait on framework-level decisions related to routes, controllers, and components, but can adapt as necessary when those decisions are made.

👍

rwjblue:
We discussed this during todays Ember core team meeting today, and we are 👍 on this proposal.  

Given that there is little conversation here, and the proposal follows directly along with Ember's preexisting conventions we are moving this forward into the final comment period.

ef4:
Can I do this?

```hbs
{{mount "my-engine" widget=(component "my-widget") }}
```

And if so, does that component properly close over the other components and helpers that it depends on, or will it try to find them inside the engine (and presumably fail)?

rwjblue:
@ef4 - It should work properly (having access to things in the app but not in the engine), but we should add tests for that in the implementation PR.

rwjblue:
@mike183 added a test to the implementation PR (https://github.com/emberjs/ember.js/pull/15174) confirming that @ef4's example does work properly.

The FCP has been going for 7 days, comments have been addressed and no new information has surfaced that would prevent us from landing this. I believe we are good to go here.

PR #226, opened 2017-05-05
Named Blocks

machty:
[Rendered](https://github.com/machty/rfcs/blob/named-blocks/text/0000-named-blocks.md)

I presented some of the ideas behind this RFC at the EmberNYC meetup; here's a [video](https://youtu.be/nxnynYOgko0?t=1422) of that. (The proposed syntax has since changed though)

### DIFF
--- a/text/0000-named-blocks.md
+++ b/text/0000-named-blocks.md
@@ -1,0 +1,730 @@
+- Start Date: 2017-05-05
+- RFC PR: (leave this empty)
+- Ember Issue: (leave this empty)
+
+# Summary
+
+Introduce syntax for passing in multiple named template blocks into a component, and
+unify the rendering syntaxes / semantics for
+blocks/primitives/component-factories passed into a component.
+
+This RFC is focused chiefly on bringing named blocks to Ember Components,
+but it was necessary to define a basic roadmap of functionality for
+Glimmer Components as well, but keep in mind that some of the Glimmer-centric details may
+change and should hence be considered non-normative.
+
+# Motivation
+
+There are limitations to composition due to the inability to pass more than one block to a component (or 2 blocks if you include the inverse block).
+
+The result of this is that Ember developers have this ultra-powerful, compositional API for overriding portions of a component, but they can only use it in one place in the component invocation; any remaining overrides/configuration needs to be expressed as data and passed in as attributes to the component when it'd be vastly preferable to just pass in a chunk of DOM.
+
+Example:
+
+```html
+{{#x-modal headerText=page.title as |c|}}
+  <p>Modal Content {{foo}}</p>
+  <button onclick={{c.close}}>
+     Close modal
+  </button>
+{{/x-modal}}
+```
+
+This works, but the moment you need to render a component in the header (rather than just `headerText`), you end up having to add more config/data/attrs to `x-modal` just to support every one of those overrides, when really you just should be able to pass in another block of DOM to define what the header looks like. The API in this proposal would allow you to express this use case via:
+
+```html
+{{#x-modal}}
+  <@header as |c|>
+    {{page.title}}
+    {{status-indicator status=status}}
+    {{close-button action=c.close}}
+  </@header>
+
+  <@main as |c|>
+    <p>Modal Content {{foo}}</p>
+    <button onclick={{c.close}}>
+       Close modal
+    </button>
+  </@main>
+{{/x-modal}}
+```
+
+and with Glimmer components:
+
+```html
+<x-modal>
+  <@header as |c|>
+    {{page.title}}
+    {{status-indicator status=status}}
+    {{close-button action=c.close}}
+  </@header>
+
+  <@main as |c|>
+    <p>Modal Content {{foo}}</p>
+    <button onclick={{c.close}}>
+       Close modal
+    </button>
+  </@main>
+</x-modal>
+```
+
+Other RFCs/addons that have attempted to address this:
+
+- [named yields](https://github.com/emberjs/rfcs/pull/72)
+- [multiple yields](https://github.com/emberjs/rfcs/pull/43)
+- [yet another named yields rfc](https://github.com/emberjs/rfcs/pull/193)
+- [ember-block-slots](https://github.com/ciena-blueplanet/ember-block-slots)
+- [ember-named-yields](https://github.com/knownasilya/ember-named-yields)
+- [local template blocks](https://github.com/emberjs/rfcs/pull/199)
+
+# Detailed design
+
+## Invocation Syntax is separate from Component type
+
+The features specified in this RFC require us to nail down
+some specifics as to how Ember and
+Glimmer components interop, which syntaxes can be used to render
+them, and the mental/teaching model behind how it all works.
+
+- Invocation Syntax (curly vs angle brackets) is conceptually separate from Component
+  type (Ember vs Glimmer component)
+- "Curly Components" is a misnomer since curly syntax can render both Ember
+  components AND Glimmer components
+- Angle-bracket syntax can only render Glimmer components
+- `{{x-foo @k=v}}` will remain invalid curly syntax due to the `@k=v`
+- The way KV pairs provided at invocation is handled depends on the
+  component type:
+  - Given `{{x-foo k=v}}`, Ember Component `x-foo` will assign/bind `v`
+    to property `k` on the component instance, which can be rendered
+    within the component layout template as `{{k}}` (same behavior as always).
+    `{{@k}}` in an Ember component layout will remain a syntax error
+  - Given `{{x-foo k=v}}`, Glimmer Component `x-foo` treats `k=v` as
+    assigning/binding arg `@k` to `v`; it will assign/bind `this.args.k`, and expose
+    the value as `{{@k}}` within the template. This example invocation
+    is functionally equivalent to `<x-foo @k={{v}} />`
+  - The mental model is that with curly syntax, `k=v` is the syntax
+    for "passing data" to a component; Ember components receive/expose
+    this data via the properties on the component instance, and Glimmer
+    components receive the data as `@arg`s.
+- Curly syntax will not be enhanced with syntax for passing HTML attrs
+  (at this time)
+- Angle-bracket syntax does not support passing positional params
+
+Implementation-wise, these varying semantics will be defined/implemented via
+[Component Managers](https://github.com/emberjs/rfcs/blob/custom-components/text/0000-custom-components.md#componentmanager).
+
+## Multi-block Syntax
+
+Both curly and angle-bracket component invocation syntax will be enhanced
+with a nested syntax for passing multiple blocks into a component.
+
+The syntax for curly invocation is as follows:
+
+```html
+{{#x-foo}}
+  <@header>
+    Howdy.
+  </@header>
+
+  <@body as |foo|>
+    Body {{foo}}
+  </@body>
+{{/x-foo}}
+```
+
+and for angle-bracket invocation:
+
+```html
+<x-foo>
+  <@header>
+    Howdy.
+  </@header>
+
+  <@body as |foo|>
+    Body {{foo}}
+  </@body>
+</x-modal>
+```
+
+As demonstrated above, the _nested_ syntax for both curly and
+angle-bracket multi-block syntax has the format `<@blockName>...</@blockName>`.
+
+This multi-block syntax cannot be mixed with other syntaxes; either ALL
+the nested "children" of a component invocation need to be
+`<@blockName>...</@blockName>` (multi-block syntax), or none of them do
+(classic single-block syntax). The presence of any non-whitespace
+character between or around `<@blockName>...</@blockName>`s is a
+compile-time error.
+
+Passing two blocks with the same name is a compiler error (though this
+might be relaxed in a future RFC).
+
+### Blocks are just (opaque) data
+
+This RFC introduces the concept that blocks are just opaque values
+passed into components as data, rather than living in what is
+essentially a separate namespace only accessible to `{{yield}}`.
+
+In the above example (with curly syntax), Ember Component `x-foo` would
+have its `header` and `body` properties set on its instance. This means,
+among other things, that there's no need for a
+[hasBlock API for JavaScript](https://github.com/emberjs/rfcs/pull/102);
+you can just use normal property lookup / computed properties / etc to
+determine whether a block is provided. This means that blocks can
+be stashed on services and rendered elsewhere, e.g. the
+[ember-elsewhere use case](https://github.com/emberjs/rfcs/pull/226#issuecomment-299949000).
+
+### Unified Renderable Syntax
+
+Rather than continuing to enhance the `{{yield}}` syntax, we should take
+this opportunity to unify the various syntaxes for rendering things,
+from blocks to primitive values to component factories.
+
+We'll use the following example component invocation to explore
+what this syntax looks like: the following (curly syntax) invocation is valid syntax for
+rendering either an Ember.Component or a Glimmer Component named
+`x-modal` and passing it 3 named blocks: `header`, `body`, and `footer`:
+
+```html
+{{#x-modal}}
+  <@header as |title|>
+    Header {{title}}
+  </@header>
+
+  <@body>
+    Body
+  </@body>
+
+  <@footer>
+    Footer
+  </@footer>
+{{/x-modal}}
+```
+
+Given the above invocation, here's how you could render these blocks:
+
+```
+{{! within ember-component layout }}
+{{this.header "ima title"}}
+{{this.body}}
+{{this.footer}}
+
+{{! within glimmer-component layout }}
+{{@header "ima title"}}
+{{@body}}
+{{@footer}}
+```
+
+Both of these Ember/Glimmer layouts would render:
+
+```
+Header ima title
+Body
+Footer
+```
+
+The mental modal here is that is that for ECs, named blocks are
+set/bound as a properties on the instance, which we're rendering the
+same way we always rendering properties on the instance. For GCs, blocks
+are just args that we're rendering with the standard @arg syntax
+
+#### Why `{{this.header}}` and not just `{{header}}`?
+
+Unfortunately, we are constrained by the fact that `{{foo}}` means:
+
+- Try and find a helper named `foo`, and render it
+- If no such helper exists, fall back to rendering property `foo`
+  on the template context
+
+There is a risk that still exists today that any time you introduce a
+new helper to your codebase, for example, a `time` helper, you will break any
+templates that try to render a `time` property via `{{time}}`. This is
+an unfortunate hangover from the past, and we don't want to continue to
+expand the scope of this footgun with this RFC. Hence, to render
+blocks/component factories, you must use `{{this.time}}` in your Ember
+Component templates.
+
+This is actually an extension to behavior introduced by the
+[Contextual Component Lookup
+RFC](https://github.com/emberjs/rfcs/blob/master/text/0064-contextual-component-lookup.md);
+today, the following is supported:
+
+```hbs
+<!-- invocation -->
+{{x-foo header=(coponent 'x-header')}}
+```
+
+```hbs
+<!-- x-foo layout -->
+{{this.header}}
+```
+
+In the above example, `{{this.header}}` will actually render the
+`x-header` component factory; this RFC merely proposes extending this
+syntax so that it can render blocks as well.
+
+Since the potential for forgetting this rule is somewhat high, we
+should consider detecting when you use `{{foo}}` syntax when `foo` is a
+component factory or a block, and provide a useful warning or error
+message to use `{{this.foo}}` instead.
+
+#### Rendering primitives
+
+Consider the following example, which modifies above example by changing
+the `footer` block to a string:
+
+```html
+{{#x-modal footer="ima footer"}}
+  <@header as |title|>
+    Header {{title}}
+  </@header>
+{{/x-modal}}
+```
+
+If `x-modal` renders footer via `{{this.footer}}`, then it'll just
+render the "ima footer" string just fine; this a nice benefit of having
+a Unified Renderable syntax
+and supports a common workflow where string args can be promoted
+to full-on blocks without having to rework the component code
+to support an alternative/parallel API.
+
+#### Call syntax with primitives (or undefined)
+
+If you try to render a block/component with args, e.g. `{{this.foo 1 2 3}}`,
+then `foo` MUST be a block or a component. If `foo` is any kind of
+non-callable primitive, including undefined, it will be an error.
+
+#### Unified Renderable Syntax: component factories
+
+The following invocation using component factories is also supported:
+
+```html
+{{#x-modal
+     header=(component 'my-modal-header')
+     footer=(component 'my-modal-footer')}}
+  <@main>
+    Main
+  </@main>
+{{/x-modal}}
+```
+
+(It's worth mentioning that since we're only defining the `main` block
+here, this could also be expressed simply as:)
+
+```html
+{{#x-modal
+     header=(component 'my-modal-header')
+     footer=(component 'my-modal-footer')}}
+  Main
+{{/x-modal}}
+```
+
+This demonstrates that the unified renderable syntax is also capable of
+rendering component factories (previously only renderable via
+`{{component header}}`).
+
+Note since we're passing a positional param `"ima title"` to `header`,
+the `my-modal-header` component would only be able to access that param if it were
+using the `positionalParams` API with (`reopenClass`), which is a bit of
+a clunky / pro-user API.
+
+As a component author, if you want to write your components to support
+passing data to both blocks (which accept positional params) and
+components (which accept KV pairs), you can pass in both formats
+of the same data, e.g.:
+
+```
+// Ember.component layout
+{{this.header headerTitle title=headerTitle}}
+
+// Glimmer Component layout
+{{@header headerTitle title=headerTitle}}
+```
+
+#### Named block params
+
+Prior to this RFC, there were only two ways to pass in overridable
+chunks of DOM to a component:
+
+- Passing in a block, which only accepts positional block params
+  (or a `(hash)` object of named params)
+- Passing in a component factory, which only accepts KV args (unless
+  you use the `reopenClass-positionalParams` api)
+
+Given that we're introducing a Unified Renderable syntax, it would be
+unfortunate if we did nothing to address this impedance mismatch between
+named and positional params. The goal is for component consumers/invokers to
+be able to pass the most "convenient" kind of Renderable for their use
+case, be it a simple primitive string value, a block if they want the
+lexical scope + block params, or a component factory for rendering
+a shared component that might be used in many places throughout the app.
+Unfortunately, the component author will have to choose whether they
+want to pass positional params (which would push consumers towards
+only using  blocks) or named params (which are presently only supported
+by component factories).
+
+Hence, for this reason (among others), it makes sense to introduce a
+syntax for named block params; with this syntax, there will be an
+organic shift towards component authors using named KV pairs for passing
+data in most cases (while still allowing positional params in certain
+simpler cases were it only really makes sense to use a block, e.g.
+control flow components that wrap `if` or `each`, etc.)
+
+Here is the syntax for named block params:
+
+```html
+{{#x-modal}}
+  <@header as |@title @data|>
+    The title: {{@title}}
+    The data: {{@data}}
+  </@header>
+{{/x-modal}}
+```
+
+There is also a "soaking" syntax which is useful in cases where nested
+blocks might introduce new named block params that clobber preexisting
+identifiers in the scope, as well as cases where spelling out each
+named block param consumes too much rightward space. The above example
+can be expressed using the soaking syntax as follows:
+
+```html
+{{#x-modal}}
+  <@header as |@...d|>
+    The title: {{d.title}}
+    The data: {{d.data}}
+  </@header>
+{{/x-modal}}
+```
+
+(The `@` is not included as part of the identifier as that would
+suggest it was a KV arg rather than essentially a hash of args.)
+
+#### Block form of Unified Renderable syntax
+
+It should be possible to pass a block TO the
+block/component-factory that's been passed into a component.
+The common use cases for this are:
+
+##### Passing a block to a component factory
+
+Given the following invocation:
+
+```
+{{x-modal header=(component 'my-header')}}
+```
+
+It should be possible for `x-modal` to pass a block to the `header`
+renderable:
+
+```
+// ember-component x-modal layout
+{{#this.header title="ima title"}}
+  I'm a block provided by the component layout template.
+{{/this.header}}
+```
+
+Assuming `my-header` had a layout of:
+
+```
+<div class="my-header-inner">
+  title is {{title}}
+  {{yield}}
+</div>
+```
+
+This would render the following (assuming `x-modal` and `my-header` are
+Ember components with `tagName: 'div'` with `classNames` set):
+
+```
+<div class="x-modal">
+  <div class="my-header">
+    <div class="my-header-inner">
+      title is ima title
+      I'm a block provided by the component layout template.
+    </div>
+  </div>
+</div>
+```
+
+##### Passing a block to a block (aka contextual components)
+
+Given the following invocation:
+
+```
+{{#x-modal}}
+  <@header as |@title @main|>
+    <div class="header-block-content">
+      title is {{@title}}
+      {{@main}}
+    </div>
+  </header>
+{{/x-modal}}
+```
+
+This would render the following (assuming the same `x-modal` layout
+as the previous example:
+
+```
+<div class="x-modal">
+  <div class="header-block-content">
+    title is ima title
+    I'm a block provided by the component layout template.
+  </div>
+</div>
+```
+
+It would also be possible to pass a component factory to the header
+block from `x-modal`'s layout:
+
+```
+// ember-component x-modal layout
+{{this.header title="ima title"
+         main=(component 'x-modal-inner-content')}}
+```
+
+The multi-block syntax can be used as well with Unified Renderable
+syntax:
+
+```
+{{#header}}
+  <@main>
+    I'm a block provided by the component layout template.
+  </@main>
+
+  <@title>
+    ima title
+  </@title>
+{{/header}}
+```
+
+##### Block form of `@`-prefixed identifiers
+
+The syntax for passing a block to an `@`-prefixed identifier
+(named block params and Glimmer `@arg`s) will be
+`{{#@thing}} ... {{/@thing}}`, e.g.:
+
+```
+{{#x-layout as |@widget|}}
+  {{#@widget as |a b c|}}
+    Hi.
+  {{/@widget}}
+{{/x-layout}}
+```
+
+### Classic single-block syntax: `main` and `else` args
+
+It would be unfortunate if component authors had to use different
+syntaxes for rendering named blocks vs the traditional "default"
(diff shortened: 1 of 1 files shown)


alexlafroscia:
What are your thoughts on the difference between this proposed syntax and the "real" WebComponents v1 concept of "named slots"?  The idea there is that your component includes something like:

```html
<!-- my-component -->
<div class="a-wrapper">
  <slot name="a">
</div>
<div class="b-wrapper">
  <slot name="b'>
</div>
<div class="default-wrapper">
  <slot>
</div>
```

and would be used like:

```html
<my-component>
  <span slot="a">I end up in a-wrapper</span>
  <span slot="a">I also end up in a-wrapper</span>
  <span slot="b">I end up in b-wrapper</span>
  <span>I end up in default-wrapper</span>
</my-component>
```

I'm not sure how important it is to try to approximate the WebComponents version of the same idea, but it might be worth thinking about.

For more information, the MDN docs on `<slot>`:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/slot

machty:
@alexlafroscia we might very well support the web components version in the future, but as it stands, it's not very composable; for instance, you wouldn't be able to forward a slot to another component for rendering (whereas in this RFC you can just pass along the block to whomever since the blocks are just `@arg`s).

Also, a lot of people (myself included) aren't fans of the behavior that everything that isn't selected out of your "block" because the "default" block.

ef4:
I presume that the lack of any Javascript-side API for the block values is intentional because we don't have any designed public API for them. It may be good to say so explicitly, and we should make sure the implementation marks any methods on them as private. 

Is `<as>` future-proof against the HTML spec?  I guess we think its mandatory arg name is sufficient? 

The RFC focuses on the multi-block use case, but the first thing I would use this for is actually block capture of the default block, like for

```hbs
{{#to-elsewhere named="my-modal"}}
  Hello world
{{/to-elsewhere}}
```

by grabbing `@default` and sending it through a service. This is obviously a dynamic case that cannot be statically optimized -- it seems the implementation will need to be able to do the fast static thing when possible but still have a slower dynamic path.

This proposal sneaks `@` arguments into curly components. That has the benefit of avoiding collision with any arguments named `default` or `else` on existing components. But I think we need to explain how you access them from Javascript. `this.args.default`? `this['@default']`?

courajs:
How is this related to glimmer's `@args`? I can't tell whether it's a completely different usage of the same symbol, or whether it's actually the same exact concept but used only in this specific place.
If it's related, can we add a note to How We Teach This?
If it's not, can we add a note to Drawbacks about overloading `@` and potential confusion?

cibernox:
I'll leave my feelings of the currently proposed API in handlebars, as I think that the usage of `as` is confusing.

Traditionally we have been using `as` to precede the pipes on blocks. I understand `as` like a way of saying "the yielded things are now known by these names". A "naming" keyword.

I find the `{{as @header |h|}}` syntax backwards on that sense. Reasons:

- The pipes are not preceded by `as`, as they used to.
- `as` is the first keyword and is not followed by pipes.
- `as` is defining a block but it's not preceded by `#`
- Is `<as` in angle-brackets syntax safe/good-idea? This is a _very minor_ concern tho, as at the end of the day our templates will be compiled and `as` is not going to be on the html.

Could we use a different keyword? I'm thinking on the `block` keyword, preceded by a `#`, and keep using as to precede the pipes?

This:
```hbs
{{#x-foo as |a|}}
  @default block {{a}}
{{as @header |h|}}
  @header block. Referenced via @header in x-foo's layout
{{as @footer}}
  @footer block. No block params.
{{/x-foo}}
```

would become
```hbs
{{#x-foo as |a|}}
  @default block {{a}}
{{#block @header as |h|}}
  @header block. Referenced via @header in x-foo's layout
{{#block @footer}}
  @footer block. No block params.
{{/x-foo}}
```

machty:
We've taken in a lot of feedback over the last few days and it's clear that the `<as>`-based separator syntax is simply too awkward / ugly, and that we can do better. TL;DR, we're leaning more towards a syntax that looks as follows:

```html
<x-modal>
  <@header>
    <span class="special-header">
      Header Text
    </span>
  </@header>

  <@main as |c|>
    <p>Modal Content {{foo}}</p>
    <button onclick={{c.close}}>
       Close modal
    </button>
  </@main>
</x-modal>
```

This syntax has become a front runner because:

- It clarifies that `@header` and `@main` are just args that you're passing into the component along with another of the other "simpler" values that are passed into the component
- Pre-existing syntax highlighters / auto-indenters handle it much more gracefully
- It avoids a lot of the overhead/boilerplate of making you type `<template>` or `<block>`

There's still quite a bit to resolve but we're feeling pretty confident about this syntax. If you're interested in mulling over alternative syntaxes, make sure you take a look at the [graveyard of alternative syntaxes that have been considered as part of this RFC](https://gist.github.com/machty/f9e14b48b118ed0534d95a43f95fa78b).

csantero:
@machty I'm guessing the following syntax would be invalid? (block params on the component's tag when in multi-block mode)

```
<x-modal as |x y z|>
  <@header>
    ...
  </@header>

  <@main as |c|>
    ...
  </@main>
</x-modal>
```

mmun:
@csantero That's right.

jonnii:
@machty do you envision people being able to do something like:

```
<x-modal @header=(component ...)>
</x-modal>
```

or maybe?

```
<@namedblock>
  shenanigans
</@namedblock>

<x-modal @header=@namedblock>
</x-modal>
```

rwjblue:
@jonnii:

```hbs
<x-modal @header=(component ...) />
```

Yes, the idea is that you can interchange between a closure component, a named block, and even a plain string. 

So you might start with:

```
<x-modal @header="Some header text" />
```

Then realize you need more HTML for the header and update to:

```
<x-modal>
  <@header>
    <div class="hero">
      <img src={{heroImageSrc}}>
      Some header text
    </div>
  </@header>
</x-modal>
```

Then you might decide that you need an actual component to back the header and refactor to:

```
<x-modal @header=(component 'my-special-header' headerImage=heroImgSrc) />
```

In all of these examples, the layout of `x-modal` might simply be:

```hbs
{{@header}}
```

---

This same interchangeability does not exist (within this RFC) if you need to yield block params or hash arguments into the renderable. Currently, passing block params and positional params is only allowed for either named blocks or components, and passing hash arguments is only allowed with closure components. For example, invoking with positional params like this:

```
{{@header someParam}}
```

The above would issue a runtime error if `@header` was a string (my first snippet above). 

This is an open question called out in the current RFC text, and could be even be addressed in future RFC's if we decide on the more conservative approach here...

chadhietala:
The latest proposal seems to imply that we should declarative way of defining blocks. Today we contextually and implicitly define them:

```handlebars
<x-modal>
  {{! Implicit block definition}}
</x-modal>
```
In the proposal we effectively are defining macros that could expand into declarative calls to something like a `{{def-block}}` helper. For instance:

```html
<x-modal>
  <@header>
    <span class="special-header">
      Header Text
    </span>
  </@header>

  <@main as |c|>
    <p>Modal Content {{foo}}</p>
    <button onclick={{c.close}}>
       Close modal
    </button>
  </@main>
</x-modal>
```

Could be expanded/de-sugared into the following.

```handlebars
{{#def-block header}}
    <span class="special-header">
      Header Text
    </span>
{{/def-block}}

{{#def-block main as |c|}}
    <p>Modal Content {{foo}}</p>
    <button onclick={{c.close}}>
       Close modal
    </button>
{{/def-block}}

<x-modal @header=header @main=main />
```

This is out of the scope for this RFC but may paint how the implementation may play out and potentially how blocks could be reused in the future.

cibernox:
My only concern with `<x-foo><@header></@header></x-foo>` is, and this is probably a question for someone involved in the design of glimmer components, glimmer intends to allow devs to use dynamic tag names. Although if that will really be possible, the syntax would probably be `<{{@prop-containing-a-tag-name}}></{{@prop-containing-a-tag-name}}>`

rwjblue:
@cibernox yes, if that sort of syntax is what we want for dynamic tagname it would be `<{{@something}}></{{@something}}>` and so it would not conflict with defining named blocks under the current proposal.

joukevandermaas:
I'm slightly confused because Ember currently does not have `@` syntax for component arguments (unlike GlimmerJS). I cannot find an RFC for this syntax anywhere. To the best of my knowledge it was just introduced to GlimmerJS? This RFC seems to assume it will be introduced to Ember, but shouldn't there be an RFC for it first then?

Does this mean this RFC depends on the new syntax landing in Ember first? Or does this RFC only apply to GlimmerJS until the syntax lands?

machty:
@joukevandermaas shipping this RFC means fleshing out a basic idea of `@arg` semantics with curly components. It's looking most likely that you can use `@arg=` notation with "curly" component invocations, and that it'll be in a separate namespace from classic `arg=`s. This approach does mean there will be an open question as to when to use `arg=` vs `@arg=`, and I think the best answer is "only use it for blocks" when you're using curly components.

luxferresum:
Do I understand it correctly, that this will basically will work exactly like [this](https://github.com/emberjs/rfcs/pull/203), and just uses a different syntax to specify block-arguments?

And will we also get a way to use this with a helper?

machty:
@luxferresum it will hit 99% of the use cases of that other RFC, but is a much more familiar/iterative/achievable API enhancement.

What does "use this with a helper" mean?

knownasilya:
What's the state of this RFC?

machty:
@knownasilya still under active development. Will follow up after this coming weekend's core team F2F, but here are a few mostly-agreed-upon developments that haven't yet been reflected in the RFC:

1. We recognize the distinction between component invocation syntax and component type: we have `{{curly}}` syntax, which can render Ember Components AND Glimmer Components, and we have `<angle>` syntax which can only render Glimmer Components.
2. With curly syntax, the format for passing "args" is and will remain `k=v`; `{{x-foo @k=v}}` will be a syntax error. If you're rendering an Ember Component, `k=v` will, as always, translate to a property set/binding on the Ember.Component instance. If you're rendering a Glimmer component with curly syntax, `k=v` will translate to passing in an arg that is referenceable as `{{@k}}` in the glimmer component layout template.
3. We're sticking with `{{#x-foo}} <@namedBlock>hey</@namedBlock> {{/x-foo}}` as the multi-block curly syntax; for EComponents, this sets the `namedBlock` property on the component instance and will be renderable as `{{namedBlock}}` in the template layout. For Glimmer components: `{{@namedBlock}}`.
4. Classic "single-block" invocation for curlies will not set `main`/`else`/`default`/`inverse` properties on the component instance; those should continue to be rendered with `{{yield to="inverse"}}`/etc.

These seem to be the best compromises for backwards compatibility but by no means final and hopefully we can flesh it out a lot more over the F2F; will be updating the official RFC shortly.

machty:
I just posted a pretty major update to the RFC, mostly fleshing out the details as to how Ember/Glimmer semantics diverge in some cases, and how component invocation syntax works with both component types. I feel _pretty_ good about where it's at, but I really haven't gotten a chance to share the full details with the rest of the core team so keep in mind that some of the details are far from settled. (But please provide whatever feedback comes to mind!)

text/0000-named-blocks.md
```diff
@@ -0,0 +1,635 @@
...
+
+### Classic single-block syntax: `main` and `else` args
+
+It would be unfortunate if component authors had to use different
+syntaxes for rendering named blocks vs the traditional "default"
+and "inverse" blocks provided by the classic single-block syntax.
+
+Hence, the blocks provided in classic single-block syntax should also
+be exposed as properties (Ember) and args (Glimmer), and should have
+conventional, meaningful names names: instead of "default" (which is a
```
ro0gr: names names

machty: names named names

text/0000-named-blocks.md
```diff
@@ -0,0 +1,635 @@
...
+- [ember-elsewhere](https://github.com/emberjs/rfcs/pull/226#issuecomment-299949000)
+  (and other similar patterns) require having access to the opaque block
+  so that it can be stashed on a service and rendered elsewhere
+- wrapper components that forward args/properties/blocks to another
+  intenal component; blocks need to be accessible as properties in order
+  to pass them into another component (otherwise you'd have to use a
+  combinatoric mess of block syntax + `if hasBlock` checks to forward
+  blocks through to the inner component)
+
+I propose an opt-in API; any Ember Component that wants `main`/`else`
```
ro0gr: Do you mean components with default and inverse blocks like today?
Do I understand correctly that you propose to extend all of them with a snippet below? Sorry if I misunderstood.

text/0000-named-blocks.md
```diff
@@ -0,0 +1,635 @@
...
+
+// Glimmer Component layout
+{{@header headerTitle title=headerTitle}}
+```
+
+#### OPEN QUESTION: "calling" string args
+
+How should we handle `{{@header some blockparam data}}`
+when `@header` is a non-"callable", simple value like a string? Should
+we loudly error? Should it just ignore those "extra" args and render
```
knownasilya: What about having a debug flag, like the ones in `config/environment.js`, and by default fail silently. But like the transitions debugging flags, output if the flag is set.

text/0000-named-blocks.md
knownasilya: Yes. Basically, you could create a component like this, and just extend any of your components from that one.

text/0000-named-blocks.md
```diff
@@ -0,0 +1,635 @@
...
+  {{this.else}}
+{{/if}}
+```
+
+##### Alternatives? Nicer opt-in syntax?
+
+The `reopenClass` syntax s clunky, and I wouldn't be surprised if people
+wanted to use unified renderable syntax all over the place. So is there
+a less painful way to opt-in? Could it just be a property set on the
+prototype rather than the class object?
```
knownasilya: Agreed on the clunkiness of `reopenClass`

text/0000-named-blocks.md
```diff
@@ -0,0 +1,635 @@
...
+    <@footer>
+      {{submit-button}}
+    </@footer>
+  {{/if}}
+</x-layout>
+```
+
+This might be desirable in the future, particularly for use cases
+involving flex-ish layouts where the component changes behavior /
+appearance based on whether blocks on passed in.
```
knownasilya: This is pretty important for anything ambitious...

text/0000-named-blocks.md
machty: Addons will be using this syntax with its intended semntics and I don't think there'd be a way to prevent a config flag from falsely enabling logs on code that doesn't belong to you.

text/0000-named-blocks.md
machty: In the future we'll likely support local block declarations (so that you can pass the same block in to multiple components / multiple args of the same component; with something like this in place you could do something like `{{x-layout footer=(if showFooter localFooterBlock)}}`. It's not great, but it's workable til we feel out all these ambitious use cases.

text/0000-named-blocks.md
joukevandermaas: It should be an error IMO, similar to when you run `{{component <something>}}` and `<something>` cannot be resolved to a known component.

machty: I don't think this can be compared to the component case, which fails for the same reason injecting an unknown service should fail. This is IMO more like how `{{yield}}` no-ops when `hasBlock` is false; we don't require you to wrap all your yields in `{{#if hasBlock}}` because, well, that would be miserable.

text/0000-named-blocks.md
joukevandermaas: How about an `{{is-block` helper that returns true if the first argument is a block, false otherwise? Example:

```hbs
{{#if (is-block @localFooterBlock)}}
  {{@localFooterBlock}}
{{/if}}
```

If the consumer did not pass in `localFooterBlock`, or if `localFooterBlock` is e.g. a string, this would not render anything.

I have already built a similar helper for the biggest app I work on, where in certain cases we allow people to optionally pass components into other components:

```hbs
{{#if (is-component localFooterBlock)}}
  {{component localFooterBlock}}
{{/if}}
```

This generally works pretty well as a pattern.

machty: Sure, that'd work, but the idea behind unified renderable syntax is that all you should really have to do is check if the arg/property is truthy or not; it's either present or it's not, and you shouldn't have to check for the presence of a specific type of renderable (be it a block or component factory), as that would arbitrarily force anyone invoking your component to use only specific kinds of renderables.

Then again... the only way to build components that support universal renderables is to commit to always passing args to universal renderables as both positional params and named k=v pairs, which, is kinda weird/bad?

Elsewhere in the RFC I demonstrated this as `{{header headerTitle title=headerTitle}}`, but maybe that's ridiculous to expect people to do, or maybe it only makes sense for addon authors to do, and app maintainers can just choose/prefer invocation styles that are consistent with the rest of the app? Will have to think on this...

machty: OK, thinking on this more, I think what we can expect is that if we include named block param syntax as part of this RFC (and we probably should), then there will be a natural/organic shift towards using named KV args when universality is desirable. In other cases like control flow (e.g. a `fancy-each`) component, yielding a positional param or two will remain a nice block-centric API for those kinds of use cases. And maybe in some cases there will be a mixture of positional and named params but I think generally speaking the API nudges folk toward named KV pairs, which is good.

text/0000-named-blocks.md
```diff
@@ -307,6 +308,61 @@ case, nor would I want to have to introduce all of the `is-block` or
 guard.
 ```
 
+#### Named block params
```
knownasilya: 👍

machty:
We're likely to FCP this on Monday. Per core team feedback I've refined a few things to clarify the Ember-centric nature of this RFC; spelling out a basic Glimmer roadmap is important but some details are likely to change over time, and hence Glimmer suggestions in this RFC should be considered non-normative.

oligriffiths:
@machty To clarify, is the reason you're not going with:

```hbs
{{#x-modal}}
{{@header}}
My header
{{/@header}}
{{/x-modal}}
```
Because the `{{@header}}` is invalid hbs syntax? It does seem a bit weird to mix the angle and curly syntax within a curly component...

machty:
@oligriffiths `{{@header}}{{/@header}}` as a (proposed) syntax means "render the `@header` arg here and pass the block containing "My header" to it.

There's a few reasons we went with angle brackets + `@`:

1. That'll (most likely) be the syntax for Glimmer component named blocks
2. `{{@header}}{{/@header}}` already has a meaning (as mentioned above)

oligriffiths:
Correct me if I'm wrong, but `{{@header}}` is *only* used in glimmer components, to render an argument.

Given that `{{@header}}` is invalid HBS syntax anyway (afaik), why not choose another invalid HBS character and keep it curly, like:
`{{~header}}my header{{/~header}}` or `{{$header}}my header{{/$header}}`

machty:
@oligriffiths You're correct (unless we decide that named arg `|@foo|` in an Ember component should be rendered via `{{@foo}}`, which some folk are leaning toward). The point is we don't want any more tokens/sigils to add to the mix. We've considered [many syntactical variants](https://gist.github.com/machty/f9e14b48b118ed0534d95a43f95fa78b) at this point and have found `<@header></@header>` to be the least bad proposal thus far.

text/0000-named-blocks.md
```diff
@@ -0,0 +1,696 @@
...
+
+```
+// Ember.component layout
+{{header headerTitle title=headerTitle}}
+
+// Glimmer Component layout
+{{@header headerTitle title=headerTitle}}
+```
+
+#### `maybe-call` built-in
```
chadhietala: I think we just need to move inline `{{if}}` to a builtin.

webark:
if you where to have a component that had a named block `{{header}}` i'm assuming if you added in a standard `{{yield}}` in there also, that would throw a compile time error?

machty:
@webark, no, today's `{{yield}}` behavior will be preserved and just render nothing if no `main` (today "default") block is provided.

courajs:
How would you pass a block out to a passed in block/component-factory, while still allowing for primitive values?
I don't see how you compose `maybe-call` and the unified renderable block form.

courajs:
Maybe we could just have an `is-primitive` or `is-callable` helper? Then instead of `maybe-call`, you would use `#if`:
```hbs
{{#if (is-primitive header)}}
  {{header}}
{{else}}
  {{#header title="some title"}}
     I'm a block
  {{/header}}
{{/if}}
```
We could keep `maybe-call` as a convenience for simple cases.

text/0000-named-blocks.md
```diff
@@ -0,0 +1,696 @@
...
+#### Glimmer Components: `@main` and `@else`
+
+Given Glimmer component invocation:
+
+```
+{{#fancy-if cond=trueOrFalse}}
+  True
+{{else}}
+  False
+{{/fancy-component}}
```
alexander-alvarez: `{{/fancy-if}}`

machty: Thanks, fixing

machty:
> How would you pass a block out to a passed in block/component-factory, while still allowing for primitive values?

> I don't see how you compose maybe-call and the unified renderable block form.

So you mean, how would you write a wrapper component that internally forwards blocks to another component such that the inner component wouldn't need to have to use `(maybe-call)` itself?

I think I understand what you're asking but could you write a quick example?

machty:
@courajs (previous comment was for you, but also:)

Yes, something like `(is-primitive)` would work, though I might prefer the inverse `(is-callable)`.

TBH, while I acknowledge the "weirdness" that is being able to call/invoke/pass args to something that's just a primitive, I still have a hard time arriving at any other conclusion than to just allow it for the following reasons:

1. Whether it's a built-in or implemented in user-space (based on `is-callable`/`is-primitve` built-ins), I suspect there will be a strong community-wide tendency to always render stuff passed into you as `maybe-call`
2. `maybe-call` and when to use it is just one more thing we'd have to teach
3. One of the arguments I've heard in favor of requiring an explicit `is-callable` check is that Ember's rendering layer has long suffered from issues of debuggability due to quiet failures/errors or flexible rendering/calling semantics, and that this is an opportunity to not preemptively open pandora's box for this new feature, but we should actually think through what debuggability issues there would be if we allowed for flexible calling semantics here: 

If you pass a string to something that the component expects to be a block/component factory, then what? Then you see that string content that you passed in rendered to the screen... which... is what you wanted to happen anyway?

The worst case of debuggability is arguably when you haven't properly wired up arg/property names in a component invocation, so a component ends up rendering nothing instead of what you intended to pass into it. Maybe we should explore APIs that prevent that particular case.

```
{{maybeUndefined 1 2 3}}
```

If we made it so that you had to do

```
{{maybe-call maybeUndefined 1 2 3}}
```

then would `maybe-call` quietly render nothing with `maybeUndefined` is undefined?

courajs:
@machty More like, how do you pass a block into something that might be a primitive?

In [Passing a block to a component factory][block-form], how do you support the user passing a primitive: `{{x-modal header="header text"}}`, while still sending the title if they pass a block/component?

Perhaps `maybe-call` has to support a block form?

[block-form]: https://github.com/machty/rfcs/blob/58fe83a3ca0ba2d01f8f85b35e37374c330143ff/text/0000-named-blocks.md#passing-a-block-to-a-component-factory

machty:
@courajs agree that `maybe-call` would have to support block form.

simonihmig:
Maybe I missed some discussions on the `maybe-call` helper, but for me it is not clear why we need to throw an error when a primitive is passed and we try to render that with block params. When a block is passed, and that block does not actually render any of the passed params, we don't error as well. 

So can't we treat the case just the same as a block that happens to have only static content. So...

```hbs
{{#x-modal header="foo"}}
  ...
{{/x-modal}}
``` 

... is treated just the same as:

```html
{{#x-modal}}
  <@header>
    foo
  </@header>
  ...
{{/x-modal}}
``` 

... even when the header in `x-modal.hbs` is rendered like `{{header headerTitle}}`?

cibernox:
My question is about the "unified renderable" concept. I proposed a while back the idea of contextual helpers. That is, a way of currying arguments to a helper but avoid its evaluation until the very end.

The conversation derived on the idea that unlike components, we needed a keyword to "wrap"/"curry" arguments and another keyword to actually execute it. I wonder how this might interact with the `maybe-call` function. 
Could a "HelperInstance" be considered a renderable and just be invoked as a block?

I'd like to have some input on the idea of unifying components, helpers and blocks as "renderable" objects.

RFC: https://github.com/emberjs/rfcs/pull/208

machty:
@simonihmig I mostly agree with you but others would point out that JavaScript already has already led by example with function call semantics where args can be omitted/ignored and arity doesn't need to match, but what JavaScript _doesn't_ have is `("string")()`, which we'd be introducing if we let `{{header 1 2 3}}` render when `header` is a string.

simonihmig:
@machty thanks for pointing that out. I am however not really able to follow that argument. 

A: in my mental model of what a template is, I would not think of `{{foo}}` as just being a simple function call. A Handlebars template is more than just some syntactic sugar on top of plain JS (hey, it's not JSX 😉). So that analogy feels somewhat unexpected...

And B: if it were so, wouldn't `{{header 1 2 3}}` resemble `("string")(1, 2, 3)`, thus `{{header}}` would actually match `("string")()`. Both of which would not work in JavaScript! But the latter is supposed to work!? 🤔

As much as I would love to see this land, I would argue to keep the new primitives needed to be introduced for this to be as few as possible! And this `maybe-call` thing seems unnecessary to me. Maybe satisfies some purist/academic thoughts, but causing more problems than it solves in the real world, especially for new users having to learn all these new concepts...

wycats:
@simonihmig In practice, you would rarely be writing code generic enough to notice the difference.

It's not precisely an academic/purist thought. In my opinion, the way to think about:

```hbs
{{someValue}}
```

is: append `someValue`. For example, if `someValue` is a DOM node, append the DOM node. If `someValue` is a safe string, append it as HTML. If `someValue` is a component (constructed via `(component)`) append the component. If `someValue` is null or undefined, append nothing. If it's a document fragment, append the document fragment. And on and on.

In this case, if `someValue` is a function, call the function and invoke its value.

There's a rule for appending many kinds of values, and functions are no different. That's very different from the syntax of:

```hbs
{{upcase value}}
```

Which means: call the function `upcase` (looked up using the normal helper lookup rules) with the `value` parameter. If `upcase` is not a helper (even if it's a local variable).

Today, if you try to write this code:

```hbs
{{#with (hash string="hello") as |c|}}
  {{c.string}}
{{/with}}
```

It'll work. But if you try to write this:

```hbs
{{#with (hash string="hello") as |c|}}
  {{c.string 1}}
{{/with}}
```

It won't.

There's a difference between appending a value and calling a "function". In today's world, the only kind of "functions" we have are helpers and variables, and if you try to call something else, it won't work. I don't want to change that here.

The reason I don't think you'll hit this much unless you're writing very generic code is that you'd be designing an API where people could pass in either a block or string, but *also* want to pass additional parameters to the block.

Let's do a quick case study. Let's imagine I'm writing a component  named `little-post` that wants to allow the caller to customize the title and I do this:

```hbs
<h1>{{header current-date}}</h1>
<div>{{body}}</div>
```

And it's intended to be called thusly:

```hbs
{{#little-post}}
  <@header as |date|>hello world today is {{date}}</@header>
  <@body>Today is my favorite day!</@body>
{{/little-post}}
```

It's true that I won't be able to write `{{#little-post header="Hello world!"}}`. However, I could write this:

```hbs
{{#little-post}}
  <@header>Hello world!</@header>
  <@body>Today is my favorite day!</@body>
{{/little-post}}
```

If the author of the component only wants you to pass a block, then no problem: pass a block.

This is analogous to wanting to call a JavaScript API that accepts a function and all you have is a string. No problem! Pass a function (`arr.map(() => "hello")`).

To me, this is not about trying to solve abstract problems that don't exist in the real world, but rather about avoiding creating even more ambiguity in the meaning of Handlebars expressions. If something looks like it's calling a function, it should be calling a function.

simonihmig:
Thanks @wycats for the thorough explanation, appreciate that! And given that, it makes much more sense to me now!

Having introduced Ember to a few folks, I still have the gut feeling that this, well, let's put it like this, it needs to be teached well! ;) Or as you said, to establish in general a best practice to design your API in a way that it's either props or blocks...

The mental modal I had described in https://github.com/emberjs/rfcs/pull/226#issuecomment-315841018, that a string given as a property is comparable to a static template in a block, was actually doing more harm to me here. I guess the better way to build a mental model around that is to think of `<@header>foo</@header>` as constituting a callable (like `() => 'foo'`), so you can safely do either `{{header}}` or `{{header 1}}`. Is that fair to say?

mmun:
@machty You should mention what happens (presumably an error) if you use two blocks with the same name.

machty:
@mmun added a note that it's to be considered a compiler error for this RFC

vlascik:
Syntax looks really nice, but is there a plan to deal incompatibility with handlebars with respect to IDEs? At least in IntelliJ/Webstorm, `<@header>Hello world!</@header>` is an error. Usually with an unsupported syntax like this it's just a question of time when will some release break IDE assistance completely (which means a significant loss of productivity).

machty:
@vlascik That's definitely an (acknowledged) risk/downside of introducing new syntax (in addition to the fact that syntax highlighters in general will need to catch up). I believe though that among the alternative syntaxes that have been proposed throughout the course of this RFC that the syntax we settled on should be the easiest for IDE's / syntax highlighters to implement; perhaps it'll just be as easy as adding `@` to a list of open/closing valid tag characters.

jgwhite:
For posterity, it'd be very interesting to know if there's been any discussion of multiple named blocks of the same name. On first glance, it's tempting to want to write something like:

```hbs
{{#power-select}}
  <@option value="foo">
    <em>Foo</em>
  </@option>
  <@option value="bar">
    <blink>Bar</blink>
  </option>
{{/power-select}}
```

While I imagine it would be possible to implement this (`@option` comes through as an array of renderables, for example, though not sure about the block args) I also imagine it's been ruled out for good reason. Would be great to capture some that reasoning in this thread.

machty:
@jgwhite It's definitely come up in discussions, but ultimately it was decided that it belonged in another RFC (this RFC defines multiple blocks with the same name as a syntax error, but that that might be relaxed in the future). There's a lot of details to flesh out though, and I don't want to distract too much away from the thread of _this_ RFC, so I'll just say: yes, it's probably on the horizon and would probably elegantly address the `power-select` use case, was well as some use cases for table components and some others.

jgwhite:
@machty wonderful, all music to my ears.

machty:
Just updated the RFC with the latest thoughts; after much discussion, it was decided that we didn't want to bloat the footgun that is the legacy helper resolution logic for 0-arity mustaches, e.g. `{{foo}}`; hence, the Unified Renderable syntax is `{{this.foo}}`. We've also settled on `{{#@foo}} ... {{/@foo}}` as the syntax for rendering / passing a block to named block params (and presumably Glimmer args in the future). For more explanation, [see here](https://github.com/machty/rfcs/blob/named-blocks/text/0000-named-blocks.md#why-thisheader-and-not-just-header).

text/0000-named-blocks.md
```diff
@@ -0,0 +1,727 @@
...
+
+The mental modal here is that is that for ECs, named blocks are
+set/bound as a properties on the instance, which we're rendering the
+same way we always rendering properties on the instance. For GCs, blocks
+are just args that we're rendering with the standard @arg syntax
+
+#### Why `{{this.header}}` and not just `{{header}}`?
+
+Unfortunately, we are constrained by the fact that `{{foo}}` means:
+
```
Panman8201: Not sure if it applies here, but there is one more rule to the list below;

- Look for a block param named `foo`, and render it

That one "wins" over all, per: https://github.com/emberjs/ember.js/issues/14413#issuecomment-251666245

tomdale:
@machty Have you thought at all about namespaced components and npm scoped packages? I think we've discussed something like:

```hbs
<package-name::component-name @foo={{bar}} />
```

Is there any ambiguity with named blocks if you have a scoped package? E.g.:

```hbs
<@scope/package-name::component-name @foo={{bar}} />
```

machty:
@tomdale I don't see any issue/ambiguity; named blocks are essentially just args, which to my knowledge don't need to be namespaced in any way, so neither would any `<@namedBlocks>...</@namedBlocks>`.

text/0000-named-blocks.md
```diff
@@ -0,0 +1,730 @@
...
+  <@option value="bar">
+    <blink>Bar</blink>
+  </option>
+{{/power-select}}
+```
+
+This RFC defines that passing multiple blocks with the same name is a
+syntax error, but it's something we might want to relax in the future
+for certain cases where you want to pass arrays of blocks, such as
+`power-select` or use cases involving tables.
```
alexander-alvarez: Does this hold true for multiple invocation at runtime? `{{#each items as |item|}}  <@option> </@option> {{/each}}`

text/0000-named-blocks.md
```diff
@@ -0,0 +1,730 @@
...
+e.g.
+
+```html
+{{#power-select}}
+  <@option value="foo">
+    <em>Foo</em>
+  </@option>
+  <@option value="bar">
+    <blink>Bar</blink>
+  </option>
```
alexander-alvarez: `</@option>`

knownasilya:
@machty maybe @tomdale meant if the package name has `@` in it, e.g. `@myorg/myaddon`, how would that work with the `@`, or is it a non-issue because the named blocks are inside the component block, versus the scoped/namespaced component is at its own level?

tomdale:
@knownasilya @machty Sorry, I should have provided an example:

```hbs
<x-modal>
  <@ember-power/select>
    Header {{title}}
  </@ember-power/select>
</x-modal>
```

How do you disambiguate between whether `@ember-power/select` is a component invocation in the implicit main block, or a named block named `@ember-power/select`?

knownasilya:
Maybe named blocks can't have a `/`?

machty:
We discussed this RFC in today's core team call, and decided it makes sense to merge it, with the caveat that since there are still open questions regarding the full story for Glimmer component invocation, there _may_ be some slight changes (mostly likely syntactical, if any) along to way to shipping this feature, but that it didn't make sense to indefinitely block implementation of named blocks in Ember on waiting for 100% of the Glimmer story to fall into place.

@tomdale's example above highlights a possible case where syntax _might_ need to shift, but such cases might also be addressed by other means (e.g. perhaps an import mechanism or something along those lines).

rwjblue:
:clap: :tada:

@machty - Great work shepherding this through the RFC process. Thank you for working so hard on it!

locks:
👏 👏

chriskrycho:
Is there a ticket tracking implementation of this feature?

mmun:
@chriskrycho No, named blocks needs some more design work to deal with angle bracket invocation. 

There's also some vague concerns about the idea and approach in general. The current proposal for named blocks doesn't have an answer for cases where you want to have multiple blocks with the same name or for situations where being able to nest blocks would be useful.

oliverlangan:
@mmun, can you elaborate a bit on the concerns, or direct us towards those discussions? Once the RFC was accepted, we decided to hold off on some refactoring until it lands—and if there is a chance that it may never be implemented we will have to explore other options.

sandstrom:
@oliverlangan There is https://github.com/ciena-blueplanet/ember-block-slots which may be a viable alternative, while you wait for named blocks.

chilicoder:
@sandstrom `ember-block-slots` seems to be abandoned and doesn't work for latest versions of Ember.

aalimovs:
@chilicoder we're using `ember-block-slots` in 3.5 with `ember-cli-shims`, works great.

sandstrom:
For anyone stumbling on this, implementation of yieldable named blocks is tracked here:
https://github.com/emberjs/rfc-tracking/issues/44#issuecomment-612478783

PR #228, opened 2017-05-22
Single File Components RFC

DHedgecock:
[Rendered](https://github.com/DHedgecock/rfcs/blob/single-file-components/text/0000-single-file-components.md)

Submitting this RFC as it is proposing a: new idiomatic usage or convention.

### DIFF
--- a/text/0000-single-file-components.md
+++ b/text/0000-single-file-components.md
@@ -5,7 +5,7 @@
 # Summary
 
 It is currently possible to create single file Ember components with component
-templates declared inline using a tagged template string assinged to the `layout`
+templates declared inline using a tagged template string assigned to the `layout`
 property. However, this pattern is not documented and support is unknown. This RFC
 proposes documenting support for this pattern.
 


text/0000-single-file-components.md
```diff
@@ -0,0 +1,111 @@
...
+1. Officially not support single file component/template definitions.
+2. Support single file component/template definitions with a pattern different from
+  the tagged template string that works today.
+
+#### Other framework solutions:
+- Templates declared using property `template` and strings (Vue).
+- Templates declared with JSX (React).
+
+#### Impact of not doing:
+Missed developer convenience where other frameworks offer single file
```
locks: You mention right away in the Summary that this is already possible to do, so I'm not sure this concern is valid.

text/0000-single-file-components.md
```diff
@@ -0,0 +1,111 @@
...
+component+template patterns and Ember does not.
+
+# Unresolved questions
+
+1. Can this pattern be supported when new Glimmer components land?
```
locks: There is no active spec for Glimmer components, I would raise this concern in an active RFC, otherwise we'll get bogged down in speculation.

text/0000-single-file-components.md
```diff
@@ -0,0 +1,111 @@
...
+})
+```
+
+For proper syntax highlighting of handlebars inside the tagged template literal our
+team has created highlighters for
+[Sublime](https://github.com/healthsparq/sublime-ember-syntax),
+[Atom](https://atom.io/packages/language-ember), and
+[VSCode](https://marketplace.visualstudio.com/items?itemName=dhedgecock.ember-syntax).
+
+A proposed change to this pattern is allowing the component template to be declared
```
locks: I want to note that there has been a Component property named `template`, which might confuse users.
Also, the way to reason about it being the layout and not the template is that the template is what's passed into the block form of a component and rendered in `{{yield}}`.

text/0000-single-file-components.md
evrowe: I think one of the concerns we've had while implementing this is that while it is supported/possible, it's not officially documented and therefore we don't know how "officially" supported this is (e.g. will support be removed in a major upcoming release that will cause us to have to rework these components?)

We've found this technique to be hugely beneficial to our workflow and codebase so ultimately I think the aim of this RFC is really two things:

1. Get a proper idea of just how well supported this pattern is and whether it is planned to continue to support it long term (or to start the conversation/thinking about that subject)
2. If the pattern will continue to be supported, we think it would be beneficial for this pattern to receive official documentation so that a larger group of people can discover it as an option.

This pattern has also been a huge help to us in terms of teaching Ember to devs with React experience.

DHedgecock: Although it's possible I think it would be good to get a definitive yes or no on if the framework intends to support it. For example, it was entirely possible to reference a namespaced component pre-Glimmer 2 like: `{{name-spaced.component}}`, but support for this wasn't intended and surprised a few people when it stopped working after 2.10.

Similarly, this pattern currently works, but as far as I know, is not 'sanctioned'. I think it's a really nice developer convenience that people would like, but want to make sure it has the 👍 as a long term pattern.

I suppose it doesn't need a full RFC, but it seemed like a different enough convention that it was worth putting out there.

text/0000-single-file-components.md
```diff
@@ -0,0 +1,111 @@
...
+
+Our team has been referring to this pattern as 'inline template declaration'. The
+syntax of declaring a template using a tagged template string can be documented in
+the Templates section of the Ember guides, perhaps in a new section called 'Declaring
+Templates Inline'. It could be presented as an alternate to defining a component's
+template using a separate .hbs file.
+
+Although this would change learning Ember as a new user, I personally think that it
+would make learning Ember easier if it is used as the primary pattern for creating
+components. Although the component generator creates the component and template files
```
locks: Module Unification was designed with this in mind, hence the heavy focus on colocation.

text/0000-single-file-components.md
```diff
@@ -0,0 +1,111 @@
...
+in the correct file locations, it may not be immediately clear to a new user that
+the file structure is required, and why the structure is required. Assigning a
+template to the `layout` (or `template`) property doesn't require understanding file
```
locks: If you are using Ember CLI, the file structure kinda is your application's architecture. Do we gain that much by avoiding the concept for components? Also see previous comment about MU.

evrowe: Absolutely the file structure is the architecture, generally speaking. That's one of the main reasons we chose Ember over three years ago and have stuck with it to this day.

Our experience in teaching Ember to new developers, however, has shown that the current pattern of separating templates from their functional code has proven confusing. This is especially true across different apps at different organizations, where some use pod structure for components (which is better in terms of keeping track of files/contexts) and others don't.

As an application grows in scale, and the number of components increases into the hundreds, being able to reduce the number of files in a repo by consolidating javascript and template code into one place becomes more and more beneficial, especially as the number of routes and corresponding templates increase.

I can understand the notion that bundling the template code for components technically "breaks" the established convention with routes and partials, and thus would require further explanation as to why components might implement templates differently than routes. Our feeling is that as more and more of our app lives in components over time (especially as we refactor controller code into high-level "container" components), this point becomes less of a concern. If/when routable components ever land in this framework, the distinction potentially becomes moot.

Granted, this is somewhat speculative, but we think it's worth considering (and we hope the core team will feel the same way)

locks: > especially as we refactor controller code into high-level "container" components

We should talk off-thread, I'm interested in knowing what you're doing.

text/0000-single-file-components.md
locks: I fully understand wanting to have some assurance that the pattern will be supported, however, it seems to be that the RFC is laying down the argument for something slightly different and more involved.
I would say that as long as `layout` is a non-deprecated, public property, that you will be able to do this. Worst case scenario would be maintaining the precompiler yourselves, which, Ember already has a precompiler for templates.

text/0000-single-file-components.md
evrowe: @locks Sure thing, would you prefer Slack or email?

text/0000-single-file-components.md
```diff
@@ -0,0 +1,111 @@
...
+lookups based on directory location.
+
+# Drawbacks
```
locks: I would like to bring up that using this patterns has some implications on component composability. See https://github.com/emberjs/ember.js/issues/15156 for a thread that touches on this.

text/0000-single-file-components.md
locks: Slack!

text/0000-single-file-components.md
evrowe: @locks That's fair; would it be more appropriate to make a PR against the official guides to add official documentation, supplemented with a PR to Ember CLI that adds a flag to the component generator to create a component with an inline template?

I'd be more than happy to work on those items myself, though we would want to make sure the core team/appropriate people feel good about that before doing so.

text/0000-single-file-components.md
DHedgecock: For later reference: We have begun using 'container' and 'presentational' components similar to the React pattern. It is helpful with orchestrating passing data based on feature flags and switching between presentation components depending on app state.
https://medium.com/@dan_abramov/smart-and-dumb-components-7ca2f9a7c7d0

It fits very nicely into the Ember DDAU conventions and makes testing dynamic displays a bit easier.

text/0000-single-file-components.md
DHedgecock: Interesting, am I right in understanding that using `layout` means that any subclass of that component will automatically inherit the defined template, whereas using a separate template file means that subclassed components will default to the `{{yield}}` template and must have the parent layout set explicitly?

text/0000-single-file-components.md
mmun: I strongly agree with the goal of this paragraph. I also agree with @locks that it should be done with care.

What is currently informally called "layout" and "template" should become "template" and "block". There's a few reasons for this.

1) I think that when most people think of a template they think of an entire `hbs` file.
2) When people say "component's layout", they also think of a template. I think it's fine to use both words when talking about components informally, but the docs should use "template".
2) Everyone already calls blocks "blocks".
3) Moreover, named blocks are... "blocks" :D

topaxi:
When I read single file, I also expected to find something about styles/css in here :)

joukevandermaas:
One thing to note is that updating this `layout` property on a component that has already been rendered does nothing (see https://github.com/emberjs/ember.js/pull/11315). For example, it cannot be a `computed` or changed based on an action.

This was initially not obvious to me when I discovered the `layout` property.

text/0000-single-file-components.md
Kerrick: We have been doing a very similar pattern in our very large Ember app -- we've been using the terms "Domain" components and "UI" components internally. The domain components replaced Controllers for us in anticipation of routable components, and the UI components are the reusable kind that we used to consider components back when Controllers were still heavily used.

Gaurav0:
Is this properly a concern of Ember or of ember-cli?

DHedgecock:
Compiling the template using a tagged template literal is handled through CLI, but I'm proposing to document the pattern as part of Ember, rather than in the Ember CLI docs.

knownasilya:
I have a blueprint for this in npm https://www.npmjs.com/package/ember-inline-component

text/0000-single-file-components.md
efx: Echoing @evrowe 's comment entirely for our team of 5 working on an application with ~176 components. The amount of files proved challenging while learning, but also introduces overhead in our day to day. Colocation seems like it can help here.

efx:
@DHedgecock rendered link is broken. could you update for those of us late to the party?

knownasilya:
FYI https://github.com/psbanka/eslint-plugin-hbs supports inline hbs syntax linting

aureliosaraiva:
I like this convencion. I think that can help on adoption the Ember by community, with less barrier of complexity with various different files for create a simple component.

urbany:
I created an addon related to this that is still a work-in-progress (waiting for module unification to be released to finish it)
https://github.com/urbany/ember-cli-single-file-components
https://github.com/urbany/ember-sfc-syntax-highlight

courajs:
@urbany That's awesome! That's more what I pictured when I saw the title of this RFC. I think if we're to make this more official, it should be through some addon like yours.

It does just *feel* nicer to me, but I think it's also a real ergonomic difference, because of the caveats @joukevandermaas mentions above (can't be a computed, or be changed once it's rendered).

But more importantly, splitting the file apart in the preprocess step should allow it to interoperate with other addons, like existing template linters.
I think it would be better to let the community play around with multiple formats via addon, then we could officialize it if one gains widespread adoption.

I'd also appreciate formalizing the semantics of the `layout` property, but I don't think we need to rely on it for single-file components.

psbanka:
@locks – do you know where this RFC stands? It seems to be inactive without any firm indication of where it should go. I think I hear you saying that since `layout` is not deprecated, that there is very little to worry about for existing developers who want to use this pattern. But there is also a sense that it would be nice to be able to formally document this approach for writing Ember components and use it in learning materials which (I believe?) won't happen unless this a more formally ratified approach to component-development?

locks:
@psbanka 

> @locks – do you know where this RFC stands?

I am not the RFC author, I can only answer as a commenter on this RFC. I am confused as to what the RFC intends to achieve. `layout` is public API, so yes, it is supported until further notice. If the current API documentation is insufficient, the best way to raise all boats is to help document it.

> But there is also a sense that it would be nice to be able to formally document this approach for writing Ember components and use it in learning materials which (I believe?) won't happen unless this a more formally ratified approach to component-development?

Speaking from a learning team perspective, I do not think that documenting a `layout`-based approach would be a fit for the Guides, because the ergonomic losses are bigger than the gains IMO, and I do not consider it a "single-file component" approach in the Vue/JSX sense.
I don't know what you mean by "formally ratified approach".

---

@courajs 

> I think it would be better to let the community play around with multiple formats via addon, then we could officialize it if one gains widespread adoption.

💯 

> I'd also appreciate formalizing the semantics of the layout property, but I don't think we need to rely on it for single-file components.

Can you expand what you mean?

mansona:
Howdy folks 👋  I just wanted to weigh in here with my perspective/experience on the idea of "single file components", it seems like most people here are _for_ the idea and I would like to give an alternative perspective.

I think if we started to officially teach this approach we would lose some of the things that are the "Magic" of the Ember Way™️. Two most notably using "Rule of Least power" and providing accessibility for *non-expert* Javascript developers.

One of the anecdotes that I always use when talking about the awesomeness of Ember is that in my company we trained up an "HTML and CSS only" designer to be 100% proficient in building an Ember app. Having a clear separation between what is the template and what is code allows you to keep the cognitive load low enough so that you can teach people to be fully blown Ember developers without much Javascript knowledge as a pre-requisite.

A lot of the benefits of other frameworks are touted as being "just javascript" which, sure, opens the door for intermediate and above JS developers to be more efficient, but closes some of the doors for beginners and people learning JS.

Maybe this raises the possibility of reviving the "cookbook" style of documentation @locks where we can show other approaches without needing them to become "official Ember Way™️" docs?

MelSumner:
I strongly urge anyone teaching folks to write code to please consider that programmers need to learn resilience. Trying to change the way Ember works because "it's hard to learn something that isn't done exactly like React" is not going to do anyone any favors in the long run. 

Learning how to learn, or that every system will be different but is probably trying to do similar things, is an absolutely vital skill for every developer.

JonForest:
FWIW I really like what is trying to be achieved here.  Having an option that we know is not going away to have everything in one file at the time of authoring would be awesome.  Way too much switching between many files right now.

However, I would also have hoped for some css capability - which there is an add-on for.
I actually like the idea of @urbany 's add-on, and would love to see that get a bit of traction.

+1 for raising this issue.  It's a good conversation to be having.

psbanka:
I believe it's a serious error to try to bring CSS into this discussion for the following reasons:

* Using @urbany 's approach, you're going to wind up with a file format that is completely alien to all editors and github. It's going to be a tooling nightmare.  We will lose:
  *  syntax-highlighting for handlebars, css and javascript
  * linting for handlebars, css and javascript 
  * TypeScript support
* If we can simply combine the template and the code together in a single file, that's a huge win and it prevents the great being the enemy of the good.
* There is far from a consensus on The Right Way to do CSS at this time, and whatever approach we pick is probably going to be wrong. Should we support LESS? SASS? PostCSS? Forget it.

Simply embedding the template in a component as a property allows us to preserve _all_ of our tooling and _in most cases_ allows us to use a single file.

knownasilya:
As a response to your concerns, see https://vuejs.org/v2/guide/single-file-components.html @psbanka

DHedgecock:
Hello, I am cleaning up outstanding PRs I have open that I am no longer active in. Best of luck!

PR #230, opened 2017-06-12
Link component hook

SpikedKira:
[rendered](https://github.com/SpikedKira/rfcs/blob/linkComponentHook/text/0000-Link-Component-Invoke-Hook.md)


0000-template.md
```diff
@@ -1,52 +0,0 @@
-- Start Date: (fill me in with today's date, YYYY-MM-DD)
```
rwjblue: This file shouldn't be deleted, can you bring it back?

SpikedKira: embarrassing :o

SpikedKira:
@rwjblue unless I'm missing something, the 3rd point is already covered.  The new `eventAction` would be a public method that calls the private `_invoke` method.  So if a user overrides, they can call `_super()` to use the default after doing any logging or w/e.

rwjblue:
Moving into FCP to close for this, I think the right path forward for folks here is to **not** use `{{link-to` at all (instead using https://github.com/emberjs/rfcs/blob/master/text/0391-router-helpers.md).

PR #233, opened 2017-06-29
Replace `any` in favour of `some` to align with regular JS

cibernox:
[Rendered](https://github.com/cibernox/rfcs/blob/add_enumerable_some/text/0000-enumerable-some.md)

### DIFF
--- a/text/0000-enumerable-some.md
+++ b/text/0000-enumerable-some.md
@@ -1,0 +1,56 @@
+- Start Date: 2017-06-29
+- RFC PR: (leave this empty)
+- Ember Issue: (leave this empty)
+
+# Summary
+
+[`any`](https://emberjs.com/api/classes/Ember.Array.html#method_any) has been
+implemented on `Ember.Enumerable` for years, however native javascript arrays have
+`Array.prototype.some` to perform the same task.
+
+This RFC proposes the addition of `Ember.Enumerable#some`, the aliasing of the old `Ember.Enumerable#any` to it,
+and encourage the usage of the first either in the docs or with deprecation to be as aligned as possible
+with native javascript arrays.
+
+# Motivation
+
+Motivation is to stay in line with web standard.
+
+# Detailed design
+
+The first step is to rename `any` to `some` and recreate `any` as simply an alias
+of the new method.
+
+The second and optional step is to encourage, possibly with a deprecation, the usage of `some`
+over `any`.
+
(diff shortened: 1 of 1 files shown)


cibernox:
@alexspeller You might have something to say as this is almost a copy of another RFC you made, with the exception that we don't have to polyfill this feature since it is available even in IE9.

alexspeller:
Definitely in favour, and also in favour of renaming isAny to isSome for consistency. I think the small amount of very easily updatable churn is worth it to keep more in line with standards.

text/0000-enumerable-some.md
```diff
@@ -0,0 +1,56 @@
...
+and encourage the usage of the first either in the docs or with deprecation to be as aligned as possible
+with native javascript arrays.
+
+# Motivation
+
+Motivation is to stay in line with web standard.
+
+# Detailed design
+
+The first step is to rename `any` to `some` and recreate `any` as simply an alias
```
Serabe: s/alias/deprecatingAlias/

https://emberjs.com/api/classes/Ember.computed.html#method_deprecatingAlias

cibernox: I wanted to leave the deprecation part optional, that is why I mentioned the deprecation as a separate step. In case we decide to implemente and deprecate, `deprecatingAlias` is indeed how I'd implement this.

Serabe:
One quick comment on this. Though documentation still says otherwise, `Array.any` is currently returning the first truthy value:

```js
let arr = emberA([undefined, null, false, 5]);
let result = arr.any(o => o);

assert.equal(result, 5);
```

Can you change the RFC to accomodate to `some` returning always a boolean? I mean, the change would be something like:

```js
{
  some(...args) {
    return !!this.any(...args);
  }
}
```

And there might be people out there relying on the fact that `any` returns it, so there needs to be a guide to tell them to use `find` instead.

Thank you!

locks:
To build on top of @Serabe's latest comment, I don't think this should be framed as aliasing, given that the two methods have slightly different semantics in practice.
Or in other words: https://github.com/emberjs/rfcs/pull/136#issuecomment-211128494.

locks:
With Octane shipped, we are no longer looking to make changes to this part of the framework. Thanks to all involved for your effort!

NullVoxPopuli:
This is because any only exists on ember array, yeah?
And octane doesn't have prototype extensions, so regular arrays don't have any.

PR #235, opened 2017-07-12
Release Channel Names

Turbo87:
[rendered](https://github.com/Turbo87/rfcs/blob/25eb943eb61567478a555d9fbf6fca121ca455b7/text/0000-release-channel-names.md)

### DIFF
--- a/text/0000-release-channel-names.md
+++ b/text/0000-release-channel-names.md
@@ -1,0 +1,56 @@
+- Start Date: 2017-07-12
+- RFC PR: (leave this empty)
+- Ember Issue: (leave this empty)
+
+# Summary
+
+The release channel names in the Ember core projects (Ember, Ember Data, and
+Ember CLI) today are: `release`, `beta` and `canary`. This RFC proposes
+to change them to: `stable`, `beta` and `master`.
+
+# Motivation
+
+Why are we doing this? What use cases does it support? What is the expected
+outcome?
+
+The `release` release channel name is confusing since it literally applies
+to *all* releases, not just the stable ones.
+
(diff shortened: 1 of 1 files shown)


workmanw:
I don't mind the change of `release` => `stable`. I'm neutral on this one.

I think `canary` => `master` could lead to more confusion. I've worked on several projects where `master` is treated as the release branch. In those cases, it's been mostly devs working from feature branches and merge into master when it's ready to ship. I know many OSS projects don't work that way, but I think reusing the term `master` is problematic for this reason.

Just my $0.02.

rwjblue:
This feels a bit like just extra churn for little value me. What am I missing? 

Also, I think that "canary" evokes the exact mental model that we want here (and is in line with other players in the ecosystem ala "chrome canary"). Whereas "master" is quite confusing for most package maintainers (where the master branch is almost overwhelmingly the branch used for stable releases / publishing).

ming-codes:
The name "alpha" was mentioned in a few places. How is this different from canary?

locks:
@ming-codes which places?

ming-codes:
https://www.emberjs.com/blog/2015/08/13/ember-2-0-released.html

https://ember-twiddle.com

Ember Twiddle has an alpha Ember version.

Searches through Ember GitHub https://github.com/emberjs/ember.js/search?utf8=✓&q=alpha&type=

bcardarella:
At this point I think this change would just result in extra confusion. Maybe the proposed names would be mean clear to new adoptees of Ember, but for existing practitioners it would be yet another change they have to absorb into their mental model of the Ember eco-system as a whole. If anything, we should be trying to reduce the churn not increase it.

PR #237, opened 2017-07-20
Remove Ember.Map, Ember.MapWithDefault and Ember.OrderedSet

Serabe:
[rendered](https://github.com/Serabe/rfcs/blob/deprecate-ember-map/text/0000-deprecation-ember-map.md)

### DIFF
--- a/text/0000-deprecation-ember-map.md
+++ b/text/0000-deprecation-ember-map.md
@@ -4,7 +4,7 @@
 
 # Summary
 
-This PR proposes the deprecation of the following classes:
+This RFC proposes the deprecation of the following classes:
 
 - `Ember.OrderedSet`
 - `Ember.Map`
@@ -14,7 +14,7 @@ These classes need to be moved to an external addon given they are private class
 
 # Motivation
 
-Theseclasses has not been used in Ember itself for a while now. They have always been private but they are used in a few addons, and in particular Ember Data is using them.
+These classes have not been used in Ember itself for a while now. They have always been private but they are used in a few addons, and in particular Ember Data is using them.
 
 # Transition Path
 
--- a/text/0000-deprecation-ember-map.md
+++ b/text/0000-deprecation-ember-map.md
@@ -18,7 +18,11 @@ These classes have not been used in Ember itself for a while now. They have alwa
 
 # Transition Path
 
-The classes will be extracted to an addon and the ones in Ember will be deprecated.
+`Ember.Map` and `Ember.MapWithDefault` will be deprecated and not extracted. There is already an addon with `Ember.OrderedSet` extracted ([@ember/ordered-set](https://github.com/emberjs/ember-ordered-set)).
+
+Ember Data is quite likely the biggest project using these classes. There is already a PR that needs merging before deprecating `Ember.Map` and `Ember.MapWithDefault` https://github.com/emberjs/data/pull/5255. Ember Data still needs to migrate to `@ember/ordered-set` to its relationship logic.
+
+Finally, these classes can be deprecated in Ember itself.
 
 # How We Teach This
 
--- a/text/0000-deprecation-ember-map.md
+++ b/text/0000-deprecation-ember-map.md
@@ -18,11 +18,11 @@ These classes have not been used in Ember itself for a while now. They have alwa
 
 # Transition Path
 
-`Ember.Map` and `Ember.MapWithDefault` will be deprecated and not extracted. There is already an addon with `Ember.OrderedSet` extracted ([@ember/ordered-set](https://github.com/emberjs/ember-ordered-set)).
+`Ember.Map` and `Ember.MapWithDefault` will be deprecated and not extracted, but not before the fix mentioned in the following paragraph is landed in Ember Data. There is already an addon with `Ember.OrderedSet` extracted ([@ember/ordered-set](https://github.com/emberjs/ember-ordered-set)).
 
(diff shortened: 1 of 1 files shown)


text/0000-deprecation-ember-map.md
```diff
@@ -0,0 +1,37 @@
...
+
+These classes need to be moved to an external addon given they are private classes and unused in Ember.js itself.
+
+# Motivation
+
+These classes have not been used in Ember itself for a while now. They have always been private but they are used in a few addons, and in particular Ember Data is using them.
+
+# Transition Path
+
+The classes will be extracted to an addon and the ones in Ember will be deprecated.
```
stefanpenner: what are the details on how ember-data will work, and how we can be backwards compat with ED. It's possible, but that needs to be fleshed out in this RFV

text/0000-deprecation-ember-map.md
```diff
@@ -0,0 +1,37 @@
...
+
+This should not impact many codebases.
+
+# Drawbacks
+
+This requires cooperation with Ember Data, the main user of these classes. It would be nice to have moved Ember Data to using the addon before releasing Ember with the deprecation so the average user does not see any deprecation warning.
+
+# Alternatives
+
+Other option would be moving these classes to Ember Data itself or leaving things as they are now.
```
jrjohnson: 👍 for this option. I recall the last time something in heavy use in Ember Data was deprecated I was hit with a terrifying number of notices that I could do nothing about. While methods now exist to hep with this it should still be avoided if possible.

text/0000-deprecation-ember-map.md
```diff
@@ -0,0 +1,37 @@
...
+
+# How We Teach This
+
+These classes being private would make this simple than other deprecations. People were not supposed to be using a private API and the few that were, would just need to use a new addon.
+
+This should not impact many codebases.
+
+# Drawbacks
+
+This requires cooperation with Ember Data, the main user of these classes. It would be nice to have moved Ember Data to using the addon before releasing Ember with the deprecation so the average user does not see any deprecation warning.
```
jrjohnson: 👍 for this option. I recall the last time something in heavy use in Ember Data was deprecated I was hit with a terrifying number of notices that I could do nothing about. While methods now exist to hep with this it should still be avoided if possible.

text/0000-deprecation-ember-map.md
buschtoens: If that is in the scope of this RFC, it might be worthwhile to take a look at switching to the native ES6 `Map`. The `MapWithDefault` behavior can be easily achieved: [Babel REPL example](https://babeljs.io/repl/#?babili=false&evaluate=true&lineWrap=true&presets=es2016%2Ces2017%2Cstage-0&targets=&browsers=&builtIns=false&debug=false&code_lz=MYewdgzgLgBA-gEwKYDMCGBXANlAamrDJAMRACcYBeGAZQE8BbAIxCwAoByZJ9bPAohwCUAbgBQY4FjQQIMALJoADgHUAllAAWAEVSYcMJAA8oSMAjmKlMAN5iYMUJChkMwKOTbJeOfIRLkADQwGkhkaExYSEK29g4hKDBsUHRKSCCJ3vr8_qQUAISU1BwoGGDuauDCsfHxWmQgAO4wYEjNACqpSACiZA1knACaIBgwDBjQMEoNAG5qyDBoMFl8fkR5MKXlUJVgMB4Kyupauj5QwuK1AL5xDhAYaQOh4ZHRl3WaahAA2oh6qwIAmQALpUZb_XyAvLvG4OOIAcyQUDYAGskHQYnZampEslPhAAHSaGSo9FCTG3eJkJEYMh7e6PAmI5FojHvByw-JOSYzQFgrRfX4rSG5cjA0lsykCwkQJES4K8_yiSnUqC0vaKogwsQ3STgSYMZRg1rNKzHHQQlnoqgAPiSNhgrOCwGpaFMCAAglAAFwtNowbRupAwK7k8R6yCsJAErAgeFsQ1KJlykogEDCZXcqMxuMJ5QEswuNRICBsMNAA)

I would be available for such a PR in Ember Data. :raising_hand_man:

locks: @buschtoens I don't think it is. Switching to native `Map` seems like a big undertaking, especially if we are to support IE9 and family :P

rwjblue:
I'd like to see this one start to gain traction now that 3.0 is "out the door".

@Serabe - Thanks for the recent round of updates! I think the task list in the transition path looks pretty close to what I was thinking, but I'd like to make it clearer that the deprecation would _only_ be landed once a released version of ember-data that included the fixes (to prevent the deprecation) shipped. Do you think that is clear enough in the current prose?

rwjblue:
Changes landed in ember-data to remove direct usage of these classes:

*  https://github.com/emberjs/data/pull/5372 - Migrates ember-data to use the `@ember/ordered-set` addon (instead of relying on `Ember.OrderedSet`)
* https://github.com/emberjs/data/pull/5255 - Migrates to using native `Map`, and removes any Ember Data usage of `Ember.Map` or `Ember.MapWithDefault`.

Both of those PR's should wind up included in ember-data@3.1.

rwjblue:
Just discussed this with the rest of the core team, and with the recent round of changes and the work in ember-data to detangle from these classes we think this is ready for final comment period.

rwjblue:
Lets do it!

PR #238, opened 2017-07-25
Ember Data JavaScript Modules API

locks:
[Rendered](https://github.com/emberjs/rfcs/blob/ember-data-modules/text/0000-ember-data-javascript-module-api.md)

Originally authored by @bmac.

### DIFF
--- a/text/0000-ember-data-javascript-module-api.md
+++ b/text/0000-ember-data-javascript-module-api.md
@@ -65,66 +65,66 @@ byt theyhave their own name and namespace in the Ember container.
 ### `@ember-data/store`
 | Global                                | Module                                                                      |
 |---                                    | ---  |
-| `DS.Store` | `import Store from 'ember-data/store'` |
-| `DS.AdapterPopulatedRecordArray` | `import AdapterPopulatedRecordArray from 'ember-data/store/adapter-populated-record-array'` |
-| `DS.DebugAdapter` | `import DebugAdapter from 'ember-data/store/debug-adapter'` |
-| `DS.FilteredRecordArray` | `import FilteredRecordArray from 'ember-data/store/filterd-record-array'` |
-| `DS.ManyArray` | `import ManyArray from 'ember-data/store/many-array'` |
-| `DS.RecordArray` | `import RecordArray from 'ember-data/store/record-array'` |
-| `DS.RecordArrayManager` | `import RecordArrayManager from 'ember-data/store/record-array-manager'` |
-| `DS._initializeStoreService` | `import { initializeStoreService } from 'ember-data/store'` |
-| `DS.normalizeModelName` | `import { normalizeModelName } from 'ember-data/store'` |
-| `DS._setupContainer` | `import { setupContainer } from 'ember-data/store'`
+| `DS.Store` | `import Store from '@ember-data/store'` |
+| `DS.AdapterPopulatedRecordArray` | `import AdapterPopulatedRecordArray from '@ember-data/store/src/src/adapter-populated-record-array'` |
+| `DS.DebugAdapter` | `import DebugAdapter from '@ember-data/store/src/src/debug-adapter'` |
+| `DS.FilteredRecordArray` | `import FilteredRecordArray from '@ember-data/store/src/filterd-record-array'` |
+| `DS.ManyArray` | `import ManyArray from '@ember-data/store/src/many-array'` |
+| `DS.RecordArray` | `import RecordArray from '@ember-data/store/src/record-array'` |
+| `DS.RecordArrayManager` | `import RecordArrayManager from '@ember-data/store/src/record-array-manager'` |
+| `DS._initializeStoreService` | `import { initializeStoreService } from '@ember-data/store'` |
+| `DS.normalizeModelName` | `import { normalizeModelName } from '@ember-data/store'` |
+| `DS._setupContainer` | `import { setupContainer } from '@ember-data/store'`
 
 ### `@ember-data/model`
 | Global                                | Module                                                                      |
 |---                                    | ---  |
-| `DS.Model` | `import Model from 'ember-data/model'` |
-| `DS.Errors` | `import Errors from 'ember-data/model/errors'` |
-| `DS.InternalModel` | `import InternalModel from 'ember-data/model/internal-model'` |
-| `DS.Relationship` | `import Relationship from 'ember-data/model'` |
-| `DS.RootState` | `import RootState from 'ember-data/model/root-state'` |
-| `DS.Shapshot` | `import Snapshot from 'ember-data/model/snapshot'` |
-| `DS.attr` | `import { attr } from 'ember-data/model'`|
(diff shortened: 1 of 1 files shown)


text/0000-ember-data-javascript-module-api.md
```diff
@@ -0,0 +1,130 @@
...
+| `DS.RESTAdapter` | `import RESTAdapter from '@ember-data/adapter/src/rest'` |
+| `DS.TimeoutError` | `import TimeoutError from '@ember-data/adapter/src/timeout-error'` |
+| `DS.errorsHashToArray` | `import { errorsHashToArray } from '@ember-data/adapters'` |
+| `DS.errorsArrayToHash` | `import { errorsArrayToHash } from '@ember-data/adapters'` |
+
+### `@ember-data/serializer`
+| Global                                | Module                                                                      |
+|---                                    | ---  |
+| `DS.Serializer` | `import Serializer from '@ember-data/serializer'` |
+| `DS.EmbeddedRecordsMixin` | `import EmbeddedRecordsMixin from '@ember-data/serializer/src/embedded-record-mixin'` |
```
josemarluedke: This should be `@ember-data/serializer/src/embedded-records-mixin` (note records as plural)

text/0000-ember-data-javascript-module-api.md
```diff
@@ -0,0 +1,130 @@
...
+
+Align Ember Data module API with [RFC #176](https://github.com/emberjs/rfcs/blob/master/text/0176-javascript-module-api.md).
+
+# Motivation
+
+This document proposes changes to the modules exported by Ember Data to make it consistent with the changes to the Ember module API proposed in [RFC #176](https://github.com/emberjs/rfcs/blob/master/text/0176-javascript-module-api.md).
+
+# Detailed design
+
+Ember Data will stick with 1 top level namespace of `@ember-data`. Under it there are 6 nested module
```
Turbo87: why `@ember-data` instead of `@ember/data` as proposed in https://emberjs.com/blog/2017/04/05/emberconf-2017-state-of-the-union.html#toc_ember-in-2017?

text/0000-ember-data-javascript-module-api.md
```diff
@@ -0,0 +1,130 @@
...
+
+# Motivation
+
+This document proposes changes to the modules exported by Ember Data to make it consistent with the changes to the Ember module API proposed in [RFC #176](https://github.com/emberjs/rfcs/blob/master/text/0176-javascript-module-api.md).
+
+# Detailed design
+
+Ember Data will stick with 1 top level namespace of `@ember-data`. Under it there are 6 nested module 
+namespaces where exposed components could live. Someday these namespaces could be moved into their own 
+top level scoped package (for example `@ember-data/serializer` could become `@ember-data/serializer`) however,
```
Turbo87: what's the difference between `@ember-data/serializer` and `@ember-data/serializer`?

text/0000-ember-data-javascript-module-api.md
```diff
@@ -0,0 +1,130 @@
...
+Why should we *not* do this? Please consider the impact on teaching Ember,
+on the integration of this feature with other existing and planned features,
+on the impact of the API churn on existing apps, etc.
+
+There are tradeoffs to choosing any path, please attempt to identify them here.
+
+# Alternatives
+
+The easiest alternative would be to continue using the current exports.
+Alternatively, we could use the current shims, which do not follow the RFC #176 heuristics as closely.
```
Turbo87: what current shims?

text/0000-ember-data-javascript-module-api.md
```diff
@@ -0,0 +1,130 @@
...
+
+# Addenda
+
+## Addendum 1 - Table of Module Names and Exports by Global
+
+### `@ember-data/store`
+| Global                                | Module                                                                      |
+|---                                    | ---  |
+| `DS.Store` | `import Store from '@ember-data/store'` |
+| `DS.AdapterPopulatedRecordArray` | `import AdapterPopulatedRecordArray from '@ember-data/store/src/adapter-populated-record-array'` |
```
Turbo87: `src` in the module import path is needed due to the new module unification semantics, which obviously makes it a little harder to build nice looking APIs based on import paths. IMHO we should first figure out how we could accomplish nice looking paths and then afterwards think about new Ember Data module APIs. Otherwise we might create more churn than needed.

text/0000-ember-data-javascript-module-api.md
locks: Bad edit from the draft.

text/0000-ember-data-javascript-module-api.md
locks: `import attr from 'ember-model/attr;` and company.

t-sauer:
What's the reasoning behind using the `@ember-data` scope?  
As far as I understood the reason behind the `@ember` scope was to work towards standalone modules (e.g. `@ember/array`) which can be removed from applications if they are not needed.  
Does the same idea apply to Ember Data? I would assume that if Ember Data is used in a project, it is probably used entirely (e.g. not only models and transforms). So probably most (if not all) applications have 6 packages just to get Ember Data working, which seems to be unnecessary complicated for new people ("Why do I need to add six packages just to use this one library?").  

What probably makes sense, is that most people don't need all the adapters and serializers that are included by default (I don't need the RESTAdapter if I only use the JSONAPIAdapter) and they could be left out. But in my opinion this should be solved by a different solution than splitting up the whole library into small chunks. For example tree shaking could make them have no costs for end users and they could stay in the main package or maybe they could be separated into their own addons. But I guess this would be a concern of a different RFC.  

So I think it makes more sense, to keep Ember Data as a single package instead of using a scope with multiple packages, to avoid unnecessary complexity.

tomdale:
One thing we discussed in the core team meeting today was keeping Ember data under the `@ember` scope, to communicate that it's an "official," core-supported package. Then we would have things like:

* `@ember/data/store`
* `@ember/data/model`

This would also allow us to break out modules we no longer wish to ship with the default distribution into packages like:

* `@ember/json-api-adapter`

lolmaus:
The link in the top comment is incorrect. Should be https://github.com/emberjs/rfcs/blob/ember-data-modules/text/0000-ember-data-javascript-module-api.md .

rwjblue:
Thanks, updated.

Samsinite:
@t-sauer while this RFC doesn't explicitly mention it (maybe it should), a meta package could be used to include all of the 6 ember-data dependencies (hopefully keeping the name `ember-data`) to make it so only one module is explicitly depended on in the package.json file.

locks:
@samsinite we already have the meta package named `ember-data`, and like RFC #176, it isn't strictly about how the packages are published, as we are scheduled to land the new modules api in 2.16 through the `ember-source` addon.

Samsinite:
@locks exactly what I assumed, was just addressing t-sauer's concern. Thanks for clarifying!

locks:
I am no longer the right person to push this forward, and it would necessitate a complete rewrite at any rate.

PR #240, opened 2017-07-29
ES Classes

pzuraq:
[rendered](https://github.com/pzuraq/rfcs-1/blob/es-classes-rfc/text/0000-es-classes.md)

### DIFF
--- a/text/0000-es-classes.md
+++ b/text/0000-es-classes.md
@@ -197,7 +197,7 @@ console.log(bar.get('bar')); // 'baz'
 ## Observers and Events
 
 Observers and events both fail to work properly when using ES Class syntax. The root
-of the issue here is how `Ember.Objec`t works at a fundamental level, and will require
+of the issue here is how `Ember.Object` works at a fundamental level, and will require
 some refactoring to fix.
 
 Currently, each time `Ember.Object.extend` is used, it stores the list of mixins and
--- a/text/0000-es-classes.md
+++ b/text/0000-es-classes.md
@@ -10,9 +10,11 @@ continue to push forward with experimental Javascript features. This includes:
 
 * Making the class `constructor` function a public API
 * Modifying some of the internals of `Ember.Object` to support existing features
+and make the usage of ES Classes cross-compatible with `Ember.Object`
 
 It does _not_ propose additions in the form of helpers or decorators, which should
-continue to be iterated on in the community as the spec itself is finalized.
+continue to be iterated on in the community as the spec itself is finalized. It also
+does not propose deprecating or removing existing functionality in `Ember.Object`.
 
 # Motivation
 
@@ -75,14 +77,35 @@ vanilla ES Classes or through `ember-decorators`, including:
 However, the following features either do not exist or do not work as a
 user familiar with `Ember.Object` would expect:
 
+* Extending from ES Classes using `extend`
 * Class properties
 * Mixins
 * Observers and events
 * Merged and concatenated properties
 
-Some of these can and should be supported by changes to `Ember.Object`, whereas
-others can be addressed by community solutions. We will address each one
-individually below.
+These features will require changes to `Ember.Object`
+
+## Extend
+
+Currently, once a class is defined using ES Classes it is not possible for users
+to extend it using the previous CoreObject style of writing and extending classes.
+This can limit the rate of adoption because ES Classes would become a trapdoor -
+once you begin using them, you must continue to use them. It would be a particularly
+thorny issue for addon developers, who may design components which their users
+expect to be able to extend and modify.
+
+This RFC proposes that `extend` be fixed on ES Classes to make them fully
+cross-compatible with the existing syntax. There are two general approaches to
+making this work:
+
+1. Modify CoreObject to use prototypes/ES Classes internally. This would bring
+CoreObject more inline with ES Classes, but would be a significant internal change.
+
+2. Modify CoreObject to have different behavior if it is extending an ES Class
+using `extend`.
+
+Both approaches should be explored and benchmarked to determine if there are an
+significant advantages to one over the other.
 
 ## Class Properties
 
@@ -283,65 +306,89 @@ not update, etc.)
 * Class property initialization can be changed such that properties are initialized
 after the constructor runs entirely, allowing them to be overwritten by values
 passed to `create`
-* Instead of fixing merged and concatenated properties, we can leave the behavior
-as is and allow community solutions to add it back in. Simple, general purpose
-decorators can be made which accomplish the same thing:
 
-  ```javascript
-  class FooComponent extends Ember.Component {
-    @concatenated classNameBindings = ['foo']
+# Topics for Future RFCs
 
-    @computed
-    get foo() { /* ... */ }
+While working on this RFC, some issues were brought into focus regarding existing
+features in CoreObject that are seen as problematic or unintuitive. In order to
+avoid bikeshedding these have been slated for discussion in future RFCs, but the
+discussion points have been included below.
 
-    @merged actions = {
-      bar() { /* ... */ }
-    }
-  }
-  ```
+## Merged and Concatenated Properties
 
-  They could also be accomplished more ergonomically with specialized decorators:
+Merged and concatenated properties are pain points for new Ember developers,
+specifically because they give no lexical hint that they are special in any way.
+Developers must know that these particular properties will be merged with the
+superclass, and there is no way to opt out of this behavior.
 
-  ```javascript
-  class FooComponent extends Ember.Component {
-    @className
-    @computed
-    get foo() { /* ... */ }
+With decorators, this same behavior can be accomplished in a much clearer and
+more straightforward way:
 
-    @action
+```javascript
+class FooComponent extends Ember.Component {
+  @concatenated classNameBindings = ['foo']
+
+  @computed
+  get foo() { /* ... */ }
+
+  @merged actions = {
     bar() { /* ... */ }
   }
-  ```
+}
+```
 
-  These approaches give us two distinct advantages over the previous behavior:
+They could also be accomplished more ergonomically with specialized decorators:
 
-  1. They are less magical. They indicate to new users that the properties are
-  special in some way, and ultimately they are just plain decorators, which are
-  compatible with ES Classes as a whole and can be reused anywhere.
-  2. They provide a way to opt out of their behavior. Currently, there is no easy
-  way to prevent properties which were marked to be merged from being merged,
-  meaning subclasses are stuck with the values that their superclass provided.
+```javascript
+class FooComponent extends Ember.Component {
+  @className
+  @computed
+  get foo() { /* ... */ }
 
-  In addition this approach is backwards compatible without a polyfill to at least Ember 1.11.
+  @action
+  bar() { /* ... */ }
+}
+```
 
-* Instead of fixing Observers and Listeners, we can leave the behavior as is and
-allow community solutions to add it back in. Class decorators can be used to modify
-the constructors at definition time and add the proto squashing:
+This approach has two distinct advantages over the existing behavior:
 
-  ```javascript
-  @evented
-  class Foo extends Ember.Object {
-    @on('init')
-    onInit() {
-      // do something
-    }
+1. It is less magical. The decorators indicate to new users that the properties are
+special in some way, and ultimately they are just plain decorators, which are
+compatible with ES Classes as a whole and can be reused anywhere.
+2. It provides a way to opt out of the behavior. Currently, there is no easy
+way to prevent properties which were marked to be merged from being merged,
+meaning subclasses are stuck with the values that their superclass provided.
+
+## Observers and Listeners
+
+Observers and event listeners are a powerful pattern that saw a lot of usage in Ember 1.
+However, it is now widely accepted that they are problematic when overused, and using
+computed properties and lifecycle hooks are better patterns in most cases.
+
+As such, rather than having events and observers turned on by default it may make
+more sense to have them be opt-in APIs. This could be accomplished by making new
+class decorators like so:
+
+```javascript
+@evented
+class Foo extends Ember.Object {
+  @on('init')
+  onInit() {
+    // do something
   }
-  ```
+}
+```
 
-  A proof of concept of this approach has been spiked out [on the ember-decorators repo](https://github.com/rwjblue/ember-decorators/tree/feature/observer-evented-class-decorators).
+Or it could be accomplished with new base classes that include the functionality:
 
-  This approach allows users to opt-in to eventing, and is backwards compatible without
-  a polyfill to at least Ember 1.11.
+```javascript
+class Foo extends EventedObject {
+  @on('init')
+  onInit() {
+    // do something
+  }
+}
+```
 
 # Unresolved questions
 
--- a/text/0000-es-classes.md
+++ b/text/0000-es-classes.md
@@ -1,0 +1,348 @@
+- Start Date: 2017-07-28
+- RFC PR: (leave this empty)
+- Ember Issue: (leave this empty)
+
+# Summary
+
+This RFC aims to solidify the usage of ES2015 Classes as a public API of Ember so
+that users can begin building on them, and projects like `ember-decorators` can
+continue to push forward with experimental Javascript features. This includes:
+
+* Making the class `constructor` function a public API
+* Modifying some of the internals of Ember.Object to support existing features
+
+It does _not_ propose additions in the form of helpers or decorators, which should
+continue to be iterated on in the community as the spec itself is finalized.
+
+# Motivation
+
+The Ember Object model has served its purpose well over the years, but now that
+ES Classes are becoming prevalent throughout the wider Javascript community
+it is beginning to show its age. With class properties at stage 3 and decorators at
+stage 2 in the TC39 process, classes are finally at a point where we can start
+integrating them into Ember.
+
+The [ember-decorators](https://github.com/rwjblue/ember-decorators) project has been
+experimenting with using ES Classes and filling out the Ember feature-set,
+allowing us to write Ember classes like so:
+
+```javascript
+export default class MyComponent extends Ember.Component {
+  didInsertElement() {
+    // do stuff
+  }
+
+  @computed
+  get foo() {
+    // do stuff
+  }
+
+  @action
+  bar() {
+    // do stuff
+  }
+}
+```
+
+Using classes makes Ember easier to teach and understand by normalizing it with
+standard Javascript coding practices, and allows us to share code and solutions
+with other frameworks and libraries. It also brings with it all the benefits of
+ES Class syntax:
+
+* More aligned with the greater Javascript community
+* Ability to share code more easily with other libraries and frameworks
+* Easier to statically analyze
+* Cleaner and easier to read (subjective)
+
+The Ember Object model already works extremely well with ES classes, as
+demonstrated above, but there several failure scenarios. Furthermore, because
+they are not officially supported as a public API, there is no guarantee
+that they will continue to work well. Thus, this RFC seeks to solidify the
+behavior of ES Classes so that the community can continue to experiment with
+new Javascript features and build on a stable API.
+
+# Detailed Design
+
+Many of the standard features of Ember classes work OOTB today, either with
+vanilla ES Classes or through `ember-decorators`, including:
+
+* Inheritance
+* Lifecycle hooks
+* Computeds
+* Injections
+* Actions
+
+However, the following features either do not exist or do not work as a
+user familiar with Ember Objects would expect:
+
+* Class properties
+* Mixins
+* Observers and events
+* Merged and concatenated properties
+
+Some of these can and should be supported by changes to Ember.Object, whereas
+others can be addressed by community solutions. We will address each one
+individually below.
+
+## Class Properties
+
+When using `Ember.Object.extend`, properties that are passed in on the object
+are assigned to the prototype of the class:
+
+```javascript
+const Foo = Ember.Object.extend({ bar: 'baz' });
+const foo = Foo.create();
+
+console.log(Foo.prototype.bar) // 'baz'
+foo.hasOwnProperty('bar') // false
+```
+
+This differs from the behavior of ES Class properties, which initialize their
+value on the instance of the class.
+
+```javascript
+class Foo {
+  bar = 'baz'
+}
+
+const foo = new Foo();
+
+console.log(Foo.prototype.bar) // undefined
+foo.hasOwnProperty('bar') // true
+```
+
+The above is essentially currently compiled down by Babel to the following:
+
+```javascript
+class Foo {
+  constructor() {
+    this.bar = 'baz';
+  }
+}
+```
+
+Property assignments like this are always done at the end of the constructor,
+and given the requirement that `super` must always be called before properties
+are assigned it is unlikely that this will change as the spec progresses.
+
+While one might intuitively expect class properties to function the same in
+ES Classes as they do with Ember Objects, this difference in behavior means that
+class properties will always be assigned after properties passed into `create`
+are initialized on the object, and thus will always win:
+
+```javascript
+const Foo = Ember.Object.extend({ testProp: 'default value' });
+
+class Bar extends Ember.Object {
+  testProp = 'default value'
+}
+
+const foo = Foo.create({ testProp: 'new value' });
+const bar = Bar.create({ testProp: 'new value' });
+
+console.log(foo.get('testProp')); // 'new value'
+console.log(bar.get('testProp')); // 'default value'
+```
+
+This behavior makes sense when you consider that it is equivalent to assigning
+values in `init` rather than on the object when it is defined. Rather than modify
+Ember.Object to treat class properties as default values, this RFC proposes that
+we accept the difference in behavior and utilize the constructor to allow users
+to set default values, as in the following example:
+
+```javascript
+class Foo extends Ember.Object {
+  constructor(props) {
+    props.testProp = props.testProp || 'default value';
+
+    super(props);
+  }
+}
+```
+
+This enforces a public API rather than allowing `create` to override values as
+it pleases, and is more inline with the behavior of components in Glimmer today -
+args that are passed into the class are distinguished from properties that are
+defined on the class.
+
+## Mixins
+
+Mixins are a contentious part of both the Ember Object model and the wider
+Javascript community - some swear by the pattern, and others believe it fundamentally
+flawed. While Ember mixins are at the core of Ember Object, the fact is that
+no standard solution for them has arisen in the wider Javascript community as
+of yet.
+
+Additionally, while concepts like computed properties, actions, and
+service injection are either unique to Ember or highly dependent on implementation,
+mixins can be implemented in a generic way which could be used across all of
+Javascript, independent of one's framework or library of choice. With that in
+mind, this RFC considers mixins out of scope and suggests that in the future Ember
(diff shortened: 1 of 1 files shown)


rwjblue:
Thanks for putting this together! I'm very much in favor...

chriskrycho:
First things first: I just want to say *this looks great*. It's a well-written and well-thought-out RFC, and I'm 100% in favor overall!

I do want to register a *strong* vote in favor of each of the alternatives raised as possibilities in the [Alternatives](https://github.com/pzuraq/rfcs-1/blob/80fec3df44276636cdfaa880a102c00f5550b1c2/text/0000-es-classes.md#alternatives) section, with some reasoning supplied for why in each case. But I'll start by giving some context.

## Qualify All The Things

While I *profoundly* appreciate Ember's strong use of conventions to accomplish things, I find both as a user and *especially* as someone trying to help others get up to speed on our app, I have a strong preference that things become *more explicit* when we're deploying new variants of existing APIs. The main complaint I hear from others on my team who come to our Ember app is that it's hard to understand what's going on, because there's so much that just magically happens. And that makes the learning curve harder. That goes especially for things like the way `classNames` and other things like it are automatically merged together. That magic is powerful when you know how it works, but it's confusing until then, and it can have surprising side effects because it's effective invisible when reading the code.

That sentiment goes double because I'm now the guy leading the charge with ember-cli-typescript: any *magic* we put in place makes it that much harder to fully support Ember from TypeScript. Now, a qualification to that sentiment is important: Even if we want TypeScript to be a first-class citizen of the ecosystem, right behind JavaScript, we should be careful not to let "easier for TS" *dictate* design decisions—even if we do want that to be a consideration. My point is simply that if we have two choices which are pretty comparable in terms of the JavaScript consumers' point of view but one is easier for TS, we should probably favor that.

So take all that as background for my position on each of these things.

## Why the "Alternatives" are better

TL;DR: less magic in every case, and therefore fewer surprises for anyone reading the code, *especially* people coming from the broader ecosystem. In other words: all the reasons we're doing this in the first place militate *strongly* in favor of doing them in the way that provokes the least surprise.

### Class Property Initialization

My (and I suspect most peoples') mental model of the behavior of `Ember.Object.create()` is that it's essentially a static factory method: it instantiates the object with whatever customizations you supply to it. That *is* the current outcome on the non-class variant, and I can't actually think of a good way to explain the proposed new behavior to another developer. As I understand the proposal, it *only* has any effect on properties which have not been declared on the class definition. So it's a way of adding properties to an instance of the class, but outside the constructor. The result is both surprising to existing Ember developers *and* (I suspect) provokes a big "Why would I ever do that?" for people coming new to the ecosystem. (The explanation of how it currently behaves was a good explanation, but it's still rather surprising!)

Making `Ember.Object.create()` actually behave as a factory which customizes the class in question is less surprising by far in my view.

### Merging and concatenating properties

As noted in my "qualifications" section above, I think these are already a rather surprising (albeit powerful) corner of the `Ember.Object` model, and I think a big part of what makes them surprising is that the merging is entirely implicit – you'd better have read the docs closely and carefully. If you're trying to figure out where a given class on a rendered component comes from, nothing in a component definition even gives a hint that there might be other sources of truth for that particular item. To the possible rejoinder that it's implied by the fact that it'll only happen if you're explicitly mixing in or inheriting from something else: sort of, but only if you know the behavior is possible in the first place, *and* it's distinct from the way all other properties on an item definition behave. In the proposal as written, `classNames` and `someOtherProperty` would behave differently from each other in `BarComponent`—but that's not apparent at all.

```javascript
const FooComponent = Ember.Component.extend({
  classNames: ['look-ma-a-foo'],
  someOtherProperty: ['also an array', 'of strings'],
})

class BarComponent extends FooComponent {
  classNames = ['hey-its-a-bar'];
  someOtherProperty = ['these are', 'also strings'];
}
```

Gladly, the docs cover this well:

> Standard CSS class names to apply to the view's outer element. This property automatically inherits any class names defined by the view's superclasses as well.

Still, I think it's better to be explicit:

```javascript
class BarComponent extends FooComponent {
  @concatenated classNameBindings = ['hey-its-a-bar'];
  someOtherProperty = ['these are', 'also strings'];
}
```

*Explicit is better than implicit* isn't a hard and fast rule, but it's a really good guideline for API design—and I think it's especially important when the behavior of one item is different from how everything else that looks like it behaves.

### Observers and listeners

I take much the same view here as I do about the merged and concatenated properties. To those considerations, I would add mainly that I increasingly think the framework should discourage the use of both in favor of computed properties and normal actions for the majority of cases. And the docs seem to agree:

> _**Note:** Observers are often over-used by new Ember developers. Observers are used heavily within the Ember framework itself, but for most problems Ember app developers face, computed properties are the appropriate solution._

Given that preference, making these more explicit not only has all the wins expressed above, but it's also a slight bump in the work required to use them, and—most importantly—it makes it *very* explicit that the property is bound in the ways it is.

## Summary

Given that the alternative proposals for merged/concatenated properties *and* observers and listeners are (a) backwards-compatible, (b) opt-in, (c) more in line with the way the rest of the JavaScript ecosystem behaves, and (d) easier to make sure we get right in TypeScript… well, those seem like pretty clear-cut cases to me. I think they're *much* better choices. And I think the alternative for `Ember.Object.create()` is far less surprising and easier to understand (and therefore easier to teach, as well).

rwjblue:
@chriskrycho - Thanks for the detailed and thoughtful response!

Believe it or not, I think that I actually agree with all very nearly all of your points, but I think the position _this_ RFC is taking is still the correct course of action. Very specifically, this RFC is _only_ focused on ensuring that `class Foo extends Ember.Object {}` functions properly (and yes this means that it has to support any/all of the existing Ember Object Model that the class you are extending from is utilizing) and that it is considered public API from a SemVer perspective. The point I'm making here is pretty subtle, but I think it is very important.

I believe the correct course of action for those still overly magical things (merged and concatenated properties, some bits of evented, observers, etc) is to continue iterating in future RFC's to work through deprecations and proper replacements independently (for each feature). If we try to replace all of Ember's Object model in a single RFC we would never be able to proceed.

rwjblue:
@chriskrycho - I don't quite grok what you mean in the "Class Property Initialization" section. Can you explain in more detail (possibly with code snippets)?

chriskrycho:
@rwjblue –

> Very specifically, this RFC is only focused on ensuring that `class Foo extends Ember.Object {}` functions properly (and yes this means that it has to support any/all of the existing Ember Object Model that the class you are extending from is utilizing) and that it is considered public API from a SemVer perspective. The point I'm making here is pretty subtle, but I think it is very important.

Totally agreed with that goal. My take was simply: if the alternatives work to support that option… let's do those; it's less to tackle in the future. That said, if the alternatives *don't* work to support that option, then I 100% agree: defer them, and if/when this RFC is accepted and implemented, write those as follow-ups.

***

> I don't quite grok what you mean in the "Class Property Initialization" section.

No problem! This actually made me go back and reread that section again (and yes, that comes out to three readings!) and I actually changed my mind on reading it again. I think that section could perhaps be clarified just a bit: the idea comes out to *don't use `.create()` with classes; always just use the constructor instead` – and I 100% agree with that. So skip that part. :wink:

pzuraq:
So to be clear, `create` is still considered the public API for creating classes, at least for those which are used in Ember. Another unmentioned part of the RFC (which should be added) is that finalization of the class must be moved out of the constructor to the `create` method - things like chains being finished. This is so that class properties can be set _before_ chains are finalized, so they won't trigger observations and whatnot (although if `set/get` aren't used will that matter?)

The current proposal is that options passed to `create` would also be passed into the constructor function, giving users an opportunity to modify them in the constructor before they are set on the class. Class properties would still "win" in this situation if there was a conflict between properties passed in and those defined on the class.

The alternative that @chriskrycho is mentioning is to instead update `create` such that it essentially does this:

```
static create(args) {
  const klass = new Class();

  for (arg in args) {
    klass[arg] = args[arg];
  }

  finishChains(klass);

  return klass;
}
```

So whatever would be passed to `create` would win, which is analogous to how it works today.

rtablada:
From a documentation and instruction side of things, while I agree that the main path should be to continue using `.extends` (at least for the time being). I think that there should be some documentation around how to use ES2015 class syntax in the official docs when true support lands based on this RFC.

I think including it in the official API as listed here and not documenting it would be a cause for confusion.

So, my thoughts would be to add documentation for `Using ES2015 Classes with Ember Objects` to the `The Object Model` section documenting the differences between `.extends` and `class Foo extends` with Ember Objects.

To clarify, this wouldn't be the pattern shown for the rest of the guides but the option would be documented in the Object Model section.

simonihmig:
1. first of all thanks as well to @pzuraq for this great RFC! Absolutely support embracing ES6 classes!
2. very much agree with @chriskrycho that reducing certain parts of Ember magic makes sense. Especially those that come quite unexpected. Probably nobody complains about the fact that putting a component module in a certain place in the folder structure will "magically" make this available to the DI container and thus any template, without having to write hundreds of lines of DI config. Because this feels quite natural and in line with expectations. But certain bits that were mentioned here just come so unexpected, that it is hard to reason about them without actually knowing the magic. But yes, all of this certainly does not need to be solved in this RFC or any other singer RFC...
3. so I think when we embrace new JS features as first class citizens in Ember, we should make sure that we do not introduce new unexpected particularities in the way they are used in the Ember ecosystem. Newer frameworks with a fresh start, that don't carry the "legacy" Ember does, are amongst other reason popular because they use just "vanilla JS". So how long our journey from "legacy Ember" to "vanilla Ember" might even take, we should make sure that we end up in a world that is as little ember-ish and as much vanilla-ish as possible (*sorry for any pain caused by that sentence!* 😬)
4. that being said, I would support the alternative in regards to class properties. I would expect, that class properties used as defaults would "just work", and that the alternative (they win over args of `.create()`) would feel like a bug to non-Ember folks. Also being kind of forced to define defaults in the constructor as suggested (if they are meant to work consistently) would be pretty much on the ember-ish side vs. vanilla-ish
5. while other parts might be fleshed out in separate RFCs, the semantics of `.create()` regarding ES6 classes need to be finalized in this RFC, as changing that later would introduce a breaking change, right?
6. I don't know if that is being meant to be covered by this RFC, but in either case maybe this could be clarified: the given examples always assume that ES6 classes are defined at the very end of the inheritance chain. But are they supposed to also work in the middle of the prototype chain? My understanding is that you cannot `.extend()` from an ES6 class currently, right? Should this work when they become public API as part of this RFC?
7. If not, they would not feel as real first class citizens. For example I could not write components in an addon as ES6 classes, because I cannot assume that they are not `.extend`ed by the addon's consumer. And on the other hand the consumer should not need to know how they were implemented to extend them. If that should work, then great! But even more important then that `.create()` works in a consistent way (cannot assume they are instantiated using `new` vs. `.create()`)

rwjblue:
@rtablada - While I am not opposed to a small paragraph or note suggesting that Ember's object model _works_ with standard ES such as `class Foo extends Ember.Object {}`, but I would be opposed to a full section as I believe that having the guides be disjointed with multiple syntaxes will make them a bit harder to reason about. Also, there are still a number of common pitfalls (to be addressed by future RFCs) that need to be ironed out before updating the entirety of the official guides and API documentation to use classes.

rtablada:
@rwjblue I'm not suggesting to update the whole guides as I had mentioned. Instead I was recommending a section under the `Ember Object Model` heading that shows that works, explains the pitfalls, and likely links to `ember-decorators` for better support.

rtablada:
One thing I'm not sure is quite covered here is the use of the `new` keyword compared to the `.creates` static method.

I think that by introducing the idea of ES2015 classes it also makes users want to use the `new` keyword along with these classes.
People that are used to ES2015 classes will be more likely to want to use `new MyClass` to instantiate their objects and IMO there should be more explanation of the subtle differences between this and `MyClass.create`.
Is there any downside to this that needs to be documented or should it "Just Work ™️"?

rtablada:
@rwjblue I was thinking after "Enumerables" so that is quick to reference.
But, as you pointed out: we don't want to change the train of thought in the middle of a section on the basics of the Object model.

screen shot 2017-07-29 at 4 05 54 pm

rwjblue:
@simonihmig 

> 3. so I think when we embrace new JS features as first class citizens in Ember, we should make sure that we do not introduce new unexpected particularities in the way they are used in the Ember ecosystem. Newer frameworks with a fresh start, that don't carry the "legacy" Ember does, are amongst other reason popular because they use just "vanilla JS". So how long our journey from "legacy Ember" to "vanilla Ember" might even take, we should make sure that we end up in a world that is as little ember-ish and as much vanilla-ish as possible (*sorry for any pain caused by that sentence!* 😬)

Totally agreed. One thing to note here, however, is that our "legacy" is one of our strongest selling points: stability with an upgrade path to the future. If for example, we decided to drop support for object model features that are currently heavily depended on, many of our users would be completely stuck and unable to upgrade. It is **very** important that we do not bifurcate the community during these sorts of changes, which is why this is being handled in small digestible parts.

> 4. that being said, I would support the alternative in regards to class properties. I would expect, that class properties used as defaults would "just work", and that the alternative (they win over args of `.create()`) would feel like a bug to non-Ember folks. Also being kind of forced to define defaults in the constructor as suggested (if they are meant to work consistently) would be pretty much on the ember-ish side vs. vanilla-ish

I totally disagree with you here.  We are actually embracing what is done in vanilla JS for class properties. Consider the following:

```js
class Foo {
  derp = "hi!";
  constructor(value) {
    console.log(this.derp); // => undefined
    this.derp = value;
  }
}

let foo = new Foo('bye!');
console.log(foo.derp); // => hi!
```

As you can see [in the proposal here](https://tc39.github.io/proposal-class-fields/#[[construct]]) the constructor is ran _before_ the class fields are applied to the instance.

If you want to accept an argument but fallback to a default value, you simply have no choice but to use the `constructor`.

> 5. while other parts might be fleshed out in separate RFCs, the semantics of `.create()` regarding ES6 classes need to be finalized in this RFC, as changing that later would introduce a breaking change, right?

Agreed.

> 6. I don't know if that is being meant to be covered by this RFC, but in either case maybe this could be clarified: the given examples always assume that ES6 classes are defined at the very end of the inheritance chain. But are they supposed to also work in the middle of the prototype chain? My understanding is that you cannot `.extend()` from an ES6 class currently, right? Should this work when they become public API as part of this RFC?

Great point!  I believe that it would work properly with the changes listed in this RFC, but we should definitely state that in the RFC specifically.

rwjblue:
@rtablada (re https://github.com/emberjs/rfcs/pull/240#issuecomment-318859486) - 

At this time we are not suggesting a change in Ember's DI system (e.g. Ember will continue to call the static `create` method to build instances).

 However, with the changes listed in this RFC you would be able to use the following:

```js
class Foo extends Ember.Object { }
let f = new Foo({ some: 'prop', goes: 'here' });
```

The main caveat (that we should continue to explore) is how we indicate to computed properties and observers that the user-land constructor is completed, and that we should begin listening to change events. This is currently done in the `constructor` just after `init` is called, but this RFC proposes to move that stage into the static `Ember.Object.create` method. Without that change, calling `super(...arguments)` in your custom classes `constructor` would enable change events and any initialization you do in the `constructor` would cause CP's and observers to fire (which is not ideal).

Moving the enabling of change events to the static `create` method is fine for existing `Ember.Object` users and for any objects instantiated by the DI system. For objects not created by the DI system (either via `owner.lookup` or `owner.factoryFor(...).create()`), we would need to provide a method that would need to be called to enable change events.

Seems good to update the RFC to at least mention this.

simonihmig:
@rwjblue 

> One thing to note here, however, is that our "legacy" is one of our strongest selling points: stability with an upgrade path to the future.

Absolutely, no question about that! There is a reason I joined the train somewhere @1.0.0-rc.x and never thought about leaving! 😉

> As you can see in the proposal here the constructor is ran before the class fields are applied to the instance.

Oh, was not aware of that, thanks for pointing that out!

However I still think that does not rule out implementing `.create()` in the alternative way, does it? I guess it depends on the way users would normally expect it to work with classes, like...

```js
return new Class(props);
```

or like 

```js
const instance = new Class();
Ember.setProperties(instance, props);
return instance;
```

cibernox:
I'll also drop my view.

I very much agree with @chriskrycho on the fact that I'd like to make concatenated properties, mixins, observers and listeners and leave as much of that as possible to some "official" decorator library.

The reasons are the same: Less magic makes Ember easier to learn and more aligned with other frameworks. I don't fully buy the idea that `class MyClass extends Ember.Object` has to support all the features that `MyClass = Ember.Object.extend()` does.

The same way angle-bracket were added and then removed and one of the reasons for that was because we wanted to take this unique oportunity of adding new syntax to change the semantics of components in a backwards incompatible way (for the better) and then Glimmer was born, I think we could take the unique opportunity of switching from `MyClass.create()` to `new MyClass` (I'd enforce/prefer `new` over the static `.create`) to cut with the past and "extract functionality from core".

We could say "If you use `class MyClass extends Ember.Object` and `new MyClass` you are opting-in to a smaller, lighter object model that requires you to be explicit and use decorators to get some behaviour that is built in when using `Ember.Object.extend`.

I don't know if this might present some problem for mix and matching. Options of mix and match are two:

- `class MyClass extend ClassCreatedUsingObjectExtend`
- `MyClass = ClassCreatedUsingES6.extend()`

If we make the ES6 classes have less features, can then a user decide to use again `ES6DefinedClass.extend` and have the features that ES6 classes lack? If that is not possible, can users be warned that once an ancestor is defined with ES6 syntax, certain features are not available anymore or force them to continue using ES6 classes for that hierarchy?

I just want to highlight that this change in syntax offers the same opportunity to rethink our object model than angle-bracket components offered to introduce components with new semantics. I appreciate backwards compatibility as much as anyone else, but it's worth considering using this to break with the past if certain edge cases if that is for the best long term.

NLincoln:
My only worry is that this is currently possible:
screenshot_20170730_181520

rtablada:
@cibernox with the compatibility scheme for Ember I think it would have to be the Object model for Ember 4.0 to keep up with compatibility concerns.

rtablada:
@NLincoln that's a good point that using `new` with `Ember.Object` is currently possible, but the differences aren't well documented and there are some edgecases that `create` resolves.

dwickern:
I can't 👍 this enough. This change will improve IDE tooling and make Ember+typescript a much better fit.

I'd like to add that method overrides should use ES6 `super` instead of `this._super`

```js
  didInsertElement() {
    this._super(...arguments);
  }
});
```

```js
export default class extends Ember.Component {
  didInsertElement() {
    super.didInsertElement();
  }
}
```

rwjblue:
We discussed this in the Ember core team call today. We are generally all in favor of this moving forward and decided to move this to final comment period.

workmanw:
First off, I'm very 👍 on this.

One topic not discussed here is `reopen` / `reopenClass`. 

Finally, I know the spirit in which it was meant, but I think we should stop short of calling Mixins, Observers, Listeners and Concatenated properties "obscure features" and "overly magical". It's one thing to call them advanced features, it's another thing to implicitly suggest they're bad and shouldn't be used. They've very widely used.

**Observers:** 1683 usages across 499 addons: [Ember Observer](https://emberobserver.com/code-search?codeQuery=observe%5Brs%5D%5C%28&regex=true)
**Mixins:** 1788 usages across 533 addons: [Ember Observer](https://emberobserver.com/code-search?codeQuery=Mixin.create)
**Listeners** 6606 usages across 1237 addons: [Ember Observer](https://emberobserver.com/code-search?codeQuery=%5BEmber%5C.%7C%7C%20%5Don%5C%28&regex=true)

pzuraq:
Mixins are one thing. They have plenty of issues, for instance with ensuring call order for functions that are overriden (what does `_super` of a mixin call? What happens in a class that has several functions that all implement `didInsertElement` and all call `_super`), but the overall issue for this RFC is that they won't be an Ember specific concept any more, but an ES class concept in general.

I agree that Listeners/Observers are not a _bad_ pattern. When used correctly, they are incredibly powerful, and concepts like Promises do not always fully close over them. The issue here is that they come out of the box, standard on every object. This adds extra weight to every single Ember object, _and_ encourages improper usage of a powerful, low level pattern.

I disagree with concatenated/merged properties. They are highly magical, and like observers/events add weight to every object. However, unlike events/observers, they are very easily solved with a simple decorator, as in the example in the RFC. 

Overall, the argument isn't to take any of these features away. It's to split them out into explicit opt-ins, ones that could likely be shared with the larger Javascript community.

rwjblue:
@workmanw:

> One topic not discussed here is reopen / reopenClass.

Agreed, this was not mentioned. There are two different aspects that are important here:

> Does the presence of `.reopen` and `.reopenClass` fundamentally change the interop system with native ES classes?

To support this we will need to move more of the current object model to "native" ES likely meaning a move away from `.ClassMixin` and `.PrototypeMixin` towards mutating the prototype directly.

> Is it likely that `.reopen` / `.reopenClass` are concepts that will carry over in a "normal" ES class based world?

Probably not.  While I think it is important for interoperability that we make sure these methods work properly, I do not think `.reopen` and `.reopenClass` are concepts that most users should have to learn and understand.

---

> Finally, I know the spirit in which it was meant, but I think we should stop short of calling Mixins, Observers, Listeners and Concatenated properties "obscure features" and "overly magical".

It seems clear to me that we **must** continue re-aligning the object model away from the custom bespoke system that we were forced to create when the underlying language lacked so many fundamentally required features towards leverage the common solutions that have been added to the language itself.  It is quite likely that this migration will eventually mean some of our current object models features will need to be deprecated.  We are absolutely not suggesting in this RFC that any new deprecations be added. Any new object model deprecations will almost certainly go through a deprecation RFC process.

The specific terms "obscure" and "magical" are a bit subjective, if those terms seemed offensive to folks, I am definitely sorry.  In my [comment](https://github.com/emberjs/rfcs/pull/240#issuecomment-318845595) I was using "magical" to refer to features whose underlying implementation so complicated that even a core team member (me 😝 ) cannot understand them without a multi-hour deep dive...

workmanw:
@rwjblue 

>  if those terms seemed offensive to folks, I am definitely sorry

No, it definitely wasn't offensive. But they are very powerful and very useful features. In particular Mixins and Observers. These are features found in many languages and frameworks. I think they've gotten a pretty bad rap in Ember because they incorrectly used in lieu of proper framework solutions (e.g. observing values before we had promises). _"When all you have is a hammer, everything looks like a nail."_

rwjblue:
@workmanw:

> But they are very powerful and very useful features. In particular Mixins and Observers. 

Totally agreed on both accounts!

RE: Mixins, this RFC is specifically suggesting that we should embrace native ES features (ES classes today do not support "mixins" like Ember has). `Ember.Mixin.create` will definitely still be a supported feature for the foreseeable future (in `Ember.Object.extend()` mode), but we should likely also consider pushing for a solution in spec land. See https://github.com/ember-decorators/ember-decorators/issues/115#issuecomment-313191923 for some additional discussion around this topic.

luxferresum:
If we get this, wouldnt this mean we can use mixing like this:

```
class Foo extends Ember.Object.extend(myMixin){}
```

dwickern:
@luxferresum Yes, that's equivalent to
```js
const BaseClass = Ember.Object.extend(myMixin);
class Foo extends BaseClass {}
```

rwjblue:
We discussed this at the Ember core team meeting today.  Lets do it. 🎉 👏

bartocc:
is this normal the merged PR file is still named "0000-es-classes.md" ?

This name positions de file at the top of the "text/" dir while it is the last PR that has been merged.

Forgive me if this is a dumb question since I am not very familiar with the RFC's merge process ;-)

rwjblue:
You are right, I was supposed to update that before merging 😱.  Updated in https://github.com/emberjs/rfcs/commit/0af0bbece6dd90cce558a1a7b3a60b3bca74c8e1.

championswimmer:
@rwjblue @chriskrycho 

Ok not sure where to ask this, think this is probably appropriate. 

Using `ember-source` and `ember-cli` 3.0


If have this Mixin - 

```javascript
// promise-resolver.js
export default Mixin.create({})
```

And I do this in my app 

```javascript
// in my app
import PromiseResolverMixin from 'ember-promise-utils/mixins/promise-resolver';

    let PromiseResolverObject = Ember.Object.extend(PromiseResolverMixin);
    let subject = PromiseResolverObject.create();

```

It works perfectly.

But if we want to use Typescript or ES6 classes

```typescript
// promise-resolver.ts
export default class PromiseResolverMixin extends Mixin<PromiseResolverMixin> {}
```
or 
```javascript
// promise-resolver.js
export default class PromiseResolverMixin extends Mixin {}
```
and then use the same code as above 
```javascript
// in my app
import PromiseResolverMixin from 'ember-promise-utils/mixins/promise-resolver';

    let PromiseResolverObject = Ember.Object.extend(PromiseResolverMixin);
    let subject = PromiseResolverObject.create();

```

We get the error
```
Error: Assertion Failed: Expected hash or Mixin instance, got [object Function]
                at new EmberError (http://localhost:7357/assets/vendor.js:24127:25)
                at Object.assert (http://localhost:7357/assets/vendor.js:24370:15)
                at Mixin.reopen (http://localhost:7357/assets/vendor.js:38773:161)
```

I get it, that the **class** itself is not a Mixin instance. 
Well, and good. Let's change it to - 

```javascript
// in my app
import PromiseResolverMixin from 'ember-promise-utils/mixins/promise-resolver';
let subject = new PromsiseResolverMixin()
```
doesn't work

also, this

```typescript
// in my app (in .ts file)
import PromiseResolverMixin from 'ember-promise-utils/mixins/promise-resolver';

class PromiseResolver extends Ember.Object<PromsiseResolverMixin> {};
let subject = new PromiseResolver();
```

doesn't work either.

I am kinda stumped here, how best to use  ES6 class definition for Mixins ?

chriskrycho:
In short, you can’t—and I wouldn’t try! The semantics just don’t map in well at this point.

I also *strongly* encourage people to avoid mixins at this point. Mostly they just cause pain.

To use a mixin defined with `Mixin.create()` in a class, you can do this:

```ts
import SomeMixin from 'the/path/to/it';

export default class MyComponent extends Component.extend(SomeMixin) {
  // normal class definition
}
```

Note that you *may* still see type errors when you do that in TypeScript – happy to talk about those in an issue on https://github.com/typed-ember/ember-cli-typescript or in the #topic-typescript room in the Ember Slack.

GCorbel:
@championswimmer I personally prefer to have Utils classes. For example, if you want to create something to archive a model, instead of create a mixin and do `myModel.archive()`, I prefer to create a class and do `Archiver.call(myModel)`. I know this is doesn't solves the problem you have but it's a good alternative, it's clean and it follows the "Single Responsibility Principle".

chriskrycho:
Also worth note: you often don't need _classes_ at all to solve this. In our experience, most stuff that we use mixins for can be replaced by plain old modules and pure in/out functions.

jacobq:
Where can one find more documentation about this? For example, when defining a model like `ember g model foo`, do I need to use [`Model.extend({...})` like the API docs show as an example](https://www.emberjs.com/api/ember-data/3.3/classes/DS/methods/attr?anchor=attr)?

```
import DS from 'ember-data';
const { Model } = DS;

export default class FooModel extends Model.extend({
  // When do things need to go here?
  prop1: DS.attr(),
}) {
  // Will this work?
  prop2 = DS.attr();
}

```

pzuraq:
@jacobq you can find the most up to date documentation in the [ember-decorators project's documentation](https://ember-decorators.github.io/ember-decorators/latest/). There is no official Ember documentation about using native classes yet because native classes have not officially shipped, and there are still potential changes that could change the way native classes are used. See RFC #337 for more details.

PR #244, opened 2017-08-16
Update 0000-find-polymorphism.md

BryanCrotaz:
Add proposal for automatic hierarchy discovery

### DIFF
--- a/text/0000-find-polymorphism.md
+++ b/text/0000-find-polymorphism.md
@@ -70,7 +70,7 @@ and get all the Lines and Circles.
 
 # Detailed design
 
-##FindAll and PeekAll
+## FindAll and PeekAll
 `store.findAll('shape', { polymorphic: true })` would return a live array that would have all the shapes and all of the
 shape subclasses.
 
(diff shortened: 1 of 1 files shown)


xaqq:
Hello,

I never understood why this wasn't either the default behavior or available through some other means.
If anything, I don't agree with your propose to have to add a `{ polymorphic: true }`, I strongly believe this should be the default behavior.

Hopefully if this RFC is implemented we may finally have a non broken store in EmberJS !

text/0000-find-polymorphism.md
```diff
@@ -151,5 +151,98 @@ different part of the app.
   3. This approach is incosistent with our current relationship polymorphism. It would require us to either deprecate
 `hasMany({ polymorphic: true })` (and we just released 1.0 :( ) or to use an inconsistent mixed approach.
 
+
+## Auto discovery of heirarchy
```
thomascchen: I believe `heirarchy` should be spelled `hierarchy.`

text/0000-find-polymorphism.md
```diff
@@ -151,5 +151,98 @@ different part of the app.
...
+
+The above would be much improved with auto discovery of heirarchy. This would remove the need for `abstractBaseClass` and the `polymorphic` flag on `find` etc could be inferred from whether a class has any descendants.
```
thomascchen: `heirachy` -> `hierarchy` here too 🙂

sandstrom:
Polymorphism for queries would be very useful!

Random thoughts:

- How about an adapter-level property `assumePolymorphism: true/false`. That way, some people can go with the more verbose `find('my-query', { polymorphic: true })` and others can have that as the default behaviour.
- With Ember Engines, can one assume that all models are loaded? If all models aren't loaded, then one cannot reliably determine the model-hierarchy, so auto-discovery wouldn't work. Not sure if this would be a problem in practice or not, but could be worth keeping in mind.

wagenet:
@runspired is this something you want to pursue?

runspired:
I don't think we need this anymore. Requests and the cache are now agnostic to type, and the identifiers RFC allows users to configure multiple types being considered as part of one identity.

PR #246, opened 2017-08-25
Array#findIndex and Array#findIndexBy

buschtoens:
[Rendered](https://github.com/buschtoens/ember-rfcs/blob/array-find-index/text/0000-array-find-index.md)

### DIFF
--- a/text/0000-array-find-index.md
+++ b/text/0000-array-find-index.md
@@ -1,0 +1,299 @@
+- Start Date: 2017-08-25
+- RFC PR: (leave this empty)
+- Ember Issue: (leave this empty)
+
+# Summary
+
+This RFC aims to add the [new ES6 Array method `findIndex`][mdn-findIndex]
+and a complimentary `findIndexBy` method to the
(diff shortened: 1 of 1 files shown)


kellyselden:
My understanding of the Ember Array API (and Object) are basically frozen as we move away from them and/or extract them?

mmun:
Kelly's right. The Ember.Array API is frozen. In the future we may move the Array/ArrayProxy classes to an addon where they can be iterated on.

If you're working with a native array and you're targeting modern browsers (i.e. not IE11) then you can just use the native Array.prototype.findIndex. Otherwise, you'll need to use something like ember-lodash.

If you need to support more general cases using `objectAt` then must first map the object to a native array using `toArray()`:

```js
post.get('comments').toArray().findIndex(comment => {
  return comment.get('title') == "Hello world"
});
```

aalimovs:
@mmun FYI, with `toArray()` the order is not guaranteed, see [here](https://emberjs.com/api/ember/3.4/classes/ArrayProxy/methods/toArray?anchor=toArray).

PR #252, opened 2017-09-25
Browser Support Changes

rwjblue:
[rendered](https://github.com/emberjs/rfcs/blob/browser-support-3.0/text/0252-browser-support-changes.md)

tldr; This RFC proposes dropping support for IE9, IE10, and PhantomJS.

### DIFF
--- a/text/0252-browser-support-changes.md
+++ b/text/0252-browser-support-changes.md
@@ -1,0 +1,100 @@
+# Browser Support Changes
+
+- Start Date: 2017-10-25
+- RFC PR: (leave this empty)
+- Ember Issue: (leave this empty)
+
+# Summary
+
+Solicit feedback on dropping support for IE9, IE10, and PhantomJS.
+
+# Motivation
+
+As Ember heads towards version 3.0, it is a good time to evaluate our browser support matrix. Ember follows Semantic Versioning, and we consider browser compatibility to be under the umbrella of those guarantees. In other words, we will continue to support whatever browsers we officially support in Ember 3.0 until Ember 4.0.
+
+We want to make this decision on the basis of the browsers that our community still needs to support, while weighing that against the costs we bear as a community to support older browsers. This RFC will lay out some of those costs, so we can decide what tradeoff is most appropriate.
+Members of the core team maintain many different kinds of apps across many different kinds of companies. Some of us work on applications with small, agile teams, while others work inside of large corporations with many engineers. When this topic came up amongst the team, we discovered that, across all these different companies and Ember apps, we did not generally support these legacy browsers.
+
+Because of this, the core team's impression is that the costs support now far exceed the benefits, and we are considering dropping support for them in Ember 3.0. Before we make the decision, we want to hear from the rest of the community. Supporting IE9, IE10, and PhantomJS incurs significant cost, both in terms of features and maintenance, and we want the community to help us think through the cost-benefit analysis.
+
+Ember is more than just the framework's code. When people use Ember, they expect to be able to use Ember's tooling, read Ember's documentation, find solutions to problems on Stack Overflow, and read tutorials produced by community members. **All of these****, including addons that follow Ember’s lead,** **are shackled to the limitations of these legacy browsers****.** By dropping support for them, people can begin to rely on the improved baseline of features.
+
+Some of the features (unavailable in IE9, IE10, or PhantomJS) that addons will be able to freely take advantage of include:
+
+
+- requestAnimationFrame ([caniuse](http://caniuse.com/#feat=requestanimationframe), [MDN](https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame))
+- CSS flexbox ([caniuse](http://caniuse.com/#search=flexbox), [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Flexible_Box_Layout/Using_CSS_flexible_boxes))
+- Websockets ([caniuse](http://caniuse.com/#feat=websockets), [MDN](https://developer.mozilla.org/en-US/docs/Web/API/WebSockets_API))
+- let ([caniuse](http://caniuse.com/#feat=let), [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/let))
+- const ([caniuse](http://caniuse.com/#feat=const), [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/const))
+- TypedArray ([caniuse](http://caniuse.com/#feat=typedarrays), [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray))
+- Geolocation API ([caniuse](https://caniuse.com/#search=Geolocation), [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Geolocation))
+- Online/offline API ([caniuse](http://caniuse.com/#feat=online-status), [MDN](https://developer.mozilla.org/en-US/docs/Online_and_offline_events))
+- XHR advanced features ([caniuse](https://caniuse.com/#feat=xhr2), [specification](https://www.w3.org/TR/2012/WD-XMLHttpRequest-20120117/))
+- HTTP2 ([caniuse](http://caniuse.com/#feat=http2), [wikipedia](https://en.wikipedia.org/wiki/HTTP/2))
+- Web Workers ([caniuse](http://caniuse.com/#feat=webworkers), [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Using_web_workers))
+- IndexedDB ([caniuse](http://caniuse.com/#feat=indexeddb), [MDN](https://developer.mozilla.org/en-US/docs/Web/API/IndexedDB_API)) 
+- WebGL ([caniuse](http://caniuse.com/#feat=webgl), [MDN](https://developer.mozilla.org/en-US/docs/Web/API/WebGL_API))
+- File API ([caniuse](http://caniuse.com/#feat=fileapi), [MDN](https://developer.mozilla.org/en-US/docs/Web/API/File))
+- PageTransitionEvent ([caniuse](http://caniuse.com/#feat=page-transition-events), [MDN](https://developer.mozilla.org/en-US/docs/Web/API/PageTransitionEvent))
+- SVG filters ([caniuse](http://caniuse.com/#feat=svg-filters), [MDN](https://developer.mozilla.org/en-US/docs/Web/SVG/Tutorial/SVG_Filters_Tutorial))
+- MutationObserver ([caniuse](http://caniuse.com/#feat=mutationobserver), [MDN](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver))
+
+Below, we’ve outlined several specific features we’re interested in using to improve the Ember framework itself. We’ve also included some other supporting arguments for this decision.
+
+## Vendor Support
+
+Microsoft dropped most support and maintenance for IE9 and IE10 on 2016-01-16 (IE9 on Vista SP2 [expired in April 2017](http://www.allyncs.com/docs/lifecyclesupport.html)).
+
+With the advent of headless Chrome and Firefox, PhantomJS is now [effectively unmaintained](https://groups.google.com/forum/#!topic/phantomjs/9aI5d-LDuNE). The default testing boilerplate for Ember-CLI-generated applications was changed to headless Chrome in Ember-CLI 2.15.
+
+## WeakMap, Map, Set
+
+From a framework perspective, being able to rely on native `WeakMap` support will allow us to remove a significant number of fallback paths that are used in browsers without `WeakMap`. Using `WeakMap` results in better developer ergonomics as it allows us to remove many of the random properties that we currently have to assign to an object which makes interacting with your objects in the devtools much less noisy. Minimal support for WeakMap was [introduced in IE11](http://kangax.github.io/compat-table/es6/#test-WeakMap).
+
+## Better ES Class Support
+
+In order to support static class methods (with inheritance) transpilers (e.g. Babel) need to leverage the `Object.setPrototypeOf` / `Object.getPrototypeOf` API's. Without the ability to rely on `Object.setPrototypeOf` we will not be able to continue iterating slowly towards leveraging ES classes as a replacement for the custom object model functionality that we have known and loved for so many years. Specifically, there is no replacement / capability to support proper inheritance with `.reopenClass`. There are several lower-fidelity hacks you might opt into, but none that we think satisfy the needs of the Ember community.
+
+Generally this means IE11 is the oldest browser we can reliably transpile ES classes for reliably.
+
+## Typed Arrays
+
+Typed arrays are not currently used in Ember, but experimentation is underway deep in the internals of glimmer-vm to be able to further reduce template size *and* the costs associated with expanding the wire format (currently a JSON structure) into a runnable program. Leveraging typed arrays would allow Ember and Glimmer apps to completely avoid the wire format to opcode compilation that currently happens before initial render, it also significantly reduces the resulting memory footprint for the same runnable program.
+
+## DOM API Improvements
+
+Although IE9 introduced JavaScript engine with support for much of ES5, it was not until IE10 that the browser began to support much of what developers consider modern web platform APIs. Littered throughout the Ember and Glimmer-VM codebase are [many](https://github.com/glimmerjs/glimmer-vm/blob/1759c16defc546b034b97e37141187652ed93859/packages/%40glimmer/runtime/lib/dom/props.ts#L54) [examples](https://github.com/glimmerjs/glimmer-vm/blob/9ecc88504c81469ba20dba3ed3f37d373a998355/packages/%40glimmer/test-helpers/lib/helpers.ts#L170) [of](https://github.com/glimmerjs/glimmer-vm/blob/bfed16af6a5ecce4fbe9f27783245fe0f8b03480/build/broccoli/transpile-to-es5.js#L25) IE9 workarounds (and [PhantomJS workarounds](https://github.com/glimmerjs/glimmer-vm/blob/1759c16defc546b034b97e37141187652ed93859/packages/%40glimmer/runtime/lib/dom/props.ts#L49), in fact). We’ve worked hard to make these fixes free at runtime for modern browsers, but some cost is unavoidable.
+
+PhantomJS in particular is a weird environment. Users must often fix Phantom-specific browser bugs, which is wasted effort since real users never run your app in Phantom. And "how to debug in Phantom" is an entire extra skill people are forced to learn. Testing your app in PhantomJS is generally a form of “testing theater”, since it fails to execute your code in a realistic environment.
+
+## requestAnimationFrame
+
+IE10 introduced support for `requestAnimationFrame`, an efficient way to schedule work in the browser environment. We’re interested in using this API to explore incremental rendering strategies, and as a way to improve Ember’s coordination with the browser when native promises are used in application code.
+
+# Detailed Design
+
+When using Ember applications in IE9, IE10, or PhantomJS Ember will cause an appropriate deprecation to be issued. The deprecation will be “until 3.0” and will reference an entry in the deprecation guide. The guide entry will describe For example:
+
+
+> Using Ember.js in IE9, IE10, or PhantomJS is deprecated and will be unsupported in Ember.js 3.0. We recommend using Ember’s 2.x LTS releases if your applications must support those browsers.
+> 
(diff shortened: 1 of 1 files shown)


cibernox:
I know that having an accurate estimate of when Ember 3.0 will be out is a serious act of spiritism, but having an estimated timeframe might affect the answer of some people, like in "we would be able to drop it mid-2018 but not before". 

Not my case tho, I'd kill both yesterday.

XaserAcheron:
As someone developing an app for corporate clients, the most important takeaway from this is that IE11 is still firmly on the supported list. That's pretty darn important since Edge isn't available on Windows 7/8 and there are plenty of browser-shy corporate environments out there that haven't made the Win10 leap, and all that jazz.

tl;dr: 👍

rwjblue:
> an estimated timeframe might affect the answer of some people, like in "we would be able to drop it mid-2018 but not before". Perhaps a reverse deadline (p.e. "not before Q2 2018") would be enough.

Personally, if it were just up to me to unilaterally decide, I would have picked 2.16 as the last 2.x version. Obviously, that is not going to be the case and we will definitely have at least a couple versions to prepare, but to a large degree the communities feedback on this RFC will help us determine the exact versioning. 

Hopefully we will have a more detailed summary of the  2.x -> 3.0 process soon, as this RFC is focusing only on the specific browser support aspect.

raido:
No objections, I am in favor 100%.

martndemus:
So will this 3.0 be more about dropping browsers and maybe trimming a little fat? Nothing big?

rwjblue:
Our policy about major versions has not changed. There is no tie to new features, and nothing will be removed/dropped that hasn't been deprecated.

Side note for all those that went through the 1.13 -> 2.0 migration: there will absolutely be no "rush to add new deprecations so that we can remove XYZ" regardless of the exact timeline we choose.

joukevandermaas:
I work on an app for a pretty big/slow moving corporate environment. We only support IE11 and evergreen browsers. We also stopped testing in PhantomJS a long time ago. This seems like a great (if 
mostly inconsequential) change from that perspective.

More generally, in my opinion there's no point in supporting a browser that the vendor has dropped support for. Perhaps it's out of scope for this RFC, but I think it makes sense to adopt a policy where you stop supporting browsers e.g. one year after the vendor stops supporting them (for example, that would imply April 2018 for IE9). This is somewhat similar to what you do for Node in Ember CLI.

shearichard:
> As someone developing an app for corporate clients, the most important takeaway from this is that IE11 is still firmly on the supported list. That's pretty darn important since Edge isn't available on Windows 7/8 and there are plenty of browser-shy corporate environments out there that haven't made the Win10 leap, and all that jazz.

Doubly so in Government - I don't like it but it's true.

jrjohnson:
My primary app is in use at medical Universities around the world. We just went through a long comment period to make the change to drop all IE support (edge and other evergreen browsers only). There were *absolutely no objections* to this change. I'm not advocating dropping IE 11 support, but we can do without 9&10 for sure.

@rwjblue I don't want to pile onto a good RFC, but I wonder if this is the right place to encourage testing builds which target only evergreen browsers separated from those which include IE 11. When we dropped support we found there were issues that needed to be handled because most things expect to be transpiled into IE9 compatible JS. When you raise that minimum threshold some issues surface and I think now that `targets.js` is so easy to find and change that some of these problems will have a big impact.

Fed03:
I approve. what about dropping IE altogether and stick with EDGE? too drastic?

XaserAcheron:
Too drastic. See my comment above.

mixonic:
@Fed03 a few reasons I think Ember should continue to support IE11:

* IE11 is still supported by MS, it is the last and only version of IE they offer security updates for.
* IE11 is popular in enterprise settings due to its inclusion of backwards compat modes for websites that target IE8.
* Over 25% of developers that participated in the [Ember Community Survey](https://www.emberjs.com/ember-community-survey-2017/#browser-requirements) expected to support IE11 this year.
* IE11 still has [more marketshare than Edge](https://www.w3counter.com/trends).

rich-poole:
No objections here, would personally like to see the end of IE11 but that's not realistic in the near future.

localpcguy:
Fully in favor of this move. Some stats on usage, from both StatCounter and NetMarketShare. I included the later despite reservations because their IE number tend to be higher due to how they do their counting.

IE9: 0.36% (SC), 0.96% (NMS)
IE10: 0.38% (SC), 0.72% (NMS)

Links:
- http://gs.statcounter.com/browser-version-partially-combined-market-share/desktop/worldwide/#monthly-201703-201708
- https://www.netmarketshare.com/report.aspx?qprid=3&qpaf=&qpcustom=Microsoft+Internet+Explorer+9.0&qpcustomb=0
- https://www.netmarketshare.com/report.aspx?qprid=3&qpaf=&qpcustom=Microsoft+Internet+Explorer+10.0&qpcustomb=0

Cryrivers:
@mixonic is it possible to have Ember 3.0 drop support for IE by default, and make it support IE 9/10 and 11 through different polyfill addons?

localpcguy:
I think usage of IE11 is still way too high to consider dropping support by default @Cryrivers, again from StatCounter and NetMarketShare. I think not having in supported by default would lead to it not being included in many cases, causing somewhere around 10% of users being unable to use the developed site. Unfortunately, that isn't a number I see decreasing any time soon, until corporations are convinced it's time to move to Edge, and Edge supports their internal applications like IE11 does, IE11 will continue to hold about the same market share. It's trend line has been pretty flat for the last year. And Microsoft has committed to supporting it until October 14, 2025 😢 


7.1% (SC), 12.7% (NMS)

- http://gs.statcounter.com/browser-version-partially-combined-market-share/desktop/worldwide/#monthly-201703-201708
- https://www.netmarketshare.com/report.aspx?qprid=3&qpaf=&qpcustom=Microsoft+Internet+Explorer+11.0&qpcustomb=0

Cryrivers:
@localpcguy i understood that. i'm just thinking of supporting legacy browsers through addons.

rwjblue:
From @Cryrivers in https://github.com/emberjs/rfcs/pull/252#issuecomment-332206627:

> is it possible to have Ember 3.0 drop support for IE by default, and make it support IE 9/10 and 11 through different polyfill addons?

From @localpcguy in https://github.com/emberjs/rfcs/pull/252#issuecomment-332221208:

> Also, I think the idea of an addon for compatibility purposes is partially addressed in the Bring Your Own Compatibility section.

Confirm, the last section of the RFC is attempting to address exactly this question. The rough summary here is that we either will fully support a given browser (which means testing in CI of Ember itself and all of our "official addons") or we will not. Having a "bring your own compatibility" system is essentially a way to weasel out of making the correct commitments to our users which then puts the burden on each individual Ember user. To me, this is very much not the "Ember way"...

theseyi:
Just to additionally point out that with Phantom JS and older browsers, developers lose access to some ES6 features like [Proxies](https://kangax.github.io/compat-table/es6/#Proxy) which is not [poly-fillable](http://babeljs.io/learn-es2015/#ecmascript-2015-features-proxies), and have to resort to less elegant plumbing code. I think it's about time Ember moved in the new world.

corrspt:
Also in favor of this, my analytics (for my app) tell me that IE9/10 is mostly non-present. My `target.js` currently only includes IE11 and evergreen browsers.

I would very much like to drop support for IE11 as well but seeing as I still get several hits with it and also MS is going to support it until 2025 makes it not an option for now.

sly7-7:
@rwjblue I'm sorry, I guess I should have make a PR for this change...

rwjblue:
No worries! I apparently did some time traveling to start this RFC 😆

pichfl:
Very much in favour of this. We work with quite a few corporate clients in finance, automotive, chemicals and health and IE 11 was the lowest we would have to support this year.

sandstrom:
Totally in favour! 🏅

Our customers are medium to large enterprises and we restricted our app to IE11 + evergreen browsers three years ago and it has worked really well. In a few cases where they didn't have IE11 their IT-department didn't mind upgrading or adding Chrome to their fleet of computers.

Would it be possible to go further and drop IE11 too? Getting onto Edge (all evergreen browsers) would be quite valuable.

Browser statistics only look at the browser version, with no data on what other browsers are available on that computer. I'd argue that 9/10 enterprise companies with Win7/8 and IE11 — the typical (only?) IE user —  will also run Chrome on all workstations.

In other words, how many of the 8% IE11 (desktop; lower on mobile) users don't have Chrome or Firefox too? I'd guess that this percentage is quite low, perhaps only 1-2%.

Curious to hear what others think about this!

luxferresum:
@sandstrom absolutly not!
In an enterprise environment you usually only have one browser because noone wants to support multiple browsers.
One reason is that this would result in a non-standarized environment.
Another reason is that many companies have a strict policy to only use software with an SLA or at least a complex support concept. This results in using IE/Edge only, because its "part of" windows, and so no additional SLA/support concept is required.

This is especially true for intranet applications, where this is required during software evaluation.

Also dont forget all theese workstations without a "full" internet access, that only exist to use intranet applications.

Usually its OK to drop IE10 support, however IE11 is required for older windows versions, and even on Windows 10 IE11 is often more popular then Edge. Also sometimes older applications dont run well in Edge, which means the entire company will use IE11 only and so every new application has to support it.

broerse:
@sandstrom Windows server 2016 still only supports IE11 and not Edge. This is the current server software also used for terminal servers. It seems we need to support IE11 for the next 10 years. [more info]( https://social.technet.microsoft.com/Forums/ie/en-US/d3f563ab-3df9-4a8b-8e8d-4b6f043b4318/edge-browser-in-windows-server-2016-rds?forum=winserverTS)

lindyhopchris:
Totally in favour of this.

Just an idea (feel free to shoot it down)... what would be nice is some sort of fallback if your Ember app gets loaded in an unsupported browser. Whether that's in core, an addon, or a cookbook in the guides - something that would enable some HTML to be inserted into the DOM where the Ember app would have gone. I'm all in favour of dropping support for older browsers but it would be good to know that my Ember app will give the user some sort of feedback if the app cannot run in the browser that is viewing it.

raido:
@lindyhopchris I don't think that is a framework's concern. It could be some nicer ember-addon but event that will not go well. All the projects I am involved in are multilingual so I'd probably like to translate the message displayed into like 25 languages so if framework provided this it would get difficult and messy quickly.

sandstrom:
@lindyhopchris The difficulty with doing this within the framework is that the framework itself may not load on unsupported browsers.

So either this check is performed in a non-ember website (say the public/marketing website for an app) or it's checked server-side, before launching into the Ember app.

lindyhopchris:
@sandstrom @raido both good points, thanks!

jenweber:
Given that Microsoft itself has declared end of life for 9 &10, I'm on board. I care about security and don't actually want people using my app full of sensitive data on old, vulnerable browsers.

Continuing support for IE 11 is essential for work with large enterprise customers.

rwjblue:
We discussed this at the weekly Ember core team meeting, and based on the overwhelming community support we are moving this into final comment period.

Kuzirashi:
In our application we've agreed with business to drop support for IE9/IE10. I think Ember is good to go with doing the same. 👍

manuel-reil:
IE 11 should be supported. Corporate customers still rely on it.

Serabe:
I'm going to propose and try to give some reasons on why Ember 3 should drop support for IE11.

According to [IE EOL](https://www.microsoft.com/en-us/windowsforbusiness/end-of-ie-support), _"only the most current version of Internet Explorer available for a supported operating system will receive technical supports and security updates"_.

About Windows Server 2016, the _End of IE Support_ page states (bold letters mine):

>the last version of Internet Explorer _[...]_ will continue to receive security updates, compatibility fixes, and technical support on **Windows 7, Windows 8.1, and Windows 10**.

So I guess that takes it out of the equation.

Windows 8.1 mainstream EOL is [1/9/2018](https://support.microsoft.com/en-us/lifecycle/search?alpha=Windows%208.1) and that is the EOL of any Windows SO with IE11 and no alternative. Windows 10 will last until [October, 2020](https://support.microsoft.com/en-us/lifecycle/search?alpha=Windows%2010), but there is Edge in there.

## Proposed timeline to Ember 3.0

Ember 2.16 is scheduled for next week, and six weeks after that, when 2.17 is out, it will become next LTS.

On March next year, Ember 2.20 would be expected, and my proposal is to make it the last release and a special kind of LTS. Instead of receiving security and critical bugfixes for 6 months, it would receive them for one year. That would make Ember 2.20 supported until late April, 2019. That is over one year after Windows 8.1 EOL, though 18 months before Windows 10 EOL.

This would make Ember 3.0 come in April next year (that's over half a year) and give everybody time to adapt.

## But my client needs IE11

Your client would be covered for 20 months after Ember 3.0, time more than enough to plan their way out of IE into Edge, Chrome, or whatever they choose. I understand that being stuck for over one year in an old version is not ideal, but it is stuck with a browser almost four years old. A browser whose vendor is actively (kinda, certainly more than what we got used from Microsoft) discouraging its use.

We should not only consider what our clients need now, but what they will need when the support ends.

## What if?

If we were to drop IE11 that would mean supporting only evergreen browsers (humor me considering Edge evergreen, please). Imagine having to support only browsers six months old. More time to spend on new features without having to worry about IE.

What if I told you?

## Drawbacks

Well, codebases will diverge and supporting an LTS for one year will be challenging. On all sides. Maybe though your bug is pretty important for you it is not considered critical, specially if there is a workaround. Maybe some bugfixes, though critical, take so much time to get done that they just aren't.

Also, moving to browsers with support for ES2015+ can have drawbacks, as an example, we are using array destructuring in loose mode now.

## Other options

### Extended Support

Ember 2.x could be supported from April, 2019 (according to previous timeline) to October, 2020 by an Extended Support (as Windows has) offered by companies (who, when, how, and how much is a very different discussion).

### Short lived Ember 3.0 with a long LTS at the end.

Letting Ember 2.16 be the last LTS and getting Ember 3 out with support for IE11 but a planned short live (2 - 3 LTS) with a longer (~18 months) LTS at the end. That would make Ember 3 cover most of the Windows 10 life cycle, having a longer LTS cover the rest of it. This option is affected, though not in the same measure, by the same drawbacks as the original proposal.

t-sauer:
The problem with dropping IE11 support is not only that you have existing applications that would get stuck on an older version, but also (and in my opinion the bigger problem): Ember won't be a competitive option for new applications.

At my job we write smaller Ember apps for customers that still need IE support from time to time and that is probably not changing for a long time. If next year we would have to build new applications with an old Ember version, we would probably have to reevaluate if Ember is the best solution for us.  
Other companies that are not already using Ember and need IE support, wouldn't even consider Ember in the first place.

mixonic:
@Serabe unfortunately your comment got lapped by the publishing of https://emberjs.com/blog/2017/10/03/the-road-to-ember-3-0.html, so I think some of your scenarios are out of alignment with our plans already. Thank you for putting in such careful thought about the topic though and making a detailed pitch.

> If we were to drop IE11 that would mean supporting only evergreen browsers (humor me considering Edge evergreen, please). Imagine having to support only browsers six months old. More time to spend on new features without having to worry about IE.

I'm not convinced the evergreen story is this simple. For example [Google's crawler runs Chrome 41](https://developers.google.com/search/docs/guides/rendering). Some evergreen browsers like the popular [UC Browser](https://en.wikipedia.org/wiki/UC_Browser) are poorly understood. I'm not sure what is reasonable to support, if last two makes any sense.

In general we might be able to get away with some kind of rolling "last 1.5 years of releases" policy which would be pretty neat. Although for an application I might suggest "last two" is fine, for a framework I think we should be more conservative.

In summary I agree with your assessment this would be a nicer world than the one pinned to IE11, but I'm not convinced it is as rosy as you think.

> Short lived Ember 3.0 with a long LTS at the end.
>
> Letting Ember 2.16 be the last LTS and getting Ember 3 out with support for IE11 but a planned short live (2 - 3 LTS) with a longer (~18 months) LTS at the end. That would make Ember 3 cover most of the Windows 10 life cycle, having a longer LTS cover the rest of it. This option is affected, though not in the same measure, by the same drawbacks as the original proposal.

I think a short 3.0 is very plausible if we are successful with the "only removals" major release strategy in 3.0. 3.0 will be the test, if it is a smooth transition then we will have proven the formula for 4.0.

localpcguy:
@Serabe awesome writeup, but I think overly aggressive. I understand the reasons behind the desire to drop IE, and I don't want to see a repeat of the IE6/7/8 days where it took forever to remove them and we were supporting multiple versions of IE in some apps.

That said, I just don't believe it's feasible to drop IE11 support in the timeframe suggested, there will be companies using it through that final 18 months before EOL and probably beyond :( . And trying to maintain an LTS for 2 1/2 years isn't feasible, and would remove resources that could be working on pushing Ember forward. 

I think revisiting dropping IE11 for v4.0 is a good goal to have, or basing in on a usage % rather than calling out IE11 by name (no specific support for browsers with less than say 3% usage, for example).

As much as I'd love to drop IE11, I think it would hurt Ember adoption, and just flat rule it out for many companies.

Serabe:
@t-sauer I can understand and relate to that. I've worked for banks in the past and they are hard companies to work with. Still, I offered a different set of options to evaluate too. And my main point is that, starting September next year, every user of IE11 will have an Edge installed, unless they pay for Extended Support.

@mixonic I tried to find the current support matrix, but it is an open issue (see https://github.com/emberjs/ember.js/issues/10566 and https://github.com/emberjs/website/issues/2870) so I wrote that considering that the closest we currently have is the saucelabs list, the latest version of every evergreen browser, and the latest few of the rest (said here https://github.com/emberjs/ember.js/issues/10566#issuecomment-285593165). Clearly, once the assumptions are invalidated, some of the conclusions are as well. In any case, supporting 1.5 years old browsers is still much better than supporting 3 years old. Even [Chrome 41 implements more features than some current browsers](https://www.chromestatus.com/features#milestone%3C%3D41) (Safari, I'm looking at you). And finally, yes. While I would love to see IE go now, I understand that is not feasible for many projects, that is why I proposed other options as well.

@localpcguy sorry if it sounded aggressive, it was far from my intention. I've been editing that post for a week now, and revisiting it I understand that I failed in some cases. I apologise for that. I really liked your idea of no specific support for browsers with less than x%. That is a nice rule, though hard to define (whose metrics? for how long do a browser need to be under that threshold to be dropped?), but worth to try to revisit for dropping IE11 before Windows 10 goes away.

Thank you all!

cibernox:
Even if we don't drop support for IE11 now, and as the 3.0 release proves, Ember does not have a fixed amount of time between mayor releases, so even if Ember 3.0 supports IE11, that doesn't mean that we have to wait 2.5 years again for Ember 4.0. 
It could come whenever we feel we can drop support for IE11, even if that is in 10 months time.

localpcguy:
@serabe FWIW, I didn't mean aggressive as in antagonistic, rather just in the sense of being bold or audacious - nothing to apologize for!

elishaebenezer:
Since Chrome and Firefox both support headless mode, why not include support for headless firefox as well?

rwjblue:
@elishaebenezer - Not sure what you mean.  We definitely support Firefox...

broerse:
@Serabe Windows Server 2016 is the current server version used for terminal servers and only supports IE11 and you can't install Edge. Lets first wait for Edge on Windows 2016 Server before thinking about dropping the default browser for Windows Server.

btecu:
@broerse I don't think Windows Server should be considered. Windows Server is for servers and so if you decide to browse the Internet on the server, I'd at least hope you'd install a decent browser.

broerse:
@btecu I see you don't use Ember in a corporate environment but Terminal Servers are used for corporate users everywhere. They can't install chrome on their thin clients.

elishaebenezer:
> Not sure what you mean. We definitely support Firefox...

@rwjblue 

Robert, 

Like Chrome from 59, Firefox from 56 supports launching in headless mode. 

Given that We are moving away from PhantomJS and started supporting headless chrome, will it not be a better idea to support "headless firefox" in environments where chrome may not be available?

rwjblue:
@elishaebenezer - Sorry for my earlier terseness. I basically meant that we definitely support headless Firefox already. Configuring a default ember-cli generated app to use headless Firefox is pretty simple (requires tweaking `testem.js` and `.travis.yml`). We chose headless chrome as the default for newly generated ember-cli apps due to popularity of the browser generally speaking, but both (and many more) browsers are definitely supported...

rwjblue:
FWIW, I personally would be totally happy to entertain migrating ember-cli's defaults for apps and addons to test in _both_ headless Firefox and headless Chrome.  That would likely be a separate RFC (at ember-cli/rfcs I believe)...

elishaebenezer:
@rwjblue Thank you very much for your clarification and proposal.

Given the latest advancements in Firefox (Quantum), it would be a very good option to consider

tomdale:
Thank you to everyone who has participated in this RFC discussion. No major new issues have been raised during the Final Comment Period, so we would like to get this merged. And you all have my deepest personal gratitude for allowing me to stop caring about IE10. You're all the best! 🎉

PR #260, opened 2017-10-18
Serializer Test Helpers

amiel:
[Rendered](https://github.com/amiel/rfcs/blob/serializer-test-helpers/text/0000-serializer-test-helpers.md)

### DIFF
--- a/text/0000-serializer-test-helpers.md
+++ b/text/0000-serializer-test-helpers.md
@@ -33,30 +33,38 @@ I think would be useful.
    Here is an example test:
 
    ```js
-   const EXAMPLE_QUERY_RECORD_RESPONSE = {
-     data: [{
-       id: '10',
-       type: 'user',
-       attributes: {
-         name: 'Tomster',
-       },
-     }],
-   };
+   import { module, test } from 'qunit';
+   import { setupTest } from 'ember-qunit';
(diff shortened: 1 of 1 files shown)


rwjblue:
In general, I’m in favor of addons providing their own test helpers to make the developer ergonomics better for testing. In this case, I would think this type of test helper would be provided by ember-data itself (as opposed to ember-test-helpers).

I think this proposal needs a bit of reworking to account for changes in the testing APIs since it was originally proposed (at the least, the examples need to be tweaked).

amiel:
@rwjblue Thanks for the response. I agree it makes sense for ember-data to provide the helpers. I will update the examples.

rwjblue:
We discussed this at the ember-data meeting today, given the new testing APIs (updating in this RFC back in March of 2018) we think that what you'd need to do to test (with only public APIs) is good enough. We are going to move this RFC into final comment period to close. 

We'd absolutely love help reviewing the guides on testing these things (as well as the blueprints) to ensure that they work well and make it clear what is going on. 

A side effect of the conversation that we had was to clarify that we think that we need to continue honing the ability to assert that a serializer does return the correct normalized form and a testing specific helper in this space is a great follow up RFC.

amiel:
👍

PR #266, opened 2017-10-31
[Ember Data] load relationships without data or links

amiel:
[Rendered](https://github.com/amiel/rfcs/blob/ember-data-load-relationships-without-data/text/0000-load-relationships-without-data.md)


runspired:
Hi @amiel 

Sorry for letting this sit so long. A few notes:

- We've improved relationship state management greatly in recent versions, including ensuring that `links` are used when necessary and ensuring that we are clear about when we hit network.
- We definitely should hit network in the no-link, no-data, not-explicitly-empty case; however, it's likely this request is swallowed due to there being no information to use in the query. We are working on a new request management layer and continuing work to refactor the relationship layer, which will help to enable us to ensure a request always happens here with proper context.
- I think a better alternative to the RFC would be an RFC to ensure that fetching a hasMany with `data` also uses the `findHasMany` code-path. Not only would such an RFC help us iterate towards a nicer query management story with this request-manager, it would be able to resolve this corner case elegantly much sooner :)  One could imagine that `findHasMany` would delegate to `findMany` as it does today in the case where there is no `link` present, to preserve existing behavior while simplifying the overall flow and enabling this case to be handled.

Thanks so much for digging into this sticky point and pointing out these issues!

amiel:
Hi @runspired. Thank you for the response.

I think your idea of

> an RFC to ensure that fetching a `hasMany` with data also uses the `findHasMany` code-path

would accomplish my goal.

I think it is basically what I meant with

> Invert the logic for deciding when to load a relationship by link vs by data ([from here](https://github.com/amiel/rfcs/blob/ember-data-load-relationships-without-data/text/0000-load-relationships-without-data.md#detailed-design))

The main reason I proposed new hooks is to ensure backwards compatibility, in case existing applications were depending on the current behavior of these hooks.

I am in between jobs right now and not in a position to look in to the ins and outs of your suggestion and write a new RFC, so I will bow out at this point, although I may be involved again depending on the nature of my next job :)

Shall I close this PR?

runspired:
@amiel I think we should close this one in favor of another one, but I hope you can be involved!

PR #267, opened 2017-11-03
RFC: Deprecate `link-to` / `disabledWhen`

bekzod:
[Rendered](https://github.com/bekzod/ember-rfcs/blob/master/text/0000-deprecate-link-to-disabled-when.md)

### DIFF
--- a/text/0000-deprecate-link-to-disabled-when.md
+++ b/text/0000-deprecate-link-to-disabled-when.md
@@ -1,0 +1,30 @@
+- Start Date: (fill me in with today's date, YYYY-MM-DD)
+- RFC PR: (leave this empty)
+- Ember Issue: (leave this empty)
+
+# Summary
+
+This RFC aims to deprecate usage of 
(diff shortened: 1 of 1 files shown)


jamesarosen:
`{{link-to}}` also has a `currentWhen` property whose implementation is backed by `routing.isActiveForRoute`. It would be neat to extract a `helper:is-active-for-route` so clients can build up more expressive logic for both `current` and `disabled`.

rwjblue:
@jamesarosen - I believe that is what [ember-router-helpers's is-active helper](https://github.com/rwjblue/ember-router-helpers/blob/master/addon/helpers/is-active.js) does...

bekzod:
@rwjblue can we move this final comment period ?

bertdeblock:
Can we close this since `@disabledWhen` [is now deprecated](https://github.com/emberjs/ember.js/blob/191389cdb3dfc893de449d12cdc71ec8c0a66e24/packages/%40ember/-internals/glimmer/lib/components/link-to.ts#L624-L639)?

PR #268, opened 2017-11-06
Rethink Acceptance Testing

rwjblue:
Continues the testing related RFC's (slowly chipping away at https://github.com/emberjs/rfcs/pull/119) by proposing migrating the acceptance testing API "unify" (though I'm not allowed to use that word in the title any more) with the newly landed and implemented ember-qunit API's (proposed in #232). 

[rendered](https://github.com/emberjs/rfcs/blob/acceptance-testing/text/0268-acceptance-testing-refactor.md)

### DIFF
--- a/text/0268-acceptance-testing-refactor.md
+++ b/text/0268-acceptance-testing-refactor.md
@@ -237,9 +237,13 @@ being proposed here:
   well as importable helpers.
 * `this.owner` will now be present and allow (for the first time 🎉) overriding
   items in the container/registry.
-* The new system will leverage the `Ember.Application` / `Ember.ApplicationInstance` split so that
-  we can avoid creating an `Ember.Application` instance per-test, and instead leverage the same system
-  that FastBoot itself uses to avoid running initializers for each acceptance test.
+* The new system will leverage the `Ember.Application` /
+  `Ember.ApplicationInstance` split so that we can avoid creating an
+  `Ember.Application` instance per-test, and instead leverage the same system
+  that FastBoot itself uses to avoid running initializers for each acceptance
+  test.
+* Implicit promise chaining will no longer be present. If your test needs to
+  wait for a given promise, it should use `await`.
 
 ## Migration
 
--- a/text/0268-acceptance-testing-refactor.md
+++ b/text/0268-acceptance-testing-refactor.md
@@ -76,11 +76,11 @@ test('should add new post', function(assert) {
 
 // **** after ****
 import { module, test } from 'qunit';
-import { setupAcceptanceTest } from 'ember-qunit';
+import { setupApplicationTest } from 'ember-qunit';
 import { visit, fillIn, click } from '@ember/test-helpers';
 
 module('Acceptance | login', function(hooks) {
-  setupAcceptanceTest(hooks);
+  setupApplicationTest(hooks);
 
   test('should add new post', async function(assert) {
     await visit('/posts/new');
@@ -102,14 +102,14 @@ The following new methods will be exposed from `ember-qunit`:
 ```ts
 declare module 'ember-qunit' {
   // ...snip... 
-  export function setupAcceptanceTest(hooks: QUnitModuleHooks): void;
+  export function setupApplicationTest(hooks: QUnitModuleHooks): void;
 }
 ```
 
 ### DOM Interaction Helpers
 
 New native DOM interaction helpers will be added to both `setupRenderingTest`
-and (proposed below) `setupAcceptanceTest`. The implementation for these
+and (proposed below) `setupApplicationTest`. The implementation for these
 helpers has been iterated on and is quite stable in the
 [ember-native-dom-helpers](https://github.com/cibernox/ember-native-dom-helpers)
 addon.
@@ -182,7 +182,7 @@ export function waitUntil<T>(Function: Promise<T>, { timeout = 1000 }): Promise<
 export function waitFor(selector: string, { count?: number, timeout = 1000 }): Promise<HTMLElement | HTMLElement[]>;
 ```
 
-### `setupAcceptanceTest`
+### `setupApplicationTest`
 
 This function will:
 
@@ -224,7 +224,7 @@ The `setupRenderingTest` function proposed in
 * setup importable `waitUntil` helper method
 * setup importable `waitFor` helper method
 
-Once implemented, `setupRenderingTest` and `setupAcceptanceTest` will diverge from each other in very few ways.
+Once implemented, `setupRenderingTest` and `setupApplicationTest` will diverge from each other in very few ways.
 
 ## Changes from Current System
 
@@ -289,11 +289,11 @@ This allows both the prior API (without modification) and the following:
 ```js
 // Option 2:
 import { module, test } from 'qunit';
-import { setupAcceptanceTest } from 'ember-qunit';
+import { setupApplicationTest } from 'ember-qunit';
 import { enableFeature } from 'addon-name-here/test-support';
 
 module('asdf', function(hooks) {
-  setupAcceptanceTest(hooks);
+  setupApplicationTest(hooks);
 
   test('awesome test title here', function(assert) {
     enableFeature(this.owner, 'feature-name-here');
--- a/text/0268-acceptance-testing-refactor.md
+++ b/text/0268-acceptance-testing-refactor.md
@@ -121,7 +121,7 @@ already adopted will have very minimal changes to make).
 
 The specific DOM helpers to be added to the `@ember/test-helpers` module are:
 
-```
+```ts
 /**
   Clicks on the specified selector.
 */
--- a/text/0268-acceptance-testing-refactor.md
+++ b/text/0268-acceptance-testing-refactor.md
@@ -87,7 +87,7 @@ module('Acceptance | login', function(hooks) {
     await fillIn('input.title', 'My new post');
     await click('button.submit');
 
-    assert.equal(find('ul.posts li')[0].textContent, 'My new post');
+    assert.equal(this.element.querySelectorAll('ul.posts li')[0].textContent, 'My new post');
   });
 });
 ```
--- a/text/0268-acceptance-testing-refactor.md
+++ b/text/0268-acceptance-testing-refactor.md
@@ -12,10 +12,13 @@ you would like.
 Unfortunately, there is a massive disparity between different types of tests.
 In acceptance tests, you use well designed global helpers to deal with async
 related interactions; whereas in integration and unit tests you are forced to
-manually deal with this asynchrony. The goal of this RFC is to unify the
-concepts amongst the various types of test and provide a single common
-structure to tests.
-
+manually deal with this asynchrony.
+[emberjs/rfcs#232](https://github.com/emberjs/rfcs/blob/master/text/0232-simplify-qunit-testing-api.md)
+introduced us to QUnit's nested modules API, made integration and unit testing
+modular, and greatly simplified the concepts needed to learn how to write unit
+and integration tests. The goal of this RFC is to leverage what we have learned
+in prior RFCs and apply that knowledge to acceptance testing. Once this RFC has
+been implemented all test types in Ember will have a unified cohesive structure.
 
 # Motivation
 
@@ -25,7 +28,7 @@ example), and assumes most (if not all) interactions are synchronous.  This is
 a major issue due to the fact that the vast majority of interactions will
 actually be asynchronous. There have been a few recent additions to
 `ember-test-helpers` that have made dealing with asynchrony better (namely
-[emberjs/rfcs#232](https://github.com/emberjs/rfcs/blob/master/text/0232-simplify-qunit-testing-api.md)
+[emberjs/rfcs#232](https://github.com/emberjs/rfcs/blob/master/text/0232-simplify-qunit-testing-api.md))
 but forcing users to manually manage all interaction based async is a recipe
 for disaster.
 
--- a/text/0268-acceptance-testing-refactor.md
+++ b/text/0268-acceptance-testing-refactor.md
@@ -1,0 +1,224 @@
+- Start Date: 2017-11-05
+- RFC PR: [emberjs/rfcs#268](https://github.com/emberjs/rfcs/pull/268)
+- Ember Issue: (leave this empty)
+
+# Summary
+
+The testing story in Ember today is better than it ever has been. It is now
+possible to test individual component/template combos, register your own mock
+components/services/etc, build complex acceptance tests, and almost anything else
+you would like.
+
+Unfortunately, there is a massive disparity between different types of tests.
+In acceptance tests, you use well designed global helpers to deal with async
+related interactions; whereas in integration and unit tests you are forced to
+manually deal with this asynchrony. The goal of this RFC is to unify the
+concepts amongst the various types of test and provide a single common
+structure to tests.
+
+
+# Motivation
+
+Usage of rendering tests is becoming more and more common, but these tests
+often still include manual event delegation (`this.$('.foo').click()` for
+example), and assumes most (if not all) interactions are synchronous.  This is
+a major issue due to the fact that the vast majority of interactions will
+actually be asynchronous. There have been a few recent additions to
+`ember-test-helpers` that have made dealing with asynchrony better (namely
+[emberjs/rfcs#232](https://github.com/emberjs/rfcs/blob/master/text/0232-simplify-qunit-testing-api.md)
+but forcing users to manually manage all interaction based async is a recipe
+for disaster.
+
+Acceptance tests allow users to handle asynchrony with ease, but they rely on
+global helpers that automatically wrap a single global promise which makes
+testing of interleaved asynchronous things more difficult. There are a number
+of limitations in acceptance tests as compared to integration tests (cannot
+mock and/or stub services, cannot look up services to setup test context, etc).
+
+We need a single unified way to teach and understand testing in Ember that
+leverages all the things we learned with the original acceptance testing
+helpers that were introduced in Ember 1.0.0.  Instead of inventing our own
+syntax for dealing with the async (`andThen`) we should use new language
+features such as `async` / `await`.
+
+# Detailed design
+
+The goal of this RFC is to introduce new system for acceptance tests that follows in the footsteps of
+[emberjs/rfcs#232](https://github.com/emberjs/rfcs/blob/master/text/0232-simplify-qunit-testing-api.md)
+and continues to enhance the system created in that RFC to share the same structure and helper system.
+
+This new system for acceptance tests will be implemented in the
+[ember-test-helpers](https://github.com/emberjs/ember-test-helpers/) library so
+that we can iterate faster while supporting multiple Ember versions
+independently and easily support multiple testing frameworks build on top of
+the primitives in `ember-test-helpers`. Ultimately, the existing [ember-testing](https://github.com/emberjs/ember.js/tree/master/packages/ember-testing) system
+will be deprecated but that deprecation will be added well after the new system has been
+released and adopted by the community. 
+
+Lets take a look at a basic example (lifted from [the guides](https://guides.emberjs.com/v2.16.0/testing/acceptance/)):
+
+```js
+// **** before ****
+import { test } from 'qunit';
+import moduleForAcceptance from '../helpers/module-for-acceptance';
+
+moduleForAcceptance('Acceptance | posts');
+
+test('should add new post', function(assert) {
+  visit('/posts/new');
+  fillIn('input.title', 'My new post');
+  click('button.submit');
+  andThen(() => assert.equal(find('ul.posts li:first').text(), 'My new post'));
+});
+
+// **** after ****
+import { module, test } from 'qunit';
+import { setupAcceptanceTest } from 'ember-qunit';
+import { visit, fillIn, click } from 'ember-test-helpers';
+
+module('Acceptance | login', function(hooks) {
+  setupAcceptanceTest(hooks);
+
+  test('should add new post', async function(assert) {
+    await visit('/posts/new');
+    await fillIn('input.title', 'My new post');
+    await click('button.submit');
+
+    assert.equal(find('ul.posts li')[0].textContent, 'My new post');
+  });
+});
+```
+
+As you can see, this proposal is unifies on Qunit's nested module syntax following
+in emberjs/rfcs#232's footsteps.
+
+## New APIs Proposed
+
+The following new methods will be exposed from `ember-qunit`:
+
+```ts
+declare module 'ember-qunit' {
+  // ...snip... 
+  export function setupAcceptanceTest(hooks: QUnitModuleHooks): void;
+}
+```
+
+### `setupAcceptanceTest`
+
+This function will:
+
+* invoke `ember-test-helper`s `setupContext` with the tests context (which does the following):
+  * create an owner object and set it on the test context (e.g. `this.owner`)
+  * setup `this.set`, `this.setProperties`, `this.get`, and `this.getProperties` to
+    the test context
+  * setup `this.pauseTest` and `this.resumeTest` methods to allow easy pausing/resuming
+    of tests
+* add routing related helpers
+  * setup `this.visit` method to visit the given url
+  * setup getter for `this.currentRouteName` which returns the current route name
+  * setup getter for `this.currentURL` which returns the current URL
+* add DOM interaction helpers (heavily influenced by @cibernox's lovely addon [ember-native-dom-helpers](https://github.com/cibernox/ember-native-dom-helpers))
+  * setup a getter for `this.element` which returns the DOM element representing
+    the root element
+  * if `jQuery` is present in the application sets up `this.$` method to run
+    jQuery selectors rooted to `this.element`
+  * setup `this.click` as an async helper to click on the specified selector
+  * setup `this.tap` as an async helper to tap on the specified selector
+  * setup `this.triggerKeyEvent` as an async helper to trigger a `KeyEvent` on the specified selector
+  * setup `this.triggerEvent` as an async helper to trigger an event on the specified selector
+  * setup `this.fillIn` as an async helper to enter text on the specified selector
+  * setup `this.waitUntil` as an async helper returning a promise which resolves when the provided callback returns a truthy value
+  * setup `this.focus` as an async helper which focuses the specified element
+  * setup `this.blur` as an async helper which unfocuses the specified element
+
+### `setupRenderingTest`
+
+The `setupRenderingTest` function proposed in
+[emberjs/rfcs#232](https://github.com/emberjs/rfcs/blob/master/text/0232-simplify-qunit-testing-api.md)
+(and implemented in
+[ember-qunit](https://github.com/emberjs/ember-qunit)@3.0.0) will be modified to add the same DOM interaction helpers:
+
+* setup a getter for `this.element` which returns the DOM element representing
+  the root element
+* if `jQuery` is present in the application sets up `this.$` method to run
+  jQuery selectors rooted to `this.element`
+* setup `this.click` as an async helper to click on the specified selector
+* setup `this.tap` as an async helper to tap on the specified selector
+* setup `this.triggerKeyEvent` as an async helper to trigger a `KeyEvent` on the specified selector
+* setup `this.triggerEvent` as an async helper to trigger an event on the specified selector
+* setup `this.fillIn` as an async helper to enter text on the specified selector
+* setup `this.waitUntil` as an async helper returning a promise which resolves when the provided callback returns a truthy value
+* setup `this.focus` as an async helper which focuses the specified element
+* setup `this.blur` as an async helper which unfocuses the specified element
+
+Once implemented, `setupRenderingTest` and `setupAcceptanceTest` will diverge from each other in very few ways.
+
+### Importable helpers
+
+Since the design of this system relies on both the test helpers being applied
+to the test context **and** the usage of `async` / `await`, a few importable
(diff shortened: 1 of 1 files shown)


text/0268-acceptance-testing-refactor.md
```diff
@@ -0,0 +1,223 @@
...
+Once implemented, `setupRenderingTest` and `setupAcceptanceTest` will diverge from each other in very few ways.
+
+### Importable helpers
+
+Since the design of this system relies on both the test helpers being applied
+to the test context **and** the usage of `async` / `await`, a few importable
+helpers are being introduced to help avoid extra noise (e.g. "rightward shift")
+in tests.
+
+This means that users will be able to use either of the following interchangably:
```
simonihmig: @Do we really want/need two equivalent ways to use those helpers, or should'n there be one "right" way to do this? In the sense of strong Ember conventions, so teams don't end up bikeshedding about the right way. And maybe also useful for easier tooling (linting rules, codemods)? Just a thought...

text/0268-acceptance-testing-refactor.md
```diff
@@ -0,0 +1,223 @@
...
+  // ...snip...
+});
+```
+
+## Changes from Current System
+
+Here is a brief list of the more important but possibly understated changes
+being proposed here:
+
+* The global test helpers that exist now, will no longer be present (e.g.
```
simonihmig: Really not present, or rather deprecated? So they can co-exist, see the following section.

rwjblue:
> But could you maybe explain the relationship between this and #119? Does it supercede the older RFC, at least with regard to acceptance testing?

Yes, basically #119 is too large and too sprawling to _actually_ implement.  I started chopping it up into smaller more bite sized RFCs (like #232 and this one) so that we can make incremental progress forward.  The main thing left in #119 that still is not addressed by either #232 or this RFC is the auto-discovery of test helpers and hooks ([this section](https://github.com/rwjblue/rfcs/blob/42/text/0000-grand-testing-unification.md#general-hooks)).

text/0268-acceptance-testing-refactor.md
```diff
@@ -0,0 +1,223 @@
...
+  * create an owner object and set it on the test context (e.g. `this.owner`)
+  * setup `this.set`, `this.setProperties`, `this.get`, and `this.getProperties` to
+    the test context
+  * setup `this.pauseTest` and `this.resumeTest` methods to allow easy pausing/resuming
+    of tests
+* add routing related helpers
+  * setup `this.visit` method to visit the given url
+  * setup getter for `this.currentRouteName` which returns the current route name
+  * setup getter for `this.currentURL` which returns the current URL
+* add DOM interaction helpers (heavily influenced by @cibernox's lovely addon [ember-native-dom-helpers](https://github.com/cibernox/ember-native-dom-helpers))
```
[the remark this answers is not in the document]
rwjblue: Yes, exactly. My plan (and discussed with @cibernox and @turbo87 a couple of times) is to essentially copy the native DOM implementations from `ember-native-dom-helpers` (the [addons own README](https://github.com/cibernox/ember-native-dom-helpers#ember-native-dom-helpers) infers that it is specifically aiming for this).

text/0268-acceptance-testing-refactor.md
```diff
@@ -0,0 +1,223 @@
...
+  * setup a getter for `this.element` which returns the DOM element representing
+    the root element
```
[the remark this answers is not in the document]
rwjblue: @simonihmig - Yes. I will update the RFC though.

text/0268-acceptance-testing-refactor.md
```diff
@@ -0,0 +1,223 @@
...
+```
+
+### `setupAcceptanceTest`
+
+This function will:
+
+* invoke `ember-test-helper`s `setupContext` with the tests context (which does the following):
+  * create an owner object and set it on the test context (e.g. `this.owner`)
+  * setup `this.set`, `this.setProperties`, `this.get`, and `this.getProperties` to
+    the test context
```
Turbo87: why do we need those methods?

text/0268-acceptance-testing-refactor.md
```diff
@@ -0,0 +1,223 @@
...
+    of tests
+* add routing related helpers
+  * setup `this.visit` method to visit the given url
+  * setup getter for `this.currentRouteName` which returns the current route name
+  * setup getter for `this.currentURL` which returns the current URL
+* add DOM interaction helpers (heavily influenced by @cibernox's lovely addon [ember-native-dom-helpers](https://github.com/cibernox/ember-native-dom-helpers))
+  * setup a getter for `this.element` which returns the DOM element representing
+    the root element
+  * if `jQuery` is present in the application sets up `this.$` method to run
+    jQuery selectors rooted to `this.element`
```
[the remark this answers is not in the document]
rwjblue: I had written a very long explanation about how you were wrong, at the end of which I realized you were right...

Suggesting that folks use `$(selector, this.element)` as needed (and from the codemod) seems absolutely fine to avoid creating a cliff.

text/0268-acceptance-testing-refactor.md
rwjblue: @simonihmig - Yes, the blueprints will use the imported helpers. The only reason I describe both is that the importable helpers _defer_ to the local versions on the test context (essentially as a desugaring).  Do you think I need to clarify that a bit?

rwjblue: I updated some verbiage in this section to hopefully clarify things a bit better.

text/0268-acceptance-testing-refactor.md
```diff
@@ -0,0 +1,223 @@
...
+confusion, making it easier to teach and understand testing in Ember.
+
+# Drawbacks
+
+* This is a relatively large set of changes that are arguably not needed (things mostly work today).
+* One of the major hurdles in upgrading larger applications to newer Ember versions, is updating their tests to follow "new" patterns.  This RFC introduces yet another "new" thing (and proposes to deprecate the old thing), and could therefore be considered "just more churn".
+
+# Alternatives
+
+* Do nothing?
```
NLincoln: An alternative (or maybe just a suggestion?) would be to make `ember-native-dom-helpers` into a semi-official addon. Instead of copying the source code into `ember-test-helpers`, why not start shifting the helper code out of the core and into an addon?

text/0268-acceptance-testing-refactor.md
rwjblue: When running tests with the old `moduleForAcceptance` API they will be present, this section is referring to tests using `setupAcceptanceTest(...)` API and is correct in saying that the global methods _will not_ be present.

text/0268-acceptance-testing-refactor.md
rwjblue: @Turbo87:

The main reason `set` / `setProperties` methods exist today, is that sets during testing trigger an assertion if not manually wrapped in a runloop. This is something that we are actively working to remove, but that effort will likely take time (as it essentially requires Ember 3.0 where we can rely on all supported platforms having a microtask queue).  

Once that deficiency is addressed, and we no longer need to use manual runloop wrapping during set operations in testing mode, we should consider deprecating (in favor of simply setting them as appropriate) them...

text/0268-acceptance-testing-refactor.md
rwjblue: Great question, I will add this to the alternatives section.

`ember-test-helpers` is an addon 😸 (completely separate from Ember's own source), supports many Ember versions simultaneously, is focused on providing consistent testing experiences for Ember users, is _already_ a dependency of every application that uses ember-qunit or ember-mocha (roughly all projects), and is a better name 😈.

text/0268-acceptance-testing-refactor.md
Turbo87: thanks for the explanation, but I still don't see in what situations one would use e.g. `this.set()` in an **acceptance** test 🤔

text/0268-acceptance-testing-refactor.md
cibernox: It is indeed an alternative, although I think it makes more sense to make `ember-test-helpers` to be THE PLACE where all "official" testing stuff (that is not qunit-specific) lives.

text/0268-acceptance-testing-refactor.md
```diff
@@ -0,0 +1,304 @@
...
+  * setup `this.waitFor` helper method
+
+### `setupRenderingTest`
+
+The `setupRenderingTest` function proposed in
+[emberjs/rfcs#232](https://github.com/emberjs/rfcs/blob/master/text/0232-simplify-qunit-testing-api.md)
+(and implemented in
+[ember-qunit](https://github.com/emberjs/ember-qunit)@3.0.0) will be modified to add the same DOM interaction helpers mentioned above:
+
+* setup `this.click` helper method
```
cibernox: I still find weird that we have two supported ways of doing the same thing. Also, specially in integration tests, it would be extraordinary that someone could do 
```
this.click = function(e) {
  assert.ok(e, 'The event is fired')
}
await render(hbs`{{my-component action=click}}`)
```

Do you think that supposed improvement in debuggeablility justifies this duality. Couldn't just the docs serve as ... documentation of what helpers exists and how to use them?

Turbo87: FYI There is one potential advantage of using `await this.click()` over `await click()` in that you could potentially run tests concurrently. The `click()` variant has no reliable way of figuring out which test just called that helper, while `this.click()` is bound to the right context.

text/0268-acceptance-testing-refactor.md
rwjblue: Ahh, I see what you are saying now, sorry about that.  I agree, I don't think we strictly need them (and I suspect 99.9% of tests will not use them anyways). I included them here so that the differences between the various types of tests are as small as possible...

text/0268-acceptance-testing-refactor.md
rwjblue: > There is one potential advantage of using await this.click() over await click() in that you could potentially run tests concurrently.

Confirm, however I tried to avoid talking about this in #119 (and here) because there are many other issues with running tests in parallel (there is only one DOM, focus and other events don't fire the same, etc) and I didn't want to open a can of worms.

cibernox: Even if I slightly favour the imported helpers, my point is about having both. I think it should be one or the other.

text/0268-acceptance-testing-refactor.md
Turbo87: I think I'd prefer to not include them, so that people don't get the wrong idea about the things that they could use

text/0268-acceptance-testing-refactor.md
rwjblue: > Do you think that supposed improvement in debuggeablility justifies this duality? Couldn't just the docs serve as ...well... documentation of what helpers exists and how to use them?

Sure, I agree that this is definitely a good suggestion also. I'll add a fleshed out version of it to alternatives.

rwjblue: > Even if I slightly favour the imported helpers, my point is about having both. I think it should be one or the other.
  
Right, totally understood. I guess I'm saying that the fact that there are both is an implementation detail. The primary public API and what blueprints will use will be the importable versions. The fact that they desugar to invoking methods on the test context is a minor detail (and _might_ enable features like what @Turbo87 alludes to in https://github.com/emberjs/rfcs/pull/268#discussion_r149123309).

text/0268-acceptance-testing-refactor.md
simonihmig: That makes sense! Maybe this should be stated to prevent misunderstandings like I had?

text/0268-acceptance-testing-refactor.md
simonihmig: Agree. Made the same point [here](https://github.com/emberjs/rfcs/pull/268#discussion_r149086095). @rwjblue just added some clarification [here](https://github.com/emberjs/rfcs/pull/268/commits/4bb68572760b76d94b19809d0da84d186e05ec8b), which seems reasonable. I would not see it as a problem to have both, as long as there is *one* ideomatic way to write your tests, that is consistently used (guides, blueprints, codemods etc.)

text/0268-acceptance-testing-refactor.md
rwjblue: Seems ok to me. We don't expect folks to need them, seems fine to avoid adding them. Will require some additional changes to the existing system in `ember-test-helpers`, but that's quite simple anyways.

text/0268-acceptance-testing-refactor.md
rwjblue: Oh, y'all did it now.  Just removed the `this.foo` versions from the RFC.  I agree having two different public API's for the same thing is confusing, and reserving a bunch of names on the test context is annoying and error prone for users. 

It is still likely that during implementation I will have _some_ mechanism to discover and access these helpers on the test context but it does not need to be public API at all...

rtablada:
@rwjblue I like the idea of moving the functions from `ember-qunit` to `ember-test-helpers`.

I would mention that the summary does a good job of describing motivations, but doesn't really give a top level explanation of what changes are being proposed and assumes that contributors are deeply in the know on testing unification and prior RFCs.
I think we should add a sentence of the actual impact on how this moves the needle forward.

text/0268-acceptance-testing-refactor.md
```diff
@@ -0,0 +1,272 @@
...
+  visit('/posts/new');
+  fillIn('input.title', 'My new post');
+  click('button.submit');
+  andThen(() => assert.equal(find('ul.posts li:first').text(), 'My new post'));
+});
+
+// **** after ****
+import { module, test } from 'qunit';
+import { setupAcceptanceTest } from 'ember-qunit';
+import { visit, fillIn, click } from 'ember-test-helpers';
```
Turbo87: as mentioned in https://github.com/emberjs/ember-test-helpers/issues/240 we might want to move to `@ember/test-helpers` for the import paths

rwjblue:
@rtablada 

> assumes that contributors are deeply in the know on testing unification and prior RFCs

I was trying to avoid this by cross-link to any prior work that I referred to, did I miss some references?

> I think we should add a sentence of the actual impact on how this moves the needle forward.

Sounds good, I've tried to do a better job and pushed those changes in b05ed2c. Mind reviewing?

rtablada:
@rwjblue seems like the main big changes are:

1. Moving helpers from `ember-qunit` to `ember-test-helpers`
2. Favor `async`/`await`
3. Change to typescript in the dead of night :P

rwjblue:
> Change to typescript in the dead of night :P

No one is suggesting this.

rtablada:
I was kidding. But the `declare module` syntax may be a bit confusing to people.

rwjblue:
@rtablada 

> Moving helpers from ember-qunit to ember-test-helpers

This isn't actually true. Nothing that `ember-qunit` is currently providing is actually moving into `ember-test-helpers`, what am I missing?

> Favor async/await

Indeed, but this isn't really the point at all. If an app doesn't want to use async/await they can happily continue to use promises. This is definitely not "summary" material IMHO.

rwjblue:
> the declare module syntax may be a bit confusing to people

Yeah, I was worried about that.  I was trying to find a way to clearly define the **exact** API (inputs, outputs, etc), so I just lifted TypeScript syntax and added some inline docs for prose.  Happy to tweak to another format, but I am pretty opposed to "just describe the API in prose" because I personally find that extremely hard to read and understand.

text/0268-acceptance-testing-refactor.md
rwjblue: Good point, updated in ebfdbe1.

rtablada:
I think just removing the wrapping `declare module 'ember-test-helpers'` still tells the full story.
Maybe switching to comments for typehints?

rtablada:
> Nothing that ember-qunit is currently providing is actually moving into ember-test-helpers, what am I missing?

I think it's still unclear what the actual being proposed is then.

[In this section](https://github.com/emberjs/rfcs/blob/acceptance-testing/text/0268-acceptance-testing-refactor.md#dom-interaction-helpers) it seems like you are proposing to move the helpers into the `ember-test-helpers` package/module.

rtablada:
IMO moving: https://github.com/emberjs/rfcs/blob/acceptance-testing/text/0268-acceptance-testing-refactor.md#changes-from-current-system higher up will make it easier for people to understand the impact before looking at the pseudo implementation.

rwjblue:
> In this section it seems like you are proposing to move the helpers into the ember-test-helpers package/module.

Yep, exactly! But they aren't moving from `ember-qunit` (they were never there). They were previously provided by an asset generated by Ember itself.

rwjblue:
> IMO moving: https://github.com/emberjs/rfcs/blob/acceptance-testing/text/0268-acceptance-testing-refactor.md#changes-from-current-system higher up will make it easier for people to understand the impact before looking at the pseudo implementation.

I actually find that having a general before and after snippet (even if not all of the details are obvious) provides a much better frame of reference for reading the rest of the prose. Not sure how to square the two opinions here... 🤔

text/0268-acceptance-testing-refactor.md
```diff
@@ -0,0 +1,278 @@
...
+    the applications root element
+  * setup importable `click` helper method
+  * setup importable `tap` helper method
+  * setup importable `triggerKeyEvent` helper method
+  * setup importable `triggerEvent` helper method
+  * setup importable `fillIn` helper method
+  * setup importable `focus` helper method
+  * setup importable `blur` helper method
+  * setup importable `waitUntil` helper method
+  * setup importable `waitFor` helper method
```
cibernox: - `waitFor` doesn't exist in `ember-native-dom-helpers`. So you like to have it added prior to the migration?

text/0268-acceptance-testing-refactor.md
```diff
@@ -0,0 +1,278 @@
...
+  test.
+* Implicit promise chaining will no longer be present. If your test needs to
+  wait for a given promise, it should use `await`.
+* The test helpers that are included by a new default ember-cli app will be no
+  longer needed and will be removed from the new application blueprint. This
+  includes:
+  * `tests/helpers/resolver.js`
+  * `tests/helpers/start-app.js`
+  * `tests/helpers/destroy-app.js`
+  * `tests/helpers/module-for-acceptance.js`
```
kategengler: Does tests/test-helper.js remain?

rwjblue: @kategengler - Yes, though IMHO, it would be better to rename it to `tests/index.js`.  It truly is the entry point to all the tests (and even more obvious now that you have to manually call `start()` to kick off the tests)...

kategengler:
How will an addon be able to provide helpers to both the old and new testing systems? Maybe, for an example, the `withFeature` helper from ember-feature-flags:

```
import Ember from 'ember';

export function withFeature(app, featureName) {
  let featuresService = app.__container__.lookup('service:features');
  featuresService.enable(featureName);
}

Ember.Test.registerHelper('withFeature', withFeature);
```

text/0268-acceptance-testing-refactor.md
cibernox: I see that you haven't included `find` and `findAll` intentionally, with the idea that that users can just "use the platform". However, specially `find` compared to `this.element.querySelector()` is much more terse. Same goes for `findAll` vs `this.element.querySelectorAll()`, with the extra annoyance that `querySelectorAll` returns a `NodeList` instance instead of an Array, which is cumbersome to use to say the list.

I'd consider adding `find` and `findAll`. I believe that people will end up using them anyway.

text/0268-acceptance-testing-refactor.md
```diff
@@ -0,0 +1,278 @@
...
+
+/**
+  Unfocus the specified selector.
+*/
+export function blur(selector: string | HTMLElement): Promise<void>;
+
+/**
+  Returns a promise which resolves when the provided callback returns a truthy value.
+*/
+export function waitUntil(() => boolean): Promise<void>;
```
cibernox: `export function waitUntil(() => boolean, { timeout = 1000 }): Promise<void>;`

text/0268-acceptance-testing-refactor.md
```diff
@@ -0,0 +1,278 @@
...
+
+/**
+  Returns a promise which resolves when the provided selector (and count) becomes present.
+*/
+export function waitFor(selector: string, count?: number): Promise<void>;
```
cibernox: `export function waitFor(selector: string, { count?: number }): Promise<void>;`

rwjblue: Great catch, updated!

text/0268-acceptance-testing-refactor.md
```diff
@@ -0,0 +1,278 @@
...
+  well as importable helpers.
+* `this.owner` will now be present and allow (for the first time 🎉) overriding
+  items in the container/registry.
+* The new system will leverage the `Ember.Application` /
+  `Ember.ApplicationInstance` split so that we can avoid creating an
+  `Ember.Application` instance per-test, and instead leverage the same system
+  that FastBoot itself uses to avoid running initializers for each acceptance
+  test.
+* Implicit promise chaining will no longer be present. If your test needs to
+  wait for a given promise, it should use `await`.
```
kategengler: I think it is worth noting that `await`ed helpers in acceptance tests will wait for all the same things they currently do.

rwjblue:
@kategengler RE: https://github.com/emberjs/rfcs/pull/268#issuecomment-343542474

Assuming the following input:

```js
import Ember from 'ember';

export function withFeature(app, featureName) {
  let featuresService = app.__container__.lookup('service:features');
  featuresService.enable(featureName);
}

Ember.Test.registerHelper('withFeature', withFeature);
```

In order for an addon to support both the existing acceptance testing system, and the new system it could replace that helper with the following:

```js
import { registerAsyncHelper } from '@ember/test';

export function enableFeature(owner, featureName) {
  let featuresService = this.owner.lookup('service:features');
  featuresService.enable(featureName);
}

export default function setupFeatures(hooks, featureName) {
  hooks.beforeEach(function() {
    setupFeature(this.owner, featureName);
  })
}

registerAsyncHelper('withFeature', function(app, featureName) {
  setupFeature(app.__container__, featureName);
});
```

This allows both the prior API (without modification) and the following:

```js
// Option 1:
import { module, test } from 'qunit';
import { setupAcceptanceTest } from 'ember-qunit';
import { setupFeature } from 'addon-name-here/test-support';

module('asdf', function(hooks) {
  setupAcceptanceTest(hooks);
  setupFeature(hooks, 'feature-name-here');

  test(...);
});

// Option 2:
import { module, test } from 'qunit';
import { setupAcceptanceTest } from 'ember-qunit';
import { enableFeature } from 'addon-name-here/test-support';

module('asdf', function(hooks) {
  setupAcceptanceTest(hooks);

  test('awesome test title here', function(assert) {
    enableFeature(this.owner, 'feature-name-here');

    // ...snip...
  });
});
```

(I added ^ to the RFC)

text/0268-acceptance-testing-refactor.md
```diff
@@ -0,0 +1,370 @@
...
+}
+
+export default function setupFeatures(hooks, featureName) {
+  hooks.beforeEach(function() {
+    setupFeature(this.owner, featureName);
+  })
+}
+
+registerAsyncHelper('withFeature', function(app, featureName) {
+  setupFeature(app.__container__, featureName);
```
kategengler: I think this is meant to call `enableFeature`

text/0268-acceptance-testing-refactor.md
```diff
@@ -0,0 +1,370 @@
...
+Ember.Test.registerHelper('withFeature', withFeature);
+```
+
+In order for an addon to support both the existing acceptance testing system, and the new system it could replace that helper with the following:
+
+```js
+import { registerAsyncHelper } from '@ember/test';
+
+export function enableFeature(owner, featureName) {
+  let featuresService = this.owner.lookup('service:features');
```
kategengler: I think this line is meant to use the passed in `owner`

text/0268-acceptance-testing-refactor.md
```diff
@@ -0,0 +1,370 @@
...
+  featuresService.enable(featureName);
+}
+
+export default function setupFeatures(hooks, featureName) {
+  hooks.beforeEach(function() {
+    setupFeature(this.owner, featureName);
```
kategengler: I think this is meant to call `enableFeature`

text/0268-acceptance-testing-refactor.md
rwjblue: DOH!  Fixed (both)!

text/0268-acceptance-testing-refactor.md
```diff
@@ -0,0 +1,370 @@
...
+
+```js
+// Option 1:
+import { module, test } from 'qunit';
+import { setupAcceptanceTest } from 'ember-qunit';
+import { setupFeature } from 'addon-name-here/test-support';
+
+module('asdf', function(hooks) {
+  setupAcceptanceTest(hooks);
+  setupFeature(hooks, 'feature-name-here');
```
kategengler: setupFeatures?

rwjblue: I renamed the example test helper to be singular (because it currently only accepts a single feature name).

rwjblue:
This was discussed at todays Ember core team meeting, and we believe this is ready to move into the final comment period.

For those that haven't reviewed it yet, now's a great time!

text/0268-acceptance-testing-refactor.md
```diff
@@ -0,0 +1,352 @@
...
+
+# Drawbacks
+
+* This is a relatively large set of changes that are arguably not needed (things mostly work today).
+* One of the major hurdles in upgrading larger applications to newer Ember versions, is updating their tests to follow "new" patterns.  This RFC introduces yet another "new" thing (and proposes to deprecate the old thing), and could therefore be considered "just more churn".
+
+# Alternatives
+
+* Do nothing?
+* Make `ember-native-dom-helpers` a default addon (removing the need for DOM interaction helpers proposed here).
```
ro0gr: Looks like `ember-test-helpers` with the `setupTest`, `moduleFor*` and company is already quite complicated.
Keeping DOM helpers implementation as a separate package seems like a win to me from the separation of concerns standpoint.

rwjblue: @ro0gr - Forcing users to be aware of the internal packaging concerns is not ideal. From their perspective, they are using a single set of cohesive tools for testing an Ember app. Whether the implementation actually lives in `ember-test-helpers` or a new addon that is depended on by (and re-exported from) `ember-test-helpers` is not exactly important. 

Also, I disagree that `ember-test-helpers` is "too big".The non-legacy portions are actually quite small at 423 lines of code (even including the legacy code its ~ only 1353 loc).

jamesarosen:
We’ve been using some of these patterns in our projects and we really like them.

The global asynchronous test helpers caused problems because we would inadvertently mix synchronous and asynchronous code, breaking tests. We changed from those to chained helpers, which were safer, but unwieldy. Using `async/await` has been a big win.

We’ve also been switching more of our tests to use ember-native-dom-helpers. These are easier for people who have JS/DOM background, but less Ember experience.

Lastly, we’ve been migrating many of our acceptance tests to integration tests, which the above two techniques make easier. Acceptance tests are slower and more brittle. They’re quite valuable, so we do still use them; I’m just saying that these new patterns allow us to write integration tests with much of the power previously reserved for acceptance tests.

Big +1 to the general approach outlined

text/0268-acceptance-testing-refactor.md
```diff
@@ -0,0 +1,352 @@
...
+
+/**
+  Unfocus the specified selector.
+*/
+export function blur(selector: string | HTMLElement): Promise<void>;
+
+/**
+  Returns a promise which resolves when the provided callback returns a truthy value.
+*/
+export function waitUntil(() => boolean, { timeout = 1000 }): Promise<void>;
```
cibernox: I realied now that, in its current implementation, `waitUntil` (and `waitFor`) do not return `Promise<void>`. The promise that returns from `waitUntil` resolves with the resolve value of the callback. The `waitFor` helper does the same, meaning that it will return a promise that resolves to one (or many) DOM elements.

I did this as a convenience for developers wanting to do assertions on the element they've been waiting for.
Example:

```js
let loadingEl = await waitFor('.loading-msg');
assert.equal(loadingEl.textContent.trim(), 'Please, wait...');
```

I am not opposed to change it, but if you think it can be handy, please, update the signature.

rwjblue: 👍 updated

rwjblue:
Made a few tweaks since my last comment:

* Updated typings for `waitFor` and `waitUntil` (based on @cibernox's feedback)
* Renamed `setupAcceptanceTest` to `setupApplicationTest`. The primary motivation here is that the method name now describes _what_ it does **not** why it is used. For example, `setupRenderingTest` is what we invoke when we will be rendering a template snippet.

raido:
This is awesome. 🎉

text/0268-acceptance-testing-refactor.md
0xadada: Many folks use `module-for-acceptance` to start/stop Mirage servers. So the new blueprint wont include that module helper, but nothing is stopping users from keeping or adding it back, correct?

Turbo87: with the new system you would call something like `setupMirage(hooks)` and it would handle starting and stopping of the server in the right hooks automatically for you

rwjblue:
This was discussed at the Ember core team meeting today, and given that no objections / suggestions / tweaks have come up during the final comment period, and it seems that most folks massively in favor of this improvement we are going to land....

FWIW, I have begun implementation and hope to be able to use this in apps soon...

PR #275, opened 2017-12-10
RFC: Yield link-to component state

sukima:
[Rendered](https://github.com/sukima/rfcs/blob/link-to-yields/text/0000-link-to-yields.md)

### DIFF
--- a/text/0000-link-to-yields.md
+++ b/text/0000-link-to-yields.md
@@ -1,0 +1,127 @@
+- Start Date: 2017-12-10
+- RFC PR: 
+- Ember Issue: 
+
+# Summary
+
+A proposal to allow the `link-to` helper to expose the active state via
+`yield`. Doing so can allow more flexable use cases and DOM layouts.
+
+# Motivation
+
+link-to will add an `active` class to its element when the route matches.
+However, many CSS frameworks or addons require a more specific layout where the
+inner child needs the `active` class but the `onclick` event needs to be on the
+parent link-to element. For example Bootstrap CSS requires the child element to
+have the `active` state to render correctly but the parent element mush have the
+onclick event handler otherwise the target area when rendered does not cover the
+entire UI (due to margins/padding).
+
+The link-to also knows how to generate a URL (href) based on the route name.
+Doing this conversion manually each tim emeans the unsertanding of route name to
+URL will be in multiple places in the app when instead could be calculated if
+exposed by the link-to helper.
+
+# Detailed design
+
+The syntax for link-to follows the convention of yielding to the consumer. It is
+expected that clicking the link-to element will perform the route transition and
+when the target route matches the current route the same element recieves the
+`active` class. The content of the element rendered to the DOM is often provided
+by the consumer as yielded content.
+
+```hbs
+{{#link-to "my-route" tagName="li"}}
+  <a href="/hard/coded">Yielded Content</a>
+{{/link-to}}
+```
+
+Which renders as:
+
+```html
+<li class="active">
+  <a href="/hard/coded">Yielded Content</a>
+</li>
+```
+
+By yielding the active and href state we can leverage the link-to logic in the
+context of our own code:
+
+```hbs
+{{! Example 1 }}
+{{#link-to "my-route" tagName="li" as |link|}}
+  <a href={{link.href}} class="{{if link.active 'active'}}">My Route</a>
+{{/link-to}}
+
+{{! Example 2 }}
+{{#link-to "my-route" tagName="li" as |link|}}
+  {{#if link.active}}
+    <span class="current-route red">Current Route</span>
+  {{else}}
+    <a href={{link.href}}>My Route</a>
+  {{/if}}
+{{/link-to}}
+```
+
+# How We Teach This
+
+The documentation for link-to will change the yield example to include a yielded
+variable called `link` which will have two properties `link.active` and
+`link.href`. A paragraph to explain that meaning of them and an example (see
+above) to illistrate a possible use case.
+
+Since this feature adds functionality to an existing API the current guides on
+the link-to helper are enough as they are since it covers most use cases. When a
+user needs more information (such as looking up `tagName` and `class`) via the
+API section they can also discover the use of this feature.
+
+This change would **not** modify how Ember is tought. It is backwards
+compatible with current practises and typical usage out in the wild. The added
+functionality need only documented once in the API for link-to.
+
+Most existing Ember users are familure with the use of yielded variables and
+adding this would not need much education beyond just letting them know it is
+available via a ChangeLog and in the API documnetaion for the link-to component.
+
+# Drawbacks
+
+A tradeoff is that the two exposed values now become public. Which means it does
+add the slight overhead to maintaining this part of the code since once made
+public it will require more scrutiny before changing.
+
+# Alternatives
+
+This section could also include prior art, that is, how other frameworks in the
+same domain have solved this problem.
+
+Per this [Stack Overflow question][1] you can nest the link-to twice to cover
+teh specific Bootstrap problem. Unfortunatly, it does not offer the flexability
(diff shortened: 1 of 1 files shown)


sukima:
[Related Implementation](https://github.com/emberjs/ember.js/pull/15835#issuecomment-347357691)

jamesarosen:
I strongly support the _overall concept_, but "active" already has [another meaning](https://developer.mozilla.org/en-US/docs/Web/CSS/:active) for `<a>` tags. I might recommend `link.current` or `link.isCurrent` to map to `current-when`.

sandstrom:
Very positive to this RFC! Seems like it should remove some hacks we're currently using to solve this issue.

I've read the RFC and from what I can tell your suggested solution should be able to supersede our hack (below). Do you agree?

**templates/something.hbs**
```hbs
{{#active-link-wrapper class='large' as |alw|}}
  {{#link-to 'myRoute' wrapper=alw}}my link text{{/link-to}}
{{/active-link-wrapper}}
```


**components/active-link-wrapper.js**
```js
import Ember from 'ember';
import { retrieveEventPath } from 'client/utils/dom';

  // Used to enlarge the click-area of links (useful for touch-devices),
  // and allows the 'active' class to exist on a parent element.
  //
  // NOTE
  // Related code is in 'reopens/link-component'.

  tagName: 'span',

  classNameBindings: ['active'],
  classNames: ['active-link-wrapper'],

  active: function() {
    return this.get('_linksList').isAny('active');
  }.property('_linksList.@each.active'),

  init() {
    this._super();
    this.set('_linksList', Ember.A());
  },

  // Act only if the target is this wrapper/element or a non-link element, do nothing when targeting the link directly.
  // Otherwise we're creating an additional synthetic click.
  click(jqEvent) {
    let event = jqEvent.originalEvent;

    // slice out elements between event target and this component element, and see if any of them is a link
    let eventPath = Array.from(retrieveEventPath(event)); // cast NodeList to Array
    let linkInPath = eventPath.slice(0, eventPath.indexOf(this.get('element'))).some((e) => e.matches('a'));

    if ((event.target === this.get('element')) || !linkInPath) {
      let l = this.get('_linksList')[0];

      if (l) {
        l._invoke(event);
      }
    }
  },

  // Scheduling needed to avoid `You modified ${label} twice in a single render` deprecation message.
  // This stuff is hacky to begin with, and will be replaced when a better option is available.
  registerLink(link) {
    Ember.run.scheduleOnce('afterRender', this, function() {
      this.get('_linksList').addObject(link);
    });
  },

  deregisterLink(link) {
    Ember.run.scheduleOnce('afterRender', this, function() {
      this.get('_linksList').removeObject(link);
    });
  },

});
```

**link-component-reopen.js**
```js
Ember.LinkComponent.reopen({

    // Communication with ActiveLinkWrapper
    //
    // Need to know of the child link that it wraps.
    // Using a computed property setter trick to register each link with the wrapper.
    // (there is no easy way to 'subclass' the link-helper, if one existed this could be less hacky)
    //
    // https://github.com/emberjs/ember.js/issues/10573
    // https://github.com/emberjs/ember.js/blob/680f997e/packages/ember-routing-htmlbars/lib/keywords/link-to.js

    _activeLinkWrapper: null,

    wrapper: Ember.computed({
      get(_key) {
        return this.get('_activeLinkWrapper');
      },
      set(key, value) {
        if (value) {
          value.registerLink(this); // value is an instance of `ActiveLinkWrapper`
          this.set('_activeLinkWrapper', value);
          return value;
        }
      },
    }),

    willDestroyElement() {
      this._super();

      let alw = this.get('_activeLinkWrapper');
      if (alw) {
        alw.deregisterLink(this);
      }
    },

  });
```

text/0000-link-to-yields.md
```diff
@@ -0,0 +1,126 @@
...
+`link.href`. A paragraph to explain that meaning of them and an example (see
+above) to illustrate a possible use case.
+
+Since this feature adds functionality to an existing API the current guides on
+the link-to helper are enough as they are since it covers most use cases. When a
+user needs more information (such as looking up `tagName` and `class`) via the
+API section they can also discover the use of this feature.
+
+This change would **not** modify how Ember is taught. It is backwards-compatible with current practices and typical usage out in the wild. The added
+functionality need only documented once in the API for link-to.
```
locks: We have a [Guide dedicated to `link-to`](https://guides.emberjs.com/v2.17.0/templates/links/), where it would likely be warranted for the API in this RFC to be covered.

wycats:
@sukima are you planning to update the RFC to replace the boilerplate from the template with some content for those sections?

My personal feedback is:

- Instead of overloading `{{#link-to`, we should pick a new name.
- The RFC should describe what would be needed to implement today's `link-to` functionality in terms of this new helper.
- In particular, does `<a` "just work" or is additional work needed to ensure that interacting with the link is correctly intercepted.

wycats:
@MelSumner What kind of accessibility considerations are relevant for JavaScript-enhanced links?

Panman8201:
> Instead of overloading {{#link-to, we should pick a new name.

@wycats but wouldn't that cause even more API surface by duplicating everything `{{link-to}}` just to also allow the yielded state? Or are you thinking a completely different API for this, perhaps in transition to something new? Personally, I'd rather see this added directly to `{{link-to}}`. It "just makes sense" and adds upon what was enabled from the [Router Service RFC](https://github.com/emberjs/rfcs/blob/master/text/0095-router-service.md).

@jamesarosen [`isActive` is already used in the Router Service](https://emberjs.com/api/ember/2.18/classes/RouterService/methods/isActive?anchor=isActive), so IMO this should stick with that name for consistency.

@sandstrom Your solution has also been available as an ember addon, [ember-cli-active-link-wrapper](https://github.com/alexspeller/ember-cli-active-link-wrapper). But I don't think this replaces that solution because `{{#active-link}}` doesn't create a clickable link itself, it just looks for _nested_ active links to apply the active class to the parent.

sandstrom:
@Panman8201 True, but if this RFC were to be implemented, we could something like below, and we'd get the class on the `li`. The only difference is that in this case, the `li` would also be clickable (but that wouldn't be an issue for us).

```hbs
{{#link-to "my-route" tagName="li" as |link|}}
  <a href={{link.href}}>My Route</a>
{{/link-to}}
```

Panman8201:
@sandstrom Ah, yeah, that would be a good fit. Guess I was thinking in the case of navbar dropdowns, where one of the many nested links could be active and the parent/container also needs the active state.

wycats:
@Panman8201 the issue with tacking more things onto `{{link-to}}` is that `{{link-to}}` as it currently stands has no choice but to opt you into a bunch of mandatory behavior (attributes that have to be computed and applied to the tag unconditionally in case CSS is using them, etc.) that make things slow.

Your API design doesn't have that problem, because it yields information into the block that you are supposed to use on your own, but it looks darn similar to the existing API, completely altering the behavior purely based on the block's arity. Importantly, the improvements to the API surface in your proposal also mean that the implementation shouldn't be shared.

In general, we haven't swapped semantics and implementation in Glimmer templates based purely on the arity of the block, and I don't really want to start now 😄

knownasilya:
This could be done in addon space first, using `RouterService#urlFor`.

```hbs
{{#url-for 'my.route' some.id as |url isActive|}}
  <a href={{url}} class={{if isActive 'active'}}>{{name}}</a>
{{/url-for}}
```

sukima:
Based on @knownasilya's [suggestion](#issuecomment-355761623) should we consider this RFC closed (unmerged) as the new API/implementation is best suited for an addon?

wycats:
@sukima I think it would be awesome for this to start out as an addon, but once it gets going, let's revisit this RFC. I really do think it makes sense for us to reform `link-to`, but the point of the Routing Service RFC was to allow people to try things out as addons first, and I think that's right here.

PR #278, opened 2017-12-11
Template-only Components

chancancode:
[Rendered](https://github.com/emberjs/rfcs/blob/template-only-components/0278-template-only-components.md)

### DIFF
--- a/0278-template-only-components.md
+++ b/0278-template-only-components.md
@@ -1,0 +1,249 @@
+- Start Date: 2017-12-11
+- RFC PR: https://github.com/emberjs/rfcs/pull/278
+- Ember Issue: (leave this empty)
+
+# Summary
+
+Introduce a low-level "flag" to remove the automatic wrapper `<div>` for
+template-only components (templates in the `components` folder that do not
+have a corresponding `.js` file).
+
+In other words, given there is NO `app/components/hello-world.js` and there
+exists `app/templates/components/hello-world.hbs` which contains the
+following markup:
+
+```hbs
+Hello world!
+```
+
+When this template-only component is invoked as `{{hello-world}}` with the
+flag unset or disabled (i.e. today's semantics), Ember will render:
+
+```html
+<div id="ember123" class="ember-view">Hello world!</div>
+```
+
+When the flag is enabled, the same invocation will render:
+
+```html
+Hello world!
+```
+
+# Motivation
+
+With today's component system (i.e. `Ember.Component`), a wrapper element (a
+`div` by default, along with an ID like `ember123` and the `ember-view` class)
+is automatically added for every component.
+
+Customizing this wrapper element (such as changing the tag name – or removing
+it altogether) requires making changes to the component's JavaScript class,
+such as:
+
+```js
+import Component from "@ember/component";
+
+export Component.extend({
+  tagName: "footer",
+  classNames: ["legalese"]
+});
+```
+
+While we acknowledge this API is quite cumbersome, it is sufficient to "get
+things done" for regular components, and Glimmer Components will address
+the usability aspect once they land.
+
+However, this API does not work for template-only components, as they do
+not have a component JavaScript class by definition. Therefore, in practice,
+template-only components always come with a `<div>` wrapper, along with the
+default `id` and `class` attributes, with no obvious ways to customize it.
+
+This is quite problematic, as it is often desirable to use a template-only
+component to organize content that requires a certain markup structure. The
+most common workaround for this problem is to use a partial instead, which
+comes with [a host of issues](https://github.com/emberjs/rfcs/pull/262). I
+will discuss other workarounds in the section below.
+
+This RFC proposes to add a global flag to remove this wrapper element around
+template-only components. This will allow the component author to specify the
+wrapper element in the component template, offering direct control over the
+tag name and other attributes. It would also allow the component to have more
+than one top-level element, or none at all.
+
+In other words, this flag changes template-only components in the app to have
+"Outer HTML" semantics. _What you type is what you get._
+
+Notably, [Glimmer Components](https://glimmerjs.com/guides/templates-and-helpers)
+have adopted the "Outer HTML" semantics long ago and the experience has been
+very positive. This would be one of the first pieces of the Glimmer.js experiment
+to make its way into Ember. We think this feature is small, self-contained but
+useful enough to be integrated back into Ember at this point.
+
+If accepted, this RFC will fully subsume the [Non-context-shifting partials](https://github.com/emberjs/rfcs/pull/262)
+RFC. We can therefore (at a later time, in a separate RFC) explore deprecating
+partials in favor of wrapper-free template-only components.
+
+# Detailed design
+
+## API Surface
+
+We should not expose the flag directly as a public API. Instead, we should
+abstract the flag with a "privileged addon" whose only purpose is to enable
+the flag. Applications will enable the flag by installing this addon. This
+will allow for more flexibility in changing the flag's implementation (the
+location, naming, value, or even the existence of it) in the future. From the
+user's perspective, it is the _addon_ that provides thisfunctionality. The
+flag is simply an internal implementation detail.
+
+We have done this before in other cases (such as the legacy view addon during
+the 2.0 transition period), and it has generally worked well.
+
+When landing this feature, it will be entirely opt-in for existing apps, but
+the Ember CLI application blueprint should be updated to include the addon by
+default. At a later time, we should provide another addon that _disables_ the
+flag explicitly (installing both addons would be an install-time error). At
+that time, we will issue a deprecation warning if the flag is *not set*, with
+a message that directs the user to install one of the two addons.
+
+## Single Global "Flag"
+
+The proposed flag will be truly global in scope. That is, setting this flag
+will change the semantics of all template-only components in the entire app,
+even for components that were included by addons.
+
+However, we believe this would not affect any addon components in practice,
+as the predominant pattern for addons to expose components currently
+necessitates a JavaScript class. Addon authors would create the component
+(with or without a JavaScript class) in the `/addon` folder, but exposing
+it for consumption in apps requires creating a corresponding JavaScript class
+in the `/app` folder to "re-export" the component. Therefore, in practice,
+it is not actually possible for addons to have a truely template-only
+component today (something to address in a future RFC).
+
+## Leakage Of `Ember.Component` Semantics
+
+While the primary purpose of this flag is to remove the wrapper element from
+template-only components, there are a few other observable semantics changes
+that comes with it as well.
+
+Currently, template-only components are "backed" by an instance of `Ember.Component`.
+That is, Ember will create an instance of `Ember.Component` and set it as the
+`{{this}}` context for the template.
+
+With the flag enabled, there will be *no* component instance for the template
+and `{{this}}` will be set to `undefined` (or `null`, perhaps). This would
+improve performance for template-only components significantly.
+
+Since there is no JavaScript file for the component, this is only observable
+in a few limited ways:
+
+1. The most noticable artifact is the component's arguments will not be
+   auto-reflected on the component instance (as there is no component
+   instance at all). Therefore, the only way to access the component's
+   arguments is to use the `{{@foo}}` syntax proposed in [RFC #276](https://github.com/emberjs/rfcs/pull/276).
+
+2. Because of the named arguments auto-reflection, it is actually possible
+   to configure the `tagName` and classes on the "hidden" component
+   instance on the invocation (e.g. `{{foo-bar tagName="footer" class="legalese"}}`).
+   This will obviously stop working, but it is also not necessary anymore
+   as the component author can simply include the tag in the template.
+   Alternatively, the component author can choose to leave out the tag
+   and let the caller wrap it in their template.
+
+3. It is possible (but very rare) to configure global injections on the
+   component type. Since no component is being instantiated here, those
+   properties will not be accessible in the template. 
+   
+   More broadly, `{{this.foo}}` or the shorthand `{{foo}}` (where it
+   would have resolved into a `this` lookup) will always be `undefined`
+   (or `null`, perhaps).
+
+## Migration Path
+
+Given the subtle semantics differences enumerated above, it is therefore
+not necessarily safe to simply turn on the flag in bigger applications
+as it is quite likely that some of the template-only components might be
+relying on one or more of these features. Further, removing the wrapper
+element might break the layout.
+
+Therefore, the only safe, mechanical transformation is to generate a
+JavaScript file for each template-only component (turning them into non-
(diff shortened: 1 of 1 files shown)


0278-template-only-components.md
```diff
@@ -0,0 +1,249 @@
...
+`{{this}}` context for the template.
+
+With the flag enabled, there will be *no* component instance for the template
+and `{{this}}` will be set to `undefined` (or `null`, perhaps). This would
+improve performance for template-only components significantly.
+
+Since there is no JavaScript file for the component, this is only observable
+in a few limited ways:
+
+1. The most noticable artifact is the component's arguments will not be
```
kategengler: How will / should the Ember Inspector behave with template-only components?

cc @teddyzeenny

0278-template-only-components.md
```diff
@@ -0,0 +1,249 @@
...
+
+We have done this before in other cases (such as the legacy view addon during
+the 2.0 transition period), and it has generally worked well.
+
+When landing this feature, it will be entirely opt-in for existing apps, but
+the Ember CLI application blueprint should be updated to include the addon by
+default. At a later time, we should provide another addon that _disables_ the
+flag explicitly (installing both addons would be an install-time error). At
+that time, we will issue a deprecation warning if the flag is *not set*, with
+a message that directs the user to install one of the two addons.
```
Turbo87: IMHO the end goal should be that the default setting for the flag is "enabled" (aka. no wrapper), and that we don't need to include the enable-addon anymore. Is that the plan?

cibernox:
I have two questions about this, on two areas that might cause breakages not listed:

1. Right now, all applications have an implicit wrapper, that is sometimes targeted in CSS as `body > .ember-view`. Will the implicit top-level outlet behave like a template-only component? If not, why?
2. A similar issue happens in testing, where there is a `#ember-testin > .ember-view` wrapper around everything. Same questions: Will that go away? Why/why now?

0278-template-only-components.md
```diff
@@ -0,0 +1,249 @@
...
+even for components that were included by addons.
+
+However, we believe this would not affect any addon components in practice,
+as the predominant pattern for addons to expose components currently
+necessitates a JavaScript class. Addon authors would create the component
+(with or without a JavaScript class) in the `/addon` folder, but exposing
+it for consumption in apps requires creating a corresponding JavaScript class
+in the `/app` folder to "re-export" the component. Therefore, in practice,
+it is not actually possible for addons to have a truely template-only
+component today (something to address in a future RFC).
```
Turbo87: it might still make sense to allow white/blacklisting to ease the transition from with-wrapper to no-wrapper in existing large-scale apps 🤔

Turbo87: I just saw that a workaround for this is proposed in the "Migration" section below, which seems fine to me 👍

0278-template-only-components.md
```diff
@@ -0,0 +1,249 @@
...
+# Detailed design
+
+## API Surface
+
+We should not expose the flag directly as a public API. Instead, we should
+abstract the flag with a "privileged addon" whose only purpose is to enable
+the flag. Applications will enable the flag by installing this addon. This
+will allow for more flexibility in changing the flag's implementation (the
+location, naming, value, or even the existence of it) in the future. From the
+user's perspective, it is the _addon_ that provides thisfunctionality. The
```
Turbo87: typo: `thisfunctionality`

0278-template-only-components.md
chancancode: The more general question is how should Ember Inspector work with Glimmer Components and other type of custom components.

I think there are two kinds of components – those that should and should not be shown in the inspector's component tree. For example, `{{render}}`, `{{mount}}`, `{{partial}}` and `{{outlet}}` are all technically implemented as components in the system, but are invisible to the inspector. You could imagine maybe one day things like `{{#if}}` could be implemented as a component as well (though unlikely), but should be invisible to the inspector as well. (These are "virtual" views in old-school Ember.)

(A related question, whether a component should be "visible" in the `Ember.Component` `parentView` (:disappointed:) hierechary.)

I think we need to decide:

1. Should template-only components be visible?
2. Separately, what are the APIs needed to expose a component to the inspector in general.

I think my default position will be that let's make template-only components (and possibly all other custom components) invisible to start while we work out the second question.

0278-template-only-components.md
chancancode: Yes. I think that should be the case in Ember 4.0.

chancancode:
@cibernox this RFC does not change that (would you like to pick that up in another RFC? I think that is a good idea, just not particularly related to this)

0278-template-only-components.md
teddyzeenny: The inspector currently displays both outlets and components. It builds the view tree by [merging two trees](https://github.com/emberjs/ember-inspector/blob/1c34df9e8e7f521b380d09d6f7d3af15c650cc78/ember_debug/libs/glimmer-tree.js#L116-L122) together: the outlet tree, and the component tree:
1) Builds the outlet tree by walking the `outletState`
2) Builds the component trees by starting with the top component and walking the component tree by getting the child views of each component instance (which are stored as an array on their respective parent instance).
3) Attaches the component sub-trees to their respective parent outlet by matching the controllers of both.
4) The final tree containing the outlets and components is displayed in the View Tree.

With the current approach, template-only components will not be visible as they're neither outlets nor component instances. While certainly not a blocker for the feature, my opinion is that eventually these components should be visible in the inspector as I would expect to see them as a user, same reason why we display outlets which are not backed by component instances. If we can find a way to build template-only component trees and match them to their respective parent outlet/component instance (just like we matched component trees with their parent outlet), then this would work. If not, then I think we would need to change our approach of how we build the view tree.

kategengler: I also think that the template-only components should be visible in the inspector. Until that is done, though, what would happen in the case of a template-only component that invokes a non-template only component? Would the tree of components just end before the template-only component or would it skip over the template-only component?

For clarification:
In, say, application.hbs:
```
{{template-only-component}}
```
In template-only-component.hbs:
```
{{not-template-only-component}}
```

teddyzeenny: @kategengler I think that would depend on the answer to @chancancode's question of "whether a component should be "visible" in the Ember.Component parentView hierechary"? My guess is it would be invisible, similar to "virtual" views, which means the view tree would just skip over the template-only component and display the instance-backed child.

chancancode: Yes. I think for now, they would be invisible by default. I _think_ we'll most likely want to keep the `parentView` tree as-is. In general I believe we want to discourage that style of programming model and use other kind of compositions, like yielding, etc.

It will also likely be a breaking change to expose anything that is _not_ `Ember.Component` as their `parentView`. For example, any code that is walking that tree with a while loop (keep getting the `parentView`) would break, as those custom component might not have the same API. (Or in this case, there is literally no component instance.) Finally, the name (parent**View**) is a misnomer at this point anyway. I think we should probably just deprecate those properties on `Ember.Component`.

Debugging (Ember Inspector) is a different story. We can provide different util methods to walk the tree intended for debugging purpose. Ideally, we should get Ember Inspector to somehow register its presence somehow before the app boots so that we would only need to track that when it is enabled.

Anyhow, that should be discussed separately IMO. Perhaps the [Custom Manager RFC](https://github.com/emberjs/rfcs/pull/213) would be the right place for it, or it could even be further spun-off into another RFC (I would prefer that if we can).

In the short term, I think the status quo would be that these new template-only components will be invisible to the view tree, which is not amazing but not fatal IMO. This is how partials work today (which is probably also not amazing), and the use cases this targets is quite similar.

sdhull:
Possibly a dumb question:
Can you already achieve this by doing `{{template-only-component tagName=''}}`?

chancancode:
I implemented the feature (behind a flag). You can try this on canary now: https://ember-twiddle.com/a2013417648c4dced26ae78e8aaa5e6a?openFiles=templates.application.hbs%2C

Reminder: since the new semantics does not auto-reflect arguments as properties, you will have to use the named arguments syntax in #276 if you want to access them.

0278-template-only-components.md
kategengler: I think invisible (ala partials) is fine, but any components in the tree below the template-only component(s) need to be visible or it would severely hinder the ability to debug.

chancancode: @kategengler correct, that is how it should work. Essentially, both the `parentView` tree and the inspector (they are more or less the same thing today) will skip over anything that is not an `Ember.Component`. Since [my patch have landed in canary](https://github.com/emberjs/ember.js/pull/15974), we should be able to boot an app and see if this is breaking anything in the inspector now.

rtablada:
@chancancode Looking at this PR I think there's some confusion in what are called "Template Only Components"

Based on recent meetup and conference talks the idea of a Template Only component seems to be understood as a component that can be statically optimized and inlined where used or used as a single lookup without the overhead of a stateful component (similar to ember-ast-helpers or the current Glimmer implementation).

While this RFC definitely removes the blocker of the surrounding div, I think that naming the removal of this div as "Template Only Components" will lead to confusion.

rtablada:
As a suggestion I think "Template Only Component HTML".

0278-template-only-components.md
kategengler: I did just that before adding my comment earlier today: In Ember Inspector, partials are just invisible, but unfortunately, template-only components cut off the tree. Since the implementation is already in canary, should I go ahead and file an issue on the inspector for this?

chancancode: @kategengler yes please! Thank you for checking, I'll look into it!

chancancode: You can also file that bug against Ember, as I suspect it's a problem on my end. Either way, as long as you mention or assign it to me I can take a look.

chancancode:
@rtablada it actually does pave the way for all of that.

[Here is the code](https://github.com/emberjs/ember.js/blob/master/packages/ember-glimmer/lib/component-managers/curly.ts) to support `Ember.Component` (actually, that doesn't even include the `Ember.Component` class itself and all its mixins!) and [here is all the code](https://github.com/emberjs/ember.js/blob/master/packages/ember-glimmer/lib/component-managers/template-only.ts) needed to support the new style template-only component in this RFC.

Even if you are not familiar with the internals, you can probably intuit that the latter is doing a lot less work. More importantly, since it does not have an instance associated with it (`create`, `getSelf` and `getDestructor` all returns `null`), the VM will be able to take advantage of that fact and do some of the optimizations you mentioned.

However, optimizability is not really a user visible change. From the user's perspective, this is mostly just a developer ergonomic improvement (which IMO stands on its own and then some). It just happens to also be a trojan horse for landing Glimmer Components and performance improvements as well.

bartocc:
I like the idea of "Outer HTML" very much, but at the moment, 2 things bother me with this RFC:

- I believe the [ember-component-css](https://github.com/ebryn/ember-component-css) addon would definitely break. If I am correct, the addon adds `componentCssClassName/styleNamespace` on every Ember.Component instance and then uses this property to render the CSS from `styles.less`. With no Ember.Component instance for template only components, ember-component-css would become unusable, and not being able to style components with this addon is a drawback that outweighs the benefit of "Outer HTML" in my opinion. Adding an empty .js file would be a workaround though, but quite cumbersome.

- We would have different behaviours for components with a .js file and components without it. This could be confusing developers. I my mental model, all components should behave the same. Should we consider using a different folder for template only components to make things clearer ?

joukevandermaas:
I agree with @bartocc; adding a javascript file shouldn't change what is being rendered. I very often start out with just a template and then add a javascript file later when I need a computed, for example.

It would be very nice to have a way to preserve the 'outer html' behavior for components with a javascript file to support this 'upgrading' scenario.

cibernox:
@joukevandermaas you can. Add `tagName: ''` to your component's JS file.

wycats:
@cibernox exactly! We already have the `outerHTML` behavior via `tagName: ''`, so a good practice for adding a JS file later is to add `tagName: ''` to the JS component once you create it.

Glimmer components have a generally better story here for avoiding the need for a `this.element` in general (through adding event handlers to the top-level element in the component).

My general hope that we can move through these RFCs quickly enough so that the correct transition here will be from "template-only components" to "new-style JS file", and I think everyone is motivated to avoid letting the current status sit for long enough for these problems to become a very serious hassle.

bartocc:
Great! I did not realize that adding `tagName: ''` was equivalent to `outerHTML`
🎉 

This gives us a pretty good transition path I believe

webark:
@bartocc things like ECC would adjust their architecture and ergonomics to account for this welcome change!

chancancode:

1. How does this affect Ember Inspector?

    For now, template-only components will be invisible to Ember Inspector. We will collaborate with @teddyzeenny to propose new APIs that would allow Ember Inspector to track the components tree more accurately.

2. It is strange that adding a JS file changes the component's behavior.

    This is indeed a change in the model. In the future with the [Custom Components API](https://github.com/emberjs/rfcs/pull/213), `Ember.Component` will no longer be the only kind of components that exists in the ecosystem. But choosing to use a certain type of component in the JS file, you are also opt-ing into the special features/semantics that comes with that breed of components. For `Ember.Component`, this means having a wrapper element that you can customize in the JS class, for Glimmer Components, that might mean a different API/feature-set, and so on.

    Over time, with the addition of Glimmer Components and other kinds of community components become more popular, we think this will feel very natural.

Meanwhile, since the feature have landed in canary behind a flag, I encourage you to give it a spin and get a feel for it, either using this [Ember Twiddle](https://ember-twiddle.com/a2013417648c4dced26ae78e8aaa5e6a?openFiles=templates.application.hbs%2C) or by using a canary build with the `ember-glimmer-named-arguments` and `ember-glimmer-template-only-components` feature flag set to `true`.

jamesarosen:
Will this break [ember-feature-flags](https://github.com/kategengler/ember-feature-flags) for template-only components? That addon [injects](https://github.com/kategengler/ember-feature-flags/blob/18a75419bbe7961b67b7f8893ae397ca715dca14/app/initializers/ember-feature-flags.js#L12) `service:features` into every component.

webark:
i assume one of these template files has to contain valid html.. It can’t be like a series of attributes and you could include in another tag, right?
It could however just include text i assume, and no markup.

chancancode:
@webark correct, it has to be valid HTML content (which, by extension, includes "just text")

chancancode:
@jamesarosen Correct. Since there are no instance in this case, `{{this.features}}` will be undefined. It is unclear that injection will/should work for components other than `Ember.Component` (my guess is no), but that is probably a question for #213.

I think your first suggestion is spot on.

chancancode:

akatov:
The link in the description is wrong! Should be https://github.com/emberjs/rfcs/blob/master/text/0278-template-only-components.md

wycats:
@akatov looks like it was fixed! Thanks!

PR #280, opened 2017-12-11
Remove Application wrapper div

cibernox:
[Rendered](https://github.com/cibernox/rfcs/blob/remove-wrapper-divs/text/0000-remove-application-wrapper.md)

### DIFF
--- a/text/0000-remove-application-wrapper.md
+++ b/text/0000-remove-application-wrapper.md
@@ -35,8 +35,8 @@ Similarly, in testing there is another `.ember-view` wrapper inside the
 
 This RFC proposes to add a global flag to remove those wrapper elements,
 effectively making the `application.hbs` template have "Outer HTML" semantics, which aligns
-well with [the changes recently proposed](#278) for template-only components,
-as well as the way Glimmer apps work.
+well with [the changes recently proposed](https://github.com/emberjs/rfcs/pull/278)
+for template-only components, as well as the way Glimmer apps work.
 
 The same flag will also remove the unnecessary extra wrapper inside the testing
 container.
--- a/text/0000-remove-application-wrapper.md
+++ b/text/0000-remove-application-wrapper.md
@@ -89,7 +89,7 @@ with regular handlebars.
 
 There is a possibility that removing the wrapper can break styles for some apps,
 but since adding the wrapper back is just of editing the `application.hbs` template,
-that is probably a despicable drawback.
+that is probably a minor drawback.
 
 There is also a non-zero chance that some testing addon is relying on the `#ember-testing > .ember-view`
 html hierarchy for some reason, and those addons would have to be updated.
--- a/text/0000-remove-application-wrapper.md
+++ b/text/0000-remove-application-wrapper.md
@@ -1,6 +1,6 @@
 - Start Date: 2017-12-11
 - RFC PR: https://github.com/emberjs/rfcs/pull/280
-- Ember Issue: (leave this empty)
+- Ember Issue: https://github.com/emberjs/ember.js/pull/15981
 
 # Summary
 
--- a/text/0000-remove-application-wrapper.md
+++ b/text/0000-remove-application-wrapper.md
@@ -9,7 +9,7 @@ Ember apps and tests.
 
 # Motivation
 
-In Ember applications today, applications are anchored to some existing html
+In Ember applications today, applications are anchored to some existing HTML
 element in the page. Usually, this element is the `<body>` of the document, but it
 can be configured to be a different one when the application is defined,
 passing a CSS selector to the `rootElement` property:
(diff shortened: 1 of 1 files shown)


text/0000-remove-application-wrapper.md
```diff
@@ -0,0 +1,99 @@
...
+the implicit application wrapper of the app (to sadly discover that is not
+even possible), but instead they will just add a wrapper only if they want,
+and in the same way they would add a wrapper in any other point of their application,
+with regular handlebars.
+
+# Drawbacks
+
+There is a possibility that removing the wrapper can break styles for some apps,
+but since adding the wrapper back is just of editing the `application.hbs` template,
+that is probably a despicable drawback.
```
Gaurav0: It's despicable! (Daffy duck voice)

debatable?

cibernox: I was using it as a synonym of negligible.

topaxi:
Yess! The current behavior gave me headaches in various styling scenarios.

text/0000-remove-application-wrapper.md
```diff
@@ -0,0 +1,99 @@
...
+
+There is also a non-zero chance that some testing addon is relying on the `#ember-testing > .ember-view`
+html hierarchy for some reason, and those addons would have to be updated.
```
Turbo87: FWIW qunit-dom supports both cases already via fallback if no ember-view child is found

text/0000-remove-application-wrapper.md
GavinJoyce: **_minor drawback_** perhaps?

sandstrom:
The RFC says *"This addon will be opt-in"*.

Does that mean that there will be an ember-cli addon that will toggle this? I think that would be bad, better to just make it an ember setting/flag. Then make it standard (and remove setting/flag) in the next major release [breaking change].

cibernox:
@sandstrom The addon will only enable a "secret" flag in Ember. It will, as some point, be enabled by default and another addon will exist to disable it. I don't strongly oppose to just make it a flag passed to `new EmberApp`, but this is also the approach proposed by @chancancode for #278 , so whatever approach we choose, it should be consistent.

sandstrom:
I don't have a strong opinion on this & I'm overall in favor of this RFC! 🏅 

But sometimes I think ember is making addons out of stuff that should be settings. Some examples from [ember-cli](https://github.com/ember-cli/ember-cli/blob/master/blueprints/app/files/package.json):

- ember-cli-app-version
- ember-export-application-global
- ember-welcome-page

Gaurav0:
@sandstrom 

This is really due to some of the philosophy behind ember, the idea of having zero configuration and strong conventions around opinions.

This way the default way of doing things is presented that can change with time in a manner that eases migration.

When there is a setting and a default value, people are conditioned to consider the setting as something they can configure for their application. They tend to look at all the settings and feel free to customize. They look at defaults as "suggestions". We want them to instead consider the defaults as "conventions", where if you change them, you are making the decision to "stray from the herd".

Another reason is "semver lawyering": if a default of a setting in the core changes, it is a violation of semver. But if an addon is added or removed in a blueprint for a new application, it isn't, "because that is not part of the core".

Yet another reason is that any code for the non-default setting can be (although in this case it won't be) removed from the core and put in the addon if appropriate api hooks in the core are provided.

chancancode:
I have implemented and landed this feature in master, so you can easily test the impact of this on canary now: https://ember-twiddle.com/?numColumns=2&openFiles=twiddle.json%2Cstyles.app.css

webark:
not that i’m opposed to these changes, but why are we not doing to standard “final comment period” on this before adding it to master? Maybe i’m miss understanding the purpose of the FCP tag though.

chancancode:
@webark I should have been more clear. It has landed in master _behind a flag_. The feature system is designed for this experimental implementation in mind – the code will be stripped in beta and release builds, and even on canary requires opting in (which is the whole purpose of the canary channel).

For this particular change, it's something we (the core team and the community) has wanted for a long time, and the main questions are really around how to roll it out gracefully so it doesn't cause problems in the ecosystem. Having a working implementation allows us to answer those questions much more easily.

This change is also very [small and self-contained](https://github.com/emberjs/ember.js/pull/15981/files#diff-427660a44f95122fbc9ffc3b2f38e4c8L101). Since all the code is already flagged, if we didn't end up wanting this change it would be easy to back it out. (The test changes are a nice refactor regardless.)

chancancode:
And to be extra clear, we are still going to explore the options here and eventually going to do a FCP. The feature cannot be enabled before we do that and merging the RFC.

webark:
Ahhh.. ok cool. Makes sense. Thanks for that clarification @chancancode !

chancancode:
:+1: The prototypes are hopefully going to make it a lot easy to access the merits and impact of this change. (For example, @kategengler found a [potential problem](https://github.com/emberjs/ember.js/issues/15977) with the template-only components RFC yesterday by testing the interaction between that feature and the Ember Inspector.)

chancancode:

1. Should this be an add-on, vs just a setting?

    I think @Gaurav0 summarized it well. The main reason is to allow the flexibility to change the implementation of the flag. For example, in a world where users are building ember in the our build pipeline, we will be able turn this into a build-time feature flag and remove the code and any runtime overhead associated with the unused mode.

Meanwhile, since the feature have landed in canary behind a flag, I encourage you to give it a spin and get a feel for it, either using this [Ember Twiddle](https://ember-twiddle.com/04ea71656db0c3286d329f81c7eab1a4?numColumns=2&openFiles=twiddle.json%2Cstyles.app.css) or by using a canary build with the `ember-glimmer-remove-application-template-wrapper` feature flag set to `true`.

jenweber:
I think this change will be great. I fought with that div in styling before.

What's the best/most likely way that someone would be alerted about this change once it's automatic, sometime in the future?  If I upgraded my app and it suddenly looked different, it would take me a while to look at the top level of the DOM for a missing container. Most other version upgrades would throw a warning but this is styling, so it's sneaky.

chancancode:
@jenweber the plan is this will be opt-in (by installing the addon) for the timeframe of 3.0. Towards the end of 3.0 we can issue a deprecation warning if you still don’t have it installed yet.

text/0000-remove-application-wrapper.md
```diff
@@ -0,0 +1,97 @@
...
+was configurable using the `ApplicationView`, but when views were removed we lost that
+ability. Right now we are stuck with a wrapper element we can't remove nor customize,
+which is why some apps target the selector `body > .ember-view` to style this element.
+
+Similarly, in testing there is another `.ember-view` wrapper inside the
+`#ember-testing` container for no good reason.
+
+This RFC proposes to add a global flag to remove those wrapper elements,
+effectively making the `application.hbs` template have "Outer HTML" semantics, which aligns
+well with [the changes recently proposed](#278) for template-only components,
```
pavloo: The link to #278 is broken here (lmk if it was left intentionally I'm gonna remove this comment)

chancancode: Good catch, fixed!

MelSumner:
This will help Ember apps become more accessible by default. 

When certain html5 elements are the direct descendant of the body element, an [additional ARIA role](https://www.w3.org/TR/wai-aria-1.1/#landmark_roles) is not required. 

While [ember-a11y-landmarks](https://github.com/ember-a11y/ember-a11y-landmarks) provides a solution for current applications, I am encouraged to see a RFC that supports better native accessibility.

chancancode:

PR #281, opened 2017-12-12
ES5 Getters

chancancode:
[Rendered](https://github.com/emberjs/rfcs/blob/es5-getters/text/0281-es5-getters.md)

### DIFF
--- a/text/0281-es5-getters.md
+++ b/text/0281-es5-getters.md
@@ -1,0 +1,180 @@
+- Start Date: 2017-12-12
+- RFC PR: https://github.com/emberjs/rfcs/pull/281
+- Ember Issue: (leave this empty)
+
+# Summary
+
+Install ES5 getters for computed properties on object prototypes, thus
+eliminating the need to use `this.get()` or `Ember.get()` to access them.
+
+Before:
+
+```js
+import Object, { computed } from '@ember/object';
+
+const Person = Object.extend({
+  fullName: computed('firstName', 'lastName', function() {
+    return `${this.get('firstName')} ${this.get('lastName')}`;
+  })
+});
+
+let chancancode = Person.create({ firstName: 'Godfrey', lastName: 'Chan' });
+
+chancancode.get('fullName'); // => 'Godfrey Chan'
+
+chancancode.set('firstName', 'ʎǝɹɟpo⅁');
+
+chancancode.get('fullName'); // => 'ʎǝɹɟpo⅁ Chan'
+
+let { firstName, lastName, fullName } = chancancode.getProperties('firstName', 'lastName', 'fullName');
+```
+
+After:
+
+```js
+import Object, { computed } from "@ember/object";
+
+const Person = Object.extend({
+  fullName: computed('firstName', 'lastName', function() {
+    return `${this.firstName} ${this.lastName}`;
+  })
+});
+
+let chancancode = Person.create({ firstName: 'Godfrey', lastName: 'Chan' });
+
+chancancode.fullName; // => 'Godfrey Chan'
+
+chancancode.set('firstName', 'ʎǝɹɟpo⅁');
+
+chancancode.fullName; // => 'ʎǝɹɟpo⅁ Chan'
+
+let { firstName, lastName, fullName } = chancancode; // No problem!
+```
+
+# Motivation
+
+Ember inherited its computed properties functionality from [SproutCore](http://guides.sproutcore.com/core_concepts_kvo.html).
+The feature was designed at a time before [ES5 getters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get)
+were widely available. This necessitated using a special function such as
+`this.get()` or `Ember.get()` to access the values of computed properties.
+
+Since all of our [target browsers](https://github.com/emberjs/rfcs/pull/252)
+support ES5 getters now, we can drop the need of this special function,
+improving developer egornomics and interoperability between other libraries
+and tooling (such as TypeScript).
+
+Note that at present, using `this.set()` or `Ember.set()` is still mandatory
+for the property to recompute properly. In the future, we might be able to
+loosen this requirement, perhaps with the help of ES5 setters. However, that
+would require more design and is out-of-scope for this RFC.
+
+`this.get()` and `Ember.get()` will still work. This RFC does not propose
+removing or deprecating them in the near term. They support other use cases
+that ES5 getters does not, such as "safe" path chaining (`get('foo.bar.baz')`)
+and `unknownProperty` (and Proxies by extension), so any future plans to
+deprecate them would have to take these features into account.
+
+Addon authors would likely need to continue using `Ember.get()` for at least
+another two LTS cycle (8 releases) to support older versions of Ember (and
+possibly longer to support proxies). It is however very unlikely that the
+everyday user would need to use this.
+
+# Detailed design
+
+The computed property function, along with any caches, can be stored in the
+object's "meta". We will then define a getter on the object's prototype to
+compute the value.
+
+One caveat is that the computed property function is currently stored on the
+instances for implementation reasons that are no longer relevant. However,
+it is possible that some developers have observed their existance and have
+accidentaly relied on these private semantics (e.g. `chancancode.fullName.get()`
+or `chancancode.fullName.isDescriptor`).
+
+Before landing this change, we should turn the property into an assertion
+for so that in these unlikely scenarios the developers will at least receive
+some warning.
+
+Another thing to consider is that there is this Little Known Trick™ to add
+Computed Properties to POJOs:
+
+```js
+import { computed, get } from "@ember/object";
+
+let foo = {
+  bar: computed(function() { return 'bar'; })
+};
+
+get(foo, 'bar'); // => 'bar'
+```
+
+In this case, there is no opportunity for us to install an ES5 getter, and
+`Ember.get` is the only solution. This is very rare in practice and is more
+or less just a party trick. We should deprecate this use case (in `Ember.get`)
+and suggest the alternative:
+
+```js
+import Object, { computed } from "@ember/object";
+
+let foo = Object.extend({
+  bar: computed(function() { return 'bar'; })
+}).create();
+
+foo.bar; // => 'bar'
+```
+
+Or simply...
+
+```js
+let foo = {
+  get bar() {
+    return 'bar';
+  }
+};
+
+foo.bar; // => 'bar'
+```
+
+# ~~How We Teach This~~
+
+For the most part, this RFC _removes_ a thing that we need to teach new
+users.
+
+It might, however, come across as slightly strange that `set()` is still
+required. However, many other libraries share the same model and
+empricially this does not appear to be an issue. For example, in React,
+you can freely access `this.state.foo` but must use `this.setState('foo', ...)`
+to update it. Even Vue has [the same API](https://vuejs.org/v2/api/#Vue-set)
+for some cases.
+
+The mental model for this is that you must use the `set()` in order for
+Ember to notice your mutations, so that it can update the caches, rerender
+things on the screen, etc.
+
+As for users who already learned to use `get()` everywhere, that would
+continue to work. Ideally, this would be a Cool Trick™ they pick up some day
+(as in "Oh, I don't have to do _that_ anymore? Cool."), at which point the
+old habbit would quickly die. If this turned out to be too confusing, we
+could always explore deprecating `this.get()`; we will just have to weight
+the cost-benefits of the confusion (if any) vs churn.
+
+# Drawbacks
+
+As mentioned, not removing `set()` at the same time might be a source of
+confusion. However, removing `set()` would require significantly more
+upfront design work, and it [might not even be possible](https://vuejs.org/v2/guide/reactivity.html#Change-Detection-Caveats)
+to completely remove the need of `set()` (as the system is designed today)
+in all cases (see `Vue.set()`).
+
+Since removing `get()` would unlock so many benefits, and since there are
+plenty of other libraries that uses the same model, the case for decoupling
+the two seems overwhemlingly positive.
+
+# Alternatives
+
+* Hold off until we also remove `set`
+* Hold off until we transition to something like [Glimmer's `@tracked`](https://glimmerjs.com/guides/tracked-properties)
+
+In my opinion, these alternatives does not make a lot of sense, as neither
+of these hypothetical systems appear to require (or would benefit from)
+having a user-land getter system.
--- a/text/0281-es5-getters.md
+++ b/text/0281-es5-getters.md
@@ -175,6 +175,6 @@ the two seems overwhemlingly positive.
 * Hold off until we also remove `set`
 * Hold off until we transition to something like [Glimmer's `@tracked`](https://glimmerjs.com/guides/tracked-properties)
 
-In my opinion, these alternatives does not make a lot of sense, as neither
+In my opinion, these alternatives do not make a lot of sense, as neither
 of these hypothetical systems appear to require (or would benefit from)
 having a user-land getter system.
--- a/text/0281-es5-getters.md
+++ b/text/0281-es5-getters.md
@@ -135,7 +135,7 @@ let foo = {
 foo.bar; // => 'bar'
 ```
 
-# ~~How We Teach This~~
+# How We Teach This
 
 For the most part, this RFC _removes_ a thing that we need to teach new
 users.


rmmmp:
If this is only for computed properties, I believe it adds another confusion since we need to use `this.get()` for normal properties.

chancancode:
@rmmmp you won't have to use `this.get()` for either in the world proposed by this RFC. (In fact you never really "have to" use `this.get()` on "normal" properties.)

rtablada:
Since this is an RFC for the underlying Ember.Object this should also remove the need for `get` with ember data and when resolving services correct @chancancode?

kategengler:
A smattering of addons, some of which are popular, use `unknownProperty` https://emberobserver.com/code-search?codeQuery=unknownProperty

I assume they would either have to change the way they work, or `get` would still be required when using those addons?

chancancode:
> when resolving services

@rtablada if you mean

```js
let obj = Object.extend({
  foo: inject.service()
}).create();
```

...then yes, in this case you should be able to do `obj.foo`

buschtoens:
Just as a general note regarding @kategengler's https://github.com/emberjs/rfcs/pull/281#issuecomment-351118896: In the distant (?) future, we'd have [Proxies](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy).

text/0281-es5-getters.md
```diff
@@ -0,0 +1,180 @@
...
+and tooling (such as TypeScript).
+
+Note that at present, using `this.set()` or `Ember.set()` is still mandatory
+for the property to recompute properly. In the future, we might be able to
+loosen this requirement, perhaps with the help of ES5 setters. However, that
+would require more design and is out-of-scope for this RFC.
+
+`this.get()` and `Ember.get()` will still work. This RFC does not propose
+removing or deprecating them in the near term. They support other use cases
+that ES5 getters does not, such as "safe" path chaining (`get('foo.bar.baz')`)
```
buschtoens: For the curious: Babel 7 will support [optional chaining](https://github.com/babel/babel/tree/master/packages/babel-plugin-proposal-optional-chaining).

wycats:
> A smattering of addons, some of which are popular, use unknownProperty https://emberobserver.com/code-search?codeQuery=unknownProperty

> I assume they would either have to change the way they work, or get would still be required when using those addons?

The way to think about this is:

In today's world, every object that inherits from `Ember.Object` has an interface that requires consumers to use `.get` to access properties. This means that "you must use .get() to access documented properties" is implicit in the documentation for any library that uses `Ember.Object`.

After this change, you would still be **able** to use `.get()` to access properties on any subclass of `Ember.Object`, but it would no longer be the case that it was **required**.

In practice, what this means is that the handful of libraries that want to use `unknownProperty` need to start explicitly documenting `.get()` as part of their interface for getting properties from their objects.

For now, it would be reasonable for users to use `.get()` when interacting with third party libraries, until libraries have generally become more explicit about that. I would also encourage the community to help improve the documentation on this front.

grapho:
Is there a separate RFC that will be introduced which will replace:
`this.set('foo', 'bar')`
with
`this.foo = bar`
?

I dont like being a negative noodle, but it seems to me that as long as set() exists.. get() should also exist for consistency sake.

That being said. I AM super excited to one day be able to set/get things with classic JS "dot style"

chancancode:
@grapho does the "How We Teach This", "Drawbacks" and "Alternatives" section not answer your question?

Besides the empirical evidence that other frameworks have the same API, you can also construct a principled explanation/mental model for this: reading data is always side-effect free, but _mutations_ are by definition effectful. Therefore, it is a feature not a bug to explicitly call out these _mutations_, both for human understanding (something something "mutations are the root of all evil" ¯\_(ツ)_/¯) and for Ember (so it knows to re-render your code).

chancancode:
> We internally use Ember.get() for interop with normal JavaScript objects, so this change shouldn't affect any part of interacting with Ember itself.

Another way to put this is that if the addon is using `unknownProperty` (or proxies) for UI objects intended for templates (e.g. yielding it), then it shouldn't be affected by this – when you do `{{foo.bar.baz}}` in the template, Ember will do the equivalent of `Ember.get(this, 'foo.bar.baz')` which will allow proxies to work.

grapho:
@chancancode good points!
I am in favor of a soft deprecation of get... just not generate deprecation warnings until actual full removal of get becomes closer to realization.

chancancode:
> It's also not uncommon for JavaScript libraries to require .get() for their objects, so it shouldn't be too problematic if libraries that use unknownProperties are documented in this way.

A quick search turned up a few examples:

* http://docs.parseplatform.org/js/guide/
* https://developers.google.com/analytics/devguides/collection/analyticsjs/accessing-trackers

Then there are many other JS APIs that use getters (as in Java-style `obj.getFoo()`):

* https://developers.google.com/maps/documentation/javascript/3.exp/reference#Map
* https://www.mapbox.com/mapbox.js/api/v3.1.1/l-marker/#api

The point is, when interacting with third-party libraries (which an addon is) in JavaScript, sometimes you work with "plain properties", sometimes you are going through something like `obj.get()`, sometimes you are going through getter methods (`obj.getFoo()`) and sometimes you are going through some even more exotic APIs (`obj.ref('foo').value()`...).

So, as long as those special objects (those that need to use `unknownProperties`, usually proxy or proxy-like objects) are documented clearly, the mental model will be that `get()` is just another API on those the object you have to learn.

As @buschtoens said, in the very long term, the community (not just the Ember community, but the wider JS community) will transition to use the native proxies and expose a more uniform API. Until then, _in the rare cases_ where you need to implement proxies in user-land (which, like ES5 getters, are not shimmable), those objects would have to go through something like `.get()` as the user-land trap, which is understandable IMO.

Gaurav0:
The RFC does not make clear how we will know when we will be able to use this. Is it available now? After all, we can access *some* properties now without `get`.

pzuraq:
@Gaurav0 I'm pretty sure this means for all computed/non-real properties specifically, which cannot be accessed without `get` currently

chancancode:
@Gaurav0 currently the mental model I would teach is "always use .get". after the feature lands, you will not need it in general, unless the library/addon tell you otherwise.

As for timing, as always, it will be available when (and if) the RFC is merged, the feature is Go'ed by the core team and when it makes it all the way to the release channel. Might be available earlier on canary behind a flag for experimentation.

Kuzirashi:
I would love to see this implemented... <3

sandstrom:
How [far out](https://caniuse.com/#search=Proxy) are [proxies](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy)? If they would allow dropping `set` and keep support for `unknownProperty` perhaps it's better to wait a little longer — taking down both `get` and `set` in one swoop.

buschtoens:
@sandstrom I think you pretty much answered your question yourself. :stuck_out_tongue: 

IE's the only browser of interest that is dragging behind. Since it will never get Proxy support, we'd basically have to wait until Ember.js drops IE support completely.

Panman8201:
Can Ember utilize [proxy-polyfill](https://github.com/GoogleChrome/proxy-polyfill) or does the polyfill not support enough of it to be worthwhile? I'm thinking ember-cli could include this based on targets.

text/0281-es5-getters.md
```diff
@@ -0,0 +1,180 @@
...
+
+Ember inherited its computed properties functionality from [SproutCore](http://guides.sproutcore.com/core_concepts_kvo.html).
+The feature was designed at a time before [ES5 getters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get)
+were widely available. This necessitated using a special function such as
+`this.get()` or `Ember.get()` to access the values of computed properties.
+
+Since all of our [target browsers](https://github.com/emberjs/rfcs/pull/252)
+support ES5 getters now, we can drop the need of this special function,
+improving developer egornomics and interoperability between other libraries
+and tooling (such as TypeScript).
```
oriSomething: I'm not sure it should be said here. But we can improve ergonomics of tooling such as TypeScript today with `Ember.get`. And if we will able to pass an Array to `Ember.get` like `Ember.get(["some", "nested", "key"])` we would even have better ergonomics with TS

chancancode: I have tried in the past, and I am pretty convinced you cannot type an Ember object successfully today, even without considering nesting. You need to convince TypeScript that they are no properties on the object itself (so obj.foo etc are type errors) but when you do `obj.get(“foo”)` or `get(obj, “foo”)` it should have the right type. I have come close with using some phantom associated type tricks but I think it’s fundamentally impossible.

bryanaka:
I am stoked about this RFC. When teaching people coming from other libs/frameworks, `get` usually comes with more questions than `set` because react's `setState` somewhat normalized the setting function

chancancode:

1. It is confusing that `set` is required, but `get` is not.

    As mentioned in the RFC and pointed out by other commenters, React (and other libraries, and our experience with Glimmer.js) has proved that this is not much of a confusion point. At the same time, there is also confusion that comes with requiring `get()` in the first place. Since this is such a big improvement in ergonomics, we also don't think it is justifiable to hold this back the benefits until `set` is also eliminated, as it is not even clear whether it could be done or not.

2. It does not work with proxies.

    When used in templates (which is true for most use cases for proxies), this is not an issues as Ember/Glimmer internally still uses `.get`. However, we can explore adding development mode assertions (similar to the "mandatory setters" assertions) when we detect that this is a problem.

Meanwhile, I am working on [implementing this feature](https://github.com/emberjs/ember.js/pull/15992) behind a flag (which can be done independently from progressing this RFC). Once the patch is merged, I'll comment on this thread so you can try the feature on Ember Twiddle, or with the canary build.

chriskrycho:
Massive 👏 here.

I have some questions and exploring I need to do on how this would work with TypeScript, most of which have to do with type resolution since `computed` would have a run-time (though not _construction_-time) effect on the type of a `class`, and unfortunately got pulled aside from work directly related to this into a feature this week. But my summary is: I think this should be great for TS consumers.

Now, being able to use getters will itself be a win, and I *suspect* that the combination of getters and (_please_ happen at the January TC39 meeting) decorators should do most of what we need for type-checking to work correctly.

The big problems *today*, for those who don't have direct experience working with TypeScript, have to do with two things: chained key lookup (which this at least partially solves) and having appropriate access to the `this` value in a computed definition (which this doesn't; but that's fine—that's something decorators will be required for). To reuse @chancancode's example, but with TS in play, using ES6/class-style syntax, and with an object to highlight the possible limitations:[[1]](#fn-1)

Before:

```typescript
import EmberObject, { computed } from '@ember/object';

class Person extends EmberObject {
  name: EmberObject<{
    first: string;
    last: string;
  }>;

  fullName = computed('name.{first,last}', function(this: Person) {
    const name = this.get('name');
    return `${name.get('first')} ${name.get('last'))}`;
    // ^ is necessary because doing it the normal way won't typecheck
    // return `${this.get('name.first')} ${this.get('name.last'))}`;
  });
}

let name = EmberObject.create({ firstName: 'Godfrey', lastName: 'Chan' })
let chancancode = Person.create({ name });

chancancode.get('fullName'); // => 'Godfrey Chan'

chancancode.set(chancancode.get('name', 'first'), 'ʎǝɹɟpo⅁');
// ^ is necessary because doing it the normal way won't typecheck
// chancancode.set('name.first', 'ʎǝɹɟpo⅁');

chancancode.get('fullName'); // => 'ʎǝɹɟpo⅁ Chan'

let updatedName = chancancode.get('name');
let { first, last } = updatedName.getProperties('first', 'last');
```

After:

```typescript
import EmberObject, { computed } from "@ember/object";

class Person extends EmberObject {
  name: EmberObject<{
    first: string;
    last: string;
  }>;
  
  fullName: computed('name.{first,last}', function() {
    return `${this.name.first} ${this.name.last}`;
  })
});

let name = EmberObject.create({ firstName: 'Godfrey', lastName: 'Chan' });
let chancancode = Person.create({ name });

chancancode.fullName; // => 'Godfrey Chan'

chancancode.set('firstName', 'ʎǝɹɟpo⅁');

chancancode.fullName; // => 'ʎǝɹɟpo⅁ Chan'

let { name: { first, last }, fullName } = chancancode; // No problem!
```

A few things to note here:

1. We *had* to separately get `name` and then `first` and `last` in the original computed body, and do a lookup before running the `set` or before doing the nested `getProperties` in order for this to type-check. TypeScript does not now (and will presumably never) have the ability to validate the type using keys like `name.first`. *This* should be fixed by this proposal, as we'll be able to just have the getters.

1. We *must* have the `this: Person` declaration in both for everything in the body of the computed property to type-check. This is the aforementioned place we'll need decorators; an ES5-style getter would have the context correctly. ES5 getters as the implementation mechanism *should* clear the path for that, I think?

In summary, I *think* that this is basically just an across-the-board win for TS consumers. :heart_eyes: 

***

**Footnotes:**

1. <a name='fn-1'></a>I'm aware of the fact that installing computed properties this way is not the same – performance-wise or behavior-wise – as using `.extends()`, since it doesn't put it on the prototype but on the instance.

rwjblue:
> In summary, I think that this is basically just an across-the-board win for TS consumers. 😍

I think it's pretty nice for us JS plebs too... :kissing_cat: :stuck_out_tongue_winking_eye:

mike-north:
It is currently possible to get a `ComputedProperty` object using `this.foo` today.
```ts
Ember.Object.extend({
   fullName: computed('firstName', 'lastName', function() {
      return `${this.get('firstName')} ${this.get('lastName')}`;
   }),
   foo() {
       let cp: ComputedProperty<string> = this.fullName; // <----------
   }
});
```
AFAIK it's very uncommon for app developers to use this directly, but I do think that this shouldn't be accidentally broken

rwjblue:
> It is currently possible to get a ComputedProperty object using this.fullName today.

It definitely is not intended (or public API), and `ComputedProperty` instances used in this way have a number of completely broken scenarios. The likely path forward for that scenario is that we will detect when it is done, and assert in development builds (likely via native proxies detecting `.get` being called and/or `isDescriptor` being accessed).

chancancode:
@mike-north as mentioned in the RFC:

> One caveat is that the computed property function is currently stored on the
instances for implementation reasons that are no longer relevant. However,
it is possible that some developers have observed their existance and have
accidentally relied on these private semantics (e.g. `chancancode.fullName.get()`
or `chancancode.fullName.isDescriptor`).
> 
> Before landing this change, we should turn the property into an assertion
for so that in these unlikely scenarios the developers will at least receive
some warning.

I have started working on that in #15992

jelhan:
Sounds great. But one question isn't cleared out for me yet. Perhaps it's obvious but I'm not sure about.

How would that effect the [need of consuming a computed property which should be observed](https://guides.emberjs.com/v2.17.0/object-model/observers/#toc_unconsumed-computed-properties-do-not-trigger-observers)? Would `.get()` still be required therefore? If so it should be made explicit in the *How We Teach That* section and maybe added as another example in *Motivation* section for use cases still requiring `.get()`.

buschtoens:
@jelhan It is not explicitly stated in the RFC, but the way I understand it, ES6 getters would basically trigger the exact same code ES5 [`property_get`](https://github.com/emberjs/ember.js/blob/71b688a0dcac85f1103c81a185d53890a742dfba/packages/ember-metal/lib/property_get.js#L22) does today. This means, nothing would change.

You still have to pseudo-consume unconsumed properties in the `init` hook, for example:

```js
init() {
  this._super(...arguments);

  this.myProperty;
}
```

theseyi:
@buschtoens to clarify, did you mean ES5 getters? I think that's what the RFC is referring to

buschtoens:
@theseyi absoultely. I updated my comment, thank you.

chancancode:
Hello! I implemented this feature (behind a flag). You can try this on canary now: https://ember-twiddle.com/7aded3d76fcc04867b426bf05f641388?openFiles=controllers.application.js%2C

text/0281-es5-getters.md
chriskrycho: For what it’s worth, you can, and we are... as long as you don’t do nested properties. It took a fairly ridiculous amount of work, though.

chriskrycho: To be precise: the types you end up with on the object itself end up having a type of `ComputedProperty<TheType>`, which the type system then properly unwraps when using `get`.

chancancode:

wycats:
For what it's worth, I think it's important that we consider the developer experience of migrating to this flag an open issue that should be addressed during implementation.

The most notable version of the problem is existing code using Ember proxies, and what kind of development mode experience we could use to help people who forget to use .get() on those objects.

chriskrycho:
Implementation question related to my above comments on TypeScript—how does this interact with the use of class instance properties?

```ts
import Service from '@ember/service';
import { computed } from '@ember/object';

class Neato extends Service {
  bar = 'a string';
  foo = computed('bar', function() {
    return this.bar.length;
  });
}
```

Will the proposed (and already-initially-implemented) solution handle this correctly?

pzuraq:
Yeah, the issue there is that currently the getters get applied once when the class is first defined (in the `proto` phase). Since class fields are per-instance, and we don't have a hook to finalize fields, there is no way to transform them into native getters. It also likely wouldn't be preferable anyways, since it would then be creating an instance of the CP per instance of the class, which could have some pretty negative perf implications.

I think the path forward for native classes and computed properties is using decorators, they're the only method we really have to modify the prototype of a class at this point.

chriskrycho:
That is the conclusion I came to as I experimented bit last night. 👍🏼 Come on, TC39, give us that decorator goodness!

wycats:
@chriskrycho since you're using TypeScript, why not just use decorators?

chriskrycho:
@wycats It's a good suggestion. Unfortunately, the result of our experiments with the ember-decorators in TS has been that they don’t work because Babel and TS have different implementations; we'd need to completely reimplement the package. As a result, we've essentially just opted to wait on decorators until we get Stage 3 decorators and Babel's and TS's implementations converge.

pzuraq:
@wycats the main issue with decorators as they stand in Typescript is with class fields. Currently, there is no way to get the initializer for a class field in TS, so some of our decorators can't be made to work with it without breaking changes (specifically thinking of `@className` and `@attribute`).

I think we can take another stab at TS compatibility once Babel finalizes the Stage 2 transforms, since we'll need to rewrite ember-decorators' internals a fair amount at that point. Ideally though we would wait for Stage 3 to avoid more churn/breaking changes.

mike-north:
For those looking to test this out in their own apps, here's a regex that works for most cases of find/replace.
#### ⚠️ Use at your own risk ⚠️
```
REPLACE
\.get\((['"]{1})([\w\-\.]+)\1\)
WITH
\.$2
```

rwjblue:
@mike-north - Hmm, I think it would be much better to use the codemod (which `npx ember-cli-update --run-codemods` would do for you automatically).

https://github.com/rondale-sc/es5-getter-ember-codemod

chancancode:
Also (as mentioned in the RFC) – it's not generally safe to replace _all_ usages of `.get` because your code or addons may return proxies, so some combination of caution, good test coverage and manual review is necessary here.

mike-north:
@rwjblue I didn't know that existed, but have had bad luck with TS and jscodeshift

rwjblue:
@rondale-sc and I recently wrote [codemod-cli](https://github.com/rwjblue/codemod-cli), to make authoring code mods much easier. One of the out of the box features is typescript support. This allowed us to update ember-qunit-codemod to “just work” for typescript files. We should do the same for es5-getter-ember-codemod...

PR #284, opened 2017-12-13
Splitting Ember into packages

ef4:
[Rendered](https://github.com/emberjs/rfcs/blob/explode/text/0000-explode.md)

### DIFF
--- a/text/0000-explode.md
+++ b/text/0000-explode.md
@@ -24,9 +24,16 @@ As of Ember 3.0, the only supported way to depend on Ember is to depend on the `
 
 We propose adding an *additional* supported way to depend on Ember. You can choose to run `ember explode`, and we will remove `ember-source` from your dependencies and replace it with `@ember/core` *plus the current complete set of already-extracted Ember packages*. `ember explode` is intended to not change the set of Ember features in your app. It just splats them out into a form that comes from separate packages.
 
-Critically, `@ember/core` includes an extra rule in its semver policy: the only supported way to upgrade to a newer `@ember/core` is to run the `ember-cli-update` command. The update command will be aware of any additional packages have been split off from `@ember/core` since your last version, and it will add them automatically to your app.
+`@ember/core` and all of the packages split out from it will share a lockstep version number. We will manage releases of these packages in two phases:
 
-This special upgrade requirement gives us the freedom to split off more packages *in any release*, without breaking anybody's app.
+ - initial phase: while we are still establishing the package boundaries, we will start at 0.1.0 and signal breaking releases each time we split out a new package.
+ - stabilization: once we have created a complete set of packages and stabilized their boundaries, we will update the version number to match `ember-source`, after which the split-up family of packages and the aggregated `ember-source` packages will stay in lock-step.
+
+## Release Cadence
+
+This proposal doesn't change Ember's release cadence. The new packages should be released on the same schedule as `ember-source` (and of course they contain the same code -- as long as you install the full set of them).
+
+Even though we expect to have many "breaking" releases of `@ember/core` during the initial phase, they will only "break" in the sense of requiring apps to add additional dependencies on new split-out packages. We can extend the `ember-cli-update` tool to handle this case automatically, making it easier for early adopters of the new package format to follow along.
 
 ## Inter-package dependencies
 
@@ -42,7 +49,7 @@ Therefore, we propose new metadata in the `ember-addon` section of `package.json
 
 ```js
   "ember-addon": {
-    "required-core-features": "3.1.0",
+    "required-core-features": "0.4.0",
     "required-additional-features": ["@ember/routing", "@ember/prototype-etxensions", "@ember/string"]
   }
 ```
@@ -64,35 +71,31 @@ For example, consider an addon with:
 
 ```js
   "ember-addon": {
-    "required-core-features": "3.1.0",
+    "required-core-features": "0.4.0",
     "required-additional-features": ["@ember/routing"]
   }
 ```
 
 If you use this addon in an app with `ember-source` 2.18.0, there's no problems, the metadata isn't affecting anything.
 
-If you use this addon in an app with `@ember/core` 3.6.0, there's still no problems, but we will be able to warn the user that they need to install `@ember/routing` plus whatever future packages happen to get split out of `@ember/core` between 3.1 and 3.6.
+If you use this addon in an app with `@ember/core` 0.6.0, there's still no problems, but we will be able to warn the user that they need to install `@ember/routing` plus whatever future packages happen to get split out of `@ember/core` between 0.4.0 and 0.6.0.
 
 In contrast, if there was a peerDependency on `@ember/core` we would break usage with `ember-source`, and *vice versa*. And if there was a peerDependency on `@ember/routing` it would break usage with any Ember version earlier than the one that split out that package.
 
 The "pure NPM" way of solving this problem is to release new versions of your addon to support new versions of Ember, but this creates a much more fragmented ecosystem than we want. We are careful not to break things very often, such that many addons really don't need to be released just to keep working with newer Ember releases.
 
 
-## Local Dependency Overrides
+## Local and Crowd-sourced Dependency Overrides
 
 We have already proposed making missing dependencies a hard build error. This is important so that weird and frustrating bugs don't crop up unexpectedly. But it needs to be possible to get oneself unstuck without being blocked by third-party code.
 
 Therefore, we propose a `config/dependency-overrides.js` file that allows you to make declarations on behalf of any addon:
 
-
 ```js
-/* 
-  If you need to edit this file, please file PRs to share these 
-  changes with the addon's authors! 
-*/
-module.exports = {
-  "liquid-fire@^0.29.0": {
-    "required-core-features": "3.4.0",
+const Overrides = require('ember-addon-dependency-overrides');
+module.exports = Object.assign({}, Overrides, {
+  "liquid-fire@0.29.0": {
+    "required-core-features": "0.4.0",
     "required-additional-features": [
       "@ember/component"
     ]
@@ -102,7 +105,11 @@ module.exports = {
 
 The point of this file is purely to suppress the build errors you would otherwise receive. 
 
-Each override is scoped to exactly the package *and* version constraint that you're using. If you change the version of the addon that you're using, the override will no longer apply. This is deliberate -- there's no way to safely know what overrides to apply to a new version of the addon.
+Each override is scoped to exactly the package *and* version that you're using. If you change the version of the addon that you're using, the override will no longer apply. This is deliberate -- there's no way to safely know what overrides to apply to a new version of the addon.
+
+The `ember-addon-dependency-overrides` module in the above example can be a community-curated configuration that tracks which popular addons really require which Ember features. This is intended to allow early adopters to help each other without a lot of duplicate work and without requiring every addon to rapidly track all the unstable `@ember/core` releases.
+
+After `@ember/core` and family have been stabilized, the expectation is that most addons will begin publishing their own dependency information and the need for overrides should taper off.
 
 ## Package Dependency Example Scenario
 
@@ -117,9 +124,9 @@ Each override is scoped to exactly the package *and* version constraint that you
     ```diff
     <     "ember-source": "3.2.0",
     ---
-    >     "@ember/core": "3.2.0",
-    >     "@ember/partial": "3.2.0",
-    >     "@ember/prototype-extensions": "3.2.0",
+    >     "@ember/core": "0.6.0",
+    >     "@ember/partial": "0.6.0",
+    >     "@ember/prototype-extensions": "0.6.0",
     ```
 
 3. Your app is expected to work exactly the same at this point.
@@ -129,18 +136,18 @@ Each override is scoped to exactly the package *and* version constraint that you
 5. `ember serve` gives you an error report that includes a message like:
 
     ```
-    Addon liquid-fire 0.29.0 may depend on `@ember/partial` to function correctly. You should either add `@ember/partial` to your app or update to a version of liquid-fire that declares `required-base-features` >= 3.1.0. If you want to test whether it actually works, you can override this warning, see https://...
+    Addon liquid-fire 0.29.0 may depend on `@ember/partial` to function correctly. You should either add `@ember/partial` to your app or update to a version of liquid-fire that declares `required-base-features` >= 0.4.0. If you want to test whether it actually works, you can override this warning, see https://...
     ```
 
-    In the above example, the message said we need to get liquid-fire to declare a required-base-features of at least 3.1.0 because (in this hypothetical example), that is the first version of `@ember/core` that doesn't include the features in `@ember/partial`.
+    In the above example, the message said we need to get liquid-fire to declare a required-base-features of at least 0.4.0 because (in this hypothetical example), that is the first version of `@ember/core` that doesn't include the features in `@ember/partial`.
     
 6. You update liquid-fire and your app is running happily now without the feature. 
 
 ### Scenario: Addon author declaring dependencies
 
-1. Ember 3.2.0 is released, and it splits out `@ember/partial` and `@ember/prototype-extensions` as their own packages.
+1. `@ember/core` 0.6.0 is released, and it splits out `@ember/partial` and `@ember/prototype-extensions` as their own packages.
 
-2. You want to update your `mega-button` addon to take advantage of this new capability. You test your addon against `@ember/core` 3.2.0 and get a test failure:
+2. You want to update your `mega-button` addon to take advantage of this new capability. You test your addon against `@ember/core` 0.6.0 and get a test failure:
 
     ```
     You're trying to use a feature that was moved out of `@ember/core` and into `@ember/partial`
@@ -148,14 +155,14 @@ Each override is scoped to exactly the package *and* version constraint that you
 
     In this case, you got a pleasant error message because we were able to leave a development-mode-only assertion in place of the removed feature. It looks like your mega-button is actually using partials after all.
     
-3. You update your addon's test scenario to use both `@ember/core` and `@ember/partial` 3.2.0. Now tests pass. This confirms that those are the only packages you depend on.
+3. You update your addon's test scenario to use both `@ember/core` and `@ember/partial` 0.6.0. Now tests pass. This confirms that those are the only packages you depend on.
     
 4. You update your package.json to include:
 
     ```
     "ember-addon": { 
       ...
-      "required-base-features": "3.2.0"
+      "required-base-features": "0.6.0"
       ...
     }
     ```
@@ -200,7 +207,6 @@ However, we are also free to create additional packages that are not in that RFC
 
 It will sometimes be convenient to leave development-mode-only stubs in place of removed code so that we can provide helpful guidance. This should be pursued whenever possible, and it can be done automatically wherever we are doing block-level code stripping.
 
-
 ## New App Blueprint
 
 We should create a new app blueprint that allows people to start with just `@ember/core`. There is no plan at this time to replace the default app blueprint -- this would be a new choice for people who want to start with a very barebones Ember, similar to what you'd get with GlimmerJS.
@@ -209,9 +215,7 @@ We should create a new app blueprint that allows people to start with just `@emb
 
 # How We Teach This
 
-The most important message we need to teach app developers is: use `ember-cli-update` whenever you're changing your Ember version. As long as we can spread that message, we can provide direct guidance the rest of the way. For example, when we reach a sufficient level of confidence in the `@ember/core` packaging, we can begin making `ember-cli-update` offer to automatically explode `ember-source` into `@ember/core` *et al*. This would be a simple code mod that should not alter any app semantics.
-
-As people begin to experiment with removing Ember packages, our best teaching opportunities are very clear and helpful feedback from `ember-cli` whenever they have a dependency issue, as illustrated in some of the example messages in this document.
+The most important message we need to teach app developers is: if you want to use `@ember/core` during the initial phase, you should use `ember-cli-update` whenever you're changing your Ember version. As long as apps follow that advice, they should experience the same level of stability they would get from `ember-source`. As people begin to experiment with removing Ember packages, our best teaching opportunities are very clear and helpful feedback from `ember-cli` whenever they have a dependency issue, as illustrated in some of the example messages in this document.
 
 The mental model for "which Ember packages am I using?" should ultimately reduce to simple rules:
 
@@ -233,4 +237,4 @@ In the case of code we want to remove, splitting it into separate packages doesn
 
 We could continue to ship Ember as a single package, and do our own Ember-specific built-time shenanigans to decide how much of it to include in the app's build. This greatly simplifies the management of user's package.json files, but it is increasingly out-of-step with wider NPM-based Javascript toolchains.
 
-We could choose to only split out new packages at major releases. This would eliminate the need for mandatory `ember-cli-update`, at the cost of either shipping many more breaking releases or waiting much longer to benefit from package splitting. I think `ember-cli-update` is the superior option because it gives us a way to refactor Ember without doing breaking changes.
+An earlier version of this document proposed having `@ember/core` and `ember-source` share version numbers even during the initial phase. In this case the `ember-cli-update` tool would be a mandatory part of `@ember/core`'s public API, allowing us to split out more packages even on minor releases. This has the benefit of making the connection between a version of `ember-source` and `@ember/core` clearer, but it has the downside of imposing weirder requirements that are out-of-step with NPM community expectations.
--- a/text/0000-explode.md
+++ b/text/0000-explode.md
@@ -1,0 +1,236 @@
+- Start Date: 2017-12-07
+- RFC PR: 
+- Ember Issue: 
+
+# Summary
+
+This RFC proposes a way to begin immediately splitting Ember into a family of separate NPM packages without breaking ecosystem compatibility.
+
+# Motivation
+
+This proposal
+
+ - enables people working on small, focused apps to start with a minimal, small core of Ember and only add more features as they are needed
+ 
+ - allows us to more aggressively isolate legacy features by making them pay-as-you-go
+ 
+ - creates an upgrade incentive for apps and addons that is "more carrot, less stick".
+ 
+# Detailed design
+
+## @ember/core Package
+
+As of Ember 3.0, the only supported way to depend on Ember is to depend on the `ember-source` npm package. This proposal does not change the meaning of `ember-source`. It will continue to be all-of-Ember in the way people already understand it, and features can only be removed from `ember-source` at major releases after an appropriate deprecation cycle.
+
+We propose adding an *additional* supported way to depend on Ember. You can choose to run `ember explode`, and we will remove `ember-source` from your dependencies and replace it with `@ember/core` *plus the current complete set of already-extracted Ember packages*. `ember explode` is intended to not change the set of Ember features in your app. It just splats them out into a form that comes from separate packages.
+
+Critically, `@ember/core` includes an extra rule in its semver policy: the only supported way to upgrade to a newer `@ember/core` is to run the `ember-cli-update` command. The update command will be aware of any additional packages have been split off from `@ember/core` since your last version, and it will add them automatically to your app.
+
+This special upgrade requirement gives us the freedom to split off more packages *in any release*, without breaking anybody's app.
+
+## Inter-package dependencies
+
+If one of the newly-split-off Ember packages depends on another, it will say so in its NPM peerDependencies. We will use exact-version constraints, which is effectively how things already work today.
+
+NPM has historically been loose about peerDependencies, so they are often ignored by developers. We propose that ember-cli should hard error for missing peerDependencies to avoid this problem (most people who think they don't really need to clean up peerDependency warnings are simply mistaken, and have latent bugs).
+
+## Addon dependencies on Ember packages
+
+Right now, every addon in the ecosystem is allowed to implicitly depend on any feature in Ember. You can't know *a priori* which features are safe to remove if you're using any given addon.
+
+Therefore, we propose new metadata in the `ember-addon` section of `package.json`:
+
+```js
+  "ember-addon": {
+    "required-core-features": "3.1.0",
+    "required-additional-features": ["@ember/routing", "@ember/prototype-etxensions", "@ember/string"]
+  }
+```
+
+`required-core-features` records the latest version of `@ember/core` that the addon has been tested against. It causes your addon to depend on all the Ember packages that were still bundled inside that `@ember/core` version. 
+
+`required-additional-features` lets the addon express dependency on features that have already been split from `@ember/core` as of the base features version.
+
+
+### Why required-core-features instead of peerDependencies?
+
+`required-core-features` says something different than a `peerDependency` constraint.
+
+We want addons to be as aggressive as possible in bumping their `required-core-features`, because that creates the maximum opportunity for apps to pick their packages *a la carte*.
+
+Simultaneously, we want addons to support the widest practical range of Ember versions (which would be their `peerDependencies` constraint on Ember, if they choose to have one), because that makes everybody's upgrades easier. 
+
+For example, consider an addon with:
+
+```js
+  "ember-addon": {
+    "required-core-features": "3.1.0",
+    "required-additional-features": ["@ember/routing"]
+  }
+```
+
+If you use this addon in an app with `ember-source` 2.18.0, there's no problems, the metadata isn't affecting anything.
+
+If you use this addon in an app with `@ember/core` 3.6.0, there's still no problems, but we will be able to warn the user that they need to install `@ember/routing` plus whatever future packages happen to get split out of `@ember/core` between 3.1 and 3.6.
+
+In contrast, if there was a peerDependency on `@ember/core` we would break usage with `ember-source`, and *vice versa*. And if there was a peerDependency on `@ember/routing` it would break usage with any Ember version earlier than the one that split out that package.
+
+The "pure NPM" way of solving this problem is to release new versions of your addon to support new versions of Ember, but this creates a much more fragmented ecosystem than we want. We are careful not to break things very often, such that many addons really don't need to be released just to keep working with newer Ember releases.
+
+
+## Local Dependency Overrides
+
+We have already proposed making missing dependencies a hard build error. This is important so that weird and frustrating bugs don't crop up unexpectedly. But it needs to be possible to get oneself unstuck without being blocked by third-party code.
+
+Therefore, we propose a `config/dependency-overrides.js` file that allows you to make declarations on behalf of any addon:
+
+
+```js
+/* 
+  If you need to edit this file, please file PRs to share these 
+  changes with the addon's authors! 
+*/
+module.exports = {
+  "liquid-fire@^0.29.0": {
+    "required-core-features": "3.4.0",
+    "required-additional-features": [
+      "@ember/component"
+    ]
+  }
+};
+```
+
+The point of this file is purely to suppress the build errors you would otherwise receive. 
+
+Each override is scoped to exactly the package *and* version constraint that you're using. If you change the version of the addon that you're using, the override will no longer apply. This is deliberate -- there's no way to safely know what overrides to apply to a new version of the addon.
+
+## Package Dependency Example Scenario
+
+*All examples in this section may use fictional package names, version numbers, and relationships. For illustration purposes only.*
+
+### Scenario: Depending on an addon with insufficient required-base-features
+
+1. Your app depends on `ember-source` 3.2.0. 
+
+2. You decide to run `ember explode` to try out this brave new world. It changes your package.json like this:
+
+    ```diff
+    <     "ember-source": "3.2.0",
+    ---
+    >     "@ember/core": "3.2.0",
+    >     "@ember/partial": "3.2.0",
+    >     "@ember/prototype-extensions": "3.2.0",
+    ```
+
+3. Your app is expected to work exactly the same at this point.
+
+4. You decide "I'm not using partials in my app", so you remove `@ember/partial`.
+
+5. `ember serve` gives you an error report that includes a message like:
+
+    ```
+    Addon liquid-fire 0.29.0 may depend on `@ember/partial` to function correctly. You should either add `@ember/partial` to your app or update to a version of liquid-fire that declares `required-base-features` >= 3.1.0. If you want to test whether it actually works, you can override this warning, see https://...
+    ```
+
+    In the above example, the message said we need to get liquid-fire to declare a required-base-features of at least 3.1.0 because (in this hypothetical example), that is the first version of `@ember/core` that doesn't include the features in `@ember/partial`.
+    
+6. You update liquid-fire and your app is running happily now without the feature. 
+
+### Scenario: Addon author declaring dependencies
+
+1. Ember 3.2.0 is released, and it splits out `@ember/partial` and `@ember/prototype-extensions` as their own packages.
+
+2. You want to update your `mega-button` addon to take advantage of this new capability. You test your addon against `@ember/core` 3.2.0 and get a test failure:
+
+    ```
+    You're trying to use a feature that was moved out of `@ember/core` and into `@ember/partial`
+    ```
+
+    In this case, you got a pleasant error message because we were able to leave a development-mode-only assertion in place of the removed feature. It looks like your mega-button is actually using partials after all.
+    
+3. You update your addon's test scenario to use both `@ember/core` and `@ember/partial` 3.2.0. Now tests pass. This confirms that those are the only packages you depend on.
+    
+4. You update your package.json to include:
+
+    ```
+    "ember-addon": { 
+      ...
+      "required-base-features": "3.2.0"
+      ...
+    }
+    ```
+    
+    because that is the latest version you have tested. And you also add:
+    
+    ```
+    "ember-addon": {
+       ...
+       "required-additional-features": ["@ember/partial"]
+       ...
+    }
+    ```
+    
+    because that is the only package you discovered you need that is outside of `@ember/core`.
+    
+5. The final result of your work is that apps who use your addon are now free to remove `@ember/prototype-extensions`, but they cannot remove `@ember/partial`.
+
+## How this relates to Svelte Builds
+
+"Svelte builds" is the idea of doing feature flagging in reverse, so that code that's deprecated but not-yet-removed can be stripped out by users who don't depend on it.
+
+This proposal is designed to enable the use of svelte builds *as an implementation detail* that apps and addons don't need to worry about. 
+
+A feature that we want to svelte away can be "moved" into a new package, and users who don't want the feature can simply drop that package. I'm using scare quotes around "move" because the new package itself can be as simple as a private build-time flag that alters which features get stripped from `@ember/core`. This strategy has already worked well in the past for other legacy support packages.
+
+## How this relates to deprecations
+
+The existing deprecation system is complementary to this proposal, and I'm not proposing any changes to deprecations.
+
+We can still deprecate features and leave them in `@ember/core`. But we can also choose to quarantine deprecated features into newly-split-out packages, making them easier for apps to drop immediately.
+
+The two systems have complementary strengths and weaknesses. Only deprecations can permanently remove code that we don't want to support anymore. But package splitting can allow apps to stop shipping deprecated code as soon as they are ready, without waiting for everyone else.
+
+## How this relates to JS Module API
+
+[RFC 176](https://github.com/emberjs/rfcs/blob/master/text/0176-javascript-module-api.md) defined a bunch of packages that are, today, not real NPM packages. This proposal gives us the infrastructure to begin making them into real packages.
+
+However, we are also free to create additional packages that are not in that RFC, particularly in cases where we want to isolate a particular feature or behavior that is on the road to deprecation and removal. Creating a new package requires an RFC (possibly a deprecation RFC).
+
+## Development-mode stubs
+
+It will sometimes be convenient to leave development-mode-only stubs in place of removed code so that we can provide helpful guidance. This should be pursued whenever possible, and it can be done automatically wherever we are doing block-level code stripping.
+
+
+## New App Blueprint
+
+We should create a new app blueprint that allows people to start with just `@ember/core`. There is no plan at this time to replace the default app blueprint -- this would be a new choice for people who want to start with a very barebones Ember, similar to what you'd get with GlimmerJS.
+
+    ember new <app-name> --blueprint @ember/core
+
+# How We Teach This
+
+The most important message we need to teach app developers is: use `ember-cli-update` whenever you're changing your Ember version. As long as we can spread that message, we can provide direct guidance the rest of the way. For example, when we reach a sufficient level of confidence in the `@ember/core` packaging, we can begin making `ember-cli-update` offer to automatically explode `ember-source` into `@ember/core` *et al*. This would be a simple code mod that should not alter any app semantics.
+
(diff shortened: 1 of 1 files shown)


knownasilya:
Beautiful!

Kerrick:
It's unclear to me from the RFC, but will Semantic Versioning be respected in `@ember/core`?

Given the given example from the RFC, let's assume Ember 3.2.0 is released and it splits out `@ember/partial` and `@ember/prototype-extensions` into their own packages. That means `@ember/core` no longer has the same API surface area as it did before, so it is a breaking change. Does that mean that `@ember/core`'s major version number will have to be bumped? If so, that means the `@ember/core` version number won't match the number from Ember 3.2.0. If not, that means `@ember/core` is abandoning Semantic Versioning.

What are the trade-offs between those, and which is the current plan?

cibernox:
I find some are unclear for addon authors. Today is common and a good practice to test your addons in beta and canary versions of Ember. The [Scenario: Addon author declaring dependencies](https://github.com/emberjs/rfcs/blob/explode/text/0000-explode.md#scenario-addon-author-declaring-dependencies) doesn't clarify enough how the `required-base-features` will work when testing with a version beyond stable. 
I have the feeling that if we don't handle that gracefully, people who want to ride the beta-wave (not to mention canary) will have to either use `ember-source` to be safe, or every time they update to a new beta, as the ecosystem will naturally lag behind, they have to heavily edit the `config/dependency-overrides.js` to satisfy all the unmet new packages.

Panman8201:
Am I in the minority of wanting one "monolithic" ember package? I'd rather have everything "ember" at my disposal, then have tree-shaking remove the un-needed stuff on build (shake not only my app code, but what pieces of ember are _not_ being used by the app and addons, etc.) Just seems like this would add so much more complexity when tree-shaking could take care of the job. Or am I missing something?

rtablada:
@Panman8201 that is achieved here. If you want "everything ember" you'll continue to use `ember-source` if you want to explicitly declare your dependencies (or dare I say add Ember features to a Glimmer app in the future) then you can include only the bits you want.

chancancode:
@Panman8201 I think you are definitely in the _majority_. I think most users wouldn't have to worry about this at all, and as long as they continue to use `ember-soure` (maybe rename to something like `@ember/all` in the future) things will work exactly the way they are today. That should still be the default in guides, CLI blueprint etc.

You only need to do this if you are very filesize-conscious (I personally think most apps – at least most "productivity" kind of desktop apps aren't in that category). But it's nice that people in that category can do it if they want to.

chancancode:
Also, I don't think tree shaking can be very effective at removing the kind of features we are talking about here. For example, Mixins, computed properties and observers are all features that are deeply weaved into different parts of the framework. Even if you are not importing the macros in your app the `Ember.Object` (for example) will have to handle all these possibilities in the constructor, so these things still ended up being imported, so from the bundler's perspective they are all "used". Template features like partials are also very difficult to tree-shake, since you don't even have an import statement.

Not that it's impossible, but the bundler would have to get _really_ smart and have a lot of knowledge about Ember's internals to do it correctly.

chancancode:
@Kerrick 

> Critically, @ember/core includes an extra rule in its semver policy: the only supported way to upgrade to a newer @ember/core is to run the ember-cli-update command. The update command will be aware of any additional packages have been split off from @ember/core since your last version, and it will add them automatically to your app.

When you run the upgrade command, it will edit your `package.json` such that the newly split packages are included in your `package.json` by default, so it behaves exactly the same way after the upgrade.

Kerrick:
> When you run the upgrade command, it will edit your package.json such that the newly split packages are included in your package.json by default, so it behaves exactly the same way after the upgrade.

@chancancode The problem with this is that semver does not consider the use of external tools to maintain compatibility -- it only considers version numbers. I understand that the proposal includes an upgrade command to help manage breaking changes in `@ember/core` that result from feature extraction, but will that tool be upgrading the `@ember/core` package's major or minor version?

chancancode:
@cibernox that is the point of the "required-core-features" system. It is basically saying "this addon was last audited as of version 3.10" – so if version 3.11 splits out two additional packages (say `@ember/computed-properties` and `@ember/observers`), then we will conservatively assume the addon requires those features by default.

So everything will still work, but your build might not be as lean as your hoped (it will be no worse than before the upgrade, though). You can use the override feature to test if anything breaks if you insist to remove those features and send a PR to the addon to "re-certify" it against 3.11.

knownasilya:
@Kerrick at the moment `ember-cli-update` has `from` and `to` flags, for specifying versions, so I'm assuming those will stay or additional flags would be added as shortcuts, e.g. `--major`.

chancancode:
> The problem with this is that semver does not consider the use of external tools to maintain compatibility -- it only considers version numbers.

@Kerrick I don't think that is true.

The only relevant concept in Semver is "public API", which is intentionally very loosely defined:

> Software using Semantic Versioning MUST declare a public API. This API could be declared in the code itself or exist strictly in documentation.

As part of the documentation, this piece of software communicates to the users that "the only supported way to upgrade is to use the upgrade command", which does whatever it needs to do to maintain the API compatibility for you.

It is also absolutely not true that you don't have to use "external tools" to maintain compatibility. In practice all node packages have the implicit requirement of "the only supported way to upgrade is to use an NPM-compatible package manager" (where "NPM-compatible package manger" is not even that well defined).

If you try to upgrade a package simply by downloading the tarball of a newer minor version and extract that into the appropriate location in your `node_modules`, it will likely not work because you will be missing some newly required dependency.

It is just a matter of clearly defining what you consider the "public API"  and communicating that clearly with your users, which is why even things like legal documents could be considered "semver-compatible".

text/0000-explode.md
```diff
@@ -0,0 +1,236 @@
...
+ 
+# Detailed design
+
+## @ember/core Package
+
+As of Ember 3.0, the only supported way to depend on Ember is to depend on the `ember-source` npm package. This proposal does not change the meaning of `ember-source`. It will continue to be all-of-Ember in the way people already understand it, and features can only be removed from `ember-source` at major releases after an appropriate deprecation cycle.
+
+We propose adding an *additional* supported way to depend on Ember. You can choose to run `ember explode`, and we will remove `ember-source` from your dependencies and replace it with `@ember/core` *plus the current complete set of already-extracted Ember packages*. `ember explode` is intended to not change the set of Ember features in your app. It just splats them out into a form that comes from separate packages.
+
+Critically, `@ember/core` includes an extra rule in its semver policy: the only supported way to upgrade to a newer `@ember/core` is to run the `ember-cli-update` command. The update command will be aware of any additional packages have been split off from `@ember/core` since your last version, and it will add them automatically to your app.
```
Turbo87: until now `ember-cli-update` is responsible for updating `ember-cli`, I fear that this change force `ember-cli` even more into lockstep with `ember-source` 🤔

text/0000-explode.md
```diff
@@ -0,0 +1,236 @@
...
+
+This special upgrade requirement gives us the freedom to split off more packages *in any release*, without breaking anybody's app.
+
+## Inter-package dependencies
+
+If one of the newly-split-off Ember packages depends on another, it will say so in its NPM peerDependencies. We will use exact-version constraints, which is effectively how things already work today.
+
+NPM has historically been loose about peerDependencies, so they are often ignored by developers. We propose that ember-cli should hard error for missing peerDependencies to avoid this problem (most people who think they don't really need to clean up peerDependency warnings are simply mistaken, and have latent bugs).
```
Turbo87: we have discussed this in the CLI calls and I am very much against a hard error without any escape valves

text/0000-explode.md
```diff
@@ -0,0 +1,236 @@
...
+  changes with the addon's authors! 
+*/
+module.exports = {
+  "liquid-fire@^0.29.0": {
+    "required-core-features": "3.4.0",
+    "required-additional-features": [
+      "@ember/component"
+    ]
+  }
+};
```
Turbo87: how does this format address overriding incompatible peer deps?

text/0000-explode.md
```diff
@@ -0,0 +1,236 @@
...
+
+As the barebones app experiences (described above under New App Blueprint) matures, we should develop a separate guide designed around how to build apps with that starting point. 
+
+# Drawbacks
+
+The biggest drawback of Ember-as-many-packages is that NPM's peerDependency support is relatively weak. Simply saying `npm install @ember/foo` won't automatically take peerDependencies into account to choose the precise version of `@ember/foo` that matches your other packages. We can easily tell you when your peerDeps are wrong, but we can't always easily fix them for you automatically, unless we start laying specialized tools on top of NPM.
+
+In the case of code we want to remove, splitting it into separate packages doesn't eliminate our need to support it (at least until the next major release lets it be truly removed). So while this proposal lets us "go faster" in some senses, it doesn't let us reduce support any faster.
+
+# Alternatives
```
Turbo87: I'm wondering if static (import) analysis to figure out what Ember features are needed would also be a valid option instead of declaring the features explicitly 🤔

Turbo87:
and FWIW I tend to agree with @Kerrick that `@ember/core` would very likely not follow semver if implemented in lockstep with `ember-source`

simonihmig:
What @chancancode said in https://github.com/emberjs/rfcs/pull/284#issuecomment-351539299 was quite helpful to me to clarify the difference to only relying on tree-shaking. Maybe something like this could be added as another "How this relates to" section?

text/0000-explode.md
```diff
@@ -0,0 +1,236 @@
...
+
+## New App Blueprint
+
+We should create a new app blueprint that allows people to start with just `@ember/core`. There is no plan at this time to replace the default app blueprint -- this would be a new choice for people who want to start with a very barebones Ember, similar to what you'd get with GlimmerJS.
+
+    ember new <app-name> --blueprint @ember/core
+
+# How We Teach This
+
+The most important message we need to teach app developers is: use `ember-cli-update` whenever you're changing your Ember version. As long as we can spread that message, we can provide direct guidance the rest of the way. For example, when we reach a sufficient level of confidence in the `@ember/core` packaging, we can begin making `ember-cli-update` offer to automatically explode `ember-source` into `@ember/core` *et al*. This would be a simple code mod that should not alter any app semantics.
```
Gaurav0: Should `ember-cli-update` be added to the default blueprint? I think it would make it easier to set up and teach.

text/0000-explode.md
```diff
@@ -0,0 +1,236 @@
...
+Right now, every addon in the ecosystem is allowed to implicitly depend on any feature in Ember. You can't know *a priori* which features are safe to remove if you're using any given addon.
+
+Therefore, we propose new metadata in the `ember-addon` section of `package.json`:
+
+```js
+  "ember-addon": {
+    "required-core-features": "3.1.0",
+    "required-additional-features": ["@ember/routing", "@ember/prototype-etxensions", "@ember/string"]
+  }
+```
```
Gaurav0: I'm really concerned this puts additional burdens on addon developers, who already have to do a lot. Will this be expected of most/all ember addons? If a single addon doesn't do this, it will be assumed that all of ember is required? Will addons be continually pressured to update "required-core-features"?

text/0000-explode.md
```diff
@@ -0,0 +1,236 @@
...
+In contrast, if there was a peerDependency on `@ember/core` we would break usage with `ember-source`, and *vice versa*. And if there was a peerDependency on `@ember/routing` it would break usage with any Ember version earlier than the one that split out that package.
+
+The "pure NPM" way of solving this problem is to release new versions of your addon to support new versions of Ember, but this creates a much more fragmented ecosystem than we want. We are careful not to break things very often, such that many addons really don't need to be released just to keep working with newer Ember releases.
+
+
+## Local Dependency Overrides
+
+We have already proposed making missing dependencies a hard build error. This is important so that weird and frustrating bugs don't crop up unexpectedly. But it needs to be possible to get oneself unstuck without being blocked by third-party code.
+
+Therefore, we propose a `config/dependency-overrides.js` file that allows you to make declarations on behalf of any addon:
```
Gaurav0: Ember should be striving for zero configuration. This adds lots more configuration that will need to be regularly updated if a user wants to use a recent version of Ember. Right now, on Ember Observer, I did a quick search and found just 17 addons whose `package.json` is up to date with the most recent version of Ember.

I strongly encourage you to rethink this.

workmanw:
This seems cool and I could see how it could make the life easier for the maintainers of Ember. That said, it seems like it would be a lot of effort and I'm struggling to see the real value it offers to Ember's end users.

text/0000-explode.md
ef4: None of this configuration is required in any app. This is purely about making it *not impossible* to spend more effort if you want to make your app smaller.

Apps are free to ignore 100% of this RFC and will *always* be free to do so.

text/0000-explode.md
ef4: Yes, whatever upgrade command we use should be in the default blueprint. It doesn't necessarily even need to be ember-cli-update, it could be something new. I just picked that one because it's the best one that already exists.

ef4:
> Given the given example from the RFC, let's assume Ember 3.2.0 is released and it splits out @ember/partial and @ember/prototype-extensions into their own packages. That means @ember/core no longer has the same API surface area as it did before, so it is a breaking change. Does that mean that @ember/core's major version number will have to be bumped? If so, that means the @ember/core version number will quickly fall out-of-sync with the ember-source version number. If not, that means @ember/core would not follow semver.

This is a misunderstanding of semver. "Public API" it not limited to things like which functions are available. It can include whatever other conditions you want to impose. What is important is that you state clearly what is the supported way to use the software, and anyone who follows those instructions is safe.

So no, we would not bump major versions of `@ember/core` when splitting out packages. That is basically the whole point of this RFC. We can just start splitting packages out of `ember-source` right now if we're willing to only do it at breaking releases.

ef4:
> Today is common and a good practice to test your addons in beta and canary versions of Ember. The Scenario: Addon author declaring dependencies doesn't clarify enough how the required-base-features will work when testing with a version beyond stable.
I have the feeling that if we don't handle that gracefully, people who want to ride the beta-wave (not to mention canary) will have to either use ember-source to be safe, or every time they update to a new beta, as the ecosystem will naturally lag behind, they have to heavily edit the config/dependency-overrides.js to satisfy all the unmet new packages.

The important thing here is that in none of those situations is anybody worse off than they are right now.

Even if you're using `@ember/core` and removing some packages, you can *always* safely upgrade to a newer version with more split-out packages *regardless* of what addons you're using. The upgrade will simply put those packages into your app's dependencies and everything continues working the way it did before.

The only time you need to do extra work is if you get motivated to try to remove some of those newly split-out packages from your app.

Kerrick:
> This is a misunderstanding of semver.

If it's a misunderstanding of semver, it is likely a common one. Everything from third-party tools like David DM and GreenKeeper to the `^` used by yarn and npm assumes that it's safe to upgrade from 3.2.0 to 3.6.0 without breaking your build. That assumption would no longer apply if upgrading `@ember/core` from 3.2.0 to 3.6.0 would break without either using the ember-cli upgrade tool or manually adding more dependencies.

> you can always safely upgrade to a newer version with more split-out packages regardless of what addons you're using.

*If you use the built-in tool and don't use any third-party tooling that has been built around the common community understanding of semver.

---

I'm all in for splitting into multiple packages, but I think that serious thought needs to be given to allowing the major version numbers of `ember-source` and `@ember/core` to drift.

Google does a similar thing with `material-components-web` -- each component (equiv. to `@ember/core` or `@ember/string`) releases a major version number when its own API has breaking changes, and the "wrapper library" (equiv. to `ember-source` or the proposed `@ember/all`) has its own separate release cycle.

text/0000-explode.md
ef4: This is an area of legitimate concern, and one I thought about a lot when writing this RFC. Here is my reasoning.

My goal here is to enable people who *choose* to spend effort on getting code out of their apps to be unblocked from doing so, and that people who don't choose to spend effort on the whole thing should not be impacted. For app authors, I think the proposal clearly achieves that goal.

For addon authors, there is this one-line change in package.json that I am indeed requiring, *if* they want to make it easy for some of their advanced users to start dropping Ember packages. To make that one-line change as easy as possible, I have introduced a mechanism (dependency-overrides.js) that allows the people who are motivated to do the work (the app authors who are trying to remove packages) to do all the testing for you.

The blueprint for addons already puts them into the strictest possible conditions (like disabling prototype extensions) for testing against the widest possible set of Ember apps. What I'm proposing is not different than that. When it's time to add a new ember-try scenario for the first release that offers `@ember/core`, best practice will be to just test against `@ember/core` instead of `ember-source` so that you would see right away if you're still compatibility, making the decision to update `required-core-features` as easy as possible.

I hear what you are saying about completely automatic code removal being better, but it is fundamentally not possible for an awful lot of the code, short of changing Ember's programming model to be much more verbose to make every kind of dependency static and explicit.

ef4:
> This seems cool and I could see how it could make the life easier for the maintainers of Ember. That said, it seems like it would be a lot of effort and I'm struggling to see the real value it offers to Ember's end users.

I agree with @chancancode that many of the apps that are using Ember today shouldn't care about this at all. You're not wrong when you question the value to *current* Ember users.

But the goal here is to enable a whole new class of apps to *become* Ember users. Apps that today would not have chosen Ember because it's "too big" or "too much to learn". We can create a more comfortable adoption path for all those apps.

Kerrick:
> But the goal here is to enable a whole new class of apps to become Ember users. Apps that today would not have chosen Ember because it's "too big" or "too much to learn". We can create a more comfortable adoption path for all those apps.

If the hope is that these sub-packages can become an adoption path for new users to the Ember ecosystem, it’s even _more_ important that they follow the common version of semver on their own, without breaking changes inside a single major release.

Let’s say a creator of that whole new class of apps decides to use `@ember/string@3.6.0` and installs it as `^3.6.0` in their non-Ember app. They don’t use `ember-cli-upgrade` because they don’t use Ember CLI or most of the rest of the Ember ecosystem.

Now let’s say that `@ember/string@3.7.0` removes the `pluralize` and `singularize` methods to extract them into `@ember/inflection` instead. Now we’ve got a user who has a broken application because their build system auto-updated `@ember/string` from `3.6.0` to `3.7.0` which removed public API methods. That user feels burned by this, so they may never continue exploring Ember and become a user.

In an alternate world, `@ember/string` was bumped to `4.0.0` instead of `3.7.0` when removing those methods. The user’s app never broke, and their tooling let them know that `@ember/string` was out-of-date. They read the release notes and see a note about installing `@ember/inflection` if they want to keep the removed methods. The user can make an informed decision and feels much more confident in the Ember ecosystem than if their build had broken out from under them.

ef4:
@Kerrick ah, you are hitting on some important points I did not consider.

I should probably say explicitly that the upgrade rule applies to `@ember/core` itself only. We would not be splitting code out of `@ember/string` without bumping major versions.

Your point still stands that `@ember/core` having a weird upgrade policy is suboptimal from a compatibility-with-the-wider-world perspective. We just face a fundamental tradeoff here. We have a particular class of changes we want to make that can be 100% automated in a reliable way. And we will always have another class of changes that are truly breaking. We need to communicate the difference to users.

We are also working within the limitations of NPM, which makes meta-packages not really work well.  It would be ideal to just have an `ember` package that depends on all the others, such that we can rearrange the others however we want, and users who depend on the meta package wouldn't care. The problem is that you can't actually force npm to do that properly unless you do it with peerDeps, and npm gives you no help in solving the constraint problem of getting a set of peerDeps that all approve of each other.

If somebody wants to figure out how to make a meta package work well, I would be happy with that alternative. I have gone pretty far down the path of trying to design that, and each time discovered I would need to do "weird" things that NPM doesn't normally do to make it work.

mehulkar:
What's the minimum package set that makes Ember, "ember"? I had thought until now that `ember-source` was the most minimal. What limited subset would `@ember/core` actually contain? (And would any of the other exploded packages actually be "ember"? E.g. `@ember/string` doesn't sound like ember, it sounds like a string library that happens to be published in ember's namespace, and that maybe Ember.js the framework uses.

locks:
> What's the minimum package set that makes Ember, "ember"? I had thought until now that `ember-source` was the most minimal.

That is the question, isn't it ;)

How I view this initiative is that we want to allow users to opt into a smaller subset of Ember than `ember-source`.  I think the examples in the RFC of `@ember/partials` and `@ember/prototype-extensions` illustrate nicely the sort of packages that a user might want to opt out ahead of their removal from ember.js, which would also be hard to do via tree shaking mechanisms as their use is fairly implicit.

> (And would any of the other exploded packages actually be "ember"?

I am not sure I fully understand this point, as we have already [broken Ember into separate packages](https://github.com/emberjs/rfcs/blob/master/text/0176-javascript-module-api.md).
What does "being ember" mean to you?

> E.g. @ember/string doesn't sound like ember, it sounds like a string library that happens to be published in ember's namespace, and that maybe Ember.js the framework uses.

[Hhhmm…](https://github.com/emberjs/rfcs/blob/master/text/0236-deprecation-ember-string.md) ;)

vasilakisfil:
If you can folks make it possible to extract ember-data outside ember to use it as a standalone API wrapper, I think it will be huge win. Apart from that, splitting ember into packages seems quite reasonable and the way to go.

jnfingerle:
I do understand and agree that this RFC is in line with "vanilla" SemVer, but as @chancancode wrote:

> In practice all node packages have the implicit requirement of "the only supported way to upgrade is to use an NPM-compatible package manager" (where "NPM-compatible package manger" is not even that well defined).

So, in essence, this proposal would introduce a way to use Ember, that is in fact outside the "NPM-compatible package manger"-ecosystem, but looks weirdly like it was part of it. @Kerrick has given a perfect example (Strings etc) how that would hurt. One way to deal with this would be to have `@ember/core` versions out of sync with `ember-source` etc. As I understand that this hurts in other ways, why not start with `0.x.y` versions for `@ember/core` while there are still many packages factored out, and start stable version numbers in sync with `ember-source` when the main work is done, but in accordance to the classic `npm` notion of SemVer?

workmanw:
@ef4 

> But the goal here is to enable a whole new class of apps to become Ember users. Apps that today would not have chosen Ember because it's "too big" or "too much to learn". We can create a more comfortable adoption path for all those apps.

Honestly this feels like a facade. I can't imagine someone deciding, "I'm just going to pick up the Ember core and the router, but not use services or components". Because even if these things are decoupled at the package level, they're still part of the same ecosystem. Guides, Docs, Tutorials, Addons, etc would all have to take extreme care to explain and enforce this.

> You're not wrong when you question the value to current Ember users.

It's also about the time and resources that go into this project that could be spent elsewhere. Not just the author's time, but those who review code changes, testing, documenting, reporting bugs that arise, etc.

ef4:
@workmanw 

> Honestly this feels like a facade. I can't imagine someone deciding, "I'm just going to pick up the Ember core and the router, but not use services or components"

The likely scenario is to start with just components. `@ember/core` would converge with [GlimmerJS](https://glimmerjs.com/). 

I absolutely agree with the premise that most real apps will eventually need most of what's in Ember. But the problem is lots of people don't understand that when they're getting started, so they pick a component-only library over Ember, and then end up reinventing all the other stuff as they go, much to their lost productivity and stability, and to our loss as a smaller community.

> It's also about the time and resources that go into this project that could be spent elsewhere

This proposal is deliberately intended as infrastructure that should make everything else go faster, not slower.  That is why I wrote this RFC in the first place. I don't happen to be a user that is obsessed with byte size. I'm motivated by strategic efforts to help the whole community ship faster.

There is still plenty of old and broken stuff floating around in Ember that nobody has stepped up to fix because there's no near-term payoff, since even deprecated stuff needs to stay around for quite a long time. This proposal creates a near-term payoff to fixing.

It also gives us general-purpose tools for backward compatibility, such that all future changes get easier to make, because "how to swap out the old code without breaking apps" becomes a problem with a standard solution. This lowers the amount of effort required for lots of features that people want.

It also creates a clear path for advanced users to experiment with next-generation replacements for things like the router. The more experimentation we unlock, the faster we can stabilize new features and bless them as the official release.

Your questions are good ones and they are forcing me to be more clear than I was in the original doc. The highest-level goal of this proposal is to make it easier to ship improvements in Ember, so that more people can overcome the intimidation and effort barriers required to help.

text/0000-explode.md
```diff
@@ -0,0 +1,236 @@
...
+
+We propose adding an *additional* supported way to depend on Ember. You can choose to run `ember explode`, and we will remove `ember-source` from your dependencies and replace it with `@ember/core` *plus the current complete set of already-extracted Ember packages*. `ember explode` is intended to not change the set of Ember features in your app. It just splats them out into a form that comes from separate packages.
+
+Critically, `@ember/core` includes an extra rule in its semver policy: the only supported way to upgrade to a newer `@ember/core` is to run the `ember-cli-update` command. The update command will be aware of any additional packages have been split off from `@ember/core` since your last version, and it will add them automatically to your app.
+
+This special upgrade requirement gives us the freedom to split off more packages *in any release*, without breaking anybody's app.
```
jnfingerle: I believe this to be wishful thinking. You'll be able to tell people that it's their own fault to have shot themselves in the foot. But people do expect packages in the npm-ecosystem to adhere to "npm-SemVer", that is, to behave just like any other npm package. Having this special requirement locks us into our own walled SemVer garden. 

Nothing against frequent breaking changes that can be mitigated by a special tool. But we shouldn't redefine what to expect from version numbers in a `package.json`.

text/0000-explode.md
jnfingerle: Keeping "npm-SemVer" would lessen the strain here. You would still have to teach people to cope with frequent major versions by using the provided magic. But you wouldn't lose people along the way that you haven't reached yet.

text/0000-explode.md
```diff
@@ -0,0 +1,236 @@
...
+
+# Drawbacks
+
+The biggest drawback of Ember-as-many-packages is that NPM's peerDependency support is relatively weak. Simply saying `npm install @ember/foo` won't automatically take peerDependencies into account to choose the precise version of `@ember/foo` that matches your other packages. We can easily tell you when your peerDeps are wrong, but we can't always easily fix them for you automatically, unless we start laying specialized tools on top of NPM.
+
+In the case of code we want to remove, splitting it into separate packages doesn't eliminate our need to support it (at least until the next major release lets it be truly removed). So while this proposal lets us "go faster" in some senses, it doesn't let us reduce support any faster.
+
+# Alternatives
+
+We could continue to ship Ember as a single package, and do our own Ember-specific built-time shenanigans to decide how much of it to include in the app's build. This greatly simplifies the management of user's package.json files, but it is increasingly out-of-step with wider NPM-based Javascript toolchains.
```
jnfingerle: Abandoning the "npm-SemVer" model won't help to keep in step with wider NPM-based Javascript toolchains.

wycats:
> Am I in the minority of wanting one "monolithic" ember package? I'd rather have everything "ember" at my disposal, then have tree-shaking remove the un-needed stuff on build (shake not only my app code, but what pieces of ember are not being used by the app and addons, etc.) Just seems like this would add so much more complexity when tree-shaking could take care of the job. Or am I missing something?

I want to echo @chancancode. I think for apps that don't already have a reason to trim down Ember, the "single package" approach is working very well, and it's something we want to continue to support. For what it's worth, Skylight (Tilde's main app) is such an app.

Ember is now and will always be a "convention over configuration" framework first, and we (or at least I) have no desire to force a menu of choices on our users if they don't want it.

On the other hand, people are increasingly trying to use Ember in more resource-constrained mobile environment, and in practice those users are doing very aggressive (and very non-standard) things to cut down the size of Ember.

This RFC is targeted at those users, and tries very hard to avoid leaking those concerns to the majority of users who are not very concerned about this issue.

It also provides an opt-in (and compatible) way for addons to declare what features they need, so that addons can help out apps that are trying to minimize their payloads.

My expectation is that addon authors would be receptive to being more explicit about which features they use in order to help out apps that really need it, but addons can continue to assume "all of Ember" compatibly if they want.

text/0000-explode.md
wycats: I don't think it's correct to say that this proposal "abandons the npm semver model".

In general, this proposal uses `peerDependencies` in the usual way. It also creates a new API (like any other API), which allows addons to ask to opt-out of features they don't want to use.

Declaring a baseline version is basically just saying "I want all of the features that were on by default in 3.5", and then the package opts out of additional features known at the time. That means that those same features will still be active in 3.6 without any changes.

In a future version, if an addon wants to drop additional features "by default", it can increment its baseline. This is not in **conflict** with npm's story, and it doesn't change anything about semantic versioning. It's just expressing something a little more subtle, but if people ignore the subtlety, the only consequence is that they pull in more code (which is the status quo).

text/0000-explode.md
wycats: Maybe a more appropriate, semver-consistent approach would be similar to what this RFC proposes for `ember-source`.

If we want to break up a package, we can leave around the original package as a "meta-package", but it is now possible to opt out of part of it.

@ef4 do you remember if we considered and rejected this when we discussed this?

knownasilya: `ember-source` would still follow semver, but would be made up of multiple packages that have their own versions. If a sub-package has a major bump, and if it is able to be integrated into `ember-source` without bumping the major version of `ember-source`, then it will otherwise it would have to wait for a major release to make it in.

wycats:
@jnfingerle I appreciate your focus on semantic versioning here, as I think it's a very important constraint.

I think it would be very productive to get specific about the ways in which you think this proposal diverges. For what it's worth, @ef4, @chancancode and I thought a lot about semantic versioning, but reading your comments, it's clear we made some mistakes.

It might just be that we got too clever or subtle with the way we interpreted semantic versioning, it may be that we explained it poorly, or it may be that we simply were wrong about the details.

I'm definitely interested in drilling in harder: can you be more specific about where you feel the semantic versioning holes are?

text/0000-explode.md
jnfingerle: In `npm`'s story I could just increase the minor version of `@ember/core`, `npm install`, and everything should work. This proposal abandons this model with the need to use special tooling outside the scope of regular `npm`-ed packages. This is what I was referring to.

knownasilya: That isn't exactly correct, because if you have multiple packages that depend on each other, you'd also have to bump the other packages manually. This tooling basically helps with that manual bit.

wycats:
> Honestly this feels like a facade. I can't imagine someone deciding, "I'm just going to pick up the Ember core and the router, but not use services or components"

Over a longer time horizon, I suspect that we will be able to get away with using tree shaking and more traditional packaging strategies to deal with this set of problems.

However, Ember itself was not written with those strategies in mind (in fact, it was written at a time where the opposite constraints were common). The point of this RFC is largely to find a transition path from Ember's original implementation strategy towards the strategies that people intuitively prefer.

A lot of the friction here comes from the fact that in order to make Ember more "pick and choose" we need to disentangle parts of Ember that have more-or-less been entangled for its entire history. Because so much of the ecosystem implicitly depends on the entanglement, it's not always possible to transparently allow people to opt out of parts of Ember without impacting other users compatibly.

The goal of this RFC is to create a more explicit way for apps that care about "opting into a smaller build" to do so very explicitly, but without changing the standard way of using Ember.

Let's keep driving towards a way to break up Ember into smaller pieces for people who want to use a smaller subset of Ember, while at the same time avoiding impacting mainline Ember use. If you feel weirded out by this RFC, it's because there are real risks to this change, but I think it's one we need to do, and I also think that we can make it nearly invisible to today's Ember users with some design effort.

wycats:
> Honestly this feels like a facade. I can't imagine someone deciding, "I'm just going to pick up the Ember core and the router, but not use services or components"

To say a bit more about my previous comment (I realize it was implicit), while it might be hard to imagine using Ember without "the router", it's not too hard to imagine using Ember without "renderTemplate", but a lot of those features are deeply entangled in the current code.

The primary goal of this RFC is to allow people to use "modern Ember style" even if it's very hard to technically separate the code. We want new features to lean on using ES6 module imports and normal tree shaking techniques, but we need a strategy that were written without that constraint in mind.

As an example future item, we might want to allow people to use Ember without `Ember.Object` (optionally), and migrate to an ES6 class style. We wouldn't want to deprecate `Ember.Object` quickly, because of how disruptive that would be to today's applications, but we need a mechanism for letting people opt out of including the `Ember.Object` code (a nontrivial amount) if they are willing to write code without `Ember.Object`.

Panman8201:
> We want new features to lean on using ES6 module imports and normal tree shaking techniques, but we need a strategy that were written without that constraint in mind.

Isn't that going backwards? Sub-packaging would have been good for the "globals" days, but imports + tree shaking is the way forward.?.

Wouldn't this also put more strain on the release folks?? Instead of releasing one package, now its many packages. Don't want to over extend peoples time..

And for releases, will _all_ package version numbers be bumped for a new release, even if nothing has changed? Ex: It would be odd to see the (latest versions) router at 3.2.4 and controller at 3.2.0.

To add, I've been concerned about ember changes before (even prior to RFC's), but they worked out Ok in the end. So I'm not overly concerned about this, just concerned that it'll add more complexity than what might be needed.

wycats:
> Isn't that going backwards? Sub-packaging would have been good for the "globals" days, but imports + tree shaking is the way forward.?.

We need to split off the (finite) code that's hard to move forward with so that we can move forward with more modern techniques going forward.

The intent is not to keep doing the package strategy indefinitely. The intent is to use the package strategy a few times to bootstrap a codebase that can use more modern strategies.

text/0000-explode.md
Gaurav0: The question for me is to what extent this RFC will impact addon developers, not app developers.

Gaurav0:
I honestly think that using 0.* version numbers for @ember/core and other subpackages, until they are stable and unlikely to break up further, might be an acceptable compromise over the semver issue.

To what extent does having parts of Ember have lockstep versioning with the overall Ember package really provide value? We could consider keeping `@ember/core`'s version be the version of `ember-source` minus 3.

ef4:
Thank you everyone for your input. I just pushed an update to the document that takes into account what I see as the two biggest points raised in this discussion:

 - because my proposed semver rules for `@ember/core` were too nonstandard, I changed the proposal to say that we should start at 0.1.0 for all the split out packages and signal breaking changes each time we split out a new package. Only after stabilizing all the package boundaries would we jump ahead to match version numbers with `ember-source` again.

 - to avoid burdening addons during the initial unstable phase, I updated the local overrides section to also include a shared community package for tracking addon feature dependencies. This would allow the early adopters of `@ember/core` who are motivated to get their apps working with less Ember packages to do most of the work without blocking on addon authors.

text/0000-explode.md
jnfingerle: @knownasilya Could you give an example where - given packages with regurlar npm/SemVer semantics - I cannot bump just one minor version number in my `package.json` without bumping others? My understanding is, that the whole point of SemVer is that I'm allowed to do that.

jnfingerle:
@wycats 
> I'm definitely interested in drilling in harder: can you be more specific about where you feel the semantic versioning holes are?

For completeness' sake: I saw the problem with introducing a new SemVer rule. While the result was still a valid variant of SemVer, it diverged from conventional npm/`package.json` SemVer, basically breaking SemVer for anyone not aware of the additional rule. This anyone could have been people (newcomers to the Ember community not yet educated, package users from the wider node/npm community not interested to go full Ember) as well as tools targeted to the npm ecosystem that Ember users might want to use. 

That said, **all this has been fixed in the new version** @ef4 pushed yesterday.

text/0000-explode.md
ef4: @jnfingerle as written, I've proposed making the separate packages have exact peerDependencies on each other. So if `@ember/routing` needs `@ember/core`, it would express that via an exact version peerDependency.

Over time we're free to relax those constraints to use `^` or `~` dependencies instead, which would enable increased mixing and matching. But the simplest thing for both consumers and contributors is to start with the most exact constraint first.

This is all just standard npm stuff now (the latest version of the RFC removed the special upgrade requirement).

ef4: @jnfingerle also I wrote the above comment before I saw that you've basically already said the same thing at the bottom of the thread. Sounds like we are approaching consensus.

Kerrick:
The new version @ef4 pushed addresses my concerns. Love this RFC now!

text/0000-explode.md
```diff
@@ -0,0 +1,240 @@
...
+
+## Inter-package dependencies
+
+If one of the newly-split-off Ember packages depends on another, it will say so in its NPM peerDependencies. We will use exact-version constraints, which is effectively how things already work today.
+
+NPM has historically been loose about peerDependencies, so they are often ignored by developers. We propose that ember-cli should hard error for missing peerDependencies to avoid this problem (most people who think they don't really need to clean up peerDependency warnings are simply mistaken, and have latent bugs).
+
+## Addon dependencies on Ember packages
+
+Right now, every addon in the ecosystem is allowed to implicitly depend on any feature in Ember. You can't know *a priori* which features are safe to remove if you're using any given addon.
```
alexander-alvarez: looks like this doesn't belong `*a priori*`

ef4: [It's intentional](https://www.merriam-webster.com/dictionary/a%20priori).

locks:
Thanks to everyone who contributed so far! The Core team discussed this RFC at the last meeting and there was consensus on moving it to FCP, so here we go :)

text/0000-explode.md
kellyselden: That is the only part that jumped out at me too. Hard errors might cause more pain than what it would be trying to solve. At least once I got peer deps unsatisfied warnings from NPM, where I knew my setup was more correct than NPM's algorithm. I think we should roll this out without hard errors, then bring that discussion up when it becomes a problem.

kellyselden:
I believe there were some good points brought up by @Turbo87 above that haven't been addressed regarding peer deps.

wycats:
@kellyselden 

> At least once I got peer deps unsatisfied warnings from NPM, where I knew my setup was more correct than NPM's algorithm

So here's the thing: virtually nobody can reasonably make this part of their normal workflow. Trying to manually resolve graphs of dependencies, while allowing duplicates "when appropriate", and resolving peer dependencies manually at the top level is just a nonsense requirement that nobody can do.

The original version of this proposal tried to avoid that problem by moving the resolution for this particular problem into side-configuration, so that Ember could handle the resolution for you and avoid the brokenness of npm's peer dep warnings.

However, there was strong pushback that we should just "go with npm" and not "reinvent the wheel". I'll accept either of the two, but if we're gonna go with peer dependencies, we need to be able to assume they work. If we think people will need to manually do work to get around brokenness, we really need to use the side-config approach.

So which is it?

jnfingerle:
@wycats 

>  I'll accept either of the two, but if we're gonna go with peer dependencies, we need to be able to assume they work.

If npm really is that broken it has to be fixed. If that isn't possible, by all means, run your own *but* don't do it inside the existing npm ecosystem disguised as part of it when it isn't. 

Maybe the peerDependencies approach with an "I know what I'm doing and don't want hard errors" config option might be a solution?

Turbo87:
> Maybe the peerDependencies approach with an "I know what I'm doing and don't want hard errors" config option might be a solution?

That's essentially what I meant with "escape valve" in https://github.com/emberjs/rfcs/pull/284#discussion_r156804814

One example where the npm peer deps warning was entirely wrong for me:

- `eslint-plugin-qunit` (with `"eslint": ">=3.18.0 <5.0.0"` peer dep)
- `ember-cli-eslint` depends on `broccoli-lint-eslint` depends on `eslint`

The nested `eslint` version was correct and satisfying the `eslint-plugin-qunit` peer dep constraint, but npm was still complaining because I don't have `eslint` as a direct dependency.

These things seem to happen quite often to me and I'm worried that enforcing peer deps in the same way will create much more pain than it solves. I'm absolutely okay with displaying warnings for them, but I'm opposed to hard errors if there is no way to turn them off.

Sidenote: we should consider how to implement this in Ember CLI without losing startup performance

tomdale:
@Turbo87 In your eslint example, isn't the issue resolved by including `eslint` as a direct dependency of the app? To me, this seems like okay before because there are no guarantees that a future release of `broccoli-lint-eslint` won't change its version to something that no longer satisfy's `eslint-plugin-qunit`'s peer dependency.

In other words, it seems okay to me to say that transitive dependencies are not allowed to fulfill peer dependency requirements. The worst case scenario is you can just adopt the dependency with the same version range as the other dependency bringing it in transitively. In your example, you could add `eslint@4` to your `package.json` and dependencies would resolve the same as before but now satisfy the peer dependency.

Is there a downside to this that I'm missing?

Turbo87:
> isn't the issue resolved by including eslint as a direct dependency of the app?

then something like `yarn run eslint` would potentially run a different ESLint version than the one that `broccoli-lint-eslint` is resolving to, which would not be such a great solution unfortunately.

> this seems okay because there are no guarantees that a future release of broccoli-lint-eslint won't change its version

I think the expectation is that if `broccoli-lint-eslint` (and `ember-cli-eslint`) update to a new major `eslint` version they would also bump majors. They are essentially exposing the ESLint APIs directly to the users so any breaking API change (including rule changes) from `eslint` would also be a breaking change for those libs.

> In your example, you could add eslint@4 to your package.json and dependencies would resolve the same as before, but now satisfy the peer dependency.

Until someone updates `eslint` to 5.0.0 and forgets about `ember-cli-eslint`. Essentially it's just a workaround for the "broken" warnings in npm and yarn, but at least they're just warnings, not hard errors.

rwjblue:
Removing Final Comment Period so that we can address recent feedback...

NullVoxPopuli:
Has the feedback been addressed?

ef4:
Update: this RFC has been on the back burner for a while but it is still quite relevant and the text holds up well.  I still think it's a good idea, I just haven't been actively championing it as working on embroider comes higher in my priorities list. 

The only major remaining objection was concern about making invalid peerDependencies into errors. I accept that the NPM ecosystem is riddled with problems that make keeping peerDeps valid difficult in general. My suggestion is:

 - reduce the scope of that feature so that it only errors on unsatisfied peerDeps in Ember's own packages. 
 - provide an environment variable to suppress the error so that people who are doing advanced debugging of the ember packages themselves can test scenarios that involve mismatched package versions.

If anybody has some time to take over and incorporate that into the text, I think this RFC is still right near the home stretch.

hakilebara:
If the latest feedback has been addressed by https://github.com/emberjs/rfcs/pull/646, it is OK to move this RFC along its path to be merged?

text/0000-explode.md
```diff
@@ -0,0 +1,242 @@
...
+
+Therefore, we propose new metadata in the `ember-addon` section of `package.json`:
+
+```js
+  "ember-addon": {
+    "required-core-features": "0.4.0",
+    "required-additional-features": ["@ember/routing", "@ember/prototype-etxensions", "@ember/string"]
+  }
```
mehulkar, line 56: Just flagging that there's going to be some amount of overlap with deprecation staging config here #649. We discussed making it possible to identify "parts of ember they do not use", but decided against it. Nothing really to do at the moment, since it's not really feasible to try to expand the scope of both RFCs and there isn't anything inherently conflicting yet. Just wanted to let you know. cc @pzuraq @rwjblue.

text/0000-explode.md
```diff
@@ -0,0 +1,242 @@
...
+In contrast, if there was a peerDependency on `@ember/core` we would break usage with `ember-source`, and *vice versa*. And if there was a peerDependency on `@ember/routing` it would break usage with any Ember version earlier than the one that split out that package.
+
+The "pure NPM" way of solving this problem is to release new versions of your addon to support new versions of Ember, but this creates a much more fragmented ecosystem than we want. We are careful not to break things very often, such that many addons really don't need to be released just to keep working with newer Ember releases.
+
+
+## Local and Crowd-sourced Dependency Overrides
+
+We have already proposed making missing dependencies a hard build error. This is important so that weird and frustrating bugs don't crop up unexpectedly. But it needs to be possible to get oneself unstuck without being blocked by third-party code.
+
+Therefore, we propose a `config/dependency-overrides.js` file that allows you to make declarations on behalf of any addon:
```
mehulkar, line 94: This was also a topic of discussion in #649. @wycats was of the opinion that overriding an addon with config is not the right path to promote. If an addon is out of date, the incentives should be to go fix the addon and use a soft fork in the meantime.

text/0000-explode.md
```diff
@@ -0,0 +1,242 @@
...
+
+## New App Blueprint
+
+We should create a new app blueprint that allows people to start with just `@ember/core`. There is no plan at this time to replace the default app blueprint -- this would be a new choice for people who want to start with a very barebones Ember, similar to what you'd get with GlimmerJS.
+
+    ember new <app-name> --blueprint @ember/core
+
+# How We Teach This
+
+The most important message we need to teach app developers is: if you want to use `@ember/core` during the initial phase, you should use `ember-cli-update` whenever you're changing your Ember version. As long as apps follow that advice, they should experience the same level of stability they would get from `ember-source`. As people begin to experiment with removing Ember packages, our best teaching opportunities are very clear and helpful feedback from `ember-cli` whenever they have a dependency issue, as illustrated in some of the example messages in this document.
```
mehulkar, line 220: `ember-cli-update` should maybe become `ember update`. Opened #653.

wycats: Good catch.

NullVoxPopuli:
What's left on this? :thinking:

jelhan:
> What's left on this? 🤔

Has implementation even started? As far as I know `ember-source` is still the only way to depend on Ember.

Some preparation work has been started for sure. Some features earlier included in `ember-source` package have been replaced by standalone packages. `@glimmer/component`, `@glimmer/tracking` and `@ember/test-helpers` are some examples.

But a common Ember all still relies heavily on fake packages like `@ember/service`, `@ember/routing` and `@ember/controller`. All of them are still provided by `ember-source`.

It even gets worse if also pulling all the fake packages into account, which are less commonly used by an application itself. By far the most packages listed in Ember API docs are fake packages provided by `ember-source`. It's still a long, long way until `ember explode` is possible as described in this RFC.

But to be honest I'm not sure if that's still the right target. Maybe instead of splitting up existing Ember we could continue to implement _new_ replacement as standalone packages. It's a slow transition path.

But I'm also not sure if there are that many use cases which require a developer to start with `@glimmer/component` and install all their way up to Ember. If it is about bundle size, maybe we could achieve the same quicker focusing on strategies like dead-code elimination (tree-shaking), deprecating features in core like `@ember/component` and excluding them from the built based on optional features.

I don't have much insights in core development. But it seemed as if doing so worked out quite well in the last two years.

ef4:
I still think the high-level goals of the RFC are good. I think we would want to do an updated iteration that takes advantage of all the progress we've already made recently across the ecosystem.

For example, we're starting to see addons shipping in v2 format now. Since V2 addons have to ship their code in a form that's statically analyzable, we can reliably tell which Ember packages they really use. So they wouldn't need the `required-core-features` and `required-additional-features` stuff that's in this RFC.

As of now, that would cover addon usage of Ember modules via JS imports, but when you combine it with https://github.com/emberjs/rfcs/blob/master/text/0496-handlebars-strict-mode.md via something like https://github.com/emberjs/rfcs/pull/779, we would have full visibility into what parts of Ember are used by any addon. That solves the hard migration problem that this RFC deals with.

Also, as of Ember 4.0 we made sure that apps are required to have support for v2 addon format via ember-auto-import >=2. That means we're free to ship ember-source itself as a v2 addon, which makes all usage of ember modules pay-as-you-go.

So I think we're actually making huge progress in moving the ecosystem toward making it possible to ship ember as "real" packages without disruption.

The next ecosystem-wide thing I'd like to do that would protect users from breakage as you move toward ember being "real packages" made out of "real modules" is deprecating all usage of `require` and `define`. But we need to ship the replacement first, which probably means making our DI system asynchronous and enabling top-level await.

webark:
If we are considering a revised RFC around this concept, would we ever consider allowing for the new "isolated" modules to not require glimmer as it's rendering engine? Or will the packages always require the assumption that the rendering engine/library is glimmer.

wagenet:
@ef4 should we leave this open still?

ef4:
This is still a good goal, but the details have likely changed since this was written and an updated proposal would be needed to move this forward.

The crux of the issue is that to maintain correctness while using separate packages we really need reliable peer dependencies, and NPM clients are still pretty terrible at peer dependencies, though the exact ways in which they are good and bad are shifting. pnpm is currently the best at correctness, but if we want to lean heavily on that we'll need an RFC asking the community to all adopt pnpm, and that might be a tough sell for some companies.

PR #287, opened 2017-12-22
Promote {{-in-element}} to public API

cibernox:
[Rendered](https://github.com/cibernox/rfcs/blob/make-in-element-public/text/0000-promote-in-element-to-public-api.md)

### DIFF
--- a/text/0000-promote-in-element-to-public-api.md
+++ b/text/0000-promote-in-element-to-public-api.md
@@ -60,15 +60,15 @@ a deprecation. I suggest keeping the deprecated private API will until the first
 
 ### Small proposed changes
 
-There is however one part of the behaviour that the core team wants to make explicit before promoting
+There is however one part of the behavior that the core team wants to make explicit before promoting
 the private API to public, and that is how the content is added to the destination when there is other
 content already there.
 
 The desired behavior is that, by default, the rendered content will **replace all the content of the destination**,
 effectively becoming the its `innerHTML`.
 In the current behaviour the rendered content is appended as the end of any existing content. This will still
-be supported by passing `nextSibling=null`, but it will not be the default anymore.
-Any other value passed to `nextSibling` must produce an error.
+be supported by passing `insertBefore=null`, but it will not be the default anymore.
+Any other value passed to `insertBefore` must produce an error.
 
 
 # How We Teach This
--- a/text/0000-promote-in-element-to-public-api.md
+++ b/text/0000-promote-in-element-to-public-api.md
@@ -23,10 +23,13 @@ time to make it public.
 # Detailed design
 
 The existing API of `{{-in-element}}` is very simple and I do not suggest making any change to it.
-It takes a DOM element as the only positional param and a block, and renders that block _inside_ the
-given element instead of where it would normally go.
 
-P.e.
+* It takes a single positional param `destinationElement` that is a DOM element, and a block
+* The given block is rendered not where it is located, but inside the given `destination` element.
+* If `destinationElement` is false/null/undefined then it doesn't render anything but it doesn't error.
+* If `destinationElement` changes the block is removed from the previous destination and added to the new one.
+
+Example usage:
 
 ```hbs
 {{#-in-element destinationElement}}
@@ -37,6 +40,7 @@ P.e.
 The current implementation only suggests creating a new `{{in-element}}` construct that is a simple
 alias of `{{-in-element}}` with the exact same params and behavior, and then, after a while, remove
 the private one.
+
 Although `{{-in-element}}` is technically private, there there is enough people using it to deserve
 a deprecation. I suggest keeping the deprecated private API will until the first LTS release of the
 3.X cycle (3.4) to be finally removed in the next one (3.5).
--- a/text/0000-promote-in-element-to-public-api.md
+++ b/text/0000-promote-in-element-to-public-api.md
@@ -24,18 +24,23 @@ time to make it public.
 
 The existing API of `{{-in-element}}` is very simple and I do not suggest making any change to it.
 
-* It takes a single positional param `destinationElement` that is a DOM element, and a block
+* It takes a single positional param `destinationElement` that is a DOM element, and a block.
 * The given block is rendered not where it is located, but inside the given `destination` element, at
 the end of it if there is any other content on the destination element.
-* If `destinationElement` is false/null/undefined then it doesn't render anything but it doesn't error.
+* If `destinationElement` is null/undefined then it doesn't render anything but it doesn't error.
+* If `destinationElement` is false/0/"" it raises an assertion in development but fails silently in production.
 * If `destinationElement` changes the block is removed from the previous destination and added to the new one. This
 process tears down the rendered content on the initial destination and renders it again on the new one, meaning
-that any component withing the block will be destroyed and instantiated again, so transient HTML state
-like the value of an input will be lost unless manually preserved somewhere else, like a service.
+that any component withing the block will be destroyed and instantiated again (calling the appropiate lifecycle hooks),
+so transient HTML state like the value of an input will be lost unless manually preserved somewhere else, like a service.
 * If the destination element is an invalid value (a string, a number ...) it throws an `parent.insertBefore is not a function` error. I think
 that throwing an error is correct but the error message could be improved.
-* If the destination element has a different context than the origin (SVG) the content respects the
-context of the parent.
+* If the destination element has a different context (like SVG) the content will be appended normally by the glimmer VM,
+which doesn't try to validate the correctness of the generated HTML. This is normal behavior in Glimmer, not
+an exception, and users must be aware that rendering invalid markup might be interpreted or auto-corrected in
+unexpected ways by the browser when in SSR mode.
+* Rendering into a foreign object (an element within an `<iframe>`) should be disallowed initially. If someone
+asks for this feature it would require an RFC to explore the consequences.
 
 Example usage:
 
--- a/text/0000-promote-in-element-to-public-api.md
+++ b/text/0000-promote-in-element-to-public-api.md
@@ -22,7 +22,7 @@ time to make it public.
 
 # Detailed design
 
-The existing API of `{{-in-element}}` is very simple and I do not suggest making any change to it.
+The existing API of `{{-in-element}}` is very simple:
 
 * It takes a single positional param `destinationElement` that is a DOM element, and a block.
 * The given block is rendered not where it is located, but inside the given `destination` element, at
@@ -57,6 +57,19 @@ the private one.
 Although `{{-in-element}}` is technically private, there there is enough people using it to deserve
 a deprecation. I suggest keeping the deprecated private API will until the first LTS release of the
 3.X cycle (3.4) to be finally removed in the next one (3.5).
+
+### Small proposed changes
+
+There is however one part of the behaviour that the core team wants to make explicit before promoting
+the private API to public, and that is how the content is added to the destination when there is other
+content already there.
+
+The desired behavior is that, by default, the rendered content will **replace all the content of the destination**,
+effectively becoming the its `innerHTML`.
+In the current behaviour the rendered content is appended as the end of any existing content. This will still
+be supported by passing `nextSibling=null`, but it will not be the default anymore.
+Any other value passed to `nextSibling` must produce an error.
+
 
 # How We Teach This
 
--- a/text/0000-promote-in-element-to-public-api.md
+++ b/text/0000-promote-in-element-to-public-api.md
@@ -19,7 +19,8 @@ time to make it public.
 
 Create a new `{{in-element}}` as simple alias of `{{-in-element}}` with the exact same signature
 and semantics and deprecate the private one.
-The deprecated private API will continue to exist until the first LTS release of the 3.X cycle (3.4)
+Although the API is technically private, there I believe there is enough people using it to deserve
+a deprecation. The deprecated private API will continue to exist until the first LTS release of the 3.X cycle (3.4)
 to be finally removed in the next one (3.5).
 
 # How We Teach This
--- a/text/0000-promote-in-element-to-public-api.md
+++ b/text/0000-promote-in-element-to-public-api.md
@@ -1,0 +1,53 @@
+- Start Date: 2017-12-22
+- RFC PR: (leave this empty)
+- Ember Issue: (leave this empty)
+
+# Summary
+
+Promote the private API `{{-in-element}}` to public API as `{{in-element}}`.
+
+# Motivation
+
+Developers need to render content out of the regular HTML flow, and `{{-in-element}}` for
+that purpose, but it remains private (or _intimate_) API. People is usually wary of using
+private APIs (and for good reason) as they may get removed at any time.
+
+If the core team and the community is happy with the current behavior of `{{-in-element}}` it's
+time to make it public.
+
+# Detailed design
+
+Create a new `{{in-element}}` as simple alias of `{{-in-element}}` with the exact same signature
+and semantics and deprecate the private one.
+The deprecated private API will continue to exist until the first LTS release of the 3.X cycle (3.4)
+to be finally removed in the next one (3.5).
+
+# How We Teach This
+
+This will be a new build-in helper and must be added to the guides and the API.
+For most usages, it will replace some community solution created with the same goal, like
+[ember-wormhole](https://github.com/yapplabs/ember-wormhole) or [ember-elsewhere](https://github.com/ef4/ember-elsewhere).
+It would be for the best to let the authors of those addons know about this feature so they can
+deprecate their packages if they feel there is no longer a need for them, or at least update their
+Readme files to let their users know that there is a built-in solution in Ember that might cover
+their needs.
+
+# Drawbacks
+
+By augmenting the public API of the framework, the framework is committing to support it for the lifespan
+of an entire mayor version (Ember 4.0).
+
+# Alternatives
+
+We can decide that the framework does not want to make public and support this feature, and continue
+to rely on community-built addons like we've done until today.
+
+# Unresolved questions
+
+Do we want to make any improvement to `{{-in-element}}` before making it public API?
+
+Some possible ideas:
+- Allow to _conditionally_ render the block in place. See https://github.com/DockYard/ember-maybe-in-element
+- Allow to receive not only DOM elements as first argument, but also strings, representing the ID of
+  other CSS selector.
+- Modify or improve the way it behaves during SSR using ember-fastboot.
--- a/text/0000-promote-in-element-to-public-api.md
+++ b/text/0000-promote-in-element-to-public-api.md
@@ -25,9 +25,17 @@ time to make it public.
 The existing API of `{{-in-element}}` is very simple and I do not suggest making any change to it.
 
 * It takes a single positional param `destinationElement` that is a DOM element, and a block
-* The given block is rendered not where it is located, but inside the given `destination` element.
+* The given block is rendered not where it is located, but inside the given `destination` element, at
+the end of it if there is any other content on the destination element.
 * If `destinationElement` is false/null/undefined then it doesn't render anything but it doesn't error.
-* If `destinationElement` changes the block is removed from the previous destination and added to the new one.
+* If `destinationElement` changes the block is removed from the previous destination and added to the new one. This
+process tears down the rendered content on the initial destination and renders it again on the new one, meaning
+that any component withing the block will be destroyed and instantiated again, so transient HTML state
+like the value of an input will be lost unless manually preserved somewhere else, like a service.
+* If the destination element is an invalid value (a string, a number ...) it throws an `parent.insertBefore is not a function` error. I think
+that throwing an error is correct but the error message could be improved.
+* If the destination element has a different context than the origin (SVG) the content respects the
+context of the parent.
 
 Example usage:
 
--- a/text/0000-promote-in-element-to-public-api.md
+++ b/text/0000-promote-in-element-to-public-api.md
@@ -8,20 +8,38 @@ Promote the private API `{{-in-element}}` to public API as `{{in-element}}`.
 
 # Motivation
 
-Developers need to render content out of the regular HTML flow, and `{{-in-element}}` for
-that purpose, but it remains private (or _intimate_) API. People is usually wary of using
-private APIs (and for good reason) as they may get removed at any time.
+Sometimes developers need to render content out of the regular HTML flow. This concept is often also
+called "portals". Some components like dropdowns and modals use this technique to render stuff close
+to the root of the page to bypass CSS overflow rules. Some apps that are embedded into static pages
+even use this technique to update parts of the page **outside** the app itself.
+
+This need need has been covered by solutions developed in the user space but it was so common that
+glimmer baked it into the VM in the form of `{{-in-element}}`, but it remains private (or _intimate_) API.
+People is usually wary of using private APIs (and for good reason) as they may get removed at any time.
 
 If the core team and the community is happy with the current behavior of `{{-in-element}}` it's
 time to make it public.
 
 # Detailed design
 
-Create a new `{{in-element}}` as simple alias of `{{-in-element}}` with the exact same signature
-and semantics and deprecate the private one.
-Although the API is technically private, there I believe there is enough people using it to deserve
-a deprecation. The deprecated private API will continue to exist until the first LTS release of the 3.X cycle (3.4)
-to be finally removed in the next one (3.5).
+The existing API of `{{-in-element}}` is very simple and I do not suggest making any change to it.
+It takes a DOM element as the only positional param and a block, and renders that block _inside_ the
+given element instead of where it would normally go.
+
+P.e.
+
+```hbs
+{{#-in-element destinationElement}}
+  <div>Some content</div>
+{{/-in-element}}
+```
+
+The current implementation only suggests creating a new `{{in-element}}` construct that is a simple
+alias of `{{-in-element}}` with the exact same params and behavior, and then, after a while, remove
+the private one.
+Although `{{-in-element}}` is technically private, there there is enough people using it to deserve
+a deprecation. I suggest keeping the deprecated private API will until the first LTS release of the
+3.X cycle (3.4) to be finally removed in the next one (3.5).
 
 # How We Teach This
 


Turbo87:
@cibernox are you aware of https://github.com/glimmerjs/glimmer-vm/pull/619?

cibernox:
@Turbo87 I knew it was public API in glimmer as it was mentioned in https://emberjs.com/blog/2017/10/10/glimmer-progress-report.html#toc_what-s-new-in-glimmer, but afaik it was not made public in Ember yet.

lukemelia:
I am in favor. It would be lovely to deprecate ember-wormhole, possibly preserving the name for a future generation of ember-wielding intergalactic travelers. 🚀

mehulkar:
Could you point to any references to the existing private/intimate API for those of us who are not intimate with it yet so we can actually see what it is? :)

Serabe:
@mehulkar https://emberjs.com/blog/2017/10/10/glimmer-progress-report.html#toc_portals

chancancode:
@cibernox Thank you for putting this together! Can you expand the RFC a bit to not assume prior knowledge of the current private API?

1. Motivation: why would you need to use this, some example use cases, and why does this API solve those problems
2. Detailed design: what arguments does it take, what does it do, what happens when you pass fasley arguments, etc

The second point is especially important – enumerating the actual API surface helps avoid leaking "accidental" behavior into the public API, and also help reviewers identify potential problems.

Panman8201:
What about having a helper that targets named outlets instead? Ex: `{{#into-outlet name="foobar"}}` Although I believe named outlets went the way of the Dodo, it seems more "ember-ish" than targeting a DOM element. This kinda feels like ember actions vs DOM events, (aside from the technical differences) one is the "ember way" and the other is the DOM way. Everyone knows what an outlet is and you wouldn't have to lookup the DOM element on a controller/component first. This would obviously require more thought (if it's even possible) but just an idea..

cibernox:
@Panman8201 One scenario that your idea doesn't cover, is rendering  content from inside an Ember app to anywhere else in the page. Even somewhere that is outside the Ember app (sometimes the app is not the entire page but only a section inside a static page)

cibernox:
@chancancode I've added your suggestions.

chancancode:
@cibernox looks great!

1. What happens when `destinationElement` is not falsy but not a DOM element? (what do we do today? it should probably assert in dev – at minimum we should explicitly state that it is not supported/undefined behavior)

2. What happens when `destinationElement` is a foreign element that belongs to a different document (e.g. in iframe?)? (I suspect this is probably an uncanny valley today, it probably "sort of works" but I bet it is not spec-compliant, or causes errors in some edge case – we probably want to disallow it)

3. What happens if you pass an element whose "context" is different than where you are now? For example:

    ```html
    <svg />
    <body>
      <div><!-- html context -->{{#in-element (jquery-select "svg")}}...<!-- suddenly in svg context -->{{/in-element}}</div>
    </body>
    ```

    Same thing when `destinationElement` is a `<math />`, `<head />`, `<script />`... etc. Does it work today? Do we expect this to cause problems? Should we disallow it for now? (I think one of the main use case is to render into `title` though.) /cc @wycats @krisselden @mmun @ef4

4. We should explicitly state that when you change `destinationElement`, the block is _torn down_ and re-rendered (it calls the `destroy` hook on components, then `create` new ones at the new location, etc). This is perhaps not the ideal behavior, but it is how it works today. Maybe we should see if there is a way for us to carve out space in the proposal so that we can change this behavior in the future to moving the elements without tearing them down. (This comes with their own complexity, e.g. what if it changes from an HTML element to an SVG element or a foreign element?)

5. We should probably clarify that this _appends_ to the end of the element (synchronously – is it problematic to guarantee that? cc @wycats). So `{{#in-element e}}foo{{/in-element}}{{#in-element e}}bar{{/in-element}}` should add `foobar` to `e`. There is a `nextSibling` argument that we support today to change this behavior, do we want to standardize that as well? If so, that comes with its own set of error cases – is falsy the same as `null`? what if `nextSibling` is not a direct children of `destinationElement`? what if it changes? etc.

6. I'm not sure how to say this – but "as usual" if you mutate the DOM yourself after rendering you might break subsequent re-render.

cibernox:
@chancancode I updated the doc

1. Today a truthy value that does not "cuac" with ".appendBefore" throws an error. I believe that the behavior is correct, but the error message should me more helpful and only `(HTML|SVG)Element`s should be valid targets. Added to the rfc.

2. I didn't add anything on iframes, but it sounds like something we want to disallow with development assertions. 

3. The rendered block should respect the context of where it is rendered. If a block is rendered inside an SVG element the content nodes should behave like children of SVG. Added to rfc.

4. Changing from HTML to SVG and vice versa should work. For what I've tested this already works because on each change of the destination element the entire block is torn down and rendered again as mentioned on the point above. Tearing everything down and rendering it again seems the safest behavior to me because of this. Added to RFC.

5. Right now it's how it works. Added to RFC unless someone sees a technical reason why this can't be guaranteed.

6. That is regular behavior in Ember and all frameworks I know. I did't see the point in mentioning as it's business as usual.

wycats:
Like @chancancode, I am excited for this RFC.

For future readers, I think something like [ember-elsewhere](https://github.com/ef4/ember-elsewhere) is a better fit for rendering components with data from one part of an Ember app into another part of an Ember app.

But there are plenty of cases (like the very common desire to render into `<title>`) that justify this API, and I was delighted to support it directly inside of Glimmer VM. Among other things, this means that the API is directly supported by the VM, and **doesn't require custom DOM escape valves**, which means it's yet another piece of common DOM work that can be modelled completely declaratively.

Let me try to answer @chancancode's questions, but these are just my personal answers (don't treat them as authoritative if you disagree 😉).

> What happens when destinationElement is not falsy but not a DOM element? (what do we do today? it should probably assert in dev – at minimum we should explicitly state that it is not supported/undefined behavior)

I think this should assert in dev and fail silently in production. In general, the philosophy of Glimmer is to allow many dynamic questions to fail silently in production to keep apps running, while being very noisy in dev (and tests) to try to catch as many of these bugs as possible ahead of time.

I also think this should assert for any value other than `null`, or `undefined`, especially the falsy values `""` and `0`. I'm not sure how I feel about `false`, but it seems semantically strange and I don't intrinsically want to support it.

> What happens when destinationElement is a foreign element that belongs to a different document (e.g. in iframe?)? (I suspect this is probably an uncanny valley today, it probably "sort of works" but I bet it is not spec-compliant, or causes errors in some edge case – we probably want to disallow it)

I think we should disallow this until we have a separate RFC that covers how we should deal with foreign elements in general. Should the foreign context adopt the element? Would that cause other issues? How would either solution play with SVG?

> What happens if you pass an element whose "context" is different than where you are now? For example:

There are two categories of "contexts" in HTML that are relevant to Glimmer:

* Namespaces, which affect the rendering of content even when the element is created using `createElement`. This means that if you put a `document.createElement('g')` into an `<svg>` element, it will not render as expected because it has the wrong namespace.
* HTML parser contexts ("in body", etc), which primarily cause elements to be ignored or moved during parse-time.

Today, we are very careful about namespaces, so if you put a block with `<a />` inside of an `<svg>` element, we will correctly create the `<a>` tag with the correct (SVG) namespace. For namespace purposes, Glimmer treats template stitching as an extension of the HTML parser, and carefully follows the parser's rules as if the whole content was parsed as a single piece of HTML (this includes things like correct handling of `<foreignElement>` and casing rules).

On the other hand, we do not (currently) reject stitching together content because the child content is not allowed by the HTML parser. In practice, this typically doesn't cause problems, with one major exception. In SSR mode, Glimmer serializes its DOM into HTML, which is then parsed on the client. If the parsing fails (for example, if a piece of content is ignored or moved), the behavior will differ between client-side rendering and SSR + rehydration.

This is a real issue and one that we should address systemically, but I don't think we need to address it in this RFC to make progress. When we do address it, we will need to handle moving a `<p>` into a `<tr>` via `{{in-element}}` the same way that we will need to handle a `{{yield}}` producing the same result. If past is prelude, it will probably be deprecated slowly and carefully, with a noisier deprecation in SSR mode.

> We should explicitly state that when you change destinationElement, the block is torn down and re-rendered (it calls the destroy hook on components, then create new ones at the new location, etc). This is perhaps not the ideal behavior, but it is how it works today. Maybe we should see if there is a way for us to carve out space in the proposal so that we can change this behavior in the future to moving the elements without tearing them down. (This comes with their own complexity, e.g. what if it changes from an HTML element to an SVG element or a foreign element?)

Teardown + rerender are today's semantics, and those semantics avoid thorny (design **and** implementation) issues that imo we shouldn't try to resolve today.

I think the RFC should be explicit about today's semantics, including the execution of `destroy` hooks, and that if we want to enable "move" semantics, we should do so in a more holistic way that would also work for moving components between lists and other desirable "move" scenarios.

> We should probably clarify that this appends to the end of the element (synchronously – is it problematic to guarantee that? cc @wycats). So {{#in-element e}}foo{{/in-element}}{{#in-element e}}bar{{/in-element}} should add foobar to e. There is a nextSibling argument that we support today to change this behavior, do we want to standardize that as well? If so, that comes with its own set of error cases – is falsy the same as null? what if nextSibling is not a direct children of destinationElement? what if it changes? etc.

I think we can get away with making it append-only for now (to avoid dealing with the cases you're describing). The only question is whether we want append vs. replace. For example, in the `<title>` scenario, you can imagine someone putting a placeholder value in the DOM so that something shows up regardless of whether the `{{in-element}}` is hit during initial render (and also while model data is loading).

I think "replacing the entire content" is probably the majority use-case, so maybe what we want to do is make `replace` the default but also allow an `position="append"` option, which would could later expand to support `nextSibling` if needed.

(in most cases, people could work around the nextSibling limitation by creating a wrapper element -- it's not usually good enough, but I suspect that for the use-cases targeted by this RFC, it would be managable for the time being).

> I'm not sure how to say this – but "as usual" if you mutate the DOM yourself after rendering you might break subsequent re-render.

We probably should have a formal description of this usual caveat. It amounts to a somewhat clear technical description around the details of bounds markers (and similar attribute concepts) in the Glimmer implementation, but we would want to turn those limitations into something understandable for humans.

For this RFC, it's probably acceptable to gesture towards "similar limitations as manually mutating a component's DOM".

simonihmig:
The following is related, but probably should not prevent progress of this RFC. I have been playing with this a bit recently (see https://github.com/kaliber5/ember-in-element-polyfill), and what I find in particular difficult is to get this to work properly with FastBoot. The problem is that `destinationElement` expects a DOM node, which is not so easy to get a reference for when running in FastBoot. `ember-wormhole` has the same problem, and solves it by
* getting access to `document` (in case of FastBoot the `simple-dom` instance) through private APIs, see https://github.com/yapplabs/ember-wormhole/blob/f6bdea79ee1d9b3ef6b7f19e378e3bfb103b2e5e/addon/utils/dom.js#L47-L63
* polyfilling `document.getElementById` (which is not available in `simple-dom`), see https://github.com/yapplabs/ember-wormhole/blob/f6bdea79ee1d9b3ef6b7f19e378e3bfb103b2e5e/addon/utils/dom.js#L26-L43

So for this to become easily usable in FastBoot I think it would be useful to either
* have a public way of getting the `document` reference (`simple-dom`) and some sort of DOM traversal support in `simple-dom`, at least `getElementById`, or even `querySelector` (see https://github.com/ember-fastboot/simple-dom/issues/53)
* or extend the `in-element` API to also accept an ID or even a full CSS selector as `destinationElement`, with a FastBoot-compatible implementation that hides the above mentioned  nasty implementation details from the user.

Gaurav0:
Another issue is that ember wormhole currently requires you to use renderInPlace in acceptance tests. Is it possible to come up with a way to fix this?

cibernox:
@Gaurav0 ember-wormhole does not require to use `renderInPlace` in tests, `ember-power-select` used it (until 2 weeks ago) and it works well, as long as the destination is on the page.

ef4:
@simonihmig you're quite right that needing access to an actual DOM node makes this API hard to use. In addition to your fastboot example, it's just fundamentally timing-dependent in a tricky way.

But I also agree that's not a problem this rfc should solve. I don't think it's even a problem glimmer should solve, because it would introduce a lot of complexity into a layer that doesn't need to have it. `in-element` is the smallest low-level primitive that lets us build solutions on top of glimmer without delving into internals.

I would like to continue making ember-elsewhere be a pleasant high-level API. It should adopt in-element as implementation, but also hide the timing dependency for you (so that you can render source and destination in either order, or even put the destination inside a conditional and turn it on and off). All that is easier done in normal "userspace" ember using a service to coordinate.

wycats:
@cibernox Can you update the RFC with some of the information in my comment (if you agree with it) or raise questions (if you don't).

I think it would be great to FCP this sucker, but @chancancode is right that we need to get the answers to his questions into the text of the RFC.

cibernox:
I updated the RFC. The only bit I'm not convinced about (or I misunderstood) is the append vs replace.

I believe that is not that uncommon to want to wormhole two thing in the same destination. Example: Two dropdown opened simultaneously (not all dropdowns are mutually exclusive).

Right now this seems to work with `{{-in-element}}`, so I'd like to keep it that way at least by default.

wycats:
@cibernox the concern that @chancancode had, which I want to reiterate, is that supporting `nextSibling` requires us to specify the behavior for various values of `nextSibling`, like null, undefined, strings, etc. It requires us to deal with `nextSibling` that isn't a child of the destination element as well as deal with what happens if it changes.

I was thinking in the near term it would be fine to just avoid those issues by going with replace semantics for now. Alternatively, if you think the semantics are important, you could figure out how it works now (in terms of all of those questions) and document it in the RFC 😄

simonihmig:
> I believe that is not that uncommon to want to wormhole two thing in the same destination. 

I would want to second that! So this would require *append*-semantics, rather than *replace*.

FWIW, `ember-cli-document-title` does not use `-in-element`, but a few addons use it already today for dropdowns, tooltips or modals, to move the overlay to the top of the DOM tree, e.g. to prevent potential clipping issues with `overflow: hidden` by ancestor DOM elements. Examples: `ember-basic-dropdown`, `ember-bootstrap`, `ember-popper`. For these use cases append-semantics are required (those UI elements are not mutually exclusive), which should be the default semantics IMHO. (and as far as I'm concerned the only one, at least for now)

Regarding the `nextSibling` thing: I don't feel a strong need to have this now, as part of this RFC...

wycats:
@simonihmig @cibernox if we think we want both replace and append semantics, but don't need to worry about arbitrary `nextSibling`, one way to accomplish that with the current implementation and semantics is to allow `nextSibling`, but only allow its value to be `null`.

That way, if you don't provide a `nextSibling`, it replaces the content, but it you do, and it's `null`, it appends. Anything else is deferred for a future RFC.

simonihmig:
@wycats I might be misunderstanding you, but are you suggesting that `{{#in-element destination nextSibling=null}}block{{/in-element}}` will append `'block'` to `destination`, while `{{#in-element destination}}block{{/in-element}}` will replace `destination.innerHTML` with `'block'`? 

It seems to me this is different from what it is doing today, which *AFAICT* is to always append, and not replace any existing nodes, at least when looking at these [tests](https://github.com/glimmerjs/glimmer-vm/blob/0e6df5e49ee79c9bb757c1b6b153c10d303f9050/packages/%40glimmer/test-helpers/lib/suites/in-element.ts#L95-L127). But I might very well miss something!?

I would appreciate to not change the semantics of what is already used in the wild with `-in-element` today (again, if I'm not misinterpreting you). Also, with no further defined API around `nextSibling`, it seems a bit weird to me, from a user's perspective, to change the append/replace semantics with a `nextSibling=null`. Something more speaking like `replace=true` or `mode='replace'` seems more appropriate to me.

But actually my suggestion would be as I said earlier to keep `{{#in-element destination}}` always appending, and leave the introduction of a replace mode and the exact specification of an eventual `nextSibling` to a future RFC.

buschtoens:
I might be a bit late to the party, but something like `{{#-before-element}}` and `{{#-after-element}}` would be quite useful. But I guess this would be addressed in a follow-up PR?

wycats:
I'm interested in @chancancode's perspective before pushing any harder on any particular resolution to the `nextSibling` issue.

chancancode:
I'll delegate the technicalities to @wycats since I don't have time to follow the discussion atm 👌

wycats:
@chancancode In that case, I would prefer for this RFC to have replace semantics by default, and append to the end of the element via `nextSibling=null`. Other values passed to `nextSibling` should produce an error.

@cibernox how do you feel about that? If you (and others) are cool with it, can you update the RFC and we can move to FCP!

cibernox:
@wycats while I don't have anything against the idea in principle, that is different from what `{{-in-element}}` does. Since this is a new API it's not technically a breaking change, it is just different.

If you see a technical reason to make this the default instead of the current one, let me know and I'll update the RFC so it can go to FCP soon-ish.

cibernox:
@wycats I added a section with the proposed changes to the API before making it public in which I document the `nextSibling=null` option to restore the current _append-at-the-end_ behavior.

I included your suggestion pristine,but  I'd prefer something more explicit like `replaceContent=false` or something alike. I find `nextSibling=null` somewhat obscure, I don't know what other think about this.

lukemelia:
I agree that `replaceContent` is much clearer than `nextSibling`

wycats:
I'm fine with `replaceContent`, but the reason I chose `nextSibling=null` originally was so that we could eventually allow `nextSibling=value` and not end up with two different ways of saying the same thing (and have to deal with what happens if you say both of them).

cibernox:
@wycats what do you think of renaming `nextSibling` to `insertBefore`? Perhaps it's just me, but since it's also the name of a well known [Node method](https://developer.mozilla.org/en-US/docs/Web/API/Node/insertBefore) it seems easier to diggest to me, even it semantically there is no difference.
@lukemelia thoughts?

cibernox:
Yes, that's the behavior as I've understood it from Yehuda's explanation, I just renamed `nextSibling` to `insertBefore`.

lukemelia:
@wycats @cibernox So how would be characterize the default value of `insertBefore`? Is it `undefined` as distinct from `null`? This seems confusing.

rwjblue:
This seems to have stalled, what needs to happen to get unblocked?

st-h:
I was trying to use the currently private -in-element api to see if I can get rid of ember-elsewhere, however I was running into an issue when making use of `{{-in-element}}` within a route template. We need to provide the `destinationElement` DOM element, however there does not seem to be a good way to query the rendered dom within a route or a controller. So, if there is a route `test` and a corresponding `test/index` route, one is not able to query the document from `test/index` for an element which is rendered in `test`. I think this is basically due to because we do not have a `didInsertElement` equivalent for route-templates.

cibernox:
I didn't get feedback after my last update. I'm fine with merging it as is, but I wanted first to propose the use of `insertBefore ` instead of `nextSibling ` as it seemed a bit more intuitive to me.

rwjblue:
I’d also like to make progress here. I don’t have terribly strong feelings on insertBefore vs nextSibling, maybe @wycats or @chancancode can help out by sharing their thoughts. 

Either way I’ll add to the next core team meetings agenda for discussion...

cibernox:
I'd like to ping with something to say on this, so we push it though the finish line.

knownasilya:
@cibernox what is the consensus on this so far?

cibernox:
The consensus was that it's good to promote it to public API and I just raised a question about wether `nextSibling` felt a bit backwards thinking to me, but I'm not too concerned really, it was just my first impression.

rwjblue:
👋 - Sorry for letting this sit so long. As far as I see it we only have the one issue around `nextSibling` remaining (please correct me if I am wrong). 

> @wycats what do you think of renaming nextSibling to insertBefore? Perhaps it's just me, but since it's also the name of a well known Node method it seems easier to diggest to me, even it semantically there is no difference.

@cibernox - I agree, `insertBefore` seems much better than `nextSibling`. Can you update to that effect?

> @wycats @cibernox So how would be characterize the default value of insertBefore? Is it undefined as distinct from null? This seems confusing.

The default value of `insertBefore` is "not used". We aren't really talking about `undefined` vs `null`, we are instead talking about specifying the parameter or not. This does pose a bit of an issue with YUIDoc style documentation, but I do not think it is inherently confusing on its own.

cibernox:
Updated!

samselikoff:
```hbs
{{#render-to '.modal-wrapper'}}
{{/render-to}}
```

samselikoff:
Love this RFC! Thanks for all your work @cibernox & others.

I know this is a last-minute drive-by 🚲 🏠 but I wanted to throw some alternative names out there to make sure we land with the best one. IMO `in-element` feels a tad bit too low-level.

```hbs
{{#in-element '.modal-wrapper'}}
{{/in-element}}

{{#render-in '.modal-wrapper'}}
{{/render-to}}

{{#render-to '.modal-wrapper'}}
{{/render-to}}

{{#render-inside '.modal-wrapper'}}
{{/render-inside}}
```

I like the idea of someone who's never seen this API before being able to look at a Handlebars template, see this, and instantly understand what's happening (as with `{{if}}`, `{{else}}` etc.). From that perspective `render-in` seems best to me.

rwjblue:
I think re-evaluating the name is perfectly fine (I do like `render-in`), but I also want to point out that this _is_ a fairly low level API so a low level name _might_ be the most appropriate...

samselikoff:
Interesting. (`render-in` is my favorite too).

Do we expect this API to ever show up in application developers' templates?

rwjblue:
TBH, I’m unsure. I do expect the majority user to be via addons that provide nicer interfaces on top (for example `ember-cli-head` or `ember-elsewhere`).

That being said, if we can find a better name that we all agree on we should totally use it...

mitchlloyd:
The name “in-element” strikes me as the right level of abstraction because you do have to think about DOM nodes (where they are) to use this API. Document.querySelector returns an Element so in-element aligns and reinforces browser terminology.

Using the word “render” in a template seems redundant. “{{#render-each}}” and “{{#render-if}}” would more imperative sounding names than “{{#each}}” and “{{#if}}”.

simonihmig:
I agree with the last comments here. This is low-level, and we shouldn't suggest users to use this directly in most cases. What's especially a bit difficult here is that `destinationElement` has to be a DOM element, not a selector (@samselikoff's examples were a bit misleading in that regard). So `in-element` seems ok to me, something like `{{#render-to ".selector"}}` could be built on top of it, in user-space.

st-h:
As a user who was excited about this feature, because it seemed like it could easily replace addons like ember-elsewhere: I tried and failed... I very much second what @simonihmig posted.

cibernox:
I think I also prefer to stick with `{{in-element}}`. Reasons:

- The existing API (that some addons/apps are using despite of being "private") is already called `{{in-element}}`.
- It is somewhat low level. It takes as argument a DOM element, not a selector as ember-wormhole/elsewere do, so the nameing makes sense.

samselikoff:
👍 I’m on board! Just figured I’d throw out the alternative. But I think
you logic makes sense.

rwjblue:
Yep, definitely makes sense. That’s what FCP is for! 😃

cibernox:
@sandstrom that makes sense, although I'm not sure if the RFC should list restrictions in such detail.

btecu:
How about we leave `in-element` private and instead add a public `render-in` that actually takes a selector instead? It seems it would be a lot more valuable for the users.

cibernox:
@btecu this proposes making this low-level primitive public so anyone can create a `{{render-in}}` component in userland relying on public this soon-to-be-public API.

btecu:
@cibernox yes, I can read, thank you. My point is, in most cases `in-element` can't be used directly. It doesn't map to `ember-wormhole` either, so everyone who wants to replace `ember-wormhole` (majority of users) will have to write something on top of `in-element`. Why not get something like `render-in` with a friendlier api?

cibernox:
@btecu Receibing a DOM element is more flexible. Imagine you want to move content to an element that doesn't have ID/class to reference it. This construct could be improved to _also_ take CSS selectors, but taking DOM elements is necessary, and that's why building a new addon (or just update ember-wormhole!) on top of this is the preferred approach.

mehulkar:
To be honest, I like the API of passing in a selector too, all else equal. It seems like a much more recognizable nomenclature to me.

@cibernox what strategies are you imagining for users to get the DOM node to pass in, in the first place? I imagine the majority case will be to use a selector?

promoting an existing API to public isn’t the worst thing in the world, but I think it will eventually evolve to accept selectors anyway, and will be the preferred pattern anyway.

Lastly, do we have some way to talk about “APIs that are meant to be used by addons” vs “APIs meant to be used in apps”?

cibernox:
@mehulkar If you _really_ want to use this API and not use the updated `ember-wormhole` built on top of it, it would be something similar to what I do in ember-basic-dropdown/ember-power-select: https://github.com/cibernox/ember-basic-dropdown/blob/850c227c0a58148056d55d41aa0e5d88656b8165/addon/components/basic-dropdown/content.js#L70-L73

rwjblue:
I don't think that the presence of good primitives (`{{in-element}}`) prevents us from _also_ adding a more sugary layer on top (`{{render-in}}`). Its just a pretty different thing, and deserves an RFC of its own. The good thing is that once this RFC lands we can actually do most of the experimentation on what works well in addon space.

In the meantime, it is pretty easy to get a nice selector based setup working well with only a simple helper:


```js
// app/helpers/query-selector.js
export Ember.Helper.helper(function([selector]) {
  return document.querySelector(selector);
});
```

```hbs
{{#in-element (query-selector 'something')}}

{{/in-element}}
```

mehulkar:
Looks like that still uses an ID selector :) (just responding to your comment about DOM node being necessary in case you wanted to move content to a node that doesn’t have a class/ID reference). The only thing I can imagine for that is this._element in components (which won’t be valid with tagless components anyway)

cibernox:
@mehulkar you're totally welcomed to open an RFC once this has been merged to make `in-element` **also** accept strings.

btecu:
@rwjblue Nothing was really preventing addons from experimenting with the private api, as @cibernox linked his use case. I believe a bit of _sugary layer_ would make a huge difference, especially for new users, otherwise they would have to learn how to create their own helper or adhoc solution to use this feature.

rwjblue:
> Nothing was really preventing addons from experimenting with the private api

The blockers are fairly clear actually:

1. lack of documentation
2. risk of using private APIs that have no SemVer guarantee

Clearly from @cibernox's example (and others) _some_ addon authors have discovered the "magical incantation" and are willing to accept the risk, but that is far from "nothing". 😸 

> I believe a bit of sugary layer would make a huge difference, especially for new users, otherwise they would have to learn how to create their own helper or adhoc solution to use this feature.

Yep, as mentioned before, I totally agree and is a nice incremental (concept wise) step forward. There is no reason to avoid adding the lower level API now, and adding a better high level API later.

chancancode:
To be very clear – we don't think anyone should be replacing the addons they are using today with this new API (perhaps the "Motivation" part of the RFC should be more clear about this). The only goal is that there are addons that uses this today and we would like to offer them more stability guarantees (and transitively, to the users of those addons) than today.

An API that takes DOM elements are more "primitive" than selectors in several sense:

- This is a Glimmer VM feature which has to work in Fastboot environments that does not have `querySelector` etc
- As other pointed out selectors only work for things that are addressable (with ID, unique classes, etc) and doesn't work for, e.g. elements that are detached from DOM (maybe you want to render into an element an element outside of the DOM tree to collect the `innerHTML`, etc)
- Edge cases like: what if the selector refers to more than one element, or that the element pointed to by the selector changes

100% agree that this API is not convenient enough for end-users to adopt, but that's why this is a primitive API for other addons to build on top.

rwjblue:
We discussed this at todays core team meeting, and are still in favor of moving forward. We are also very much interested in further RFCs in this space (e.g. possibly a more user-facing proposal like `{{render-in 'some-selector'}}`).

Thanks to everyone who participated in this RFC!

0xadada:
I just used `{{-in-element}}` to implement a modal window component, and this would give me that warm fuzzy feeling about putting it into multiple production projects.

PR #289, opened 2017-12-28
Consume node_modules with Ember-CLI

rtorr:
This RFC needs help, but wanted to help get the ball rolling. This focuses on the user experience, more than the what needs to change in Ember's internals (that will take a bit more time and others are probably more suited).

### DIFF
--- a/text/0000-node-modules.md
+++ b/text/0000-node-modules.md
@@ -1,0 +1,110 @@
+* Start Date: 2017-12-17
+* RFC PR: (leave this empty)
+* Ember Issue: (leave this empty)
+
+# Node Modules
+
+## Summary
+
+Introduce the ability to consume packages from node_modules, without wrapping it in a plug-in.
+
+```bash
+npm install cool-thing
+```
+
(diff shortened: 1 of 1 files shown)


Panman8201:
ember-cli 2.15.0 enabled this functionality, no? I was just starting to move an addon over to this, instead of using bower deps.

https://github.com/ember-cli/ember-cli/releases/tag/v2.15.0
https://github.com/ember-cli/ember-cli/pull/7045

If anything, I don't think it's really documented at all...

rtorr:
I realistically don’t have time to take this on right now =(

wycats:
@Panman8201 The goal of this RFC **was** to make importing from `node_modules` "just work" without further steps.

The main blocker for that kind of thing in today's Ember is making it possible for Ember CLI to track what modules are being used. I believe that the CLI team is working on that.

I enumerated some of the relevant questions that would need to be answered in this Twitter thread, especially around CJS modules:

https://twitter.com/wycats/status/946176381095895040

PR #290, opened 2017-12-28
headings- case consistency

MelSumner:
- Changed an instance of title case to sentence case for consistency.


andreafrost:
@MelSumner looks like the same inconsistencies are here: https://github.com/emberjs/rfcs/blob/master/deprecation-template.md

PR #292, opened 2018-01-05
Setup mdbook

sivakumar-kailasam:


.travis.yml
```diff
@@ -0,0 +1,17 @@
+language: rust
+script:
+  - rev=$(git rev-parse --short HEAD)
+  - (cargo install mdbook --vers 0.0.26 --force || true)
```
buschtoens: Why `|| true`? If this step fails, wouldn't the CI script fail on the next line, since `mdbook` might not be available?

sivakumar-kailasam: We copied this over from the rust RFC repo. Unsure about this line's significance

buschtoens: Thanks 👌

PR #294, opened 2018-01-12
Make jQuery optional

simonihmig:
[Rendered](https://github.com/simonihmig/rfcs/blob/jquery-optional/text/0000-optional-jquery.md)

### DIFF
--- a/text/0000-optional-jquery.md
+++ b/text/0000-optional-jquery.md
@@ -106,6 +106,56 @@ production), that provide some useful error messages for the developer:
   should throw an assertion stating that jQuery is not available and that `this.element` and native DOM APIs should be
   used instead.
 
+### Introducing `ember-jquery-legacy` and deprecating `jQuery.Event` usage
+
+Event handler methods in components will usually receive an instance of [`jquery.Event`](http://api.jquery.com/category/events/event-object/) as an argument, which is very
+similar to native event objects, but not exactly the same. To name a few differences, not all properties of the native 
+event are mapped to the jQuery event, on the other hand a jquery event has a `originalEvent` property referencing the
+native event. 
+
+The updated event dispatcher in Ember 3.0 is capable of working without jQuery (similar to what 
+`ember-native-dom-event-dispatcher` provided for Ember 2.x). When jQuery is not available, it will naturally not be 
+able to pass a `jquery.Event` instance but a native event instead. This creates some ambiguity for addons, as they 
+cannot know in advance how the consuming app is built (with or without jQuery).
+
+For code that does not rely on any `jQuery.Event` specific API, there is no need to change anything as it will continue
+to work with native DOM events. 
+
+But there are cases where jQuery specific properties have to be used (when jQuery events are passed). This is especially
+true for the `originalEvent` property, for example to access `TouchEvent` properties that are not exposed on the 
+`jQuery.Event` instance itself. So there has to be a way to make the code work with either jQuery events or native 
+events being passed to the event handler (especially important for addons). Moreover this should be done in a way that 
+uses native DOM APIs only, to support the migration away from jQuery coupled code.
+
+To solve this issue another addon `ember-jquery-legacy` will be introduced, which for now will only expose a single
+`normalizeEvent` function. This function will accept a native event as well as a jQuery event (possibly distinguishing
+between those two modes at build time, based on the jQuery integration flag), but will always return a native event 
+only. 
+
+This will allow addon authors to work with both event types, but start to only use native DOM APIs:
+
+```js
+import Component from '@ember/component';
+import { normalizeEvent } from 'ember-jquery-legacy';
+
+export default Component.extend({
+  click(event) {
+    let nativeEvent = normalizeEvent(event);
+    // from here on use only native DOM APIs...
+  }
+})
+```
+
+To encourage addon authors to refactor their jQuery coupled event code, the use of `jQuery.Event` specific APIs used for
+jQuery events passed to component event handlers should be deprecated and a deprecation message be shown when accessing 
+them (e.g. `event.originalEvent`). Care must be taken though that this warning will not be issued when `normalizeEvent` 
+has to access `originalEvent`. 
+
+Also for apps that do not want to transition away from jQuery and would be overloaded with unnecessary warnings, the 
+deprecations should be silenced when the jQuery integration flag is explicitly set to true (and not just true by 
+default). By doing so users effectively state their desire to continue using jQuery, thus any needless churn should
+be avoided for them.
+
 ### Testing
 
 Ember's test harness has been based on jQuery for a long time. Most global acceptance test helpers like `find` or 
@@ -171,6 +221,12 @@ The section on component tests should not use `this.$()` anymore as well, and in
 use `this.element` to refer to the component's root element, and use the new DOM interaction helpers instead of jQuery 
 events triggered through `this.$()`. 
 
+### Deprecation guide
+
+The deprecation warnings introduced for using `jQuery.Event` specific APIs should explain the use of the 
+`normalizeEvent` helper function to migrate towards native DOM APIs on the one side, and on the other side the effect of 
+setting the jQuery integration flag to explicitly opt into jQuery usage thus suppressing the warnings.
+
 ### Addon migration
 
 One of the biggest problems to easily opt-out of jQuery is that many addons still depend on it. Many of these usages 
@@ -221,30 +277,4 @@ Continue to depend on jQuery.
 
 ## Unresolved questions
 
-### jquery.Event vs. native events
-
-Event handler methods in components will usually receive an instance of `jquery.Event` as an argument, which is very
-similar to native event objects, but not exactly the same. To name a few differences, not all properties of the native 
-event are mapped to the jQuery event, on the other hand a jquery event has a `originalEvent` property referencing the
-native event. 
-
-The updated event dispatcher in Ember 3.0 is capable of working without jQuery (similar to what 
-`ember-native-dom-event-dispatcher` provided for Ember 2.x). When jQuery is not available, it will naturally not be 
-able to pass a `jquery.Event` instance but a native event instead. This creates some ambiguity for addons, as they 
-cannot know in advance how the consuming app is built (with or without jQuery).
-
-* should addon authors be encouraged to add checks for the type of event (like `event instanceof jQuery.Event`) if 
-they rely on jQuery event specifics? (which is already the case when they just need access to the native event 
-`originalEvent`)
-* should we aim to remove that ambiguity and pass only native events at some time, regardless of jQuery being available
-or not? This would be a breaking change, so can be done for Ember 4.0 only. If not, the type checks mentioned above 
-would have to stay around basically forever
-* for the above case, one *could* add deprecation warnings when jQuery specific properties like `originalEvent` are 
-accessed. However this can be misleading, as it may be required to use these jQuery specifics if a jQuery event is 
-passed, so there is technically no other way to prevent the deprecation warning to be triggered. 
-* is there a way to start passing *only* native events, which are augmented by jQuery specific properties like
-`originalEvent` (e.g. `event.originalEvent = event`) to not break things, thus following SemVer while still 
-transitioning to native events only. This would allow all addons to assume native events without any type checks. 
-Something similar seems to have been discussed for the 
-[jQuery 4.0 Event Design](https://github.com/jquery/jquery/wiki/jQuery-4.0-Event-Design#avoid-the-need-for-a-jqueryevent-wrapper),
-but seems to be controversial and possibly error prone.
+None so far.
--- a/text/0000-optional-jquery.md
+++ b/text/0000-optional-jquery.md
@@ -1,0 +1,248 @@
+- Start Date: 2018-01-11
+- RFC PR: (leave this empty)
+- Ember Issue: (leave this empty)
+
+# Make jQuery optional
+
+## Summary
+
+For the past Ember has been relying and depending on jQuery. This RFC proposes making jQuery optional and having a well 
+defined way for users to opt-out of bundling jQuery. 
+
+## Motivation
+
+### Why we don't need jQuery any more
+
+One of the early goals of jQuery was cross-browser normalization, at a time where browser support for web standards was 
+incomplete and inconsistent, and Internet Explorer 6 was the dominating browser. It provided a useful and convenient
+API for DOM traversal and manipulation as well as event handling, that hid the various browser differences and bugs from
+the user. For example `document.querySelector` wasn't a thing at that time, and browsers were using very different event 
+models ([DOM Level 0, DOM Level 2 and IE's own proprietary model](https://en.wikipedia.org/wiki/DOM_events#Event_handling_models)). 
+
+But this level of browser normalization is not required anymore, as today's browsers all support the basic DOM APIs well
+enough. Even more so that the upcoming Ember 3.0 will drop support for all versions of Internet Explorer except 11.
+
+Furthermore Ember users will need to directly traverse and modify the DOM or manually attach event listeners in very 
+special cases only. Most of these low level interactions are taken care of by Ember's templates and its underlying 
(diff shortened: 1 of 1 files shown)


rwjblue:
Big 👍  here, thank you for taking the time to work on it!

text/0000-optional-jquery.md
```diff
@@ -0,0 +1,248 @@
...
+
+### Introduce `@ember/jquery` package
+
+Currently Ember CLI itself is importing jQuery into the app's `vendor.js` file. To decouple it from this task, and
+to allow for some better flexibility in the future, the responsibility for importing jQuery is moved to a dedicated 
+`@ember/jquery` addon.
+
+To not create any breaking changes, Ember CLI will have to check the app's dependencies for the presence of this addon. 
+If it is not present, it will continue importing jQuery *unless* the jQuery integration flag is disabled.
+In the other case, it will stop importing jQuery at all, and delegate this responsibility to the addon.
```
Gaurav0: If it is present,

text/0000-optional-jquery.md
```diff
@@ -0,0 +1,248 @@
...
+
+To nudge users to install `@ember/jquery` when they need jQuery, some warning/deprecation messages should be issued when
+the addon is *not* installed and the integration flag is either not specified or is set to true. Only in the case the 
+app is actively opting out of jQuery integration the addon is not needed.
```
Gaurav0: The `@ember/jquery` addon should be placed in the default blueprint for a period of time so as to ease migration.

text/0000-optional-jquery.md
```diff
@@ -0,0 +1,248 @@
...
+
+Apart from testing (see below), Ember features some APIs that directly expose jQuery, which naturally cannot continue 
+to work without it. For these APIs some assertions have to be added when running in "no jQuery" mode (and not in 
+production), that provide some useful error messages for the developer:
+
+* `Ember.$()`
+  should throw an assertion stating that jQuery is not available.
+* `this.$()` in components
+  should throw an assertion stating that jQuery is not available and that `this.element` and native DOM APIs should be
+  used instead.
```
Gaurav0: Should these throw deprecation warnings before jquery is removed?

text/0000-optional-jquery.md
```diff
@@ -0,0 +1,248 @@
...
+
+Ember CLI and the `@ember/jquery` addon will also look for `jquery-integration` in this configuration file, and will 
+opt-out of importing jQuery when this file is present and the flag is set to `false`.
+
+## How we teach this
+
+### Guides
+
+The existing "Managing Dependencies" chapters in the Ember Guides as well as on ember-cli.com provide a good place to 
+explain users how to set the jQuery integration flag by means of the mentioned addon that handles this flag.
```
Gaurav0: Clarify this is not the `@ember/jquery` addon

text/0000-optional-jquery.md
```diff
@@ -0,0 +1,248 @@
...
+
+### Addon migration
+
+One of the biggest problems to easily opt-out of jQuery is that many addons still depend on it. Many of these usages 
+seem to be rather "accidental", in that the full power of jQuery is not really needed for the given task, and could 
+be fairly easily refactored to use only native DOM APIs.
+
+For this reason this RFC encourages addon authors to not use jQuery anymore and to refactor existing usages whenever 
+possible! This certainly does not apply categorically to all addons, e.g. those that wrap jQuery plugins as 
+components and as such cannot drop this dependency.
```
Gaurav0: Addons that require jquery should state that they do so.

At some point the end goal should be that such addons could simply include `@ember/jquery` in their dependencies and not have additional instructions.

text/0000-optional-jquery.md
```diff
@@ -0,0 +1,248 @@
...
+
+A good amount of that churn can be mitigated by having a codemod that migrates tests (see "Testing" above).
+
+## Alternatives
+
+Continue to depend on jQuery.
+
+## Unresolved questions
+
+### jquery.Event vs. native events
```
Gaurav0: In my opinion, this question needs to be somewhat resolved before this RFC can proceed, as it pretty much derails the whole show.

text/0000-optional-jquery.md
rwjblue: Yep, absolutely agreed.

text/0000-optional-jquery.md
rwjblue: No, I don't think so. Importantly, this RFC is _not_ suggesting that usage of jQuery specific features are deprecated.  It's providing for a way to opt-out of that feature set which is coherent.

There certainly may be an additional deprecation RFC down the road, but for now we are trying to ensure that either option (using jQuery integration features or _not_ including jQuery) are viable coherent options for Ember apps...

text/0000-optional-jquery.md
rwjblue: Yes, great point. Currently we are thinking that it will be `@ember/optional-features` ([WIP repo here](https://github.com/emberjs/ember-optional-features))...

text/0000-optional-jquery.md
rwjblue: > Addons that require jquery should state that they do so.

Totally agreed! The current line of thinking is that addons that require jQuery (or any of the other optional features that we are adding opt-out capabilities for) can as `@ember/optional-features` if any given feature is enabled or disabled, this allows an addon that is (for example) wrapping a jQuery plugin ensure that jQuery integration is enabled (and `@ember/jquery` is present properly). 

> At some point the end goal should be that such addons could simply include @ember/jquery in their dependencies and not have additional instructions.

_Possibly_, but for now I think that we want to leave control of opting in or out of features in the hands of the application specifically.

text/0000-optional-jquery.md
```diff
@@ -0,0 +1,248 @@
...
+they rely on jQuery event specifics? (which is already the case when they just need access to the native event 
+`originalEvent`)
+* should we aim to remove that ambiguity and pass only native events at some time, regardless of jQuery being available
+or not? This would be a breaking change, so can be done for Ember 4.0 only. If not, the type checks mentioned above 
+would have to stay around basically forever
+* for the above case, one *could* add deprecation warnings when jQuery specific properties like `originalEvent` are 
+accessed. However this can be misleading, as it may be required to use these jQuery specifics if a jQuery event is 
+passed, so there is technically no other way to prevent the deprecation warning to be triggered. 
+* is there a way to start passing *only* native events, which are augmented by jQuery specific properties like
+`originalEvent` (e.g. `event.originalEvenet = event`) to not break things, thus following SemVer while still
```
rwjblue: typo: s/originalEvenet/originalEvent/

text/0000-optional-jquery.md
rwjblue: Yep, I agree that we need to reach a consensus on the best path forward here.

rwjblue:
RE: The `jQuery.Event` vs native event situation, I think we have additional options that are not listed. The simplest solution that comes to my mind is:

Create a small addon that detects if jQuery integration is enabled or disabled (at build time), and provide a consistent interface to always convert to a native event.  Something like:

```js
// jQuery enabled version
export default function(jQueryEvent) {
  return jQueryEvent.originalEvent;
}
```

```js
// jQuery disabled version
export default function(event) { 
  return event;
}
```

This utility function could be used where needed like:

```js
import normalizeEvent from '@ember/some-awesome-name';

  click(event) {
    let nativeEvent = normalizeEvent(event);
    // do stuff...
  }
})
```

I think there are a number of variations of this setup that would work nicely...

text/0000-optional-jquery.md
simonihmig: Exactly. I had a second follow up RFC planned, that is more aggressive (and thus maybe controversial) about pushing away from jQuery, including deprecations as mentioned. But this would be *complimentary* to this one, so basically the second step (if adopted).

text/0000-optional-jquery.md
simonihmig: > If it is present,

That's what I meant with _In the other case_, but I can rephrase this.

simonihmig:
> Create a small addon that detects if jQuery integration is enabled or disabled (at build time), and provide a consistent interface to always convert to a native event.

Love it! 

I guess this would also allow to add some deprecation messages when accessing `event.originalEvent` directly, as mentioned in the chapter's 3rd bullet point. With some shenanigans we could make sure that this deprecation is not triggered when accessing this through the `normalizeEvent` function.

This would also allow the move to only pass native events (would be a breaking change) painless, as addons can be refactored immediately to only use native event semantics, even when jQuery events are still passed for now.

simonihmig:
Just to give some more context, as this has been touched here: https://github.com/emberjs/rfcs/pull/294#discussion_r161353291...

This RFC is about providing an officially supported way to opt-out of jQuery. It is a non breaking change, and so far no deprecations are planned for this!

There will probably be a second follow up RFC, that is more aggressive (and thus maybe more controversial) about pushing away from jQuery, including deprecating jQuery coupled APIs like `this.$()`. Basically it would change the semantics from this one's "jQuery included by default, allow opt-out" to "removed by default, allow opt-in".

But that would be complimentary to this one and would obviously require the opt-out proposed here to be already possible, so it would basically be the second step (if adopted).

dbbk:
I'm not sure enabled by default is the best approach. Also, I'm a bit confused, does this mean there would be two addons - one called something like "ember-disable-jquery" and one for "@ember/jquery"? I think it'd be much more elegant if jQuery was disabled by default and users could simply install `@ember/jquery`, and tools like `ember-cli-update` could do that automatically.

simonihmig:
> does this mean there would be two addons - one called something like ember-disable-jquery

Managing the opt-out flag would be the responsibility of a dedicated addon, presumably ` ember-optional-features`, which is not restricted to jQuery. See [implementation outline](https://github.com/simonihmig/rfcs/blob/jquery-optional/text/0000-optional-jquery.md#implementation-outline)

> I think it'd be much more elegant if jQuery was disabled by default

I am also in favor of being more aggressive about moving away from jQuery *at some point in time*, but as I said that could be the subject of another follow up RFC. To really remove the existing jQuery support from `ember-source` (like `this.$()`) would be a breaking change, so can be done at Ember 4.0 at the earliest (with prior deprecation). Enabled by default is what we have today, and this RFC just opens the door to disable...

What we probably could do earlier is to opt-out by default in the default app blueprint, which would affect *new*  apps, but would not break things for existing ones. But even that I think we should do carefully, when the most popular addons are known to work fine without jQuery, to not create a frustrating developer experience.

rwjblue:
> I think it'd be much more elegant if jQuery was disabled by default and users could simply install @ember/jquery, and tools like ember-cli-update could do that automatically.

I agree that is the longer term plan, but again this RFC is solely focused on _allowing_ opting out to begin with. I believe as part of that plan, the default addon blueprint will begin testing without jQuery as well as with it and that will help make using Ember without jQuery easier and easier as addons update to absorb that change. 

I absolutely agree with the goal of having `ember new foo` simply not use jQuery at all, we just need to move with care. 😸

dbbk:
By 'elegant' I meant it seems strange to me that a user would have to _install_ something to _remove_ something else. I agree that jQuery should be provided out of the box by default for now. But I think conceptually it doesn't make sense to have "lack of jQuery" being a new dependency.

rwjblue:
> By 'elegant' I meant it seems strange to me that a user would have to install something to remove something else. 

Gotcha, sorry for misunderstanding!

> But I think conceptually it doesn't make sense to have "lack of jQuery" being a new dependency.

Considering this RFC alone, I would completely agree with you. However, the actual intention here is to have a single `@ember/optional-features` addon that allows opting out of a number of currently defaulted features ([application-template-wrapper](https://github.com/emberjs/rfcs/blob/master/text/0280-remove-application-wrapper.md) and [template-only-component-wrapper](https://github.com/emberjs/rfcs/blob/master/text/0278-template-only-components.md) are the other two in addition to this one which would be `jquery-integration`). As features continue to be made optional (ultimately including being stripped from the build when not enabled) in Ember itself this list will grow even larger....

simonihmig:
I just added a [new section](https://github.com/simonihmig/rfcs/blob/jquery-optional/text/0000-optional-jquery.md#introducing-ember-jquery-legacy-and-deprecating-jqueryevent-usage) with a possible solution to the `jQuery.Event` vs. native issue, based on @rwjblue's suggestion [here](https://github.com/emberjs/rfcs/pull/294#issuecomment-357402503)! 

Please review! 🧐

text/0000-optional-jquery.md
```diff
@@ -0,0 +1,280 @@
...
+But there are cases where jQuery specific properties have to be used (when jQuery events are passed). This is especially
+true for the `originalEvent` property, for example to access `TouchEvent` properties that are not exposed on the 
+`jQuery.Event` instance itself. So there has to be a way to make the code work with either jQuery events or native 
+events being passed to the event handler (especially important for addons). Moreover this should be done in a way that 
+uses native DOM APIs only, to support the migration away from jQuery coupled code.
+
+To solve this issue another addon `ember-jquery-legacy` will be introduced, which for now will only expose a single
+`normalizeEvent` function. This function will accept a native event as well as a jQuery event (possibly distinguishing
+between those two modes at build time, based on the jQuery integration flag), but will always return a native event 
+only.
```
Gaurav0: Will it be necessary for ember-jquery-legacy to be a dependency or can it be just a devdependency?

text/0000-optional-jquery.md
```diff
@@ -0,0 +1,280 @@
...
+    let nativeEvent = normalizeEvent(event);
+    // from here on use only native DOM APIs...
+  }
+})
+```
+
+To encourage addon authors to refactor their jQuery coupled event code, the use of `jQuery.Event` specific APIs used for
+jQuery events passed to component event handlers should be deprecated and a deprecation message be shown when accessing 
+them (e.g. `event.originalEvent`). Care must be taken though that this warning will not be issued when `normalizeEvent` 
+has to access `originalEvent`.
```
Gaurav0: This may require passing something that extends `jquery.Event`.

text/0000-optional-jquery.md
```diff
@@ -0,0 +1,280 @@
...
+(minified and gzipped). This not only increases the loading time, but also parse and compile times, thus increasing the 
+total time to interactive. This is especially true for mobile devices, where slow connectivity and weak CPU performance 
+is not uncommon.
+
+Having jQuery not included will improve the suitability of Ember for mobile applications considerably. Even
+if the raw number is not that huge, it all adds up. And it plays together with other efforts to make leaner Ember builds
+possible, like enabling tree shaking with the new [Module API](https://github.com/emberjs/rfcs/blob/master/text/0176-javascript-module-api.md),
+moving code from core to addons (e.g. the [`Ember.String` deprecation](https://github.com/emberjs/rfcs/blob/master/text/0236-deprecation-ember-string.md))
+or the ["Explode RFC"](https://github.com/emberjs/rfcs/blob/explode/text/0000-explode.md). In that regard removing the 
+dependency on jQuery is a rather low hanging fruit with an high impact.
```
Gaurav0: Considering the implementation complexity, describing this as "low hanging fruit" seems ridiculous.

text/0000-optional-jquery.md
simonihmig: Will have to be a dependency (for an addon), as `normalizeEvent` needs to be imported in actual code, see the little code example! Only needed though if you currently require any jQuery specifics like `originalEvent`.

text/0000-optional-jquery.md
simonihmig: Which implementation complexity are you referring to? For Ember itself, most of the hard work has already been done, what is proposed here should not be too hard to implement. 

If you refer to user's apps, sure that might involve a lot of work to get rid of jQuery depending on the amount of jQuery code (so "your milage may vary"). However nobody forces you to do so, and for apps with less of an jQuery legacy, removing it could be a big and rather easy win!

rwjblue:
We discussed this at the core team meeting on 2018-01-19, and believe it is ready for final comment period.

Making it so...

rwjblue:
Thanks to all those that helped push this forward! There have been no new concerns brought up while this has been in the final comment period, and the core team is still very excited to move forward with it.

✂️

PR #297, opened 2018-01-19
Deprecating Ember.Logger

lupestro:
This is the RFC for the work to deprecate Ember.Logger.

[Rendered](https://github.com/lupestro/rfcs/blob/lupestro-deprecate-logger/text/0297-deprecate-ember-logger.md).

### DIFF
--- a/text/0297-deprecate-ember-logger.md
+++ b/text/0297-deprecate-ember-logger.md
@@ -1,6 +1,6 @@
 - 2018-01-17
 - RFC PR: 0297
-- Ember Issue: (leave this empty)
+- Ember Issue: https://github.com/emberjs/ember.js/issues/16231
 
 # Deprecation of Ember.Logger
 
--- a/text/0297-deprecate-ember-logger.md
+++ b/text/0297-deprecate-ember-logger.md
@@ -75,8 +75,13 @@ affect this behavior.
 ### Codemod 
 
 Provide a codemod that developers can use to switch references to the methods 
-of `Ember.Logger` to use the corresponding `console` methods instead. I 
-will definitely need help here.
+of `Ember.Logger` to use the corresponding `console` methods instead. 
+
+The codemod will need to replace `Ember.Logger.debug` calls with `(console.debug || 
+console.log)(<arguments>)` or something similar to take in stride the situation 
+where console.debug isn't defined.
+
+I will definitely need help here.
 
 ### Add-On Developers
 
@@ -162,29 +167,4 @@ a shim for users.
 
 ## Unresolved questions
 
-### How do we deal with `Logger.debug` in the codemod? 
-
-Do we provide separate options for those who might use Node versions 
-earlier than 9 and those who are confident they will only use Node 
-version 9 or later? 
-
-Do we have the codemod inject a polyfill for `console.debug` that calls 
-`console.log`? 
-
-Do we provide one separately for the user to apply? 
-
-Or does that become a fastboot concern, since that's the primary driver 
-for running ember projects in node?
-
-### What do we do about the eslint `no-console` flag? 
-
-Some developers are using `Ember.Logger` right now to work around it, while 
-others are providing their own logging service, and others are just providing 
-override comments. 
-
-Currently, `no-console` is in the default set of flags set with eslint, 
-and ember-cli doesn't override the defaults in its set of flags to turn it off. 
-
-Should we turn it off in the ember-cli settings? Or should that be left as a user decision? 
-
-I'm inclined toward the latter. 
+None at this point. The answers from prior drafts have been promoted into the text. 
--- a/text/0297-deprecate-ember-logger.md
+++ b/text/0297-deprecate-ember-logger.md
@@ -1,5 +1,5 @@
 - 2018-01-17
-- RFC PR: (leave this empty)
+- RFC PR: 0297
 - Ember Issue: (leave this empty)
 
 # Deprecation of Ember.Logger
--- a/text/0297-deprecate-ember-logger.md
+++ b/text/0297-deprecate-ember-logger.md
@@ -26,19 +26,10 @@ lighter.
 
 For the most part, this is a 1:1 substitution of the global `console` object 
 for `Ember.Logger`. However, Node only added support for `console.debug` in 
-Node version 9. To support earlier versions of Node, our codemod will need to 
-use `console.log`, rather than `console.debug`, as the replacement for 
-`Logger.debug`, except for add-ons that specify that they only support Node 
-version 9 and beyond.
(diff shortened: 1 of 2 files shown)


mmun:
I support this change. Happy to help implement as well.

ef4:
This is good, and it looks like a nicely separable piece of work that doesn't require any deep experience with ember internals.

text/0297-deprecate-ember-logger.md
```diff
@@ -170,15 +161,24 @@ a shim for users.
...
-codemod inject a polyfill for console.debug that calls console.log? Do we
-provide one separately for the user to apply? Or does that become a fastboot 
-concern, since that's the primary driver for running ember projects in node.
-
-What do we do about the eslint `no-console` flag? Some developers are 
-using `Ember.Logger` right now to work around it. Others are providing 
-their own logging service. If we encourage the use of a Babel plugin 
-to strip console calls, should we remove `no-console` from the default 
-flags that `ember-cli` ships?
+How do we deal with `Logger.debug` in the codemod?
```
locks: A suggestion: Try making the question a sub-header and not use bullet points for all of the answer.

text/0297-deprecate-ember-logger.md
```diff
@@ -70,9 +61,15 @@ Note: None of the uses of `Ember.Logger` in `ember.js` or `ember-data` involve
...
 
 Add deprecation warnings to the implementation: `ember-console\lib\index.js`.
-Bear in mind that the deprecation mechanism currently calls `Logger.warn`, so 
-that code should be changed _first_ or this change will be very difficult 
-to debug.
+Bear in mind that the deprecation mechanism in `ember-debug` currently calls 
+`Logger.warn`, so the `ember-debug` code should be changed _first_ or adding 
+the deprecation warning will create a deep recursion.
+
+The log and debug messages emitted by Ember for deprecations and assertions
```
locks: You might want to slightly rephrase this section to specifically indicate that it is `Ember.warn` and friends that are noops in production.

paulyoder:
What happens in the situation where the `Ember.Logger` methods are wrapped in another function to provide additional functionality?

For example, my app has an initializer that sends all log messages to Rollbar. Here is what it looks like for the `Ember.Logger.info` method:

```
let infoLogger = Ember.Logger.info;
Ember.Logger.info = function() {
  Rollbar.info.apply(Rollbar, arguments);
  infoLogger.apply(this, arguments);
};
```

How would I change this code in light of this RFC? Would I just override `console.info` directly then?

bendemboski:
@paulyoder I would think you'd just override `console.info` directly -- that's what many libraries that don't depend on a framework like Ember do, e.g. TrackJS.

bendemboski:
I think this makes sense. As far as the eslint question, I personally wouldn't want to turn off `no-console` because I frequently scatter `console.log()`s and `console.count()`s and whatnot in my code while I'm debugging something, and want `eslint` to remind me to remove them before checking in the code. So I would opt for something like this:

```javascript
no-console: ["error", { allow: ["info", "warn", "error"] }]
```

This gives me a good set of options for logging to the console, but also allows me to use `console.log()` or `console.debug()` (and various other console methods) with the eslint safety net.

Personally, I'd advocate for something like ☝️ over just disabling `no-console` in `ember-cli`'s blueprints, although I'm uncertain of the merits of doing ☝️ over just leaving it on and letting the user decide.

webark:
I think that separate from this RFC, it would be nice to have another RFC that deals with the "debugging and logging story" of ember. I think there is where a lot of these kind of issues could come out, and you could tap into some parts of ember and other extensible tools to get a richer logging and debugging experience.

locks:
@webark send that RFC over ;)

rwjblue:
@lupestro - Great job putting this together!!

I only have a few responses to the unanswered questions section:

> How do we deal with Logger.debug in the codemod?

I think replacing it with `(console.debug || console.log)(<whatever arguments>)` would be perfectly fine (and be an accurate codemod). I think `Logger.debug` is fairly rare, and having the codemod roughly guarantee existing functionality is a good default...

> What do we do about the eslint no-console flag?

I think we do nothing. Leaving `Logger.log`'s strewn about a codebase just as a way to work around the `no-console` rule is much worse than simply adding `// eslint-disable-line no-console` (the codemod _could_ do this for the user automatically also).

rwjblue:
We discussed this at the core team meeting today, and we are general happy to move forward on this RFC. Since it seems like conversation has largely settled here, we are moving it into the final comment period.

akatz:
Rendered?: https://github.com/lupestro/rfcs/blob/cbef6c30cc6915c57ba116597c98a7c2ab864c69/text/0297-deprecate-ember-logger.md

machty:
I'm the proud maintainer of the addon with the most uses of `Ember.Logger` (ember-concurrency), and technically e-c supports Embers all the way back to 1.13. If I want to continue that long-term support I'll have to have some kind of code that fulfills a similar purpose as Ember.Logger; what's the proper strategy for me going forward?

lupestro:
Currently, the best recommendation I can think of is to build a small wrapper of your own, preferably as a service suitable for dependency injection, possibly even available to your clients for interception. (Ember.Logger itself has something like a single functional line for each kind of console call it performs, so your wrapper shouldn't need to be any bigger.) 

I'm hoping other reviewers will have a better suggestion. I'll make sure the best suggestion makes it into the deprecation guide, since this is a question that is likely to be asked a lot. Should I say anything in the RFC text at this late date?

MelSumner:
Was the option 2 considered as a bridge, or will that just prolong the pain?

> Extract Ember.Logger into its own (tiny) @ember/console package as a shim for users.

rwjblue:
@MelSumner - I think that is likely fine but under a different name (e.g. `@ember/legacy-logger` or somesuch).

@machty - I believe that the answer here is more situational (e.g. it depends how an addon uses `Ember.Logger`). In the ember-concurrency cases, I have specific suggestions below.

* [Warning for not using maxConcurrency](https://github.com/machty/ember-concurrency/blob/9b8f0a386671ff7b6e62d3cd62c064b408c5e9ac/addon/-task-property.js#L436) - This should be swapped for `deprecate` (the message explicitly states that it is a deprecation). This change is backwards compatible and avoids any deprecations proposed by this RFC.
* [Logging the cancellation reason when manually debugging](https://github.com/machty/ember-concurrency/blob/9b8f0a386671ff7b6e62d3cd62c064b408c5e9ac/addon/-task-instance.js#L329) - I would change this usage to directly invoke `console.log`. This would mean that someone could not debug manually on IE < 11, but that seems like an extremely unlikely scenario...

PR #298, opened 2018-01-19
Lint Component Lifecycle Events

rwwagner90:
[Rendered](https://github.com/rwwagner90/rfcs/blob/deprecate-component-lifecycle-events/text/0298-deprecate-component-lifecycle-events.md)

### DIFF
--- a/text/0000-deprecate-component-lifecycle-events.md
+++ b/text/0000-deprecate-component-lifecycle-events.md
@@ -1,4 +1,4 @@
-- Start Date: (fill me in with today's date, 2018-01-19)
+- Start Date: 2018-01-19
 - RFC PR: (leave this empty)
 - Ember Issue: (leave this empty)
 
@@ -27,6 +27,17 @@ The exact deprecation message will be decided later, but something along the lin
 Using `Ember.on` event listeners for component lifecycle hooks is deprecated.
 Please move your code from `myHandler: Ember.on('didRender'` into the `didRender(){}` method.
 ```
+
+Ember does not appear to rely on these events internally, however there are at least a couple of mentions of 
+these events in the comments, referring specifically to `on('didInsertElement')` 
+[here](https://github.com/emberjs/ember.js/blob/master/packages/ember-metal/lib/run_loop.js#L140) 
+and [here](https://github.com/emberjs/ember.js/blob/master/packages/ember-runtime/lib/ext/function.js#L149) 
+We will want to update those, and any other comments to reflect the new way of doing things.
+
+We will also likely want to add some new tests, that ensure that deprecation messages are logged, when someone attempts to use the events.
+
+We would likely want to implement this in the same way the `didInitAttrs` deprecation was implemented. i.e. we should maintain an array of
+component lifecycle hooks, and check if `hooks.includes(eventName)` and issue a deprecation if it does.
 
 ## How we teach this
 
--- a/text/0000-deprecate-component-lifecycle-events.md
+++ b/text/0000-deprecate-component-lifecycle-events.md
@@ -62,6 +62,7 @@ myHandler: on('didRender', () => {
 After:
 ```js
 didRender() {
+  this._super(...arguments);
   console.log('didRender!');
 }
 ```
--- a/text/0000-deprecate-component-lifecycle-events.md
+++ b/text/0000-deprecate-component-lifecycle-events.md
@@ -1,0 +1,62 @@
+- Start Date: (fill me in with today's date, 2018-01-19)
+- RFC PR: (leave this empty)
+- Ember Issue: (leave this empty)
+
+# Deprecate Component Lifecycle Events
+
+## Summary
+
+The goal of this RFC is to remove component lifecycle events, such as `Ember.on('didRender'`,
+in favor of enforcing usage of the lifecycle methods themselves i.e. `didRender(){}`
+
+## Motivation
+
+When using events, timing is not guaranteed, so it is possible to have unintended
+consequences and have things execute in an unexpected order. Using the methods directly
+does guarantee timing and solidifies the path forward on one recommended way of using
+lifecycle hooks.
+
+## Detailed design
+
+Ember will start logging deprecation messages that state which event you used and which
+method it should be converted to.
+
+The exact deprecation message will be decided later, but something along the lines of:
+
+```
+Using `Ember.on` event listeners for component lifecycle hooks is deprecated.
+Please move your code from `myHandler: Ember.on('didRender'` into the `didRender(){}` method.
+```
+
+## How we teach this
+
+The docs already recommend against using `Ember.on` for lifecycle hooks, but they do currently
+still mention it is possible.
+
+```
+While didInsertElement() is technically an event that can be listened for using on(),
+it is encouraged to override the default method itself, particularly when order of execution is important.
+```
+
+We will want to update the docs to either not mention using `on()` at all or to explicitly say *not* to use it.
+
+Other than updating the docs, the deprecation messages should be explanatory enough that people understand the path
+forward and convert all their `on()` style lifecycle hooks to the actual methods.
+
+eslint-plugin-ember, which is also now used in the default ember-cli blueprints, enforces this rule as well with
+[https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/no-on-calls-in-components.md](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/no-on-calls-in-components.md).
+
+## Drawbacks
+
+The only potential drawback, as far as I know, is that large code bases relying on lifecycle events will
+experience a potentially large amount of churn as they convert things over. We could potentially provide a
+codemod, if we think this will be problematic.
+
+## Alternatives
+
+This decision seems to be fairly universally supported, and I am not aware of any proposed alternatives.
+
+## Unresolved questions
+
+The only potential uncertainty is the exact deprecation message we want to issue when people use the
+deprecated events.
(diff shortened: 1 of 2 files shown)


text/0000-deprecate-component-lifecycle-events.md
```diff
@@ -43,9 +43,38 @@ We will want to update the docs to either not mention using `on()` at all or to
...
+Before:
+```js
+myHandler: on('didRender', () => {
+  console.log('didRender!');
+})
+```
+
+After:
+```js
+didRender() {
```
krisselden: needs super

text/0298-deprecate-component-lifecycle-events.md
```diff
@@ -0,0 +1,103 @@
...
+The exact deprecation message will be decided later, but something along the lines of:
+
+```
+Using `Ember.on` event listeners for component lifecycle hooks is deprecated.
+Please move your code from `myHandler: Ember.on('didRender'` into the `didRender(){}` method.
+```
+
+Ember does not appear to rely on these events internally, however there are at least a couple of mentions of 
+these events in the comments, referring specifically to `on('didInsertElement')` 
+[here](https://github.com/emberjs/ember.js/blob/master/packages/ember-metal/lib/run_loop.js#L140)
```
mehulkar: should change these to permalinks at a tag, so they won't be dead in the future!

rwwagner90: Good call! Updating now.

cibernox:
I'm all for removing those events. Specially thinking on the upcoming ES6 class syntax, defining a function and calling `super` will be much more streamlined with regular javascript.

Gaurav0:
The main use I have for lifecycle events is in mixins, where I don't want to override a method the user may want to override without calling _super

Panman8201:
> The main use I have for lifecycle events is in mixins, where I don't want to override a method the user may want to override without calling _super

An addon example of that is: https://github.com/alexspeller/ember-cli-active-link-wrapper/issues/10#issuecomment-169076245

rogatty:
What about the case where you need an observer to fire as part of the initialization process? [Ember guides recommend](https://guides.emberjs.com/v2.18.0/object-model/observers/#toc_observers-and-object-initialization) to use `on('init', observer(...))`. Is there a way to achieve this effect without the lifecycle events?

rwwagner90:
@rogatty can you not just put your observer code in a plain function and call that function inside `init`?

chadhietala:
I'm +1 for this because it will finally remove the this long standing issue where basically have to call hooks 2 times. In our apps we already leverage ESLint rules via https://github.com/ember-best-practices/eslint-plugin-ember-best-practices that already check for the [presence of evented APIs](https://github.com/ember-best-practices/eslint-plugin-ember-best-practices/blob/master/guides/rules/no-lifecycle-events.md#no-lifecycle-events).

rogatty:
@rwwagner90 I believe there were some issues with that approach ~3 years ago. After your suggestion, I tried to reproduce them in Ember Twiddle and found nothing. Thanks, we can now get rid of this dinosaur from our old code.

josemarluedke:
I give this a huge "+1". In our app we have been avoiding this for a long time already. It also removes a confusion for people learning ember, since there is two ways of doing the same thing.

rwjblue:
I am also in favor of moving forward on this, these event based hooks are a pretty significant trap...

text/0298-deprecate-component-lifecycle-events.md
```diff
@@ -0,0 +1,103 @@
...
+## Motivation
+
+When using events, timing is not guaranteed, so it is possible to have unintended
+consequences and have things execute in an unexpected order. Using the methods directly
+does guarantee timing and solidifies the path forward on one recommended way of using
+lifecycle hooks.
+
+## Detailed design
+
+Ember will start logging deprecation messages that state which event you used and which
```
rwjblue: AFAICT the RFC only gives examples of the events we are talking about, but the detailed design section should also include an explicit list of events.

I _believe_ this is the current list of events that are component specific ([from the curly component manager](https://github.com/emberjs/ember.js/blob/7ff756a87d5757d05588136d41516bcd5f5a0fd1/packages/ember-glimmer/lib/component-managers/curly.ts)):

* `willRender`
* `willInsertElement`
* `didInsertElement`
* `didRender`
* `didUpdateAttrs`
* `didReceiveAttrs`
* `willUpdate`
* `willRender`
* `didUpdate`
* `didRender`

rwjblue:
@Gaurav0:

> The main use I have for lifecycle events is in mixins, where I don't want to override a method the user may want to override without calling _super

I agree that this is the most common use case for these hooks, but I also think that it is one of the main reasons to move forward with this RFC. Often this is done by mixin authors defensively as a way to protect against component authors not properly `_super`ing. I think this is an anti-pattern. It makes reviewing/understanding what the mixin does much more complicated (because now _any_ method could be `didReceiveAttrs`). Instead, mixins should be clear about their API's and what methods they expect to be called (either having the user invoke their custom named method from the components own hook method, or implementing the hook method and `_super`ing).

Gaurav0:
@rwjblue @locks Should we update APIs, Guides, Docs etc. to indicate that _super should always be called in each of these lifecycle hooks just as we do for `init`?

rwwagner90:
> @rwjblue @locks Should we update APIs, Guides, Docs etc. to indicate that _super should always be called in each of these lifecycle hooks just as we do for init?

@Gaurav0 I say yes.

rwwagner90:
@rwjblue I added the list of events you provided 👍

rwwagner90:
What else do we need here?

rwjblue:
Soooo, when I brought this up at a recent Ember.js Core Team call there were a few folks that had concerns.  I'd love to have them comment here so we can discuss the concerns together...

@wycats / @dgeb - Would y'all mind sharing the concerns you brought up?

rwwagner90:
@rwjblue @wycats @dgeb I would love to have an open discussion about your concerns, if you wouldn't mind voicing them here please 😃

dgeb:
@rwwagner90 @rwjblue I apologize for taking so long to comment on this. I'm actually in favor of moving forward with this RFC, but cautiously.

Eliminating these lifecycle events and forcing usage of lifecycle methods will of course eliminate the ability of outside listeners to respond directly to these events. My concern was in verifying that tooling may currently rely on wiring up these listeners for potentially legit purposes such as auditing component lifecycle events throughout an application. Maybe this is the kind of use case we want to support in non-production builds or maybe it's not worthwhile even then.

I don't consider this concern to be a major blocker. It may be entirely imagined and no such usages exist in the wild. It just seems worthwhile to do a code search in emberobserver (although that search might be quite noisy). Perhaps it would be sufficient to announce FCP (Final Comment Period) and see if this raises alarms for anyone.

rwwagner90:
@dgeb I'm in favor of announcing FCP and seeing if anyone comes out of the woodwork in objection to it. 

@rwjblue what do you need from me here to help this along?

mike-north:
This is a great initiative. One thing that might be tricky to codemod/detect (since connections are defined in hbs) is parent/child components interacting with eahother via these events. It would be helpful for the deprecation guide to provide some nudging toward more sustainable patterns for these kinds of things (i.e. observer-free pattern for: child invokes ember-concurrency task in response to parent attrs changing)

tomdale:
We discussed this at the core team meeting today and agree that this RFC is ready for Final Comment Period. We had a few folks out this week, so we don't have consensus to merge this, but we do think it's appropriate to FCP and give folks a chance to raise any remaining issues before we discuss further.

rwwagner90:
@tomdale Thanks for the update! How long do we wait on FCP before merging?

rwwagner90:
Any updates on when we might want to merge this?

machty:
I had some confusions that I discussed with @rwwagner90:

- `on('init')` is included as one of the lifecycle hooks to be deprecated for Components, even though it's a general Ember Object lifecycle hook
- that said, we are not (for this RFC) deprecating `.on('init')` for regular (non-Component) Ember Objects
- `.on('click')` and other non-lifecycle Component hooks (that still use the `.on()` API) will not be deprecated as part of this RFC

tomdale:
At the core team meeting today, @wycats raised some concerns about whether it's worth deprecating `on('init')` and making people refactor existing mixins, which impacts downstream consumers, rather than just asking people to move to non-mixin patterns. We have asked him to comment here with his thoughts, but since there are some open concerns, we are not ready to merge or close this RFC yet.

rwwagner90:
@tomdale thanks for the update! If `init` is the only hook there is concern around, we could exclude `init` from this.

tomdale:
We discussed this again at the core team meeting. We agree that people should move away from this but we also are leery of a full-blown deprecation creating additional churn, when the real end-game is that people eventually refactor away from mixins and ideally towards Glimmer components.

We discussed that perhaps this should be a lint rather than a full-blown deprecation, as a way to nudge people towards the right path, but without the heavy-handedness of a deprecation warning.

rwwagner90:
I discussed this with @tomdale and will be updating this RFC to suggest using lint to enforce this.

rwwagner90:
@tomdale I adjusted some things to use lint, but there is already a lint rule for this stuff, so do we even want to do anything? I suppose we could have more specific rules, but let me know what we should do here.

rwwagner90:
Should we close this?

rwjblue:
Hmm. Perhaps? I still agree with the overall goal here, but I suspect that doing a change now is probably a bit of a disservice to folks. Making them do extra work to update their `Ember.Component` calls then in some future time asking them to refactor them to `@glimmer/component` seems not great.

What do you think?

rwwagner90:
@rwjblue I think there is nothing to do anyway, since these lint rules already exist. Are we able to just close or is there a process?

locks:
@rwwagner90 as the author you are free to withdrawn the proposal at any time 😁

PR #300, opened 2018-02-04
RFC Process Update

locks:
[Rendered](https://github.com/emberjs/rfcs/blob/rfc-process-update/text/0300-rfc-process-update.md)

### DIFF
--- a/text/0300-rfc-process-update.md
+++ b/text/0300-rfc-process-update.md
@@ -21,41 +21,47 @@ Proposals by both Core and community members are discussed and refined,
 with the result coming out much stronger.
 Other projects like React also adopted an RFC process, lending external validation to this idea.
 
-Following its initial implementation, shortcomings of the RFC process as well as new requirements were identified,
-which we will try to address in this proposal.
+During this time, the community and the core teams have identified shortcomings of the RFC process as well as new requirements,
+which this proposal intends to address:
 
 ### Confusion about emberjs/rfcs and ember-cli/rfcs
 
 The Ember project currently has two separate RFC processes for Ember.js and Ember CLI.
 
-This leads to confusion because community members must keep track of two different repositories in order to monitor progress.
-The submitter must consider these differences during the lifecycle of the RFC.
+This leads to confusion because the community needs to track of two different repositories for progress.
+For contributors there is the overhead of having to decide where to file their RFC if the proposal involves both projects,
+as well as being aware of the differences in the process during its lifecycle.
 
-There are also some cases where it is not clear exactly where the RFC should be submitted.
-This can also lead to multiple RFCs being written, one for each repository.
-While multiple RFCs might be warranted in some cases,
-it should be a choice instead of a limitation of the current processes.
-
-### Process does not cover sufficient areas of development
+### The process does not cover sufficient areas of development
 
 RFCs to emberjs/rfcs and ember-cli/rfcs concern themselves mostly with features or deprecations to ember.js and ember-cli respectfully, with some ember-data proposals in emberjs/rfcs.
 
-Once again taking inspiration from Rust, they have experimented expanding the types of RFCs proposed
-[crates.io default ranking](https://github.com/rust-lang/rfcs/blob/master/text/1824-crates.io-default-ranking.md)
-[default URLs for documentation](https://github.com/rust-lang/rfcs/blob/master/text/1826-change-doc-default-urls.md).
-The latter one, for example, would have been a useful RFC when designing the new [API docs](https://github.com/ember-learn/ember-api-docs).
+We do not have a process for some kinds of initiatives, like a website redesign,
+information architecture suggestions, SEO suggestions, and the like.
 
 ### Lingering RFCs
 
-The emberjs/rfcs repository currently has [54 issues](https://github.com/emberjs/rfcs/issues) and [62 PRs](https://github.com/emberjs/rfcs/pulls) open, and ember-cli/rfcs has [183 issues](https://github.com/ember-cli/ember-cli/issues) and [49 PRs](https://github.com/ember-cli/ember-cli/pulls).
-A significant percentage of these have not been active in the recent past.
+The emberjs/rfcs repository currently has [86 issues](https://github.com/emberjs/rfcs/issues) and [67 PRs](https://github.com/emberjs/rfcs/pulls) open, and ember-cli/rfcs has [21 issues](https://github.com/ember-cli/ember-cli/issues) and [34 PRs](https://github.com/ember-cli/ember-cli/pulls).
+A percentage of these have not been active in the recent past.
 
 We have kept PRs and issues open so people could more easily find the discussions,
 but this has instead given a negative feeling of staleness, as RFCs linger open without new feedback.
 
+### Gaps between RFC proposal, implementation work, and release
+
+At the moment the process underspecifies what happens when an RFC is accepted and merged.
+Even with the presence of the statusboard,
+it has been hard to communicate to the community the status of the various initiatives.
+
 ## Detailed design
 
-To address the issues raised in the “Motivation” section, I will propose changes to the current process.
+To address the issues raised in the “Motivation” section, I propose the following changes:
+
+- Add frontmatter to RFC files denoting which teams are involved
+- Unify emberjs/rfcs and ember-cli/rfcs
+- Define how core champions are assigned
+- Add the concept of FCP to close
+- Define how to track the implementation after an RFC is accepted
 
 ### Implement Core Champion
 
@@ -146,12 +152,7 @@ None at the moment.
 
 ## Unresolved questions
 
-### How do Champions get assigned to an RFC?
-
-More enterprising authors will likely reach out to team members on their own,
-but the added burden of having to seek a Champion might discourage people from authoring RFCs.
-
-The various teams could agree to review new RFCs during the weekly calls and assign Champions?
+None at the moment.
 
 ---
 
--- a/text/0300-rfc-process-update.md
+++ b/text/0300-rfc-process-update.md
@@ -1,0 +1,161 @@
+---
+tags: ember, ember-cli
+---
+
+- Start Date: 2018-02-04
+- RFC PR: https://github.com/emberjs/rfcs/pull/300
+- Ember Issue: (leave this empty)
+
+# RFC Process Update
+
+## Summary
+
+Introduce the idea of Core Champion to the Ember.js RFC process, merge Ember.js and Ember CLI RFCs repositories,
+and add metadata to make cross-project RFCs easier.
+
+## Motivation
+
+The Ember community has been using the Request For Comments (RFC) process to great effect over the last few years.
+The RFC process has proved invaluable.
+Proposals by both Core and community members are discussed and refined,
+with the result coming out much stronger.
+Other projects like React also adopted an RFC process, lending external validation to this idea.
+
+Following its initial implementation, shortcomings of the RFC process as well as new requirements were identified,
+which we will try to address in this proposal.
+
+### Confusion about emberjs/rfcs and ember-cli/rfcs
+
+The Ember project currently has two separate RFC processes for Ember.js and Ember CLI.
+
+This leads to confusion because community members must keep track of two different repositories in order to monitor progress.
+The submitter must consider these differences during the lifecycle of the RFC.
+
+There are also some cases where it is not clear exactly where the RFC should be submitted.
+This can also lead to multiple RFCs being written, one for each repository.
+While multiple RFCs might be warranted in some cases,
+it should be a choice instead of a limitation of the current processes.
+
+### Process does not cover sufficient areas of development
+
+RFCs to emberjs/rfcs and ember-cli/rfcs concern themselves mostly with features or deprecations to ember.js and ember-cli respectfully, with some ember-data proposals in emberjs/rfcs.
+
+Once again taking inspiration from Rust, they have experimented expanding the types of RFCs proposed
+[crates.io default ranking](https://github.com/rust-lang/rfcs/blob/master/text/1824-crates.io-default-ranking.md)
+[default URLs for documentation](https://github.com/rust-lang/rfcs/blob/master/text/1826-change-doc-default-urls.md).
+The latter one, for example, would have been a useful RFC when designing the new [API docs](https://github.com/ember-learn/ember-api-docs).
+
+### Lingering RFCs
+
+The emberjs/rfcs repository currently has [54 issues](https://github.com/emberjs/rfcs/issues) and [62 PRs](https://github.com/emberjs/rfcs/pulls) open, and ember-cli/rfcs has [183 issues](https://github.com/ember-cli/ember-cli/issues) and [49 PRs](https://github.com/ember-cli/ember-cli/pulls).
+A significant percentage of these have not been active in the recent past.
+
+We have kept PRs and issues open so people could more easily find the discussions,
+but this has instead given a negative feeling of staleness, as RFCs linger open without new feedback.
+
+## Detailed design
+
+To address the issues raised in the “Motivation” section, I will propose changes to the current process.
+
+### Implement Core Champion
+
+To make sure that RFCs receive adequate support from the team, Ember CLI has implemented the idea of a champion associated with each RFC.
+The goal is that in seeking a champion from the team,
+the RFC author starts a dialogue with the team and gets some early feedback.
+The champion is then responsible for representing the RFC in team meetings, and for shepherding its progress.
+
+This will be imported into emberjs/rfcs.
+
+### Improve the lifecycle of RFCs
+
+To address the triaging and inactivity problem of RFCs, I introduce the concept of FCP to close.
+
+Closing an RFC should be viewed as another triaging tool, not as a rejection of the RFC.
+Sometimes a rewrite of an RFC would be so fundamental that it would benefit of a fresh discussion in a new thread.
+Sometimes the original author is no longer active (Champions should help here as well),
+and someone else might want to take over the work in a new RFC.
+Sometimes the timing might not be right, or the feature might have been addressed some other way, and yes,
+sometimes it might be something that is not aligned with the team's values for the project.
+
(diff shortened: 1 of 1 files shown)


text/0000-rfc-process-update.md
```diff
@@ -0,0 +1,166 @@
...
+
+## Motivation
+
+Inspired by Rust, Ember implemented a Request For Commits (RFC) process with the intent to involve the community in discussing the future of the framework.
+The RFC process has proved invaluable.
+Proposals by both Core and community members are discussed and refined,
+with the result coming out much stronger.
+Other projects like React also adopted an RFC process, lending external validation to this idea.
+
+After having been involved with the RFC process for some time, and from feedback from the community, some issues with the current implementation of the process have been noted.
```
jenweber: Following the success of its initial implementation, the RFC process can be further refined to address some issues identified by the community and core team members.

text/0000-rfc-process-update.md
```diff
@@ -0,0 +1,166 @@
...
+
+### Confusion about emberjs/rfcs and ember-cli/rfcs
+
+The Ember project currently has two separate RFC processes for Ember.js and Ember CLI.
+
+This leads to confusion due to having to keep track of two different repositories to track progress.
```
jenweber: This leads to confusion because users must keep track of two different repositories in order to monitor progress.

text/0000-rfc-process-update.md
```diff
@@ -0,0 +1,166 @@
...
+It also increases the burden on the submitter of an RFC, as the two processes have diverged slightly,
+making the submitter be aware of these differences during the lifecycle of the RFC.
```
jenweber: The submitter must consider these differences during the lifecycle of the RFC.

text/0000-rfc-process-update.md
```diff
@@ -0,0 +1,166 @@
...
+
+There are also some cases where it is not clear exactly which repository the RFC should be submitted to.
```
jenweber: exactly where the RFC should be submitted.

text/0000-rfc-process-update.md
```diff
@@ -0,0 +1,166 @@
...
+
+Once again taking inspiration from Rust, they have experimented expanding the types of RFCs proposed
+[crates.io default ranking](https://github.com/rust-lang/rfcs/blob/master/text/1824-crates.io-default-ranking.md)
+[default URLs for documentation](https://github.com/rust-lang/rfcs/blob/master/text/1826-change-doc-default-urls.md).
+The latter one, for example, would have been a useful RFC when designing the new [API docs](https://github.com/ember-learn/ember-api-docs).
+
+Rust has RFCs for documentation, website, internals, tools.
+We only have emberjs and ember-cli due to them having been implemented separately.
+
+This makes it so that a potential contributor does not know where to submit a proposal.
```
jenweber: drop, moved up higher

text/0000-rfc-process-update.md
```diff
@@ -0,0 +1,166 @@
...
+
+### Lingering RFCs
+
+The emberjs/rfcs repository currently has [54 issues](https://github.com/emberjs/rfcs/issues) and [62 PRs](https://github.com/emberjs/rfcs/pulls) open, and ember-cli/rfcs has [183 issues](https://github.com/ember-cli/ember-cli/issues) and [49 PRs](https://github.com/ember-cli/ember-cli/pulls).
+A significant percentage of these have not been active in the recent past.
+
+We have kept PRs and issues open so people could more easily find the discussions,
+but this has instead given a negative feeling of staleness, as RFCs linger open without new feedback.
```
jenweber: Maybe give an example or cite a source

text/0000-rfc-process-update.md
```diff
@@ -0,0 +1,166 @@
...
+## Detailed design
+
+To address the issues raised in the “Motivation” section, I will propose changes to the current process.
+
+### Implement Core Champion
+
+To make sure that RFCs receive adequate support from the team, Ember CLI has implemented the idea of a champion associated with each RFC.
+The goal is that in seeking a champion from the team,
+the RFC author starts a dialogue with the team and gets some early feedback.
+The champion is then responsible for representing the RFC in team meetings, and for shepherding its progress.
```
jenweber: no comma

text/0000-rfc-process-update.md
```diff
@@ -0,0 +1,166 @@
...
+
+This will be imported into emberjs/rfcs.
```
jenweber: This champion model will be adopted for emberjs/rfcs.

jenweber: again, maybe link to an example in ember-cli

text/0000-rfc-process-update.md
```diff
@@ -0,0 +1,166 @@
...
+
+### Improve the lifecycle of RFCs
+
+To address the triaging and inactivity problem of RFCs, I introduce the concept of FCP to close.
```
jenweber: should say "concept of Final Comment Period (FCP) to close"

text/0000-rfc-process-update.md
```diff
@@ -0,0 +1,166 @@
...
+
+Closing an RFC should be viewed as another triaging tool, not as a rejection of the RFC.
+Sometimes a rewrite of an RFC would be so fundamental that it would benefit of a fresh discussion in a new thread.
+Sometimes the original author is no longer active (Champions should help here as well),
+and someone else might want to take over the work in a new RFC.
+Sometimes the timing might not be right, or the feature might have been addressed some other way, and yes,
+sometimes it might be something that is not aligned with the team's values for the project.
+
```
jenweber: Make the above a bulleted list

text/0000-rfc-process-update.md
```diff
@@ -0,0 +1,166 @@
...
+A good example of this is the [Named Blocks RFC](https://github.com/emberjs/rfcs/blob/master/text/0226-named-blocks.md#motivation),
+which lists in the motivation section previous attempts at similar ideas.
```
jenweber: which covers previous attempts in the "Motivation" section.

text/0000-rfc-process-update.md
```diff
@@ -0,0 +1,166 @@
...
+
+Like the FCP to merge process, once an RFC is marked as FCP to close there will be a period of one week where people can raise new concerns.
+After that period of one week, the respective team will review and close the RFC or extend the period for another week.
+
```
jenweber: I think this is the most likely section people may object to. Try to anticipate the objections and work them in here. i.e. "won't this discourage submitters from participating in the future?"

text/0000-rfc-process-update.md
```diff
@@ -0,0 +1,166 @@
...
+### Merge ember-cli/rfcs into emberjs/rfcs
+
+Part of removing confusion about the process is having a central repository for all Ember RFCs.
```
jenweber: There will be a central repository for all Ember RFCs.

text/0000-rfc-process-update.md
```diff
@@ -0,0 +1,166 @@
...
+
+To achieve merging ember-cli/rfcs into emberjs/rfcs the following will be done:
```
jenweber: Here is how ember-cli/rfcs can be merged into emberjs/rfcs:

text/0000-rfc-process-update.md
```diff
@@ -0,0 +1,166 @@
...
+
+* Remove active/completed concepts from ember-cli/rfcs
+  * Add front matter
```
jenweber: I don't know what this is, but if other people do, you could keep as-is

text/0000-rfc-process-update.md
jenweber: Also as a user, I would want to see formal affirmation that a decent number of core team members _would_ champion other RFCs before I'd feel good about this section. If this falls to just a few people, it will not work and we'll burn out them out

text/0000-rfc-process-update.md
```diff
@@ -0,0 +1,166 @@
...
+  * Copy both active and completed RFC files into `text` of emberjs/rfcs
+* Transfer active PRs and Issues to emberjs/rfcs
+* Archive ember-cli/rfcs
+
+There are some concerns about links breaking when we move the files to emberjs/rfcs,
+but given the fact that ember-cli/rfcs had the concept of active/completed by moving the files into different folders,
+links were already being broken.
+
```
jenweber: Phrase more positively - It is expected that some links will break in the process of moving files to emberjs/rfcs. This does not introduce any new concerns, since ember-cli/rfcs already had a process to move files into active/completed folders, which also breaks the links.

text/0000-rfc-process-update.md
```diff
@@ -0,0 +1,166 @@
...
+the RFC front matter should be oriented towards those teams.
+
+A list of the teams and respective projects should be added to the instructions,
+possibly with the addition of per-team instructions on specifics of the project.
+Additional templates might be created as well, such a design work template.
+
+### Provide more guidance for post-RFC
+
+At the moment it is not clear for some contributors what happens once an RFC is merged.
+How and when should work to implement an RFC be done?
```
jenweber: Needs another sentence or two. Maybe just specifying that the instructions will be added to a CONTRIBUTING.md file or something. It would be good to see a draft and link it here.

text/0000-rfc-process-update.md
```diff
@@ -0,0 +1,166 @@
...
+
+## How We Teach This
+
+To ensure that contributors are updated on the RFC process and the process is clear,
+the documentation should be improved in a couple of ways.
+
+The README should be updated to include information about Champions,
+how they are assigned, what they do.
```
jenweber: and what they do

text/0000-rfc-process-update.md
```diff
@@ -0,0 +1,166 @@
...
+
+## Drawbacks
+
+### Adjustment period
+
+There are active RFCs in ember-cli/rfcs. Moving these discussions would be onerous, so they should be kept there until completion, and no new RFCs accepted.
+
+### Permalinks to ember-cli/rfcs proposals
+
+Moving the RFC files from ember-cli/rfcs (active or completed) to emberjs/rfcs can be seen as a breaking change, and could lead to someone linking to ember-cli/rfcs and then the RFC being updated in emberjs/rfcs. However, ember-cli/rfcs already suffers from a linking problem due to the active/completed folders, as RFCs need to be moved from one to the other even after being accepted.
```
jenweber: Repetition with "There are some concerns about links breaking..." I recommend removing the "There are concerns" section and working it into here

text/0000-rfc-process-update.md
```diff
@@ -0,0 +1,166 @@
...
+None at the moment.
+
+## Unresolved questions
+
+### How do Champions get assigned to an RFC?
+
+More enterprising authors will likely reach out to team members on their own,
+but the added burden of having to seek a Champion might discourage people from authoring RFCs.
+
+The various teams could agree to review new RFCs during the weekly calls and assign Champions?
```
jenweber: Maybe list off a few more ways to do this, in a bulleted list, and drop the "?"

text/0000-rfc-process-update.md
jenweber: It's kind of a side comment. I'm hoping that team members will chime in and say, "yes I want to do this," not necessarily that people should be listed here

text/0000-rfc-process-update.md
```diff
@@ -0,0 +1,166 @@
...
+# RFC Process Update
+
+## Summary
+
+Introduce the idea of Core Champion to the Ember.js RFC process, merge Ember.js and Ember CLI RFCs repositories,
+and add metadata to make cross-project RFCs easier.
+
+## Motivation
+
+Inspired by Rust, Ember implemented a Request For Commits (RFC) process with the intent to involve the community in discussing the future of the framework.
```
ro0gr: I'm confused about RFC abbreviation. I've heard at least about three interpretations:
  - request for **change**
  - request for **comments**
  - and now request for **commits**

Does anybody knows for sure what does "C" stand for?

locks: It's a typo, should be Request For Comments [[1]](https://github.com/emberjs/rfcs#ember-rfcs) [[2]](https://github.com/rust-lang/rfcs/#rust-rfcs).

Gaurav0:
Thanks for this. I have one additional suggestion: a standard time period after which an FCP to close should be expected. Perhaps 6 months. The idea should be that RFCs should not be perceived to be treated differently based on author and all RFCs should have sufficient time for comment, no matter how unpopular.

Gaurav0:
One other concern: there should not be a rather of FCP to close immediately after this is accepted.

text/0000-rfc-process-update.md
sandstrom: How about `people/participants/contributors` over `users`? 😄

joukevandermaas:
The part about assigning champions is not entirely clear to me. Does the RFC author seek out a champion? Or do they just submit the RFC and wait for one to assign themselves?

If the former, it may help to have some semi-standardized way of contacting core team members for this purpose (perhaps a slack channel). This should be mentioned somewhere in the repo so people know where to start.

If the latter, it may be a good idea to have some time-frame in which a decision is made and communicated (even if the decision is that no one has time to champion that RFC for now). My ballpark suggestion would be about 6 weeks after opening the RFC (so about one release cycle). I know time-frames are difficult to commit to, so maybe this is impossible.

rwjblue:
> The part about assigning champions is not entirely clear to me. Does the RFC author seek out a champion? Or do they just submit the RFC and wait for one to assign themselves?

I think that in practice, interested team members are _falling over themselves_ to champion ideas that they like and/or want to advance.  I know that I do... But I think the answer here is actually "both". Advocating for the RFC both in its own comments, in `#dev-rfcs` in slack, on the forum, etc is likely the best way...

> If the latter, it may be a good idea to have some time-frame in which a decision is made and communicated (even if the decision is that no one has time to champion that RFC for now). My ballpark suggestion would be about 6 weeks after opening the RFC (so about one release cycle). I know time-frames are difficult to commit to, so maybe this is impossible.

I'm not sure this is a great idea. I don't think we need more process just for process sake, all these rules are just one more thing that we have to maintain/adhere to/etc.  I definitely think that a nice default timeframe to `FCP to close` after not successfully engaging a champion is warranted (I believe that @Gaurav0 also suggested this).

rwjblue:
I'd love to see this move forward, @locks are you still planning on pushing this?

mehulkar:
Is it possible to move ember-cli into the emberjs org as well? It's a core part of the Ember experience now and IMO if RFCs are merged, it should be located closer to the library also. IMO, many of the other repos on ember-cli could also be moved into `github.com/emberjs` or maybe github.com/emberjs-core-addons` or something similar.

ctcpip:
> Each RFC will require a champion from the primary core team to which the RFC has been marked relevant. The champion must be found by the opener of the RFC or other community member.

So basically if a core member can't be convinced to champion, the RFC is dead in the water?  This does not seem community-driven to me.  It sounds like an oligarchy.  I understand the motivation behind it, but it seems like it's taken a step too far.

ctcpip:
@MelSumner -- I don't mean to be pedantic, but that section could use some work in terms of elaborating on how exactly the required champion manifests.  Yes the core team is community, but is there any concern that as their numbers are limited, they may be reluctant to champion otherwise-worthy RFCs because they are already very busy?  I may be interpreting the intent too strictly, but if so, I think that section could be given a bit more detail. 

> "A section on 'Finding a champion' will be added to the instructions on proposing an RFC."

Maybe that would have more specifics?  The language just seems to be a bit discouraging to me and implies that without a champion, you may as well not submit an RFC.  For example, in some projects, if you miss some teeny little instruction when submitting an issue, the maintainer will close it outright, even though it's a valid issue.  I'm not saying that's analogous to what's happening here, but from an outsider looking in, it may look similar.  As I understand it, this comes at a time when increasing community involvement has been a priority, so some sensitivity around something like this I think is warranted.

To put it another way, I'm not cynical about the intent of the RFC or of the folks proposing it.  I'm concerned about the language (and maybe part of the process itself?) possibly discouraging contribution by implying (or seeming to imply) that this additional requirement be met before being heard.

kategengler:
The Champion requirement has been in place on ember-cli/rfcs since early 2017. It was implemented to attempt to prevent RFCs from stagnating. It has served ember-cli well. 

The current reality is that if an RFC does not have a core team member advocating for it -- making sure the core team(s) read it, ensuring all RFC comments are addressed, gathering consensus for FCP, for merging -- is that the RFC stays open and stagnates. That's not to say the core team doesn't pay attention to new RFCs (we certainly do), but that if an RFC doesn't entice someone on the team to push it forward, members may read it and forget it. The champion requirement makes this an explicit requirement and provides a guide for fulfilling it. It reflects that RFCs that are most likely to be accepted are ones for which the author "lobbies" -- by sharing ideas in discord, chatting on the forums, giving talks, writing, etc. 

Mel is correct when she says "assume the best". While the numbers of core team members are limited, it is certainly not small: https://emberjs.com/team/ and in practice, team members will be excited to champion any RFC they think improves Ember. Getting a core team member on board to champion would be a reasonable metric to decide whether it is worth spending the time writing an RFC.

Gaurav0:
> So basically if a core member can't be convinced to champion, the RFC is dead in the water? This does not seem community-driven to me. It sounds like an oligarchy. I understand the motivation behind it, but it seems like it's taken a step too far.

The final decision is in the hands of the core team anyway. The RFC Process is about soliciting community input, not creating some sort of democracy. At least Ember's core team is relatively large and comprises many different points of view, and they have proven that they generally don't ignore community needs.

mehulkar:
> Everything is possible 😈 but I don't personally think this would add much value. It is quite nice to have a grouping of packages that relate to "the CLI tool"...

Some counter points:

- Considering that it’s basically impossible to use Ember without CLI (no mention in guides), I think the value add is that users don’t need to hunt down another org. 
- The website is merged, the RFCs are merged, it seems weird for the code not to live in the same place. 
- GH org’s repo search and tags are a pretty decent way to organize / find repos, IMO. 
- in the (very) long term, I expect that much of ember tooling will converge with non-ember tooling, so the need for supporting CLI packages will go down.

But that said, I haven’t vetted all the repos in ember-CLI yet so it’s possible I’m missing something.

PR #308, opened 2018-02-16
Transition Away From Property Fallback Behavior

chadhietala:
[Rendered](https://github.com/chadhietala/rfcs-1/blob/remove-prop-fallback/text/0000-deprecate-property-lookup-fallback.md)

### DIFF
--- a/text/0000-deprecate-property-lookup-fallback.md
+++ b/text/0000-deprecate-property-lookup-fallback.md
@@ -4,7 +4,7 @@
 
 # Summary
 
-Deprecate the fallback behavior of resolving `{{foo}}` by requiring the usage of `{{this.foo}}` as syntax to refer to properties of the templates' backing component. This would be the default behavior in Glimmer Components.
+Beging the transition to deprecate the fallback behavior of resolving `{{foo}}` by requiring the usage of `{{this.foo}}` as syntax to refer to properties of the templates' backing component. This would be the default behavior in Glimmer Components.
 
 For example, given the following component class:
 
@@ -86,9 +86,31 @@ Furthermore, by enforcing the `this` prefix, tooling like the [Ember Language Se
 
 # Transition Path
 
-The `{{this.foo}}` syntax has always been supported Handlebars syntax and thus can be used in every version of Ember. However, no guides or documentation promote its usage. While applications can start migrating today to this more explicit syntax, we can surface information about what properties on the class were accessed allowing developers to get a concise list of expressions that should be migrated. This can be surfaced because we are already disambiguating the syntax at runtime.
+We intend this to be a _very slow_ process as we understand it is a large change. Because of this we will be doing a phased rollout to help guide people in transtion. Below is an outline of how we plan to roll this change out.
 
-Using this information, we will also have the ability to create a codemod that will help migrate templates.
+**Phase 1:**
+- Add [template lint rule](https://github.com/ember-template-lint/ember-template-lint/pull/392) to [ember-template-lint](https://github.com/ember-template-lint/ember-template-lint) as an **opt-in** rule
+- Document the [codemod infrastructure](https://github.com/dyfactor/dyfactor) and [codemod](https://github.com/dyfactor/dyfactor-plugin-disambiguate-locals). Make it available for early adopters
+- Start updating docs to use `this.`
+
+**Phase 2:**
+- Add the lint rule by default _in the apps_ `.template-lintrc.js`
+- Complete doc migration to use `this.`
+
+**Phase 3:**
+- Enable the lint rule by default in the `recommended` config
+
+**Phase 4:**
+- Introduce deprecation **app only** fallbacks
+
+**Phase 5:**
+- Introduce deprecation for **any** fallbacks
+
+**Phase 6:**
+- Rev major to 4.0.0
+
+**Phase 7:**
+- Remove fallback functionality post 4.3.0 LTS
 
 # How We Teach This
 
--- a/text/0000-deprecate-property-lookup-fallback.md
+++ b/text/0000-deprecate-property-lookup-fallback.md
@@ -32,7 +32,47 @@ Currently, the way to access properties on a components class is `{{greeting}}` 
 
 The first problem with this approach is that the `{{greeting}}` syntax is ambiguous, as it could be referring to a local variable (block param), a helper with no arguments, a closed over component, or a property on the component class.
 
-This can often lead to confusion for readers of the template. Upon encountering `{{greeting}}` in a component's template, the reader has to check all of these places: first you need to scan the surrounding lines for block params with that name; next you check in the helpers folder to see it there is a helper with that name (it could also be coming from an addon!); finally, you check the component's JavaScript class to look for a (computed) property.
+## Exemplar
+Consider the following example where the ambiguity can cause issues:
+
+You have a component class that looks like the following component and template:
+
+```js
+import Component from '@ember/component';
+import computed from '@ember/computed';
+
+export default Component.extend({
+  formatName: computed('firstName', 'lastName', function() {
+    return `${this.firstName} ${this.lastName}`;
+  });
+});
+```
+
+```hbs
+<h1>Hello {{formatName}}!</h1>
+```
+
+Given `{ firstName: 'Chad', lastName: 'Hietala' }` Ember will render the following:
+
+```html
+<h1>Hello Chad Hietala!</h1>
+```
+
+Now some time goes on and someone adds a `formatName` helper that looks like the following:
+
+```js
+export default function formatName([firstName, lastName]) {
+  return `${this.firstName} ${this.lastName}`;
+}
+```
+
+Due to the fact helpers take precedence over property lookups our `{{formatName}}` now resolves to a helper. When the helper runs it doesn't have any arguments so our template now renders the following:
+
+```html
+<h1>Hello !</h1>
+```
+
+This can be a refactoring hazord and can often lead to confusion for readers of the template. Upon encountering `{{greeting}}` in a component's template, the reader has to check all of these places: first you need to scan the surrounding lines for block params with that name; next you check in the helpers folder to see it there is a helper with that name (it could also be coming from an addon!); finally, you check the component's JavaScript class to look for a (computed) property.
 
 Like [RFC#0276](https://github.com/emberjs/rfcs/blob/68812bf2d439c6bb77ad491e0159b371b68c5c35/text/0276-named-args.md) made argument usage explicit through the `@` prefix, the `this` prefix will resolve the ambiguity and greatly improve clarity, especially in big projects with a lot of files (and uses a lot of addons).
 
@@ -44,6 +84,8 @@ Furthermore, by enforcing the `this` prefix tooling like the [Ember Language Ser
 # Transition Path
 The `{{this.foo}}` syntax has always been supported Handlebars and thus can be used in every version of Ember, however no guides or documentation promote its usage. While applications can start migrating today to this more explicit syntax, we can surface information about what properties on the class were accessed allowing developers to get a concise list of expressions that should be migrated. This can be surfaced because we are already disambiguate the syntax at runtime.
 
+Using this information, we will also have the ability to create a codemod that will help migrate templates.
+
 # How We Teach This
 
 `{{this.foo}}` is the way to access the properties on the component class. This also aligns to property access in JavaScript.
@@ -51,8 +93,6 @@ The `{{this.foo}}` syntax has always been supported Handlebars and thus can be u
 Since the `{{this.foo}}` syntax has worked in Ember.Component (which is the only kind of components available today) since the 1.0 series, we are not really in a rush to migrate the community (and the guides, etc) to using the new syntax. In the meantime, this could be viewed as a tool to improve clarity in templates.
 
 While we think writing {{this.foo}} would be a best practice for new code going forward, the community can migrate at its own pace one component at a time. However, once the fallback functionality is eventually removed this will result in a "Helper not found" error.
-
-We can also encourage the community to supplement this effort by wiring linting tools and code mods.
 
 ## Syntax Breakdown
 The follow is a breakdown of the different forms and what they mean.
--- a/text/0000-deprecate-property-lookup-fallback.md
+++ b/text/0000-deprecate-property-lookup-fallback.md
@@ -72,7 +72,7 @@ Due to the fact helpers take precedence over property lookups our `{{formatName}
 <h1>Hello !</h1>
 ```
 
-This can be a refactoring hazord and can often lead to confusion for readers of the template. Upon encountering `{{greeting}}` in a component's template, the reader has to check all of these places: first you need to scan the surrounding lines for block params with that name; next you check in the helpers folder to see it there is a helper with that name (it could also be coming from an addon!); finally, you check the component's JavaScript class to look for a (computed) property.
+This can be a refactoring hazard and can often lead to confusion for readers of the template. Upon encountering `{{greeting}}` in a component's template, the reader has to check all of these places: first you need to scan the surrounding lines for block params with that name; next you check in the helpers folder to see it there is a helper with that name (it could also be coming from an addon!); finally, you check the component's JavaScript class to look for a (computed) property.
 
 Like [RFC#0276](https://github.com/emberjs/rfcs/blob/68812bf2d439c6bb77ad491e0159b371b68c5c35/text/0276-named-args.md) made argument usage explicit through the `@` prefix, the `this` prefix will resolve the ambiguity and greatly improve clarity, especially in big projects with a lot of files (and uses a lot of addons).
 
--- a/text/0000-deprecate-property-lookup-fallback.md
+++ b/text/0000-deprecate-property-lookup-fallback.md
@@ -4,7 +4,7 @@
 
 # Summary
 
-Deprecate the fallback behavior of resolving `{{foo}}` by requiring the usage of `{{this.foo}}` as syntax to refer to properties of the templates's backing component. This would be the default behavior in Glimmer Components.
+Deprecate the fallback behavior of resolving `{{foo}}` by requiring the usage of `{{this.foo}}` as syntax to refer to properties of the templates' backing component. This would be the default behavior in Glimmer Components.
 
 For example, given the following component class:
 
@@ -33,6 +33,7 @@ Currently, the way to access properties on a components class is `{{greeting}}` 
 The first problem with this approach is that the `{{greeting}}` syntax is ambiguous, as it could be referring to a local variable (block param), a helper with no arguments, a closed over component, or a property on the component class.
 
 ## Exemplar
+
 Consider the following example where the ambiguity can cause issues:
 
 You have a component class that looks like the following component and template:
@@ -52,7 +53,7 @@ export default Component.extend({
 <h1>Hello {{formatName}}!</h1>
 ```
 
-Given `{ firstName: 'Chad', lastName: 'Hietala' }` Ember will render the following:
+Given `{ firstName: 'Chad', lastName: 'Hietala' }`, Ember will render the following:
 
 ```html
 <h1>Hello Chad Hietala!</h1>
@@ -66,36 +67,38 @@ export default function formatName([firstName, lastName]) {
 }
 ```
 
-Due to the fact helpers take precedence over property lookups our `{{formatName}}` now resolves to a helper. When the helper runs it doesn't have any arguments so our template now renders the following:
+Due to the fact that helpers take precedence over property lookups, our `{{formatName}}` now resolves to a helper. When the helper runs it doesn't have any arguments so our template now renders the following:
 
 ```html
 <h1>Hello !</h1>
 ```
 
-This can be a refactoring hazard and can often lead to confusion for readers of the template. Upon encountering `{{greeting}}` in a component's template, the reader has to check all of these places: first you need to scan the surrounding lines for block params with that name; next you check in the helpers folder to see it there is a helper with that name (it could also be coming from an addon!); finally, you check the component's JavaScript class to look for a (computed) property.
+This can be a refactoring hazard and can often lead to confusion for readers of the template. Upon encountering `{{greeting}}` in a component's template, the reader has to check all of these places: first, you need to scan the surrounding lines for block params with that name; next, you check in the helpers folder to see if there is a helper with that name (it could also be coming from an addon!); finally, you check the component's JavaScript class to look for a (computed) property.
 
 Like [RFC#0276](https://github.com/emberjs/rfcs/blob/68812bf2d439c6bb77ad491e0159b371b68c5c35/text/0276-named-args.md) made argument usage explicit through the `@` prefix, the `this` prefix will resolve the ambiguity and greatly improve clarity, especially in big projects with a lot of files (and uses a lot of addons).
 
 As an aside, the ambiguity that causes confusion for human readers is also a problem for the compiler. While it is not the main goal of this proposal, resolving this ambiguity also helps the rendering system. Currently, the "runtime" template compiler has to perform a helper lookup for every `{{greeting}}` in each template. It will be able to skip this resolution process and perform other optimizations (such as reusing the internal [reference](https://github.com/glimmerjs/glimmer-vm/blob/master/guides/04-references.md)
 object and caches) with this addition.
 
-Furthermore, by enforcing the `this` prefix tooling like the [Ember Language Server](https://github.com/emberwatch/ember-language-server) does not need to know about fallback resolution rules. This makes common features like ["Go To Definition"](https://code.visualstudio.com/docs/editor/editingevolved#_go-to-definition) much easier to implement since we have semantics that mean "property on class".
+Furthermore, by enforcing the `this` prefix, tooling like the [Ember Language Server](https://github.com/emberwatch/ember-language-server) does not need to know about fallback resolution rules. This makes common features like ["Go To Definition"](https://code.visualstudio.com/docs/editor/editingevolved#_go-to-definition) much easier to implement since we have semantics that mean "property on class".
 
 # Transition Path
-The `{{this.foo}}` syntax has always been supported Handlebars and thus can be used in every version of Ember, however no guides or documentation promote its usage. While applications can start migrating today to this more explicit syntax, we can surface information about what properties on the class were accessed allowing developers to get a concise list of expressions that should be migrated. This can be surfaced because we are already disambiguate the syntax at runtime.
+
+The `{{this.foo}}` syntax has always been supported Handlebars syntax and thus can be used in every version of Ember. However, no guides or documentation promote its usage. While applications can start migrating today to this more explicit syntax, we can surface information about what properties on the class were accessed allowing developers to get a concise list of expressions that should be migrated. This can be surfaced because we are already disambiguating the syntax at runtime.
 
 Using this information, we will also have the ability to create a codemod that will help migrate templates.
 
 # How We Teach This
 
-`{{this.foo}}` is the way to access the properties on the component class. This also aligns to property access in JavaScript.
+`{{this.foo}}` is the way to access the properties on the component class. This also aligns with property access in JavaScript.
 
-Since the `{{this.foo}}` syntax has worked in Ember.Component (which is the only kind of components available today) since the 1.0 series, we are not really in a rush to migrate the community (and the guides, etc) to using the new syntax. In the meantime, this could be viewed as a tool to improve clarity in templates.
+Since the `{{this.foo}}` syntax has worked in Ember.Component (which is the only kind of component available today) since the 1.0 series, we are not really in a rush to migrate the community (and the guides, etc) to using the new syntax. In the meantime, this could be viewed as a tool to improve clarity in templates.
 
-While we think writing {{this.foo}} would be a best practice for new code going forward, the community can migrate at its own pace one component at a time. However, once the fallback functionality is eventually removed this will result in a "Helper not found" error.
+While we think writing `{{this.foo}}` would be a best practice for new code going forward, the community can migrate at its own pace one component at a time. However, once the fallback functionality is eventually removed this will result in a "Helper not found" error.
 
 ## Syntax Breakdown
-The follow is a breakdown of the different forms and what they mean.
+
+The follow is a breakdown of the different forms and what they mean:
 
 - `{{@foo}}` is an argument passed to the component
 - `{{this.foo}}` is a property on the component class
@@ -103,12 +106,15 @@ The follow is a breakdown of the different forms and what they mean.
 - `{{foo}}` is a helper
 
 # Drawbacks
+
 The largest downside of this proposal is that it makes templates more verbose, causing developers to type a bit more. This will also create a decent amount of deprecation noise, although we feel like tools like [ember-cli-deprecation-workflow](https://github.com/mixonic/ember-cli-deprecation-workflow) can help mitigate this.
 
 # Alternatives
+
 This pattern of having programming model constructs to distinguish between of the backing class and arguments passed to the component is not unique to Ember.
 
 ## What Other Frameworks Do
+
 React has used `this.props` to talk about values passed to you and `this.state` to mean data owned by the backing component class since it was released. However, this approach of creating a specific object on the component class to mean "properties available to the template", would likely be even more an invasive change and goes against the mental model that the context for the template is the class.
 
 Vue requires enumeration of `props` passed to a component, but the values in the template suffer from the ambiguity that we are trying to solve.
@@ -116,11 +122,14 @@ Vue requires enumeration of `props` passed to a component, but the values in the
 Angular relies heavily on the dependency injection e.g. `@Input` to enumerate the bindings that were passed to the component and relies heavily on TypeScript to hide or expose values to templating layer with `public` and `private` fields. Like Vue, Angular does not disambiguate.
 
 ## Introduce Yet Another Sigil
+
 We could introduce another sigil to remove ambiguity. This would address the concern about verbosity, however it is now another thing we would have to teach.
 
 ## Do Nothing
-I personally don't think this is an option, since the goal is to provide clarity for applications as they evolve over time and to provide a more consise mental model.
+
+I personally don't think this is an option, since the goal is to provide clarity for applications as they evolve over time and to provide a more concise mental model.
 
 # Unresolved questions
-Soon...
+
+Soon…
 
--- a/text/0000-deprecate-property-lookup-fallback.md
+++ b/text/0000-deprecate-property-lookup-fallback.md
@@ -127,11 +127,14 @@ Angular relies heavily on the dependency injection e.g. `@Input` to enumerate th
 
 We could introduce another sigil to remove ambiguity. This would address the concern about verbosity, however it is now another thing we would have to teach.
 
+## Change Resolution Order
+
+The other option is to reverse the resolution order to prefer properties over helpers. However this has the reverse problem as described in the exemplar.
+
 ## Do Nothing
 
 I personally don't think this is an option, since the goal is to provide clarity for applications as they evolve over time and to provide a more concise mental model.
 
 # Unresolved questions
-
-Soon…
+TBD
 
--- a/text/0000-deprecate-property-lookup-fallback.md
+++ b/text/0000-deprecate-property-lookup-fallback.md
@@ -111,7 +111,7 @@ The largest downside of this proposal is that it makes templates more verbose, c
 
 # Alternatives
 
-This pattern of having programming model constructs to distinguish between of the backing class and arguments passed to the component is not unique to Ember.
+This pattern of having programming model constructs to distinguish between the backing class and arguments passed to the component is not unique to Ember.
 
 ## What Other Frameworks Do
 
--- a/text/0000-deprecate-property-lookup-fallback.md
+++ b/text/0000-deprecate-property-lookup-fallback.md
@@ -62,7 +62,7 @@ Now some time goes on and someone adds a `formatName` helper that looks like the
 
 ```js
 export default function formatName([firstName, lastName]) {
-  return `${this.firstName} ${this.lastName}`;
+  return `${firstName} ${lastName}`;
 }
 ```
 
--- a/text/0000-deprecate-property-lookup-fallback.md
+++ b/text/0000-deprecate-property-lookup-fallback.md
@@ -4,7 +4,7 @@
 
 # Summary
 
-Beging the transition to deprecate the fallback behavior of resolving `{{foo}}` by requiring the usage of `{{this.foo}}` as syntax to refer to properties of the templates' backing component. This would be the default behavior in Glimmer Components.
+Beginning the transition to deprecate the fallback behavior of resolving `{{foo}}` by requiring the usage of `{{this.foo}}` as syntax to refer to properties of the templates' backing component. This would be the default behavior in Glimmer Components.
 
 For example, given the following component class:
 
--- a/text/0000-deprecate-property-lookup-fallback.md
+++ b/text/0000-deprecate-property-lookup-fallback.md
@@ -108,9 +108,10 @@ We intend this to be a _very slow_ process as we understand it is a large change
 
 **Phase 6:**
 - Rev major to 4.0.0
+- Add assert for fallback behavior
 
 **Phase 7:**
-- Remove fallback functionality post 4.3.0 LTS
+- Remove fallback functionality in 4.5, post 4.4.0 LTS
 
 # How We Teach This
 
--- a/text/0000-deprecate-property-lookup-fallback.md
+++ b/text/0000-deprecate-property-lookup-fallback.md
@@ -59,7 +59,7 @@ Given `{ firstName: 'Chad', lastName: 'Hietala' }`, Ember will render the follow
 <h1>Hello Chad Hietala!</h1>
 ```
 
-Now some time goes on and someone adds a `formatName` helper that looks like the following:
+Now some time goes on and someone adds a `formatName` helper at `app/helpers/fortmatName.js` that looks like the following:
 
 ```js
 export default function formatName([firstName, lastName]) {
--- a/text/0000-deprecate-property-lookup-fallback.md
+++ b/text/0000-deprecate-property-lookup-fallback.md
@@ -26,6 +26,8 @@ One would refer to the `greeting` property as such:
 
 Ember will render "Hello, Chad".
 
+To make this deprecation tractable, we will provide a codemod for migrating templates.
+
 # Motivation
 
 Currently, the way to access properties on a components class is `{{greeting}}` from a template. This works because the component class is one of the objects we resolve against during the evaluation of the expression.
--- a/text/0000-deprecate-property-lookup-fallback.md
+++ b/text/0000-deprecate-property-lookup-fallback.md
@@ -1,0 +1,86 @@
+- Start Date: 2-15-2018
+- RFC PR:
+- Ember Issue:
+
+# Summary
+
+Deprecate the fallback behavior of resolving `{{foo}}` by requiring the usage of `{{this.foo}}` as syntax to refer to properties of the templates's backing component. This would be the default behavior in Glimmer Components.
+
+For example, given the following component class:
+
+```js
+import Component from '@ember/component';
+export default Component.extends({
+  init() {
+    super(...arguments);
+    this.set('greeting', 'Hello');
+  }
+});
+```
+
+One would refer to the `greeting` property as such:
+
+```hbs
+<h1>{{this.greeting}}, Chad</h1>
+```
+
+Ember will render "Hello, Chad".
+
+# Motivation
+
+Currently, the way to access properties on a components class is `{{greeting}}` from a template. This works because the component class is one of the objects we resolve against during the evaluation of the expression.
+
+The first problem with this approach is that the `{{greeting}}` syntax is ambiguous, as it could be referring to a local variable (block param), a helper with no arguments, a closed over component, or a property on the component class.
+
+This can often lead to confusion for readers of the template. Upon encountering `{{greeting}}` in a component's template, the reader has to check all of these places: first you need to scan the surrounding lines for block params with that name; next you check in the helpers folder to see it there is a helper with that name (it could also be coming from an addon!); finally, you check the component's JavaScript class to look for a (computed) property.
+
+Like [RFC#0276](https://github.com/emberjs/rfcs/blob/68812bf2d439c6bb77ad491e0159b371b68c5c35/text/0276-named-args.md) made argument usage explicit through the `@` prefix, the `this` prefix will resolve the ambiguity and greatly improve clarity, especially in big projects with a lot of files (and uses a lot of addons).
+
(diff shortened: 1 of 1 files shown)


text/0000-deprecate-property-lookup-fallback.md
```diff
@@ -0,0 +1,126 @@
...
+}
+```
+
+Due to the fact helpers take precedence over property lookups our `{{formatName}}` now resolves to a helper. When the helper runs it doesn't have any arguments so our template now renders the following:
+
+```html
+<h1>Hello !</h1>
+```
+
+This can be a refactoring hazord and can often lead to confusion for readers of the template. Upon encountering `{{greeting}}` in a component's template, the reader has to check all of these places: first you need to scan the surrounding lines for block params with that name; next you check in the helpers folder to see it there is a helper with that name (it could also be coming from an addon!); finally, you check the component's JavaScript class to look for a (computed) property.
```
wycats: "hazard"

wycats: Note, this problem makes it very hard to "tool" -- "surrounding block param" is purely local, so you can write a syntax colorer or completion tool that understands it. But other kinds of fallback are much less local and are therefore much harder to tool.

rwjblue:
Thanks for writing this up @chadhietala!  I am definitely in favor of reducing the mental overhead of scanning a template and "knowing" where things are coming from....

text/0000-deprecate-property-lookup-fallback.md
```diff
@@ -0,0 +1,126 @@
...
+
+While we think writing {{this.foo}} would be a best practice for new code going forward, the community can migrate at its own pace one component at a time. However, once the fallback functionality is eventually removed this will result in a "Helper not found" error.
+
+## Syntax Breakdown
+The follow is a breakdown of the different forms and what they mean.
+
+- `{{@foo}}` is an argument passed to the component
+- `{{this.foo}}` is a property on the component class
+- `{{#with this.foo as |foo|}} {{foo}} {{/with}}` the `{{foo}}` is a local
+- `{{foo}}` is a helper
```
wycats: I think we can merge the last two, so that `{{foo}}` is always "look up foo in the current scope", and if a variable is not in scope, it falls back to the file system "scope" to finish the search.

text/0000-deprecate-property-lookup-fallback.md
```diff
@@ -0,0 +1,126 @@
...
+
+# Drawbacks
+The largest downside of this proposal is that it makes templates more verbose, causing developers to type a bit more. This will also create a decent amount of deprecation noise, although we feel like tools like [ember-cli-deprecation-workflow](https://github.com/mixonic/ember-cli-deprecation-workflow) can help mitigate this.
```
wycats: I think we would want to say that we must have a codemod before enabling the deprecation due to the noise it would cause. We should also think about whether we can scope the deprecation to the user's app; for example, we could consider making it a build-time deprecation, rather than a runtime deprecation.

text/0000-deprecate-property-lookup-fallback.md
```diff
@@ -0,0 +1,126 @@
...
+
+## What Other Frameworks Do
+React has used `this.props` to talk about values passed to you and `this.state` to mean data owned by the backing component class since it was released. However, this approach of creating a specific object on the component class to mean "properties available to the template", would likely be even more an invasive change and goes against the mental model that the context for the template is the class.
+
+Vue requires enumeration of `props` passed to a component, but the values in the template suffer from the ambiguity that we are trying to solve.
+
+Angular relies heavily on the dependency injection e.g. `@Input` to enumerate the bindings that were passed to the component and relies heavily on TypeScript to hide or expose values to templating layer with `public` and `private` fields. Like Vue, Angular does not disambiguate.
+
+## Introduce Yet Another Sigil
+We could introduce another sigil to remove ambiguity. This would address the concern about verbosity, however it is now another thing we would have to teach.
```
wycats: And in light of `this.` having a meaning in JS, and generally being perceived as relatively low-noise to JS developers, it seems worse to  introduce another sigil.

wycats:
It's worth adding "Do not change the default in Glimmer components" to the alternatives section.

wycats:
I'd also mention a codemod in the summary section, since without a codemod, this change will suck pretty hard.

For context, I think we could use a "runtime aided" codemod here, where we boot the app, compile all the templates, and enumerate exactly where the Glimmer compiler resolved a `{{foo}}` into `{{this.foo}}`. @rwjblue has suggested using Puppeteer to automate the process, which sounds awesome to me.

text/0000-deprecate-property-lookup-fallback.md
```diff
@@ -0,0 +1,126 @@
...
+
+As an aside, the ambiguity that causes confusion for human readers is also a problem for the compiler. While it is not the main goal of this proposal, resolving this ambiguity also helps the rendering system. Currently, the "runtime" template compiler has to perform a helper lookup for every `{{greeting}}` in each template. It will be able to skip this resolution process and perform other optimizations (such as reusing the internal [reference](https://github.com/glimmerjs/glimmer-vm/blob/master/guides/04-references.md)
+object and caches) with this addition.
+
+Furthermore, by enforcing the `this` prefix tooling like the [Ember Language Server](https://github.com/emberwatch/ember-language-server) does not need to know about fallback resolution rules. This makes common features like ["Go To Definition"](https://code.visualstudio.com/docs/editor/editingevolved#_go-to-definition) much easier to implement since we have semantics that mean "property on class".
```
wycats: In other words, it makes resolving a name to one of these purely local:

- local variable
- property on the component
- external helper or component

sohara:
I think this will represent an significant reduction in ambiguity in reading templates in Ember that will be especially helpful for developers new to the framework. Our main app is now 5 years old, currently on Ember 2.18 and I can see us adopting this very shortly after it lands.

text/0000-deprecate-property-lookup-fallback.md
```diff
@@ -0,0 +1,126 @@
...
+
+Since the `{{this.foo}}` syntax has worked in Ember.Component (which is the only kind of components available today) since the 1.0 series, we are not really in a rush to migrate the community (and the guides, etc) to using the new syntax. In the meantime, this could be viewed as a tool to improve clarity in templates.
+
+While we think writing {{this.foo}} would be a best practice for new code going forward, the community can migrate at its own pace one component at a time. However, once the fallback functionality is eventually removed this will result in a "Helper not found" error.
+
+## Syntax Breakdown
+The follow is a breakdown of the different forms and what they mean.
+
+- `{{@foo}}` is an argument passed to the component
+- `{{this.foo}}` is a property on the component class
```
Turbo87: why is it `this.foo`, but not `this.@foo`?

cibernox: My understanding is that `this.` makes reference to be a property in the component instance, and `@foo`-like arguments do not require a component to even exist, like it will happen in components without a javascript file, that become something similar to functional partials.

kellyselden:
Any reduced resolving complexity makes future tree-shaking of templates easier. Perhaps that can be added to motivations, or it's too handwavy and we can leave it out. Just thinking out loud.

locks:
I am for this RFC. It brings to mind the [Not Explicit](https://boats.gitlab.io/blog/post/2017-12-27-things-explicit-is-not/) blog post from a member of the Rust community. I like how the RFC allows us to know more about templates by reading solely their source.

I feel like the disambiguation and alignment with JavaScript syntax described by this RFC will also help prepare the transition to one-way bound Glimmer templates by widening the gap between properties of the component's JavaScript class, and arguments passed in.

balinterdi:
I like this, it will make reading templates much easier at a glance. Does this RFC address the same problem (helper name vs. property lookup) in controllers?

cibernox:
I like this a lot.
I am curious about how the codemod would work, as it needs to gather information in runtime (running the tests like _ember-cli-deprecation-workflow_?). 
I'm also wondering if, once the usage of `@arg` is enabled by default, the same codemod could be used for the exact opposite purpose and deprecate usage of `this.arg` for arguments passed from templates.

Gaurav0:
I'm worried the verbosity issue and deprecation spew will lead to a lot of complaints.

dwickern:
👍 My app just broke when I installed [ember-math-helpers](https://github.com/shipshapecode/ember-math-helpers). A template `<button onclick={{add}}>` started calling the `add` helper instead. Fortunately the helper throws an exception if you pass no arguments or I wouldn't have noticed.

chadhietala:
@cibernox I believe the plan will be to launch the app with Puppeteer and collect the information. Once that is done, use that information to inform the codemod. This should cover a very large portion of the cases.

vlascik:
This solves a non-existent problem (I haven't encountered it once, and I don't know of anyone that has, really - just don't name your properties after those 20, let's say 50 helpers tops you usually use - it's not that hard), and adds a huge amount of completely arbitrary verbosity to templates.
Please don't do it.

tim-evans:
I would greatly appreciate this— I've bumped into issues around this several times when making addons. For [ember-page-title](https://github.com/tim-evans/ember-page-title) I used a [handlebars plugin](https://github.com/tim-evans/ember-page-title/blob/latest/lib/plugins/translate-helper-name.js) to disambiguate calls between `{{this.title}}` (a variable) and `{{title}}` (a helper), by renaming all mustache statements with arguments to `{{title}}` to `{{page-title}}`.

✨ Thanks so much for putting this together @chadhietala ! ❤️

text/0000-deprecate-property-lookup-fallback.md
```diff
@@ -0,0 +1,140 @@
...
+## Syntax Breakdown
+
+The follow is a breakdown of the different forms and what they mean:
+
+- `{{@foo}}` is an argument passed to the component
+- `{{this.foo}}` is a property on the component class
+- `{{#with this.foo as |foo|}} {{foo}} {{/with}}` the `{{foo}}` is a local
+- `{{foo}}` is a helper
+
+# Drawbacks
```
joukevandermaas: A drawback I think is missing: copying old examples will now lead to deprecation warnings. This will probably confuse new-comers who are learning Ember. There are a *lot* of blogposts, stack overflow answers, video courses, etc. that probably use the old syntax. I think it's very confusing for beginners to see deprecation warnings when they just started following some basic examples.

andrewcallahan:
This makes a lot of sense, however it's making me think more about what to do about actions in the template.

If I was new to Ember and I had to call `{{this.property}}` to display a property, then I'd also expect that to call an action I would have to do something like `<button onclick={{this.actionProperty}}</button>`. Instead with this new proposal, we only have to explicitly call `.this` for a property `{{this.property}}` but not for a action, which would still use the non-`.this` syntax `<button onclick={{action 'action'}}</button>`.

That is pretty confusing to me. In order to keep things consistent and simple, I'd prefer only one way to access anything from the .js file using `.this`, which would including actions. So assuming the actions hash stays, to call an action using `.this` would be something like `<button onclick={{this.actions.actionProperty}}</button>`. That would make everything clear and obvious that every variable in the template is coming from the .js file.

However, I realize that we can't actually do this, since an action can refer to either an action defined on the actual .js file or an action that bubbles up to the controller. Therefore, I don't really have a solution, but am coming to the conclusion that actions face a similar ambiguity problem as properties do now, which this RFC has forced me to think about.

jenweber:
I am really on the fence from a dev UX perspective. On one hand, I like the clarity that distinguishes props passed in vs declared on the component.

On the other hand, I wonder to what extent our templates will be full of a ton of "this", reducing readability. The majority of my template variables would use "this" but I build complex weird stuff. In other people's apps, what percentage of template variables are passed in vs declared on the component? 

I also find "renaming" properties passed to child components to be a stumbling block to new devs. But it's possible the clarity that "this" brings would actually help. I'm going to ask some coworkers who are newer to Ember before I can say how I think it affects learning and/or how to teach it.

averydev:
I very much agree with the goal of disambiguating helpers from properties that are local to the backing `.js` file but it does seem to me that adding `this` adds a ton of noise to templates in a not great way. 

I think that the syntax proposed optimizes for the less used case (helpers) rather than the majority case (local properties) being that local properties make up a much larger volume of what is dynamic in a template than helpers.

I want to offer an alternative where instead we alter the syntax of helpers. This would build on existing syntax that folks are familiar with in handlebars and Ember: adding parenthesis for helpers rather than requiring `{{this.foo}}`. 

It's easiest to explain with an example. 
For instance, we're already doing the following in case of the commonly used ember-truth-helpers:
```
{{#if (eq 1 1)}}
 Truth!
{{/if}}
```

What if that were just extended that to when the helper is the first element as well? Here are a couple of options.
Option 1: 
`{{(capitalize 'some text')}}`
Combined with another helper
`{{(capitalize (concat 'some' 'text'))}}`
Or with hash arguments
`{{(capitalize text='some text')}}`

Option 2:
Just wrap the helper name in parenthesis
`{{(capitalize) 'some text'}}`
`{{(capitalize) (concat 'some' 'text')}}`
`{{(capitalize) text='some text')}}`

One downside is what to do about built in helpers like `{{link-to }}` or inline `{{if}}` I could see arguments for modifying those too, as well as arguments to treat them like reserved words. Personally I'd probably just leave them and say they're reserved. 

One nice thing about this is that parenthesis are associated with functions and that's not a bad thing in this case since it's a reasonable mental model.

Another bonus of changing helpers instead (either this way or some other way is that a code mod that was only looking for helpers would be much easier to pull off than one that replaced every local variable with `{{this.foo}}` and would impact MUCH less code.

averydev:
@andrewcallahan I've always found it hard to explain the difference between:
```
{{my-component
doSometthing = (action 'doSomething')}}
```
and
```
{{my-component
doSomething = (action doSomething)}}
```

I'd love to see the `@foo` syntax continue through to actions
```
{{my-component
doSometthing = (action doSomething)}} << local to the actions hash
```
and
```{{my-component
doSomething = (action @doSomething)}} << passed in
```

Turbo87:
> I think that the syntax proposed optimizes for the less used case (helpers) rather than the majority case (local properties) being that local properties make up a much larger volume of what is dynamic in a template than helpers.

This is also my main concern. I do understand the benefits of being explicit, but the increased verbosity is quite significant and has the potential to make people quite unhappy about being forced to use it.

Since components will use angle-brackets soon we can probably ignore those, so we only need to disambiguate between properties and helpers. I'm wondering if the fact that helpers usually take arguments would be enough to disambiguate? I'm trying hard to remember, but I don't think I've ever seen a helper invocation that did not take any arguments.

luxferresum:
> the increased verbosity is quite significant

I totally agree. Also helpers have the much more confusing syntax then properties, so if we change something we should change helpers. `foo=(concat bar baz)` always felt strage, I've written `foo=concat(bar baz)` so many times. And if we really want to align with `js` this would be the logical step. Then we could just require `()` for helpers, even when no arguments are given `foo=helpMe()` or `{{helpMe()}}`.
I'm not sure if there is a path forward to something like this, but in general I just think we should change helpers, not properties.


Also this would basically mean `@` means `this.args`, not only `args`. So in js I do `this.prop` and `this.args.arg`, but in the template `{{this.prop}}` and `{{this.args.arg}}`.

andrewcallahan:
Another option to consider is to keep the current syntax of calling properties in the template without `.this` the same - but in order to use a helper you must first import it and define it into the .js file of the component that uses it.

```
import eq from 'ember-truth-helpers'

  eq: eq
}
```

It's obviously much more work to use helpers this way, but it would both solve the ambiguity problem of helpers vs properties, and also allow the cleaner syntax of not using `.this` in the template.

chilicoder:
I believe requiring `this.` will bring too much clutter to our templates, especially complex ones. 

Approving this rfc as it is will make the situation when 100% of users will have to write more code for nothing only to solve problems that have 1% (or even less) users. 

I see two alternatives here
1) introduce different syntax for helpers. Something like `{{my-helper -> foo bar}}`.
2) instead of `{{this.firstName}}` take `{{.firstName}}`

Personally I like approach 1 more. Feels more expressive.

wycats:
It's worth calling out that this proposal is part of a series of proposals, and probably shouldn't be considered on its own, especially from the perspective of teaching and learning.

---

First, a quick divergence. Towards the end of last year, the core team discussed a problem we've had historically: it's difficult for us to make incremental progress on landing features, because we often feel the need to nail down the whole picture in high fidelity before we can make progress on the parts.

In an effort to make progress on some long-standing goals that the community has (notably Glimmer Components and plain JS Classes), we decided it would be acceptable to break the work into smaller RFCs. We wanted the RFCs to each add incremental value (to avoid thrashing in people's codebases), but  we also knew that one consequence of this would be that the smaller RFCs would, themselves, not always show their value in terms of the bigger picture.

One way that we have attempted to mitigate this problem has been to lean on the Glimmer.js experiment, which is an end-to-end sketch of what Ember APIs could look like, supporting the great majority of use-cases that Ember supports in the view layer.

The best way to look at these individual RFCs is by thinking of Glimmer.js as a "Living RFC" prototype for the big picture that you can play around with and even build production apps with. Because of how much ground it covers, it's still a sketch of a future vision, but because it works, it's an unusually high fidelity sketch.

The smaller Ember.js RFCs should therefore be seen as attempts to iterate towards the world sketched out in the Glimmer.js vision. (Aside: because of Ember's strong compatibility goals and RFC process, there will likely need to be changes to the Glimmer.js vision once we start trying to integrate, just as any big-picture vision gets tweaked as it gets implemented for real).

---

Back to talking about this RFC.

Glimmer.js templates have a number of differences from Ember templates, which bear on the conversation in this thread.

## Arguments are referred to via `@name`

In Glimmer templates, arguments are not "splatted" onto the component object as in Ember. Instead, you refer to arguments in templates via `@name`.

For example, if you invoke a component this way in a Glimmer.js template:

```hbs
<Accordion @title={{name}}>
  Some body text
</Accordion>
```

The `Tab` template would look like this:

```hbs
<div class="accordion">
  <h1>{{@title}}</h2>
  <section>{{yield}}</section>
</div>
```

RFC #276 proposed adding `@names` to Ember, and it was enabled in Ember 3.1 beta.

You can still write a computed property in your class to normalize some data passed into a component, but as of RFC #276, there is a strong distinction between using data passed "as-is" to the component, and special processing that you did in your component class. Speaking as an Ember user, figuring out "where the value came from" (was it passed "as-is" when the component was called, or did the component class do something to change what was passed in) is one of the bigger sources of cognitive overhead for me in reading templates. 

The consequence of this feature is that many places where you would have written `{{title}}` can be migrated to `{{@title}}`, which clearly indicates that you just have to look at the template that called the component to understand what `@title` means.

## Actions are not specially nested

Since Glimmer components are plain JS classes, there is no nested `actions` hash. Instead, normal JS methods turn into actions, and `{{action someMethod}}` simply binds the method to the component (so `this` works correctly).

The original motivation for the actions hash in Ember, for some historical context, is that people kept making `destroy` actions and getting confused about why that had bizarre behavior. I expect that in Ember, we'll add some sort of `@action` decorator on top of actions to help identify mistaken overrides, while at the same time auto-binding these methods for you.

But that's a topic for another RFC 😄 

Making actions "just methods" addresses another possible source of inconsistency in idiomatic Ember `(action 'foo')` that people have raised above.

## This RFC

Since the beginning of Handlebars, the `{{foo}}` syntax was ambiguous: it could refer to a helper named `foo` or `this.foo`. Handlebars also provided tools for being clear (explicit `this.` was always permitted) but it was not mandatory.

This problem has gotten a little worse as we have added more ways to "call" helpers and components to Glimmer. For example:

```hbs
{{#with (component some-component) as |c|}}
  {{! these mean different things }}
  {{c}}
  {{this.c}} {{! assuming that this.c is a contextual component }}
{{/with}}

{{this.c argument}} {{! call the component at this.c }}
{{c argument}} {{! call a helper named c }}
```

None of this is particularly "fatal" in the technical sense, but it makes the meaning of variable names much less regular. If you have the mental model that `{{c}}` is a shorthand for `{{this.c}}` and `{{this.c arg}}` works, it's reasonable to assume that `{{c arg}}` means the same thing.

We have tried various rules over time (the "dot rule" and similar rules in #226), but they add more confusion about the justification for those syntactic rules than they aid in understanding.

As described in the RFC, the most notable consequence of this in day-do-day use is that helpers shadow properties on the component, but that's really just the canary in a more pernicious coalmine.

Assuming that there's a `title` helper in scope:

```hbs
{{title}} {{! invokes the title helper }}
{{format title}} {{! passes this.title to format }}
```

It's not simply that `name` is shadowed by the helper, but also that the `title` name means something different depending on the syntactic context in which it's used.

We might be used to all of these quirks, but in my opinion they create much more teaching and learning confusion than the cost of the increased verbosity. As we have added block params, contextual components, named blocks and more features, the inconsistencies have become bigger and bigger.

## Syntactic Future

In this future world, Handlebars expressions mean the same thing no matter where they appear:

- `this.foo` looks up a property on the current component
- `foo` looks up a local variable in scope (helpers are in scope at the top level of components)
- `@foo` refers to arguments passed directly into the component

I think this world is much much easier to understand, teach and learn than the world we have today.

---

To circle back around to the beginning of this comment, one of the main consequences of the incremental approach to landing these RFCs is that the value of the sequence of RFCs is mostly in the combination of all of them. In fact, some of the individual RFCs might make things worse if they were accepted and implemented alone, without the rest of them.

The core team made an explicit decision to allow these RFCs to be approved and implemented in a more incremental way, because trying to combine them into a single, fully detailed RFC is impossible and was preventing the community from making progress along a path towards Glimmer components.

The Glimmer.js implementation more-or-less serves as a "vision statement" for the collection of RFCs. It isn't meant to be a high-fidelity spec, but rather a rough sense of the end state we're aiming towards. This particular RFC actually goes beyond what Glimmer.js did, largely because after making the other changes (`@names`, un-nested actions, tracked getters), it became more clear that the verbosity cost (in full context) was lower than it would have been in isolation.

One consequence of the incremental approach is that we have landed parts of the whole picture (such as `@names`) before it really makes sense to recommend that apps migrate to the new world.

Apps that attempt to track the features as they land will end up with a lot of churn as they are forced into workarounds that would be obviated a release or two later. On the other hand, enthusiastic early adopters can start using the features we have already landed, and because of Ember's usual compatibility story, can be adopted incrementally in existing apps (one `@name` at a time).

It's useful to think of this collection of features as currently "in-progress". We are approving RFCs and landing features as we implement them, with plans to refresh the docs to reflect idiomatic changes once the collection of features has landed. In the meantime, most Ember users should continue to use existing idioms, which work holistically today.

In terms of this RFC, the "syntactic future" described above is possible holistically pretty soon, and we should only flip on the deprecation once we:

1. have a high-fidelity codemod that can reliably update your app at once (and doesn't immediately introduce deprecation spew from addons)
2. are ready to update the guides to reflect the "syntactic future" described above

Sorry for the lengthy comment; it's probably somewhat overdue.

wycats:
I originally was going to put this inside the above comment but it got too long.

I picked a random medium-sized template from Handlebars and analyzed updating it with these changes: https://gist.github.com/wycats/7e187fdf995f29873adf8392edc16ecd

I found the results interesting: the wins came from disambiguating local variable lookup from property lookups in this case.

kellyselden:
@wycats It might make sense to switch the order around a bit. After RFC acceptance, immediately having the official docs/tutorials/blogposts use the new format, letting it sit for a while, then turn the deprecation on. This could let the syntax "settle" in the community, then the deprecations won't feel as jarring when they do start coming out.

wycats:
@kellyselden I think that's sensible as long as we've reached a "coherence checkpoint" with the list of features that we've landed.

That said, since we expect the codemod to be reliable, I don't think we need to wait that long.

chilicoder:
@wycats thank you for the broader explanation of the context.
Then, since we don't know the broader picture, the only wish I can express is not to indicate internal property with 5 characters. We use `@` to indicate external property, which is only 1 character.  Please, don't put much clutter into the templates.

vlascik:
Imho with changes like these, there should be a cost/benefit analysis:
1. Cost
- direct, in lines of code: greping through `travis-web`, it has about 500 computed properties. I'll assume some will be referenced multiple times, there will be component variables, some code got deleted, so on a project of similar size, `this.` would have to be typed ~1500 times.
I'll assume 5x this. = 1 line of code = $10-20; 300 lines of code = ~$3000 - $6000. Not a huge number, but not zero either.
- indirect - every copy/paste of a template from any article would instantly add a deprecation, which would be problematic, especially for new users. Every developer will have to familiarize himself with the deprecation, file issues, create PRs, wait for them to merge. I would say on the low side, everyone would have to spend 2-4 hours on this deprecation alone at least. 
The cost? [number of ember developers] * [average number of hours spent] * [hourly rate] = nice, round, probably six figure sum.
2. Benefit
- apps won't accidentally break if you add a new helper with conflicting name. I'd argue that this is a very rare occurrence, chances of conflict are very remote - there are 50 helpers in most applications tops, 90%+ of them are named as short verbs (inc, mult, take, slice). It's a common practice to name variables as nouns. Examples that were mentioned were "formatName" (should be formattedName), "title" (could have been named setTitle or pageTitle), and a variable named "add". 
Imho, this problem could be easily solved by a linter rule (no-helpers-without-arguments?) or an assertion inside the helper.
- added clarity for some - I don't think this has been an issue in last 7 years for most people, we would have heard about it a lot sooner otherwise. If it's just a question of style, it would be better solved by a linter rule.
I haven't seen any feature mentioned that would be impossible to implement without this breaking change yet.
Is it really worth the cost and effort of the community? Imho it isn't.

cibernox:
@vlascik I believe you are ignoring the fact that a codemod will be provided that will do it automatically. I do not know if it will cover 100% of the usages, but other codemods, like the one for the new qunit api was very good. I migrated a huge codebase in ~2h.

rtablada:
As an instructor that has worked with new developers, transitioning developers of all stripes, startups, and corporate teams this would definitely help clarify scoping and property use.
It is definitely clearer and cleaner to know where a property is coming from vs where helpers or global values are used.
It also does help keep consistency in applications.
There have been times where we've had to use `this.foo` in templates because of helper naming collision.
We've also run into odd times where there is helper vs variable collision between addons or engines which is much harder to solve as you don't have direct control of those templates.

While it is more verbose, this is a step that will help clarify usage and improve standardization within templates.

From a scoping and binding aspect, I can see that this could help improve VM and template optimizations since it helps to bind the actual available scope.

---

This all starts with education since `this.foo` works now with Ember and as we throw deprecations we do need to be wary of supporting these deprecations and the reasoning behind it.
But that learning is well worth the effort.

---

For teams that absolutely are dead set on not using a `this.` syntax I could foresee a build step transformation that adds `this.` before properties it knows to be locally scoped.
However this runs into an issue of not easily surfacing variable to helper collision.

Personally I think this build step would hide the underlying problem and as more work is done at lower levels to take advantage of  compile time scoping and bindings, this becomes more of an issue.

vlascik:
@cibernox I'm not, but you can only codemod existing code. And codemod won't fix copy/pasted code either. It's a hurdle, if there would be a good technical reason and no way around it, sure, why not, but I don't see this being the case.
@rtablada I don't think helpers are resolved at runtime, I would imagine this is done at compile. As for clarity, purely cosmetic changes should belong to a lint rule. Deprecation at a cost of breaking code in every article ever written is just too much, and really won't help learning either.

les2:
1. change helper syntax (e.g., `{{:a-helper}}`), or
2. allow explicitly importing helpers in template similar to ES module import, including aliasing the helper in the template to avoid name clash, or
3. allow importing helpers in component js and exposing as properties on the component instance, maybe via an injection helper like service injection - `aHelper: helper()` (don't like this option since it implies that template only commoners couldn't easily use helpers)

but please, please don't destroy the joy of handlebars templates by requiring `this.` for all property accesses.

1. it's not actually necessary to enable disambiguating helpers from properties - 3 alternatives listed above
2. the ember language server integration can be robustly implemented using alternatives above
3. properties are easily 10:1 more common than helpers in hbs - so we should optimize the ergonomics of the former at the expense of the latter if a trade-off is necessary.

$0.02

ro0gr:
An ability to access class members w/o need to wtite `this.` in C# made me really happy sometime.

I agree with existing literature and examples concern. Unfortunatelly we can't migrate all the literature with a codemod 😄

That's being said I'd prefer to make sure we can't disambiguate props vs helpers from the helpers side before moving to `this.` props. 

As far as I understand all the helpers should be known at build time. So maybe we can tweak helpers at build time to disambiguate it from props?

Like exising helper invocation:
```hbs
  <strong>{{capitalize "string"}}</strong>
``` 
During the build becomes:
```hbs
<strong>{{helper "capitalize" "string"}}</strong>
```

For the reference: there is a related [RFC](https://github.com/cibernox/rfcs/blob/contextual-helpers/text/0000-contextual-helpers.md) proposing `{{helper` and `{{invoke-helper` helpers.

btecu:
I agree with @Turbo87 and @averydev.

The helpers are much less common than component properties and if anything, those should be addressed. Most helpers have arguments so that should be enough but if more clarity is required why not do as @averydev suggested and include parenthesis?

cibernox:
@vlascik I do not think that the copy-pasting counterargument is realist.

If your codemod updates all your templates, copy-pasting within your project will work, and if you can go to stackoverflow and copy-paste a snippet of code which author used the exact same names for every single property than you, without making any change to it, I'd buy some lottery because that is truly your lucky day.

I buy the argument than having to modify all the templates in your app would be an awful process if performed manually, but if the codemod does a decent job, the real world impact should be pretty low.

As of the `this.` being considered noisy, that is indeed a downside. But in javascript code we type `this.` all the time and we don't really think a lot about it much. I think I'd could get used to it in a couple days. I most certainly prefer `this.`, which has a _very clear_ meaning for any javascript developer, than inventing a new special sigil to express a known concept.

For me the real option that could be added to the RFC, is between forcing `this.` on properties **VS** forcing wrapping helpers in parens (`{{(helper-name)}}`). I do favour the `this.` approach but it would be good to explain why this other apparently valid option is not the chosen one.

rtablada:
@vlascik there is some runtime resolution since there is ambiguity for late registered or runtime registered components and helpers.

This does create a cost.

I think while helper vs property is important, also this RFC has impact on locally scoped variables, blocks, and yielded variables. These actually pose just as much or more of a conflict and confusion.

wycats:
> For me the real option that could be added to the RFC, is between forcing this. on properties VS forcing wrapping helpers in parens ({{(helper-name)}}). I do favour the this. approach but it would be good to explain why this other apparently valid option is not the chosen one.

Good question.

I think it's worth reading my linked gist, which shows that there's another pretty common source of ambiguous syntax other than helpers.

The goal of this RFC is to make it possible to identify whether something is a property lookup "at a glance" (i.e. with only very local mental parsing). Today, `{{foo}}` might mean also mean a local variable, and I personally have gotten confused about this repeatedly (including just this morning, trying to show an example for this thread).

There are also other ambiguities, like:

```hbs
{{prop}} {{! this.prop }}
{{prop variable}} {{! invokes the helper prop }}
{{this.prop variable}} {{! invokes this.prop if it's a component }}

{{! but... }}

{{#with some.value as |prop|}}
  {{prop}} {{! local variable }}
  {{prop foo}} {{! call the local variable foo if it's a component }}
{{/with}}
```

In some circumstances, `{{prop}}` behaves identically to a local variable, but in other circumstances it doesn't.

This also causes some inconsistencies when using the named block style instead of yield.

Caller:

```
{{#tab-bar as |active|}}
  {{#if active}}
    This tab is active.
  {{else}}
    This tab is inactive.
  {{/if}}
{{/tab-bar}}
```

```hbs
{{@block this.active}} {{! invoke the 'default' block with this.active as a parameter }}

{{! same as }}

{{yield this.active}}

{{! also same as }}

{{this.block this.active}}

{{! but not the same as }}

{{block this.active}} {{! this invokes the helper named block }}
```

In other words, if you think of `foo` as a shorthand for `this.foo`, it doesn't always work, for a variety of reasons.

The TL;DR is that there are a number of inconsistencies that already crop up due to the helper problem, but are actually bigger than just helpers, and the mental confusion will become worse as we wrap up other long-standing goals, like named blocks.

For those of us who were around when we dropped context-shifting helpers for block params, I think the verbosity/clarity tradeoff is very similar:

```hbs
{{! old style }}

{{#each items}}
  <h1>{{title}}</h1>
  <div>{{body}}</div>
{{/each}}

{{! new style }}

{{#each items as |item|}}
  <h1>{{item.title}}</h1>
  <div>{{item.body}}</div>
{{/each}}
```

There's a lot more typing in the new style, but it's a lot clearer what's going on. It's easy to say the same thing about context-shifting helpers: most of the time people weren't referring to the outer context, so why impose all of this extra typing burden. It would have been possible to do a similar cost-benefit analysis: how often were people screwing up and meaning to refer to something in the outer context?

In retrospect, those concerns were wrong. The old style imposed a lot of cognitive overhead that we had gotten used to at the time, and imposed unnecessary programming-model workarounds like itemController that turn out to be completely unnecessary when using block parameters.

In our case, this manifests as addons doing things like using properties with multiple leading underscores to be extra sure that they won't conflict with helpers provided by the app, for example. It also manifests in the choices we use for helper names: you better not pick a name that collides with a (now or future) property, or you're gonna silently break some code.

These costs are already accounted for by existing users, but they are pervasive and expensive.

vlascik:
@cibernox I still remember starting out with ember, in the beginning I was copy/pasting code from tutorials all the time. Just take a look at e.g. http://yoember.com/, those examples would break in 50 places instantly. A lot of people start this way. Why make it harder if you don't have to.
@btecu actually, after greping travis-web, it seems helpers are used more often than local variables. Having a distinct syntax for helpers would be probably even more work.
@rtablada as far as I can tell, opcodes for Get and Helper are different in the wireformat, not sure what MaybeLocal means though, maybe it is meant for cases like these. The rest of the lookups that cannot be resolved at compile will probably be done at boot time only once, so I don't think the runtime cost will be that significant. There would probably have to be tens or hundreds of thousands of these for it to be noticeable, and that's not realistic, real number probably won't reach a hundred even on large apps.
@wycats thanks for more examples, but I honestly can't say I've seen these issues, if I had, the fix probably took 30 seconds and wasn't really that memorable. You really have to try to bump into most of them (then again, I've seen enough bad code in my lifetime to know better than name variables by three-five letter verbs, maybe it's different for others). It definitely can happen. I just don't think it does often enough to justify the churn. YMMV.

rtablada:
@vlascik I think you may be missing the idea of deprecation vs breaking change.

This RFC would introduce a deprecation which would be a warning for how and why the code would break in the future not how it is currently breaking. In terms of making it harder if you don't have to. This RFC mentions: 

> While it is not the main goal of this proposal, resolving this ambiguity also helps the rendering system. Currently, the "runtime" template compiler has to perform a helper lookup for every {{greeting}} in each template. It will be able to skip this resolution process and perform other optimizations (such as reusing the internal reference object and caches) with this addition.

> Furthermore, by enforcing the this prefix, tooling like the Ember Language Server does not need to know about fallback resolution rules. This makes common features like "Go To Definition" much easier to implement since we have semantics that mean "property on class".

These are definite improvements that could implement and utilize this new standard and change.

---

@wycats thanks for recognizing the cognitive steps that we've become accustom to.

---

Not sure that this has been mentioned much here, but the use of `this` definitely brings us more in line with other frameworks and can help adoption. Especially if used with single file components it is a much easier cognitive model to understand.

---

@ro0gr I also do a lot of C# development and appreciate the lack of need for `this`, but I also work in languages like Elixir and Ruby where it can be confusing where the use of `^` vs `@` is needed versus `self.` or even a mix of these.
While linting and static analysis can help with this, it is still a big cognitive barrier to match with.
Also, I think it's good to try and match scoping of JS where possible since this is the world that Ember developers will be most familiar with.

vlascik:
@rtablada if it's a noisy deprecation you will have to/want to solve it right away anyway. New users will get confused by deprecation spew just a little less than they would be by an error. Everyone will have to spend the same amount of time on it.
Tools can understand this logic, generally I don't think looking into helpers hash (which you already have to have) before variable hash (this is something e-l-s probably doesn't have anyway) is a problem.
Doing helper lookup at runtime doesn't sound right, the wire format has separate Get and Helper ops. I imagine most unambiguous calls will be resolved at compile time. For the rest, it should be possible to do it once at boot, in which case even if your app has 1000 {{variables}}, cost of lookups should negligible. I don't know the details of the implementation though, I might be wrong. 
Anyway, I'm not going to argue this point further, there's little to add anymore.

rtablada:
> if it's a noisy deprecation you will have to/want to solve it right away anyway

Yes, deprecations can be a bit scary, but there is tooling that we can do to make this easier.
The deprecation can/should link to the codemod to solve this.
From working with developers of all skills and experience this is a standard practice and something that is common in other ecosystems.

---

> I don't think looking into helpers hash (which you already have to have) before variable hash (this is something e-l-s probably doesn't have anyway) is a problem

This continues to be the case.
Currently look up is:

1. Components
2. Helpers
3. Local/scoped variables 
4. Property fallback

---

> I imagine most unambiguous calls will be resolved at compile time.

This is exactly what this RFC wants to solve.

The property fallback is ambiguous and is something built into the `get` operation in the Glimmer VM interop for Ember.

Right now this is limited to a few things that are ambiguous (I will label where they are ambiguous)

* Block Scope - Component Blocks, Helper Blocks etc (mostly can be determined at compile time)
* Bundled Helpers and Components (can be determined at compile time)
* Late Registered Helpers/Components (HAS to be determined at runtime)

This means that for every `{{foo}}` there is an ambiguity that has to be checked to determine that no late registered helpers or components called `foo` have been added.
The Op codes know they can skip compiled helpers and block scope variables, but there still needs to be a check for late registered helpers and components.

Using `{{this.foo}}` leaves no ambiguity the value is bound to the current `this` context.

The point of this RFC is also to open up new possibilities.
That means that we are making room for the future not just now.

vlascik:
Alright, if the resolution is a part of get, it should absolutely go. Do people really add helpers after initializers run? Hm.

rtablada:
Yeah, there are definitely usecases where async helpers are added. Even helpers added using initializers would be considered as late registered or runtime registered helpers.

rtablada:
@cibernox I think the `{{(helper arg arg)}}` is interesting but becomes then confusing because local attributes and properties can be functions so is `{{(helper arg arg)}}` really just clarity for helpers or render time function invocation in general?

I think there could be some UX improvements to function invocation, but that could be more achieved by syntax highlighting and beautification rules.

cibernox:
@vlascik My point was that if you are copying code from somewhere, you will be replacing `{{#each posts as |post|}}` to `{{#each pets as |pet|}}` (because you happen to be building an app for vets). There is no such a thing like an ember template you can copy without making modifications anyway, so at the same time you change `posts` => `pets` you can also change `posts` => `this.pets` with just a few extra chars. 

With a good deprecation message that tells you exactly what to do, I don't think this will be confusing, and definitively not after the second time you do it. The concept of `this.` is something we all understand.

The real issue is transforming existing usages automatically. If the codemod does a good job, this should be relatively painless.

averydev:
The more I think about it, I think I'm resistant to adding `{{this.var}}` just because I've been doing it the same way for so long. The point that we pepper `this.myVar` all over every .js file and don't think twice about it is a good indicator that this won't matter once we get used to it. 

I'm going to switch now, and tell y'all how it feels in a week :)

webark:
would having the resulting optimization be implimented, and the current fallback behavior be included as a sperate package be benificial? since `foo` and `this.foo` are both currently valid, this would enable a longer “early adopter” period to help ease past the “trough of disillusionment” and allow more time for addon authors, guides, posts, stackoverflow answers, etc, get updated to the new syntax, while allowing devs who opt in be able to get the benefits of the added clarity and optimization. Then the package that keeps the behavior could have more time before it adds the depreciation message, and potential cutoff of support.

miguelcobain:
Just wanted to added that ember guides could warn new users about the "copy pasting from online examples" problem.
And the deprecation log would do that as well. I think there are other stronger arguments for the "no this." side than this.

rwjblue:
@webark - I don't fully understand what you mean. Using `{{this.foo}}` (instead of `{{foo}}`) has been possible since Ember 1.0.0. It is absolutely possible to begin using what this RFC proposes in a completely backwards compatible way...

webark:
@rwjblue totally. maybe i was misunderstanding this RFC, but my understanding was the order of events would be, (some steps might happen simultaneously, but thinking the general exposure progress)

# Deprecate message added to core
# guides updated 
# code mod introduced
# Fallback behavior removed
# legacy fallback package added

instead i was trying to say what if it was more like..

# guides updated
# legacy fallback package added
# Fallback behavior removed
# code mod introduced
# Deprecate message added to legacy package

this would avail people to “opt in” earlier and push the depreciation message and the dev’s choice (product/qa dictation) to update their code down the pipeline.

Gaurav0:
Please note, If we follow semver, I do believe we will have to continue to support property fallback (with deprecation) until 4.0. Afterward, if a legacy support package is added, It will be considerably more time (until 4.5) before any performance improvement can be realized.

rtablada:
@Gaurav0 to some degree. For default yes this is true.

I think that in `@ember/optional-features` we might be able skip some of the ambiguity and have the VM skip the local property lookup.
Right now if you don't use `{{this.foo}}` or you don't reference something that the compiler knows is a reference to a component, helper, or local var: there's essentially a big `?` on how to look the value up (including property fallback).

If with `optional-features` the config says "I'm not using any ambiguous references" then we can optimize for that as an opt in before 4.5.

MelSumner:
While I can understand the reasons why that are presented in this RFC, and I understand that it's going to help things be better/move forward in Ember, I can anticipate some risks in enterprise environments (which likely have less to do with Ember and more to do with enterprise environments). Even if the migration tooling is perfect, this could be a bit challenging.

Gaurav0:
@MelSumner can you clarify? I know there's more code and lots of internal addons, but is there something else?

text/0000-deprecate-property-lookup-fallback.md
IAMZERG: I couldn't agree more @joukevandermaas.  That was one of my biggest hang-ups when I got started (around 1.13, still a noob).

jenweber:
I have added some additional data points to my [original comment above](https://github.com/emberjs/rfcs/pull/308#issuecomment-368190582). I did a little bit of digging through my templates and talked to a new-ish Ember dev, and most of my concerns have been addressed.

luxferresum:
> For those of us who were around when we dropped context-shifting helpers for block params, I think the verbosity/clarity tradeoff is very similar

I think thats a *very* good argument. I'm still not 100% convinced, but I think if we *really* need a shorter syntax for `this.`, we can always add this later.

lennyburdette:
As someone who wrote coffeescript for way too long, the `@foo` vs `this.foo` distinction is confusing. In coffee they mean the same thing. 

I can imagine a future where I prefer to be even more explicit and use `this.args.foo` and `this.bar`, forgoing any magic sigils. Are there any risks to that approach?

I’m otherwise in favor of the proposal, and I love the idea of a runtime-assisted codemod with puppeteer.

luxferresum:
@lennyburdette we need the sigil for angle bracket components to differ between arguments and HTML attributes, so we can do this:

    <my-component class="foo" @class=this.classForComponent />

and later spread the arguments inside `my-component` to a DOM node. While here we dont need the `@` its a logical assumption to use it:

    <div ...attributes>
      my class is {{@class}} and {{this.args.class}}
    </div>

ctusch:
My first reaction was also to keep addressing properties without a prefix and instead prefix helper invocations. The logic being that all variables in scope (be it template-local or the component level) should be accessed in a consistent manner. Of course passed-in variables already break that consistency.

After reading @wycats last post I agree with prefixing component properties with `this.`. But I think this still leaves too much ambiguity between local variables and helpers. So I think ultimately helpers should also be prefixed or otherwise distinguished.

I might add that I'm a big fan of explicitness.  😅

miguelcobain:
A real world example that just happened to me. ember-mdi addon uses `{{icon}}` to render the icon property. ember-leaflet addon exposes an `{{icon ...` helper to the app.
If you install both addons, ember-mdi will stop working because `{{icon}}` will resolve to the ember-leaflet helper.

This particular case could also be solved by https://github.com/emberjs/rfcs/pull/309

tomdale:
@lennyburdette:

> I can imagine a future where I prefer to be even more explicit and use this.args.foo and this.bar, forgoing any magic sigils. Are there any risks to that approach?

One downside of this is that it makes `args` dynamic, meaning a JavaScript component could expect to change the behavior of the `args` object and have that change reflected in the template. Currently, `@name` arguments take a fast path in the Glimmer VM. This will be even more important in a hypothetical WebAssembly future, where we can do almost all of the work for `@names` in WebAssembly but would have to round trip out to JavaScript quite a bit to reify `this.args.name`.

chadhietala:
After discussion with the core team this afternoon, we are 👍 on moving this RFC into the final comment period. We understand this is a large change and thus we have outlined what the [rollout strategy](https://github.com/emberjs/rfcs/pull/308/commits/515ceb3ef1afff321660df834af68c8a54442a84) will be.

kellyselden:
I think people will get used to the `{{this.x}}` syntax pretty quickly, but maybe I'm an optimist?

text/0000-deprecate-property-lookup-fallback.md
```diff
@@ -4,7 +4,7 @@
 
 # Summary
 
-Deprecate the fallback behavior of resolving `{{foo}}` by requiring the usage of `{{this.foo}}` as syntax to refer to properties of the templates' backing component. This would be the default behavior in Glimmer Components.
+Beging the transition to deprecate the fallback behavior of resolving `{{foo}}` by requiring the usage of `{{this.foo}}` as syntax to refer to properties of the templates' backing component. This would be the default behavior in Glimmer Components.
```
MelSumner: "Beging" -> "Beginning"  ?

btecu: "Beging" -> "Begin"

josemarluedke:
I have been using this syntax in one new project and I have to say that I love it. It makes the templates feel more natural.

rtablada:
@chadhietala looking at the video with Dyfactor, would it be possible to use this and run the test suite for the app? While this may not hit ever permutation of a template, for tested apps it could be a way to cover most of the application surface.

rwjblue:
@rtablada - Yep, thats definitely something that should work.

luxferresum:
Why removing the fallback at `4.3.0`? shouldn't it be the `4.0.0` major release that removes the fallback?

text/0000-deprecate-property-lookup-fallback.md
```diff
@@ -0,0 +1,162 @@
...
+- Introduce deprecation **app only** fallbacks
+
+**Phase 5:**
+- Introduce deprecation for **any** fallbacks
+
+**Phase 6:**
+- Rev major to 4.0.0
+
+**Phase 7:**
+- Remove fallback functionality post 4.3.0 LTS
```
rwjblue: The LTS would be 4.4.0, and the fallback could be removed in 4.5.0.

text/0000-deprecate-property-lookup-fallback.md
```diff
@@ -0,0 +1,162 @@
...
+- Enable the lint rule by default in the `recommended` config
+
+**Phase 4:**
+- Introduce deprecation **app only** fallbacks
+
+**Phase 5:**
+- Introduce deprecation for **any** fallbacks
+
+**Phase 6:**
+- Rev major to 4.0.0
```
rwjblue: Should mention here that the fallback would now assert unless the legacy addon is present.

luxferresum: THX, the hint for the legacy addon here is what I was missing.

knownasilya:
Can this be merged or does FCP need to be restarted?

rwjblue:
@knownasilya - We need to tweet out RE: the FCP (from the emberjs account), and I keep forgetting to get it done. AFAIK, there isn't anything blocking other than that procedure though...

rwjblue:
We reviewed this at todays core team meeting, and are still in favor of moving forward. This RFC process was super helpful in coming up with a rollout strategy that works for everyone, thank you!

bartocc:
there is no ember issue referenced for this RFC. Has the implementation work already started?

PR #309, opened 2018-03-02
Module Unification Namespaces

mixonic:
[Rendered](https://github.com/mixonic/rfcs/blob/mu-namespaces/text/0000-module-unification-namespaces.md)

### DIFF
--- a/text/0000-module-unification-namespaces.md
+++ b/text/0000-module-unification-namespaces.md
@@ -1,0 +1,510 @@
+# Module Unification Namespaces
+
+- Start Date: 2018-03-02
+- RFC PR:
+- Ember Issue:
+
+# Module Unification with Ember Addons
+
+## Summary
+
+Ember addons can use the module unification filesystem layout (i.e. an addon
+with a `src/` directory) in place of the legacy `app/` or `addon/` folders. The
+`src/` folder encapsulates a module unification namespace, and Ember
+applications can access components, helpers, and services from that namespace.
+Usually this takes the form of an invocation of the style
+`my-namespace::my-thing`.
+
+## Motivation
+
+[RFC #143](https://github.com/emberjs/rfcs/blob/master/text/0143-module-unification.md)
+introduced a new filesystem layout for Ember applications we've called "Module
+Unification". The RFC described semantics for "namespaces" and the concrete
+example of components from another namespace or within an addon namespace being
+invoked as `{{my-namespace::my-component}}` in the section [Addon
+modules](https://github.com/emberjs/rfcs/blob/master/text/0143-module-unification.md#addon-modules).
+
+The syntax of `::` allows us to specify an **explicit** namespace for an
+invocation. The lookup should only occur within the prefixed namespace. The RFC
+described this explicit behavior. If our design stops here, an addon author
+would need to type `{{my-addon::` before any component/helper/service
+invocations.
+
+This RFC makes the additional suggestion that addons have an **implicit**
+namespace. This makes writing an addon much like writing an app. Things in the
+`src/` directory are invoked by their names, and things in other namespaces
+require the `::` syntax.
+
+## Detailed design
+
+Only **services**, **components**, and **helpers** from an addon can be resolved
+by an application.
+
+To help the RFC be approachable I'll describe behaviors using plausible
+filenames. The actual implementation is based on module unification specifiers,
+so filenames used below may have alternative forms that are valid.
+
+An addon name of `gadget` or `gadget-lib` is used in these examples.
+
+Additionally I'll be using `{{` to invoke components in examples. It is implied
+that `(component` would work with the string version of the `{{` invocation.
+
+### Explicit namespaces
+
+#### Explicit namespaces in templates
+
+To invoke a specific item from an addon, the addon's namespace must be passed as
+a prefix to the invocation. For example an invocation of `{{gadgets::try-me}}`
+from an application would attempt each of the following ordered list:
+
+* With addon namespace:  `gadgets/src/ui/components/try-me/template.hbs`
+* With addon namespace:   `gadgets/src/ui/components/try-me/component.js`
+* An invocation with `::` should never be treated as a property. Throw.
+
+Since the presence of `::` makes it unambiguous that a component or helper is
+being invoked, the "dash rule" is loosened. For example an invocation of
+`{{gadgets::tryme}}` from an application would attempt each of the following
+ordered list:
+
+* With addon namespace:  `gadgets/src/ui/components/tryme/template.hbs`
+* With addon namespace:  `gadgets/src/ui/components/tryme/component.js`
+* An invocation with `::` should never be treated as a property. Throw.
+
+This permits Ember addons to have an in-template API that is isolated from
+conflict but not overly verbose. For example `{{animated::explode}}` would be
+valid as a component or helper, and within the same app you could have
+`{{mind::explode}}` available.
+
+#### Explicit namespaces in services
+
+Services likewise specify namespaces with a prefix. For example an injection via
+`inject(‘ember-simple-auth::session)` would resolve with these steps:
+
+* With addon namespace:  `ember-simple-auth/src/services/session.js`
+* No such service was found. Throw.
+
+#### Namespaces containing `ember-` or `ember-cli-`
+
+In [RFC #143: Module Unification - Main
+Modules](https://github.com/emberjs/rfcs/blob/master/text/0143-module-unification.md#main-modules).
+it was described that packages starting with `ember-` or `ember-cli-` would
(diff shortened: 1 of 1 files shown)


mixonic:
This was discussed in the Friday core meeting today. Highlights of what we covered and some notes:

  - Implicit namespaces in templates
  - Implicit namespaces in services
    - Can we close over the addon scope and have a special service inject for the addon?
    - Can we make a mini-rfc for throwing a build-time error
  - Invoking from an NPM scope
    - wycats doesn’t like the `@`, wants to just permit w/o `@`
    - Can we rely on `import` down the line
    - Can we rely on the component helper down the line
    - has a cost, people won’t want to use scoped packages
  - ember- and ember-cli- stripping
    - We have new consensus to not do this
  - Namespace and source APIs on owner, and the limits of those APIs (no way to address factory specific injections, options)
  - Resolver API, implement via expandLocalLookup 

I will make updates the the RFC today and leave a note when I do.

rmmmp:
> Only **services**, **components**, and **helpers** from an addon can be resolved by an application.

Was this always the case? I'm currently developing an internal addon which will expose some models. There's no blueprint for it in Ember CLI but I can manually create them myself. It works but I'm worried if I should abandon that approach. If it should continue to work, will the invocation also change? e.g. `store.findRecord('my-name-space::user')`

mixonic:
@rmmmp Regarding your specific use case: I do not suggest any change to Ember Data here. The RFC provides very specific examples of where the `::` syntax can be used. There is no implication you could pass it to `findRecord`, for example.

This RFC does outline support at the low level (`factoryFor` for example) for namespaces. Ember Data could choose to support namespaced models using the `::` syntax, and this doesn't explicitly disallow that.

However it also doesn't rush to add that support. First, it would be basically impossible to make all of Ember and it's various libraries work with `::` syntax. I'd like to keep our progress iterative. Second, I consider the `::` syntax a medium-term solution. Other parts of the Ember team have discussed formalizing absolute specifiers (a canonical ID for a given resolution) or leaning more on the naive ES `import` system to reference factories. That motivates the limited design space of this RFC.

mixonic:
The RFC has been updated:

* Invoking something from an NPM scoped package now requires use of the `{{component` helper. It is not suggested that you can invoke these packages directly as `{{@scope/package::name}}` since the leading `@` leads to some confusing wrt `@arg`, and because the `/` makes a closing tag difficult to process visually.
* Stripping of `ember-` and `ember-cli` is explicitly **removed**, and change compared to the original module unification RFC. Experienced developers would appreciate the stripping behavior, but new developers would certainly find it confusing and surprising.

Gaurav0:
> Only services, components, and helpers from an addon can be resolved by an application.

I think the reasoning behind this should be explicitly explained and added to the RFC. I think there are many use cases for this:

Some addons add new type(s) that must be automatically resolved. A prominent example is `ember-data`. With this limitation, they could not namespace implementations of those types.

A major enterprise company I used to work for chose to extend the DSL for routing. In order to implement, they created a custom resolver with virtual types which resolved to "cards" rather than routes. With this limitation, they could not combine this approach with namespaces.

Glimmer Components and other template invoked types that have not yet been fully specified may also require this.

ro0gr:
I'm a bit afraid of an idea to write in a such verbose way on ф regular base:

```hbs
{{#package-name::component-name as |smth|}}
  ...
{{/package-name::component-name}}
```

I think the most annoying part is a closing tag: `{{/package-name::component-name}}` . 

One of the regular build error I get related to invalid closing tag. With a package name it becomes even more complicated to write. Maybe good tooling may help here.

Also I'd like to have an ability to write angle bracket components without a namespace("`<some-package::`"?):
```html
<MyCoolComponent>
</MyCoolComponent>
```
I find it more familiar.

I'm cool with a necessity to be more verbose with look-ups in js but I'd prefer to leave templates ergonomic. If I have an ability to import namespaces I would choose this way.

sandstrom:
@mixonic Will there be an easy way of importing a package, say `wildlife` and then choose to make it available to the app as `wild` instead?

In other words, are we forced to use the NPM package name, or can we 'rename' them on import (something akin to JS modules `import foo as bar from 'foo'`).

csantero:
We have a large library of components in several npm-scoped addons, and it's very frequent in our apps' templates to see, given a component called `standard-table` provided by an addon called `@my-namespace/my-addon`:

```
   {{#standard-table}}
   {{/standard-table}}
```

Under module unification, it appears that we'd have to invoke this as:

```
    {{#component "@my-namespace/my-addon::standard-table"}}
    {{/component}}
```

I think it's safe to say nobody on our team is going to be happy to have to do this.

@sandstrom's suggestion would be very helpful for us. For example, if we could globally map `@my-namespace/my-addon` to just `my-addon`, then we could do:

```
    {{#my-addon::standard-table}}
    {{/my-addon::standard-table}}
```

rmmmp:
I second @csantero's concern. We're going to have the same issues in our apps and I can definitely say that it's not gonna be a fun experience for us.

mixonic:
I want to thank people for their feedback here and I encourage others to continue adding their reviews. EmberConf has kept me a bit busy 😬 but I have updates I want to make this weekend and I'll reply to comments as well. Thanks for your patience.

tim-evans:
Would it make sense to remove the npm scope if the app is in the same npm scope?

For example, at Condé Nast, we have a bunch of npm packages published with the `@condenast` scope. If our application `@condenast/cms` has the same scope as our addon library `@condenast/ember-ui`, would it be sensible to drop `@condenast`?

csantero:
> Would it make sense to remove the npm scope if the app is in the same npm scope?

@tim-evans That wouldn't actually work for our use-case since our apps (unlike our addons) don't use an npm scope. I guess we could change it but that would involve modifying a lot of files, as every ES6 import of app-defined modules would have to be rewritten.

gossi:
With all this explicitely I have strong tendency we are heading into a world where we are forced to read and write idiotic templates. MU and angle brackets for me are about clarity (file/folder-wise) and catching up to web components of just writing `<MyComponent>` and we are good to go. Addons are a good way for shared code, especially components. Now, if using a component forces me to write:

```
{{use my-name as MyName from '@scoped/pkg-name::my-name'}}
<MyName/>
```

then this is not a world, where I just can go with `<MyName>`. It may be more explicit in case another developer might be unclear about where that component is coming from, sure. Though for me as the main developer it will become very frustratring and exhausting repeating these imports over and over again. The DX will suffer and decrease with all these non-enjoyable-complex-explicit-statements.

The question for myself will rise, if I develop a MU or classic addon? MU for developing clarity of the new file structure or classic addon for an enjoyable consume of my addon/components of it?

In order to keep up DX and joy, while also denying auto-imported components and keep a bit of explicity by manually declaring imported components (helpers, services, ..., ?) with the chance of aliasing them (in case I depend on two addons that both have a component of the same name).
The idea is, to allow for a configuration file at application level where I can map imports to their desired names.

In a slack discussion with @knownasilya @lifeart and @ef4 this could be at `config/aliases.js` rsp. `config/aliases/(components,helpers,services).js` and could look like this:

```js
// config/aliases.js

module.exports = {
  '@myorg/myaddon': 'sausage',
  '@myorg/myaddon/components': {
    'my-component': 'MyComponent',
    'my-other-component': 'SomethingCompletelyDifferent'
  }
}
```

So you can still do `<sausage::table>`.

Also @ef4 suggested, this could be an addon at first, though required APIs should exist then.

ef4:
Another option for what @gossi is describing (whether as an addon or an eventually first-class feature) would be to create a preamble template. A preamble template would be an `hbs` file that works as if it was pasted into the beginning of all your other templates.

This has the benefit of not introducing a second syntax for doing the same thing. Your preamble would be able to just say `{{use my-name as MyName from '@scoped/pkg-name::my-name'}}` to make `MyName` available throughout the app.

This can obviously be abused, and I would be more comfortable shipping it as an addon first and seeing how people use it in real projects before making it a standard feature.

ro0gr:
I think today we have concept which looks similar to `use`. Contextual components is a way to yield configured components:
```hbs
{{#my-addon as |i|}}
  {{i.header}}
  {{i.body}}
  {{i.footer}}
{{/my-addon}}
```
It may be a bad idea but what if in MU world default components are supplied with a namespaced components? 

The only thing I don't like about the current way to access context is a reliance on order, so there is an imagined destruction syntax with angle bracket:
```html
<SomeAddon as { A } >
  <A>
</SomeAddon>
```
or even an inline(simple) form:
```html
<SomeAddon as { A } />
<AnotherAddon as { B } />

<A />
<B />
```
I feel there are lots of edges but I like this syntax.

mixonic:
Closed in favor of https://github.com/emberjs/rfcs/pull/367

PR #311, opened 2018-03-11
Introduce `<AngleBracketInvocationSyntax />`

chancancode:
[Rendered](https://github.com/emberjs/rfcs/blob/AngleBracketInvocation/text/0000-angle-bracket-invocation.md)

### DIFF
--- a/text/0000-angle-bracket-invocation.md
+++ b/text/0000-angle-bracket-invocation.md
@@ -156,10 +156,12 @@ an internal block, and `...attributes` is the syntax for yielding to this
 internal block. Since the `yield` keyword is not available in element position,
 a dedicated syntax is needed.
 
-> Open Question: Should classic components (`Ember.Component`) implicitly add
-> `...attributes` to the hidden wrapper element? If not, should it be an error
-> for developers to add `...attributes` to a classic component's template? Are
-> either of those approach going to create a migration/compatibility risk?
+Classic components (`Ember.Component`) will implicitly have an `...attributes`
+added to the end of the wrapper element (if `tagName` is not an empty string),
+after any attributes added by the component itself (using `attributeBindings`,
+`classNames` etc). This means that attributes provided by the caller will
+override (replace) those added by the component (except for `class`, which is
+merged).
 
 ### Block
 
@@ -179,12 +181,7 @@ that can be passed using the named blocks syntax.
 The last piece of the angle bracket invocation syntax is the closing tag, which
 is mandatory. The closing tag should match the tag name portion of the opening
 tag exactly. If no block is passed, the self-closing tag syntax `<FooBar />`
-can also be used.
-
-> Open Question: What should `{{has-block}}` return for components invoked with
-> the self-closing tag? On one hand, returning true would be quite counter-
-> intuitive; on the other hand, returning false would mean self-closing tags are
-> not semantically equivalent to an immediately closed tag.
+can also be used (in which case `{{has-block}}` will be false).
 
 ### Dynamic Invocations
 
--- a/text/0000-angle-bracket-invocation.md
+++ b/text/0000-angle-bracket-invocation.md
@@ -104,7 +104,7 @@ be normalized using the `dasherize` function, which is the same rules used by
 existing use cases, such as service injections. This allows existing components
 to be invoked by the new syntax.
 
-Another benefit of the captial letter rule is that we can now support component
+Another benefit of the capital letter rule is that we can now support component
 names with a single word, such as `<Button>`, `<Modal>` and `<Tab>`.
 
 > Note: Some day, we may want to explore a file system migration to remove the
@@ -183,7 +183,7 @@ can also be used.
 
 > Open Question: What should `{{has-block}}` return for components invoked with
 > the self-closing tag? On one hand, returning true would be quite counter-
-> intuitive; on the other hand, returning false would mean self-closing tags and
+> intuitive; on the other hand, returning false would mean self-closing tags are
 > not semantically equivalent to an immediately closed tag.
 
 ### Dynamic Invocations
@@ -231,7 +231,7 @@ The following example will be a syntax error:
 
 > Note: The [named blocks RFC](https://github.com/emberjs/rfcs/blob/master/text/0226-named-blocks.md)
 > proposed to use the `<@foo>...</@foo>` syntax on the invocation side to mean
-> proivding a block named `@foo`. This will create a conflict with this RFC.
+> providing a block named `@foo`. This will create a conflict with this RFC.
 > Since the named blocks RFC has not been implemented yet, we propose to change
 > that syntax to avoid the conflict, for example, `<@foo=>...</@foo>` or
 > `<Block @foo>...</Block>`.
--- a/text/0000-angle-bracket-invocation.md
+++ b/text/0000-angle-bracket-invocation.md
@@ -144,7 +144,7 @@ is a new feature that is not available in the classic invocation style.
 
 These attributes can be accessed from the component template with the new
 `...attributes` syntax, which is available only in element positions, e.g.
-`<div ...atributes />`. Using `...attributes` in any other positions, e.g.
+`<div ...attributes />`. Using `...attributes` in any other positions, e.g.
 `<div>{{...attributes}}</div>`, would be a syntax error. It can also be used on
 multiple elements in the same template. If attributes are passed but the
 component template does not contain `...attributes` (i.e. the invoker passed
--- a/text/0000-angle-bracket-invocation.md
+++ b/text/0000-angle-bracket-invocation.md
@@ -207,34 +207,36 @@ Because `Option` is the name of a local variable (block param), the `<Option>`
 invocation will invoke the yielded value instead of looking for a component
 named "option".
 
-More generally, the tag name portion can be any path expressions:
+To avoid collision with HTML, only local variables starting with a capital
+letter can be invoked this way. No other forms of expressions are allowed in
+this position at this moment:
 
 ```hbs
-<this.foo>...</this.foo>
-<@foo>...</@foo>
+{{!-- these do not work! --}}
+<this.foo />
+<@foo />
 
 <MyForm as |f|>
-  <f.input>...</f.input>
+  <f.input />
+  <f />
 </MyForm>
 ```
 
-The only exception is that it is illegal to invoke a local variable starting
-with a lower case letter without a path segment, as it will conflict with HTML.
-The following example will be a syntax error:
+This restriction may be relaxed in the future. In the time being, the `let` can
+be used to rename these into an invocable form:
 
 ```hbs
-{{!-- don't do this! --}}
-<FooBar as |foo|>
-  <foo />
-</FooBar>
+{{#let ... as |FooBar|}}
+  <FooBar />
+{{/let}}
 ```
 
 > Note: The [named blocks RFC](https://github.com/emberjs/rfcs/blob/master/text/0226-named-blocks.md)
 > proposed to use the `<@foo>...</@foo>` syntax on the invocation side to mean
-> providing a block named `@foo`. This will create a conflict with this RFC.
-> Since the named blocks RFC has not been implemented yet, we propose to change
-> that syntax to avoid the conflict, for example, `<@foo=>...</@foo>` or
-> `<Block @foo>...</Block>`.
+> providing a block named `@foo`. This will create a potential conflict with
+> this future extension. Since the named blocks RFC has not been implemented
+> yet, we propose to change the block-passing syntax to `<@foo=>...</@foo>` to
+> avoid any confusion and potential conflict.
 
 ## How we teach this
 
--- a/text/0000-angle-bracket-invocation.md
+++ b/text/0000-angle-bracket-invocation.md
@@ -204,36 +204,145 @@ Because `Option` is the name of a local variable (block param), the `<Option>`
 invocation will invoke the yielded value instead of looking for a component
 named "option".
 
-To avoid collision with HTML, only local variables starting with a capital
-letter can be invoked this way. No other forms of expressions are allowed in
-this position at this moment:
+Similar to curly invocations, most valid Handlebars path expressions are
+invokable in this manner:
 
 ```hbs
-{{!-- these do not work! --}}
-<this.foo />
-<@foo />
+{{!-- LOCAL VARIABLES --}}
 
-<MyForm as |f|>
-  <f.input />
-  <f />
-</MyForm>
-```
+{{#form-for model=user as |f|}}
+  {{f.fieldset}}
+    {{f.input name="username" type="text"}}
+    {{f.input name="password" type="password" }}
+  {{/f.fieldset}}
 
-This restriction may be relaxed in the future. In the time being, the `let` can
-be used to rename these into an invocable form:
+  {{!-- is equivilant to --}}
 
-```hbs
-{{#let ... as |FooBar|}}
-  <FooBar />
-{{/let}}
+  <f.fieldset>
+    <f.input @name="username" @type="text" />
+    <f.input @name="password" @type="text" />
+  </f.fieldset>
+{{/form-for}}
+
+{{!-- NAMED BLOCKS OR CURRIED COMPONENTS --}}
+
+{{@content}}
+
+{{!-- is equivilant to --}}
+
+<@content />
+
+{{!-- THIS LOOKUP --}}
+
+{{#this.container}}
+  {{this.child}}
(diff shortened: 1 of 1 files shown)


sclaxton:
I’m pretty opposed to overriding the proposed syntax for passing named blocks structurally, i.e. `<@foo></@foo>`. In my mind this syntax is pretty important for representing the mental model that named blocks are really just the same as passing the block as an argument. It also seems more natural syntactically that thing inside naked angle brackets should treated as static names rather than dynamic expressions. Could we not instead have dynamic invokation be expressed with curlies? `<{{@foo}}></{{@foo}}>` seems to convey dynamic-ness to me more than `<@foo></@foo>`.

mmun:
> In my mind this syntax is pretty important for representing the mental model that named blocks are really just the same as passing the block as an argument.

I strongly agree with this and it's one of the reasons why I prefer the `<@foo=>...</@foo>` block-passing syntax (with the equal sign) over bare `<@foo>...</@foo>`.

The second reason is that the `<@foo>...</@foo>` block-passing syntax prevents us from having a `<form.submit>...</form.submit>` syntax for invoking contextual components. This feature is very important to me. When I've worked with JSX in the past I've really app I don't want to be forced to write `<{{f.input}}>`

I think it is less clear  we were to use `<@foo></@foo>` to pass blocks
The other reason is that it's very important to me that I don't need to write `<{{f.input}}>` to

mmun:
> In my mind this syntax is pretty important for representing the mental model that named blocks are really just the same as passing the block as an argument.

I strongly agree with this and it's one of the reasons why I prefer the `<@foo=>...</@foo>` block-passing syntax (with the equal sign) over bare `<@foo>...</@foo>`.

The second reason is that the `<@foo>...</@foo>` block-passing syntax prevents us from having a `<form.submit>...</form.submit>` syntax for invoking contextual components. This feature is very important to me. When I've worked with JSX in the past I've really enjoyed the simplicity of being able to do this.

sclaxton:
@mmun those are fair points, however it still feels weird to me to allow dynamic expressions outside of curlies. Can someone speak to the rationale behind that decision? With the exception of maybe the `...arguments` keyword in glimmer it seems like we’ve kept dynamic expressions within curlie braces, this seems like a pretty big departure from that.

tchak:
Is it really not possible to support positional params?

text/0000-angle-bracket-invocation.md
```diff
@@ -0,0 +1,259 @@
...
+component template does not contain `...attributes` (i.e. the invoker passed
+some attributes, but the invokee does not take them), it will be a development
+mode error.
+
+It could be thought of that the attributes in the invocation side is stored in
+an internal block, and `...attributes` is the syntax for yielding to this
+internal block. Since the `yield` keyword is not available in element position,
+a dedicated syntax is needed.
+
+> Open Question: Should classic components (`Ember.Component`) implicitly add
```
joukevandermaas: I'm assuming 'classic component' is a reference to the component manager RFC? Initially this paragraph was (maybe still is) very confusing to me. I guess I can understand the point of the question if I assume the following:

1. 'Classic components' (current-day Ember components) do not support `...attributes`, so throwing an error when they are not used seems wrong (at least initially)
2. Glimmer components do support it so they will work as described in the paragraphs above

Can someone please correct my understanding if it's wrong?

text/0000-angle-bracket-invocation.md
```diff
@@ -0,0 +1,259 @@
...
+
+The last piece of the angle bracket invocation syntax is the closing tag, which
+is mandatory. The closing tag should match the tag name portion of the opening
+tag exactly. If no block is passed, the self-closing tag syntax `<FooBar />`
+can also be used.
+
+> Open Question: What should `{{has-block}}` return for components invoked with
+> the self-closing tag? On one hand, returning true would be quite counter-
+> intuitive; on the other hand, returning false would mean self-closing tags and
+> not semantically equivalent to an immediately closed tag.
```
joukevandermaas: I really don't get this. They *are* not semantically equivalent. An immediately closed tag has a block, but it's empty. Is there some technical reason why they should be the same?

mmun: I agree. `<Foo></Foo>` should have a block with no content whereas `<Foo />` should not have a block passed at all (similar to `{{some-component}}`).

mmun: The RFC was alluding to some of the technicalities around self-closing tags in the HTML spec, but I think it's ok to make small refinement to the semantics for angle bracket component invocations.

mmun:
@tchak My gut feeling is that positional arguments don't really fit well the angle bracket component invocation syntax, but I'd be interested to hear compelling use cases or syntax ideas. Keep in mind that any syntax ideas should be compatible with https://github.com/emberjs/rfcs/pull/112.

text/0000-angle-bracket-invocation.md
mmun: Yes, that seems correct.

I think that classic components should fully support `...attributes` and that we should automatically insert `...attributes` to the wrapping element if it doesn't appear in the template and `tagName: ''` hasn't been used.

This would allow people to immediately start using HTML attributes with their classic components and remove the need for hacky add-ons like https://github.com/mmun/ember-component-attributes.

averydev:
I'm curious whether the long term goal is for there to only be the angle bracket syntax, or whether there would always be curlies and angle bracket options?

Gaurav0:
As a practical matter, I would like to be able to convert to angle bracket syntax, especially for the one way binding advantages. However, my large work app has lots of nested components and not being able to traditionally nest components in directories pretty much means we will have most of our components remain curly components indefinitely. I think this is pretty much true of any large enterprise app.

I think a way to invoke components nested in directories globally is absolutely necessary if large enterprise apps are ever going to be able to migrate. While it is possible to hack the resolver, I would prefer a shared solution.

text/0000-angle-bracket-invocation.md
```diff
@@ -0,0 +1,259 @@
...
+```
+
+Because `Option` is the name of a local variable (block param), the `<Option>`
+invocation will invoke the yielded value instead of looking for a component
+named "option".
+
+More generally, the tag name portion can be any path expressions:
+
+```hbs
+<this.foo>...</this.foo>
```
sclaxton: What types are permissible for `this.foo` in this case? Curried component?

courajs:
@chancancode this proposal doesn't include mention of the one-way binding semantics @Gaurav0 mentions, correct? That was part of the original angle brackets proposal, but has been removed here?

courajs:
I'm slightly concerned about how the capitalization interacts with native HTML elements. Right now, `<Button>Press</Button>` will render the native browser button tag. Does this need to come with the specification that native elements must always be specified lower-case in templates?

courajs:
I think `{{has-block}}` should probably return false for self-closing tags, and true for immediately-closed tags. Then, these would correspond more closely with curly invocation - self-closing is the inline invocation form, and immediately closed is the block form.

Since a lot of components use `has-block` to render default content, there should be a way to send no block. If it returned true for self-closing tags, we'd need to specify some other way to pass no block. Currently you'd have to do something like `<Comp @main={{null}} />`, but adding some other additional syntax seems likely to feel awkward as well.

Panman8201:
> I think `{{has-block}}` should probably return false for self-closing tags, and true for immediately-closed tags.

Agreed, it seems more semantically correct. IMO, if it returns false for immediately-closed tags then it would be more appropriately named `{{has-content}}`. Perhaps add both for additional state information.?.

chancancode:
I have updated the RFC to reflect a few things we discussed in the EmberConf core team meeting:

* For now, `<Foo />` (where `Foo` is a local variable) will be the only kind of dynamic invocation allowed. This makes all component invocations uniformly start with a capital letter and allow for more time to collect feedback on the other forms of dynamic invocations.

* Classic components will have `...attributes` added to the wrapper element.

* `<Foo />` will have `{{hasBlock}}` set to false.

chancancode:
> I'm slightly concerned about how the capitalization interacts with native HTML elements. Right now, <Button>Press</Button> will render the native browser button tag. Does this need to come with the specification that native elements must always be specified lower-case in templates?

That is technically true, but I think it is very rare and almost never happens in practice. If we want to be super careful, we can put in a warning/error now to clear the space before rolling it out. Are you asking because you have seen this in the wild?

chancancode:
> I'm curious whether the long term goal is for there to only be the angle bracket syntax, or whether there would always be curlies and angle bracket options?

@averydev As stated in the RFC, curly invocation is here to stay. I think in the long term, curlies will be primarily used for things that are like control-flow (e.g. `animated-each`).

chancancode:
@Gaurav0 @courajs This proposal does _not_ include the one-way binding semantics from the original proposal. This proposal is really just just an alternative syntax for the existing invocation style (with some feature differences – one has positional arguments, the other has attributes).

The two-way binding semantics generally "doesn't matter" as long as you don't use it in the component classes. To the extent that (and in the cases where) they matter, they affect curly invocations just as much, and since we are proposing that they are both here to stay and there are good use cases for both, it is not really feasible to fix one but not the other.

I think we have better tools to fix that now (optional features, deprecations, etc), so I am pretty optimistic that we can find a way to fix that broadly. But _if_ we somehow cannot do that, the divergent in semantics between the two invocation styles will be more of a problem in the long term.

rwjblue:
> the divergent in semantics between the two invocation styles will be more of a problem in the long term

I strongly agree. The current proposal supports gradual migration and straight-forward interop over broad sweeping changes. This will ultimately translate into fewer "cliffs" for existing applications...

wycats:
@rwjblue @chancancode I agree with this. Since we don't think we're going to remove "curly invocation", and plan to keep it around for control-flow style abstractions (liquid-if), we're going to have to deprecate auto-mut for those cases anyway.

Additionally, Glimmer component classes do not see two-way bindings, since the Glimmer component manager doesn't create two-way properties on the component class, and also sets an immutable `args` object. The fact that angle-bracket components still nominally create two-way bindings is more-or-less a consequence of the fact that angle bracket components can be used to invoke classic component classes in this proposal.

(Aside: given this, we will need to decide how `<Input>` should be written using Glimmer component classes. We don't want to reify `mut` onto `args` today because of auto-mut, but maybe we would be willing to reify it after other changes?)

Gaurav0:
Please specify if <A/b> will be allowed to invoke nested components.

rwjblue:
> Please specify if `<A/b />` or `<A/b>...</A/b>` will be allowed to invoke nested components.  

This is not supported, I agree the RFC should be a tad more explicit in the `Tag Name` section.

rwjblue:
@Gaurav0:

> As a practical matter, I would like to be able to convert to angle bracket syntax, especially for the one way binding advantages. 

There is no behavior change associated with angle bracket invocation. The support of two way bindings is entirely based on the base class of your component. This RFC does not propose to change that at all.

> However, my large work app has lots of nested components and not being able to traditionally nest components in directories pretty much means we will have most of our components remain curly components indefinitely. I think this is pretty much true of any large enterprise app.

I agree that this complicates migration. It seems absolutely fine to continue using curly invocation (with nested directory structures) indefinitely! There is no proposal to deprecate or make curly component invocation "second class". 

Please note though, that there are a number of ways that you can obtain grouping behaviors already (and even more on the way!). For example, you could use a file-name prefix along with in repo addons as a mechanism to namespace components (both physically and in the template). 

If you look ahead a bit, you can see there are even more tools to aid in this. The new file system layout (being actively pushed forward by the community) provides both lighter weight namespaces and a very clear way to invoke them from the template. In addition to light weight namespaces, we are also adding local lookup capabilities. When leveraging local lookup you can easily scope components to a specific subdirectory so that they are only available within specific routes or components. The combination of these two things provide an even better mechanism for grouping that folders do!

> I think a way to invoke components nested in directories globally is absolutely necessary if large enterprise apps are ever going to be able to migrate. 

I agree to disagree with you. 😺 I believe the new file system layout with both namespaced packages and local lookup are **wonderful** paths forward to avoid that nested directory invocation.

> While it is possible to hack the resolver, I would prefer a shared solution.

I agree that modifying the resolver is not a good long term strategy. 

> Note: The current RFC draft does not explicitly forbid slashes. 

Yep, absolutely should be updated to make it much clearer.

> This should be added if they are not going to be supported as previously expected.

This statement is pretty confusing.  I am not aware of **any** prior proposal that allowed nested directory invocation.

lmcardle:
What’s the plan to support contextual components. I agree with @mmun, this is an important feature with components today.

mmun:
I still want a follow-up RFC for `<f.input>` but it still needs works and its not necessary for merging the simple angle bracket invocation style. In the mean time {{f.input}} will continue to work...

lmcardle:
@mmun, you can do this in jsx? I’m curious how they accomplish this, do you have a reference by any chance?

mmun:
@lmcardle It wouldn't be using JSX, just inspired and validated by it. It would be implemented by changing the Glimmer parser.

text/0000-angle-bracket-invocation.md
```diff
@@ -204,36 +204,145 @@ Because `Option` is the name of a local variable (block param), the `<Option>`
...
 
-This restriction may be relaxed in the future. In the time being, the `let` can
-be used to rename these into an invocable form:
+> Note: The [named blocks RFC](https://github.com/emberjs/rfcs/blob/master/text/0226-named-blocks.md)
+> proposed to use the `<@foo>...</@foo>` syntax on the invocation side to mean
+> providing a block named `@foo`, which creates a conflict with this proposal.
+> [RFC #317](https://github.com/emberjs/rfcs/pull/317) propose to change the
+> block-passing syntax to `<@foo=>...</@foo>` to avoid this conflict.
+
+Notably, based on the rules laied out above, the following is perfectly legal:
```
mmun: typo: laid

text/0000-angle-bracket-invocation.md
```diff
@@ -0,0 +1,367 @@
...
+```
+
+## Motivation
+
+The original [angle bracket components](https://github.com/emberjs/rfcs/pull/60)
+RFC focused on capitalizing on the opportunity of switching to the new syntax
+as an opt-in to the "new-world" components programming model.
+
+Since then, we have switched to a more iterative approach, favoring smaller
+RFCs focusing on one area of improvment at a time. Collectively, these RFCs
```
alexlafroscia: Should be `improvement`

text/0000-angle-bracket-invocation.md
```diff
@@ -204,36 +204,145 @@ Because `Option` is the name of a local variable (block param), the `<Option>`
...
+Second, while Handlebars technically allows `{{foo/bar}}` as an equivilant
+alternative to the `{{foo.bar}}` path lookup (and therefore `foo/bar` is
+technically a valid Handlebars path expression), it will not be supported in
+angle bracket invocation. This is both because the `/` conflicts with the HTML
+closing tag syntax, and the fact that Ember overrides that syntax with a
+different semantic.
+
+In today's semantics, `{{foo/bar}}` does not try to lookup `this.foo.bar` and
+invoke it as a component. Instead, it is used as a filesystem scoping syntax.
+Since this feature will be rendered unnecessary with [Module Unification](https://github.com/emberjs/rfcs/blob/master/text/0143-module-unification.md),
```
Gaurav0: Disagree that this feature is unnecessary. 📁 are the most natural way to organize components and are widely used in large applications. None of the alternatives presented in module unification are intuitive or easily transitioned to.

rwjblue:
After discussion with the core team this afternoon, we are 👍 on moving this RFC into the final comment period.

oligriffiths:
Can I ask why `<Foo @bar=123 />` is not supported, that the value must be {{123}}?

chilicoder:
> `<FooBar @foo={{123}} @bar={{this.bar}} />`
I can imagine myriad of `@` in my components) can I get `<FooBar {{foo=123 bar=this.bar}} />` please?

rmmmp:
@chancancode I didn't see anything that mentions `{{yield to="inverse"}}`. I'm guessing syntax would be the same?

e.g.

```hbs
<Grid as |item|>
  <GridItem @item={{item}}>
{{else}}
  No item available
</Grid>
```

rwjblue:
@rmmmp:

> I didn't see anything that mentions {{yield to="inverse"}}. I'm guessing syntax would be the same?

Quoting from the RFC:

> There is no dedicated syntax for passing an "else" block directly. If needed, that can be passed using the named blocks syntax.

chadhietala:
Since 14 days have passed with no additional unaddressed comments, I'm going to merge this RFC.

bravo-kernel:
Should the label not be removed?

PR #314, opened 2018-03-22
HTML Attribute and Property Rationalization

chadhietala:
[Rendered](https://github.com/emberjs/rfcs/blob/c09ddf39a5db024cb44aa0ae089c8cf250e5a805/text/0000-attribute-property-rationalization.md)

### DIFF
--- a/text/0000-attribute-property-rationalization.md
+++ b/text/0000-attribute-property-rationalization.md
@@ -66,20 +66,73 @@ This proposal:
 - brings clarity and [addresses confusion](#appendix-a) as how to set a property on an element
 - enables the usage of [custom elements](https://custom-elements-everywhere.com/)
 - fixes HTML serialization issues for server-side rendering
-- removes the need for `TextSupport` mixin for supporting input events
-- removes the need for `{{textarea}}` and `{{input}}`
 
 ## Detailed design
 
-To move to this more explicit syntax in a backwards compatable we will convert well-known properties to use the `{{prop}}` and `{{on}}` modifiers via a codemod and retain the runtime setting rules for `attributeBindings`.
+To move to this more explicit syntax in a backwards compatible we will convert well-known properties to use the `{{prop}}` and `{{on}}` modifiers via a codemod and retain the runtime setting rules for `attributeBindings`.
 
 ### Transforming Through Codemod
-Due to the declarative nature of HTML we are able to see exactly which attributes are being set on which elements. This allows us to understand where we need to set properties and where we can safely rely on setting attributes. Using the list of [HTML attributes](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes) and [WAI-ARIA attributes](https://www.w3.org/TR/wai-aria-1.1/) we can then determine which attributes in templates need to be converted to use either `{{prop}}` or `{{on}}`.
+Due to the declarative nature of HTML we are able to see exactly which attributes are being set on which elements. This allows us to understand where we need to set properties and where we can safely rely on setting attributes. Using the list of [HTML attributes](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes) and [WAI-ARIA attributes](https://www.w3.org/TR/wai-aria-1.1/) we can then determine which attributes in templates need to be converted to use either `{{prop}}` or `{{on}}`. We will create a "best effort" codemod to transform all of these cases by default.
 
 ### Runtime Rules
 With Ember Components we have the notion of `attributeBindings` to set attributes on the wrapping element as defined by the `tagName` on the component class. For this case we will continue to use the setting rules we have today. Notably, [Glimmer Components](https://glimmerjs.com/guides/templates-and-helpers) have adopted the "Outer HTML" semantics long ago and the experience has been very positive. This means that when Glimmer Components become available in Ember they would be covered by the codemod as described above.
 
-Furthermore, we would deprecate the usage of `{{textarea}}`, `{{input}}`, and the `TextSupportMixin`. This will require a deprecation guide for showing developers how to convert to use these new primitives.
+### `{{prop}}` Static Semantics
+
+- {{__prop__ _HashPair_}}
+  - {{__prop__ _HashPair_ ... }}
+
+- _HashPair_ __:__
+  - _Key_ __=__ _Value_
+
+- _Key_ __:__
+  - _PathExpression_
+  - _StringLiteral_
+
+- _Value_ __:__
+  - _PathExpression_
+  - _StringLiteral_
+  - _BooleanLiteral_
+  - _NullLiteral_
+  - _UndefinedLiteral_
+  - _NumberLiteral_
+
+### `{{prop}}` Runtime Semantics
+The `{{prop}}` modifier is installed on the element once the element is placed in the DOM. At that time it will loop over the `HashPair`(s) setting the properties on to the element. No normalization is done for the `Key` or `Value`. In other words what you see is what you get.
+
+The `{{prop}}` modifier will install a cache on the modifier that it uses to compare against if and when the `Value`(s) change. This is to ensure that `Value`(s) that may force layout don't force layout unnecessarily.
+
+There are no destruction semantics. Becuase of this event setting should not go through `{{prop}}`.
+
+### `{{on}}` Static Semantics
+
+- {{__on__ _EventName_ _EventCallBack_ _EventOptions?_}}
+
+- _EventOptions?_ __:__
+  - _Key_ __=__ _Value_
+
+- _Key_ __:__
+  - __passive__
+  - __capture__
+
+- _Value_ __:__
+  - _PathExpression_
+  - _BooleanLiteral_
+
+- _EventName_ __:__
+  - _PathExpression_
+  - _StringLiteral_
+
+- _EventCallBack_ __:__
+  - _PathExpression_
+  - _SubExpression_
+
+### `{{on}}` Runtime Semantics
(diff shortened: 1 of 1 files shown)


Panman8201:
Sooo.... we're bringing back a variant of `{{bind-attr}}`?? 😕

knownasilya:
Seems like an implementation of core element modifiers https://github.com/emberjs/rfcs/pull/112

The old way still exist, but works in a different way under the hood. The above would only be in certain situations. @chadhietala maybe an example is required of when the above would be required and when not?

chadhietala:
@knownasilya Yes this is introducing more modifiers like #112 outlined. We are actively working on revising that RFC. This RFC may make more sense when #112 is updated.

@Panman8201 Not quite. `{{bind-attr}}` was used for introducing dynamic attributes to any element and still had the implicit semantics as described in this RFC. There are only a handful of places where we think `{{prop}}` would need to be used. Below is an outline of some unexpected consequences of how we do things today, along with some use cases we feel like will most likely would be affected.

## Weird Cases That Work Today

These cases would require the `{{prop}}` modifier. For simplicity sake the below examples use static strings, but the result and migration is the same if they are dynamic.

### `innerText`

```hbs
<div class="red" innerText="Inner Text">
- What in the World???
</div>
```
Renders:

```
<div class="red">
Inner Text - What in the World???
</div>
```

Would be re-written to:

```
<div class="red" {{prop innerText="Inner Text"}}>
- What in the World???
</div>
```

### `innerHTML`

```hbs
<div class="red" innerHTML="<h1>Inner HTML</h1>">
- What in the World???
</div>
```

Renders:

```hbs
<div class="red">
<h1>Inner HTML</h1> - What in the World???
</div>
```

Would be re-written to:

```
<div class="red" {{prop innerHTML="Inner HTML"}}>
- What in the World???
</div>
```

### `textContent`

```hbs
<div class="red" textContent="Text Content">
- What in the World???
</div>
```

Renders:

```hbs
<div class="red">
Text Content - What in the World???
</div>
```

Would be re-written to:

```
<div class="red" {{prop textContent="Text Content"}}>
- What in the World???
</div>
```

## More Reasonable Cases
As mentioned in the RFC majority of the time you want what happens by default. These cases should only pertain to built in "user input" elements and custom elements. Consider the following html:

```html
<input value="Chad" />
```

When the browser parses this and creates the DOM it will `setAttribute` of `value` to `"Chad"`, however if the user types into the `<input />` the HTML attribute is not updated but rather the property on the `HTMLElement`. [Example](https://jsfiddle.net/66aLjmjj/6/). So in Ember's case when you have a dynamic value that is being used on a "user input" element you would need to use `{{prop}}`. This would include properties such as `value` on `<input />`, `selected` on `<radio />`, `selected` on `<option />`. Happy to give more examples.

balinterdi:
@chadhietala Would `disabled` and `checked` be other examples where you'll have to use `{{prop}}`?

text/0000-attribute-property-rationalization.md
```diff
@@ -0,0 +1,174 @@
...
+- _EventName_ __:__
+  - _PathExpression_
+  - _StringLiteral_
+
+- _EventCallBack_ __:__
+  - _PathExpression_
+  - _SubExpression_
+
+### `{{on}}` Runtime Semantics
+The `{{on}}` modifier is installed on the element once the element is placed into the DOM. At that time we will call `addEventListener` with the `EventName` and `EventCallBack`. Since an element can have multiple `{{on}}` modifiers installed on it, we will associate the `EventCallBack` and `EventName` with the element so we can call `removeEventListener` during destruction of the element. The `addEventListener` will be called with `{ capture: true }` if `EventOptions` are not passed.
```
karthiick: @chadhietala Why capture phase as default instead of bubbling phase? Any reason?

piotrpalek:
I dislike this for DX reasons mostly. It's pushing a lot of manual labor to developers and additional complexity to the templates. I'd argue that the templates will get unreadable with all of the sigils in them, it's already somewhat hard to read some templates when there is a lot of stuff going on.

Also, do we really need yet another way to define actions / events? I think our current setup is [complex enough](https://www.youtube.com/watch?v=bt9MRkf5Mus&feature=youtu.be&t=14450):

image

I guess my main issue is that the current setup *mostly* works and is pretty convenient to work with. It feels like this proposal optimizes for the edge cases instead of the "90%" use case. Maybe an alternative would be to go the other way, meaning leave the heuristics as they are (or improve if necessary) and provide tools (like `{{prop}}` or `{{on}}`) to specify what we want to do in the instances we need to.

chadhietala:
@balinterdi Yes.

@piotrpalek For what it's worth I help maintain an app that has 120,000 lines of handlebars (2589 templates). I did a quick audit of just `<input />` and `<option />` usage, of the 752 usages (268 `<option />` and 484 `<input />`) identified roughly 54 places where `{{prop}}` would need to be used.

piotrpalek:
@chadhietala I've tried to collect my thoughts about this proposal:

What I don't like:
 - it will force existing and new Ember.js users to learn the (muddy) attributes vs properties html semantics
 - it makes using Ember.js less convenient
 - it adds complexity to things we already have (event listeners/action handlers)
 - it increases (imo significantly) the learning curve of Ember

What I wish it would looks like:
 - I would expect for Ember to do the right thing for the common case by default (as it does now)
 - Maybe these element modifiers could be added, but we wouldn't deprecate/remove/discourage the current behavior? This would allow their usage for those special cases without forcing new/existing users to learn when to use which
 - If the current semantics are unclear perhaps I'd be good to specify them somewhere to clear it up (could also have the benefit of exposing uncosistencies / bugs)
 - If the Custom Components use case is the most important one, maybe we could use the same way Glimmer.js [distinguishes](https://www.emberjs.com/blog/2017/10/10/glimmer-progress-report.html#toc_component-attributes) between properties and attributes (`@` sigil vs lack of it)

Useful discussions from the React ecosystem:
- https://github.com/facebook/react/issues/7249
- https://github.com/facebook/react/issues/11347

text/0000-attribute-property-rationalization.md
```diff
@@ -0,0 +1,174 @@
...
+- _Value_ __:__
+  - _PathExpression_
+  - _StringLiteral_
+  - _BooleanLiteral_
+  - _NullLiteral_
+  - _UndefinedLiteral_
+  - _NumberLiteral_
+
+### `{{prop}}` Runtime Semantics
+The `{{prop}}` modifier is installed on the element once the element is placed in the DOM. At that time it will loop over the `HashPair`(s) setting the properties on to the element. No normalization is done for the `Key` or `Value`. In other words what you see is what you get.
```
jelhan, line 101: The timing is problematic for custom elements.

Custom elements (web components) have lifecycle hooks similar to `@ember/component`. One of them is `connectedCallback`. It's executed when the element is insert into the DOM.

The property is not available in that hook if `{{prop}}` modifier runs after the element had been insert into DOM.

I'm aware that this is a limitation of modifiers in general and not related to `{{prop}}` modifier proposed here only. The limitation is discussed in detail at #652.

text/0000-attribute-property-rationalization.md
jelhan: This seems to be outdated. There was a separate [RFC for `{{on}}` modifier](https://github.com/emberjs/rfcs/pull/471), which is merged. It has been implemented and shipped as part of Octane.

text/0000-attribute-property-rationalization.md
```diff
@@ -0,0 +1,174 @@
...
+
+We should also add a section to the docs about how to interop with native web components as this unlocks the primitives to properly support them.
+
+## Drawbacks
+The drawback of this is that we are making developers know about some nuanced details of the updating semantics of HTMLElements. The tradeoff is that serve-side rendering works as expected and developers have a clear path for adopting natvie web components.
+
+## Alternatives
+There are a couple different alternatives. We could keep `{{textarea}}` and `{{input}}` around and somehow special case them as special components that always set properties. However, this means we would have to expand helper support to things like `<option />`. It's also odd to be creating primitives just for the sake of updating semantics.
```
jelhan, line 154: I may have missed it but is this RFC proposing to deprecate `<Textures>` and `<Input>` built-in components?

text/0000-attribute-property-rationalization.md
```diff
@@ -0,0 +1,174 @@
...
+
+The other option is to take Angular's approach and always set properties. However, this creates the inverse problem from serve-side rendering. Since you can slot an Element instance with any field it is unclear how we serialize any attribute. This means we would have to do all attribute setting during server-side rendering and then set properties on the client, but this may surface even more issues.
+
+## Unresolved questions
+
+- Can we remove `TextSupportMixin`?
+- Can we remove `{{textarea}}`?
+- Can we remove `{{input}}`?
```
jelhan, line 162: I don't see how these questions are related to this specific RFC.

text/0000-attribute-property-rationalization.md
```diff
@@ -0,0 +1,174 @@
...
+- Do we really want `{{on}}` to have updating semantics for `EventName`?
+- Who is responsible for binding the context for the `EventCallback`?
```
jelhan, line 164: These two questions seem to be outdated as on modifier has already been shipped.

text/0000-attribute-property-rationalization.md
```diff
@@ -0,0 +1,174 @@
...
+```
+
+While the codemod will re-write existing instances of this, we can warn developers in development builds that they should set the value using the `{{prop}}` modifier. We can also expand the rules of [Ember Template Lint](https://github.com/rwjblue/ember-template-lint) to be a linting error for these limited and well known cases. The guides should be updated calling out these examples and should be referenced in any warning message that is raised.
+
+We should also add a section to the docs about how to interop with native web components as this unlocks the primitives to properly support them.
+
+## Drawbacks
+The drawback of this is that we are making developers know about some nuanced details of the updating semantics of HTMLElements. The tradeoff is that serve-side rendering works as expected and developers have a clear path for adopting natvie web components.
+
+## Alternatives
```
jelhan, line 153: As a third option we could stick with the current rules but additionally provide fine granular control to developer where needed. E.g
 `{{attr}}` and `{{prop}}` modifiers that run before the element is insert into the DOM.

wagenet:
Is there still interest in moving this forward? If so, I can help get it finalized.

PR #319, opened 2018-03-24
`html-safe` helper

mmun:
[Rendered](https://github.com/emberjs/rfcs/blob/html-safe-helper/text/0319-html-safe-helper.md)

### DIFF
--- a/text/0319-html-safe-helper.md
+++ b/text/0319-html-safe-helper.md
@@ -1,0 +1,42 @@
+- Start Date: 2018-03-24
+- RFC PR: https://github.com/emberjs/rfcs/pull/319
+- Ember Issue: (leave this empty)
+
+# `html-safe` helper
+
+## Summary
+
+This RFC proposes to add an `html-safe` template helper for creating `SafeString`s in templates.
+
+The helper would be invoked as `(html-safe trustedHTML)` and return a `SafeString` that wraps the string of trusted HTML using the `htmlSafe` function in `@ember/string`.
+
+## Motivation
+
+Ember should encourage people to use `SafeString` and one way to do that is to make them easier to use. Sometimes it can be inconvenient to use the `htmlSafe` function because it would require you to create a component when you otherwise wouldn't need one. This is especially common inside of {{#each}} blocks. If we had a `html-safe` helper than you could express this in the template.
+
+## Detailed design
+
+The `html-safe` helper will be implemented similar to other helpers:
+
+```js
+import { helper } from '@ember/component/helper';
+import { htmlSafe } from '@ember/string';
+
+export function htmlSafe([trustedHTML]) {
+  return htmlSafe(trustedHTML);
+}
+```
+
+It's worth noting that the `htmlSafe` function already handles edge cases like `null` and `undefined` gracefully so it works well as a Handlebars helper.
+
+## How we teach this
+
+This helper should be mentioned in the [API docs](https://emberjs.com/api/ember/release/classes/Ember.Templates.helpers). I couldn't find a security section in the guides, but if it exists it should also go there. The warning message that you get when you use a non-SafeString in an insecure place could also mention this helper.
+
+## Drawbacks
+
+As usual, adding new helpers increases the surface area of the API and file size but in this case it is justified because the file size change is extremely small and its making it easier to build secure apps.
+
+## Alternatives
+
+While it's very easy to write this helper yourself, it costs us very little to provide it in Ember ever so slightly reduces the barrier to writing more secure apps.


text/0319-html-safe-helper.md
```diff
@@ -0,0 +1,42 @@
...
+
+## Summary
+
+This RFC proposes to add an `html-safe` template helper for creating `SafeString`s in templates.
+
+The helper would be invoked as `(html-safe trustedHTML)` and return a `SafeString` that wraps the string of trusted HTML using the `htmlSafe` function in `@ember/string`.
+
+## Motivation
+
+Ember should encourage people to use `SafeString` and one way to do that is to make them easier to use. Soemtimes it can be inconvenient to use the `htmlSafe` function because it would require you to create a component when you otherwise wouldn't need one. This is especially common inside of {{#each}} blocks. If we had a `html-safe` helper than you could express this in the template.
```
rwjblue: s/soemtimes/sometimes/

mmun:
@rwjblue Great point. I hadn't considered it as a complete alternative to `{{{ ... }}}`, since I usually use it in dynamic styles.

kellyselden:
I think this should mention the trade-off of new API vs the existing triple curly way of doing it. Also, any confusion of which one is better to use, and if the existing {{{}}} should be a deprecation candidate.

mmun:
@kellyselden :+1: One of the benefits is that `html-safe` is a subexpression so you can pass the result around.

Question: does `{{{ }}}` work in attributes like `style`?

kellyselden:
I didn't think about message passing benefits. That's a good one.

> Question: does {{{ }}} work in attributes like style?

Just tested. Yes it does.

lukemelia:
I think this is a good idea. Pretty much all of our apps have this exact helper.

ef4:
I’m wondering about the name. It seems like a security troll because it’s easy for someone to assume that this *makes* a thing safe, when the opposite is true. 

(I know we already use this name in JS, and even there it’s not wonderful. But making it even more convenient to find and use means more beginner level devs will see it and use it.)

Some of the comments here are making me wonder if even RFC readers are thinking of the full security implications of what htmlSafe does. Switching from triple curlies to this helper does not change your security — it just makes it harder to notice if you’re doing something dangerous. I’m especially concerned that people think this is a good way to avoid triple curly linter warnings, because it’s just a way to elude those very legitimate security warnings. Any project that lints for triple curlies should also lint for this new helper. 

I’m still in favor because I agree it’s better to be able to separate the location where you know a value is safe from the value where it’s ultimately bound into the DOM. But I want us to consider names that convey the gravity of what you’re doing.  Just to throw one out:`trust-this-html`

balinterdi:
`unescaped` or `raw`?

ef4:
One more naming consideration: we should probably keep `html` as part of the name, because being safe/trusted/raw for use in HTML is different from being safe/trusted/raw for other contexts like CSS. In the future we may want to provide a stronger built-in distinction.

GavinJoyce:
`raw_html` or `trusted_html` perhaps?

paddyobrien:
Perhaps `dangerously-render-html` borrowing from reacts `dangerouslySetInnerHTML`

GavinJoyce:
If we do use a name other than `html-safe` (which I think we should), it would be valuable if this name could be used consistently across the framework. If we used a name like `trusted-html`, we could also use it as follows in place of `htmlSafe`:

```js
import { trustedHtml } from '@ember/string';

trustedHtml('<div>someString</div>');
```

> Perhaps dangerously-render-html borrowing from reacts dangerouslySetInnerHTML

I'm not sure that this is a good name as the function isn't rendering anything.

cibernox:
My points are:
1. Will the triple curly (`{{{}}`) be deprecated? If so, I believe it should be part of of this RFC.
2. Others have suggested that `safe-html` can misguide users. While I may see their point, I think that having the javascript API and the HTMLbars API aligned is good and that potential confusion it's not worth it.

ef4:
> I think that having the javascript API and the HTMLbars API aligned is good and that potential confusion it's not worth it.

Yeah, I think that's an argument for adding the new name (I think `trustedHTML` is my favorite of the suggestions so far) as both a helper and as an importable Javascript function, and we would deprecate the old `htmlSafe` function.

mmun:
We discussed this at the core team meeting on 2018-03-30 and decided we are not comfortable with this proposal and encourage the author to rework the existing APIs and new template helper to be clearer in their intent.

localpcguy:
We've had the exact concerns that @ef4 mentioned above about the name implying that it somehow makes the html "safe", and I've had to explain that to coworkers (after I understood it myself.)

I personally like `raw` - potentially tacking `HTML` on the end (`rawHTML`). I like `trustedHTML` also. And agree the JS helper should also be updated.

GavinJoyce:
From `The Ember Times - Issue #40`:

screen shot 2018-03-31 at 01 41 42

I've seen many people make this mistake, it does sound like `htmlSafe` will make a string safe.

locks:
Rust went the opposite approach for doing something similar, the [`unsafe`](https://doc.rust-lang.org/book/second-edition/ch19-01-unsafe-rust.html) block denotes code that isn't checked by the Rust compiler but the programmer pinky swears is actually safe.

balinterdi:
> We discussed this at the core team meeting on 2018-03-30 and decided we are not comfortable with this proposal and encourage the author to rework the existing APIs and new template helper to be clearer in their intent.

@mmun The author is you, right? :) What was the main objection? Would coming up with a good name make the intent clearer?

mmun:
@balinterdi yes and yes. I think even the Rust people don't think `unsafe` blocks are the named accurately. It's more like a `manually_verfied_to_be_safe` block. We probably want to use something `trust` related, like in @ef4 and @GavinJoyce.

wycats:
> Rust went the opposite approach for doing something similar, the unsafe block denotes code that isn't checked by the Rust compiler but the programmer pinky swears is actually safe.

I think that, in retrospect, this has caused more confusion than something like `trust` (because it conflates functions that are unsafe to use and a promise that you are using unsafe functionality safely).

I believe that `trust` is the right way to express what is happening here. TL;DR "Hey Glimmer, you can trust this string".  Typically, you'd use this when the string came from your keyboard, or because you already escaped it for the context in which it will be interpolated.

wycats:
Also, the thing I don't like about "dangerous" or "unsafe" is that it doesn't really tell you why what you're doing is unsafe, and people are willing to stab at it to make errors go away. On the other hand, "trust" is talking about the string ("you can trust the string"). I still think people will likely stab at it, but when they're doing so, they will more likely understand what they're saying.

wycats:
I also think that Rust's `unsafe` is too coarse.

Instead of this current API:

```rs
// some bytes, in a vector
let sparkle_heart = vec![240, 159, 146, 150];

let sparkle_heart = unsafe {
    String::from_utf8_unchecked(sparkle_heart)
};

assert_eq!("💖", sparkle_heart);
```

I think this API would be nicer:

```rs
// some bytes, in a vector
let sparkle_heart = unsafe { AssertUTF8(vec![240, 159, 146, 150]) };

let sparkle_heart = String::from(sparkle_heart);
assert_eq!("💖", sparkle_heart);
```

We have a similar situation with context: I want it to be possible to assert that some string has been validly escaped for HTML, CSS, script or whatever, rather than just marking a string as generally trusted.

GavinJoyce:
@mmun what's the current status of this? Let me know if you don't have time to rework this RFC, I'd be happy to take it on

locks:
@GavinJoyce if you are still interested, I would suggest you take the wheel!

GavinJoyce:
@locks 👍 I'll pick this up and work on a new RFC, thanks

GavinJoyce:
Here's a new RFC which proposes that we:

 * Deprecate `htmlSafe` in favor of a new `trustedHtml` function.
 * Deprecate `{{{` in favor of a new `trusted-html` handlebars helper.

https://github.com/emberjs/rfcs/pull/443

rwjblue:
@mmun - Any objections to closing in favor of iterating further over in #443?

Issue #306, opened 2018-02-12
Allow passing a function as the key for {{each}}


locks:
Currently, `{{each}}` in templates have a `key` property that receives a string, which Ember sets for you using `guidFor`.

The `key` property only supports passing a string at the moment, but it would be useful to enhance it to also receive a function.
This function would have to be provided either by the current context (controller or component), or generated by a helper, like happens nowadays with `action`.

rwjblue:
https://github.com/emberjs/rfcs/pull/321 addresses this...

wagenet:
Closing since there's a related RFC.

PR #321, opened 2018-03-24
Key functions for `each`/`each-in`

mmun:
[Rendered](https://github.com/emberjs/rfcs/blob/key-functions/text/0321-key-functions.md)

Closes https://github.com/emberjs/rfcs/issues/306

### DIFF
--- a/text/0321-key-functions.md
+++ b/text/0321-key-functions.md
@@ -1,0 +1,74 @@
+- Start Date: 2018-03-24
+- RFC PR: https://github.com/emberjs/rfcs/pull/321
+- Ember Issue: (leave this empty)
+
+# Key Functions for `each`/`each-in`
+
+## Summary
+
+This RFC proposes that we support passing functions to the `key` argument of the `each` and `each-in` helpers. These functions would be called for each item and would be expected to return they key value used in diffing.
+
+For example,
+
+```js
+import Component from '@ember/component';
+
+export default Component.extend({
+  itemKey(item) {
+    return `${item.factoryNumber}-${item.partNumber}`;
+  }
+});
+```
+
+```hbs
+{{#each @items key=itemKey as |item|}}
+  {{item-details item=item}}
+{{/each}}
+```
+
+## Motivation
+
+The `key` argument to `each` and `each-in` determines the key used in the Glimmer diffing algorithm. Here is a description of the currently accepted values?
+
+`key` argument          | Computed key
+------------------------|---------------------------------------------
+`@identity`             | `guidFor(item)`
+`@index`                | index of the item in the array / object keys
+`@key` (`each-in` only) | the key of the key-value pair
+other strings           | `get(item, path)`
+otherwise               | error
+
+If a key is not specified then `@identity` strategy is used.
+
+Unfortunately, unless one of the default strategies works for your use case, you're stucking using a path on your object as the key. This can be problematic when you need to assemble a composite key from multiple values as in the example above. It is especially problematic when dealing with objects whose fields are out of your control (e.g. a moment date, an immutable.js record, a frozen object, etc.). Another less common reason you might want to use a key function is if your keys are expensive to compute and you want to cache them in a `WeakMap`.
+
+## Detailed design
+
+If `key` is a function, then Glimmer's diffing algorithm will call this function to determine the item's key.
+
+For `each`, the function would be called with arguments `keyFn(item, index)`.
+For `each-in`, the function would be called with arguments `keyFn(key, value)`.
+
+The key function should not be invoked with a `this`.
+
+One additional benefit of key functions is that they can be used to model the existing strategies in a uniform way:
+
+`key` argument | `each` key function                | `each-in` key function
+---------------|------------------------------------|-----------------------------------
+`@identity`    | `(item, index) => guidFor(item)`   | `(key, value) => guidFor(value)`
+`@index`       | `(item, index) => index`           | N/A
+`@key`         | N/A                                | `(key, value) => key`
+other strings  | `(item, index) => get(item, path)` | `(key, value) => get(value, path)`
+
+## How we teach this
+
+This feature just needs to be documented in the API docs and possibly the guides. We may want to reframe the existing (string-based) key documentation to be explained as shorthands for key functions.
+
+## Drawbacks
+
+There aren't any drawbacks I can see besides the usual concerns of increased file size and API surface. I believe that the increased flexibility of keys and making the built-in key strategies possible to be written in user space sufficiently offset these concerns.
+
+## Alternatives
+
+- Leave things as they are. I don't think this is a good idea.
+- Think of an alternative way to solve this problem. I can't think of an option that is better than this.


text/0321-key-functions.md
```diff
@@ -0,0 +1,74 @@
...
+
+If a key is not specified then `@identity` strategy is used.
+
+Unfortunately, unless one of the default strategies works for your use case, you're stucking using a path on your object as the key. This can be problematic when you need to assemble a composite key from multiple values as in the example above. It is especially problematic when dealing with objects whose fields are out of your control (e.g. a moment date, an immutable.js record, a frozen object, etc.). Another less common reason you might want to use a key function is if your keys are expensive to compute and you want to cache them in a `WeakMap`.
+
+## Detailed design
+
+If `key` is a function, then Glimmer's diffing algorithm will call this function to determine the item's key.
+
+For `each`, the function would be called with arguments `keyFn(item, index)`.
```
Turbo87: this seems strange because in an array the `index` is sort of like that `key`, which means in one case we're passing `key, value` and `value, key` in the other case 🤔

text/0321-key-functions.md
```diff
@@ -0,0 +1,74 @@
...
+For `each-in`, the function would be called with arguments `keyFn(key, value)`.
+
+The key function should not be invoked with a `this`.
```
Turbo87: what does that mean?

text/0321-key-functions.md
mmun: The arguments are intended to match the order of the block params for those helpers.

text/0321-key-functions.md
mmun: That glimmer should call the key function as `keyFn(...args)` as not to bind `this` to any particular value.

Turbo87: but with https://github.com/emberjs/rfcs/pull/308 would we then need to use `key=this.someUnboundFunction`? would seem strange if that was actually not bound to `this` in that case 🤔

mmun: I don't think it's strange. For example,

```hbs
{{some-component onClick=this.someUnboundFunction}}
```

is also not bound. This is generally true (albeit a bit annoying!) in Handlebars today.

Turbo87: yeah, true, I guess passing around `this.someUnboundFunction` in JS also doesn't have to mean that it's bound...

Turbo87: > The key function should not be invoked with a `this`.

I think due to #308 we should rephrase this as something like "The key function will not be bound to `this` automatically", since in the future we will have to call it with `this` if #308 gets merged.

mmun: :+1: That phrasing is better.

text/0321-key-functions.md
```diff
@@ -0,0 +1,74 @@
...
+
+```hbs
+{{#each @items key=itemKey as |item|}}
+  {{item-details item=item}}
+{{/each}}
+```
+
+## Motivation
+
+The `key` argument to `each` and `each-in` determines the key used in the Glimmer diffing algorithm. Here is a description of the currently accepted values?
```
rwjblue: The ending question mark seems pretty out of place here. Is this a description of the current support, or is the RFC asking for clarification?

text/0321-key-functions.md
```diff
@@ -0,0 +1,74 @@
...
+`key` argument | `each` key function                | `each-in` key function
+---------------|------------------------------------|-----------------------------------
+`@identity`    | `(item, index) => guidFor(item)`   | `(key, value) => guidFor(value)`
+`@index`       | `(item, index) => index`           | N/A
+`@key`         | N/A                                | `(key, value) => key`
+other strings  | `(item, index) => get(item, path)` | `(key, value) => get(value, path)`
+
+## How we teach this
+
+This feature just needs to be documented in the API docs and possibly the guides. We may want to reframe the existing (string-based) key documentation to be explained as shorthands for key functions.
```
rwjblue: I'd prefer to remove "just" here, otherwise 👍

text/0321-key-functions.md
mmun: https://www.youtube.com/watch?v=tqNhEzrWQpY

rwjblue: Yeah, I don't know what I was thinking?

text/0321-key-functions.md
```diff
@@ -0,0 +1,74 @@
...
+For `each`, the function would be called with arguments `keyFn(item, index)`.
+For `each-in`, the function would be called with arguments `keyFn(key, value)`.
+
+The key function should not be invoked with a `this`.
+
+One additional benefit of key functions is that they can be used to model the existing strategies in a uniform way:
+
+`key` argument | `each` key function                | `each-in` key function
+---------------|------------------------------------|-----------------------------------
+`@identity`    | `(item, index) => guidFor(item)`   | `(key, value) => guidFor(value)`
```
cibernox: One possible suggestion I have is to reverse the order of the arguments in one of the functions. I believe that the value should be in the same position in both.

`key` argument | `each` key function                | `each-in` key function
---------------|------------------------------------|-----------------------------------
`@identity`    | `(item, index) => guidFor(item)`   | `(value, key) => guidFor(value)`

or

`key` argument | `each` key function                | `each-in` key function
---------------|------------------------------------|-----------------------------------
`@identity`    | `(index, item) => guidFor(item)`   | `(key, value) => guidFor(value)`

I have the feeling that it would make the table simpler:

`key` argument |  key function for both
---------------|------------------------------------
`@identity`    | `(_, value) => guidFor(value)`  
`@index`/`@key`       | `(ref, _) => ref`   
other strings  | `(_, value) => get(value, path)`

However I also see the convenience on `{{each}}` taking the arguments in the same order that `.forEach` and `{{each-in}}` having them as (`key`, `value`)

mmun: That order seems plausible since it's the order you get when calling the Array.prototype.entries and Map.prototype.entries iterators, however I think it still makes more sense to mirror the block param order. Additionally, I don't like promoting array index to the first slot because you almost never want to use it as a key (unless you know the whole #each block is state-less and no one will ever add a stateful component into it in the future, see https://medium.com/@robinpokorny/index-as-a-key-is-an-anti-pattern-e0349aece318).

cibernox: Yeah, I'm not really opposed to the proposed order, I just thought that this one presented a "chance for unification", in case that was desirable or simplified the implementation in any form. I'm happy either way.

bgentry:
I'm definitely a :+1: on this.

With [ember-apollo-client](https://github.com/bgentry/ember-apollo-client), we're always just passing either POJOs wrapped in an `EmberObject` or `NativeArrays`. Out-of-the-box, the arrays do not work cleanly with Ember unless you pass `key="id"` or similar to avoid re-rendering the components for the entire list. That's because the POJOs themselves do not have the same guid, so Ember thinks they're different.

However, users typically already have to tell Apollo how to get a unique identifier from their objects by configuring a [`dataIdForObject()`](https://www.apollographql.com/docs/react/advanced/caching.html#normalization) function. That gets used for cache normalization purposes and typically just defaults to `obj => obj.id` for apps with globally unique IDs. I would love to use this same function to make it easier for users to get correct behavior on re-renders. The suggestion from @mmun was to build an `{{#apollo-each}}` helper that gets AST transformed to reference the key function. This RFC would make that possible.

The proposal itself seems straightforward and I would love to see it built.

wagenet:
Seem like there has been some solid interest for this. I'm going to see if we can get some better review and a proper yay or nay on it.

wagenet:
I'm moving this FCP to close since it's had virtually no activity in the last few years. If someone wants to pick this up and make sure the RFC is up to date, we can re-open.

PR #322, opened 2018-03-24
Deprecation of copy and Copyable

lupestro:
Here's the initial complete draft

### DIFF
--- a/text/0322-deprecate-copy-copyable.md
+++ b/text/0322-deprecate-copy-copyable.md
@@ -49,22 +49,17 @@ The route  `packages/ember-routing/lib/system/route.js` has one shallow copy, bu
 
 The `copy()` methods in `packages/ember-metal/lib/map.js` and  `chains.js` and their use in `meta.js`, and  `map_test.js` are unrelated.
 
-During the deprecation period, the `Ember.copy` method and the `NativeArray.copy` methods will carry a deprecation warning. At the end of this period, we will remove the deprecated elements:
-* The deprecated copy() method and the Copyable mixin
-* The use of Copyable in NativeArray
-* The deprecated  copy() method in NativeArray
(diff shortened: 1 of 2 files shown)


text/0322-deprecate-copy-copyable.md
```diff
@@ -0,0 +1,137 @@
...
+
+If we change  `Ember.copy` to consistently always use the array copy mechanism to copy arrays, even arrays that implement `Copyable`, the `NativeArray.copy` method should never be called as a side-effect of calling `Ember.copy`. By deprecating `NativeArray.copy` as well, the only users who ever see this deprecation will be those who are calling it directly.
+
+Even if we don't make this change in `Ember.copy`, we absolutely must do so in the version of `copy` in the add-on. We must also exempt arrays from the restriction on `EmberObject` and `Copyable`, since we will be removing `Copyable` from `NativeArray`.
+
+Those using the add-on would need to mechanically adjust any uses of  `myArray.copy(deep)` to  `copy(myArray, deep)` in order to avoid the deprecation message.
+
+I cannot see a way to notify the user through deprecation at runtime about their own objects that are extending `Copyable` and contain `copy()` methods that will no longer be called when the deprecated `Ember.copy()` method goes away. Our best bet for this might be to supply a new eslint warning that flags the import of `Copyable`.
```
rwjblue: We can implement a deprecation at “extend” time such that anything including the copyable mixin received the deprecation (even if copy is never called on it), happy to deep dive the details with you in slack...

rwjblue:
Let’s do this!!

text/0322-deprecate-copy-copyable.md
lupestro: That would be welcome - but that would include all arrays, no?

cibernox:
This looks great. This is a general JS problem that should not be part of Ember.

text/0322-deprecate-copy-copyable.md
lupestro: "at extend time" - Since `extend` is defined in `system.core-object`, also in `ember-runtime`, implementing a deprecation at that level might be architecturally sound. The check would be a small tax on every extend, though, which seems severe. @rwjblue Was that where you intended to put it? Or somewhere else in the chain of activity "at extend time"?

runspired:
I have removed all usage of `copy` from ember-data https://github.com/emberjs/data/pull/5436

rwjblue:
Thank for all the hard work here, I super excited to be moving forward!

PR #328, opened 2018-04-27
Include `ember-cli-build-notifications` by default

mattmcmanus:
[Rendered](https://github.com/mattmcmanus/rfcs/blob/master/text/000-build-notifications.md)

### DIFF
--- a/text/000-build-notifications.md
+++ b/text/000-build-notifications.md
@@ -6,7 +6,11 @@
 
 ## Summary
 
-Adding ember-cli-build-notifications as a default addon will increase developer productivity and improve the feedback cycle of day-to-day development. It will also assist new ember devs by quickly notifying them of errors they do not yet know to look at for. The default settings should should be verbose, as that would be most helpful to new Ember developers. 
+`ember-cli-build-notifications` uses `node-notifier` to provide cross platform system notifications for `ember-cli` build events (specifically on build success and build error).
+
+![](https://github.com/pdud/ember-cli-build-notifications/raw/master/example.png)
+
+Adding ember-cli-build-notifications as a default addon will increase developer productivity and improve the feedback cycle of day-to-day development. It will also assist new ember devs by quickly notifying them of errors they do not yet know to look at for. The default settings should be verbose, as that would be most helpful to new Ember developers. 
 
 ## Motivation
 
@@ -17,14 +21,17 @@ Confusion as a result of the state of the build consumes a surprising amount of 
 There are several parallel paths that we can take to communicate the state of a build. Recent developments in the new error page are one of them.  Having `ember-cli-build-notifications` alert on **all** builds is another. Not all problems result in a build error, so the absence of a notification is also a useful tool in debugging.
 
 Some of these scenarios include: 
-* Addons, like `ember-cli-typescript` or tools like `watchman` breaking and not to starting a new build
-* Saves in linked packages not causing builds to start
+* Addons, like `ember-cli-typescript` or tools like `watchman` breaking, as well as saves linked pacakges not starting a new build
+  * Not seeing a success/error notification when you normally do will indicate something is off
 * Page refreshed not happening after successful builds due to livereload disconnecting
-* Making only style changes and not realizing the new css was already pushed
+  * When you see a success message but the page hasn't refreshed you know it's time to refresh the page on your own
+* Making only style changes and not realizing the new css was already pushed to the browser
+  * Seeing the success notificaiton makes it plain that your changes are now present, rather than wondering a build error has occured
 
 Build notifications give automatic insight into these scenarios in an unobtrusive way. 
 
 ## Detailed design
+
 1. Include `ember-cli-build-notifications` as part of the default kit
 2. Generate it’s config file at `{app}/config/build-notifications.js`
 3. Configure it to notify on success and notify, including sound, on failure
@@ -51,7 +58,8 @@ The best thing about this approach is that it teaches itself. The alerts will ar
 
 ## Drawbacks
 
-* Some may find the notifications noisy or unnecessary. If so, they can easily tweak the config or remove the add-on. 
+* Some may find the notifications noisy or unnecessary. If so, they can easily tweak the config or remove the add-on
+* The default for OSX is to include the all the notifications in the notification center, which is a lot. This is easily configured, but it can only done after the fact (as far as I know)
 * It’s another dependency
 
 ## Alternatives
--- a/text/000-build-notifications.md
+++ b/text/000-build-notifications.md
@@ -1,4 +1,4 @@
-- Start Date: (fill me in with today's date, YYYY-MM-DD)
+- Start Date: 2018-04-27
 - RFC PR: (leave this empty)
 - Ember Issue: (leave this empty)
 
--- a/text/000-build-notifications.md
+++ b/text/000-build-notifications.md
@@ -1,0 +1,71 @@
+- Start Date: (fill me in with today's date, YYYY-MM-DD)
+- RFC PR: (leave this empty)
+- Ember Issue: (leave this empty)
+
+# Include `ember-cli-build-notifications` by default
+
+## Summary
+
+Adding ember-cli-build-notifications as a default addon will increase developer productivity and improve the feedback cycle of day-to-day development. It will also assist new ember devs by quickly notifying them of errors they do not yet know to look at for. The default settings should should be verbose, as that would be most helpful to new Ember developers. 
+
+## Motivation
+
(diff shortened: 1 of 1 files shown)


rtablada:
I think system level notifications could be really helpful for developers.

Coming from a project where builds take about 1-2 seconds on incremental builds and initial loads take a while, I usually know when I have triggered a new build.

But working on some smaller 3.0+ projects the build times and refreshes are often completed before I am in the browser, this can lead to sometimes not realizing the refresh has even occurred.
A system notification could help this.

---

Playing devil's advocate the notification queue on Mac can get really cluttered really quickly since it keeps a backlog of notifications.
I haven't worked with `ember-cli-build-notifications` to know if it pollutes this too badly, but I have run into issues with other test/build notifiers that would fill up the queue and make it basically unusable for finding other application messages.

text/000-build-notifications.md
```diff
@@ -0,0 +1,71 @@
+- Start Date: 2018-04-27
+- RFC PR: (leave this empty)
+- Ember Issue: (leave this empty)
+
+# Include `ember-cli-build-notifications` by default
+
+## Summary
+
+Adding ember-cli-build-notifications as a default addon will increase developer productivity and improve the feedback cycle of day-to-day development. It will also assist new ember devs by quickly notifying them of errors they do not yet know to look at for. The default settings should should be verbose, as that would be most helpful to new Ember developers.
```
simonihmig: One "should" should be enough 😉

text/000-build-notifications.md
```diff
@@ -0,0 +1,71 @@
...
+Some of these scenarios include: 
+* Addons, like `ember-cli-typescript` or tools like `watchman` breaking and not to starting a new build
+* Saves in linked packages not causing builds to start
+* Page refreshed not happening after successful builds due to livereload disconnecting
+* Making only style changes and not realizing the new css was already pushed
+
+Build notifications give automatic insight into these scenarios in an unobtrusive way. 
+
+## Detailed design
+1. Include `ember-cli-build-notifications` as part of the default kit
```
simonihmig: For people like me, who didn't know of this addon before, it would be helpful to have a short summary what it actually does, so we don't have to look for the github page. 

Also maybe elaborate how it would solve the scenarios you described in the previous section?

patocallaghan:
While I think this will be a good change, similar to others concerns, I can definitely see myself changing the config to only turn it on for error builds. I only want to be notified when there’s a problem, it may become quite noisy otherwise.

I see ember-cli-build-notification’s default is not to turn this on for successful builds. Is there any reason not to keep this behaviour here too?

mattmcmanus:
@patocallaghan I tried to lay out my thoughts on why Success notifications are helpful in the scenarios above. We've been using the addon at Yapp for a while now. Over time, I've turned on the success messages for nearly all of our internal apps and addons. It eliminates so many possible questions about the status of the build. I'm also a person who generally disables notifications because I find them really distracting. That's never really been the case with this addon.

I suppose it's worth trying to make some distinctions. The apps we're working on are BIG. The incremental rebuild can take several seconds and it's really easy to disengage. The notification, whether success or error, is always an actionable thing for me.

mehulkar:
I’m not familiar with this addon, but one thing to consider is that it’s very likely that two people on the same team will disagree with the config for this. I would be much more happier about this being included by default if I could control the setting for myself locally (and not using system/OS settings). Some kind of environment variable config or a local rc file that we can gitignore would be ++.

sandstrom:
Neat functionality! 

Two thoughts:
1. I think it's better if the notifications are only shown for errors by default.
2. There must be an easy way (dropping file in `~/.ember-cli-rc` or similar) of setting this per user (overriding project defaults). For a team of 10+ people it's almost guaranteed that there will be different preferences on notification behaviour.

stefanpenner:
> There must be an easy way (dropping file in ~/.ember-cli-rc or similar) of setting this per user (overriding project defaults). For a team of 10+ people it's almost guaranteed that there will be different preferences on notification behaviour.

+ 1

> Ideally they shouldn't be kept in the macOS notification center history (too much noise, not relevant with a log in this case).


+1

mattmcmanus:
@mehulkar @sandstrom @patocallaghan @stefanpenner Could I impose on you to give the addon a try for a couple days with the settings I propose? The real test will be when you turn it off after the couple days is whether you miss the notifications. That's how I really knew I liked the setup is because whenever I'm on a project that doesn't have it I feel blind. 

> There must be an easy way (dropping file in ~/.ember-cli-rc or similar) of setting this per user (overriding project defaults). For a team of 10+ people it's almost guaranteed that there will be different preferences on notification behaviour.

As far as I'm aware, this isn't functionality that ember-cli currently has, right? That could be really useful. 

> Ideally they shouldn't be kept in the macOS notification center history (too much noise, not relevant with a log in this case).

I agree and I'm exploring whether this is an option.

rwjblue:
@mehulkar @sandstrom @patocallaghan @stefanpenner - Did y'all give this a try? How did that pan out?

I'm personally keen to move forward here as I think that the overall ergonomics are much better, but I definitely want to make sure we have investigated various options / concerns / etc...

mattmcmanus:
A nice status update: @pdud (the developer of the addon) [just released an update to the addon](https://github.com/pdud/ember-cli-build-notifications/issues/23#issuecomment-396453842) that puts all notifications in a single group. So it will never take more than on item in notification center.

sandstrom:
@rwjblue I haven't tried it. I have no new comments besides [the three I've outlined above](https://github.com/emberjs/rfcs/pull/328#issuecomment-386218874).

rwjblue:
@sandstrom - Thank you for cross linking!

> Notification should only be shown for errors by default.

This would also have been my default preference. I am not massively opposed to having it enabled all the time, but I'm not sure that notifications for success is _as_ helpful as notifications of failures...

> There must be an easy way (dropping file in ~/.ember-cli-rc or similar) of setting this per user (overriding project defaults). For a team of 10+ people it's almost guaranteed that there will be different preferences on notification behaviour.

Totally agreed! ember-cli has long needed this capability. We basically need one more type of config file (that is not intended to be checked in) in the project for personal per-project + per-user preferences, and this seems like as good a time as any to build that in. Just spit-ballin' here, but `.ember-cli` would continue to be the main checked-in project level config file and we would add a `.ember-cli.local` (or some other better name ;p ) that would be excluded from git...

> Ideally they shouldn't be kept in the macOS notification center history (too much noise, not relevant with a log in this case).

Agreed. I believe that the most recent update to the addon makes this really nice, thank you for pointing it out!

stefanpenner:
> There must be an easy way (dropping file in ~/.ember-cli-rc or similar)

Isn't this already a thing? my `~/.ember-cli` file contains:
```
cat ~/.ember-cli 
{
  "yarn": true
}
```

and works like a charm.

stefanpenner:
@rwjblue this works for me, i am 👍

sandstrom:
@stefanpenner Yep, [you are right](https://github.com/ember-cli/ember-cli/blob/master/tests/unit/settings-file/precedence-settings-test.js), I didn't know that file already existed.

rwjblue:
@stefanpenner - We currently have (listed in order of "precedence":

* per-user config (`~/.ember-cli`)
* per-project config (`<project-root>/.ember-cli`)

I was suggesting adding another to that list, for user-specific overrides of project config, which would be `.gitignore`d by default.

This would allow projects to have good default/general settings, but also allow users to override those for their own personal preference.  A couple of examples where this would be really nice:

* setting a default `--test-port` / `--port` for a specific project (allows users running many projects at once to still run `ember s` instead of `ember s -p 9999`).
* allowing customizing launchers with `ember test --server` (some folks may want `--no-launch`, others only want `--launch="firefox"`, etc)


--- 

Regardless, of this added level of configuration, I think this RFC is _still_ a good idea...

patocallaghan:
I did try out the addon but unfortunately I haven’t been doing much feature work so I couldn’t really make a proper assessment of it. 

I think the least controversial option to move the RFC along would be to introduce it with only error notifications enabled. The behaviour to turn it on all the time could be revisited if we see many people adopt that usage in the wild.

I also didn’t know about the ~/.ember-cli settings override so that could probably be enough to introduce it and allow users to configure it personally. The per-project user config, while nice, could probably be done in another RFC rather than block this one.

stefanpenner:
> I was suggesting adding another to that list, for user-specific overrides of project config, which would be .gitignored by default.

ah ok

locks:
We discussed this RFC at the last Framework team meeting and decided to initiate the final comment period to Close this RFC. Since this proposal was first introduced there have developments both in Ember (build error page) and the general ecosystem which makes this proposal not as relevant.

PR #329, opened 2018-05-01
Start RFC to deprecate the use of Ember Evented in Ember Data

bmac:
[Rendered](https://github.com/emberjs/rfcs/blob/5a1e8cddd3c1a210c2d60c7a9ac021dbfbedc370/text/1234-deprecated-ember-evented-in-ember-data.md)

### DIFF
--- a/text/1234-deprecated-ember-evented-in-ember-data.md
+++ b/text/1234-deprecated-ember-evented-in-ember-data.md
@@ -1,0 +1,84 @@
+- Start Date: 2018-05-01
+- RFC PR: (leave this empty)
+- Ember Issue: (leave this empty)
+
+# Deprecate Usage of Ember Evented in Ember Data
+
+## Summary
+
+`Ember.Evented` functionality on `DS.Model`, `DS.ManyArray`,
+`DS.Errors`, `DS.RecordArray`, and `DS.PromiseManyArray` will be
+deprecated and eventually removed in a future release. This includes
+the following methods from the
+[Ember.Evented](https://www.emberjs.com/api/ember/2.15/classes/Ember.Evented/methods/on?anchor=off)
+class: `has`, `off`, `on`, `one`, and `trigger`. Additionally the
+following lifecycle methods on `DS.Model` will also be deprecated:
+`becameError`, `becameInvalid`, `didCreate`, `didDelete`, `didLoad`,
+`didUpdate`, `ready`, `rolledBack`.
+
+## Motivation
+
+The use of `Ember.Evented` is mostly a legacy from the pre 1.0 days of
+Ember Data when events were a core part of the Ember Data programming
+model. Today there are better ways to do everything that once needed
+events. Removing the usage of the `Ember.Evented` mixin will make it
+easier for Ember Data to eventually transition to using native ES2015
+JavaScript classes and will reduce the surface area of APIs that Ember
+Data must support in the long term.
+
+## Detailed design
+
+`Ember.Evented` mixin will be scheduled to be removed from the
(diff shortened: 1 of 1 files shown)


dwickern:
[ember-data-change-tracker](https://github.com/danielspaniel/ember-data-change-tracker) depends on these events ([source](https://github.com/danielspaniel/ember-data-change-tracker/blob/ec673be0f41315c6a3ce662218d05ed25941da74/addon/model-ext.js)). Will there be an alternative?

btecu:
Some of the events (`didLoad`) are better than computed (no overhead).

joukevandermaas:
The motivation section does not mention any reason for removing lifecycle hooks like `didLoad`. These can definitely not be replaced by a computed in all scenarios. It's not obvious to me why they have to go or what the migration path will be.

gossi:
If the idea behind this RFC is to strive towards ES classes rsp. allowing a better "programming model" to narrow the gap to traditional OOP-related programming languages and their design patterns then I'd say `DS.Model` would have blank lifecycle methods of `protected` visibility state and would allow specific models to overwrite those methods for their particular use case.

Removing events in favor of overwritable methods sounds good to me as this is what happened with components afair.

runspired:
@dwickern there is already a path forward for ember-data-change-tracker with the ModelData RFC.

@btecu using the `didLoad` event is far from "no overhead". Computeds are much cheaper.

@joukevandermaas Can you provide an example use of `didLoad` that computeds do not handle?

@gossi lifecycle methods have been debated before for models; however, for the most part there isn't a clear conceptual model for what the lifecycle of a model even is. That said, the `ModelData` RFC allows for alternative model layers to experiment with this in their own ways.

joukevandermaas:
@runspired anything that has side-effects really, which is the reason to use lifecycle hooks generally. For example, kicking off a long-running task. Are you saying there is overhead to `didLoad` even if people don't use it? How was that solved for components in Ember itself?

runspired:
@joukevandermaas 

> anything that has side-effects really, which is the reason to use lifecycle hooks generally.

Side-effect driven development isn't something we seek to support. Kicking off additional work from the promise returned when fetching a record would achieve the same thing in a non-side-effect way. If you seek to centralize this pattern then fetching via a service that has these known additional behaviors would provide a clear abstraction.

> Are you saying there is overhead to didLoad even if people don't use it?

Yes. Evented accounts for somewhere between 10 and 30% of our cost to instantiate a record and an even higher percentage of the cost when mutating a record, this even after we did things to make it substantially faster around version 2.10.

> How was that solved for components in Ember itself?

At a superficial level, the answer for this difference is that Components have a clear lifecyle, which makes certain patterns easy to optimize. Models do not. At a deeper level the specific issue is that in order to support these events we had to add significant complex anti-patterns throughout the internals. This has made the internals difficult for even regular contributors to understand, and prevented many needed optimizations from taking place.

joukevandermaas:
@runspired thanks for the clear answer 👍 this makes sense to me. As long as there is a clear migration path I'm OK with anything.

mehulkar:
Could the deprecation contain the message that if the user wants to retain this behavior they can mixin Ember.Evented into their model definition manually; i.e. `DS.Model.extend(Ember.Evented, {})`? Would this be equivalent to the way things are now?

runspired:
@mehulkar possibly; however, this is primarily for events that we trigger.

sandstrom:
@bmac I'm in favour! 👍

hjdivad:
Discussed during the call today and we're happy to merge 👍 

The only change from the current text is that the deprecations will last through an LTS cycle.

Issue #307, opened 2018-02-14
Deprecate Globals Resolver


Gaurav0:
We should do this soon as people should be given time to transition.

PR #331, opened 2018-05-08
Deprecate Globals Resolver

Gaurav0:
[Rendered](https://github.com/Gaurav0/rfcs/blob/deprecate_globals_resolver/text/0331-deprecate-globals-resolver.md)

For https://github.com/emberjs/rfcs/issues/307


text/0000-deprecate-globals-resolver.md
```diff
@@ -0,0 +1,82 @@
...
+
+# Motivation
+
+Over the past years we have transitioned to using Ember-CLI as the main way
+to compile Ember apps. The globals resolver is a holdover and primarily
+facilitates use of Ember without Ember-CLI.
+
+# Transition Path
+
+Primarily, the transition path is to recommend using Ember-CLI.
```
rwjblue: Not sure anything needs to change here per-se, but using Ember without ember-cli is already officially not supported.

Gaurav0: I know. I mainly wrote this up so we could get some deprecation warnings set up.

text/0000-deprecate-globals-resolver.md
```diff
@@ -0,0 +1,115 @@
...
+
+# Alternatives
+
+Without doing this, we will have to continue to ship and maintain this rarely used functionality.
+We don't believe this is a reasonable alternative.
+
+# Unresolved questions
+
+There has never been a transition guide for transitioning an old codebase to Ember-CLI.
+Do we want to create one at this late date?
```
rwjblue: I think that we certainly _could_ do it, but I don't think that it would be a requirement for moving forward with this RFC.  As it stands today, the migration path away from globals mode to a more natural ember-cli built application is something _most_ of our community has already absorbed, and the effort of writing a thorough migration guide at this point far outweighs the upside...

text/0000-deprecate-globals-resolver.md
```diff
@@ -0,0 +1,115 @@
...
+    {{#each model as |item|}}
+      <li>{{item}}</li>
+    {{/each}}
+  </ul>
+</script>
+```
+
+# Transition Path
+
+Primarily, the transition path is to recommend using Ember-CLI.
```
rwjblue: One small clarifying point for this transition path, is that ember-cli's own default resolver [ember-resolver](https://github.com/ember-cli/ember-resolver) currently still extends from the globals resolver. In order to implement this RFC, and deprecate the globals resolver we need to change the ember-cli resolver so that it does _not_ extend from the globals resolver (otherwise even ember-cli users would get a deprecation).

JackEllis:
I've never used this.

andyhot:
Will the global resolver be moved to its own repo (once it's deprecated from ember)?

rwjblue:
This RFC does not propose that, but it definitely seems possible if someone does like it / want it.

Gaurav0:
@andyhot Are you using the globals resolver? How can we best help you transition?

text/0000-deprecate-globals-resolver.md
```diff
@@ -0,0 +1,122 @@
...
+
+One small detail required to implement this RFC: ember-cli's own default resolver,
+[ember-resolver](https://github.com/ember-cli/ember-resolver)
+currently still extends from the globals resolver.
+In order to implement this RFC, the ember-cli resolver will need to be changed
+so that it does *not* extend from the globals resolver, or otherwise ember-cli users
+will get a deprecation warning as well.
```
mixonic: We can't remove the global resolver base class for the ember-cli classic resolver, that would be a breaking change.

I can think of two options:

**Simple option**

Have the global resolver log a deprecation message at runtime. When `resolve` or any other method on the resolver is hit, deprecation message. Then we remove a) the global resolver and b) ember-cli classic resolver's extension in 4.0.

**Complex, faster option**

In the ember-cli classic resolver, deprecate any runtime calls where there is fallback to the globals mode resolver. This would be a deprecation in ember-cli's resolver. We could bump a major version of ember-cli-resolver removing the base class and release it in ember-cli after an LTS of ember-cli.

Then separately (or concurrently with some cleverness) we could deprecate any use of the ember-cli globals resolver.

text/0000-deprecate-globals-resolver.md
```diff
@@ -0,0 +1,122 @@
...
+- &lt;script type="text/handlebars" data-template-name="path/to/template"&gt;
+
+Use of any of the above should trigger a deprecation warning, with a target
+of version 4.0
+
+# Motivation
+
+Over the past years we have transitioned to using Ember-CLI as the main way
+to compile Ember apps. The globals resolver is a holdover and primarily
+facilitates use of Ember without Ember-CLI.
```
mixonic: I did a quick review, and it does look like the [ember-cli classic resolver](https://github.com/ember-cli/ember-resolver/blob/master/addon/resolvers/classic/index.js) replaces all methods on the [globals resolver](https://github.com/emberjs/ember.js/blob/master/packages/@ember/application/globals-resolver.js) 👍

text/0000-deprecate-globals-resolver.md
mixonic: I think a traditional approach to deprecation RFCs is that we write the deprecation documentation here in the RFC. Once we pick a deprecation strategy I'd like to see that added.

text/0000-deprecate-globals-resolver.md
rwjblue: > In the ember-cli classic resolver, deprecate any runtime calls where there is fallback to the globals mode resolver. This would be a deprecation in ember-cli's resolver. We could bump a major version of ember-cli-resolver removing the base class and release it in ember-cli after an LTS of ember-cli.

This was what I was planning on doing.

text/0000-deprecate-globals-resolver.md
Gaurav0: Comment added

text/0000-deprecate-globals-resolver.md
Gaurav0: Unless there are any objections, I'll put this in the RFC?

text/0000-deprecate-globals-resolver.md
Gaurav0: I would appreciate it if anyone can PR some documentation against this RFC.

I'm really not sure what to say except "Do not use globals resolver, use ember-cli".

text/0000-deprecate-globals-resolver.md
rwjblue: Yes, sounds good to update with that.

text/0000-deprecate-globals-resolver.md
rwjblue: @Gaurav0 put together a summary for the deprecation guide over in https://github.com/ember-learn/deprecation-app/issues/153

ming-codes:
I think global resolver is still needed for the case where you can start using Ember by linking in via script tag. This was one of the selling point of Vue.

I didn’t know this is still possible before I read this rfc.

mehulkar:
I've actually been looking forward to migrating a legacy app to Ember by way of the globals resolver (would be a nice onboarding strategy instead of doing a full rewrite). But that said, I can pretty trivially start with an older version of Ember to accomplish this.

rtablada:
@mehulkar are you looking to migrate a legacy Ember app or a server rendered site over to Ember? You can still use Ember to replace page by page without global resolver.

ef4:
> I didn’t know this is still possible before I read this rfc.

For practical purposes, it's not. It's documented nowhere, it supports almost nothing you would want to add to your app, and we already dropped support for anyone not building with ember-cli back at 3.0.

If somebody wants to do a project to make a pleasant "try Ember with zero install" setup, that would be great. It's definitely doable (and Ember Twiddle already demonstrates most of what you'd need to get working). But it would _not_ use the globals resolver, which -- far from being a nice selling point for Ember -- is part of the crufty history that we want to get rid of precisely because it makes adoption and innovation harder. 

A nice Try Ember experience deserves to be designed for that purpose. The globals resolver is not it.

Gaurav0:
I drafted a deprecation guide here: https://github.com/ember-learn/deprecation-app/pull/155

mixonic:
@Gaurav0 thanks for the deprecation draft.

FWIW, I think 90% of users who will run into this are *not* legacy non-ember-cli users. I believe we got those users to upgrade already. I believe it will be users of Ember-CLI who are using `App.MyModel = ` and other globals items sprinkled through their otherwise modern codebase. Thus I think you could more simply tell people to "don't use `App.FooBarComponent` for example, instead create a file with the appropriate name per Ember documentation: `app/components/foo-bar.js`.

@Gaurav0 can you update your RFC to include a link to the deprecation guide draft? Thank you.

I expect we will FCP this on Friday and merge it next week!

mehulkar:
@rtablada it would have been a non-ember app. I would have tried to use it in exactly the way documented in this RFC (i.e. stick some script tags into my html that creates and renders an app in a specific div). There's no reason that I need to use latest Ember to try this out though.

Gaurav0:
@mixonic Do you want a link to the pull request (https://github.com/Gaurav0/deprecation-app/blob/deprecation_guide_globals_resolver/content/ember/v3/deprecate-globals-resolver.md) or to where it will end up (https://emberjs.com/deprecations/v3.x#toc_ember-deprecate-globals-resolver) once merged and published?

mixonic:
@Gaurav0 to the pull request would be great. The idea is that someone reviewing the RFC should be able to read the proposed deprecation guidance either in the RFC itself or wherever it is elsewhere.

Thank you!

Gaurav0:
> FWIW, I think 90% of users who will run into this are not legacy non-ember-cli users. I believe we got those users to upgrade already. I believe it will be users of Ember-CLI who are using App.MyModel = and other globals items sprinkled through their otherwise modern codebase. Thus I think you could more simply tell people to "don't use App.FooBarComponent for example, instead create a file with the appropriate name per Ember documentation: app/components/foo-bar.js.

@mixonic I haven't run into this. In fact, most people don't know what the global namespace is for their app. Unless I made a mistake trying this, it doesn't work by default, even if they have a global namespace via https://github.com/ember-cli/ember-export-application-global and have enabled it in production mode. The ember-cli resolver takes over. Maybe one can get it working, but I'm not sure how.

rwjblue:
Discussed with the core team again today, and we are 👍 on moving this back into final comment period.

rwjblue:
👏 lets do it!

tomdale:
Thank you @Gaurav0, and congrats on the merged RFC!

PR #332, opened 2018-05-11
[EMBER DATA] record links & meta

runspired:
https://github.com/runspired/rfcs/blob/ember-data-record-links-and-meta/text/0000-ember-data-record-links-and-meta.md#ember-data-record-links--meta

### DIFF
--- a/text/0000-ember-data-record-links-and-meta.md
+++ b/text/0000-ember-data-record-links-and-meta.md
@@ -1,0 +1,224 @@
+- Start Date: (fill me in with today's date, 2018-05-10)
+- RFC PR: (leave this empty)
+- Ember Issue: (leave this empty)
+
+# Ember Data Record Links & Meta
+
+## Summary
+
+Enable users to associate `links` and `meta` information with individual records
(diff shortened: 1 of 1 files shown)


text/0000-ember-data-record-links-and-meta.md
```diff
@@ -0,0 +1,224 @@
...
+      }
+    }
+  ]
+})
+```
+
+--------------------------------
+
+Links & Meta on objects used as `ResourceIdentifiers` (e.g. to link to another resource within a relationship)
+ will not be used for the associated resource and will be silently ignored.
```
HeroicEric: Why will these be ignored?

jgwhite:
Great proposal, would be a very welcome addition to ED.

Couple of nice-to-haves:

- Examples of what one might use `meta` and `links` for
- A code snippet demonstrating how one would rename links/meta in the serializer

Do you think links/meta would also be processed by the serializer to normalize keys?

sandstrom:
@runspired This would be very useful! 👍

PR #335, opened 2018-05-29
RFC to deprecate `component#sendAction`

cibernox:
[Rendered](https://github.com/cibernox/rfcs/blob/rfc-deprecate-send-action/text/0000-deprecate-send-action.md)

### DIFF
--- a/text/0000-deprecate-send-action.md
+++ b/text/0000-deprecate-send-action.md
@@ -1,4 +1,4 @@
-- Start Date: 2019-05-29
+- Start Date: 2018-05-29
 - RFC PR: (leave this empty)
 - Ember Issue: (leave this empty)
 
--- a/text/0000-deprecate-send-action.md
+++ b/text/0000-deprecate-send-action.md
@@ -32,6 +32,98 @@ explanation that will make super-easy for developers to migrate to closure actio
 As it is mandatory with new deprecations, a new entry in the deprecation guides will be added
 explaining the migration path in depth.
 
+To refresh what the migration path would look like in the typical use case.
+
+**BEFORE**
+```js
+// parent-component.js
+export default Component.extend({
+  actions: {
+    sayHi() {
+      alert('Hello user!');
+    }
+  }
+})
+```
+
+```hbs
+{{!-- parent-component.hbs --}}
+{{child-component salute="sayHi"}}
+```
+
+```js
+// child-component.js
+export default Component.extend({
+  actions: {
+    sendSalute() {
+      this.sendAction('salute');
+    }
+  }
+});
+```
+
+```hbs
+{{!-- child-component.hbs --}}
+<button {{action "sendSalute"}}>Send salute</button>
+```
+
+**AFTER**
+```js
+// parent-component.js
+export default Component.extend({
+  actions: {
+    sayHi() {
+      alert('Hello user!');
+    }
+  }
+})
+```
+
+```hbs
+{{!-- parent-component.hbs --}}
+{{child-component salute=(action "sayHi")}}
+```
+
+```js
+// child-component.js
+export default Component.extend({
+  actions: {
+    sendSalute() {
+      this.salute();
+    }
+  }
+});
+```
+
+```hbs
+{{!-- child-component.hbs --}}
+<button {{action "sendSalute"}}>Send salute</button>
+```
+
+However _closure actions_ allow to be less verbose, so the same behavior could be attained using
+less intermediate calls
+
+```js
+// parent-component.js
+export default Component.extend({
+  actions: {
+    sayHi() {
+      alert('Hello user!');
+    }
+  }
+})
+```
+
+```hbs
+{{!-- parent-component.hbs --}}
+{{child-component salute=(action "sayHi")}}
+```
+
+```hbs
+{{!-- child-component.hbs --}}
+<button onclick={{@salute}}>Send salute</button>
+```
+
 ## How we teach this
 
 There are no new concepts to teach, but the removal of an old concept now considered outdated.
--- a/text/0000-deprecate-send-action.md
+++ b/text/0000-deprecate-send-action.md
@@ -90,6 +90,10 @@ export default Component.extend({
   actions: {
     sendSalute() {
       this.salute();
+      // if the salute action is optional you'll have to guard in case it's undefined:
+      // if (this.salute) {
+      //   this.salute()
+      // }
     }
   }
 });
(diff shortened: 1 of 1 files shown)


simonihmig:
Totally in favor of this, as we have discussed before! 👍

Maybe for some more context, it should be mentioned that not only the pure `sendAction()` method is deprecated, but essentially alongside it also the underlying concept of "bubbling string actions", which the use of `sendAction()` was required for? 

Which is IMHO way to magic, and an incompatible way to solve the same problem closure actions do (e.g. no return values).

miguelcobain:
Should we document what the alternative sendAction is?

cibernox:
@miguelcobain I'm not sure if the RFC should contain an in depth explanation of closure actions, but I wrote a blog post about it a while ago.

cibernox:
@Gaurav0 I think you're conflating `send` with `sendAction`. The latter is only available in components, and it's the one being replaced by closure actions. `send` it's used to call actions on controllers and routes, and I don't advocate its removal (yet)

miguelcobain:
@cibernox I just thought it would be clearer if the RFC had a "before and after" example, for the sake of completeness.
Could be useful for people that want to migrate their code, writing codemods or even just for understanding what this deprecation is about in a practical sense.

josemarluedke:
I'm very much in favor to this deprecation. I also think that would be beneficial to have some examples of migrating off `sendAction`.

mydea:
I am very much in favor of this, but I would suggest to maybe think about including the `route-action` helper from https://github.com/DockYard/ember-route-action-helper by default? As without it, you cannot use bubbling actions on routes anymore, which we do use quite often and find pretty helpful. Otherwise, we'd need to move all our actions to the controllers, which is not always desirable for us.

miguelcobain:
Just wanted to bring the question of optional actions. `sendAction` calls an action *if present*. It's useful in multiple scenarios.

So that means that:
```js
this.sendAction('salute');
```

is not equivalent to

```js
this.salute();
```

because `this.salute();` will error if `salute` is not passed in as a component argument.

Will we recommend users to go with the following for these cases?

```js
if (this.salute) {
  this.salute();
}
```

I believe this would be the code that better represents what `this.sendAction('salute');` does.
It also assumes [#281 RFC](https://github.com/emberjs/rfcs/blob/master/text/0281-es5-getters.md), if I'm not mistaken.

jasonbelmonti:
>Just wanted to bring up the question of optional actions. sendAction calls an action if present. It's useful in multiple scenarios.

Agreed.

simonihmig:
> Will we recommend users to go with the following for these cases?

An alternative would be to define `salute() {}` as a noop function on the component, so you can remove the guard. That would be defined on the component's prototype (not on each instance), so should not be very wasteful (see also the `Ember.K` deprecation RFC). This also provides a more obvious place to add docs for that action.

cibernox:
> Just wanted to bring up the question of optional actions. sendAction calls an action if present. It's useful in multiple scenarios.

@miguelcobain I added the changes on the examples to clarify that. 
However I do think that the approach taken by `sendAction` of calling the action only if it exists may be counterproductive when it comes to code clarity. I prefer to guard with an `if` or to define an empty function, as it makes very clear wheter the action is optional or mandatory. I like that being more explicit.

miguelcobain:
@cibernox I agree 100%.

jasonbelmonti:
@simonihmig @miguelcobain @cibernox: While I agree that `sendAction` is "magical" and makes code less clear, `sendAction` is an attempt at addressing some important ideas:
1. Easily invoking optional (potentially undefined) actions
2. Preventing components from being tightly coupled to their invocation context

Consider this antipattern:
```javascript
if (this.salute) {
  // by relying upon the return value of `this.salute`
  const result = this.salute();
  // this component implementation is tightly coupled to the implementation of `salute`
  this._postProcessSalute(result):
}
```
In the case of `sendAction`, the inability to play with returned values from the passed action is probably a good thing.  It helps encourage DDAU patterns that delegate action handling _up_ the component hierarchy.

cibernox:
> In the case of sendAction, the inability to play with returned values from the passed action is probably a good thing.

I want to add to what @simonihmig said that having return values when you invoke an action **is not a violation of the DDAU principle**. Functions are as good data as any other, so they can be passed down, and it's perfectly normal to expect functions to have return values.

knownasilya:
Yeah it enables you to work very well with async code when an action returns a promise. It opens up plenty of creative ways to use actions.

rwjblue:
We discussed this in today’s core team meeting, and we were all enthusiastic about moving forward here. 

We are moving this into the final comment period and will merge the RFC (which enables implementation to begin) in about a week (barring any new issues / concerns).

akashdsouza:
For actions in route, if I didn't want to write a dummy action with `send()` to simply bubble it up, would doing something like [this](https://medium.com/ember-titbits/quest-3-how-to-call-an-action-on-a-route-using-closure-actions-7f9ab00dddae) be the right solution?
```
// application/template.hbs
{{my-component editUser=(action send 'editUser')}}
```

```
// my-component/component.js
import Ember from 'ember';
  actions: {
     editUser() {
        this.editUser();
     }
  }
});

Caltor:
@akashdsouza I think you need to have `this.get ('editUser')();` in your action.

knownasilya:
@Caltor you don't need `.get` for computed properties any more in 3.x (0 or 1 not sure). Although you do for models since they use a proxy.

Caltor:
@knownasilya ah sorry we're still on 2.18.

JackEllis:
I use it a few times in a big codebase I work on but I see no problem with insisting on closure actions. And then this (https://github.com/DockYard/ember-route-action-helper) also exists

jrock2004:
I just went through an upgrade from 2.4 to 1.18 and had to change alot of stuff to closure actions. https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/closure-actions.md has a good guide of the old bad way to the new closure action way.

rwjblue:
We discussed this at todays core team meeting, and are in favor of moving forward.

Thanks to everyone who participated in this RFC!

ohhorob:
> Rendered

The link is broken; Is there another reference to the content of this RFC?

pangratz:
@ohhorob All RFCs are published on https://emberjs.github.io/rfcs, this one is at https://emberjs.github.io/rfcs/0335-deprecate-send-action.html

PR #337, opened 2018-06-15
Native Class Constructor Update

pzuraq:
[rendered](https://github.com/pzuraq/emberjs-rfcs/blob/native-class-constructor-update/text/0000-native-class-constructor-update.md)

### DIFF
--- a/text/0000-native-class-constructor-update.md
+++ b/text/0000-native-class-constructor-update.md
@@ -185,7 +185,7 @@ class EmberObject {
        return this;
     }
 
-    deprecate('using `new` with EmberObject has been deprecated. Pleas use `create` instead.', false, {
+    deprecate('using `new` with EmberObject has been deprecated. Please use `create` instead.', false, {
       id: 'object.new-constructor',
       until: '3.5.0'
     });
@@ -227,8 +227,20 @@ This would be a breaking change that could negatively affect early adopters.
 
 ## Alternatives
 
-We could leave the behavior as is, and choose a method for defaulting to
+* We could leave the behavior as is, and choose a method for defaulting to
 standardize on.
+
+* We could make this change behind a feature flag and require users to opt-in
+to the new behavior, like optional features that currently exist. This would
+have to be a build time feature flag, since the area is very performance
+sensitive. Given native classes are not yet public API, if we were to do this we
+should probably still default to enabling the new behavior and recommending it
+as the preferred path.
+
+* We could not deprecate `new EmberObject` altogether, and instead only
+deprecate passing properties to the constructor. While this would work as a
+temporary solution, it may also encourage users to continue using EmberObject
+instead of switching to native classes, which is ultimately the long term goal.
 
 ## Unresolved questions
 
--- a/text/0000-native-class-constructor-update.md
+++ b/text/0000-native-class-constructor-update.md
@@ -162,6 +162,9 @@ assigned to the instance.
 It was previously possible to use `new` syntax with EmberObject. While this
 was not considered public API, it has technically worked and been under test
 since the early days of Ember, and may fall under the category of intimate API.
+Ideally, we would deprecate this usage as a private/intimate API, which would
+mean supporting it through the next LTS version, and dropping support after
+(currently, this would mean dropping it at `v3.5.0`).
 
 We can continue to support this behavior in a backwards compatible way while
 deprecating it with one final tweak to the change above:
@@ -182,7 +185,11 @@ class EmberObject {
        return this;
     }
 
-    deprecate('using new with EmberObject has been deprecated', true);
+    deprecate('using `new` with EmberObject has been deprecated. Pleas use `create` instead.', false, {
+      id: 'object.new-constructor',
+      until: '3.5.0'
+    });
+
     initialize(this, props);
   }
 
--- a/text/0000-native-class-constructor-update.md
+++ b/text/0000-native-class-constructor-update.md
@@ -1,0 +1,220 @@
+- Start Date: 2018-06-14
+- RFC PR: (leave this empty)
+- Ember Issue: (leave this empty)
+
+# Native Class Constructor Update
+
+## Summary
+
+Update the behavior of EmberObject's constructor to defer object
+initialization.
+
+## Motivation
+
+Using native class syntax with EmberObject has almost reached full feature
+parity, meaning soon we'll be able to ship native classes and begin recommending
+them. This will do wonders for the Ember learning story, and will bring us in
+line with the wider Javascript community.
+
+However, early adopters of native classes have experienced some serious
+ergonomic issues due to the current behavior of the class constructor. The issue
+is caused by the fact that properties passed to `EmberObject.create` are
+assigned to the instance in the root class `constructor`. Due to the way that
+native class fields work, this means that they are assigned _before_ any
+subclasses' fields are assigned, causing subclass fields to overwrite any value
+passed to `create`:
+
+```js
+class Foo extends EmberObject {
+  bar = 'baz';
+}
+
+let foo = Foo.create({ bar: 'something different' });
+
+console.log(foo.bar); // 'baz'
+```
+
+This has made adoption very difficult, and is a consistent stumbling block for
+new users of native class syntax in Ember. Worse yet, it makes writing a codemod
+for converting to native class syntax very difficult because we don't have a
+clear target.
+
+For instance, given the above class, how would we convert the class field? Let's
+go through the various options:
+
+```js
+class Foo extends EmberObject {
+  // Does not work, for the reasons described above
+  bar = 'baz';
+
+  // Does not cover all cases. If we did `Foo.create({ bar: false })` it would
+  // still assign the default.
+  bar = this.bar || 'baz';
+
+  // This works, but is very verbose and not ideal
+  bar = this.hasOwnProperty('bar') ? this.bar : 'baz';
+
+  // This is one of the community accepted solutions, but it requires lodash
+  bar = _.defaultTo(this.bar, 'baz');
+
+  // This is another community accepted solution, but it requires
+  // @ember-decorators/argument, which is a separate library
+  @argument foo = 'bar';
+}
+```
+
+None of these is ideal. Instead, we can change the behavior of the constructor
+and the `create` method to circumvent this issue.
+
+This change _would_ be a breaking change to the behavior of native classes
+today, and a change from the previous class RFC. This will impact early adopters
+and should be made with that in mind. However, because native classes never
+officially shipped as part of Ember's public API (an announcement was not made,
+docs have not been written, etc), this RFC proposes that the change would _not_
+be considered a breaking change _for the purposes of semver_. This would allow
+us to ship the change during the Ember v3 release cycle, and prevent more code
+from being built on top of the previous behavior.
+
+## Detailed design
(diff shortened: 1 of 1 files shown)


text/0000-native-class-constructor-update.md
```diff
@@ -0,0 +1,220 @@
...
+}
+```
+
+Much easier to codemod! There are other subtle differences between native class
+fields and EmberObject properties, such as the fact that class fields are
+assigned each time a class is initialized, but these are easier to work around.
+
+### Injections and the `init` hook
+
+One side effect of this change is that injections will not be available on the
```
chriskrycho: It's worth noting that we've done an unofficial survey of some of the largest apps *known* to be using ES6 classes in production via TypeScript and have found something like 2 instances of the constructor using injected services *total* across them. I don't think we have enough data to say that in the RFC, but I wanted to bring it up here for the record in the RFC discussion.

text/0000-native-class-constructor-update.md
```diff
@@ -0,0 +1,220 @@
...
+and the `create` method to circumvent this issue.
+
+This change _would_ be a breaking change to the behavior of native classes
+today, and a change from the previous class RFC. This will impact early adopters
+and should be made with that in mind. However, because native classes never
+officially shipped as part of Ember's public API (an announcement was not made,
+docs have not been written, etc), this RFC proposes that the change would _not_
+be considered a breaking change _for the purposes of semver_. This would allow
+us to ship the change during the Ember v3 release cycle, and prevent more code
+from being built on top of the previous behavior.
```
chriskrycho: It's worth note that the design landed on here *does not break* the behavior of any of the fallback solutions the community has had to embrace to deal with the pain point.

As such, this is a case where it's a "breaking change" in the strictest sense but is *extremely* unlikely to break any real-world code, because the real-world code has necessitated the use of one of those workarounds.

text/0000-native-class-constructor-update.md
```diff
@@ -0,0 +1,220 @@
...
+
+We can document the exact behavior of the new constructor in the API docs for
+EmberObject. Most details won't have to change since this change only affects
+native class syntax, which has not been documented much officially. We can also
+demonstrate the behaviors of classes throughout the guides and API docs.
+
+One thing we should make clear is that EmberObject will likely be deprecated
+in the near future, and that ideally for non-Ember classes (things that aren't
+Components, Services, etc.) users should drop EmberObject altogether and use
+native classes only. For the purposes of codemodding, however, this will not be
```
chriskrycho: This last sentence is a bit unclear; I actually have no idea what it means. Could you reword?

text/0000-native-class-constructor-update.md
```diff
@@ -0,0 +1,220 @@
...
+possible, since EmberObject based classes still have many differences in common
+usage from standard native classes.
+
+## Drawbacks
+
+This would be a breaking change that would negatively affect early adopters.
```
chriskrycho: I'd suggest "could" rather than "would".

text/0000-native-class-constructor-update.md
pzuraq: Coming back to it, I think it was pretty unclear too so I'm going to drop it. Basically, it's not possible to easy codemod a class which was base on `EmberObject` to one which is _not_ based on `EmberObject`, for a number of reasons:

- Common usage of static methods (`create` is used everywhere, and tracking down and converting each usage would be _very_ difficult without types in non TS apps)
- Mixins. If they're used once in the inheritance tree, we can't convert.

The point is, we shouldn't be recommending that people use `EmberObject` at all now that we have native classes, and we should guide users toward converting any existing classes.

text/0000-native-class-constructor-update.md
```diff
@@ -0,0 +1,228 @@
...
+
+class EmberObject {
+  constructor(props, maybeDefer) {
+    // ..class setup things
+
+    if (maybeDefer === DEFER_INIT) {
+       return this;
+    }
+
+    deprecate('using new with EmberObject has been deprecated', true);
```
rwjblue: I'd like to see this fleshed out to include the deprecation timeline, what would the `until` be here? I'd _personally_ like to see this be considered a private API deprecation, where `until` would be the next LTS+1...

chriskrycho: I'd also support that timeline.

pzuraq: I think the sooner we can remove it, the better. Will update the RFC to state that it will be `until: '3.5.0'`

text/0000-native-class-constructor-update.md
```diff
@@ -0,0 +1,235 @@
...
+class EmberObject {
+  constructor(props) {
+    // ..class setup things
+  }
+
+  static create(props) {
+    let instance = new this(props);
+
+    Object.assign(this, props);
+    this.init();
```
jamesarosen: I think you mean
```js
Object.assign(instance, props);
instance.init()
```

rtablada:
I'm excited for native classes and better parity is 👏.

Deprecating `new Object` feels like a step backwards though. Requiring `create` seems off in the broader sense of reducing magic and making Ember more comfortable for other JS users.

pzuraq:
Yeah, I hear you there. Not being able to use `new` has been problematic in teaching the object model, and is definitely non-standard and not what we want in the future. I wanted to make sure that this decision was made in the larger context of the future of `EmberObject` in general, which is why I also put forward #338 at the same time as this RFC.

The only time users currently actually manually use `create` are either when A. they have extended `EmberObject` directly for their own classes or B. they are writing unit tests. Generally, all other Ember constructs that extend from `EmberObject` are managed by the container, which means users should never have to manually call `create` or `new` for them. 

As proposed in #338, the long term goal would be to remove `EmberObject` from the framework altogether. Classes that users have defined that directly extend `EmberObject` should actually become native classes, with no base class, and no constructor behavior. However, we _cannot_ codemod current classes that extend `EmberObject` to these, because of the static class methods like `create`. We would have to find every instance of the class being created and update it to do something like:

```js
let foo = new Foo();
Object.assign(foo, props);
```

This could work in _some_ cases, but I don't think it could for all of them safely. The idea with this change is that we do this conversion in two parts:

1. Make converting to native class syntax _with_ `EmberObject` codemoddable, and easy. Get the community on board, and get everything updated as quickly as possible.
2. Manually update classes that extend `EmberObject` to no longer need to extend it. This may mean adding a constructor to those classes, it may not. Every class will likely have its own needs and ideal API.

In either case, I believe we should start updating guides, blog posts, addon documentation, etc. to use native classes instead of extending from `EmberObject`, and generally start to treat it as deprecated and embrace native class syntax.

pzuraq:
Also, another alternative that has come up is potentially not deprecating `new EmberObject()`, but deprecating `new EmberObject(props)`. I think this would be a decent middle ground, although it could lead to some confusing branching behavior with `init`.

rtablada:
I think deprecating with props is better than deprecating `new` altogether as public (especially since there was an RFC just in the last year to make `new` more acceptable).

I'm ok with something like:

```
let report = new MyReport();
report.setProperties(props);
```

Where `create` is just a static shorthand for ☝️?

rtablada:
From a learning side of things I think we'd need to document the differences in

* constructor
* init
* create

pzuraq:
I would be OK with it, as long as we made it very, very clear that you should not be extending from `EmberObject` in general. Overall momentum should be towards writing pure native classes and converting existing classes, and being able to do

```js
let report = new MyReport();
setProperties(report, props); // minor tweak
```

Is essentially a forwards compatible way to do that.

text/0000-native-class-constructor-update.md
```diff
@@ -0,0 +1,235 @@
...
+and the `create` method to circumvent this issue.
+
+This change _would_ be a breaking change to the behavior of native classes
+today, and a change from the previous class RFC. This will impact early adopters
+and should be made with that in mind. It would _not_ be a change that breaks the
+behavior of the community solutions to class fields mentioned above, and all
+other changes would be relatively easy to create a safe codemod for (essentially
+converting `constructor` -> `init` in affected classes), so the impact _should_
+be minimal.
+
```
Gaurav0: An alternative not mentioned here is to make the change behind a feature flag enabled by an add-on. This would make the new behavior opt in. We are doing this already for making jQuery optional, for example.

pzuraq: That would be a reasonable alternative, we should definitely add it to the alternatives section.

In practice, I think it would be best to ship the target behavior as the actual/only public API and polyfill its behavior to older versions of Ember (already looking into how we can do that, it _should_ be possible). Unlike a lot of the other optional feature flags, we don't _technically_ have a public API that we need to continue to support.

This does beg the question, when does an RFC spec become public API? The process in intentionally fuzzy here, because details may change in implementation (see Named Blocks).

In this case, we've talked to lots of early adopters and they have so far been very supportive of this late-stage update to the original RFC. However, if there are any users of native class syntax out there who would be very against it, we definitely want to hear from them!

oligriffiths:
I'd like to second the points raised that advising again using `new` seems like a backwards step.
It seems like allowing `new` *WITH* `props` is the thing that we don't want to support, due to needing these backwards compatible hacks.
It may be worth moving `EmberObject.create` to a separate method that is not inherited by subclasses, something like:

```js
// @ember/object`

class EmberObject {
  constructor(props) {
    // ..class setup things
  }

  static create(props) {
    return createObject(this, props);
  }
}

function createObject(className, props){ 
    const instance = new className;
    if (typeof props === 'object') { 
        Object.assign(instance, props); 
    }
    return instance; 
}

export default EmberObject;
export createObject;
```
```js
import createObject from @ember/object;

const foo = createObject(Foo, {my: 'prop'});
```

pzuraq:
To be clear, I would say the advisement is _not_ to not use `new`, it's to not use `EmberObject`. In general, modern Ember apps can be written without using `EmberObject` or `.create()` at all, and we should be recommending that folks actually use native constructors the way they were meant to be used:

```js
// Before
let MyClass = EmberObject.extend();

let foo = MyClass.create({ bar: 'baz' });

// After
class MyClass {
  constructor(bar) {
    this.bar = bar;
  }
}

let foo = new MyClass('baz');
```

This approach is the way native classes in Javascript are moving in general and what we want to target. Mass assigning properties to the class via `Object.assign` is not really inline with the way native classes work, and not something we should be recommending, so my only concern with continuing to allow `new EmberObject()` is that it would actually encourage folks to keep using `EmberObject` and not actually transition to native classes, like the above.

oligriffiths:
> it would actually encourage folks to keep using EmberObject and not actually transition to native classes, like the above.

Makes total sense. Thanks

rwjblue:
Discussed this in todays Ember.js Core Team meeting. Some of the take aways from the convo were:

* We should reach out to @wycats for verbiage help in the `new EmberObject()` deprecation (and deprecation guide) 
* Guidance should continue to be "use `init`", not `constructor` when extending from a base class that extends from `Ember.Object`
* We should include a small section in the object model section of the guides explaining _why_ we still advise using `init` for now.

None of these were blocking concerns, and we are collectively 👍 on moving this into final comment period.

rwjblue:
There has been no additional feedback since this RFC entered the final comment period, so its time has come...

locks:
Can you update the "How we teach this" section?

PR #338, opened 2018-06-15
Native Class Roadmap

pzuraq:
[rendered](https://github.com/pzuraq/emberjs-rfcs/blob/b47e7f9ec4f02c7d27d50de64691130e7d22747d/text/0000-native-class-roadmap.md)

### DIFF
--- a/text/0000-native-class-roadmap.md
+++ b/text/0000-native-class-roadmap.md
@@ -1,0 +1,265 @@
+- Start Date: 2018-06-14
+- RFC PR: (leave this empty)
+- Ember Issue: (leave this empty)
+
+# Native Class Roadmap
+
+## Summary
+
+Now that we’ve (almost) achieved feature parity for using native classes to
+extend from EmberObject, the next step for Ember is to move toward using and
+recommending native classes by default, with the long term goal of deprecating
+EmberObject and removing it from the framework. This RFC seeks to outline where
+we want to go (e.g. what Ember looks like without EmberObject), constraints we
+face, and potential paths forward.
+
+## Design Constraints
+
+In order to figure out the path forward to native classes, we need to first
+develop an idea of what the future of Ember looks like once we have fully
+adopted them and removed EmberObject entirely. Below are several design
+considerations for the new object model:
+
+- **It should leverage the platform.** We should use and encourage standard
+Javascript classes and idioms wherever possible. This means leveraging classes,
+class fields, and decorators for most features that the object model provides.
+- **Ember should remain as neutral as possible about the implementation details.**
+Part of the reason we are going to have migration pains as a community moving
+forward is that Ember is tied to the behavior of EmberObject in a myriad of
+small ways. In an ideal world, these details would be opaque to Ember, allowing
+users to implement and change-out object models at their own convenience. This
+would have made the transition much easier, and will provide flexibility in the
+community for pursuing new language features and libraries (e.g. mixins)
+- **It should be possible to adopt incrementally.** This will be one of the largest
+changes the community has ever gone through, on the same level if not larger than
+dropping Views and adopting Components. EmberObject is pervasive - it's been
+used for every construct imaginable in Ember apps and addons due to its
+versatility and simplicity. Ideally users will be able to make incremental
+changes toward pure native classes by first rewriting their existing classes
+using native class syntax (as is already possible) and then dropping EmberObject
+altogether.
+- **We should consider the long tail of adoption, particularly for addons.** There
+will be a long period where the community will need to support older versions of
+Ember at the same time as newer versions. Ideally it should be possible to
+support both the new and old model at once without separate code paths, or the
+overlap period should be long enough that addons can drop EmberObject without
+breaking Ember versions that are still part of their support matrix.
+- **If possible, conversion should be automate-able.** This is not a hard
+constraint, since it limits us to having very similar behavior to the existing
+object model. However, if it is possible to create scripts that speed up
+adoption it will help the community tremendously, so if we are at an impasse on
+particular behaviors or implementations, we should consider whether one would be
+easier to automate towards.
+
+## Detailed design
+
+As mentioned above, one of the main reasons this conversion is going to be a
+difficult and long process is the fact that both user code and Ember code are
+very intertwined with the existing object model. Specifically:
+
+- They rely on the behavior of static methods such as `create`, `extend`, and
+  `destroy`.
+- They rely on the behavior of EmberObject's constructor
+- They rely on a non-trivial amount of functionality provided via inheritance
+- They rely on features of EmberObject's inheritance model that do not exist in
+native Javascript (Mixins, concatenated/merged properties, nuances of observers
+and listeners, etc.)
+
+The fact that these are so intertwined makes it very difficult (if not
+impossible) for us to rewrite the core classes that exist in Ember without
+breaking user land code.
+
+One strategy we can implement to prevent this from happening again in the future
+is to make the details of classes opaque to Ember. Rather than providing base
+classes that can be extended by users to add functionality and implement their
+applications, we can rewrite Ember to treat user classes as delegates. In other
+words, Ember will make no assumptions about the behavior of a class, other than
+that it fulfills a certain interface.
+
+For instance, a Component can be any class which receives an `args` hash and
+optionally implements any of the component lifecycle hooks:
+
+```js
+export default class FooComponent {
+  didInsertElement() {
+    if (this.args.isShowing) {
+      // do something
+    }
+  }
+}
+```
+
+This method allows users to provide base-less classes, classes which do not
+extend anything, to Ember as well. Ember could continue to provide empty base
+classes to extend from, but it wouldn't have to.
+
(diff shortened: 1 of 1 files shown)


text/0000-native-class-roadmap.md
```diff
@@ -0,0 +1,265 @@
...
+
+One strategy we can implement to prevent this from happening again in the future
+is to make the details of classes opaque to Ember. Rather than providing base
+classes that can be extended by users to add functionality and implement their
+applications, we can rewrite Ember to treat user classes as delegates. In other
+words, Ember will make no assumptions about the behavior of a class, other than
+that it fulfills a certain interface.
+
+For instance, a Component can be any class which receives an `args` hash and
+optionally implements any of the component lifecycle hooks:
```
snewcomer: Non important, curious question - Would this in effect reduce lookup time for properties/non existent properties, especially for things are are high up the prototype chain?

text/0000-native-class-roadmap.md
```diff
@@ -0,0 +1,265 @@
...
+pattern, since Ember would have to support both the old Ember object model and
+the new base classes at the same time until the switch. It would also likely be
+fairly confusing to users, since they would have to make sure they were
+importing the correct classes.
+
+## Unresolved questions
+
+If we go this direction, it means we cannot specify any base class behavior. One
+issue that arises is the fact that dependency injections will not be available
+to the class _during_ object construction. They can be assigned _after_ only.
```
snewcomer: Does this result affect initializers/instance-initializers?

pzuraq: Initializers and instance-initializers are (as I understand them) essentially pure functions that run at the beginning off app creation/instantiation. They would still receive the app definition/instance and container as such, and wouldn't be directly impacted by this change.

However, and injections which are setup by initializers _would_ be affected. All injections are currently assigned by this mechanism, and assigning after would make them unavailable. 

It's worth noting that only potential solutions mentioned in the RFC which would work for injections done this way would be the `init` hook and passing the injections to the constructor as named parameters. Personally, I am not a fan of the ability to unilaterally inject items onto any type thing (e.g. `store` automatically injected to routes/controllers) and think it would make sense to recommend users manually declare the dependency.

text/0000-native-class-roadmap.md
pzuraq: In theory yes, though I wouldn't count on it being that huge of a win. In aggregate though, who knows 😄

text/0000-native-class-roadmap.md
raido: Perhaps in v4 we have init() to support this and in Ember v5 we drop it and make everything explicit?

lupestro:
Super minor text nit and personal pet peeve: “a myriad of”
“Myriad” is a number, not a collective noun. It means “ten thousand”. If swapping in “ten thousand” sounds funny, make that sound right, then go back to myriad, so “myriad things” is better usage. 
(I will now retire my fussy school-marm hat for the evening. :) )
This was a *great* writeup. I’m glad the ball is rolling on this.

rtablada:
A thought on Mixin migration. I'm not 💯sure of the validity or plausibility of this, but would it make sense to have a legacy `apply-mixin` function for use in classes?

Here's an example I'm thinking (inspired by 

```js
import ApplicationRouteMixin from 'ember-simple-auth/mixins/application-route-mixin';
import applyMixin from '@ember/apply-mixin';

class ApplicationRoute extends Route {
  // Stuff
};

export default applyMixin(ApplicationRoute, ApplicationRouteMixin);
```

Or a different  API might be more like:

```js
import ApplicationRouteMixin from 'ember-simple-auth/mixins/application-route-mixin';
import createClassFromMixin from '@ember/create-class-from-mixin';

const createApplicationRoute = createClassFromMixin(ApplicationRouteMixin, ...any other mixins);

class ApplicationRoute extends Route {
  // Stuff
};

export default createApplicationRoute(ApplicationRoute);
```

pzuraq:
Pre-removal of `EmberObject`, I'm not sure it would be necessary. We can still use `class Foo extends EmberObject.extend(Mixin)`, so they are effectively supported. It would just add another way to do the same thing, though it may be clearer. I'm not opposed to it personally.

Post-removal of `EmberObject`, I don't think it should be part of Ember. It would be very difficult, if not impossible, to maximize all of the benefits of removing `EmberObject` and switching to native classes, _and_ make an `applyMixins` which would faithfully apply the mixins and result in the exact same final class. 

That said, I'm considering lots of tiny edge cases and small behaviors that are tied to Ember/Mixins. I think there's definitely room to _make_ a library that applies mixins in _mostly_ complete fashion, so legacy users can use mostly the same mixin definitions with relatively minor changes. I also think that there is room for a much better general mixin solution that actually leverages native classes (for inspiration, see [this article that proposes one such system](http://justinfagnani.com/2015/12/21/real-mixins-with-javascript-classes/)).

dfreeman:
The wider JS community (or at least a vocal subset) seems to like that 'functional mixin' pattern — the [pipeline operator proposal](https://github.com/tc39/proposal-pipeline-operator#mixins) calls out how it would make applying them nicer, and the original author of that post has [a stage 1 proposal](https://github.com/justinfagnani/proposal-mixins) for formalizing them with dedicated syntax.

nruth:
@lupestro the myriad usage is fine going by the [OED](https://en.oxforddictionaries.com/definition/myriad). Hoping Ember isn't classical history :)

lupestro:
@nruth <sigh/> And I hear rumors that prepositions are now accepted to end sentences with. <Oh/><Ow/> <Ow/> (“...mere anarchy is loosed upon the world...”) I am finally old enough to experience the eternal rules of language changing in even one brief lifetime. (But of course they have ever done so...) NVM, carry on, carry on! 😀

text/0000-native-class-roadmap.md
pzuraq: I think that would be an option if we decided to go with the other DI options and drop `init` altogether. It would give us a good timeline to clean things up on as well, and falls neatly inline with the proposed timeline for removing EmberObject.

text/0000-native-class-roadmap.md
```diff
@@ -0,0 +1,265 @@
...
+
+    constructor(time, date) {
+      this.currentTime = time.now();
+      this.currentDate = date.today();
+    }
+  }
+  ```
+
+
+3. **Utilize private global state to provide the container to injections during
```
luxferresum: I like this, but then it should be done for `getOwner()` so that even a manual call to `getOwner(this)` inside the constructor will work.

pzuraq: I agree, this would make the most sense if we pursued this strategy. Will update the RFC to reflect that

pzuraq:
I had a great conversation with @mike-north the other day about this, he pointed out a few concerns with this pattern going forward. I'm posting them here for discussion, but they should definitely be addressed by updates to the RFC at some point:

1. Interface-based design can be very cumbersome dependending on the way it is done. Other frameworks have implemented it, in particular Angular 2+, in a form where interfaces are scoped done to very specific functionality that _any_ class can implement. For instance, the `onInit` interface tells Angular that a class should fire the `ngOnInit()` hook:

  ```ts
  import { Component, OnInit } from '@angular/core';
  
  @Component({})
  export class HeroDetailComponent implements OnInit {
    ngOnInit() {
    }
  }
  ```

  This can be very confusing, because apparently without the `OnInit` interface being used `ngOnInit` will not fire at all (part of Angular's inferencing of type information). It can also become combinatorially complex, as every lifecycle hook must become an interface, resulting in massive unparseable component definitions.

  When writing this proposal, I envisioned interfaces representing a full type, like they are today. Typescript users would actually end up with very similar code to what exists currently:

  ```ts
  import { Component } from '@ember/interfaces';

  export default class HeroDetailComponent implements Component {
    didInsertElement() {

    }
  }
  ```

  And native Javascript users would not extend a class at all. This would keep definitions of interfaces and types minimal and not lead to combinatorial complexity, though it could be less flexible than the alternative.

2. If base classes are not used, then there is no way to specify the placement of the call to `super` for a hook. This means that users cannot easily execute code both before and after a hook, which is less flexible:

  ```js
    didInsertElement() {
      // do something before

      super.didInsertElement();

      // do something after
    }
  }
  ```

  This can be mostly mitigated with "bookend" hooks, such as `willInsertElement`/`didInsertElement`, `willDestroy`/`didDestroy`, etc. It should also be noted that users can still maintain their own class hierarchy's and control calls to super in those, so this would only be a concern with framework code.

3. If base classes are not used, there is no easy way to intercept arguments to a hook, or the return value of a hook. The effects here have to be studied more closely, as there are times when arguments should generally _not_ be used or modified, and times where it may make sense to modify them. It may also be possible to modify arguments in bookend hooks, though that would come with its own set of problems and concerns.

tomdale:
Removing the Octane label from this, as this is a more future-looking RFC that needs to be updated to reflect the current state of the world, and is unlikely to happen in the Octane timeframe.

pzuraq:
This RFC is pretty out if date at this point. Closing it for now, and will reopen sometime post Octane when we’ve had time to update it.

PR #339, opened 2018-06-15
Router link component and routing helpers

NullVoxPopuli:
my first RFC! 

[rendered](https://github.com/emberjs/rfcs/blob/5c6beae045ae4863828b9823414b80849c8d5308/text/0000-router-link-and-helpers.md)

### DIFF
--- a/text/0000-router-link-and-helpers.md
+++ b/text/0000-router-link-and-helpers.md
@@ -1,0 +1,56 @@
+- Start Date: (fill me in with today's date, YYYY-MM-DD)
+- RFC PR: (leave this empty)
+- Ember Issue: (leave this empty)
+
+# (RFC title goes here)
+
+## Summary
+
+> One paragraph explanation of the feature.
+
+## Motivation
+
+> Why are we doing this? What use cases does it support? What is the expected
+outcome?
+
+## Detailed design
+
+> This is the bulk of the RFC.
+
+> Explain the design in enough detail for somebody
+familiar with the framework to understand, and for somebody familiar with the
+implementation to implement. This should get into specifics and corner-cases,
+and include examples of how the feature is used. Any new terminology should be
+defined here.
+
+## How we teach this
+
+> What names and terminology work best for these concepts and why? How is this
+idea best presented? As a continuation of existing Ember patterns, or as a
+wholly new one?
+
+> Would the acceptance of this proposal mean the Ember guides must be
+re-organized or altered? Does it change how Ember is taught to new users
+at any level?
+
+> How should this feature be introduced and taught to existing Ember
+users?
+
+## Drawbacks
+
+> Why should we *not* do this? Please consider the impact on teaching Ember,
+on the integration of this feature with other existing and planned features,
+on the impact of the API churn on existing apps, etc.
+
+> There are tradeoffs to choosing any path, please attempt to identify them here.
+
+## Alternatives
+
+> What other designs have been considered? What is the impact of not doing this?
+
+> This section could also include prior art, that is, how other frameworks in the same domain have solved this problem.
+
+## Unresolved questions
+
+> Optional, but suggested for first drafts. What parts of the design are still
+TBD?
--- a/text/0000-router-link-and-helpers.md
+++ b/text/0000-router-link-and-helpers.md
@@ -1,6 +1,6 @@
 - Start Date: 2018-06-15
-- RFC PR: 
-- Ember Issue: 
+- RFC PR:
+- Ember Issue:
 
 # Add a Link component and routing helpers
 
@@ -14,60 +14,104 @@ Currently, on ember#canary, the Angle Bracket Invocation feature doesn't support
 
 ## Detailed design
 
-Helpers to be implemented:
- - `transition` 
-    this is a helper that would tie into the router to be able to achive the same functionality as the simple usage of `LinkTo`
+#### tl;dr:
 
-    example:
-      
+*Phase 1*
+- add named arguments to `LinkTo`: `@routeName` and `@models`
+- add built-in helpers for custom link-like behavior
+
+*Phase 2*
+- add `Link` component with `@to` named argument -- all other named arguments will be the params
+- deprecate `LinkTo`
+
+*Phase 3*
+- remove `LinkTo` and the corresponding deprecation
+
+
+#### Phase 1
+*Add Named Arguments to `LinkTo`*
+
+  `@routeName` and `@models` will allow us to smoothly transition away from `{{link-to}}` into the `<AngleBracket />` invocation-style.
+
+  example:
+
+    ```hbs
+      {{!-- before --}}
+      {{#link-to 'posts.edit' post.id}}
+        Edit
+      {{/link-to}}
+
+      {{!-- After --}}
+      <LinkTo @routeName='posts.edit' @models={{hash postId=post.id}}>
+        Edit
+      </LinkTo>
+    ```
+
+  The `<AngleBracket />` invocation with named arguments is more verbose, but trade-off here is the added clarity of what argument is used for what purpose -- especially with respect to the models / model ids.
+
+*Helpers to be implemented*
+
+ - `transition`
+    this is a helper that would tie into the router to be able to achieve the same functionality as the navigational usage of `LinkTo`
+
+    examples:
+
+      ```hbs
+      <button {{action (transition to='posts.edit' postId=post.id)}} />
+        Edit
+      </button>
+
+      <button {{action (transition to='login' replace=true)}} />
+        Login
+      </button>
       ```
-      <button {{action (transition to='posts.edit' postId=post.id}}
-      ```
-    
- - `replace-with`
-    this is same as transition, but replaces the current URL and current history entry. This may be an argument to transition instead. e.g.: `(transition ... replace=true)`
-    
+
+
+
  - `is-route-active`
     this returns a boolean representing whether or not the current route matches the passed argument.
-    
+
     example:
-    
+
+      ```hbs
+      <button
+        {{action (transition to='posts.edit' post=model.post)}}
+        class={{if (is-route-active 'posts.edit' post=model.post) 'selected'}} />
       ```
-      <button 
-        {{action (transition to='posts.edit' post=model.post)}} 
-        class={{if (is-route-active name='posts.edit' post=model.post) 'selected'}} />
-      ```
-      
+
    this may need to support multiple invocation styles, for when there aren't parameters and the dev wants to be concise:
 
-   example:
-   
-     ```
+   examples:
+
+     ```hbs
      <button ... class={{if (is-route-active 'posts') 'selected'}} />
+      No model params provided
+     </button>
      ```
 
-Inspiration for helpers implementation could come from https://github.com/BBVAEngineering/ember-route-helpers
+#### Phase 2
 
- Components to be implemented:
- - `Link`
- 
+*Add `Link` Component*
+
 `Link` could just be tagless component with the following template:
+
 ```hbs
 {{!-- components/link/template.hbs --}}
-<a 
-  {{action (transition ...@to)}} 
-  class="{{if (is-route-active ...@to) 'active'}}"
->
+<a
+  {{action (transition @to models=@models)}}
+  class="{{if (is-route-active @to models=@models) 'active'}}">
 
-{{yield}}
+  {{yield}}
 
 </a>
- ```
+```
 
 Usage:
 ```hbs
-<Link @to=(array 'posts.edit' post)>Edit Post</Link>
+<Link @to='posts.edit' @models={{post}}>Edit Post</Link>
 ```
+
+where `@models` can be a `hash`, `array`, or just a single object or id.
 
 
 **Deprecation** `LinkTo` aka `link-to`
@@ -93,7 +137,7 @@ Without positional params, the alternative may need to use the `array` helper fo
 
 There are two alternatives:
 
-1. Do nothing: 
+1. Do nothing:
   - people will find that the `LinkTo` usage with angle bracket invocation to be somewhat awkward, maybe unintuitive.
 2. Only implement the routing helpers
   - people could define their own linking components however they desire
@@ -105,3 +149,9 @@ There are two alternatives:
 - For routes with multiple segment parameters, it may become hard to keep track of things. e.g.: `(array 'posts.edit.comments.reply' post comment.id)` how do I know where the parameters are going to go without looking at the router.js file? (though, this is an existing question)
 
 - Would it make sense to somehow have a way to statically confirm that the route path is valid? right now, it looks like a string -- how would I know if I typed it wrong?
+
+## Inspiration / Code taken from
+- https://github.com/alexspeller/ember-cli-active-link-wrapper
+- https://github.com/BBVAEngineering/ember-route-helpers
+- https://github.com/peec/ember-transition-helper
+- https://github.com/rwjblue/ember-router-helpers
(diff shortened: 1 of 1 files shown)


text/0000-router-link-and-helpers.md
```diff
@@ -0,0 +1,107 @@
...
+   example:
+   
+     ```
+     <button ... class={{if (is-route-active 'posts') 'selected'}} />
+     ```
+
+Inspiration for helpers implementation could come from https://github.com/BBVAEngineering/ember-route-helpers
+
+ Components to be implemented:
+ - `Link`
```
rwjblue: I believe that we can be more iterative here, and add the following named arguments to the existing `link-to`:

* route (the route to link to)
* models (an array of model / primitive ids / etc)

This would make transforming existing `{{link-to}}` usages very easy, and still largely satisfy (what I thought) was the main concern: desire to use the newer angle bracket invocation style.

NullVoxPopuli: desire to use the newer angle bracket invocation style was my original concern.  but I couldn't come with up with named argument for `link-to` that didn't feel awkward, or implicitly imply functionality that might not be there, such as would be the case with `{{link-to route=''}}` (what else could we link to?)

rwjblue: Ha! I was actually thinking exactly the opposite. `<Link` seems like it should manage _all_ links, not just ones within the routing structure.

Also, I think there is value in iterating forward within the component we all know / love / hate / etc. Adding a couple of named arguments to `{{link-to}}` seems like it would unlock using angle bracket invocation and be a fairly small change/tweak to the internal code.

NullVoxPopuli: lol, guess the only way forward is with the smallest amount of changes. I'm totes good with named args in linked to. :)
less change = less potential for confusion

rtablada:
I think naming it `RouterLink` would make it clearer that this is router coupled.

text/0000-router-link-and-helpers.md
rtablada: I would be 👍for making `LinkTo` work with named attrs and even named models.

I actually looked at this when I was first playing with <AngleBracket> components.

With a router.js (ignore that the route inheritance wouldn't really make sense

```
this.route('profile', { path: '/:userId' }, function() {
	this.route('friend', { path: '/friend/:friendId' })
})
```

My ideal `LinkTo` invocation would be

```
<LinkTo @routeName="profile.friend" @models={{hash userId=model.user.id friendId=friend.id}}>Some Test</LinkTo>
```

I think this could actually be done completely backwards compatible while allowing positional params too (with deprecation later).

gossi:
I would rather see `<Link @to='route.name'>...</Link>` for simple use cases that don't need params. For those that need them: `<Link @to='route.name' @params=(array ....)>...</Link>` that might be an easier to grasp API, especially for beginners.

text/0000-router-link-and-helpers.md
NullVoxPopuli: ooo, @rtablada I like that

NullVoxPopuli:
So, how about this:

Phase 1:
 - add named arguments to `LinkTo`: `@routeName` and `@models`

Phase 2:
 - add `Link` component with `@to` and `@params` 
 - deprecate `LinkTo`

_Question_: RouterLink vs Link? (personally, I favor `Link`, though, because components aren't imported like they are in react, maybe having the context of `RouterLink` or `Router::Link` (or however namespaces or going to work) would make sense, too? idk)

Phase 3:
- remove `LinkTo` and deprecation

sandstrom:
I haven't read the full RFC yet (I will).

One issue that immediately comes to mind is this discussion:
https://github.com/emberjs/rfcs/pull/275

This addon is also related to the aforementioned RFC:
https://github.com/alexspeller/ember-cli-active-link-wrapper

rtablada:
@nullvoxpopuli @params might leak internal component implementation and I don’t think we could use that attribute name for a long time as Ember component uses @params for positional params.

Maybe I’m missing something but after adding named args, why would we need to make a completely different component with different attribute names? Seems counter intuitive? At best renaming to the Router namespace.

text/0000-router-link-and-helpers.md
rtablada: The names are a bit different and I would need to ask someone a bit more knowledgeable with the router but here's a proof of concept of the changes needed to make params or named attributes work:

https://github.com/rtablada/link-to-params/commit/5236ac5d3495faf59879a760fb30ccab4f6dfaaf

The links I tested this with are:

```hbs
<LinkTo @params={{array "foo" 1 2}}>Positional Params</LinkTo>

<LinkTo @targetRouteName="foo" @models={{array 1 2}}>Named Params</LinkTo>
```

NullVoxPopuli: hmm, we'd probably want to alias `targetRouteName` for the public api, yeah?
I'm glad what you have is so simple though!

chrism:
I really like idea of a `Link` component with named arguments and deprecating the current `link-to`.

My comment is more of an open question—could we use this as an opportunity to also address the beginner bear trap of passing a model versus an id ie whether or not the model hook gets fired on the route?

I feel like this might be a good moment to see if there is a way to reduce the potential confusion, though I don't feel confident proposing a solution myself.

chrism:
> I think if the params/model argument has a hash, it could help? having only positional args really confused me when I was first learning ember.

Definitely. I think that using the convention of `modelname` vs `modelId` to highlight the distinction is a great idea—especially if the documentation makes it clear exactly what the difference means in practice (ie whether or not to use the model hook).

Whether or not that would completely resolve the issue I don't know, but my feeling is that it would be a big help in clarifying the two different use cases.

btecu:
I wonder if there's anything we can borrow from `ember-href-to`.
cc @GavinJoyce

gossi:
If that works out, since `@postId` is some kind of a wildcard argument to the component (couldn't be defined at creating the component, since it is unknown to it). If it does, I'm happy in favor of it.

luxferresum:
@gossi first this is ember internal so it could use private API.

But I think its even possible with public API. `Object.keys(this.attrs)`/`Object.keys(this.args)` inside `didReceiveAttrs` should do the trick:

```
    didReceiveAttrs() {
      this.set('to', Object.keys(this.attrs).reduce((to, key) => {
         to[key] = this.attrs[key]
         return to;
      }, {}));
    }
```

However then we need a way to pass it to the `transition` and `is-route-active` helpers. Because we don't have a `...` operator in handlebars yet we can't do `{{action (transition ...this.to)}}`, but essentially thats what we want. So maybe `transition could accept a special named argument  which could be used for this:

```
    {{transition to=@to params=this.to}}
```

while this could be useful as public API it could kept private and only used by the `<Link>`  component.

---

could it maybe also be a good idea to expose a `href-to` helper? because the current proposed `<Link>` component misses a `href`, and a custom `href` could be useful for a custom version of the `<Link>` component. Essentially this is `ember-href-to`.

---

And one last thing, maybe it should be specified what should happen if the user does `<Link to='posts' onclick={{action 'foo}}>`.

text/0000-router-link-and-helpers.md
```diff
@@ -0,0 +1,161 @@
...
+
+The purpose of this is to add a single component, `Link`, that wraps up some routing helpers for default usage of linking / routing.
+
+## Motivation
+
+Currently, on ember#canary, the Angle Bracket Invocation feature doesn't support positional arguments. For clarity, and with some inspiration from react-router, It may be beneficial for Ember to include a `Link` component, similar to the existing `LinkTo` (`link-to`), but with only named arguments. This will help with clarity of intent, and also allow for easier transitioning from other communities who already have named-argument links.
+
+## Detailed design
+
+#### tl;dr:
```
locks: You went from an H2 to an H4 here.
I would suggest dropping this title, however, and writing a small intro explaining your rationale regarding splitting the work into phases, and why each specific phase.

text/0000-router-link-and-helpers.md
```diff
@@ -0,0 +1,161 @@
...
+
+# Add a Link component and routing helpers
+
+## Summary
+
+The purpose of this is to add a single component, `Link`, that wraps up some routing helpers for default usage of linking / routing.
+
+## Motivation
+
+Currently, on ember#canary, the Angle Bracket Invocation feature doesn't support positional arguments. For clarity, and with some inspiration from react-router, It may be beneficial for Ember to include a `Link` component, similar to the existing `LinkTo` (`link-to`), but with only named arguments. This will help with clarity of intent, and also allow for easier transitioning from other communities who already have named-argument links.
```
locks: Positional arguments are not planned for angle bracket invocations, I suggest updating this sentence to remove any "but in the future it might" implicit suggestion.

text/0000-router-link-and-helpers.md
```diff
@@ -0,0 +1,161 @@
...
+</a>
+```
+
+Usage:
+```hbs
+<Link @to='posts.edit' @models={{post}}>Edit Post</Link>
+```
+
+where `@models` can be a `hash`, `array`, or just a single object or id.
+
```
locks: Extraneous newlines, I think?

text/0000-router-link-and-helpers.md
```diff
@@ -0,0 +1,161 @@
...
+
+
+
+**Deprecation** `LinkTo` aka `link-to`
+
+The goal of `Link` and the route helpers is to provide a flexible way of routing from the template while providing a sample component with sensible defaults. This would achieve the exact same functionality as `LinkTo`, so `LinkTo` would no longer be needed and could eventually be removed.
+
+It's possible we could write a codemod to auto-convert everyone's non-angle-bracket invocation of `{{#link-to ...` to the new angle bracket component: `Link`
```
locks: This feels more of an open question than a detailed design, due to the uncertainty of the proposal.

text/0000-router-link-and-helpers.md
```diff
@@ -0,0 +1,161 @@
...
+
+## How we teach this
+
+We'll want to make it very clear that the traditional `LinkTo` technique of linking will still be available, but will eventually be deprecated.
+
+The documentation and guides would need to be incrementally upgraded to inform people about the new linking strategy -- but becaus the old way would still work, having some docs say `LinkTo` instead of `Link` wouldn't be the worst thing.
```
locks: Typo in "becaus"

text/0000-router-link-and-helpers.md
```diff
@@ -0,0 +1,161 @@
...
+
+The biggest drawback is that all routing documentation would be out of date, and there is a lot of routing documentation and blog posts throughout the web.
+
+Without positional params, the alternative may need to use the `array` helper for the `@to` argument... which could almost make people wonder why `LinkTo` didn't get a named argument for the route / route params. With the `Link` component, the arguments would read more ergonomically -- `<Link @to=...` vs `<LinkTo @route=...`. The latter implies that things other than routes could be linked to via the `LinkTo` component, which could cause confusion about the intended usage.
+
+## Alternatives
+
+There are two alternatives:
+
+1. Do nothing:
```
locks: I'm thinking we should skip including this, because that is the alternative for every RFC, and the body of the RFC is an active argumentation against doing nothing.

courajs:
Why bother adding the named params to link-to, if we're immediately afterwards going to add a new component to supersede it? It seems like unnecessary work - just introduce the new component and deprecate the old one.

NullVoxPopuli:
> Why bother adding the named params to link-to, if we're immediately afterwards going to add a new component to supersede it? It seems like unnecessary work - just introduce the new component and deprecate the old one.

@courajs I'd be a fan of this. If other people are, I'll remove those relevant parts of the RFC. :)
(and then also address @locks' comments)

NullVoxPopuli:
Updated. Also started some work over here: https://github.com/rwjblue/ember-router-helpers/pull/46

buschtoens:
Things that seem to be missing:

- `href`
- `replace=true`

NullVoxPopuli:
@buschtoens thanks for bringing those up! Those lend themselves nicely to named arguments :)

text/0000-router-link-and-helpers.md
```diff
@@ -114,46 +89,34 @@ Usage:
...
 
+Starting with @rwjblue's [ember-router-helpers](https://github.com/rwjblue/ember-router-helpers) as a base, `is-active` is already implemented, along with `route-params`, `transition-to`, and `url-for`.
 
+A [PR](https://github.com/rwjblue/ember-router-helpers/pull/46) has been started that aims to implement the `transition` helper (or just rename `transition-to`), as well as implement the `Link` component.
 
 
-**Deprecation** `LinkTo` aka `link-to`
 
-The goal of `Link` and the route helpers is to provide a flexible way of routing from the template while providing a sample component with sensible defaults. This would achieve the exact same functionality as `LinkTo`, so `LinkTo` would no longer be needed and could eventually be removed.
+**Deprecation** `link-to` aka `link-to`
```
gossi: that doesn't make sense anymore

gossi:
I saw, you changed some of the `is-route-active` helpers to `is-active` (not all, you missed one :p). Though the `is-active` is very generic and has a potential for conflicts (also MU works on `{{use}}`), this should be precise like the `is-route-active` before.

text/0000-router-link-and-helpers.md
NullVoxPopuli: I edited too quickly. lol

NullVoxPopuli:
@gossi I think that makes sense -- I'll revert to is-route-active :)

NullVoxPopuli:
@buschtoens `replace` is mentioned in the `transition` helper.
and `href` would just be the anchor attribute yeah?

(I'll add `...attributes` to the example implementation of `Link`)

NullVoxPopuli:
@buschtoens, @gossi:  updated

luxferresum:
Just to clarify any misunderstanding:

I think we also need a `href` helper. Something like `(href to='item' item='12')`. This is needed for the `Link` component. The `Link` component should always have a correct `href` attribute. This is currently true for the `link-to` component.

NullVoxPopuli:
thanks @luxferresum I'll get that added! :)
and also add it to ember-router-helpers if it isn't in there already. 

The goal of this RFC is to have equivalent capabilities from `{{link-to}}`, but with any custom implementation.

buschtoens:
I made [`ember-link`](https://github.com/buschtoens/ember-link) which gives you a renderless version of `<LinkTo>` and complies with the API proposed in [RFC 459 "Angle Bracket Invocations For Built-in Components"](https://github.com/emberjs/rfcs/blob/angle-built-ins/text/0459-angle-bracket-built-in-components.md#linkto), but is reasonably close to this PR.

```hbs
<Link
  @route="some.route"
  @models={{array 123}}
  @query={{hash foo="bar"}}
as |l|>
  <a
    href={{l.href}}
    class={{if l.isActive "is-active"}}
    {{on "click" l.transitionTo}}
  >
    Click me
  </a>
</Link>
```

text/0000-router-link-and-helpers.md
```diff
@@ -0,0 +1,124 @@
...
+
+*Helpers to be implemented*
+
+ - `transition`
+    this is a helper that would tie into the router to be able to achieve the same functionality as the navigational usage of `link-to`
+
+    examples:
+
+      ```hbs
+      <button {{action (transition to='posts.edit' postId=post.id)}} />
```
jrjohnson: This is an accessibility red flag to me, links should be links. While the `transition` helper probably has value in other places I'd like to see an example that doesn't just turn a button into a styled link.

NullVoxPopuli:
Superseded by (at the time of writing) to-be-implemented functionality described by https://github.com/emberjs/rfcs/blob/master/text/0391-router-helpers.md#event-dispatcher which would be much more ergonomic

PR #340, opened 2018-06-19
Deprecate Ember.merge

rwwagner90:
[Rendered](https://github.com/rwwagner90/rfcs/blob/deprecate-ember-merge/text/0340-deprecate-ember-merge.md)

### DIFF
--- a/text/0340-deprecate-ember-merge.md
+++ b/text/0340-deprecate-ember-merge.md
@@ -1,5 +1,5 @@
 - Start Date: 2018-06-19
-- RFC PR: (leave this empty)
+- RFC PR: https://github.com/emberjs/rfcs/pull/340
 - Ember Issue: (leave this empty)
 
 # Deprecate Ember.merge in favor of Ember.assign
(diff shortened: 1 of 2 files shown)


0xadada:
I'm 👍 because it bring the name more inline with vanilla js

text/0340-deprecate-ember-merge.md
```diff
@@ -0,0 +1,81 @@
...
+
+The only drawback, that I can think of, is people would need to convert `Ember.merge` to 
+`Ember.assign`, but this would be a very easy change and could easily be done via codemod.
+
+## Alternatives
+
+The impact of not doing this, is we continue to have two functions that do basically the same thing,
+which we need to maintain. 
+
+Another alternative, could be to remove both `Ember.merge` and `Ember.assign`, in favor of `Object.assign`
```
les2: are there any benefits to using the ember assign over ecma script version? if so, it would be helpful to briefly speak to the benefits here.

rwjblue: Object.assign is only available on some platforms (e.g. not on IE11). Ember.assign uses native Object.assign when present and falls back to a simple (not fully spec compliant) polyfill.

rwwagner90:
@rwjblue what do you need from me here to help this along?

tomdale:
We discussed this at the core team meeting today and agree that this RFC is ready for Final Comment Period.

rwjblue:
Merging...

PR #343, opened 2018-06-26
Ember Data Deprecate Transforms

Alonski:
[Rendered](https://github.com/Alonski/rfcs/blob/ember-data-deprecate-transforms/text/0000-ember-data-deprecate-transforms.md)

### DIFF
--- a/text/0000-ember-data-deprecate-transforms.md
+++ b/text/0000-ember-data-deprecate-transforms.md
@@ -1,0 +1,66 @@
+- Start Date: 2018-06-24
+- RFC PR: (leave this empty)
+- Ember Issue: (leave this empty)
+
+# (Ember Data Deprecate Transforms)
+
+## Summary
+
+> This RFC suggests deprecating Ember Data [transforms](https://www.emberjs.com/api/ember-data/release/classes/DS.Transform).
+
+Example:
+
+```javascript
+// app/models/person.js
+import DS from 'ember-data';
+
+export default DS.Model.extend({
+  name: DS.attr('string'), //<--
+  age: DS.attr('number'), //<--
+  admin: DS.attr('boolean'), //<--
+  birthday: DS.attr('date') //<--
+});
+```
+
+## Motivation
+
+> Ember Data `transforms` have, for some time, been the cause of confusion for Ember Data users. Anything that can be done with `transforms` can also be done using Serializers.
+>
+> Part of the confusion was aided by the [no-empty-attrs](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/no-empty-attrs.md) ESLint Plugin Ember Rule. This rule was included as a recommended rule at one point in time in the plugin.
+>
+> `Transforms` are not *types*. They allow serializing and deserializing model `attrs`. However their use often causes people to assume that they are *types*.
+>
+> Another issue, `transforms` can have a bad effect on performance when they are used incorrectly, or at all.
+>
+> tl;dr `Transforms` are confusing and don't provide any added value that can't be achieved using serializers.
+
+## Detailed design
+
+> `Transforms` will be deprecated and any use of `transforms` will cause a deprecation warning, notifying the user that this method is deprecated and providing a link to an explanation of how to achieve the same outcome with serializers.
+>
+> An Ember Addon can be built to help any current users of `transforms` migrate to not using them over time.
+
+## How we teach this
+
+> We need to remove mentions of `transforms` in the Ember Guides and API:
+> [Defining Models](https://guides.emberjs.com/release/models/defining-models/#toc_transforms),
+> [Customizing Serializers](https://guides.emberjs.com/release/models/customizing-serializers/#toc_creating-custom-transformations),
+> [API DS.Transform](https://www.emberjs.com/api/ember-data/release/classes/DS.Transform)
+>
+> Instead of these we will need to teach how to do the same things that were possible with `transforms` but with serializers.
+>
+> Once we remove the use of `transforms` from the guides and instead explain how to use serializers to achieve the same result, new users will learn to use serializers and not `transforms`.
+
+## Drawbacks
+
+> Currently `transforms` are being used and taught and deprecating them will remove a long used feature of Ember Data.
+>
+> We will need to write up more information on serializers to cover for the lack of `transforms`. However this isn't necessarily a bad thing.
+
+## Alternatives
+
+> We could not do this and just leave `transforms` as part of Ember Data. We could try to educate users about the actual uses of `transforms` and remove any notions of using them as `attr` *types*.
+
+## Unresolved questions
+
+> ?


mydea:
We do use transforms quite a lot in our apps, and I don't really see how it would be easy to move this to serializers. Sure, it would be _possible_, but would be much more verbose. Some examples:

Our server can sometimes return null for array fields, so we have an `array` transform that ensures we end up with an empty array instead. We use this in multiple different models, and would have to add serializers for all of them.

Then, we also have some common objects, which the server sends as inlined JSON objects, and we convert to actual class instances in transforms (and back to JSON when serializing). Again, this is used many times, and could theoretically be done in a serializer, but would be much more verbose. Currently, we only have special serializers for really custom, one-off functionality, and all generic functionality that we need multiple times in different models can be handled in transforms. Removing transforms and requiring to do all of this in serializers seems very un-DRY to me. It would also mean that we'd basically need a custom serializer for each and every of our models (which we currently don't).

Maybe I'm missing something here - but what would the actual migration strategy be for a generic transform that's used in multiple models?

knownasilya:
I would change the syntax instead, keeping string, number, boolean, and date (basically making them types) and making custom transform syntax as:

```js
dateUpdated: attr('date', {
  transform: 'moment-date'
})
// just a suggested API
```

buschtoens:
@knownasilya What about non-standard types (i.e. custom classes) and arrays?

runspired:
@mydea 

> I don't really see how it would be easy to move this to serializers

Basically the issue here is that one bad pattern (per-type serializers) is endorsing another bad pattern as a compromise (transforms). That said, another way of looking at this is that you've already done exactly the thing you say you fear, just on the model instead of in the serializer where it belongs.

> Our server can sometimes return null for array fields, so we have an array transform that ensures we end up with an empty array instead. We use this in multiple different models, and would have to add serializers for all of them.

You could do this once in the `application` serializer.

> Then, we also have some common objects, which the server sends as inlined JSON objects, and we convert to actual class instances in transforms (and back to JSON when serializing). Again, this is used many times, and could theoretically be done in a serializer, but would be much more verbose.

It would likely be less verbose if done in the application serializer, especially if each class implements `toJSON`. That said, pushing custom classes as values on ember-data models is highly bug-prone and not something we have in the past officially supported. In addition to "dirty-tracking" not working for them, you would have to be careful to never mutate the reference.  It is however an ability that `RecordData` unlocks.

> Removing transforms and requiring to do all of this in serializers seems very un-DRY to me.

Transforms are anything but DRY, as they require you to do extensive additional work to be able to ensure mutations conform to a hidden and unofficial "type" that they lead you to believe exists, and must be repeated on every `attr`.

> It would also mean that we'd basically need a custom serializer for each and every of our models (which we currently don't).

A single application serializer will do the trick. The serializer is always given the information of the model being serialized / deserialized.

> what would the actual migration strategy be for a generic transform that's used in multiple models?

To use your array example.

```js

const EMPTY_ARR_ATTRS = {
  'dashboard-user': ['settings', 'friends']
};

function normalizedEmptyArrAttrs(modelName, srcHash) {
    let attrsToNormalize = EMPTY_ARR_ATTRS[modelName];
    let attributes = {};

   if (attrsToNormalize !== undefined) {
     attrsToNormalize.forEach(attr => {
       attributes[attr] = srcHash[attr] === null ? [] : srcHash[attr];
     });
   }

  return attributes;
}

class Serializer {
  extractAttributes(modelClass, resourceHash) {
     let attributes = {};
     let { modelName } = modelClass;
     let srcAttributes = resourceHash.attributes;

     // ... do other things as necessary ...

    Object.assign(attributes, normalizeEmptyArrAttrs(modelName, srcAttributes));

     // ... do other things as necessary ...
  }
}
```

runspired:
@Alonski thanks for putting this together to get the conversation started!

We discussed this recently on the team meeting and it seems likely that instead of deprecating transforms we just leave them as a feature of the current model layer (yes, this implies a different model layer might exist, this was something enabled by the `RecordData` RFC and `ember-m3` is one such alternative layer).

We will be splitting ember-data into multiple packages soon (RFCs coming), and while deprecating transforms would allow us to keep `serializers` and `DS.Model` as separate packages it wouldn't be too bad to keep them as a single unit for folks to replace with something that better works for them.

JCBarry:
We use transforms a fair bit as well.  Testing as it pertains to a transform is really nice, and does a nice job performing a "separation of duties" task.  I always feel like the serializers end up with a LOT of code (this may just be because I work with APIs that don't conform to JSONAPI), but when that happens it becomes cumbersome to test/maintain.  So, I'm not big advocate of letting serializers bear the brunt of this workload, especially considering there is already a reasonable pattern in place.

Reading the RFC, it seems the main issue is the confusion regarding types and I believe this confusion stems mostly from the design of the `DS.attr()` function.  In most cases, the argument passed to that function (ex. `string`, `number`, `date`, etc) could all be reasonably perceived as a "type".  The documentation for `DS.attr()` also references the first param as a "type" and by making a transform, it feels like you're making another "type" because of the way you end up using the transform: 

``` javascript
  name: DS.attr('string'), // seems like a type
  age: DS.attr('number'), // seems like a type
  profile: DS.attr('markdown'), // custom transform, is it "markdown" a type?  
});
```
Perhaps as an alternative, transforms become options with a type of "object".  So in practice, it could work along these lines:
``` javascript
  name: DS.attr('string'), // seems like a type
  age: DS.attr('number'), // seems like a type
  profile: DS.attr('object', { 
    transform: 'markdown'
  }), // Clearly and object, but uses a transform.
});
```

Again, most of my apprehension here is to avoid shoving more responsibilities onto the serializers. However, I can appreciate the delicate nature of transforms and their impact on performance if used incorrectly.

luketheobscure:
Effective use of transforms has provided a great developer experience in our applications. 

We use a custom transform when dealing with units on a model that provides access to things like conversion rates and translated names. All a developer has to do is declare `fooUnit: attr('unit')`. Forcing the developer to do that work in the serializer seems sub-optimal, especially since most models will inherit from the application serializer.

christophersansone:
Long time Ember user here.  Transforms are one of my favorite parts of Ember Data and one of the main reasons I haven't moved to something else.  Key benefits:

 * It automatically converts the data to the type of value that you define.  That's awesome, right?
 * The `attr()` definitions help to annotate directly on the model which properties are server-based and what type of value they hold.
 * It minimizes the need to define custom serializers for every model type.  I use JSON:API and have rarely had to declare a custom serializer.
 * It allows the developer to write a converter once across all serializers.

One of my most often used custom transforms is `date-only`, when the property is a date without a time.  The server sends a date in the format of `YYYY-MM-DD` and expects the same format in return.  It becomes a Javascript Date object and I can safely pass it into `moment` or any date picker and just know that, not matter how mangled the time may get along the way, only the date will be sent to the server.

I do believe there are strong reasons to keep transforms.  If transforms did not exist, we could solve the same problem today through custom serializers, though the concept of an isolated, reusable attribute transform seems like a good abstraction.

For me, it would help a lot to see what the proposed "new way" would be to replace a transform. Usually the RFCs have an "old way" vs. "proposed new way".  Any chance we can have one here?

runspired:
Closing as it seems folks would prefer to leave transforms on the current model layer and as mentioned above we feel it would be less disruptive to enable folks to drop that model layer all at once vs in pieces.

I would encourage these folks to investigate the powers of serializers more thoroughly, as these sorts of transformations will nearly always be more clear when done there: this includes the case where one only uses one or a small handful of serializers (such as one per API).

knownasilya:
@runspired maybe a blog post or stream would help guide people in how to do this properly, in a testable way?

sandstrom:
I have three main comments:

1. I agree with all of what @JCBarry said above(https://github.com/emberjs/rfcs/pull/343#issuecomment-400790664). Basically that the blur between attribute types and transforms could definitely be improved, but I don't think we should remove transforms. 

2. I also think that transforms as concept (if separate from the attribute type), could be moved to the serializer layer, in other words, the transforms that should be applied to different model attributes should be specified in the serializer, not the model. 

3. But it's still very useful to keep the transform concept: shared logic for transforming scalar values and data structures. I don't think they should be removed.

gustaff-weldon:
We breathe transforms. 
We have special ones that convert slugs like `male`, `en-us` to objects with more data, or date strings to Moment objects for easier manipulation. 
Thanks to transforms being very simple `serialize/deserialize` constructs they are highly reusable. We reuse them across many models eg. `attr('gender')`, no serialiser logic needed.
This RFC does not seem to propose any DRY alternative - eg. maybe moving transforms to serializers' `attrs` definition instead?

mehulkar:
> Transforms are not types.

Maybe a silly question, but can someone explain why this is? If a transform can guarantee that a property will be a string or number or someCustomThing, why is it wrong to think of it as a type?

runspired:
@mehulkar transforms have zero guarantees. They are only used by the serializer, not the model (so calls to `set` simply clobber them), their state isn't tracked, and almost any serializer customization will result in them not being used at all.

PR #351, opened 2018-08-01
Component Manager Bounds

chadhietala:
[Rendered](https://github.com/emberjs/rfcs/blob/4f541c350cd4f366eed66fa80ef320bb38656b20/text/0000-Component-Manager-Bounds.md)

### DIFF
--- a/text/0000-Component-Manager-Bounds.md
+++ b/text/0000-Component-Manager-Bounds.md
@@ -17,9 +17,9 @@ Today, the custom components are not given direct access to the DOM structure; i
 
 In [RFC#2013](https://github.com/emberjs/rfcs/blob/master/text/0213-custom-components.md#detailed-design) we outlined the component lifecycle and how a custom component is associated with a manager. This design builds off off of that existing work and concepts.
 
-### `elementHook` capability
+### `bounds` capability
 
-The primary way of introducing new functionality to the component manager is through capabilities. To allow custom component managers to announce that they support element related hooks, we will introduce a `elementHook` flag to the capabilities. In doing so, bounds will be dispatched to a `didRenderLayout` method on the manager and `willDestroyLayout` will be called during destruction.
+The primary way of introducing new functionality to the component manager is through capabilities. To allow custom component managers to announce that they support element related hooks, we will introduce a `bounds` flag to the capabilities. In doing so, bounds will be dispatched to a `didCreateBounds` method on the manager and `willDestroyBounds` will be called during destruction.
 
 ```js
 import { capabilities } from '@ember/component';
@@ -27,15 +27,15 @@ import EmberObject from '@ember/object';
 
 export default EmberObject.extend({
   capabilities: capabilities('3.5', {
-    elementHook: true
+    bounds: true
   }),
 
   // ...
 });
 ```
 
-### `didRenderLayout` hook
-Once the rendering engine has constructed the bounds for the component the `didRenderLayout` hook will be called with the `component` and a `Bounds` instance.
+### `didCreateBounds` hook
+Once the rendering engine has constructed the bounds for the component the `didCreateBounds` hook will be called with the `ComponentStateBucket` and a `Bounds` instance.
 
 
 ```js
@@ -47,9 +47,8 @@ export default EmberObject.extend({
     elementHook: true
   }),
 
-  didRenderLayout(component, bounds) {
+  didCreateBounds({ component }, bounds) {
     component.bounds = bounds;
-    component.didInsertElement();
   }
   // ...
 });
@@ -57,11 +56,11 @@ export default EmberObject.extend({
 
 #### Timing Semantics
 
-`didRenderLayout` has the following timing semantics:
+`didCreateBounds` has the following timing semantics:
 
 **Always**
 - called **before** `didCreateComponent`
-- called **after** the children's `didRenderLayout` hook is called
+- called **after** the children's `didCreateBounds` hook is called
 - called **after** DOM insertion
 
 **May or May Not**
@@ -71,9 +70,9 @@ export default EmberObject.extend({
 
 In other words, components **should not** rely on the timing of DOM insertion outside of the component's own bounds.
 
-### `willDestroyLayout` hook
+### `willDestroyBounds` hook
 
-To balance `didRenderLayout` we will expose a hook called `willDestroyLayout` that is invoked during destruction. It will receive the `component` instance.
+To balance `didRenderLayout` we will expose a hook called `willDestroyBounds` that is invoked during destruction. It will receive the `ComponentStateBucket` instance.
 
 ```js
 import { capabilities } from '@ember/component';
@@ -84,12 +83,11 @@ export default EmberObject.extend({
     elementHook: true
   }),
 
-  didRenderLayout(component, bounds) {
+  didRenderLayout({ component }, bounds) {
     component.bounds = bounds;
-    component.didInsertElement();
   },
 
-  willDestroyLayout(component) {
+  willDestroyBounds({ component }) {
     component.willDestroyElement();
   }
   // ...
@@ -98,11 +96,11 @@ export default EmberObject.extend({
 
 #### Timing Semantics
 
-`willDestroyLayout` has the following timing semantics:
+`willDestroyBounds` has the following timing semantics:
 
 **Always**
 - called **before** `destroyComponent`
-- called **after** the children's `willDestroyLayout` hook is called
+- called **after** the children's `willDestroyBounds` hook is called
 
 **May or May Not**
 - be called in the same tick as DOM removal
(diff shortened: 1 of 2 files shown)


text/0000-Component-Manager-Bounds.md
```diff
@@ -0,0 +1,177 @@
...
+## Motivation
+Historically, classic Ember components have had a built in escape hatch for doing DOM manipulation via `didInsertElement`. This allowed for components to encapsulate some 3rd party JavaScript library or do some fine grain DOM manipulation. We feel that is ability to drop down to a lower level abstraction makes components powerful.
+
+Today, the custom components are not given direct access to the DOM structure; internally known as `Bounds`. This limitation means that custom components cannot easily encapsulate functionality in the same way that classic Ember components can. Expanding the capabilities of custom components to expose the `Bounds` is in order to increase the usefulness of the abstraction.
+
+## Detailed design
+
+In [RFC#2013](https://github.com/emberjs/rfcs/blob/master/text/0213-custom-components.md#detailed-design) we outlined the component lifecycle and how a custom component is associated with a manager. This design builds off off of that existing work and concepts.
+
+### `elementHook` capability
```
rwjblue: `elementHook` doesn't really seem to fit with the other capability names. Suggestions:

* `elementAccess`
* `domAccess`
* `domAccessCallbacks`
* `domCallbacks`
* `elementCallbacks`
* `bounds` (I like this one best 😄)

Thoughts?

text/0000-Component-Manager-Bounds.md
```diff
@@ -0,0 +1,177 @@
...
+export default EmberObject.extend({
+  capabilities: capabilities('3.5', {
+    elementHook: true
+  }),
+
+  // ...
+});
+```
+
+### `didRenderLayout` hook
```
rwjblue: I know that this is the name internally in the glimmer-vm, but I don't think its particularly relevant as a name.

Possibly: `didCreateBounds`?

text/0000-Component-Manager-Bounds.md
```diff
@@ -0,0 +1,177 @@
...
+import EmberObject from '@ember/object';
+
+export default EmberObject.extend({
+  capabilities: capabilities('3.5', {
+    elementHook: true
+  }),
+
+  didRenderLayout(component, bounds) {
+    component.bounds = bounds;
+    component.didInsertElement();
```
rwjblue: Based on the prior RFC `didInsertElement` is more appropriately called in `didCreateComponent` (which was an optional `asyncLifecycleCallbacks` capability), I don't think we should recommend changing that...

text/0000-Component-Manager-Bounds.md
rwjblue: Actually, can we avoid creating a new hook here and instead add `bounds` as a second argument to the already existing `didCreateComponent` hook?  `didCreateComponent` is part of the `asyncLifecycleCallbacks` capability, and per the prior RFC is where `didInsertElement` should go. 

The name seems much better, and the intent seems **very much** in line with what we are doing here...

text/0000-Component-Manager-Bounds.md
```diff
@@ -0,0 +1,177 @@
...
+  capabilities: capabilities('3.5', {
+    elementHook: true
+  }),
+
+  // ...
+});
+```
+
+### `didRenderLayout` hook
+Once the rendering engine has constructed the bounds for the component the `didRenderLayout` hook will be called with the `component` and a `Bounds` instance.
```
rwjblue: This phrasing is slightly off, the hook does not receive "the component" it receives the opaque result of calling `createComponent` (aka the "state bucket"). Different component managers _may_ return the instance itself but it may also return some other POJO/state bucket.

text/0000-Component-Manager-Bounds.md
```diff
@@ -0,0 +1,177 @@
...
+- called **after** DOM insertion
+
+**May or May Not**
+- be called in the same tick as DOM insertion
+- be called in the same tick as `didCreateComponent`
+- have the the parent's children fully initialized in DOM
+
+In other words, components **should not** rely on the timing of DOM insertion outside of the component's own bounds.
+
+### `willDestroyLayout` hook
```
rwjblue: Assuming we can keep the bounds as a second arg to `createComponent` (one of my suggestions above), this can become `willDestroyComponent`...

text/0000-Component-Manager-Bounds.md
```diff
@@ -0,0 +1,177 @@
...
+
+To balance `didRenderLayout` we will expose a hook called `willDestroyLayout` that is invoked during destruction. It will receive the `component` instance.
```
rwjblue: it doesn't get the component instance, it gets the "return value from createComponent" (may be a state bucket, may be a component instance)

text/0000-Component-Manager-Bounds.md
```diff
@@ -0,0 +1,177 @@
...
+### `Bounds` instance
+
+The `Bounds` instance that is passed to `didRenderLayout` has the following interface.
+
+```ts
+
+type FirstNode = Node;
+type LastNode = Node;
+
+interface Bounds {
```
rwjblue: `Bounds` doesn't get `.parentElement`?

The interface in glimmer-vm does have `parentElement` property:

https://github.com/glimmerjs/glimmer-vm/blob/e30b3d64232a0972365c8cbb1dd8e236a0f33dcf/packages/%40glimmer/runtime/lib/bounds.ts#L4-L9

chadhietala: Due to the lack of timing guarantees of construction of the parent do think this actually useful?

text/0000-Component-Manager-Bounds.md
chancancode: I don't mind bike shedding about the name, but I don't think we want to bundle it with `didCreateComponent`. There may be situations that you need one but not the other, and we want to keep things as granular as possible. Plus, bundling the hooks forces their timing to be the same, which we probably don't want (and they are in fact different today).

text/0000-Component-Manager-Bounds.md
chancancode: In general, I think we need to recommend _against_ calling any kind of user hooks here.

text/0000-Component-Manager-Bounds.md
chancancode: I think we need to provide access to the parent. What we cannot guarantee is that the parent's children is fully constructed. (And possibly that the grandparent may not be construed yet.)

By the way, I think you meant to say:

```ts
interface Bounds {
  firstNode: Node | Bounds;
  lastNode: Node | Bounds;
 }
```

text/0000-Component-Manager-Bounds.md
rwjblue: Sounds good, I agree.  Still would prefer another name than `didRenderLayout` it _really_ doesn't feel cohesive with the other component manager method names.

lennyburdette:
Has anyone proposed a react-style “refs” api? Instead of relying on the framework to create special properties, we could have a declarative api that allows decorating the template with special attributes that create JavaScript-accessible node references. 

Something like `<div ref="a" />` to be accessed as `this.refs.a` in lifecycle hooks.

I’m just curious. Both APIs could exist simultaneously.

JCBarry:
Not frequently, but I've had the need for this in more complex components and the workarounds aren't pretty or easy to test.  👍

chancancode:
We discovered a problem with this design in the core team meeting last week. The intent of this is to offer functionality akin to classic component's `this.element`, `didInsertElement` and `willDestroyElement`. The common use case is to attach and cleanup jQuery or native event listeners.

On the surface, it seems like this gets the job done. However, consider the following component template:

```hbs
{{#if this.isSelected}}
  <strong>selected</strong>
{{else}}
  <em>select me</em>
{{/if}}
```

Ignoring the whitespace nodes, the bounds object will point to the `strong` or `em` element depending on the component's `this.isSelected` state, which could change back and forth multiple times during the lifecycle of the component. Therefore, if you attached an event listener on the element at initial render time, you may not get a chance to detach it before the element goes away, thus leaking memory. Dereferencing the nodes referenced by the bounds and then saving them off or closing over them with a closure has the same issue.

Of course, in a sense this is no different than attaching event listeners directly to descendants of `this.element`, since they are also subject to change without notice. However, as there is a stable root element, the event listeners are almost always attached there in practice.

There is still a class of use cases that is somewhat compelling and does not have the problem. For example, a component may use the bounds to "autofocus" the first input element. This only involve accessing the bounds' content synchronously and does not require any cleanup. In general, these kind of "snapshotting" use cases are fine. However, they may call for more hooks, like one that runs after the bounds' content has (potentially) been updated so they could reapply the logic.

So we can adapt this proposal for that kind of "snapshotting" use cases, perhaps adding the update hook and possibly removing the teardown hook (cause it is not needed, and may be misleading). We can then explore alternatives for the original use case, perhaps a modifier/ref-based design or an additional capability that ensures (at runtime) that the component has a stable root element. However, my worry is that since the original use case is the most common and compelling one, having this API around without a solution for that use case will mislead our users and causing them to misuse this API.

chancancode:
Relevant: https://github.com/emberjs/ember.js/pull/12500

chadhietala:
Closing in favor of #353

PR #353, opened 2018-08-16
Modifiers

chadhietala:
[Rendered](https://github.com/emberjs/rfcs/blob/new-modifiers/text/0000-modifiers.md)

### DIFF
--- a/text/0000-modifiers.md
+++ b/text/0000-modifiers.md
@@ -74,11 +74,34 @@ export default class extends Modifier {}
 
 During rendering and teardown of a target element, any attached element modifiers will execute a series of hooks. These hooks are:
 
+* `willInsertElement`
 * `didInsertElement`
 * `didUpdate`
 * `willDestroyElement`
 
 It is important to note that in server-side rendering environments none of the lifecycle events are called.
+
+#### `willInsertElement` semantics
+`willInsertElement` is called when the element has been constructed, but has yet to be inserted into the DOM. Any attributes that where defined within the template will be present. This hook is specifically to be used for programmatically attatching additional attributes or properties that must be there before the element is in-DOM. Some examples of this are `style`, `selected`, `value`, etc. This hook is called once.
+
+This hook has the following timing semantics:
+
+**Always**
+- called **after** all child modifiers `willInsertElement` hook are called
+- called **after** template defined attributes are attached
+- called **before** DOM insertion
+
+Below is an example of how this hook could be used:
+
+```js
+import Modifier from '@ember/modifier';
+
+export default class extends Modifier {
+  willInsertElement() {
+    this.element.style = 'background-color: green;'
+  }
+}
+```
 
 #### `didInsertElement` semantics
 
@@ -101,6 +124,11 @@ import Modifier from '@ember/modifier';
 
 export default class extends Modifier {
   this.listenerOptions = undefined;
+
+  willInsertElement() {
+    this.element.style = 'background-color: green;'
+  }
+
   didInsertElement([ eventType, callback ], eventOptions) {
     this.listenerOptions = [
       eventType,
@@ -131,6 +159,11 @@ import Modifier from '@ember/modifier';
 
 export default class extends Modifier {
   this.listenerOptions = undefined;
+
+  willInsertElement() {
+    this.element.style = 'background-color: green;'
+  }
+
   didInsertElement([ eventType, callback ], eventOptions) {
     this.listenerOptions = [
       eventType,
@@ -159,6 +192,11 @@ import Modifier from '@ember/modifier';
 
 export default class extends Modifier {
   this.listenerOptions = undefined;
+
+  willInsertElement() {
+    this.element.style = 'background-color: green;'
+  }
+
   didInsertElement([ eventType, callback ], eventOptions) {
     this.listenerOptions = [
       eventType,
--- a/text/0000-modifiers.md
+++ b/text/0000-modifiers.md
@@ -11,7 +11,7 @@ This RFC introduces the concept of user defined element modifiers. Unlike a comp
 Below is an example of the element modifier syntax:
 
 ```hbs
-<button {{add-event-listener 'click' (action 'save')}}>Save</button>
+<button {{effect 'fade-in'}}>Save</button>
 ```
 
 This RFC supercedes the [original element modifiers RFC](https://github.com/emberjs/rfcs/pull/112) and is intended to replace the [`this.bounds` RFC](https://github.com/emberjs/rfcs/pull/351).
@@ -258,6 +258,16 @@ Element modifers are intended to be directly installed onto HTML elements and ca
 ```
 
 This also means that they have no relationship to `...attributes`.
+
+### Relationship With Other Modifiers
+
+An element may have multiple modifiers installed on it.
+
+```hbs
+<button {{effect 'fade-in'}} {{effect 'spin'}}>Wat!</button>
+```
+
+The modifiers are evaluated from left to right. This means that the `fade-in` effect's hooks will run before `spin` effect's.
 
 ## Examples
 
--- a/text/0000-modifiers.md
+++ b/text/0000-modifiers.md
@@ -18,11 +18,11 @@ This RFC supercedes the [original element modifiers RFC](https://github.com/embe
 
 ## Motivation
 
-Classic component instances have a `this.element` property which provides you a single DOM node as defined by `tagName`. The children of this node will be the DOM representation of what you wrote in your template. Templates are typically referred to having `innerHTML` semantics in classic components since there is a single wrapping element that is the parent of the template. These semantics allow for components to encapsulate some 3rd party JavaScript library or do some fine grain DOM manipulation.
+`Ember.Component` instances have a `this.element` property which provides you a single DOM node as defined by `tagName`. The children of this node will be the DOM representation of what you wrote in your template. Templates are typically referred to having `innerHTML` semantics in `Ember.Component`s since there is a single wrapping element that is the parent of the template. These semantics allow for components to encapsulate some 3rd party JavaScript library or do some fine grain DOM manipulation.
 
-Glimmer components have `outerHTML` semantics, meaning what you see in the template is what you get in the DOM, there is no `tagName` that wraps the template. While this drastically simplifies the API for creating new components it makes reliable access to the a component's DOM structure very difficult to do. As [pointed out](https://github.com/emberjs/rfcs/pull/351#issuecomment-412123046) in the [Bounds RFC](https://github.com/emberjs/rfcs/pull/351) the stability of the nodes in the `Bounds` object creates way too many footguns. So a formalized way for accessing DOM in the Glimmer component world is still needed.
+Custom components have `outerHTML` semantics, meaning what you see in the template is what you get in the DOM, there is no `tagName` that wraps the template. While this drastically simplifies the API for creating new components it makes reliable access to the a component's DOM structure very difficult to do. As [pointed out](https://github.com/emberjs/rfcs/pull/351#issuecomment-412123046) in the [Bounds RFC](https://github.com/emberjs/rfcs/pull/351) the stability of the nodes in the `Bounds` object creates way too many footguns. So a formalized way for accessing DOM in the custom component world is still needed.
 
-Element modifiers allow for stable access of the DOM node they are installed on. This allows for programatic access to DOM in Glimmer templates and also offers a more targeted construct for cases where classic components were being used. The introduction of this API will likely result in the proliferation of one or several popular addons for managing element event listeners, style and animation.
+Element modifiers allow for stable access of the DOM node they are installed on. This allows for programatic access to DOM in custom component templates and also offers a more targeted construct for cases where classic components were being used. The introduction of this API will likely result in the proliferation of one or several popular addons for managing element event listeners, style and animation.
 
 ## Detailed design
 
@@ -36,7 +36,7 @@ An element modifier is invoked in "element space". This is the space between `<`
 <b {{crum bing='whoop'}} zip="bango">Hm...</b>
 ```
 
-Element modifiers may be invoked with params or hash arguments.
+Element modifiers may be invoked with positional and/or named arguments.
 
 ### Definition and lookup
 
@@ -54,7 +54,7 @@ MU paths:
   src/ui/routes/posts/-components/whipperwill/modifier.js
 ```
 
-In Module Unification, modifiers live within the generalized collection type "components" [as specified](https://github.com/dgeb/rfcs/blob/module-unification/text/0000-module-unification.md#components). Modifiers, like component and helpers, are eligible for local lookup. For example:
+In Module Unification, modifiers live within the generalized collection type "components" [as specified](https://github.com/emberjs/rfcs/blob/master/text/0143-module-unification.md#components). Modifiers, like component and helpers, are eligible for local lookup. For example:
 
 ```
 MU paths:
@@ -150,7 +150,7 @@ export default class extends Modifier {
 
 #### `willDestroyElement` semantics
 
-`willDestroyElement` is called during the destruction of a template. It receives no arguments.
+`willDestroyElement` is called during the destruction of the element. It receives no arguments.
 
 Below is an example of how this hook could be used:
 
@@ -184,9 +184,24 @@ This hook has the following timing semantics:
 
 **Always**
 - called **after** all children modifier's `willDestroyElement` hook is called
+- called **after** all children component's `willDestroyComponent` hook is called.
 
  **May or May Not**
 - be called in the same tick as DOM removal
+
+### Relationship With Components
+
+Element modifers are intended to be directly installed onto HTML elements and can not be invoked on a component invocation. They also cannot be passed as arguments to a component. The examples below will result in a compile error.
+
+```hbs
+<FooBar {{draggable x=x y=y}} />
+```
+
+```hbs
+<FooBar (draggable x=x y=y) />
+```
+
+This also means that they have no relationship to `...attributes`.
 
 ## Examples
 
@@ -279,7 +294,7 @@ In terms of guides, I believe we should add a section to the "Templates" section
 
 The drawbacks of adding element modifiers largely deal with explaining when to use a classic component for encapsulating some DOM manipulation versus when to use an element modifier. That being said it is likely that the recommendation would be to use element modifiers if you are just manually modifying the DOM.
 
-This API also doesn't attempt to create a corollary of `this.element` for Glimmer Components and instead offers a different API for altering DOM nodes directly. This expands the surface area of Ember's API.
+This API also doesn't attempt to create a corollary of `this.element` for custom components and instead offers a different API for altering DOM nodes directly. This expands the surface area of Ember's API.
 
 ## Alternatives
 
--- a/text/0000-modifiers.md
+++ b/text/0000-modifiers.md
@@ -70,6 +70,24 @@ import Modifier from '@ember/modifier';
 export default class extends Modifier {}
 ```
 
+### Resolution
+
+As mentioned in the [invocation section](#invocation), an modifier must be invoked in element space. If an element modifier is invoked outside of element space, it will go through the existing resolution logic.
+
+#### Invocation With Parameters
+
+- *If* closure component, then unroll arguments and invoke
+- *Else If* helper, then invoke
+- *Else* throw "Helper not found" error
+
+#### Invocation Without Parameters
+
+- *If* clusure component, then unroll arguments and invoke
+- *Else If* helper, then invoke
+- *Else If* local, look up value and place in DOM
+- *Else If* property on context, look up value and place in DOM
+- *Else* resolve to `undefined`
+
 ### Lifecycle Hooks
 
 During rendering and teardown of a target element, any attached element modifiers will execute a series of hooks. These hooks are:
--- a/text/0000-modifiers.md
+++ b/text/0000-modifiers.md
@@ -1,0 +1,242 @@
+- Start Date: (fill me in with today's date, YYYY-MM-DD)
+- RFC PR: (leave this empty)
+- Ember Issue: (leave this empty)
+
+# Element Modifiers
+
+## Summary
+
+This RFC introduces the concept of user defined element modfiers. Unlike a component, there is no template/layout for an element modifier. Unlike a helper, an element modifier does not return a value.
+
+Below is an example of the element modifier syntax:
+
+```hbs
+<button {{add-event-listener 'click' (action 'save')}}>Save</button>
+```
+
+This RFC supercedes the [original element modifiers RFC](https://github.com/emberjs/rfcs/pull/112) and is intended to replace the [`this.bounds` RFC](https://github.com/emberjs/rfcs/pull/351).
+
+## Motivation
+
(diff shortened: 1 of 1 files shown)


rtablada:
@chadhietala what does this mean for overloaded `helpers` vs `modifiers`?

Like `action` is both?

Is that something we want to allow as public API for other users?

How do we keep that compatible with current helpers and AST transforms like https://github.com/mmun/ember-component-attributes?

mike-north:
Do you have any concerns around making it easy to static-analyze potential misuse of helpers and modifiers? As it stands, we'd have to know more about the implementation of these two things (`my-helper`, `my-modifier`) to detect problems at build-time.

```hbs
<button onClick={{my-modifier "does not emit a value"}} >
```

```hbs
<button {{my-helper "emit a value into nothingness"}}>
```

text/0000-modifiers.md
```diff
@@ -0,0 +1,242 @@
...
+It is important to note that in server-side rendering environments none of the lifecycle events are called.
+
+#### `didInsertElement` semantics
+
+`didInsertElement` is called when the element is in the DOM and it's children are attached. The element is available on `this.element` of the modifier instance. Unlike classic components, the `didInsertElement` hook receives the positional and named params as arguments in the same way helpers do. This hook is only called once.
+
+This hook has the following timing semantics:
+
+**Always**
+- called **after** any children modifiers `didInsertElement` hook are called
```
xg-wang: 'any' => 'all'?

text/0000-modifiers.md
```diff
@@ -0,0 +1,242 @@
...
+  willDestroyElement() {
+    this.element.removeEventListener(...this.listenerOptions);
+  }
+}
+```
+
+This hook has the following timing semantics:
+
+**Always**
+- called **after** any children modifier's `willDestroyElement` hook is called
```
xg-wang: 'any' => 'all'?

chadhietala:
@mike-north this is going to result in a compilation error. In the case of the modifier you would get an error like

`No modifier named “my-helper” could be found`

In the case of the modifier being used outside of element space you will get a helper not found error. 

Effectively the location narrows the type.

text/0000-modifiers.md
```diff
@@ -0,0 +1,242 @@
...
+
+This hook has the following timing semantics:
+
+**Always**
+- called **after** any children modifiers `didInsertElement` hook are called
+- called **after** DOM insertion
+
+ **May or May Not**
+- be called in the same tick as DOM insertion
+- have the the parent's children fully initialized in DOM
```
sandstrom: Overall very positive to this RFC!

Minor thing: I guess "the parent's children" is the same as "the elements siblings"? If that's the case, would the term sibling be better?

text/0000-modifiers.md
```diff
@@ -0,0 +1,242 @@
...
+```
+Classic paths:
+
+  app/element-modifiers/flummux.js
+  app/element-modifiers/whipperwill.js
+
+MU paths:
+
+  src/ui/components/flummux/element-modifier.js
+  src/ui/routes/posts/-components/whipperwill/element-modifier.js
```
dfreeman: It seems like there's a discrepancy between the type name of `element-modifier` and the base package + class names of `@ember/modifier` + `Modifier`. Is there a reason to have the leading `element` for one but not the other?

Relatedly, how does a dashed type name play with [Rule 3](https://github.com/dgeb/rfcs/blob/module-unification/text/0000-module-unification.md#module-type) of the MU resolution rules? What would the identifier for a named `element-modifier` export be?

text/0000-modifiers.md
mike-north: I'd also like clarification. If `children` is sibling elements, can't those belong to sibling components?

gossi:
Some notes that I quickly come up with:

1) Modifiers !== Components. period. So don't put them into `src/ui/components` but e.g. `src/ui/modifiers`. Don't mix. It just happens I have to work on components that were helpers (and placed there), it is messy. Lesson learned, please don't repeat.

2) As the `Component` is basically the umbrella to encapsulate the logic around a component (or to bridge it to a 3rd party library), this is now spread across various modifiers?

3) Will these modifiers be connected to the `Component` (if there is one) or will they work in isolation? If so, the concept of a component managing its template/DOM is kinda eliminated? Or that's how I feel about it at the moment.

4) With `helper`, `modifier` and `Component` it increases complexity. There should be a clear path forward. E.g. `<div {{my-whatever}} ...>` has to be clear where to find it in the code and what it is.

Although I see the benefits of modifiers (e.g. attaching listeners to elements wo/ js code). At the same time I see this critical as well as we will now declaratively set our listeners which is done programmatically in vanilla JS. With native classes Ember is catching up with recent javascript developments modifiers it looks like there is a chance we are splitting apart again? Components mentally always connect to Web Components (and shadow DOM in some way). Ember should more align with these concepts in order to onboard new ember devs.

chadhietala:
@luxferresum as mentioned before there are 2 implementations of `{{action}}` today that has the interfaces aligned, however they behave different. The `{{action}}` helper returns a bound method that is either assigned directly to an element or is closed over and passed. The `{{action}}` modifier gains access to the underlying element and registers `data-` attributes that are used by the event dispatcher.

To draw corollary, JS decorators have different implementations based on where the decorator is used. For instance a [class decorator](https://www.typescriptlang.org/docs/handbook/decorators.html#class-decorators) cannot be used as a [accessor decorator](https://www.typescriptlang.org/docs/handbook/decorators.html#accessor-decorators) unless the implementation allows for both.

@gossi some answers to your questions

1) Module Unification has [already specified](https://github.com/dgeb/rfcs/blob/module-unification/text/0000-module-unification.md#components) that "components" is a generic term that encompasses components, specialized components, helpers, and modifiers.

2) As mentioned in the RFC Glimmer components are effectively "fragments" and thus there is not a single `this.element` that represents the component, so we need a way to give raw assess to the underlying DOM. I wrote [another RFC](https://github.com/emberjs/rfcs/pull/351) about providing components raw access to the underlying DOM and there were [many issues](https://github.com/emberjs/rfcs/pull/351#issuecomment-412123046) with it.

3) Does not foreclose on the idea see the alternatives section this RFC.

4) Position narrows the type.

luxferresum:
I *do* understand how `{{action}}` works. I wanted to point out that currently `{{action}}` is quite hard to teach and there are many blog posts about how it works. So I hope we will eventually get rid of classic actions and don't reintroduce anything like this again with modifiers.

text/0000-modifiers.md
```diff
@@ -0,0 +1,242 @@
...
+
+Glimmer components have `outerHTML` semantics, meaning what you see in the template is what you get in the DOM, there is no `tagName` that wraps the template. While this drastically simplifies the API for creating new components it makes reliable access to the a component's DOM structure very difficult to do. As [pointed out](https://github.com/emberjs/rfcs/pull/351#issuecomment-412123046) in the [Bounds RFC](https://github.com/emberjs/rfcs/pull/351) the stability of the nodes in the `Bounds` object creates way too many footguns. So a formalized way for accessing DOM in the Glimmer component world is still needed.
+
+Element modifiers allow for stable access of the DOM node they are installed on. This allows for programatic assess to DOM in Glimmer templates and also offers a more targeted construct for cases where classic components were being used. The introduction of this API will likely result in the proliferation of one or several popular addons for managing element event listeners, style and animation.
```
pablobm: `s/programatic assess/programmatic access/`?

samselikoff:
This is an awesome RFC, exactly the kind of low-level primitive Ember needs to foster more experimentation. I think the power of abstractions that can be made with these sorts of things are under appreciated.

`<button {{on 'click' (action 'save')}}>Save</button>` might be a more enticing example (on vs. add-event-listener).

I think these modifiers could be used for so many useful things – animation for one. They are more like a horizontal composition/mixin pattern than a hierarchical/tree-like composition pattern (and many problem areas lend themselves to horizontal rather than vertical methods of reuse). Cannot wait to get my hands on these!

sandstrom:
@chadhietala Great examples! 🏅  We've built something similar to the performance marking example, but with components. Using modifiers would be much cleaner!

And although I liked the preceding RFC, this one is easier to understand and more powerful/generic. 👍

text/0000-modifiers.md
```diff
@@ -0,0 +1,323 @@
...
+
+## Drawbacks
+
+The drawbacks of adding element modifiers largely deal with explaining when to use a classic component for encapsulating some DOM manipulation versus when to use an element modifier. That being said it is likely that the recommendation would be to use element modifiers if you are just manually modifying the DOM.
+
+This API also doesn't attempt to create a corollary of `this.element` for Glimmer Components and instead offers a different API for altering DOM nodes directly. This expands the surface area of Ember's API.
+
+## Alternatives
+
+The alternative to this is to create a "ref"-like API that is available in [other client side frameworks](https://reactjs.org/docs/refs-and-the-dom.html). This may look like the following:
```
rwjblue: I think that modifiers (as proposed here) allows a "ref-like api" to be implemented. Slightly modified example:

```hbs
<section>
  <h1 {{ref this "heading"}}>Hello!</h1>
  <p>How are you?</p>
</section>
```

Could easily be implemented as a modifier:

```js
export default class RefModifier extends Modifier {
  didInsertElement([component, name]) {
    component[name](this.element);
  }

  willDestroyElement([component, name]) {
    component[name](null);
  }
}
```

In other words, I don't think "refs" is an alternative to modifiers but instead is a valid use case of them...

gossi:
@chadhietala I did some rereading and get a better understanding. Some of this is now more clear to me.

1) For how I understand this:

- *modifiers:* will be placed inside an element tag, e.g. `<div {{on 'click' (action 'huibuuh')}}>`
- *helpers:* will be placed outside of an element tag, e.g. `{{log 'blah'}}`

If that's correct, please add it somewhere at the top to give an immediate better understanding.

2) Can you please provide two more examples, that show the interaction of modifiers with a controller/component. And a second one (more in direction towards MU) how a component will work with local modifiers and how the file layout will look like (more to give us an impression what we will 
be able to do with Ember Octane :tada:).

3) Although the MU RFC mentions to put everything component related into `src/ui/components` I still think, this is a bad idea. It's good to group them under `src/ui` but should be `src/ui/helpers`, `src/ui/modifiers`, ... maybe there will be a future RFC. Just want to rise awareness for now.

**Questions**

1) With handlebars you rarely interact with the DOM directly (unless applying third party libraries). Will modifiers be the (only?) location where you should do this?

2) That would be a follow up on @rwjblue of passing the `element` in the constructor to apply and manage attributes. I do like this idea. Though contrary to this, this will be an unknown set of attributes managed by a modifier that are non visible in the template (hidden UI). E.g. just use attributes as we would of today: `<input disabled={{this.disabled}}>`. The question to this is: Is there a use-case where a modifier MUST be used instead of just going the regular way we already have.

text/0000-modifiers.md
```diff
@@ -0,0 +1,323 @@
...
+  willDestroyElement() {
+    this.element.removeEventListener(...this.listenerOptions);
+  }
+}
+```
+
+This hook has the following timing semantics:
+
+**Always**
+- called **after** all children modifier's `willDestroyElement` hook is called
```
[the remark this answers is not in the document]
chadhietala: Added more detail

text/0000-modifiers.md
```diff
@@ -70,16 +70,57 @@ import Modifier from '@ember/modifier';
...
+
+#### Invocation With Parameters
+
+- *If* closure component, then unroll arguments and invoke
+- *Else If* helper, then invoke
+- *Else* throw "Helper not found" error
+
+#### Invocation Without Parameters
+
+- *If* clusure component, then unroll arguments and invoke
```
rwjblue: clusure -> closure

text/0000-modifiers.md
```diff
@@ -70,16 +70,57 @@ import Modifier from '@ember/modifier';
...
 
+* `willInsertElement`
 * `didInsertElement`
 * `didUpdate`
 * `willDestroyElement`
 
 It is important to note that in server-side rendering environments none of the lifecycle events are called.
 
+#### `willInsertElement` semantics
+`willInsertElement` is called when the element has been constructed, but has yet to be inserted into the DOM. Any attributes that where defined within the template will be present. This hook is specifically to be used for programmatically attatching additional attributes or properties that must be there before the element is in-DOM. Some examples of this are `style`, `selected`, `value`, etc. This hook is called once.
```
rwjblue: Might need a newline above this to have the header show up properly...

rwjblue: "attributes that where defined" -> "attributes that were defined"

tomdale:
We discussed this RFC at the core team meeting today and there were a few concerns raised that we don't yet have consensus on.

First, there were some concerns about `willInsertElement` and whether it was the right solution to the use cases @rwjblue raised given trouble we've had in the past about timing guarantees and subtle bugs caused by DOM ordering quirks. @wycats can expand on some of the details here.

Second, we discussed whether this API puts the cart before the horse in terms of implementing the high-level feature before the low-level capability. Specifically, implementing the element modifier equivalent of the Component Manager API would allow us to experiment with specific implementation and API details before committing to a high-level API as laid out in this RFC.

There were some additional concerns around what arguments are passed to hooks and how, but that discussion was tabled until we get a resolution on whether we should merge a Modifier Manager RFC first, which would make the concern moot for the time being.

Lastly, one item that we didn't have time to discuss but has been on my mind: I don't love the name "element modifier" from a teachability perspective. I feel like it emphasizes the details of how it's implemented vs. the value they provide to the programming model.

For example, if I write an addon that tracks whenever a user clicks on an element and annotates it with a custom tag, like this:

```hbs
<div {{track-click 'User Profile'}}></div>
```

I don't think it's obvious or important that it's "modifying" the element. To me, it looks like a helper that can be used with an element, rather than producing a value. Maybe "element helper" is a better term?

sandstrom:
How about *element enhancements* or *element augmentations* (or simply augmentations)? I think 'element helper' is easily confused with regular helpers.

topaxi:
Angular calls these things *directives*, I'm not sure if they directly map to the same thing, but from a learning perspective, I think this might help at least developers with [AngularJS](https://docs.angularjs.org/guide/directive)/[Angular](https://angular.io/api/core/Directive) knowledge.

text/0000-modifiers.md
```diff
@@ -0,0 +1,394 @@
...
+# Element Modifiers
+
+## Summary
+
+This RFC introduces the concept of user defined element modifiers. Unlike a component, there is no template/layout for an element modifier. Unlike a helper, an element modifier does not return a value.
+
+Below is an example of the element modifier syntax:
+
+```hbs
+<button {{add-event-listener 'click' (action 'save')}}>Save</button>
```
simonihmig: First, I totally like this RFC! Just this example was a bit confusing for me when first reading this, as it does not show any real value added here, IMHO. To my understanding, this would be equivalent to:
```hbs
<button {{action 'save'}}>Save</button>
```
At least from the user's perspective (ignoring internals like the action modifiers use of `data-ember-action` and the resulting event delegation).

Maybe another example thats show something *not* easily possible today would be nice, like
```hbs
<button {{effect 'fade-in'}}>Save</button>
```

ryanto:
I'm excited for this one... it's going to unlock some cool ideas.

Would it be possible to partially apply element modifiers like we do with the component helper? I'm imagining contextual element modifiers here...

```hbs
{{yield (hash
  my-component=(component "my-component" x=x)
  my-element-modifier=(element-modifier "my-element-modifier" x=x)}}
```

Maybe useful for modifiers that could be considered related, like a drag and drop zone?

```hbs
{{#drag-and-drop-modifiers as |drag|}}
  <div {{drag.draggable}}></div>
  <div {{drag.droppable}}></div>
{{/drag-and-drop-modifiers}}
```

averydev:
> I don't think it's obvious or important that it's "modifying" the element. To me, it looks like a helper that can be used with an element, rather than producing a value. Maybe "element helper" is a better term?

Agreed! Also, when you're just talking about DOM, "Element" is clear and specific, but when I first read this I wasn't clear that `Element` was referring to an `HTML-Element`. That got me thinking perhaps `Node` rather than `Element`?

"Node-Helper"
"Node-Enhancer"  per @sandstrom's  but more closely parallels "Helper" semantic
"Node-Modifier"

knownasilya:
Node makes you think node.js, so might not be a good candidate either.

chadhietala:
#373 Was merged and I intend to create an addon that exposes this API so people can try it out.

gossi:
`sparkles-modifiers`  ✨ 🎇 ❇️ 🌟 ? 😂

buschtoens:
Motivated by https://github.com/emberjs/rfcs/pull/415#issuecomment-458563339 ([`@ember/render-modifiers`](https://github.com/emberjs/ember-render-modifiers)) I've published [`ember-on-modifier`](https://github.com/buschtoens/ember-on-modifier), which implements the exemplary `{{on}}` modifier.

luxferresum:
@buschtoens why would I do `{{on "click" this.foo}}` instead of `onclick={{this.foo}}`? Also whats is there a difference between `<div {{on "click" this.foo}}>` and `<div {{action this.foo}}>`?

---
And how can `willInsertElement` implemented while [modifier managers](https://emberjs.github.io/rfcs/0373-Element-Modifier-Managers.html) when modifier managers dont have an even for this moment? Dont we need this first?

knownasilya:
@luxferresum element modifier managers RFC was merged and the feature implemented here: https://github.com/emberjs/ember.js/pull/17143

luxferresum:
@knownasilya But `installModifier`, the first event that receives the element, will always be "called after DOM insertion".

knownasilya:
@luxferresum maybe by caching the value and element?

chriskrycho:
Hey folks, what's the status on this?

NullVoxPopuli:
Based on the proposed implementation, and how everyone is using https://github.com/ember-modifier/ember-modifier,
I think this RFC needs updating

text/0000-modifiers.md
```diff
@@ -0,0 +1,404 @@
...
+
+```hbs
+<FooBar {{draggable x=x y=y}} />
+```
+
+```hbs
+<FooBar (draggable x=x y=y) />
+```
+
+This also means that they have no relationship to `...attributes`.
```
lvegerano: If I understand this, that means that a developer would not be able to forward the modifier. If so:
- What are the drawbacks of forwarding a modifier?
- What does that means for the `{{on}}` modifier? (currently can be forwarded)
  - Not giving modifiers the ability to forward and not changing `{{on}}` creates an inconsistent API
  - Changing `{{on}}` would be a breaking change

Thoughts?

Panman82: I think this is simply outdated information, as modifiers _are_ "forwarded" in `...attributes`. The RFC for that was created and merged _after_ this RFC was last updated: https://github.com/emberjs/rfcs/pull/435

chriskrycho:
Given that ember-modifiers has been implemented and has a nice API, is there anything blocking updating this text and merging it, with ember-modifiers as the official API? If so, what are the blockers? Is there a path where we can ship a public modifier manager API for other addons to use, even if we end up needing to iterate on that design later?

jelhan:
> Given that ember-modifiers has been implemented and has a nice API, is there anything blocking updating this text and merging it, with ember-modifiers as the official API? If so, what are the blockers? I.e. is there a path where we can ship a public modifier API for other addons to use, even if we end up needing to iterate on that design later?

The modifier manager is not supporting modifying an element *before* it's insert into DOM. This limits some use cases, which should be supported. Please find more details about this in #652.

The public API for functional modifiers exposed by `ember-modifier` doesn't seem to be flexible enough to support an additional lifecycle hook, which runs _before_ the element is inserted. Not saying that such will be added. Not sure if it should be usable through a functional modifier approach if it lands. But we should resolve these questions *before* committing to a public high-level API for writing modifiers.

wycats:
> Given that ember-modifiers has been implemented and has a nice API, is there anything blocking updating this text and merging it, with ember-modifiers as the official API? If so, what are the blockers? I.e. is there a path where we can ship a public modifier API for other addons to use, even if we end up needing to iterate on that design later?

@chriskrycho from my perspective, the biggest issue is unifying a bunch of features (modifiers, helpers, destroyables and ultimately usables) around a single high-level API.

From the perspective of modifiers, the biggest inconsistency between what I hope we ultimately do and what ember-modifiers does is: https://github.com/ember-modifier/ember-modifier/blob/master/addon/-private/class/modifier-manager.ts#L37

In the absence of a modifier supplying updating hooks, I would like these APIs to default to teardown/recreate across the board, which would be a breaking change if we stabilized `ember-modifier` into Ember right now.

lukemelia:
> Given that ember-modifiers has been implemented and has a nice API, is there anything blocking updating this text and merging it, with ember-modifiers as the official API? If so, what are the blockers? I.e. is there a path where we can ship a public modifier API for other addons to use, even if we end up needing to iterate on that design later?

Per [this open issue](https://github.com/emberjs/rfcs/issues/655), the question of whether and how args should be passed to `destroyModifier` should be reconsidered. As the issue describes, the current spec'd behavior as implemented by ember-modifier (destroy consuming the modifier's args) makes it easy to create confusing bugs.

NullVoxPopuli:
I think we discussed that using a proxy would allow lazy consumption of tracked tags. I volunteered to implement that behavior, but am waiting on some refactoring work to land from @pzuraq

wycats:
@lukemelia I totally agree with this. In general, destructors should be used in a way that's compatible with JS' new [finalizer](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/FinalizationRegistry) feature, which goes further than your request here and actually doesn't even make the object itself available to finalizers. This is because it's way to easy to accidentally resuscitate an object in its own finalizer, resulting in confusing bugs.

Finalizers address the problem by allowing you to supply a *different* piece of metadata that gets passed into the destructor, so you can do parameter-specific things during destruction without waking the object back up.

That was a long-winded way of saying I agree with your design proposal, and also think we should consider aligning [RFC #580 (destroyables)](https://emberjs.github.io/rfcs/0580-destroyables.html) with the finalizer design. @pzuraq 

Note that we would *not* want to use the finalizer design directly for our destroyables, because we want to offer more timing guarantees for resource cleanup. Our destroyables are closer to the [explicit resource management](https://github.com/tc39/proposal-explicit-resource-management) JS proposal.

(I mentioned the finalizer design because @lukemelia made me realize that the kinds of bugs that the current design enables are similar in spirit to the kinds of mistakes that the finalizer design is trying to avoid).

NullVoxPopuli:
Was this accidentally implemented? Do we still need this RFC

jelhan:
> Was this accidentally implemented? Do we still need this RFC

I'm not sure if the RFC is this still needed. But I don't think it's implemented.

`ember-modifiers` is not part of default blueprint. User-defined modifiers are also not captured in the guides.

Maybe the RFC could be updated to add existing `ember-modifiers` package to application blueprint and cover it in the guides?

NullVoxPopuli:
With https://github.com/emberjs/rfcs/pull/757
I don't think we need ember-modifier to be in the blueprint

jelhan:
> With #757 I don't think we need ember-modifier to be in the blueprint

I'm not sure if I agree. That would make state-less, function-based modifiers a first-class concern with built-in support while state-full, class-based modifiers are user-land. If one is teached in the guides but the other not, that may make discoverability of the other option difficult.

NullVoxPopuli:
It's _an improvement_ over today's current state of the discoverability of any modifiers difficult.

_but_, I'd say _most_ modifiers don't actually need state. State can be useful, but I think we'd want a separate RFC for figuring out class-based modifiers as default.

wagenet:
Is this something we still need? We do have modifiers now.

locks:
As mentioned in the description of the mentioned PR, https://github.com/emberjs/rfcs/pull/811 supersedes this, so we can close it!

PR #362, opened 2018-08-22
Ember Data: Return Promise from DS.Model.save()

fivetanley:
[Rendered](https://github.com/fivetanley/rfcs/blob/deprecate-promise-object-save/text/0000-ember-data-return-promise-from-ds-model-save.md)

### DIFF
--- a/text/0000-ember-data-return-promise-from-ds-model-save.md
+++ b/text/0000-ember-data-return-promise-from-ds-model-save.md
@@ -6,7 +6,7 @@
 
 ## Summary
 
-DS.Model.save() will return an Ember.RSVP.Promise instead of a PromiseObject.
+DS.Model.save() will return an RSVP.Promise instead of a PromiseObject.
 
 ## Motivation
 
@@ -19,7 +19,7 @@ DS.Model.save() will return an Ember.RSVP.Promise instead of a PromiseObject.
 
 ## Detailed design
 
-Introduce a new feature flag, “ds-return-promise-from-model-save”. When enabled, DS.Model.save() will return a RSVP.Promise that resolves with the model if the save was successful, or rejects with an error if the save fails. When disabled, a PromiseObject will be returned to keep today's behavior, but accessing properties on the PromiseObject or calling non-Promise functions (.then, .finally, .catch) on the PromiseObject will issue a deprecation warning.
+Introduce a new feature flag, “ds-return-promise-from-model-save”. When enabled, DS.Model.save() will return an RSVP.Promise that resolves with the model if the save was successful, or rejects with an error if the save fails. When disabled, a PromiseObject will be returned to keep today's behavior, but accessing properties on the PromiseObject or calling non-Promise functions (.then, .finally, .catch) on the PromiseObject will issue a deprecation warning.
 
 Deprecation Plan:
 
--- a/text/0000-ember-data-return-promise-from-ds-model-save.md
+++ b/text/0000-ember-data-return-promise-from-ds-model-save.md
@@ -1,0 +1,65 @@
+* Start Date: 08/22/2018
+* RFC PR: (leave this empty)
+* Ember Issue: (leave this empty)
+
+# Return a Promise from DS.Model.save()
+
+## Summary
+
+DS.Model.save() will return an Ember.RSVP.Promise instead of a PromiseObject.
+
+## Motivation
+
+* The API documentation already documents the return value as a Promise.
+* Remove dependency on promise proxies
+* Async Consistency - The PromiseObject encourages a usage pattern of sometimes-async, sometimes-sync behavior. We want to [reduce zalgo](http://blog.izs.me/post/59142742143/designing-apis-for-asynchrony).
+* API consistency - While the PromiseObject proxies properties, the PromiseObject does not proxy methods on the underlying model. For instance, if you try to call a method like destroyRecord on the model, you will get a “not a function” error because the method is called on the Proxy and not the underlying model. This encourages users for Ember Data to reach into private implementation details of the promise proxy, such as using proxy.get('content') to call methods.
+* Enable new functionality in Ember Data while making backwards compatibility with older versions of Ember easier.
+* The documentation of using PromiseProxy APIs has been replaced over time in the guides and API documentation to use more typical JavaScript usage of Promises, such as async/await, instead of PromiseProxies.
+
+## Detailed design
+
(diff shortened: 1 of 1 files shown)


btecu:
This is great!

text/0000-ember-data-return-promise-from-ds-model-save.md
```diff
@@ -0,0 +1,65 @@
...
+* Enable new functionality in Ember Data while making backwards compatibility with older versions of Ember easier.
+* The documentation of using PromiseProxy APIs has been replaced over time in the guides and API documentation to use more typical JavaScript usage of Promises, such as async/await, instead of PromiseProxies.
+
+## Detailed design
+
+Introduce a new feature flag, “ds-return-promise-from-model-save”. When enabled, DS.Model.save() will return an RSVP.Promise that resolves with the model if the save was successful, or rejects with an error if the save fails. When disabled, a PromiseObject will be returned to keep today's behavior, but accessing properties on the PromiseObject or calling non-Promise functions (.then, .finally, .catch) on the PromiseObject will issue a deprecation warning.
```
runspired: We should go directly from canary => beta => release with the deprecation on. I don't believe we need or should use a feature flag here to guard the new world, as promiseObjects do function as normal promises if used as such.

That said, @rwjblue may have some thoughts, and using the flag as suggested below would let us code-strip for users that opt-in to a post-proxy world.

text/0000-ember-data-return-promise-from-ds-model-save.md
```diff
@@ -0,0 +1,65 @@
...
+
+Deprecation Plan:
+
+The deprecations can be implemented by adding deprecations around the existing PromiseObject. However, we can't include the deprecations in the actual PromiseObject class because [it is the base class for the relationship proxies](https://github.com/emberjs/data/blob/master/addon/-legacy-private/system/promise-proxies.js#L85). Instead, we can create and use a new DeprecatedPromiseObject for the DS.Model.save() function. It functions like a PromiseObject but deprecates the following behavior:
+
+    * Accessing properties on the PromiseObject returned from DS.Model.save()
+        * Use deprecateProperty from Ember to deprecate all the properties that the PromiseProxy Mixin provides. The list of properties can be found here: https://github.com/emberjs/ember.js/blob/master/packages/ember-runtime/lib/mixins/promise_proxy.js
```
runspired: In the deprecation for this, we should point users towards ember-concurrency or creating their own PromiseObject wrappers if they do desire to continue using the promise resolution flags.

text/0000-ember-data-return-promise-from-ds-model-save.md
```diff
@@ -0,0 +1,65 @@
...
+
+For users relying on this behavior, they will have to refactor their code to either use patterns like async/await, ember-concurrency, or ember-promise-helpers. Alternatively, if all they want to access are properties on the model, they can use the model instead.
+
+## Alternatives
+
+The impact of not doing this prevents further changes in Ember Data and requires more effort to maintain compatibility between Ember Data and older versions of Ember.
+
+## Unresolved questions
+
+* It's quite possible Ember Data users use the isSettled, isRejected, isFulfilled, and isPending properties of the  PromiseProxyMixin to display state in their UI. For example, you may have a component that calls record.save(), and uses isPending to show a loading spinner or isRejected to show UI for error states. Should we point them to community alternatives like [ember-promise-helpers](https://github.com/fivetanley/ember-promise-helpers) or [ember-concurrency](http://ember-concurrency.com/docs/introduction/) in the deprecation message for these properties?
```
runspired: ha, got to the end and saw this. I think yes!

luxferresum:
In my opinion`.save()` is not really the problem. The primary problem are relations, followed by `findAll`/`findRecord`/`query`/`queryRecord`. And to solve them we need a way to work seamlessly with computed properties, templates and promises. There are different ideas and solutions like `await` helpers. [I even have my own little helper for computed properties](https://github.com/luxferresum/ember-computed-promise). However I think we need something *official*.

And I'm not sure if I like the idea to change the result of `save` at a different time then the result of relationships and `findAll`/`findRecord`/`query`/`queryRecord`. This just creates two different APIs, but does not really solve the problem. Its easier if `save` just returns the same as `findRecord`.

igorT:
This looks great to me as well

text/0000-ember-data-return-promise-from-ds-model-save.md
runspired: I take this back and we should use a feature flag.

rwjblue:
Chatted about this in todays ember-data meeting, the general consensus is that we'd like to move forward.

text/0000-ember-data-return-promise-from-ds-model-save.md
runspired: Is mentioned further down.

hjdivad:
Discussed during today's meeting; we've already agreed this looks like a good path forward; putting it in FCP.

pete-the-pete:
COOL https://github.com/pete-the-pete/data/tree/promise-after-save <--
hopefully that can come in handy.

HeroicEric:
I think this might be a good opportunity to do the same refactoring for `Model.reload()`, which also currently returns a `PromiseObject` for all of the same reasons mentioned in this RFC.

Gaurav0:
This has been in FCP for months?

runspired:
After discussion today we decided that the return value here needs to be a custom thenable, basically a `Promise` with three additional properties for knowing the state

```
isPending
isResolved
isRejected
```

We are otherwise 👍 on merging this

knownasilya:
Ping

runspired:
I'll mention at the next meeting that we forgot to merge, we should go ahead and merge :)

text/0000-ember-data-return-promise-from-ds-model-save.md
snewcomer, line 65: We could also implement something on top of Promise that exposes these flags.

https://github.com/snewcomer/ember-stateful-promise/blob/main/addon/decorators/stateful-function.js

snewcomer: I guess their method could be wrapped with something similar to this or ember-concurrency as well.  Almost every `save` we deal with uses derived state to help the UI along and prevent user initiated bugs. 

So I take that back.  Returning a simple Promise is better.

snewcomer:
supplanted by https://github.com/emberjs/rfcs/pull/795!

PR #367, opened 2018-08-27
Module Unification Packages (MU with Ember Addons)

mixonic:
An iteration on and replacement of https://github.com/emberjs/rfcs/pull/309

[Rendered](https://github.com/mixonic/rfcs/blob/mu-packages/text/0000-module-unification-packages.md)

---

> [RFC #143](https://github.com/emberjs/rfcs/blob/master/text/0143-module-unification.md) introduced a new filesystem layout for Ember applications we've called "Module Unification".
>
> Ember addons can use the module unification filesystem layout (i.e. an addon with a `src/` directory) in place of the legacy `app/` or `addon/` folders. The `src/` folder encapsulates a module unification "package", and Ember applications can access components, helpers, and services from that package.
>
> This RFC rejects some parts of the Module Unification RFC, such as `::` component invocation syntax and implicit invocation of an addon's `main`. These changes are largely motivated by conflicts between these designs and new features in Ember and NPM.

> This RFC describes a new helper `{{use}}` for explicit importing of a component or helper from a package. This replaces the `::` syntax in templates.

> Service injections do not currently have source information encoded. Here we propose adding that information via an AST transform and argument to
`inject.service`. This additional information provided by Ember CLI at build
time mirrors how `source` is already a compile-time addition for templates.

### DIFF
--- a/text/0000-module-unification-packages.md
+++ b/text/0000-module-unification-packages.md
@@ -1,0 +1,992 @@
+# Module Unification Packages
+
+- Start Date: 2018-03-02
+- RFC PR:
+- Ember Issue:
+
+# Module Unification with Ember Addons
+
+## Summary
+
+[RFC #143](https://github.com/emberjs/rfcs/blob/master/text/0143-module-unification.md)
+introduced a new filesystem layout for Ember applications we've called "Module
+Unification".
+
+Ember addons can use the module unification filesystem layout (i.e. an addon
+with a `src/` directory) in place of the legacy `app/` or `addon/` folders. The
+`src/` folder encapsulates a module unification "package", and Ember
+applications can access components, helpers, and services from that package.
+
+This RFC rejects some parts of the Module Unification RFC, such as
+`::` component invocation syntax and implicit invocation
+of an addon's `main`. These changes are largely motivated by conflicts between
+these designs and new features in Ember and NPM.
+
+This RFC describes fully how packages work in Ember, and how apps and addons will migrate
+from "classic" `app/` and `addon/` folders to the modern `src/` folder.
+It introduces new APIs:
+
+* Accessing a component or helper from a package requires the `{{use}}` helper.
+* The service injection API gains a new argument `package`. It also gains
+  an argument `source` which Ember CLI will add to all invocation with
+  an AST transform.
+* `lookup`, `factoryFor` and other owner APIs gain new `package` and `source` APIs.
+
+Throughout this document the term "absolute specifier" is used. This refers to a
+serialization of a module's identity in Module Unification. Absolute specifiers
+are not specified in
+[RFC #143](https://github.com/emberjs/rfcs/blob/master/text/0143-module-unification.md)
+or anywhere else, and this RFC does not attempt or require a formal definition
+of the serialization. Traditionally we've used something like
+`component:/my-app/ui/components/foo-bar` as the absolute specifier format.
+
+## Motivation
+
+[RFC #143](https://github.com/emberjs/rfcs/blob/master/text/0143-module-unification.md)
+introduced a new filesystem layout for Ember applications we've called "Module
+Unification". The RFC described semantics for ["Addon modules"](https://github.com/emberjs/rfcs/blob/master/text/0143-module-unification.md#addon-modules) and the concrete
+example of components from an addon being
+invoked as `{{my-addon::my-component}}`.
+
+The syntax of `::` was intended to allow us to specify an **explicit** package for an
+invocation. The lookup would only occur within the prefixed package name. The RFC
+touches on these lookup rules and the syntax for explicit invocation only lightly.
+
+Contemporaneous with the Module Unification discussion in ember, NPM announced a new feature: Scoped packages. Scoped
+packages have a composite name in the format `@${scope}/${packageName}`.
+For example:
+
+```js
+// import from a "traditional" NPM package.
+//
+// Import the default export of the moment package.
+// Often this is node_modules/moment/index.js
+import moment from 'moment';
+
+// Import from a "scoped" NPM package.
+//
+// Import the default export of a scoped package.
+// For example node_modules/@angular/router/index.js
+import { RouterModule } from '@angular/router';
+```
+
+**The addition of `@npmscope/` as a valid part of package names is in conflict with
+the original design of Module Unification package invocation syntax.** For
+example here are several invocation examples which would be valid presuming
+Module Unification's original package proposal, NPM scopes, and angle bracket
+invocation syntax:
+
+```hbs
+{{#@npmscope/package-name::component-name}}
+  Invoke a component from a scoped NPM package.
+{{/@npmscope/package-name::component-name}}
+
+{{#@npmscope/package-name}}
+  Would invoke an implicit component "main" in the package.
+{{/@npmscope/package-name}}
+
+<@Npmscope/PackageName::Name>
+  Invoke a component from a scoped NPM package.
+</@Npmscope/PackageName::Name>
+
+<@Npmscope/PackageName>
+  Some content. Would invoke an implicit component "main" in the package.
+</@Npmscope/PackageName>
+```
+
+Because RFC \#143 was a) not designed with NPM scopes in mind, b) was not
+designed with `@var` syntax fully explored, and c) was not designed with
+Ember's consensus `<AngleBracket>` syntax the examples above violate several design constraints:
+
+- **Use of `@` when invoking a component from a scoped package should be avoided.**
+  Recent versions of Ember and Ember RFCs have introduced the sigil `@` to
+  templates with a specific meaning. `{{@someArg}}` reads an argument passed to
+  a component, and arguments are themselves passed with the `@` prefix (for
+  example `<MyComponent @someArg={{someObj}}>`). Arguments can also be invoked
+  (`<@someArgWithClosureComponentValue>` per [RFC 226](https://github.com/emberjs/rfcs/blob/master/text/0226-named-blocks.md#unified-renderable-syntax)) or be created from a block (`<@someArg= >` per [RFC
+  317](https://github.com/emberjs/rfcs/blob/named-block-syntax/text/317-named-block-syntax.md#summary)). As the meaning of `@` is very specific and important in Ember templates,
+  we should not create visual ambiguity and confusion by using it for non-variable component lookup. `<@scope/package::name>` is a poor invocation syntax for a
+  component from an NPM scoped package because it is visually ambiguous with
+  Ember's existing `@` usage.
+- **Use of `/` when invoking a component from a scoped package should be avoided.**
+  When using angle brackets multiple closing tags are very difficult to
+  disambiguate from multiple closing tags. We’ve already discouraged use of `/` in
+  component names by disallowing arbitrary directories in Module Unification layout apps. For comparison here are two plausible strings in an Ember template which
+  are difficult to disambiguate:
+    - `</@Npmscope/PackageName::Name>`
+    - `</@Npmscope></PackageName::Name>`
+- **NPM package names should be treated as literal strings.** The issues addressed
+  by this RFC arose
+  because we made an assumption that NPM package names would be a single word or
+  a dasherized series of words. That was incorrect, as both `@` and `/` are now
+  valid parts of an NPM package name. In the future, we should assume that
+  new features may be added to NPM with additional characters. So, to be
+  future-safe our design should
+  treat NPM package names as strings where any characters would be valid.
+
+I suggest two further constraints:
+
+- **Use of NPM scopes should not be discouraged.** It could be tempting to accept a design
+  which works well for “normal” NPM packages but requires more effort for users of
+  scoped packages. By accepting one of these options we would a) make Ember more
+  difficult to learn since you must learn two different solutions for the same
+  problem of addon component/helper/service invocation and b) by making un-scoped packages more
+  ergonomic we would discourage addon authors from using the scoped packages
+  feature. In practice, scoped packages seem like a good feature for authors to
+  consider and in some corporate settings use of them may not be optional.
+- **Resolution rules should be consistent across parts of Ember.** For example
+  normalization rules for package names in templates and in service injections
+  should be the same. Given that a developer is using Ember Data they should not
+  refer to it in some places as EmberData and others as `ember-data`.
+
+These are the problems and constraints that motivate this RFC.
+
+## Detailed design
+
+### Removal of RFC \#143 Module Unification features
+
+The following features from RFC \#143 are removed from the updated design:
+
+* **Drop the single line package+name syntax (`::`)**. This syntax is the root
+cause of several issues in templates outlined in the motivation section. This RFC does not
+suggest any replacement "single-line" invocation syntax in templates. The
+`::` syntax was also implied to be used with service injections. As
+we're removing the micro-syntax from templates, we will also remove it from
+the service injection API.
+* **Drop implicit "main" rules from module unification.** Main modules and their
+implicit resolution was [introduced in #143](https://github.com/emberjs/rfcs/blob/master/text/0143-module-unification.md#main-modules).
+Because implicit main is a
+*resolution rule* and not an invocation syntax rule, there is no way for us to
+know if `<Foo />` is an app component or a package name + main without
+significant analysis of the program at build time. As a runtime
+resolver is sure to be part of common Ember usage for a long while, it seems
+beneficial not to introduce this ambiguity. Additionally, the whole-program knowledge
+needed to disambiguate `<Foo>` may make it difficult to write tooling
+for templates.
+  * Resolution of "main" is not entirely removed from MU. For example
+  `src/router.js` should still be resolved via
+  `resolveRegistration('router:main')`. Only the implicit main of a package
+  is removed.
+* **Remove the stripping of `ember-` or `ember-cli-` prefixes from
+package names.** [RFC #143: Module Unification - Main
+Modules](https://github.com/emberjs/rfcs/blob/master/text/0143-module-unification.md#main-modules).
+describes stripping these prefixes during build time, and so invocations
+referencing those package names can ignore the prefix.
+For example `{{ember-simple-auth::login}}` would be incorrect for a component
+in the ember-simple-auth addon, and `{{simple-auth::login}}` would be correct.
+As this RFC suggests removing single-line invocation entirely, this
+convenience is unnecessary. Ember should be conservative about how
+it changes the names of packages. A new developer installing a package should
+not need to internalize a bunch of naming rules. Knowing the name of the
+package itself should be sufficient.
+
+The removal of these features of course requires that we provide alternatives.
+These follow.
+
+### New Package APIs
+
+Without using the `owner` API, only **services**, **components**, and **helpers** from a module unification
+package can be invoked by an application. The modules contained in an addon's
+package are defined by the files in the addon's `src/` directory.
+
+Both addons and app have a package. For this reason writing Module Unification
+addon and app code should be very consistent. Modules in the
+`src/` directory are invoked by their bare names, and invoking/injecting/looking up things from another package will require a special API.
+
+A summary of new behaviors follows (specific detailed designs each have a section of this RFC):
+
+* Invoking a component, helper, or service without an explicit package
+  will always look to the **implicit** package of that lookup. For an addon
+  or app, this means that first local lookup is considered for resolution then
+  their top-level directory for that module under their own `src/`.
+* To invoke a component or helper from another package, the `{{use}}` helper
+  is defined by this RFC.
+* To implicitly inject a service from a package the `source` argument
+  is added to the injection API. To explicitly inject a service from another package the `package` argument is defined
+  for the injection API.
+* To look up or resolve an arbitrary module in local lookup the `source`
+  argument is defined for `lookup`, `factoryFor`, and `resolveRegistration`. To explicitly look up or resolve an arbitrary module the `package` argument is added.
+
+Each of these changes is described below in detail.
+
+To help the RFC be approachable I'll describe behaviors using plausible
+filenames. An actual implementation for any of these examples would be
+based on Module Unification resolution rules, so most filenames used in
+examples have alternative forms which are valid.
+
+### Implicit packages for templates
+
+An application using Module Unification already has an implicit package.
+From an app you can invoke
+`{{try-me}}` and it looks in the `src/ui/components/` directory. As an app author
+you are not required to be
+explicit and provide the app name.
+
+This RFC suggest addon templates behave in the same manner. An invocation of
+`{{try-me}}` in an addon will resolve:
+
+* Local lookup `my-addon/src/ui/components/invoking-component/try-me/component.js`
+* Top-level lookup `my-addon/src/ui/components/try-me/component.js`
+
+Metadata on compiled Glimmer templates already contains the path on disk of
+a template. This metadata should be expanded to also include an absolute
+specifier for the template.
+
+Using the `source` metadata on a template (or another appropriate build-time value), all lookups from that template will determine the package of the templates
+and use it for any implicit invocations in the template itself.
+
+### The `{{use}}` helper
+
+This RFC describes a new helper `{{use}}` for explicit importing of a component or
+helper from a package. This replaces the `::` syntax in templates.
+
+The `{{use}}` helper's functionality is a subset of JavaScript's `import` keyword, but the syntax is
+different. `{{use}}` is required to bring a component or helper from another
+package (and not from a directory or file) into the template's scope as a symbol.
+
+The `{{use}}`
+helper has unique syntax compared to any keywords present in Glimmer today
+because it introduces a symbol to the template without a block.
+
+To start with an example: In this template the `{{use}}` helper imports a component `Widget` from the
+`gadget` addon.
+
+```hbs
+{{! invokes node_modules/gadget/src/ui/components/Widget/component.js }}
+
+{{use Widget from 'gadget'}}
+<Widget @options=someOptions @value=someValue />
+```
+
+The semantics of `{{use}}` are the following:
+
+- `{{use}}` introduces a symbol to template that corresponds to the
+  component or helper (in the above example `Widget`) from the package (in the above example `gadget`). Normal
+  invocation rules apply to the symbol: To be invoked with `{{` it must have a
+`-`, to be invoked with a `<` it must start with a capital letter.
+- There are no "default" imports. The component or helper being imported is
+  always named.
+- The symbol introduced by `{{use}}` is an invokable symbol. It has identical
+  semantics to a symbol with a closure component value. The resolution of the
+  component module is (semantically if not literally) performed at the time of
+  `{{use}}`, not at the time of invocation. Because the symbol's value is
+  already resolved when the component is invoked, there are no normalization
+  rules (`CapCase` -> `caps-case`) applied to symbols introduced by `{{use}}`.
+- `{{use}}` can use `as` to map a component or helper name to a differently
+  named symbol.
+- `{{use}}` can only be used to access absolute package names. There are no
+  relative paths relating to file names or directories.
+- `{{use}}` introduces a hoisted symbol for the entire template. The helper, like
+  `import`, must be at the top level of the template (not inside a block, element, or subexpression)
+  and cannot be used dynamically. For example it cannot be invoked inside an `{{#if}}` block.
+- If a symbol introduced by `{{use}}` is already present in a template a
+  compilation error is thrown for `Duplicate declaration ${BindingIdentifier}`.
+  This mirrors the behavior of ES imports.
+
+The syntax of `{{use}}` can be described using syntax spec language
+comparable to the [ES `import`
+spec](https://www.ecma-international.org/ecma-262/6.0/#sec-imports):
+
+* *UseDeclaration*:
+  * `{{` `use` ImportsList FromClause `}}`
+* *ImportsList*:
+  * *ImportSpecifier*
+  * *ImportsList*, *ImportSpecifier*
+* *ImportSpecifier*:
+  * *ImportedBinding*
+  * *IdentifierName* `as` *ImportedBinding*
+* *FromClause*:
+  * `from` *ModuleSpecifier*
+* *ModuleSpecifier*
+  * *StringLiteral* - The module unification package from which a components
+    or helper is being imported. Usually
+    an NPM package name.
+* *ImportedBinding*
+  * *BindingIdentifier* - The symbol used to reference the import.
+
+And what follows are several examples of valid usage.
+
+##### Using a component with mustaches
+
+```hbs
+{{! use package-name/src/ui/components/component-name/component.js }}
+
+{{use component-name from 'package-name'}}
+
+{{component-name}}
+
+{{#component-name}}
+  Some content
+{{/component-name}}
+```
+
+```hbs
+{{! use @npmscope/package-name/src/ui/components/component-name/component.js }}
+
+{{use component-name from '@npmscope/package-name'}}
+
+{{#component-name}}
+  Some content
+{{/component-name}}
+```
+
+##### Using a component with angle brackets
+
+```hbs
+
+{{! use package-name/src/ui/components/Name/component.js }}
+
+{{use Name from 'package-name'}}
+
+<Name />
+
+<Name>
+  Some content
+</Name>
+```
+
+```hbs
+{{! use @npmscope/package-name/src/ui/components/Name/component.js }}
+
+{{use Name from '@npmscope/package-name'}}
+
+<Name>
+  Some content
+</Name>
+```
+
+##### Re-assignment of the imported component to a new symbol
+
+`{{use}}` introduces a symbol to the template, and can choose a local name.
+
+```hbs
+{{! use @npmscope/package-name/src/ui/components/Name/component.js }}
+
+{{use Name as component-name from 'package-name'}}
+
+{{component-name}}
+
+
+{{use Name as scoped-package-component-name from '@npmscope/package-name'}}
+{{scoped-package-component-name}}
+```
+
+```hbs
+{{! use @npmscope/package-name/src/ui/components/component-name/component.js }}
+
+{{use component-name as Name from 'package-name'}}
+
+<Name />
+
+
+{{use component-name as ScopedPackageName from '@npmscope/package-name'}}
+<ScopedPackageName />
+```
+
+
+##### Multiple Imports
+
+`{{use}}` can import multiple components separated by `,`.
+
+```hbs
+{{! use ember-power-select/src/ui/components/Option/component.js }}
+{{! use ember-power-select/src/ui/components/Select/component.js }}
+
+{{use
+  Option,
+  Select as PowerSelect
+from 'ember-power-select'}}
+
+<PowerSelect>
+  <Option />
+</PowerSelect>
+```
+
+##### `{{use}}` causes and error for duplicate declarations
+
+This template will throw an error `Duplicate declaration "ComponentName"`:
+
+```hbs
+{{use ComponentName from 'package-name'}}
+{{use ComponentName from 'other-package-name'}}
+
+<ComponentName />
+```
+
+To use the second import an author must provide a unique import binding.
+
+##### `{{use}}` declaraions are hoisted
+
+Mirroring the behavior of ES `import`, `{{use}}` declarations are hoisted
+in a template. The following is not suggested (and perhaps we should
+lint against it) but would be valid:
+
+```hbs
+<ComponentName />
+
+{{use ComponentName from 'package-name'}}
+```
+
+### Impacts of `{{use}}` on the `{{component}}` helper
+
+The `{{component}}` helper that exists in Ember today can look up any component
+in an application or addon. All components share a single namespace (all modules
+are merged into `app/`) and `{{component 'foo-bar'}}` simply resolves a full
+name of `component:foo-bar`.
+
+This design has a notable flaw: If a template uses the `{{component}}` helper
+then **all components from the app or addons** must be in memory when the
+template is rendered so they can be synchronously resolved. This is a behavior
+of the `{{component}}` helper we would not like to persist into the module
+unification design.
+
+For example the following are valid uses of `{{component}}` in Ember today:
+
+```hbs
+{{! will resolve `component:foo-bar`, could be in an app or addon }}
+{{component 'foo-bar'}}
+
+{{! will read the value of the binding and resolve that string, could be in an app or addon }}
+{{component someBindingToAString}}
+
+{{#some-component-yielding-component as |closureComponent|}}
+  {{! invoke a closure component }}
+  {{component closureComponent}}
+{{/some-component}}
+```
+
+In a module unification template (a template in `src/`) there is no way to
+invoke a component from a module unification addon without `{{use}}`. Because
+`{{use}}` merely introduces a symbol into the template's context, **there is no
+way to dynamically lookup an addon component in module unification**.
+
+For example the following is a valid use of `{{use}}` and `{{component}}`:
+
+```hbs
+{{! my-app/src/ui/routes/index.hbs }}
+
+{{! invoke gadgets/src/ui/components/foo-bar/component.js }}
+{{use foo-bar from 'gadgets'}}
+
+{{component foo-bar}}
+
+{{component (component foo-bar)}}
+
+{{! yield the component to the invocation site }}
+{{yield (component foo-bar)}}
+```
+
+```hbs
+{{! gadgets/src/ui/components/baz-qux/template.hbs }}
+
+{{! invoke gadgets/src/ui/components/foo-bar/component.js }}
+{{component 'foo-bar'}}
+
+{{component (component 'foo-bar')}}
+
+{{! yield the component to the invocation site }}
+{{yield (component 'foo-bar')}}
+```
+
+However the following are *NOT* valid:
+
+
+```hbs
+{{! my-app/src/ui/routes/index.hbs }}
+
+{{! Each of these invocations would render nothing }}
+
+{{use foo-bar from 'gadgets'}}
+
+{{component 'foo-bar'}}
+
+{{component (component 'foo-bar')}}
+
+{{component symbolWithStringValueOfFooBar}}
+
+{{! yields a closure component that renders nothing }}
+{{yield (component 'foo-bar')}}
+```
+
+Since the import of a component from an addon is explicit, even with the `{{component}}`
+helper, it is easy to analyze templates and tree-shake modules across package
+bounderies. For example component modules from module unification addons can
+be tree-shaken based on a) their declarative use in the app modules via
+`{{use}}`, b) their declarative use in the addon itself via standard component
+invocation, and c) the presence of a dynamic `{{component}}` helper in the addon
+(the presence of which would defeat any tree-shaking of the addon's top-level
+modules).
+
+It remains impossible to tree shake
+package (app or addon) modules when the dynamic and bound `{{component}}` helper is present.
+Despite this, the ability to tree-shake modules across package bounderies and
+with local lookup rules many common cases
+are solved in this approach.
+
+### `prelude.hbs`
+
+In many cases `{{use}}` will be more verbose than today's `app/`-merged global
+namespace. For example where today Ember apps might write:
+
+```hbs
+{{#power-select options=names as |name|}}
+  {{name}}
+{{/power-select}}
+```
+
+In an MU world with angle bracket invocation we will write:
+
+```hbs
+{{use Select from 'ember-power-select'}}
+<Select @options=names as |name|>
+  {{name}}
+</Select>
+```
+
+The overhead of an extra line is not expected to be burdensome. From a developer
+experience perspective:
+
+* There is no `{{use}}` statement for any component or helper from the app
+  itself.
+* An explicit import makes it easy for multiple addons to share the same
+  single-word component name without conflict. This is a notable benefit.
+  Both could even be used in the same template if one is re-named.
+* In many apps the component from a given addon is not used in hundreds or tens
+  of templates, but in only a few. For example an app using Google Maps via
+  ember-g-map may only have a single template showing a map even if there are
+  hundreds of templates. The extra verbosity in that case isn't burdensome.
+* For apps that are extremely large it is common for an addon like
+  ember-power-select to be wrapped in an app-specific component providing
+  appropriate defaults or UI for that app. Thus the `{{use}}` will often be
+  minimized to a few files even if the addon component is rendered in many places.
+
+Despite this, in some common use cases `{{use}}` may be too verbose a tool for bringing
+components and helpers into scope. For example:
+
+* A translation helper used throughout a codebase, for example `{{t 'name'}}`
+  would mean writing `{{use t from 'ember-intl'}}` at the top of every file.
+* The popular ember-truthy-helpers package adds
+  operator-ish helpers to Ember templates such as `(and itemA itemB)`. Importing
+  these in each template would be painful.
+
+Addons which provide globally available helpers or components are a valid use
+case we would like to address. To do so this RFC suggests a `prelude.hbs` file
+which is effectively prepended to every template in an app.
+
+#### Detailed design for `prelude.hbs`
+
+An Ember application can provide a file `src/prelude.hbs`. At compile time this
+file is prepended to every template in the application. For example given the
+following two files:
+
+```hbs
+{{! src/prelude.hbs }}
+{{use Select from 'ember-power-select'}}
+```
+
+```hbs
+{{! src/ui/routes/posts/template.hbs }}
+<Select @options=names as |name|>
+  {{name}}
+</Select>
+```
+
+At build time and before template compilation they would be combined to a
+single file:
+
+```hbs
+{{! src/prelude.hbs }}
+{{use Select from 'ember-power-select'}}
+{{! src/ui/routes/posts/template.hbs }}
+<Select @options=names as |name|>
+  {{name}}
+</Select>
+```
+
+* This file contains all the information the compiler, and any template tooling,
+  will need to resolve all non-app components staticly.
+* The template compiler may, during compilation, choose to strip `{{use}}`
+  calls that add an unused identifier to the template. As such making
+  a `{{use}}` global can still have a very low runtime and packaging overhead.
+* The unused identifier stripping is to the exclusion
+  of templates which have a `{{component}}` helper. Because a template with a
+  `{{component}}` helper does not know until runtime what component is being
+  invoked, staticly un-used identifiers must be retained for possible dynamic
+  invocation.
+* Because duplicate declarations are not permitted (you cannot `{{use}}` two
+  things with the import binding `Select` without renaming one of them)
+  Once an addon has claimed a name in a prelude the app will not, in its own
+  templates, be able
+  to write `{{use}}` for that name without a re-assignment.
+
+If an addon wants to provide a global component/helper the following options
+are available:
+
+* The addon author can use a post-install generator to re-export their component
+  into the app's package scope, making it act like any other component or helper
+  in the app. This is effective but not very upgrade-friendly, as the app and
+  not the addon decide what will be makde global.
+* The addon author can use a broccoli transform to add their re-export to the
+  applications `src/` tree. This technique would be better, but is clumsy to
+  implement and makes writing good tempalate tooling impossible.
+* The addon author can write a broccoli transform to alter hbs templates and add
+  appropriate `{{use}}` statements. This suffers the same flaws as the previous
+  suggestion.
+
+### Implicit packages for services
+
+Service injections do not currently have source information encoded. Here we
+propose adding that information via an AST transform and argument to
+`inject.service`. This additional information provided by Ember CLI at build
+time mirrors how `source` is already a compile-time addition for templates.
+
+For example given a service `main` in the addon `gadget`, an injection
+will maintain the `gadget` package:
+
+```js
+// gadgets/src/services/main.js
+import Service, { inject } from '@ember/service';
+
+export default Service.extend({
+  maguffin: inject()
+});
+```
+
+This snippet injects a service `maguffin` from the addon while preserving the implicit
+package of `gadgets` (`gadgets/src/services/maguffin.js`). The implicit
+package is encoded at build time by an AST transform which adds a `source`
+option to the injection. The post-transform file would look like:
+
+```js
+// gadgets/src/services/main.js
+import Service, { inject } from '@ember/service';
+
+export default Service.extend({
+  maguffin: inject('maguffin', {
+    source: 'gadgets/src/services/main'
+  })
+});
+```
+
+All uses of Service `inject` would have `source` added, including those in
+components or helpers.
+
+Note: Because services are not configured in the Ember module unification config to be
+an available as a nested collection, services are never
+subject for local lookup resolution despite using `source`.
+
+
+### Explicit packages for service injections
+
+The argument `package` is added to the `inject` API to specify an
+explicit package. The package provided by this argument is absolute.
+Any implicit package is over-ruled.
+
+For example:
+
+```js
+export default Ember.Component.extend({
+
+  // inject src/services/geo.js
+  geo: inject(),
+
+  // inject node_modules/ember-stripe-service/src/services/store.js
+  checkoutService: inject('stripe', { package: 'ember-stripe-service' }),
+
+  // inject node_modules/ember-simple-auth/src/services/session.js
+  session: inject({ package: 'ember-simple-auth' })
+
+});
+```
+
+### Explicit packages for `owner` APIs
+
+Owner objects (the `ApplicationInstance` instance) in Ember should have new
+options introduced that allow developers to programmatically interact with these
+features.
+
+##### `owner.lookup()`
+
+```js
+lookup(fullName, {
+  source, // Optional source of the lookup
+  package // Explicit package
+});
+```
+
+##### `owner.factoryFor()`
+
+```js
+factoryFor(fullName, {
+  source, // Optional source of the lookup
+  package // Explicit package
+});
+```
+
+##### `owner.resolveRegistration()`
+
+```js
+resolveRegistration(fullName, {
+  source, // Optional source of the lookup
+  package // Explicit package
+});
+```
+
+##### `owner.hasRegistration()`
+
+```js
+hasRegistration(fullName, {
+  source, // Optional source of the lookup
+  package // Explicit package
+});
+```
+
+#### Registration specific APIs
+
+Several owner APIs are specific to registrations. Without the introduction of
+absolute specifiers, several of these APIs have ambiguous cases that arise once
+package and source are added features to the container. The following changes
+are suggested:
+
+**unregister**
+
+This API accepts a fullName and removes an entry in the registry and clears any
+instances of it from the singleton caches (including the singleton instance
+cache).
+
+This API has ambiguity without large changes. A resolution will always win over
+a registration. What if you resolve a factory with local lookup,  should
+unregister clear all resolved items with the passed `fullName` regardless of
+`source` including their singleton instances? Only the cached item with the
+`fullName` and no `source`? Only if the cache is for something that was in the
+registry and not resolved?
+
+We lack a canonical representation of some factories after adding source and
+package features. For example "main" export from an addon can be referenced by
+several different lookup paths (with an explicit package, with any of several
+sources). In the future, we're considering absolution specifiers the path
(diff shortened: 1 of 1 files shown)


webark:
Since `use` is a module import statement in a handlebars file, why not just use `import` as the reserved word?

webark:
That makes sense to me. Also, if you used `import` you’d want to put in the fully qualified path, rather then just the name of the component, which would get annoying and overly verbose. Was just concerned about introducing something that is so similar but yet subtly different could lead to confusion and “that’s clever eyerolls”

But i think the amount that is different is easily argued as a need for another word, with using a similar convention.

mixonic:
@rmmmp I'm not sure how to read this statement:

> So the store is saved from the proposed API. Did I understand that properly?

When crossing any package boundary (via templates or JS) you must specify the package. For example I suggest an Ember Data API of the following:

**In an app, explicit package** (in RFC: [Explicit packages for service injections](https://github.com/mixonic/rfcs/blob/mu-packages/text/0000-module-unification-packages.md#explicit-packages-for-service-injections))

```js
// my-app/src/ui/components/foo-bar/component.js
import { inject } from '@ember/service';

  store: inject({ package: 'ember-data' })
});
```

**In ember-data itself, implicit package** (in RFC: [Implicit packages for services](https://github.com/mixonic/rfcs/blob/mu-packages/text/0000-module-unification-packages.md#implicit-packages-for-services))

```js
// ember-data/src/services/something.js
import Service, { inject } from '@ember/service';

export default Service.extend({
  store: inject() // babel transform will add the `source` arg for you
});
```

Many people have commented that they rather like `store: inject()` without a package. This RFC does not propose a way that a full MU version of Ember Data could continue to offer that API. If Ember Data *really* wanted to continue to provide `store: inject()` under MU layout, it would need to do one of the following non-standard things:

* Add a broccoli transform that adds a file to the application's `src/` tree at `src/services/store.js`
* Have users manually install a re-export file at `src/services/store.js`
* Have the post-install hook add the file
* Use one of those three methods with an initializer to do some fancy stuff
* Use a babel transform to add the `package:` arg to all uses of `inject`

Basically it would be non-standard and I don't suggest that there is a good solution here.

luxferresum:
Can I pass a string to the `use`helper? Like `{{use string from bla}}` and then `bla` being dynamic? It doesnt sound like it is possible.

Is there a way to basically `{{use}}` all components from a package and then, at runtime decide which one to use? Dynamic lookup can be *very* useful. And I *do* understand that some static `{{use}}` is needed, but I think it should be possible to use an entire package and then dynamically select the component.

rmmmp:
@mixonic got it. I think I was the one who misunderstood things. This paragraphs has a lot of fancy words in it that it confused me 😛 

> The concept of a data store is a generic software design tool. That Ember Data has claimed the word "store" so effectively in the Ember user's mind that another library providing a "store" seems unimaginable limits the potential of the framework.
>
> This RFC proposes no design-level mitigation for the extra argument injecting a store. Ember Data itself could continue to support store: inject() via special-case broccoli tooling should it choose to.

mixonic:
> Can I pass a string to the `use` helper? Like `{{use string from bla}}` and then bla being dynamic? It doesnt sound like it is possible.

Correct, this is not suggested. The syntax section for `{{use}}` (under [The `{{use}}` helper](https://github.com/mixonic/rfcs/blob/mu-packages/text/0000-module-unification-packages.md#the-use-helper), scroll down) specifically calls out the *ModuleSpecifier* as a *StringLiteral*. "The module unification package from which a component or helper is being imported. Usually an NPM package name."

I would like these import statements to be hints for improving Ember's template compilation and application packaging. If the source of the import is dynamic we lose lots of information.

> Is there a way to basically `{{use}}` all components from a package and then, at runtime decide which one to use?

Sure, you could explicitly import all the components from the package and the use the `{{#if}}` helper to decide which one was rendered. That would be explicit in the template, but still be dynamic at runtime. 😉 :shipit: 

What is *not* provided are two things I believe you are implying. **First**, something like JavaScript's "import all named exports" feature:

```js
import * as name from "module-name";
```

This would then perhaps allow you to invoke a component with `{{component name.Foo}}`. The problem is that Ember's resolve is traditionally lazy, so even at build time we really can't be sure that 100% of loadable things are in the module map. At build, we can't guarantee the `*` at build is really the same as `*` would be at runtime.

Sidestepping this and requiring statically named imports seems like a good idea for now. Maybe in the future Ember can decide build-time imports are reliable at runtime, but it just isn't true today.

**The second thing** implied is that this renders the component `foo-bar`:

```hbs
{{use foo-bar from 'my-addon'}}
{{component computedPropertyWithStringOfFooBar}}
```

However the component helper doesn't do this today. Strings passed to the component helper are resolved in the current package. A string passed to the `{{component}}` helper does not mean "check for a local variable with that name".

There was discussion about if the component helper should be changed to support this. I decided against advocating for that change for a few reasons:

* It is not actually a backwards compatible change. Today `{{#with (component 'foo-bar') as |c|}}{{component 'c'}}{{/with}}` would render nothing. If we made this change, it would render the `foo-bar` component. How big of a breaking change this is would be up for debate.
* It is a weird change to make. The `{{component}}` helper is already pretty unique in the Glimmer template language. This change would also give it a macro-esq ability to take runtime information (the string) and combine it with build-time information (the variable names).
* Although this RFC does not allow the "dynamically invoke anything from another package" approach, it certainly continues to allow apps and addons to have dynamic `{{component}}` usage in their own package-space. So if an addon is indeed highly dynamic, it can certainly `{{yield (component someDynamicString)}}`. The available components for dynamic invocation would be limited by the template with the `yield` statement, and only the components from that one addon would be defeated in tree-shaking. An addon can provide a highly dynamic API if it desires.
* I'm not convinced there are many (any?) real-world cases where you want to dynamically invoke some component from another package. If there are any, the workarounds of `{{#if}}` statements and using `{{yield}}` to pass a dynamic component out of a package seem like ample workarounds. I would welcome another perspective here. Happy to kick around practical cases in `#st-module-unification` on community Slack.

rtablada:
> {{use}} is required to bring a component or helper from another package (and not from a directory or file) into the template's scope as a symbol

This is limiting (not in a good way).

For local packages there’s a use for aliasing or renaming components. I think that having the `use` helper would be a great fit for this.

Having the `use` helper work like import for modules in browser by requiring absolute paths makes sense and makes local import and dependency import feel a bit more natural.

For clarification I don’t mean that all local component declaration should require a `use` statement. But I’m only recommending that `use MyComponetnt as NewAlias from 'my-app'` should work.

luxferresum:
@mixonic thank you very much for clarification.
I agree that building a wrapper component for dynamic component calls is totally feasible. So the package would provide a component with this template:

```hbs
{{component @name}}
```

I totally agree that this works for now. Can't wait for MU now.

toranb:
@mixonic I'd love to see a migration path example that takes into account "shipping with sensible defaults" as I've described in great detail below 

https://github.com/ember-cli/ember-cli/issues/7680

Kerrick:
It seems that as specified, `prelude.hbs` could include arbitrary text and/or HTML, and that could lead to that text rendering _all over_ apps. Imagine this `prelude.hbs` file:

```hbs
{{use Select from 'ember-power-select'}}}
```

Now suddenly the user has `}`s rendering in dozens of places across the screen.

Perhaps it would be best to say that the _only_ contents `prelude.hbs` can have are `{{use}}` statements, and any text nodes, DOM nodes, or component/helper invocations cause a compilation error?

rwjblue:
Great point @Kerrick! In addition to limiting `prelude.hbs` to `{{use` statements and text nodes, I think we should _also_ specify that white space text nodes before and after `use` statements is stripped.

webark:
if the prelude is limited to “use” import statements, does it being a special one off .hbs file cause more confusion then it being a config block or js file?

Also, where does this file live? can it be scoped to a “level” i.e. the values in `two/level/prelude.hbs` get added/overwritten by `three/more/levels/prelude.hbs`?

It’s mentioned that only one use can be done, and that “Once an addon has claimed a name in a prelude the app will not, in its own templates, be able to write {{use}} for that name without a re-assignment.” Shouldn’t the use from the individual template over write the prelude rather then the other way around? Isn’t that the point? Otherwise, if an addon developer adds a generic item to the prelude, we are kind of back in the same boat we currently are if we can’t override it on an individual basis. Or did i read that passage wrong?

rmmmp:
I think the prelude.hbs has a potential to cause confusion with application.hbs

How should we describe things so this wouldn't confuse newcomers?

dfreeman:
This occurred to me over [on the Modifier Managers RFC](https://github.com/emberjs/rfcs/pull/373#discussion_r216442668) when I read a `setModifierManager` snippet there, but it equally applies to `setComponentManager`, which already exists.

In the context of this RFC, how do these functions work when referencing a manager from an addon? Would it be something like `setComponentManager('sparkes', { package: 'sparkles-component' }, SparklesComponent)`?

cibernox:
I like to explicitly import the components and helpers I want to use in my templates, so I'm generally in favor or this. 

I even _think_ that we could take this one step further and also require to explicitly import even components from your very app, but I think this is a good compromise for now. If next year we realize that explicitly importing components from your app can help with tree shaking it seems that it wouldn't be terribly hard to opt-in into a "strict mode" that enforces explicit imports. At the end in React apps imports are explicit and it's not a big deal.

Also, this could open a door to lazy-load components when they are imported with a very similar syntax some day in the future.

webark:
> Also, regarding the prelude, it's supposed to be used for components or helpers that are used suuuuuper often

My main concern is that there is no way to enforce this, and if an addon developer can add all of their components to this file, which definitions take presedent over your own in your app, we are basically where we started.

cibernox:
@webark Personally other than the helpers in ember-truth-helpers, I struggle to find any other example of something I use sooooo often that explicitly importing it would become annoying. Maybe `{{href-to}}` comes to mind. I wouldn't be too pissed if the `prelude.hbs` doesn't exist at all if the truth-helpers are moved to core. 

But I can see people feeling otherwise. As long as addons cannot add things to your prelude implicitly (they obviously can if they use the `afterInstall` hook to edit the file) I can live with it.

chriskrycho:
Yeah, I'd argue that the behavior should be both *discouraged* and *explicit*; as long as it is both of those things, the user can trivially undo it if they so choose.

webark:
> Once an addon has claimed a name in a prelude the app will not, in its own templates, be able to write {{use}} for that name without a re-assignment.

As i read that, if an addon author defined a “not” helper, that would be something you could not redefine, and arevstick with there’s. And if two addon authors do the same, not sure what would happen.

I get that this behavior should be discouraged, but without any enforcement, you lend yourself to the some of the same issues that drove the creation of this addon.

I the pretending of use statements to some, or all template files, is something that if the community decides is a priority, a extra community addon could be created that would do just that, and people could opt in that way. I feel (personally) that the “prelude” feature shouldn’t be included as part of this effort.

And if we hold to local helpers and components don’t need to be “use”d, then you could just create your own “t” helper that just inherits from i18n’s for example.

chriskrycho:
Right, but the explicitness of the prelude is the other save: if you don't want something that an addon added to your prelude – since it's in the actual file `my-app/src/prelude.hbs` on disk – you just remove it. Delete that post-install-added line from the file and it's gone. Or remap it to what you *do* want it globally available as, or whatever else. The fact that it's just a file in your app's `src` folder means the addon can add it automatically, but it can't _require_ it to be there.

webark:
hmm.. I did not read it as lines that where continually added to the consuming apps prelude. That makes more sense, and would be more manageable.

Does this concatenation happen automatically? or would someone have to write there own custom post install hook in their addon to opt in.

chriskrycho:
Yeah, re-reading I see where the ambiguity comes from there. It might be worth a small update to the text to indicate that an addon can only do that by running a post-install hook that updates the `src/prelude.hbs`.

mixonic:
@rtablada you said:

>> `{{use}}` is required to bring a component or helper from another package (and not from a directory or file) into the template's scope as a symbol
>
> This is limiting (not in a good way).
>
> [snip]
>
> For clarification I don’t mean that all local component declaration should require a use statement. But I’m only recommending that `{{use MyComponetnt as NewAlias from 'my-app'}}` should work.

Your suggested invocation of `{{use}}` would do what I believe you are expecting. It would make the component from `src/ui/components/MyComponetnt/component.js` available as `<NewAlias>` in that template.

`{{use}}` is **required** to bring a component or helper from another package into the template's scope, but you **may**  use it for aliasing a component or helper that was implicitly in scope.

Did you have a deeper concern here? I'm pretty strongly against "use is just importing things from disk". This design intentionally keeps resolutions shallow, you're only importing top-level components or helper from a package. If your codebase feels like it requires importing from an arbitrary folder, I think that is a signal telling you to create a new package for those items (in the current design, a new addon).

@sandstrom I think @bryanaka has is right. My justification for not using `''` is that you are indeed populating the value of a symbol. For example in JavaScript we would find the awkward syntax to set and read a variable value.

```js
let 'foo' = someFun();
anotherFunc(foo);
```

The contents of a string should not have any relationship to the symbols current in scope of the programming language.

@toranb and I talked over some of his concerns. I think the high level summary is that ember-redux has addon code that relies on addon `app/` directories being merged into `app/` directories, granting the addon code access to arbitrary paths on disk without knowing the app's name.

A note from our conversation:

> It sounds like there are two design constraints common to addons which ember-redux attempts to reject:
>
> 1) Explicitly coupling your addon to the filesystem layout of an app is bad. Usually Ember's DI system is a layer of abstraction for this, and has in the past allowed us to move from globals namespace to modules to pods and in the future a new filesytem. Addon's that couple to what is on disk will create pain or at least inconsistency for their users.
> 2) All addons should basically play by the same rules. For example if I use ember-simple-auth, the ideas around where files are found on disk for working with that addon should be similar to ember-redux. In MU those rules are already being bent, as we now allow addon's to modify the collections and types expected by MU (and collections/types can have different configs for local-lookup, co-location, etc). I do think there are some things that have co-locations rules, like styles, that don't really play with this system. Tests are maybe another example. Those tend not to be application JavaScript though.
>
> If we agree ember-redux rejects those constraints, I see two helpful technical outlets. I'm not sure if they are adequate, you will need to noodle on them.
>
> 1) Broccoli still exists. You can move a file into the app's src/ tree at build time that does some magic. Is this weird? Oh yeah. But you are trying to break convention so that seems like proportional pain.
> 2) factoryFor, lookup etc can reach into the app code, or into your addon's code.

@Kerrick said:

> It seems that as specified, `prelude.hbs` could include arbitrary text and/or HTML, and that could lead to that text rendering all over apps.

We discussed this at our strike team standup, and there is strong consensus that `prelude.hbs` should only permit `{{use}}` and that whitespace should be disregarded. This is a notable change that needs to be mentioned in the RFC.

This will result in `prelude.hbs` being a special kind of template which will throw when you use any non `{{use}}` helper at compile time. It will also disregard whitespace. From a teaching perspective, I think the mental model of "this runs before each template runs" still makes sense. As learners outgrow that mental model they will quickly run into exceptions that we can use to point at useful documentation.

@webark the `prelude.hbs` lives at `src/prelude.hbs`. There is only one prelude file for a give package. There is no "stacking" of multiple preludes. `{{use}}` statements do not override other statements, instead an exception is thrown if you attempt to add the same symbol to a template twice. Devs can use the aliasing features of `{{use}}` to rename a conflicting import, or if they find themselves doing that too often they can alias the original name in the app's `prelude.hbs`.

Thanks for catching this text in the detailed design for `prelude.hbs` section:

> Once an addon has claimed a name in a prelude the app will not, in its own templates, be able to write `{{use}}` for that name without a re-assignment.

It was from a previous iteration of the RFC, and I will remove it from this version! Totally an oversight, please disregard it. There is no stacking of multiple `prelude.hbs` files in anyway.

@cibernox @chriskrycho I think you both got dragged into a confusing conversation with @webark over this mistake. Please take note.

@rmmmp `application.hbs` in module unification is at `src/ui/routes/application/template.hbs`. I hope that the very different placement of `src/prelude.hbs` on the file system is enough to differentiate it for users. Additionally we're going to add to the RFC a compiler behavior limiting `prelude.hbs` to only include `{{use}}` statements, which means anyone confused about how to use these two files will quickly run into helpful errors.

@dfreeman that is a pretty good question. I'm wary of this RFC having the responsibility to figure out how every module in MU works in this world. In fact I don't think your question about how managers work relates directly to this RFC. Let me kick it around in chat and perhaps we have a good simple answer.

rtablada:
@mixonic thanks for clarification. I wanted to make sure it was also available for local project aliasing.

mixonic:
@dfreeman ok, so that API is important to rationalize in light of this RFC since it uses container lookup strings (which I did not realize it did). I think we can addresses it per your suggestion of passing `package` as an option, although it likely means we should reserve the second argument for options at all times. Further discussion at: https://github.com/emberjs/rfcs/pull/373#discussion_r217794940

knownasilya:
Addons will probably want to auto add something to a `prelude.hbs` on install or addon generator, so maybe a mention of ember-cli API that would make that possible, otherwise each addon author might implement their own and that might not be great.

This api should handle collisions and warn the user that they should manually add/rename the {{use}} statement that was being added, instead of waiting for the user to start the app and get that error and backtrack.

webark:
Since the prelude is so specific, can’t have whitespace, can’t have anything but use, can’t have same names, etc.. is there a reason this needs to be a `.hbs` file?

If we gave it a more general name like .. global-component-aliases.js or something, and made it a plain js array with values like
```
[{
  name: 'my-component',
  alias: 'MyComponent', (optional alias)
  package: ‘all-my-components’
}]
```
you wouldn’t have to use an ast to parse it, wouldn’t mislead people to think they could put other global values in there, it would be less “special” being a plain js file, (could easily just validate the export), and have a name that is more descriptive for other people to understand what it’s doing without needing special information.

Kerrick:
Perhaps then it would be a good idea to allow HBS comments in the prelude file? That way the pre-generated file can have an explainer comment with a link to the docs. Plus, that would allow users to “section” their prelude better.

webark:
> The reason I argued for hbs is that it means there's not a second API to learn.

@ef4 ya. it’s not a large api (2-3 properties in one file name/property an the app). you could carry that same api into the component.js file, and have more of an explicit importing and setting premis, but that’s diverging too far from the RFC. 

what about the name prelude..? I still feel it’s a little too clever/cutesy.. and would prefer something more specific. That could be just me though. I just imagine the eye rolls from my coworker’s when i explain it.

BTW.. i think this whole notion explicitly importing is a huge improvement 👍 it’s just the prelude part that seems weird. to keep to the spirit of keeping core lean, and extending off of it, could the prelude part be something that is implemented in a separate package rather then directly in core?

ef4:
> what about the name prelude..? I still feel it’s a little too clever/cutesy.. and would prefer something more specific. 

We didn't make the name up, it's been in very long use by other languages. See for example [Haskell](https://hackage.haskell.org/package/base-4.11.1.0/docs/Prelude.html) and [Rust](https://doc.rust-lang.org/std/prelude/).

webark:
@ef4 Rust and haskell aren’t two very common languages, so many people who do have familiarity with a variety, might not have come across that. maybe we could add that too the RFC so that it’s more clear that we are basing this off of similar conversations used in other languages?

webark:
> I still feel it’s a little too clever/cutesy.. and would prefer something more specific. That could be just me though. I just imagine the eye rolls from my coworker’s when i explain it.

Sorry for phrasing this in this way. It was a flippant remark, and wasn’t honoring the time or thought that was put into this proposal or this process. I apologize.

sclaxton:
Not sure if this has been brought up offline or online before, but is it possible to, instead of introducing new hbs syntax/semantics, leverage JS imports to accomplish the same goals? I'm not sure I have enough context on implementation details to even posit a detailed design for this sort of thing but my "dream" design for something like this would be something like:

```
// my-app/src/ui/components/my-component/component.js

import Child from 'some-addon/components/child';

  ...
}

MyComponent.use(Child)
```

And have that somehow register the imported component to the template namespace. This kind of allows hbs to remain "just a template" and puts some sugar over explicitly resolving the names with the resolver.

I'm fully ready to be schooled on why this is not a good idea / not possible implementation wise haha. I'm not even sure I like this more than `{{use}}` + preamble.hbs, but just wanted to put it out there in a best effort to keep hbs as minimal as possible.

rtablada:
@sclaxton I think that a mechanism for adding components via JS is a good idea in addition to the `use` syntax for templates, but from a syntax perspective I think having this declaration at the top of the template make a lot of sense from being able to clearly know what's going on.

When in a template seeing:

```
{{use Foo from "my-package"}}
```

This would make it pretty clear where `Foo` is coming from when I see this two lines later:

```
<Foo @ar1="myValue" />
```

Having this local component naming and importing be a JS syntax concern means jumping back and forth to understand where the `Foo` Component is coming from.

mixonic:
@sclaxton and @webark I think @ef4 and @rtablada did a good job of summarizing why we think using a template syntax for this is helpful. I'll add the general statement that, IMO, we have created a poor design if app authors regularly need to consider the contents of other files to understand the one they are reading. In general.

@sclaxton said:

> @rtablada seems like most {{use}} invocations would be in the the preamble.hbs anyway, so seems like you’d have the same “file jumping” problem, no?

I don't believe this is accurate. I expect an application to have one or two global level imports. Something like `{{t}}` for translation or `{{and}}`. I don't expect that every time an app author installs an addon they will immediately want to add it's public components as a global. For example, we want Ember Power Select to use component names that make sense like `<Select>`. If every addon's public API is aggressively added to the `prelude.hbs`, you won't be able to read templates and understand your app, and you'll end up with naming conflicts. Basically, you'll end up with something like the common shared namespace we have today via `app/` merging.

I expect that if you consider what percentage of all your app's templates use addon X, it is a very low percentage. The `prelude.hbs` is an escape valve for things like translation helpers that violate the norms. Although it necessarily takes up a notable amount of our design head-space in the RFC and our discussion, it is not something a developer should need in their first week with the framework.

@knownasilya said:

> Addons will probably want to auto add something to a prelude.hbs on install or addon generator, so maybe a mention of ember-cli API that would make that possible, otherwise each addon author might implement their own and that might not be great.

And he suggested an API. Because we want to discourage addon authors from changing the app's `prelude.hbs` without a serious weighing of the implications (such as stomping any same-named components already in the app with the addon's components!) I do not suggest such an API in this RFC. I'm not against us either a) implementing an addon which provides a nice API for this or b) if there is demand designing and adding an API to Ember CLI itself. I think @rwjblue was thinking we could do "a" as we port ember-intl and truth helpers to MU.

sclaxton:
@mixonic I'm not sure I fully grok how jumping from `hbs` to `js` to see how component names are resolved is any different/more burdensome than the current state of things where you have to go to `js` to see how a computed property or action is implemented...

chriskrycho:
Two points there:

1. It’s definitely easier for tooling!
2. More importantly, there are plenty of times (and likely to be far more of them going forward!) where there is no backing JS file, and there are performance wins to be had when we know there is no dynamic JS backing a given template (Glimmer can do more optimizations in that context). Requiring JS lookup would mean *any* component which wanted to `use` a component from another package would have to have backing JS.

sclaxton:
@chiskrycho definitely solid points. I’m mainly worried about this hurting readability in a big way as far as noise in templates. It’s already hard to grok more complex templates currently, can’t really imagine the fatigue involved in the templates I have in mind where every component is annotated with `{{use}}`

chriskrycho:
@sclaxton One thing I’ve done often with proposed changes which feel like that is to just take one of those components and rework it as if the proposed change were true. I did that with AngleBracket and named arg invocation, for example, and it took me from slightly skeptical to completely sold in about half an hour.

Obviously your mileage may vary, but I’d encourage you to do that exercise with this proposal and one of the templates you’re worried about, and then report back on what you find; it would be useful input to the thread in my view. Assume that you have truth helpers or internationalization in prelude and then only `{{use}}` other external packages and see how it comes out. (And if your code is shareable, putting it here as an example would be 💯 .)

rtablada:
@sclaxton 

> It’s already hard to grok more complex templates currently, can’t really imagine the fatigue involved in the templates I have in mind where every component is annotated with `{{use}}`

1. The `use` keyword only needs to be used once per template per component (similar to `import` in a JS module.
2. Local application components still are looked up using the standard resolver (although as @mixonic and I discussed in some of my earlier questions, the `use` helper can be used as a local alias or for explicit clarity if your team desires that)
3. For widely used 3rd party helpers and components the `preamble.hbs` can be used to your teams desire to reduce the `use` keyword
4. It does add lines of code, but from experience in multiple languages locally bound explicit `import`/`use` statements for third party modules helps a lot for readability and understanding of what's going on.

---

Another thing that hits when talking about using pure JS based imports is the idea that in Ember templates aren't always tied to a single component JS.
Let's say you have a `MyThing` component and a `MyModifiedThing` component which share the same template/layout with different JS logic.
These can be something like:

```js
// my-thing/component.js
import layout from './template';

  layout = layout
}
```

And:

```js
// my-modified-thing/component.js
import layout from '../my-thing/template';

  layout = layout

  // Stuff that makes this one a bit different
}
```

If component importing and aliasing was in in only JS.
Then the same component imports would need to be duplicated in both components and you add a heavily layer of coupling.

This is actually a problem in Vue apps that use importable `template` modules (instead of Single file components).
Adding another component invocation requires updating across every JS module that uses the shared `template`.

sclaxton:
I've been thinking about what rubs me the wrong way about this change for the last few days and why I feel like adding this syntax would make templates a little harder to reason about, in particular for beginners. I've come to the conclusion that it's mostly because `{{use}}` introduces a micro syntax which overloads the current `{{helper}}` syntax to make it look/function like JS imports, but not quite (as per the RFC `use` is a subset of JS import functionality wise). It's this "almost like JS  import but not quite" that's pretty confusing IMO. The syntax kind of pretends that templates are module-like and I'm skeptical about whether it's helpful to think about templates in this way (e.g. templates can't export anything, templates can only "import" helpers and components, not arbitrary values). The real goal, as laid out by this RFC, is to disambiguate symbols in a template as a reference to a particular component in a particular addon. I wonder if we can accomplish this goal without treating templates like pseudo-modules and without adding module-like micro-syntax.

One way to do this could be to introduce syntax for explicit "compiler hints", i.e. statements that are syntactically different from helpers so they can explicitly exist to disambiguate symbols. A possible candidate syntax could be something like:

```
{{decl! 'addon-name/components/my-component' as MyComponent}}

<MyComponent />
```

Essentially the `{{decl! <string-literal> as <symbol>}}` syntax would allow you to map a full component path to a symbol in the template. This has the double advantage of not using vanilla helper syntax for something that really is just a compile time mapping while also not treating templates like modules.

cibernox:
@sclaxton I might see why you don't like that `{{use}}` looks like a helper. I don't share your concern but I see where it comes from.

I fail to see tho how the alternative you propose is any different in semantics than `use` except in the syntax. The both "import" a helper/component for a path and gives it the alias you want.

sclaxton:
@cibernox My intent in the alternative is to avoid having the same semantics as `import`, i.e. the same semantics as modules in general, since templates are very different from modules. To me the semantics of "lookup component from a location and bind to a symbol" is pretty different than "import a component into this template for use", but maybe I am just splitting hairs here?

For example, if we didn't care about being able to statically analyze the binding of symbols to components we could just use inline `let` and the normal component lookup:

```
{{let MyComponent (component 'my-component@addon-name')}}
```

The alternative I proposed is essentially just aiming to be a statically analyzable version of the above.

mixonic:
@sclaxton You've raised a number of issues, and I think I would need you to narrow down your criticism to something specific instead of simply suggesting an alternative in order to respond. Even in your alternatives you've iterated through a few suggestions, and I'm not sure which warrants a response :-/

I think the closest statement to a thesis was:

> I've come to the conclusion that it's mostly because `{{use}}` introduces a micro syntax which overloads the current `{{helper}}` syntax to make it look/function like JS imports, but not quite (as per the RFC use is a subset of JS import functionality wise).

I would not categorize `{{use}}` as a micro-syntax because it is valid HBS as we write it today. It explicitly lives within current syntax.

However I agree that the semantics are unique. For example it introduces a symbol into the template. This seems like an easy hurdle though, we've been discussing an inline form of `{{let}}` for years now. Nobody has really pushed back on this.

I agree that `{{use}}` is similar to `import` but not the same. However I'm not convinced that changing `{{use}}` to `{{decl!}}` would improve anything in this regard. At the end of the day we're building a namespacing/import system here. Whatever we design we author will both benefit (easier to learn) and suffer (confusing differences) in its comparison to other systems.

> This has the advantage of not conveying implicitly that addons and templates are both somehow being module-like, while keeping some of the convenient syntactic elements.

"module-like" only arises in comparison to `import`. `{{use}}` imports a component from a package. Nothing is named "module" in this system. The source package in a `{{use}}` statement is explicitly a package. You cannot `{{use Foo from 'some/module/on/disk'}}` which I agree would make for a lot more confusion.

---

Making some edits to the RFC now, I'll comment with a summary after.

knownasilya:
One thing I haven't seen mentioned is anything about how this affects the possibility of dynamic `{{use}}`. I know that it's not part of this RFC, but I think some reference to how the two might interact or if anything from this RFC might (or not) be a blocker for a dynamic `{{use}}`.

Two scenarios come to mind for a dynamic `{{use}}`.

1. A bundle shared between multiple engines.
2. Splitting bundles based on things like media queries, ember-responsive or https://flexi.readme.io/docs/viewport-specific-templates

mixonic:
Updates per review:

* `{{use}}` and package options are available to files in `src/` and `app/`.
* Whitespace will be stripped from `prelude.hbs`.
* Only `{{use}}` helpers are permitted in `prelude.hbs`. Anything else throws.
* Make clear that actual concatenation may not occur in the implementation of `prelude.hbs`.
* Remove an incorrect line about `{{use}}` statements over-riding each other: That was left over from a previous draft and was misleading.

btecu:
@mixonic, referring back to a comment of yours about the `component` helper. We currently have a list of addons that provide some special form components and we have some specialized dynamic forms that require us to use the `component` helper to dynamically render the specific component. Will this local `use` work for that or what's the story there?

cibernox:
@mixonic I understand than `{{use}}` attempts to implement a subset of the static form of import, but I wonder if, even making clear that it's outside of the scope of this RFC, an async version of `use` could exist, much like the runtime `import().then` statement in modern javascript.

If that is even a posibiliy reversing the syntax and having a block is one option:

Some example that comes to mind is loading engines:
```hbs
{{#if user.isAdmin}}
  {{#use 'ember-admin' as |admin-panel|}}
    {{mount admin-panel}}
  {{/use}}
{{/if}}
```

A non-block form could work but it would make easier having a loading state.

mixonic:
I've pushed another commit with example steps for migrating Ember Data and Ember Power Select.

---

@btecu I'm not sure exactly what case you are asking about, but I suspect it is this: You have an addon which provides components and you would like to use `{{component someDynamicString}}` in your app (or another addon) to invoke those components directly.

And the simple answer is **no**. You cannot invoke a component imported via `{{use}}` by any dynamic use of the `{{component}}` helper. Static use would be fine, but that isn't what you're looking for to satisfy the above case.

You can read https://github.com/mixonic/rfcs/blob/mu-packages/text/0000-module-unification-packages.md#impacts-of-use-on-the-component-helper for more details on this and examples.

There are good reasons for us to not support this, including the compile-time analysis it unlocks so we can tree shake components across package boundaries despite the `{{component}}` helper being present.

For your use case there are two pretty decent workarounds, IMO.

1. **Use `{{#if` `{{else if}}` `{{else}}` blocks.** `{{use Foo from 'foo-addon'}}{{#if isThing}}<Foo />{{else}}{{component someDynamicString}}{{/if}}`, etc. `{{use}}` is only needed for components from another package, so if you've just got a few this works fine.

2. **Build an API for your addon that doesn't require dynamic invocation in the app.** You can still yield out components from an addon based on the `(component` helper, so do something like:

```hbs
{{! my-addon/src/ui/components/FormInputFor/template.hbs }}
{{yield (component someDynamicFormInputBasedOnArgument)}}
```

```hbs
{{! my-app/src/ui/routes/edit/template.hbs }}
{{use FormInputFor from 'my-addon'}}
<form>
  {{#each formFields as |field|}}
    <FormInputFor @type={{field.type}} />
  {{/each}}
</form>
```

Since angle bracket invocation doesn't implicitly add an element, there is no downside to using a higher order component to yield the dynamic bit. In fact I think this is basically always a better way to provide a highly dynamic API to an end-user app author anyway. Much more descriptive, trivial to document/test/re-use, refactoring-friendly, and has access to local lookup to isolate those yielded components.

@cibernox I've got nothing against a lazy version of `{{use}}`. At some point all of Ember's rendering is going to end up being async anyway (I hope). That said I'm totally focused on unblocking our progress on the filesystem issues, and I agree this RFC is probably not the spot for that design.

@knownasilya I think the same goes for any dynamic use (I presume you mean something like `{{use Foo from someDynamicString}}`). I'm not sure I see the utility in adding that API, but regardless it seems additive to this RFC.

---

Thanks for all the thorough reads btw ❤️ I really appreciate the feedback and discussion here!

luxferresum:
Again, how would I override addon services in the future?
For example [here](https://github.com/runspired/rfcs/blob/ember-data-custom-records/text/0000-ember-data-model-factory-for.md) such an API is proposed. But how could I override an addon service to implement hooks if the service is namespace specific?

sclaxton:
@mixonic I agree that I was essentially saying two separate things: 1) I think `{{use}}` statements could have a syntax that sets them apart from helper statements in hbs, signaling them as statements that do not have run time output/effects and 2) I have trouble with `{{use}}` semantics because it does not jive with my previous mental model for component passing/invocation.

For (1), I feel like adding a syntax for non-runtime statements might be useful as a general concept. For example, for `i18n` it's often desirable to colocate translatable string definitions in the template in which the translated string will be used, for example:

```
{{t-def! "A string to be translated in {:lang}" key="an_i18n_string"}}

<p>
  {{t "an_i18n_string" lang=translatedLangName}}
</p>
```

In the above syntax, it's more clear that `t-def` is compiled away to some string translation repository and is not really a helper but rather an annotation that has no run time output/effects. Another use case for this syntax could be user-defined syntactic macros:

```
<div style="margin: {{calc! offset + width - magicNumber}};"></div>

{{! expands to... }}

<div style="margin: {{plus offset (minus width magicNumber)}}"></div>
```

For (2), I believe I've finally fully figured out my trouble getting my head around the implications of `{{use}}`'s semantics. First off I'll apologize because I clearly glossed over a key bullet point in the description of `{{use}}`'s semantics that would have saved some discussion if I had internalized it immediately: `{{use}}` introduces a symbol that is bound to a "component factory", i.e. "closure component", etc. More succinctly, this RFC proposes that components always be first class values in templates. I did not explicitly internalize that implication at first and just had a sort of feeling that something big had changed but I couldn't exactly express it. Up until this RFC, I would say that components have been equivalent to C functions and function pointers. You can call them statically or explicitly convert them into something you can pass around via `(component)`. Now components will be passable values everywhere it seems. IMO this should be called out super explicitly as like a main section of this RFC. This changes how we think about components and addons in a subtle, but I think pretty important way. From now I've always thought of addons as purely namespaces that export component *names* to be referenced statically and evaluated in templates. Now we have addons that implicitly export values, namely component/helper factories. TBH I've always been confused about component factories, i.e. the return value of `(component)`. What are they?? They're basically undocumented in any explicit way and pretty fuzzy as a concept, evidenced by the fact that people can't agree on a name (I've heard component factories, closure component, component instances, curried components, partially applied components, etc...). If we're moving towards this concept of component factories as the only method of component/helper invocation then I would really want some explicit/canonical documentation of the concept.

cibernox:
Now I see @sclaxton point. I still don't agree with the idea that we need a different syntax, but it would be a good idea to clarify that an imported component is not a value you can pass around like if you used the let/with helper and passed a contextual component.

My understanding is that you can't import a component this way and pass it as a value to other components, correct?

knownasilya:
@mixonic by dynamic I meant components coming from a bundle/script that hasn't been loaded yet. Hence the two examples I made in my first post.

cibernox:
@sclaxton I re-read it and I think you're right. An imported component behaves like a closure component and can be passed around as such.

That was unexpected to me but on a second though it could make some composition patterns I use nicer (avoiding passing strings that get resolved with the `component` keyword). This could also help minimize one of the caveats of `component`, which is that since it receives a string you can't discover in build time what components are unused. With this in *some* situations could be known.

samselikoff:
I was curious if a YFM (YAML Front Matter)-like solution was ever discussed. One could imagine using ES6 imports to define local lookup in a template:

```hbs
---
import { PowerSelect, Option } from 'ember-power-select';
---

<PowerSelect>
  <Option />
</PowerSelect>
```

(Apologies if I missed the discussion somewhere.)

knownasilya:
I could see a YFM solution that has JS in it becoming confusing. Where does the JS end, can I define my whole class in the template (single file components)?

samselikoff:
@knownasilya We could start by only allowing `import` statements. The benefit is not introducing a new `import`-ish syntax & being able to leverage existing tooling around ES6 imports.

knownasilya:
I think the problem with that is that regular imports allow more features that aren't supported by `{{use}}`, like default imports and relative imports.

chriskrycho:
And what would this import even mean?

```js
export class Visitor {
  // implements a version of the visitor pattern
}
```

```hbs
---
import { Visitor } from 'visitor';
---
```

You have now way to instantiate it, and even if you did it would have no meaning in a template context. You could forbid things which don’t have templates semantics, but then you have to do crazy analysis of what’s being imported. You could just trust users not to do it, but it’s a an extra complication to explain in the docs (“You can import anything, but you can only use things in your templates that follow these rules: …”).

The semantics of `{{use}}` are *new*, but they’re clear and appropriate to their context. Glimmer templates aren’t Just JS; for both good and I’ll they are their own distinct programming language, and are best treated (indeed designed) as such.

samselikoff:
I'm not sure about the analysis involved but I noticed CRA has a nice error message when trying to [render a plain class in the render function](https://codesandbox.io/s/z62pmqp00x):

image

samselikoff:
Also, while it is true that Glimmer templates aren't just JS, they were designed specifically for the problem of declarative rendering. ES6 imports (and more generally the concept of looking up files via filepaths) is a well-trodden path. It would be worth considering how much of it we can build on top of.

chriskrycho:
Thinking about @sclaxton's "What exactly *is* the `{{component}}` helper and what are the bound names of components in Glimmer templates, exactly?" question—

I've long been conceiving of [components as *functions*](https://www.chriskrycho.com/2018/higher-order-components-in-emberjs.html). In that framing, `{{component ...}}` ends up being something like [`Reflect.apply`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Reflect/apply) in JavaScript: a way of applying arguments to the function with an arbitrary context. (Or, if you think of components primarily as classes, `{{component ...}}` is like [`Reflect.construct`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Reflect/construct).) Either way, you have a named reference to an invocable *thing*—a function or a class—and then a tool for invoking it abstractly.

mixonic:
@sclaxton I think a build-time system for translations would be kind of neat, but I disagree it requires special syntax. For example this could be done today with a template AST transform. You mentioned:

> it's more clear that `{{!t-def` is compiled away to some string translation repository and is not really a helper but rather an annotation that has no run time output/effects.

I agree with this statement. However I don't agree the compile-time/run-time distinction is so important to a template reader/author that the distinction needs to be clear in the first place. What mis-reading of `{{use}}` is so concerning that a whole new syntax is needed to prevent developers from confusion?

@chriskrycho's description of components as functions matches my own mental model. Since we're talking about defining a symbol, there is of course some part of variable name vs. value discussed in this RFC.

I call `{{foo-bar}}` a "component invocation" in training and documentation. I would call `{{t.foo}}` a closure component and an "invokable". This is a hard thing to name in JavaScript as well. When you see `foo;` as a variable, what tells you it has a value of a function? Nothing really. I would say in `{{use Foo from 'baz'}}` that `Foo` has the value of a component, just like in `let foo = () => {};` `foo` has a value of a function.

I'm going to review some of the writing in regards to symbols and invocations in the RFC. I'll look for anyplace I can clarify definitions vs invocations, and for any unique behaviors around invocation introduced by the `{{use}}` helper. AFAIK, there are none. A symbol from `{{use}}` should be equivalent to a closure component.

@samselikoff Your proposal is unique, and I believe it would clearly raise several complicated design questions if we wanted to pursue it to completion. Is there something in particular you are reacting to in the proposed design?

This RFC does not propose importing via file name. Instead `{{use}}` imports top-level components from a package (an addon). I believe there are several good reasons for us to continue pursuing that goal:

* Imports done this way are independent of the filesystem, but also statically resolvable given whole-program knowledge at build time. Independence from the filesystem is a benefit that is paying off for Ember right now: We're designing and moving to a better on-disk layout with out trying to re-write everyone's templates. Despite the additional abstraction of a resolver, Module Unification has clearly documented rules for resolution which mean we can statically know what file on disk maps to which module. The abstraction has a cost in that app authors haven't simply written out the string path name for every components they use, but by using a static set of resolution rules we gain a benefit in future flexibility without much loss of performance/treeshaking benefit.
* Ember templates don't require any imports today. If you've written some React in a large codebase, you will have seen files with many, many lines of imports. I believe the ergonomic sweet spot is somewhere in the middle. In Ember you don't know where `{{foo}}` is coming from because is could be from several places on disk or an addon, but in React you don't intuitively know where `<Foo />` is from because it could be anyplace on the filesystem. In both systems you end up hunting. This RFC tries to strike balance by permitting implicit lookups from the current package (basically things close to that template on disk) and requiring explicit lookups for things from other packages (usually things in `node_modules.`).
* I strongly doubt we will build consensus around any design that is 100% explicit.
* Bare file paths as import sources don't provide us a great way to make certain components private. An application would be able to import any file in an addon. This issue exists today in JavaScript.

@luxferresum In this RFC there is no way for an application to override a service provided by an addon. This will be a notable change, and something should be added to the section on migrating Ember Data which details this change.

Ember Data (and less commonly other addons) use a system of defaults that relies on the merged `app/` directory. They place a file of their own in the addon `app/` folder (for example at `ember-data/app/services/store.js`) and if you provide your own file in your apps `app/` folder (for example `my-app/app/services/store.js`) then the app's file will "win".

This system can lead to confusion: For example with Ember Data a reader of an application cannot know what methods on an injection of `store` come from the addon code or from the app code. Another example: If you inject the Ember routing service with `inject('router')` you probably expect the Ember routing API, but if the application has a file at `my-app/app/service/router.js` they may have a completely different service there.

This is a manifestation of @toranb's concerns, and I expect it is more common. The mitigations would include those I described in https://github.com/emberjs/rfcs/pull/367#issuecomment-421431201:

* An addon may: `import config from 'config/environment'; export default Whatever.extend({ store: inject({ package: config.modulePrefix });` and use it's post-install step to install a re-exported service in the app's `src/` directory. This is probably the easiest, best path forward.
* Broccoli still exists, and the addon can still do the same trick with a virtual file in the app's `src` tree. This is obviously an unfortunate solution though.
* `factoryFor`, `lookup` etc still have access to the application's modules and instances from addon code.

This definitely warrants an edit to this RFC clarifying how Ember Data needs to handle this.

samselikoff:
@mixonic Thanks for describing the abstraction & listing out the benefits.

I'm in 100% agreement around a balance between implicit & explicit lookups. I think requiring explicit lookups for 3rd-party code makes total sense.

I suppose the main thing I'm reacting to is that `{{use}}` is a new abstraction for looking up modules, and abstractions are risky and hard to get right. ES6 already has a battle-tested and way to look up modules that's well-understood by the JavaScript community. Using ES6 modules (in some form) to perform explicit lookups does not preclude us from implicitly looking up certain components in our templates.

I think if we could come up with a way to register explicit lookups using ES6 modules, it might let us do way less work, since we'd get to piggy back on existing ES6 tooling. I could see our approach being lower-level and less ergonomic, but more familiar and easy to understand. Others could then experiment with higher-level abstractions (like `{{use}}`) that improve the ergonomics, but an ES6-based API would let us solve much of the resolution issues we have today, and avoid the "magic" of a new abstraction that some have expressed concerns over.

(There are several approaches we could take for an ES6-based API. The YFM example I posted was just a thought. We could also do something in the Component file, a la [Vue's Local Registration](https://vuejs.org/v2/guide/components-registration.html#Local-Registration)).

```js
import ListItem from './list-item';

  components: {
    ListItem
  }
  
});
```

I do understand I'm coming in at the tail end of this, and I personally would like to see us _end up_ with something that looks very much like `{{use}}`, but I do wonder if there's something lower-level that's a smaller jump from what we have today.

Ultimately I just wanted to make sure that an ES6 module-based API was considered and discussed, even if we ultimately move in the direction of making our own abstraction.

sclaxton:
@mixonic I understand the intuition behind component closures, I just think that since it's basically becoming THE primary primitive in the framework that we shouldn't rely on developers having the right intuition and rather should have explicit canonical documentation and a consistent nomenclature in documentation for this concept.

knownasilya:
One of the main things that has been coming back up in the chat is loosing the ability to organize components in deeply nested trees. A few people have brought this up @Alonski, @lougreenwood, @nullvoxpopuli, and sol and Robbo (from Discord) and I think it's worth at least mentioning in this RFC.

I do see the value in nesting purely from an organizational perspective, where things go into categories or areas and make it easier to reason about and navigate components. People that have mentioned this concern have many components, some 400+ I think.

Packages are a sort of solution, but they only support one level of grouping and require a few levels of folders to even get started.

I'm for punting on nested groupings and revisiting in a followup RFC, but I think that needs to be mentioned here if it's the way we are going to do that. In that case, the classic mode should not be deprecated before there a solution for nesting.

jeffhertzler:
From Discord in response to the discussion mentioned above:

> We have 300+ components, a significant chunk are organized in directories like this. Some of our uses can be solved by yielding out the children or only referencing the children from within the parent. Many many others could not. Like others here, we commonly use the directory structure for namespacing.

rtablada:
We use nested components for namespacing. Most of these namespaces match either route groups for parts of our application or could be easily moved to in-app-addons from what I see.

We'd probably migrate slowly for those as we migrate to angle brackets and MU folder structure.

luxferresum:
@mixonic

> An addon may: import config from 'config/environment'; export default Whatever.extend({ store: inject({ package: config.modulePrefix }); and use it's post-install step to install a re-exported service in the app's src/ directory. This is probably the easiest, best path forward.

how can the addon then use this code from addon code? Is there a way to `inject` a service from the app namespace without knowing the app namespaces name? And the same for `factoryFor` and `lookup`?

And the same for components? Is there any way to use a app component from addon code?

----

@knownasilya isn't this already part of the original RFC?

> 2. No slashes in component names. The existing system allows this, but we don't want to support invocation of nested components in Glimmer Components.
> ...
> 8. Local relative lookup for components and helpers needs to work.

Local relative lookup *will* work right?

Alonski:
@knownasilya Thanks for speaking our concerns.
I use nested component structures with a lot of inheritance for certain components. Especially in our UI Component Styleguide addon.
We do this for example: `{{button-/primary/small}}`
I have been living with the fact that  Angle Bracket Invocation will mean we will have to migrate from using `/`. However maybe this can be solved as I and my team like using this nested folder structure pattern. We have multiple hundreds of components.

NullVoxPopuli:
> Local relative lookup will work right?

it already does: https://gitlab.com/NullVoxPopuli/emberclear/tree/master/packages/frontend/src/ui/components/chat-entry :)

`<ChatEntry />` renders `<EmbedsMenu />`

knownasilya:
> No slashes in component names. The existing system allows this, but we don't want to support invocation of nested components in Glimmer Components.

@luxferresum yes it is mentioned, but I think since packages is it's own RFC and it promotes local single level grouping, it should consider the nesting scenario explicitly. I don't think it's enough to refer back to the original RFC because the '/' was avoided in the actual component name, e.g. `<Namespace/SomeComponent />`, but this RFC promotes `{{use}}` + packages as a sort of alternative to that.

The original RFC threw out the idea of grouping all together, but this RFC supports one level of grouping. So far it seems like one level isn't enough for many people.

webark:
i must have missed the nested part. Are you not able to use `{{use}}` with nested components? or only top level ones?

chriskrycho:
Is there a blocker to supporting a `use * from`? Here's the motivating example: we have a family of components which we'll be grouping together into a package and possibly extracting to a shared addon. In the old proposal, I'd have used name-spacing for that: `Olo::Button`.

In the new world proposed here—as the text stands—I can either import those all *individually* in the prelude (and, if I want name-spacing of any sort, `use Button as OloButton` for all of them), or I have to do `use` for them in every template which uses them. For a library of shared UI components, this is annoying. It's not the end of the world, but it's annoying.

If we had the ability to `use * as Alias`, however, we could do `{{use * as Olo from 'olo-shared-components'}}` in the prelude, and this would allow us to then do `<Olo.Button />` in our application.

ro0gr:
Can we support a short form of `use`, which would expose all the public components of the package? 

```hbs
  {{use "ember-bootstrap"}}

  {{!-- we can use all the public package components with a short form use! --}}
  {{bs-tooltip}}
``` 
I think it also can be convenient in prelude.

knownasilya:
I'm totally for the @chriskrycho proposal, but only with a name. An import all from addon, like `{{use "ember-bootstrap"}}` is exactly the kind of ambiguity I think MU and this RFC is trying to avoid/prevent. There is no way to know what is being used or how one addon might affect the imports of another.

samselikoff:
@chriskrycho I think the fact that ECMA modules already support `import * as Alias` is another point in column for an ECMA-based approach, or at least an approach that desugars to ECMA imports.

hemlok:
> If we had the ability to `use * as Alias`, however, we could do `{{use * as Olo from 'olo-shared-components'}}` in the prelude, and this would allow us to then do `<Olo.Button />` in our application.

I just want to echo @chriskrycho's suggestion to allow `* as Alias` as a pattern. It would really unblock building powerful design systems in ember addons for consumption here at Intercom 👍

I think its also worth noting that this goes some way to address your concern @mixonic 
> Ember templates don't require any imports today. If you've written some React in a large codebase, you will have seen files with many, many lines of imports. I believe the ergonomic sweet spot is somewhere in the middle. 

When you can export groups of components with an alias you dramatically reduce the length and number of import statements, while simultaneously adding a lot of context as to which components are grouped and indicate that they work together, or have the same source (I've noticed this pattern  becoming more common in big react project component imports). Allowing `* as Alias` goes a long way to hitting that ergonomic sweet spot 👌

lougreenwood:
Maybe I'm mis-understanding, but would `* as Alias` also allow arbitrary levels of nesting?

chriskrycho:
No, the rules around nesting would be unchanged; the *only* thing I'm suggesting with the addition of `* as Alias` (or the `{{use 'package-name' as Alias}}`, which I also quite like) is being able to provide an alias for the whole set of components exported at the top level, with *exactly* the same rules around top-level exports etc. as you have in the existing proposal.

lougreenwood:
Ok, in that case, +1 me in :) - this could be very powerful for allowing nice semantic names for components in templates. I can see this being very useful for the general Ember teaching story & guides too.

Also, if we begin to see common alias names emerge for component aliases in the community, it's another point towards the "It's easy to onboard onto an Ember project" story.

luxferresum:
@chriskrycho would this work?

```
{{use * as Alias from 'package'}}
{{component (get Alias 'dynamicLookupString')}}
```

mixonic:
> Is there a blocker to supporting a `{{use * as Alias from 'package'}}` pattern?

I think a follow-on RFC would be a good place to add that feature. One downside of this feature is that since `Alias` can be yielded or passed into JS space, you cannot tree-shake out un-used components from a template using this style. The same downside applies to alternative of simply yielding the dynamic component of course, but in that case the downside is opt-in for the addon author instead of the application author.

I'm not convinced there is a case where `* as` is required over having the addon provide a component that yields the hash.

Basically we're considering a `{{use}}` feature giving app authors:

```hbs
{{use * as FormHelpers from '@org/form-helpers'}}
<FormHelpers.Form>
  <FormHelpers.Input />
</FormHelpers.Form>
```

vs. (and there are a few ways this could be done in practice)

```hbs
{{use form-helpers from '@org/form-helpers'}}
{{form-helpers as |f|}}
  <f.Form>
    <f.Input />
  </f.Form>
{{/form-helpers}}
```

@samselikoff said:

> Ultimately I just wanted to make sure that an ES6 module-based API was considered and discussed, even if we ultimately move in the direction of making our own abstraction.

I believe @wycats and @dgeb are exploring a design on that front. I think there are notable challenges but it is plausible. For example we need a system for services (and arbitrary lookups) to come from addon's as well, so perhaps services become class-based injections instead of name-based? That removes flexibility from the DI/container system. And IMO front-matter or some other syntax-switching would look weird once you try to build single-file components. The Vue system still means swapping to a class-based system, and implies the need for a JS file just to reference components from another package: I think that is a bit clumsy. If there is some uncertainty about using `{{use}}` for replacing nested components I expect the Vue system would be even worse for that case.

I've been on leave for a few weeks and missed an informal conversation at the most recent F2F that seems to have made an impact on some core peeps. I'm pretty happy with this RFC (I think clarifications to the Ember Data migration section would be nice but the conversations here have covered most of that). It would probably be helpful if some framework core people engaged here. I'll raise it at the next standup I attend regardless.

dgeb:
Just wanted to follow up on what @mixonic said ASAP. It's true that @wycats and I and several others on the core team discussed the implications of using ES imports for templates at the last F2F in Chicago. The more we discussed it, it seemed obvious to us that we won't have legitimately done our due diligence on this design without fully considering this potential solution. 

At this point we're doing some exploratory work right now, which involves a preliminary design and pulling on threads in the MU design to understand its full implications. We will surface this work as soon as possible if it seems to us like it's got a chance of working well. Either way, we'll report back here so that we can engage with the community, make a final decision, and get some traction on the implementation ASAP.

Thanks to everyone for your patience in what has turned into a very long process.

lougreenwood:
I'm leaving this thought here instead of discord for posterity...

One of my concerns around MU/Packages is that it seems to remove the ability to use directories for organising components. Personally, I like to group many components which are used in a similar context together, for example, I currently have a `/components/forms/` dir where I place all of my forms - this helps me to:

* quickly find a component by type
* easily create new components based on previous implementations which may follow a similar implied interface
* has benefits for introducing new team members to the project - less cognitive load needed to become familiar with the project shape

When I heard about MU I felt quite upset that this pattern is seemingly disallowed because of the no nested components and the switch to local/global components.

But since then I've started on a new project and I've been exposed a lot to contextual components through some work I did on `ember-bootstrap` - this work give me a new perspective and what I think could be a nice way to avoid the problem of seemingly losing nested component organisation in MU-world...

**My idea is to make use of the factory pattern for components.**

So, in my form example above - currently I invoke a form by doing `{{forms/some-fancy-form model=@model}}`

However, with the factory pattern, I would create some new `<FormFactory />` component. The JS for the component could contain a list of defined local form components which the factory can pull from a cleanly organised sub-dir, so invocation would now look something like:

```
<FormFactory @form="someFancyForm" @model=@model />
```

``` js
// components/form-factory.js

export default class FormFactoryComponent extends Component {
    someFancyForm="components/forms/some-fancy-form";
    someOtherJuicyForm="components/forms/some-other-juicy-form";

    @argument
    @type("string")
    form = null;

    @argument
    @required
    model = null;

    @computed("form")
    get formComponent() {
        // TODO: validation
        return this[this.form];
    }
}
```

```
// templates/form-factory.hbs
{{if hasBlock}}
    {{yield
        (hash form=(component this.formComponent model=@model))
    }}
{{else}}
    {{component this.formComponent model=@model}}
{{/if}}
```

Developing this idea further, these simple Factory-like components could use some type of custom component using RFC 213 to minimise performance impact - I expect in my case I'll make use of a fair number of these component factories across my application to allow me to keep my semantic component organisation - knowing that these are minimal components likely with no runtime hooks or bindings to add weight could be a bonus.

I'm going to test out this idea over the coming days, but I just wanted to share my idea and get any feedback - maybe I missed something obvious, or maybe people are already doing something similar?

rtablada:
@lougreenwood

> One of my concerns around MU/Packages is that it seems to remove the ability to use directories for organising components.

The idea of the `use` helper is to allow these nested component trees to still work with the new angle bracket syntax (nested components work with the old curly brace component syntax and the component helper)

lougreenwood:
@rtablada 

I'm not sure that's the current intention, from all of the discussions I've seen here and in Discord `{{use}}` won't resolve nested components and any discussion of some kind of implementation for nested invocation should happen after this RFC is implemented and we are more aware of the problem space:

https://github.com/emberjs/rfcs/pull/367#issuecomment-427071114
https://github.com/emberjs/rfcs/pull/367#issuecomment-425228971

If I'm wrong, that's even better 🎉

rtablada:
Bringing some discussion from Discord and fresh 👀 to this RFC.

I think a lot of the discussion here has been about the particulars of the `use` helper and how it mitigates issues with <AngleBracket /> syntax and module imports.

I think since the syntax of `use` is the sticking point and question of a lot of discussion here IMO it should be moved to a separate RFC since there seems to be desire for it to be used outside of package/addon invocation.

While the rest of this package RFC requires an import API for addon/package component resolution, IMO it makes more sense to split those out (and maybe indicate any details/changes needed here)

text/0000-module-unification-packages.md
```diff
@@ -0,0 +1,1073 @@
...
+
+For example:
+
+```js
+export default Ember.Component.extend({
+
+  // inject src/services/geo.js
+  geo: inject(),
+
+  // inject node_modules/ember-stripe-service/src/services/store.js
```
pangratz:
```suggestion
  // inject node_modules/ember-stripe-service/src/services/stripe.js
```

mixonic:
Good morning! I'm closing this RFC today in favor of [SFC & Template Import Primitives](https://github.com/emberjs/rfcs/pull/454) and other possibly-to-be-written RFCs focused on a new filesystem layout for Ember.

The `{{use}}` part of the this RFC effectively lost consensus in October of last year, but framework core had to do a lot of soul searching to figure out exactly where to go next. The path still isn't totally clear, but there is a strong desire to have template imports simply be ES imports.

Putting aside the exact design addressing imports: There was some suggestion that this RFC could simply strip out the `{{use}}` bits and be ready to go. I don't think that is true. For one, the comments here would be largely irrelevant making further discussion confusing. And second, the motivations and constraints this RFC (and the MU RFC) were based on change **dramatically** given an ES import based template import system.

An "explicit import" system means *only* explicit imports. In the same way this this RFC suggested removing the implicit resolution of an MU package (an addon's) "main", explicit imports would replace the need for local lookup and even top-level component lookup in a package. You import any component you need, all the time, and always based on the path on disk.

That design obliterates most of the clever bits in the original MU RFC:

* As mentioned above this RFC already gutted the "main" resolution rule for packages.
* You don't need local lookup as a resolution rule, instead you import from a subdirectory.
* Without local lookup you don't need "private collections" in MU. We do lose something, the ability for a component to be private to a specific template, but we eliminate a bunch of design complexity and Ember-specific file layout rules.
* You don't need implicit lookup at the top level as a resolution rule. If templates are 100% explicit, then a template in an addon doesn't need implicit resolution rules to find its own top-level components. If we remove that behavior then the implicit scoping for JS files (as described in this RFC) and service injections should also be removed. All service injections can explicitly state their package.
* You obviously don't need `{{use}}` or any other package+component name invocation. Instead you import a component/helper/modifier from an addon based on it's path on disk in the NPM package.

Furthermore as far as creating a migration path for Ember app code: Templates with no implicit scope don't care about resolution rules, and so don't intersect with a new filesystem layout. Their design doesn't require we put components in `src/ui/components/`, for example, because everything is imported.

That leaves the migration path for Ember applications to a new filesystem layout wide open. We need to figure out a design for templates that considers layout on disk, and might use it as a carrot, but it decouples templates and component entirely from a resolvable system.

So what is left for MU, or any resolving system, to tackle?

A major fault in Ember's addon story is the merged `app/` directory. Much of the MU design tried to build a robust system that could replace that with properly scoped modules. We still need to address that fault.

Addons in Ember often provide only two kinds of resolved modules: Services and components. I've seen models provided as well. I'm unsure about how public the API is, but Ember developers can already inject a service from an addon's `addon/` dir with the `@` sign:

```js
export default EmberComponent.extend({
  session: inject('my-addon@session')
});
```

We could RFC the `src/` dir design for services as in this RFC with explicit packages:

```js
export default EmberComponent.extend({
  session: inject({ package: 'my-addon' })
});
```

...but if this is the only impact of that change I doubt it is worth adding `src/` to addons. There is maybe some middle ground here to be explored.

In general `src/` may end up as a tool to help Ember nudge and lint toward new best practices on the file system like co-located tests, single-file components, addon-providede styles, and co-located routes and templates. I think it may be possible to do some of that design work in parallel to the template imports design, but it certainly would not be a simple stripping of `{{use}}` from this RFC.

Thank you to the community members and contributors who helped hash out the nitty-gritty of this design! By fleshing out these ideas so concretely we demonstrated the best of the RFC process: We rejected the design *before* it caused churn for our apps.

PR #369, opened 2018-08-31
Deprecate Computed Overridability and `.readOnly()`

pzuraq:
[rendered](https://github.com/pzuraq/emberjs-rfcs/blob/deprecate-computed-clobberability/text/0000-deprecate-computed-clobberability.md)

### DIFF
--- a/text/0000-deprecate-computed-clobberability.md
+++ b/text/0000-deprecate-computed-clobberability.md
@@ -123,6 +123,18 @@ functions elsewhere and replace them with a proxy getter and the mandatory
 setter assertion, respectively. In the long term, making computeds more
 transparent in this way would be ideal, but it is out of scope for this RFC.
 
+## Deprecation Timeline
+
+This change will be a breaking change, which means we will not be able to change
+the behavior of `computed` until Ember v4.0.0. Until then, we should provide two
+deprecation warnings:
+
+1. When users invoke the default override-setter behavior
+2. When users call `.readOnly()` on a computed property
+
+The warnings should explain the deprecation, and recommend that users do not
+rely on setter behavior or opting-in to read only behavior.
+
 # How We Teach This
 
 In general, we can teach that computed properties are essentially cached native
--- a/text/0000-deprecate-computed-clobberability.md
+++ b/text/0000-deprecate-computed-clobberability.md
@@ -131,9 +131,24 @@ want to continue using `.readOnly()` up until overriding has been fully removed
 to ensure they are using properties safely. With that in mind, the ordering of
 events should be:
 
-1. Deprecate the default override-setter behavior immediately, remove it in
-Ember v4
-2. Deprecate the `.readOnly()` modifier in Ember v4, remove it in v5
+1. Ember v3
+  * Deprecate the default override-setter behavior immediately. This means that
+    a deprecation warning will be thrown if a user attempts to set a
+    non-`readOnly` property which does not have a setter. User's will still be
+    able to declare a property is `readOnly` without a deprecation warning.
+  * Add optional feature to change the deprecation to an assertion after the
+    deprecation has been released, and to show a deprecation when using
+    the `.readOnly()` modifier.
+  * After the deprecation and optional feature have been available for a
+    reasonable amount of time, enable the optional feature by default in new
+    apps and addons. The main reason we want to delay this is to give _addons_
+    a chance to address deprecations, since enabling this feature will affect
+    both apps and the addons they consume.
+2. Ember v4
+  * Remove the override-setter entirely, making non-overrideable properties the
+    default.
+  * Make the `readOnly` modifier a no-op, and show a deprecation warning when it
+    is used.
 
 The warnings should explain the deprecation, and recommend that users do not
 rely on setter behavior or opting-in to read only behavior.
--- a/text/0000-deprecate-computed-clobberability.md
+++ b/text/0000-deprecate-computed-clobberability.md
@@ -1,0 +1,146 @@
+- Start Date: 2018-08-30
+- RFC PR: (leave this empty)
+- Ember Issue: (leave this empty)
+
+# Summary
+
+Deprecate computed clobberability and `computed().readOnly()` in favor of
+read-only computeds as the default.
+
+# Motivation
+
+Computed properties have existed in Ember long before class syntax and native
+accessors (getters and setters) were readily available, and as such they have a
+few notable behavioral differences. As we move toward adopting native class
+syntax and using a decorator-based form of computeds, it makes sense to
+reconcile these differences so that users can expect them to work the same as
+their native counterparts.
+
+The main and most notable difference this RFC seeks to deprecate is computed
+clobberability. There are some other notable differences, including the caching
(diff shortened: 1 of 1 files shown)


text/0000-deprecate-computed-clobberability.md
```diff
@@ -0,0 +1,146 @@
...
+
+Computed properties have existed in Ember long before class syntax and native
+accessors (getters and setters) were readily available, and as such they have a
+few notable behavioral differences. As we move toward adopting native class
+syntax and using a decorator-based form of computeds, it makes sense to
+reconcile these differences so that users can expect them to work the same as
+their native counterparts.
+
+The main and most notable difference this RFC seeks to deprecate is computed
+clobberability. There are some other notable differences, including the caching
```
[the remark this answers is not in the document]
pzuraq: `override` seems like a more explanatory term without context. Maybe also `overload`. I'll update the language of the RFC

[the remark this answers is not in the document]
mike-north: might be easier to just put a dummy term in there for now. Naming is hard, but very important in this case

mike-north:
I really like this proposed change! Yet one more place where we're adjusting defaults to set as many developers and use cases as possible up for success.

RE: decorators -- have we considered this ⬇️ option for stuff that chains off of the return of `Ember.computed()` like `.readOnly()` or `.clobberable()`?

```ts

@computed.foo.bar
get myProp() { }

```

Personally, I find something like this to be less readable
```ts
@foo
@bar
@computed
get myProp() { }

```

rwjblue:
> have we considered this ⬇️ option for stuff that chains off of the return of Ember.computed() like .readOnly() or .clobberable()?
>
> ```js
> @computed.foo.bar
> get myProp() { }
> ```

Yep, @pzuraq and I discussed this yesterday actually. I think our conclusion was something like what you wrote:

```js
@computed('first', 'last').thing()
```

pzuraq:
@mike-north I actually just went over the spec again and realized we're more limited than I thought. Decorators can only be chains of properties with one final call at the end:

```js
// valid
@foo
@foo.bar
@foo()
@foo.bar()

// invalid
@foo().bar
@foo().bar()
```

This is going to affect `readOnly` and `volatile` too unfortunately 😕

mike-north:
This syntax is still valid. Remember we're just capturing dependent keys and **flagging the property in some manner** to indicate readonly-ness. We only need to capture arguments once
```ts
class {
   @computed.volatile.readOnly('firstName', 'lastName')
   get fullName() {
		...
   }
}
```
There are some [very popular projects](https://github.com/chalk/chalk) that have chaining like this, albeit in a conventional function call instead of a function that returns a decorator.

balinterdi:
@pzuraq Thank you for putting this together.

While we still can't just use ES classes and decorators in Ember (or can we already?) how would we replace the following if this RFC gets implemented (since it would deprecate computed overridability):

```js
  showErrors: computed('_showErrors', {
    get() {
      return this._showErrors || { email: false, password: false };
    },
    set(key, value) {
      this.set('_showErrors', value);
      return this._showErrors;
    }
  }),
```

(The transition path shows an example but it uses an ES class.)

Thank you.

pzuraq:
Hey Balint, the example you posted would be perfectly valid since the computed property provides a setter. What would not be valid is this:

```js
const Foo = EmberObject.extend({
  showErrors: computed('_showErrors', {
    get() {
      return { email: false, password: false };
    }
  })
});

let foo = Foo.create();

Foo.set('showErrors', { email: true, password: true }); // Throws an error because `showErrors` does not have a setter
```

Does that make sense?

balinterdi:
@pzuraq Yes, it does and I then wholeheartedly support this RFC :) Thank you.

Gaurav0:
I don't object to the spirit of this RFC, but I need to point out that this is a breaking change and will likely mess up poorly written legacy apps. I strongly advise that we throw a deprecation warning when calling set() on a computed property without a setter, and then make them readonly by default in 4.0.

pzuraq:
@Gaurav0 that was the original intent of this RFC, but the deprecation aspect of it was only implied by the title and the summary. I'll add an extra section to the Transition Path section to clarify that the transition would follow the standard deprecation path for breaking changes, showing a warning until v4.0.0

text/0000-deprecate-computed-clobberability.md
```diff
@@ -0,0 +1,146 @@
...
+notable differences, including the caching behavior of the `return` value of
+setter functions, which may be addressed in future RFCs.
+
+## Overridability
+
+When defining a native getter without a setter, attempting to set the value will
+throw a hard error (in strict mode):
+
+```js
+function makeFoo() {
```
tragiclifestories: nitpick - the indentation in this block is messed up

text/0000-deprecate-computed-clobberability.md
```diff
@@ -123,6 +123,18 @@ functions elsewhere and replace them with a proxy getter and the mandatory
...
 transparent in this way would be ideal, but it is out of scope for this RFC.
 
+## Deprecation Timeline
+
+This change will be a breaking change, which means we will not be able to change
+the behavior of `computed` until Ember v4.0.0. Until then, we should provide two
+deprecation warnings:
+
+1. When users invoke the default override-setter behavior
+2. When users call `.readOnly()` on a computed property
```
Gaurav0: If we deprecate this, users have no way of getting a hard error (not a deprecation warning) when they clobber without getting deprecation warnings by using readOnly.

I recommend only deprecating .readOnly() in the 4.x era. Yeah, like I said, this is a pain.

text/0000-deprecate-computed-clobberability.md
```diff
@@ -0,0 +1,176 @@
...
+want to continue using `.readOnly()` up until overriding has been fully removed
+to ensure they are using properties safely. With that in mind, the ordering of
+events should be:
+
+1. Ember v3
+  * Deprecate the default override-setter behavior immediately. This means that
+    a deprecation warning will be thrown if a user attempts to set a
+    non-`readOnly` property which does not have a setter. User's will still be
+    able to declare a property is `readOnly` without a deprecation warning.
+  * Add optional feature to change the deprecation to an assertion after the
```
rwjblue: We chatted about this today in the core team meeting, and decided to try to leverage svelte here instead of an optional feature. This doesn't change the timeline at all here, just the verbiage.

Can you tweak this from `Add optional feature` to `When the application is configured with svelte, the deprecation will be an assertion`?

webark:
@pzuraq did the `overridable` computed macro ever get into an addon somewhere?

pzuraq:
It did not, but it should be relatively easy to write to interop with something like [macro-decorators](https://github.com/pzuraq/macro-decorators). I don't think it'd be easy to write to interop with `@computed`, unfortunately, due to the mandatory setter.

alexlafroscia:
I ended up building an `overriable` macro as an addon here:

https://github.com/alexlafroscia/ember-overridable-computed

It doesn't support being used as a decorator yet, but I'll do that eventually!

PR #372, opened 2018-09-07
[RFC ember-data] modelFactoryFor

runspired:
[rendered](https://github.com/runspired/rfcs/blob/ember-data-custom-records/text/0000-ember-data-model-factory-for.md#ember-data--modelfactoryfor)

### DIFF
--- a/text/0000-ember-data-model-factory-for.md
+++ b/text/0000-ember-data-model-factory-for.md
@@ -60,7 +60,7 @@ interface Store {
 }
 ```
 
-Users interested in providing a custom class for their `records` and whom override `modelFactoryFor`,
+Users interested in providing a custom class for their `records` and who override `modelFactoryFor`,
  would not need to also change `modelFor`, as this would be the `klass` accessible via the `factory`.
 
 Users wishing to extend the behavior of `modelFactoryFor` could do so in the following manner:
@@ -129,33 +129,34 @@ export default class CustomModel extends EmberObject {
 Custom classes for models should expect their constructor to receive a single argument: an object with *at least*
  the following.
  
-- A `recordData` instance available on the `recordData` `Symbol` and accessible via `getRecordData` (see below)
+- A `recordData` instance accessible via `getRecordData` (see below)
 - Any properties passed as the second arg to `createRecord`
-- An `owner` available on the owner `Symbol` and accessible via `Ember.getOwner`
+- An `owner` accessible via `Ember.getOwner`
 - Any DI injections
 - any other properties that `Ember` chooses to pass to a class instantiated via `factory.create` (currently none)
 
-### getRecordData/setRecordData
+### getRecordData
 
 Every `record` (instance of the class returned by `modelFactoryFor`) will have an associated [RecordData](https://github.com/emberjs/rfcs/pull/293)
  which contains the backing data for the id, type, attributes and relationships of that record.
  
 This backing data can be accessed by using the `getRecordData` util on the `record` (or on the `createArgs` passed to
- a record).
+ a record). Using `getRecordData` on a `record` is only guaranteed after the record has been instantiated. During
+ instantiation, this call should be made on the `createArgs` object passed into the record.
  
-If providing a custom class that does not inherit from `EmberObject`, you *MUST* either use `assign` (see example 2)
- or `setRecordData` (see example 4) to add the symbol for the backing `recordData` instance onto your record instance.
- If inheriting from `EmberObject` calling `super` with the `createArgs` is sufficient.
-
 **Example 4**
 
 ```javascript
-import { getRecordData, setRecordData } from 'ember-data';
+import { getRecordData } from 'ember-data';
 
 export default class CustomModel {
   constructor(createArgs) {
+    // during instantiation, `recordData` is available by calling `getRecordData` on createArgs
     let recordData = getRecordData(createArgs);
-    setRecordData(this, recordData);
+  }
+  someMethod() {
+    // post instantiation, `recordData` is available by calling `getRecordData` on the instance
+    let recordData = getRecordData(this);
   }
   destroy() {
     // ... do teardown
--- a/text/0000-ember-data-model-factory-for.md
+++ b/text/0000-ember-data-model-factory-for.md
@@ -1,0 +1,82 @@
+- Start Date: 2018-09-06
+- RFC PR: (leave this empty)
+- Ember Issue: (leave this empty)
+
+# ember-data | modelFactoryFor
+
+## Summary
+
+Promote the private `store._modelFactoryFor` to public API as `store.modelFactoryFor`.
+
+## Motivation
+
+This RFC is a follow-up RFC for [#293 RecordData](https://github.com/emberjs/rfcs/pull/293).
+
+Ember differentiates between `klass` and `factory` for classes registered with the container.
+ At times, `ember-data` needs the `klass`, at other times, it needs the `factory`. For this reason,
+`ember-data` has carried two APIs for accessing one or the other for some time. The public `modelFor`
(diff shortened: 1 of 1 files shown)


knownasilya:
Could this work side by side with the alternative method?

runspired:
@knownasilya yes of course :)

runspired:
We discussed this on our weekly meeting and decided to put this RFC on hold for 2 weeks until the Ember f2f in Chicago after EmberCamp. @dgeb will champion discussion on improving Ember's DI / spec'ing `factory` since that is outside of our domain.

Additionally, we should verify that `DS.Model` is no longer depended upon internally, and that a theoretical factory could simply accept `modelData` as a prop during initialization and that be enough.

runspired:
After further discussion, we've update the proposal with the expected requirements for custom classes and use of `factoryFor`.

text/0000-ember-data-model-factory-for.md
```diff
@@ -0,0 +1,189 @@
...
+}
+
+declare function factoryFor<K extends keyof FactoryMap>(factoryName: K): FactoryMap[K];
+
+interface Store {
+  modelFactoryFor(modelName: string): ReturnType<typeof factoryFor>;
+}
+```
+
+Users interested in providing a custom class for their `records` and whom override `modelFactoryFor`,
```
hjdivad: typo: s/whom/who

text/0000-ember-data-model-factory-for.md
```diff
@@ -0,0 +1,189 @@
...
+  constructor(createArgs) {
+    super(createArgs);
+  }
+}
+```
+
+Custom classes for models should expect their constructor to receive a single argument: an object with *at least*
+ the following.
+ 
+- A `recordData` instance available on the `recordData` `Symbol` and accessible via `getRecordData` (see below)
```
hjdivad: I'd recommend not mentioning the `Symbol`, or rephrasing to make it clear that it's an implementation detail and not part of the contract.

eg

> - A `recordData` instance accessible via `getRecordData(createArgs)`

text/0000-ember-data-model-factory-for.md
```diff
@@ -0,0 +1,189 @@
...
+- Any properties passed as the second arg to `createRecord`
+- An `owner` available on the owner `Symbol` and accessible via `Ember.getOwner`
```
hjdivad: I don't think the `Ember.getOwner` docs mention anything about a `Symbol`; let's just say the owner is accessible via `Ember.getOwner`

text/0000-ember-data-model-factory-for.md
```diff
@@ -0,0 +1,189 @@
...
+
+Every `record` (instance of the class returned by `modelFactoryFor`) will have an associated [RecordData](https://github.com/emberjs/rfcs/pull/293)
+ which contains the backing data for the id, type, attributes and relationships of that record.
+ 
+This backing data can be accessed by using the `getRecordData` util on the `record` (or on the `createArgs` passed to
+ a record).
+ 
+If providing a custom class that does not inherit from `EmberObject`, you *MUST* either use `assign` (see example 2)
+ or `setRecordData` (see example 4) to add the symbol for the backing `recordData` instance onto your record instance.
+ If inheriting from `EmberObject` calling `super` with the `createArgs` is sufficient.
```
hjdivad: Why is it necessary for the record to have a backref to the record data?

We should certainly recommend that people do this, but I don't see why we should require it.  I'd also recommend we make the public api here `setRecordData` and not allow the record data to be set via `assign`, to simply the docs if nothing else.

runspired: @hjdivad implementations of `Model` and `RecordData` need to be able to convert from a `record` to a `record-data` when `createRecord` is called or when a relationship is mutated. 

Overtime we may be able to loosen that requirement, although personally I think that even if no longer necessary it is a good one.

RE `assign`, since we are using a `Symbol`, `Object.assign` and the Ember polyfill `"just work"`. `Object.keys` and `for ... in` do not work unless using Ember's `Symbol` polyfill.

screen shot 2018-10-04 at 1 42 44 pm

hjdivad: Yes, `assign` can work but it will be simpler if we have a single public API, `setRecordData`.  `assign` works as a side effect of using a `Symbol`, but this is an implementation detail (you could imagine supporting the APIs with `WeakMap`s for instance).

> @hjdivad implementations of Model and RecordData need to be able to convert from a record to a record-data when createRecord is called or when a relationship is mutated.

That's true for `DS.Model`s but need not be the case for other model classes.  That `record.set('belongsToRelationship', otherRecord)` invokes `recordData.setBelongsTo` is an implementation detail of `DS.Model` -- what part of the system necessitates that this is done by *all* model implementations?

I think it's important that we avoid these kinds of requirements to allow users an escape valve for relationship semantics when they need it.

runspired: > I think it's important that we avoid these kinds of requirements to allow users an escape valve for relationship semantics when they need it.

I feel as though we won't be able to have any-level of decent interop between various `RecordData` and `Model` implementations if we don't have this constraint. While I agree the existing relationship layer could (and should) be refactored to be less coupled, I think it's mechanics derives in part from the need for an interop story.

runspired: > Yes, assign can work but it will be simpler if we have a single public API, setRecordData. assign works as a side effect of using a Symbol, but this is an implementation detail (you could imagine supporting the APIs with WeakMaps for instance).

I can get behind this idea and will update the language.

runspired: I think if we consider `recordData` as a "weak-mapped" thing (even if symbol based), we may be able to remove the requirement. Users would need to call `recordDataFor` on `createArgs` from within the constructor, but afterwards we could weakMap to the instance or symbolicate it ourselves and calls to `recordDataFor` on the instance would "just work"

hjdivad: > I feel as though we won't be able to have any-level of decent interop between various RecordData and Model implementations if we don't have this constraint. While I agree the existing relationship layer could (and should) be refactored to be less coupled, I think it's mechanics derives in part from the need for an interop story.

That interop should be driven by `RecordData`s; the relationships already deal with these objects and should be as indifferent as can be to the particular records associated with some particular `RecordData`.

There's no dispute that `RecordData`s need to know what record to return at various times: the basic mechanism of caching an invocation of `factory.create(...)` works well here.  Model implementations can set up edges between `RecordData`s via the public `RecordData` API.  Why is that insufficient to achieve interop?

runspired: @hjdivad because if you are a `record` and someone calls `record.set` with a different `record` then either the `record` or the `record-data` for the receiving `record` needs to be able to lookup the `record-data` instance associated with it. In absence of this ability, you end up needing to expose various other public interfaces to make things work. That said, I think the "weak-map after create" approach will allow us to make this work just fine.

hjdivad: > @hjdivad because if you are a record and someone calls record.set with a different record then either the record or the record-data for the receiving record needs to be able to lookup the record-data instance associated with it.

Sure, so it's recommended to `setRecordData` to achieve this.  But that's not the same thing as it being required.  In particular I don't think the rest of the system should assume that records (besides `DS.Model`s) have a backref of any kind to their `RecordData`.

hjdivad:
Discussed this at the core team meeting today, and we are moving this to final comment period.  We'll aim to merge this next week, barring additional comments.

runspired:
As an update to this RFC, we landed an implementation only to discover several fatal flaws when interopiing with Ember's object factories that lead to a revert. A new approach is being drafted by @igorT, and this RFC should be considered *dead*

PR #378, opened 2018-09-18
RFC to collocate tests in in-repo-addons

tmquinn:
[Rendered](https://github.com/tmquinn/rfcs/blob/collocate-in-repo-addon-tests/text/0000-collocate-in-repo-addon-tests.md)

### DIFF
--- a/text/0000-collocate-in-repo-addon-tests.md
+++ b/text/0000-collocate-in-repo-addon-tests.md
@@ -22,12 +22,14 @@ Having a simple syntax for executing tests associated with an addon is useful an
   - `ember test --ir-addon foo-bar`
   - The flag's use would create a list of tests to run that are collocated in that in-repo-addon's test folder. The tests will still be included all in one bundle, so this is basically just running the modules that are associated with the co-located tests.
 
-## Not covered in this RFC
+## Not covered in this RFC (but will be covered in a following RFC)
 - Splitting the test bundles
   - Tests should be separately bundled to allow running these tests without loading other tests.
   - In very large application with many in-repo-addons, there is significant overhead in loading all of the application's tests, when only a single in-repo-addon's tests are being tested
   - Expect another RFC coming soon to address this functionality
-
+- Option to run addpn tests with dummy app
+  - Tests for an in-repo-addon/engine could be chosen to either run against the main app, or a (shared) dummy app.
+  
 ## How we teach this
 
 - Create a guide chapter on in-repo-addons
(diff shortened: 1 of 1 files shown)


luketheobscure:
I'm assuming that in-repo lazy engines would be covered under this? We might need to specify that by default the in-repo tests would use the `engineResolverFor` logic when setting up the resolver.

jeanduplessis:
I would very much love to see the inclusion of having a dummy app as mentioned in the "Unresolved questions" section.

tmquinn:
@luketheobscure Could you expand? I will make it a little clearer in the RFC that only the location of the test source code will change. The built assets will remain identical to the non-collated version.

tmquinn:
@jeanduplessis I will try and get that RFC up as soon as possible. I'm intentionally keeping them separate as I have the code to make this RFC happen, but there is a lot of discussion that will go into the idea of separating the test output into separate bundles or having dummy apps in the in-repo-engines. Although that is certainly the end goal and the bit I am most excited to see happen!

runspired:
Should this RFC be in https://github.com/ember-cli/rfcs ?

jamesarosen:
See also ember-cli/rfcs#60

tmquinn:
> Should this RFC be in https://github.com/ember-cli/rfcs ?

I heard that ember-cli/rfcs might be deprecated soon. I was going to inquire if I should still move this to ember-cli, but after a recent write up, should we just leave it here?

rwjblue:
> Should this RFC be in ember-cli/rfcs ?

tldr; no, I don't think so. https://github.com/emberjs/rfcs/pull/300 has details (needs some edits though 😉 @locks), but I think we would generally like to reduce confusion and use a single RFCs repo.

text/0000-collocate-in-repo-addon-tests.md
```diff
@@ -0,0 +1,48 @@
...
+
+Having a simple syntax for executing tests associated with an addon is useful and time-saving.
+
+## Detailed design
+
+- Indicate that an addon has tests that are collocated
+  - use `includeTestsInHost: true` in `index.js` to indicate collocated tests
+- Implement CLI flag that only runs the tests for a specified addon
+  - `ember test --ir-addon foo-bar`
+  - The flag's use would create a list of tests to run that are collocated in that in-repo-addon's test folder. The tests will still be included all in one bundle, so this is basically just running the modules that are associated with the co-located tests.
```
ro0gr: Since all the test are included into a single bundle why do we need `--ir-addon` flag? Build takes a while, building several `--ir` tests may be annoying. 

What if we provide something like _"Engines"_ or _"Bundles"_ dropdown in qunit interface?

wagenet:

PR #384, opened 2018-10-04
[DEPRECATION RFC ember-data] deprecate errors changing record state

runspired:
[Rendered](https://github.com/runspired/rfcs/blob/ember-data-errors-manipulation-deprecation/text/0000-ember-data-deprecate-errors-changing-valid-state.md#ember-data--deprecate-errors-changing-record-state)

### DIFF
--- a/text/0000-ember-data-deprecate-errors-changing-valid-state.md
+++ b/text/0000-ember-data-deprecate-errors-changing-valid-state.md
@@ -1,0 +1,65 @@
+- Start Date: 2018-10-04
+- RFC PR: (leave this empty)
+- Ember Issue: (leave this empty)
+
+# Ember Data | Deprecate Errors changing record-state
+
+## Summary
+
+Deprecates the `add` `remove` and `clear` methods of `DS.Errors` in favor of `addError`
+ `removeError` and `clearErrors`.
+
+## Motivation
+
+In `ember-data`, adding, removing, or clearing errors from a record's `errors` object
+ will affect the validation state of the record. Previously a [soft-deprecation](https://github.com/emberjs/data/pull/3853) of
+ this behavior was added; however, it provided no clear path to resolution. 
+
+By deprecating these methods entirely, and providing replacement methods that do not
+ affect the state of the `record`, we provide end users a clear path and a clear timeline
+ for resolving the deprecation, and a mechanism through with users can determine if this
+ change in behavior requires any additional changes in their application code.
+
+## Detailed design
+
+The warnings in `DS.Errors` `add` `remove` and `clear` will be upgraded to deprecations
+ targeting the next LTS `3.8`. This means the feature and these methods will be removed
+ in `3.9`.
+
+New methods will be added to `DS.Errors` that do not affect record state. `addErrors` `removeErrors`
+ and `clearErrors`. These methods will have the same signatures as before, listed below, where
+ `member` is one of `type` `id` or the property name of an `attribute` or `relationship`.
+
+```typescript
+interface Errors {
+  addErrors(member: string, messages: string|Array<string>): void;
+  removeErrors(member: string): void;
+  clearErrors(): void;
+}
+```
+
+## How we teach this
+
+`DS.Errors` is not well documented today, so this is an opportunity for us to revisit and
+ clean up.
+
+## Drawbacks
+
+- For apps using `Errors` that do not rely on the changes to record state, this will feel like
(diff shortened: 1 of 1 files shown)


sandstrom:
Great suggestion! 🏅 

How about adding the methods on the model itself, i.e. `model.addErrors` etc?

To me it would feel more consistent with the current API, for example accessing relationships are via `model.belongsTo` (it's not `model.relationships.belongsTo`).

text/0000-ember-data-deprecate-errors-changing-valid-state.md
```diff
@@ -0,0 +1,65 @@
+- Start Date: 2018-10-04
+- RFC PR: (leave this empty)
+- Ember Issue: (leave this empty)
+
+# Ember Data | Deprecate Errors changing record-state
+
+## Summary
+
+Deprecates the `add` `remove` and `clear` methods of `DS.Errors` in favor of `addError`
+ `removeError` and `clearErrors`.
```
lindyhopchris: This uses the singular whereas the design uses the plural.

hjdivad:
@sandstrom i'd rather not scope creep into changing the API itself; the goal of this rfc is to follow-through on the (now quite old) deprecation and make it actionable for users.

---

On the question of moving the API to the model, I'd prefer to avoid adding more API surface area to a class users already subclass a lot.

> To me it would feel more consistent with the current API, for example accessing relationships are via model.belongsTo (it's not model.relationships.belongsTo).

This is a good point about consistency although i think the safer way to consolidate this would be to move `belongsTo` &c. to `record.relationships` rather than to have a larger API on `DS.Model`.

hjdivad:
@sandstrom I do like the idea of a flag rather than moving the methods.  But the downside is the flag is all or nothing: in a large codebase it can be important to be able to migrate deprecations incrementally.

Adding a flag to the call site *does* allow for incremental migration and from my point of view is roughly equivalent to making a new method.  I don't have a strong opinion between the two choices.  As i see it, each has a slight edge over the other:

1. Adding a flag creates more work for people migrating as they add the flag and then later should drop it for cleanup once the deprecation is removed
2. But creating less pleasant method names impacts new projects

I could go either way (although again I'm not a fan of a global flag due to the impact on incremental conversion).

@runspired @dgeb ?

runspired:
@sandstrom @hjdivad 

I agree with David on being opposed to a global flag. I am additionally against the local flag and in favor of the changes here as proposed for three reasons. I list them here in order of importance:

Assuming a local opt-in API like the following:

```typescript
add(member: string, message: string|string[], optIntoFuture: boolean)
```

1) My largest complaint with this syntax is that it leads to a double deprecation lifecyle.

Not only would users have to opt-in at each call site to eliminate the deprecation and ensure their code did not break, but once the deprecation completed they would need to receive a new deprecation telling them to drop the third argument.

2) My second largest complaint with this syntax is that the syntax today is misleading. `add` `remove` and `clear` are evocative of the `set` `delete` and `clear` APIs of `Map` and `WeakMap`. Given their behaviors and the nature of the data structure (a mapping of errors), one might expect the APIs to behave similarly, but they don't.

`set(key, value)` doesn't differentiate between value being an `array` vs not, but `add/addErrors` does.

`delete(key)` doesn't leave behind an empty array, but `remove/removeErrors` does. Also note, that while you can add a single error to an attribute, remove removes all of them, not just the one.

`clear` is the only API of the three that is semantically the same.

Both because I dislike the confusion with the mechanics of true `Map` classes, and because of the differences in function signatures, the names of the existing APIs make me uneasy.

sandstrom:
@runspired Good arguments!

1. In my opinion we could skip the second deprecation, since the third argument will be ignored. But I understand if you'd like to do it "the right way" and make it a two-phase thing.

2. Not really sold on the confusion risk with map, I think people will understand that `add` on different objects will have different effects.

All this said though, I'm only the voice of one person and if you as the RFC champion disagree I think we should go ahead with this PR anyway 😄 

Even though I have some concerns (1 & 2) it's overall a good change that's presented in this RFC! 🏅

hjdivad:
Moving this to final comment period.  We'll aim to merge next week, barring any additional concerns.

Thanks very much @sandstrom @lindyhopchris for taking the time to review and give feedback. 🎉

runspired:
We landed a different approach to resolving this due to a desire to reconsider how we propagate errors through the system more holistically.

PR #388, opened 2018-10-13
Add new basic helpers to Ember

cibernox:
[Rendered](https://github.com/cibernox/rfcs/blob/new-basic-handlebars-helpers/text/0000-basic-template-helpers.md)

### DIFF
--- a/text/0000-basic-template-helpers.md
+++ b/text/0000-basic-template-helpers.md
@@ -67,6 +67,7 @@ and move them into Ember.js itself or even the Glimmer VM, in a fully backwards 
 I propose to add to core at least:
 
 - `eq`
+- `not-eq`
 - `not`
 - `and`
 - `or`
@@ -81,7 +82,6 @@ I propose to not add:
 - `is-empty` (uses `Ember.isEmpty`)
 - `is-equal` (uses `Ember.isEqual`)
 - `xor`      (not very common)
-- `not-eq`   (can be expressed as `(not (eq a b)`)
 
 When this feature is implemented, we would update `ember-truth-helpers` to automatically remove
 the promoted helpers from the code when the Ember is above a certain version number.
--- a/text/0000-basic-template-helpers.md
+++ b/text/0000-basic-template-helpers.md
@@ -1,0 +1,86 @@
+- Start Date: 2018-10-13
+- RFC PR: (leave this empty)
+- Ember Issue: (leave this empty)
+
+# (RFC title goes here)
+
+## Summary
+
+Add new built-in template helpers to perform basic operations in templates, identical to the
+helpers in [ember-truth-helpers](https://github.com/jmurphyau/ember-truth-helpers)
+
+## Motivation
+
+It is a very common necessity to almost every Ember app to perform certain operations like compare
+by equality, negate a value, or perform boolean operations, and often the most convenient place to
+do it is in the templates.
+Because of that, `ember-truth-helpers` is probably the single most installed addon either directly
+or transitively in the entire ecosystem.
+
+The fact that this addon is so popular is very telling and Ember should consider to move into the
+core of the templating engine at least the most popular helpers.
+
+One of the shocking moments developers that are familiar with React, Vue or Angular have when trying Ember
+is that they cannot, at least out of the box, perform the most basic comparisons and boolean operations
+they are so used to in JSX or Vue/Angular templates.
+
+Furthermore, if the [RFC #367](https://github.com/emberjs/rfcs/pull/367) eventually gets merged, these
+helpers are the perfect candidates to create adoption friction in the community because of how
+pervasive its usage it in so many templates.
+
+Albeit less important, implementing these helpers in the core _could_ enable clever trickery in the
+Glimmer VM, like evaluating them at compile time when the arguments are constant.
+
+## Detailed design
+
+The process consists on deciding what helpers from `ember-truth-helpers` we consider the most important
+and move them into Ember.js itself or even the Glimmer VM, in a fully backwards compatible way.
+
+I propose to move to core at least:
+
+- `eq`
+- `not`
+- `and`
+- `or`
+- `gt` and `gte`
+- `lt` and `lte`
+
+Those helpers are very low lever and generally useful in both Ember and Glimmer.
+
+I propose to not move:
+
+- `is-array` (uses `Ember.isArray`)
+- `is-empty` (uses `Ember.isEmpty`)
+- `is-equal` (uses `Ember.isEqual`)
+- `xor`      (not very common)
+- `not-eq`   (can be expressed as `(not (eq a b)`)
+
+When this feature is implemented, we would update `ember-truth-helpers` to not ship the helpers that
+were moved into the core if the version of Ember is
+
+## How we teach this
+
+As any new API, will have to be documented in the guides with examples.
+
+## Drawbacks
+
+We are increasing the public API of the framework, and every line of code is a liability, although
+those helpers are extremely straightforward.
+
+## Alternatives
+
+An alternative path would be to include `ember-truth-helpers` in the default blueprint for apps an
+addons.
+
+## Unresolved questions
+
+The main unresolved question is what helpers we deem worthy of being moved into the core, and also
+if we want any other helpers not mentioned above.
+
+Other helpers that come to mind that _could_ also be worth adding:
+
+- `array` (there is an RFC for it: https://github.com/emberjs/rfcs/pull/318)
+- `add`, `subtract`, and other arithmetic operators.
+- `{{#await promise as |value|}}` to render a block conditionally if a promise resolves or fails.
+
+If there is interest in them, I advocate creating standalone RFCS for them.
--- a/text/0000-basic-template-helpers.md
+++ b/text/0000-basic-template-helpers.md
@@ -86,6 +86,61 @@ I propose to not add:
 When this feature is implemented, we would update `ember-truth-helpers` to automatically remove
 the promoted helpers from the code when the Ember is above a certain version number.
 
+The implementation details of each helper:
+
+#### `{{not}}`
+
+- Unary operation. Throws an error if not called with **exactly** one argument.
+- Equivalent of `!<argument>`
+
+#### `{{eq}}`
+
+- Binary operation. Throws an error if not called with **exactly** two arguments.
+- Equivalent of `<arg1> === <arg2>`
+
+#### `{{not-eq}}`
+
+- Binary operation. Throws an error if not called with **exactly** two arguments.
+- Equivalent of `<arg1> !== <arg2>`
+
+#### `{{gt}}`
+
+- Binary operation. Throws an error if not called with **exactly** two arguments.
+- Equivalent of `<arg1> > <arg2>`
+- Lazy: If the first argument is `null` or `undefined`, the second argument is never evaluated
+
+#### `{{gte}}`
+
+- Binary operation. Throws an error if not called with **exactly** two arguments.
+- Equivalent of `<arg1> >= <arg2>`
+- Lazy: If the first argument is `null` or `undefined`, the second argument is never evaluated
+
+#### `{{lt}}`
+
+- Binary operation. Throws an error if not called with **exactly** two arguments.
+- Equivalent of `<arg1> < <arg2>`
+- Lazy: If the first argument is `null` or `undefined`, the second argument is never evaluated
+
+#### `{{lte}}`
+
+- Binary operation. Throws an error if not called with **exactly** two arguments.
+- Equivalent of `<arg1> <= <arg2>`
+- Lazy: If the first argument is `null` or `undefined`, the second argument is never evaluated
+
+#### `{{and}}`
+
+- Binary or greater operation. Throws an error if called with **less than** two arguments.
+- Equivalent of `<arg1> && <arg2> && ... && <argN>`. That means it returns the last truthy value or the first falsy value.
+- Definition of truthiness: The same the `&&` operator has in javascript.
+- Lazy: It starts evaluating arguments in order and short-circuits as soon as one of them is falsy.
+
+#### `{{or}}`
+
+- Binary or greater operation. Throws an error if called with **less than** two arguments.
+- Equivalent of `<arg1> || <arg2> || ... || <argN>`. That means it returns the first truthy value or the last value.
+- Definition of truthiness: The same the `||` operator has in javascript.
+- Lazy: It starts evaluating arguments in order and short-circuits as soon as one of them is truthy.
+
 ## How we teach this
 
 The introduction of these helpers does not impact the current mental model for Ember applications.
--- a/text/0000-basic-template-helpers.md
+++ b/text/0000-basic-template-helpers.md
@@ -2,7 +2,7 @@
 - RFC PR: (leave this empty)
 - Ember Issue: (leave this empty)
 
-# (RFC title goes here)
+# Promote helpers from ember-truth-helpers into Ember
 
 ## Summary
 
@@ -55,8 +55,8 @@ I propose to not move:
 - `xor`      (not very common)
 - `not-eq`   (can be expressed as `(not (eq a b)`)
 
-When this feature is implemented, we would update `ember-truth-helpers` to not ship the helpers that
-were moved into the core if the version of Ember is
+When this feature is implemented, we would update `ember-truth-helpers` to automatically remove
+the promoted helpers from the code when the Ember is above a certain version number.
 
 ## How we teach this
 
--- a/text/0000-basic-template-helpers.md
+++ b/text/0000-basic-template-helpers.md
@@ -1,5 +1,5 @@
 - Start Date: 2018-10-13
-- RFC PR: (leave this empty)
+- RFC PR: https://github.com/emberjs/rfcs/pull/388
 - Ember Issue: (leave this empty)
 
 # Add new basic template helpers to Ember
--- a/text/0000-basic-template-helpers.md
+++ b/text/0000-basic-template-helpers.md
@@ -89,7 +89,10 @@ the promoted helpers from the code when the Ember is above a certain version num
 
 ## How we teach this
 
-As any new API, will have to be documented in the guides with examples.
+The introduction of these helpers does not impact the current mental model for Ember applications.
+
+In addition to API and Guides documentation with illustrative examples of usage of the various helpers,
+and explanation of their short circuiting nature might be warranted.
 
 ## Drawbacks
 
@@ -98,8 +101,10 @@ those helpers are extremely straightforward.
 
 ## Alternatives
 
-An alternative path would be to include `ember-truth-helpers` in the default blueprint for apps an
+An alternative path would be to include `ember-truth-helpers` in the default blueprint for apps and
 addons.
+However, this alternative loses strength due to the fact that it is not possible to implement short
+circuiting helpers in Ember's userspace.
 
 ## Unresolved questions
 
--- a/text/0000-basic-template-helpers.md
+++ b/text/0000-basic-template-helpers.md
@@ -7,29 +7,58 @@
 ## Summary
 
 Add new built-in template helpers to perform basic operations in templates, identical to the
-helpers in [ember-truth-helpers](https://github.com/jmurphyau/ember-truth-helpers)
+helpers in [ember-truth-helpers](https://github.com/jmurphyau/ember-truth-helpers).
+
+This is a resurrection of [RFC #152](https://github.com/emberjs/rfcs/pull/152) that was opened over
+by @martndemus two years ago, updated to reflect the new start of the things in Ember, with the glimmer
+VM powering our templates.
 
 ## Motivation
 
 It is a very common necessity to almost every Ember app to perform certain operations like compare
-by equality, negate a value, or perform boolean operations, and often the most convenient place to
-do it is in the templates.
-Because of that, `ember-truth-helpers` is probably the single most installed addon either directly
-or transitively in the entire ecosystem.
+by equality, negate a value or perform boolean operations, and often the most convenient place to
+do it is right in the templates.
+Because of that, `ember-truth-helpers` is probably the single most installed addon that exists, either
+directly by apps or indirectly by other addons that those apps use.
 
-The fact that this addon is so popular is very telling and Ember should consider to move into the
-core of the templating engine at least the most popular helpers.
+The fact that this addon is so popular is very telling and Ember.js should consider moving into the
+core of the templating engine at least some of those helpers.
 
-One of the shocking moments developers that are familiar with React, Vue or Angular have when trying Ember
-is that they cannot, at least out of the box, perform the most basic comparisons and boolean operations
-they are so used to in JSX or Vue/Angular templates.
+A second reason is that I believe it would help making Ember more approachable by newcomers
+that have _some_ experience in other frameworks. One of the most shocking moments that developers
+that are familiar with React, Vue or Angular experience when trying Ember is that they cannot,
+at least out of the box, perform the most basic logical comparisons and operation they are so used to
+in JSX or Vue/Angular templates.
 
-Furthermore, if the [RFC #367](https://github.com/emberjs/rfcs/pull/367) eventually gets merged, these
-helpers are the perfect candidates to create adoption friction in the community because of how
-pervasive its usage it in so many templates.
+A third reason is that if we implement those super common helpers in the Glimmer VM, that would open
+a important vector of low level optimization.
 
-Albeit less important, implementing these helpers in the core _could_ enable clever trickery in the
-Glimmer VM, like evaluating them at compile time when the arguments are constant.
+Consider the following template:
+
+```hbs
+{{#if (and @featureEnabled this.expensiveComputedProperty @model.asyncEDRelationship.length)}}
+  {{!-- some logic --}}
+{{else}}
+  {{!-- some logic --}}
+
+{{/if}}
+```
+
+Because of the way Ember helpers work, all their input parameters are eagerly evaluated by the
+Glimmer VM and passed to the helpers. This might include computationally expensive computed properties,
+or hitting code paths that trigger network requests.
+
+Implementing helpers like `and` or `or` at a lower lever would allow those helpers to be evaluated in
+short circuit, so if `@featureEnabled` is false, neither of the following arguments will ever evaluated.
+
+Going one step further into the optimizing compiler world, if **every** invocation of the component
+with this template receives `@featureEnabled={{false}}`, the compiler could even completely remove
+the conditional and the truthy branch of the if from compiled output.
+
+The last reason is that if the [RFC #367](https://github.com/emberjs/rfcs/pull/367) eventually gets merged,
+these helpers are the perfect candidates to create adoption friction in the community because of how
+pervasive its usage it in so many templates, forcing them to explicitly importing them on most templates
+or adding them to the proposed `prelude.hbs` file.
 
 ## Detailed design
 
@@ -77,10 +106,14 @@ addons.
 The main unresolved question is what helpers we deem worthy of being moved into the core, and also
 if we want any other helpers not mentioned above.
 
+In particular, I'm sitting on the fence about `not-eq`.
+It is not _really necessary_, in the same way `{{unless foo}}` can also be expressed as `{{#if (not foo)}}`,
+but it may be convenient.
+
 Other helpers that come to mind that _could_ also be worth adding:
 
-- `array` (there is an RFC for it: https://github.com/emberjs/rfcs/pull/318)
+- `array` (there is an [RFC](https://github.com/emberjs/rfcs/pull/318) and its being implemented)
 - `add`, `subtract`, and other arithmetic operators.
 - `{{#await promise as |value|}}` to render a block conditionally if a promise resolves or fails.
 
-If there is interest in them, I advocate creating standalone RFCS for them.
+If there is interest in them, I advocate creating standalone RFCS for each one of them.
--- a/text/0000-basic-template-helpers.md
+++ b/text/0000-basic-template-helpers.md
@@ -2,11 +2,11 @@
 - RFC PR: (leave this empty)
 - Ember Issue: (leave this empty)
 
-# Promote helpers from ember-truth-helpers into Ember
+# Add new basic template helpers to Ember
 
 ## Summary
 
-Add new built-in template helpers to perform basic operations in templates, identical to the
+Add new built-in template helpers to perform basic boolean operations and comparisons in templates, identical to some of the
 helpers in [ember-truth-helpers](https://github.com/jmurphyau/ember-truth-helpers).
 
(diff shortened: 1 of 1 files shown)


text/0000-basic-template-helpers.md
```diff
@@ -0,0 +1,86 @@
...
+Glimmer VM, like evaluating them at compile time when the arguments are constant.
+
+## Detailed design
+
+The process consists on deciding what helpers from `ember-truth-helpers` we consider the most important
+and move them into Ember.js itself or even the Glimmer VM, in a fully backwards compatible way.
+
+I propose to move to core at least:
+
+- `eq`
```
Turbo87: will that be `==` or `===` semantics?

text/0000-basic-template-helpers.md
```diff
@@ -0,0 +1,86 @@
...
+The main unresolved question is what helpers we deem worthy of being moved into the core, and also
+if we want any other helpers not mentioned above.
+
+Other helpers that come to mind that _could_ also be worth adding:
+
+- `array` (there is an RFC for it: https://github.com/emberjs/rfcs/pull/318)
+- `add`, `subtract`, and other arithmetic operators.
+- `{{#await promise as |value|}}` to render a block conditionally if a promise resolves or fails.
+
+If there is interest in them, I advocate creating standalone RFCS for them.
```
Turbo87: I agree, this RFC should be scoped to add the mentioned logic helpers, the others would need dedicated discussions.

text/0000-basic-template-helpers.md
```diff
@@ -0,0 +1,86 @@
...
+One of the shocking moments developers that are familiar with React, Vue or Angular have when trying Ember
+is that they cannot, at least out of the box, perform the most basic comparisons and boolean operations
+they are so used to in JSX or Vue/Angular templates.
+
+Furthermore, if the [RFC #367](https://github.com/emberjs/rfcs/pull/367) eventually gets merged, these
+helpers are the perfect candidates to create adoption friction in the community because of how
+pervasive its usage it in so many templates.
+
+Albeit less important, implementing these helpers in the core _could_ enable clever trickery in the
+Glimmer VM, like evaluating them at compile time when the arguments are constant.
```
rwjblue: Another example would be to short circuit in helpers like `eq`, `and`, and `or`. Today’s version of ember-truth-helpers _always_ evaluates all of the arguments (eg Ember.get’s them), but if implemented in the VM we could avoid evaluation of all arguments after we know the result. 

Take the following example:

```hbs
{{#if (and first second third)}}

{{/if}}
```

Today, we evaluate all of those args, but if implemented in the VM we can avoid evaluating `third` if `second` were falsey.

simonihmig: Somewhat related to the `and` semantics: I just saw that the current implementation follows the JS semantics in that it does not necessarily return a strict boolean value, but rather the last argument (if evaluating to a truthy value) or the last falsy argument (if evaluating to false). So `(and "a" "b")` would evaluate to `"b"` (similar to `'a' && 'b'`).

I am not entirely sure I would copy over these JS semantics to hbs-land. At least it was a bit unexpected for me, as I have never used `and` or `or` outside of an `{{#if}}` block. But that can be discussed. If these semantics should remain, this could maybe be pointed out in the RFC, for guiding the future implementation in core...

cibernox: IMHO the current behavior makes sense. It would be very limiting to coerce values to a boolean. I've used the helper that way.

text/0000-basic-template-helpers.md
cibernox: If we want to maintain backwards compatibility it has to be `===`. And even if compatibility wasn't a concern I'd still prefer triple equal

webark:
#152

just including the existing open RFC on this topic for cross reference, especially with regards to the “short circuiting”, abd why this needs to be done in core space... 😞

cibernox:
@webark I've expanded the information about short circuiting and how it can help making apps do less work. 
I've also expanded the information about how an optimizer compiler (that right now is that smart, but it maybe some day) could even completely remove some branches in compile time from the program altogether.

It would be nice if all helpers could short circuit by default, but that's just one of the four reasons to include those helpers into templating engine. I think it's still valuable even disregarding possible optimizations. Allowing all helpers to pull arguments as they use them would be a separate RFC.

webark:
@cibernox thanks. and ya. i do t want to get into a “purity” debate. I (having never used ember truth helpers in many apps) always found that if i wanted to reach for one of these, more often then not it was due to a data structure that needed improvement upon revisiting, or that the chunk of template where i was wanting these warranted a new component itself. I will miss this “cenvention check”, but i feel the community has clearly spoken in favor of this, and there are clear benefits to including it in core. so i’m 👍 even if it’s with a 😞

Gaurav0:
Why not just make ember-truth-helpers part of the default blueprint?

cibernox:
@Gaurav0 in the **Alternatives** section that possibility is discussed and I expose some arguments why I think this should be in the core.

jonnii:
I also find myself by default installing ember-composable-helpers (https://github.com/DockYard/ember-composable-helpers) whenever starting a new ember applications. If there's an appetite for adding new helpers I'd also advocate for:

 * toggle
 * optional
 * contains
 * inc
 * dec

derekwsgray:
> @cibernox thanks. and ya, i don't want to get into a “purity” debate. I (having never used ember truth helpers in many apps) always found that if i wanted to reach for one of these, more often then not it was due to a data structure that needed improvement, or that the chunk of template where i was wanting these warranted a new component itself. I will miss this “convention check”

Totally agree. Have never used them, and found whenever someone on my team tried to add them it was inevitably !DRY and/or like you point out above. If they're going in, I'd much prefer default blueprint over core, so I could remove them. I do agree there are some decent use-cases I've been shown, but in general they require too much discipline to keep things from getting hairy. Or if there was an addition to the ember linter plugin...

rwjblue:
> Or if there was an addition to the ember linter plugin...

FWIW, we do already have a similar concept in the `simple-unless` rule. The goal is to essentially limit the totally amount of complexity in logical expressions with `{{unless` but as you might imagine I believe that these logical expressions could use the same underlying logic to determine “complexity”...

cibernox:
@derekwsgray There is one particular use case that is very annoying to do without some basic logic helpers:

```hbs
{{#each jsonPayload.users as |user|}}
    {{#if (eq user.type "admin")}} 
      Some admin stuff
    {{else}}
      Non admin stuff.
    {{/if}
{{/each}}
```
You want to do some very simple check on every iteration. You can do it creating a component for the loop or wrapping your POJOs (to create a `isAdmin` computed property for instance). You can find ways around without adding any logic to your templates, but in those scenarios a little bit of logic in the template is perfectly acceptable. I think that 100% logicless templates can go against convenience in this case.

Gaurav0:
One of the problems with adding stuff like this to core is what happens when a user defines a helper with the same name? Which one wins? How do we ensure that users update ember-truth-helpers to a version without "core" ones?

cibernox:
@Gaurav0 From the RFC.
> When this feature is implemented, we would update ember-truth-helpers to automatically remove the promoted helpers from the code when the Ember is above a certain version number.

This update can happen in a patch as it's not a breaking change and anyone that has a floating version number will eventually get it. We might want to be smart and detect that if they are only using of the ported helpers (no `is-equal` or `is-empty`...) and ask them to remove the addon.

Gaurav0:
>When this feature is implemented, we would update ember-truth-helpers to automatically remove the promoted helpers from the code when the Ember is above a certain version number.

Doesn't this rely on the user updating to the new patch version of ember-truth-helpers? (Not guaranteed, I feel like I'm the only one at my company who ever tries to yarn upgrade.)

cibernox:
@Gaurav0 correct, if they don't update and the user-defined helpers clover the default ones (which I don't know if it's the case) they wouldn't get the improvement.
Would you prefer Ember throwing an error if a user defines a helper whose name collides with a default one? That error could ask people to update (or remove) their version of ember-truth-helpers.
That approach sounds like it would be a bit more annoying but would bring more people on the latest best practice.
I also believe that `ember-cli-update`(which is becoming more popular by the day is it is the recommended way to update versions now ) can run additional tasks, and one of them could bump ember-truth-helpers to the latest version if your app happens to use it.

Gaurav0:
>Would you prefer Ember throwing an error if a user defines a helper whose name collides with a default one?

I think that would be a great idea. Of course, it would have to be a deprecation warning in the current 3.x cycle, and an error in 4.x actually. Too late for this situation.

derekwsgray:
Agree with the first part of @webark - the 'admin' example given is not great. The templates become harder to maintain against code changes as you say.

Download numbers don't really sway my opinion (if they did, I'd be using a different framework, har har!). I respect the thought-leaders and proponents of best-practices. If the core ember team thinks this is a good idea, I respect that.
I do concede the argument for "attracting more people in from other frameworks" is persuasive (though doesn't speak to whether this should be blueprint vs. core.)

text/0000-basic-template-helpers.md
```diff
@@ -0,0 +1,86 @@
...
+## Unresolved questions
+
+The main unresolved question is what helpers we deem worthy of being moved into the core, and also
+if we want any other helpers not mentioned above.
+
+Other helpers that come to mind that _could_ also be worth adding:
+
+- `array` (there is an RFC for it: https://github.com/emberjs/rfcs/pull/318)
+- `add`, `subtract`, and other arithmetic operators.
+- `{{#await promise as |value|}}` to render a block conditionally if a promise resolves or fails.
```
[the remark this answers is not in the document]
chancancode: I agree 😄 👉 👉 👉 http://github.com/tildeio/ember-async-await-helper/

mmun:
I've been advocating for this for years. Thanks for writing it up.

rondale-sc:
I created a little polyfill to make these helpers available in glimmer applications:

https://github.com/rondale-sc/glimmer-boolean-helpers-polyfill

Eventually would be neat to have `ember-truth-helpers` depend on this package to share implementation and have a central location where we could inactivate if/when glimmer-vm has these helpers.

mixonic:
Should these be operators instead of helpers? Is is potentially confusing that `(and a b)` would not realize the value of `b`, but `(someBespokeHelper a b)` would?

An operator alternative would be something like `(a && b)`, which is clearly not a helper.

rwjblue:
> Is is potentially confusing that `(and a b)` would not realize the value of b, but `(someBespokeHelper a b)` would?

Firstly, I think the fact that it does or does not evaluate `b` is not terribly important to the programming model (just like many of the other optimizations that glimmer-vm implements). 

Secondly, I think instead of moving to create a whole new construct we should instead _make_ it possible to do in `someBespokeHelper`. Specifically, this would mean making a "helper manager" and adding a capability to receive references instead of values (this exact concept was already mentioned as part of the custom components RFC [here](https://github.com/emberjs/rfcs/blob/master/text/0213-custom-components.md#follow-up-rfcs)).

cibernox:
Update: Since feedback so far has been very positive, I went ahead and implemented those helpers in a PR: https://github.com/emberjs/ember.js/pull/17177/files

rondale-sc:
Updated glimmer-boolean-helpers-polyfill to include `not-eq`:

https://github.com/rondale-sc/glimmer-boolean-helpers-polyfill/tree/v1.1.1

sukima:
**Disclaimer:** I write this solely for *Devil's Advocacy*. I wanted to make sure that other side of the coin was represented before this merged. In truth I *personally* do not hold a stake in this PR. I'd be happy either way.

There is one argument that @webark eluded to in that adding more out-of-the-box logic operators to the templates can (and in my experience will) result in abuse of them. What I mean by this is that in many of the apps I've used that had truth-helpers the least friction to writing code is to have complicated and highly nested blocks in templates. Mainly because when your building out a component typically a coder is already in the thick of HTML markup and the last thing they want to do is open up another file, consider a name, and context switch to another language.

The unfortunate side-effect is in maintenance. When I see a rendered ember app I know where I want to make a change. Often I will open the template and think oh it is right inside that bold tag or in that `<section>` only to find a 100+ line block of `if/else/eq/not-eq/and/or/etc.` and get completely lost.

Now this is not saying truth helpers are to blame. Obviously this is a factor of lazy programming. However, it may behoove us as leaders in the ecosystem to offer alternatives. Truth helper although provide many benefits they also can help facilitate people shooting themselves in the foot. If the de-facto standard is that any logic lives in JS land while HBS land is solely for interfacing with those values it means we've pushed the whack-a-do to JS and the HBS can remain more focused on layout then on business logic.

Ok, *devil's advocate* hat off... to bring this back into focus I don't believe it is reasonable for the lack of a feature to drive good coding practices. As such the framework should facilitate flexibility to reach as many situations as possible and adding truth-helpers is a huge step in that direction. In fact there are cases when having truth helpers it improves readability and code quality. I am :+1: this RFC.

My take away is that as a framework we provide features and flexibility to facilitate better ergonomics. As leaders in the ecosystem it is our responsibility to teach and promote good coding practices. I would then propose that we provide some disclaimers in the guides to explain that although truth-helpers are flexible and useful that a programmer should be weary about abusing them in templates and consider moving complicated logic into the JS file where it is easier to  reason about.

locks:
@sukima
> There is one argument that @webark eluded to in that adding more out-of-the-box logic operators to the templates can (and in my experience will) result in abuse of them.

I think it is important to keep in mind that this RFC does not add new capabilities to templates. You can already create your own helpers, or install addons with collections of them, both are a command away.
For this reason, I don't think the slippery slope argument is particularly strong in this case.

> If the de-facto standard is that any logic lives in JS land while HBS land is solely for interfacing with those values it means we've pushed the whack-a-do to JS and the HBS can remain more focused on layout then on business logic.

I think stating it as "any logic lives in JS" erases some nuance that you touch upon afterwards.
There is presentation logic and business logic, and the tension between the two is the everyday challenge that we face when developing applications.

We already try to communicate in the project's documentation (and [linting rules](https://github.com/emberjs/rfcs/pull/388#issuecomment-430080203)) that you should not overcomplicate template logic and keep it to presentational concerns.
I would like to add that having these helpers be part of the core package, and documented as such, should actually actually contribute to more widespread knowledge and practices.

chancancode:
For the template complexity argument: I think as with functions in code, when a template gets complex, it's a good time to refactor them into smaller pieces. Sometimes, that means extracting logic into custom helpers, and that's perfectly fine. But increasingly more often, the answer is the break things down into smaller components, just as you would break down big functions into smaller ones. And in (very) modern Ember, we have a very good solution for that – template-only components. With these, it becomes super easy to pull parts of your templates out into their separate files.
    
In this world, my opinion is _not_ having these helpers would end up hurting template readability. It will add enough frictions for those simple components use cases that you may just not want to do the extraction anymore (you would have to have to add an extra JavaScript file just so you can put a silly computed property (or computed macros?) in it).
    
Reducing the arbitrary "cliff" of when you need to "fallback" into JavaScript for simple things that you would like to express in templates is ultimately a win for templates readability imo.

chancancode:
I think we should consider using `neq` instead of `not-eq` to match the verbosity of the other helpers (otherwise, why not `lt-eq`?).

chancancode:
It was proposed that `eq` should use `===` semantics, which I think is probably a good idea. However, what about `lte` or `gte`? What should `(lte false "")` or `(gte false 0)` be? Should it be `true`, `false` or an error? Is it really okay that the ordering operators disagree with the equality operators? (It is illegal/undefined behavior for [`Eq`](https://doc.rust-lang.org/std/cmp/trait.Eq.html) to disagree with [`Ord`](https://doc.rust-lang.org/std/cmp/trait.Ord.html) in Rust, for example.) I suspect there will be some pretty bad composition fallout here if we are not careful.

chancancode:
How many arguments should these each of these helpers take? The straightforward answer would be exactly two. However, for `eq`, `and`, `or`, etc, that seems a bit arbitrary and it _seems_ like a good idea to allow more than two arguments. Should we allow one or zero arguments? What should their evaluation value be? Or should it be an error? A runtime error or a compile-time syntax error? Should `lt`, `lte`, `gt` and `gte` also take multiple arguments? What are the semantics there?

chancancode:
Would these operators implicitly coerce the result into booleans, or would it passthrough the operands? It _seems_ like a good idea to _not_ do that, for use cases like `(or @posts @reblogs)` where you would expect this to pass through _either_ the value of `@posts` or `@reblogs`. However, what should this evaluate into if both of those are falsy? Should it be `false`, `undefined` or the value of the last falsy argument (whatever `@reblogs` is)? What about `not`? Should `(not @foo)` be `(if @foo ??? @foo)` or `(if @foo false true)`? What about `lt`, `lte`, `gt` and `gte`?

chancancode:
Most importantly, what truthiness semantics should these helpers (at least `and`, `or`, `not`) assume? Currently, the handlebars truthiness semantics consider the JavaScript falsy values *as well as* empty arrays falsy. Presumably, these logic operators have to agree with the existing boolean semantics otherwise it would be really award if the `if` operator disagree with the logic operators. However, the Ember also added *proxies*, to make things like Ember Data relationships work. (see https://github.com/emberjs/ember.js/blob/master/packages/%40ember/-internals/glimmer/tests/utils/shared-conditional-tests.js).This is super difficult to support, but IMO we may not have much of a choice here. I think a consistent boolean semantics across the language is pretty important, and we should just figure out how to deprecate and remove proxies from Ember across the board when the time is right.

chancancode:
I think as it is written today, this RFC left some pretty important semantics under-specified (I didn't realize the reach and impact of these questions until I start following the implementation work). As I mentioned in my first comment, I would like to see these helpers added, but I think we should halt the FCP for now and restart it only when we have satisfactory answers to all of these questions.

cibernox:
@chancancode I updated the RFC with details of the implementation of each helpers. I kept the semantics in `ember-truth-helpers` with the only edition of lazy evaluation.
That means in my update the definition of truthyness still includes empty arrays, which is the only part that makes me cringe.

Gaurav0:
@chancancode I don't really have any opinion on `neq` vs `not-eq`, but given that ember-truth-helpers is in wide use let's stick with `not-eq` so people don't have to do anything to migrate.

wycats:
> I think as it is written today, this RFC left some pretty important semantics under-specified (I didn't realize the reach and impact of these questions until I start following the implementation work). As I mentioned in my first comment, I would like to see these helpers added, but I think we should halt the FCP for now and restart it only when we have satisfactory answers to all of these questions.

In light of @chancancode's concerns about the need to specify these helpers with the existing truthiness semantics of Ember, I'm going to halt the FCP. I hope that we can wrap up the specification quickly and re-proceed to FCP.

Gaurav0:
> Most importantly, what truthiness semantics should these helpers (at least and, or, not) assume? Currently, the handlebars truthiness semantics consider the JavaScript falsy values as well as empty arrays falsy. Presumably, these logic operators have to agree with the existing boolean semantics otherwise it would be really award if the if operator disagree with the logic operators. However, the Ember also added proxies, to make things like Ember Data relationships work. (see https://github.com/emberjs/ember.js/blob/master/packages/%40ember/-internals/glimmer/tests/utils/shared-conditional-tests.js).This is super difficult to support, but IMO we may not have much of a choice here. I think a consistent boolean semantics across the language is pretty important, and we should just figure out how to deprecate and remove proxies from Ember across the board when the time is right.

I've been using ember-truth-helpers for some time, and I find this surprising. I'm wondering how many apps/addons rely on this stuff.

cibernox:
@wycats let me know when you discuss it, specially the truthiness part.

sukima:
Hold on!

> we should just figure out how to deprecate and remove proxies from Ember across the board

Is unfair. Proxies serve a real world problem and are well understood as a Object Oriented Pattern. So much so that they are now part of the JavaScript core specs. Why would we get rid of them?!

* https://www.youtube.com/watch?v=sClk6aB_CPk
* https://medium.com/dailyjs/how-to-use-javascript-proxies-for-fun-and-profit-365579d4a9f8
* https://www.oodesign.com/proxy-pattern.html

cibernox:
@sukima I believe @chancancode was talking about `Ember.ObjectProxy`, `Ember.ArrayProxy` and `Ember.PromiseProxyMixin`, not about ES6 proxies.

rwjblue:
@sukima - FWIW, what @chancancode was referring to was to removing the (fairly archaic at this point) custom proxy concepts in _favor of_ "just using the platform", **but** it is also out of scope for this RFC...

cibernox:
What is the state of the discussion of truthiness for these new helpers? I believe some helpers like `{{or}}` will become even more common give default values once `@`arguments are more used in glimmer components:

```hbs
<button type={{or @type "button"}}>...</button>
```

GavinJoyce:
> 2\. We are copying ember-truth-helpers

I don't have any reason to think that any helpers in `ember-truth-helpers` have an undesirable implementation, I think that we should be careful to ensure that we are designing the behaviour of these new built in helpers from first principles though.

Panman8201:
> What is the state of the discussion of truthiness for these new helpers? I believe some helpers like `{{or}}` will become even more common to give provide default values once `@`arguments are more used in glimmer components:
> 
> ```
> <button type={{or @type "button"}}>...</button>
> ```

Related discussion here: https://github.com/emberjs/rfcs/issues/479

IMO, we should also add a `{{default}}` helper (or such) to return the second parameter when the first is `undefined`. Because, `{{or}}` would return the second parameter when the first is falsy, which might be needed.

miguelcobain:
> IMO, we should also add a {{default}} helper (or such) to return the second parameter when the first is undefined. Because, {{or}} would return the second parameter when the first is falsy, which might be needed.

In database systems there is an operation called `coalesce`. It takes N arguments and returns the first that is non-null.
Example in Postgres: https://www.postgresql.org/docs/9.2/functions-conditional.html#FUNCTIONS-COALESCE-NVL-IFNULL

I could see myself using a `coalesce` helper that returns the first non-undefined argument.
Example:
```hbs
{{coalesce @arg1 @arg2 this.value "default value"}}
```

cibernox:
I think that a default helper could be nice in cases where false is a valid value. 

Regardless of that, we must resuscitate the discussion about truthyness. As far as I can remember that was the only controversy 

/cc @chancancode @wycats

pzuraq:
I think a `{{nullish}}` coalescing operator may be more inline with where JS is heading (though that is only a stage 1 proposal still). I also think it may be a better idea to try to get the primitive/basic helpers finalized and added first, before we look into adding a defaulting helper.

text/0000-basic-template-helpers.md
```diff
@@ -0,0 +1,178 @@
...
+
+- `eq`
+- `not-eq`
+- `not`
+- `and`
+- `or`
+- `gt` and `gte`
+- `lt` and `lte`
+
+Those helpers are very low lever and generally useful in both Ember and Glimmer.
```
locks:
```suggestion
Those helpers are very low level and generally useful in both Ember and Glimmer.
```

mehulkar:
Apart from the truthiness discussion, adding more helpers blurs the line between Handlebars and Ember's flavor of Handlebars. I imagine as more added, it will be confusing to user just where to go for documentation on any part of the template they are looking at.

An explicitly allowed list of template helpers (or template imports or strict mode #496) would help mitigate this problem, as it would be easier to see where these magic bits come from.

Or we could decide this is a moot point, considering that package.json contains "ember-cli-htmlbars" and not "handlebars".

Panman8201:
FWIW, I'd like to see this move forward so that I can support optional dynamic tag name (once implemented) in a template only Glimmer Component from an addon without forcing another dep by using [ember-truth-helpers](https://github.com/jmurphyau/ember-truth-helpers).

```hbs
{{#let (element (or @tagName "div")) as |DynamicElement|}}
    <DynamicElement />
{{/let}}
```

chriskrycho:
@Panman8201 I believe you could just use an inline `if` to work around as well?

```hbs
{{#let (element (if @tagName @tagName "div")) as |DynamicElement|}}
    <DynamicElement />
{{/let}}
```

(I'd also like this to move forward, just noting it's not *blocking*!)

text/0000-basic-template-helpers.md
```diff
@@ -0,0 +1,178 @@
...
+{{else}}
+  {{!-- some other logic --}}
+{{/if}}
+```
+
+Because of the way Ember helpers work, all their input parameters are eagerly evaluated by the
+Glimmer VM and passed to the helpers. This might include computationally expensive computed properties,
+or hitting code paths that trigger network requests.
+
+Implementing helpers like `and` or `or` at a lower lever would allow those helpers to be evaluated in
```
pangratz:
```suggestion
Implementing helpers like `and` or `or` at a lower level would allow those helpers to be evaluated in
```

mcfiredrill:
Is something blocking this? It's unclear to me what the status is.

mixonic:
I raised this in a few recent FW core meetings and it fell to @rwjblue and I to leave some of the feedback here. I'll work to get that done today or Friday.

cagrimmett:
I'm a new Ember dev and we use @rwjblue's `ember-truth-helpers` addon (and I didn't know it was an addon, assumed it was core.) When I spun up a side project app, I was surprised/frustrated when these basic helpers weren't working. Should definitely be part of core! I use it in at least half of my templates.

mixonic:
Framework core discussed this at the September F2F: https://github.com/emberjs/core-notes/blob/master/ember.js/2019-09/F2F_Ember_Core_Team_20190915.md#basic-helpers-rfc

I'll try to summarize: @chancancode left a string of comments [above](https://github.com/emberjs/rfcs/pull/388#issuecomment-436378641), and he raised a few different points which we think remain unaddressed. That said, I think comments were a bit ad-hoc in their presentation and the F2F conversation may help clarify the over-arching concerns. 

My goal in raising the discussion at F2F was to try and find a path forward, and I think everyone is aligned on that goal. We all want to see some more convenience helpers added to the core framework.

### HBS v. JS truthiness, `{{and}}` `{{or}}` `{{not}}`

@chancancode raised the questions of "What truthiness semantics should these helpers (at least `and`, `or`, `not`) assume?" [here](https://github.com/emberjs/rfcs/pull/388#issuecomment-436386850). Just below @cibernox explained "`eq` behaves like `===` and `not-eq` behaves like `!==`, just like in ember-truth-helpers."

The problem of adopting JS semantics is that HBS **already has truthiness semantics of its own**, at least of some of the cases here. Where possible **we would prefer to base additional helpers on current semantics**. This helps the template language maintain cohesion and allows to us implement more complex helper atop simple tools.

For example instead of defining `(and)` by [deferring to JS `&&` semantics](https://github.com/cibernox/rfcs/blob/new-basic-handlebars-helpers/text/0000-basic-template-helpers.md#and) as currently suggested it should be defined via HBS semantics:

```hbs
{{and a}} is an expression of {{if a a false}}
{{and a b}} is an expression of {{if a (if b b false) false}}
{{and a b c}} is an expression of {{if a (if (if c c false) false) false}}
```

By expressing the design as HBS we can avoid needing to consider differences in how HBS and JS may decide truthiness, and the question of "what value is returned in what case" becomes explicit. The glimmer engine can additionally implement optimizations like shortcutting evaluation of subexpressions as a VM optimization instead of deferring to JS.

Of course just like JS truthiness can be subtle and sometimes confusing, HBS truthiness is probably the same. Some notable warts are empty array behaviors and the difference in the behavior of a blank string vs. a blank `SafeString`. Any proposal for building on `(if` should at least enumerate the current behaviors, but probably should consider changing those behaviors out of scope.

This is also trivially possible with `(not)`: 

```hbs
{{not a}} is an expression of {{if a false true}}
```

`{{or}}` is a bit more complex but still possible:

```hbs
{{or a}} is an expression of {{if a a false}}
{{or a b}} is an expression of {{if a a (if b b false)}}
{{or a b c}} is an expression of {{if a a (if b b (if c c false))}}
```

Given that these are a reasonable design compromise (which I expect they are?) we can look at more meaty challenges. I think a stand-alone RFC for `(and)` `(or)` `(not)` based on HBS semantics would be easy to advance.

### `{{eq}}` `{{neq}}`

We spend some time at the F2F thinking about if HBS already has equality semantics. The consensus answer is "probably not", although there was some suggestion you could consider the `{{#each}}` DOM stability semantics as being dependent on some measure of value equality.

After discussion there was some loose consensus that `===` in JS would be appropriate for template equality semantics.

An open issue was the equality of safe strings. Does `{{eq htmlSafeStringOfFoo anotherHtmlSafeStringOfFoo}}`? Since creating a safe string creates an object a `===` comparison would have two safe strings from the same actual string fail an equality comparison.

`{{neq}}` could be implemented atop the `{{eq}}` semantics as `{{if (eq a b) false true}}`

### `{{lt}}` `{{gt}}`

These helpers also require a definition. I think there is some amount of discussion possible around if non-numeric values should ever be permitted as arguments. In JS `'abc' > NaN === false`. Does that make any sense? Maybe we should not lean on the JS versions of these operators as we build the HBS versions.

### How to move forward

My own proposal to get work landed in the quickest manner and to keep the conversations focused is that this work be split into 3 RFCs.

* RFC: Adding Logical Operators to Templates (`{{and}}` `{{or}}` `{{not}}`)
  * This should enumerate examples of what is truthy and falsy in the current implementation
* RFC: Adding Equality Operators to Templates (`{{eq}}` `{{neq}}`)
* RFC: Adding Numerical Comparison Operators to Templates (`{{lt}}` `{{gt}}`)

This is only my suggestion, as I think the logical operator RFC would be trivial where the other two will benefit from being more focused. At the least I know it would aid framework core in having discussion and moving each forward to consensus.

cibernox:
Great to see that this can finally move. My only concern with those helpers obeying handlebars'  trutiness semantics is that it is different from what `ember-truth-helpers` do. 

We'd have to plan to allow people to migrate without too much pain. I'm particularly worried about addons that use ember-truth-helpers, not so much about application code.

I agree that splitting this RFC into three RFCs makes things easier. The one for the equality operators a the one for numerical comparisons are less controversial as they would be identical to those in Ember-truth-helpers and can be implemented quickly.

I'll try to write all/some of those RFCs this week, but by any means if anyone wants to help, by all means, just call dibs on one of them in this thread.

cibernox:
@mixonic I've open three separate RFCs as suggested, feel free to close this one now.

rwjblue:
Thanks @cibernox! We discussed this in todays core team meeting and are moving this RFC into "FCP to Close" (in favor of the broken out / smaller RFCs).

Panman82:
Any thoughts on where a `{{coalesce}}` helper might best fit in the separated RFCs? Or maybe it's own RFC? Or not at all?? (It was just a discussion I brought up earlier.) I've created a Gist of what I figured it would look like / how it would function: https://gist.github.com/Panman82/04d96915d8854f20ede4a4788fcb11e3

locks:
@Panman82 I would submitted it on its own, as I think it would benefit from focused discussion!

locks:
The FCP period has lapsed without further discussion, so I'll be closing this RFC.
Thanks everyone for participating, see you in the discussions for https://github.com/emberjs/rfcs/pull/560, https://github.com/emberjs/rfcs/pull/561 and https://github.com/emberjs/rfcs/pull/562 !

PR #391, opened 2018-10-22
Router Helpers

chadhietala:
[Rendered](https://github.com/emberjs/rfcs/blob/new-router-helpers/text/0000-router-helpers.md)


rwjblue:
I'm definitely in favor of this, thanks for working on it!

raido:
Less magic and more control for advanced UX 👍

text/0000-router-helpers.md
```diff
@@ -0,0 +1,442 @@
...
+**Before**
+
+```hbs
+{{#link-to 'profile'}}Profile{{/link-to}}
+```
+
+**After**
+
+```hbs
+<a href={{url-for 'profile'}} class={{is-active 'active'}}>Profile</a>
```
gossi: Is this correct for `{{is-active}}`? According to the signature you mentioned above, this should be `class={{if (is-active 'profile') 'active'}}` or is there some magic going on between `is-active` and `url-for` helper? I'd like the short syntax you used here, thought then it looks like overloading in that very special location?

gossi:
How does it related to #339 ? Will it make that one obsolete?

text/0000-router-helpers.md
rwjblue: Agree here, I believe it should be updated to use the inline `if` snippet above.

lennyburdette:
Is `is-active` too generic of a name? Would `is-route-active` be more appropriate? 

(Or will I not care soon because it will be a matter of `{{use is-active as is-route-active from "@ember/route-helpers"}}` to distinguish them? 😁)

luxferresum:
So I think the primary difference to #339 is that this assumes ordered params and not named params?

juggy:
I am not 100% sure what `transition-to` adds to the html of the anchor. Does it add the `href` url and an action? Something like:
```
<a href={{url-for "my.route" model}} onclick={{transition-to-action 'my.route' model}}> Link </a>
```
So it would be a shortcut?  (made up `transition-to-action` for the example) 

Can I have a component use the `transition-to` as an action?
```
{{my-comp onSave=(transition-to 'my.route')}}
```

chadhietala:
@gossi and @luxferresum that is the primary difference in regards to the helpers, but this RFC also makes `<a href="/">Home</a>` also "just work".

@lennyburdette these APIs are intended to map directly onto those provided my the [router service](https://github.com/emberjs/rfcs/blob/master/text/0095-router-service.md#routerservice).

@juggy Yes it adds the `href` in the modifier form. I think we could have a sub-expression API as well.

lennyburdette:
Mirroring the router service API makes a lot of sense, but I would argue that the API is actually `router.isActive()` not just `isActive` on its own.

I also agree that a `transition-to` helper would be helpful in addition to the `transition-to` element modifier, but I am concerned about creating another point of confusion similar to `{{action}}`/`(action)`. Maybe it'll be clearer after the Element Modifier Manager RFC is implemented and element modifiers become a documented feature of the framework?

sandstrom:
A thought on the drawback:

> By proxy this may cause people to encapsulate all of these primitives into a single component and thus creating a user-land version of {{link-to}}. -- this RFC

The documentation around this could include a short snippet on how to put together a link-to component (under an 'advanced' section or similar), as a starting point for anyone that needs to encapsulate app-specific logic into a link component.

*(I also don't think this drawback/scenario is very likely, most people should be fine with `<a {{transition-to 'profile' model}}>Profile</a>`)*

rwjblue:
@sandstrom

> The documentation around this could include a short snippet on how to put together a link-to component (under an 'advanced' section or similar), as a starting point for anyone that needs to encapsulate app-specific logic into a link component.

There is a “kitchen sink” example [here](https://github.com/emberjs/rfcs/blob/new-router-helpers/text/0000-router-helpers.md#kitchen-sink) which shows **all** the things that a `{{link-to` does.

alexparker:
> This RFC expands the surface area of the templating layer by exposing the primitives that make up {{link-to}}
Would also be worth exploring not just breaking apart the helpers, but parts for building ones own light-weight link components. I recently just ended up creating a `nav-link` component of my own instead of link-to and using the new `router` service api has been really really nice with that.

sandstrom:
@alexparker Perhaps I misunderstand your question, but the purpose of this RFC is precisely about making it easier for you to make your own nav-link components.

text/0000-router-helpers.md
```diff
@@ -0,0 +1,442 @@
...
+
+### Anchor Tags
+
+We currently do not have a good solution for transitioning solely based on HTML anchors defined in the templating layer. For instance let say you are using [Ember Intl](https://github.com/ember-intl/ember-intl) to do internationalization for your application. Ember Intl uses the [ICU message format](http://userguide.icu-project.org/formatparse/messages) for the actual translation strings and supports having HTML within the string. Now lets say you want to put a link in a translation string and have it work like `{{link-to}}` works. In that case you either have to roll your own solution or use something like [Ember-href-to](https://github.com/intercom/ember-href-to). Another example where this would be useful is that links within markdown produced by addons like [Ember-CLI-Showdown](https://github.com/gcollazo/ember-cli-showdown) would just work. API's like `RouterService#transitionTo` can transition an application using relative URLs and we have an opportunity to leverage this functionality to support this use case.
```
locks:
```suggestion
We currently do not have a good solution for transitioning solely based on HTML anchors defined in the templating layer.
For instance, let's say you are using [Ember Intl](https://github.com/ember-intl/ember-intl) to do internationalization for your application.
Ember Intl uses the [ICU message format](http://userguide.icu-project.org/formatparse/messages) for the actual translation strings and supports having HTML within the string.
Now, let's say you want to put a link in a translation string and have it work like `{{link-to}}` works.
In that case, you either have to roll your own solution or use something like [Ember-href-to](https://github.com/intercom/ember-href-to).
Another example where this would be useful is that links within markdown produced by addons like [Ember-CLI-Showdown](https://github.com/gcollazo/ember-cli-showdown) would just work.
APIs like `RouterService#transitionTo` can transition an application using relative URLs and we have an opportunity to leverage this functionality to support this use case.
```

text/0000-router-helpers.md
```diff
@@ -0,0 +1,442 @@
...
+
+### Extensibility Of `{{link-to}}`
+
+In 2.11 we moved `LinkComponent` from `private` to `public` largely because there was no other way to modify the behavior of `{{link-to}}` and it had effectively become de-facto public API. That being said, it is less than desirable to `reopen` or `extend` framework objects to gain access to the functionality to create some application specific primitive. For example [Ember Bootsrap](https://github.com/kaliber5/ember-bootstrap/) extends the [`LinkComponent`](https://github.com/kaliber5/ember-bootstrap/blob/master/addon/components/base/bs-dropdown/menu/link-to.js) and then layers more functionality on top of it. Addons would be better served if they had access to more primitive functionality.
```
locks:
```suggestion
In 2.11 we moved `LinkComponent` from `private` to `public` largely because there was no other way to modify the behavior of `{{link-to}}` and it had effectively become de-facto public API.
That being said, it is less than desirable to `reopen` or `extend` framework objects to gain access to the functionality to create some application specific primitive.
For example, [Ember Bootstrap](https://github.com/kaliber5/ember-bootstrap/) extends [`LinkComponent`](https://github.com/kaliber5/ember-bootstrap/blob/master/addon/components/base/bs-dropdown/menu/link-to.js) and then layers more functionality on top of it.
Addons would be better served if they had access to more primitive functionality.
```

text/0000-router-helpers.md
```diff
@@ -0,0 +1,442 @@
...
+
+`{{link-to}}` adds some convienent, yet not obvious, classes to the element. These classes are:
+
+- `active`: applied to any `{{link-to}}` that is on the "active" path
+- `disabled`: applied depending on the evaluation of `disabled=someBool`
+- `loading`: applied if one or more of the models passed are `undefined`
+- `ember-transitioning-in`: applied to links that are about to be `active`
+- `ember-transitioning-out`: applied to links that are about to be deactivated
+
+The issue with these class names is that they are not declared anywhere in your templated and are provided by the `LinkComponent` as `classNameBindings`. This effectively creates a set of reserved class names that are highly prone to colissions in your typical application.
```
locks:
```suggestion
The issue with these class names is that they are not declared anywhere in your templated and are provided by the `LinkComponent` as `classNameBindings`.
This effectively creates a set of reserved class names that are highly prone to collisions in your typical application.
```

text/0000-router-helpers.md
```diff
@@ -0,0 +1,442 @@
...
+
+Furthermore, addons like [ember-cli-active-link-wrapper](https://github.com/alexspeller/ember-cli-active-link-wrapper) and [ember-bootstarp-nav-link](https://github.com/zoltan-nz/ember-bootstrap-nav-link) do a ton of work arounds to get things like the `.active` class to show up on wrapping elements instead of the element directly. This is a great example that shows we are missing some primitives.
```
locks:
```suggestion
Furthermore, addons like [ember-cli-active-link-wrapper](https://github.com/alexspeller/ember-cli-active-link-wrapper) and [ember-bootstarp-nav-link](https://github.com/zoltan-nz/ember-bootstrap-nav-link) do a ton of workarounds to get things like the `.active` class to show up on wrapping elements instead of the element directly.
This is a great example that shows we are missing some primitives.
```

text/0000-router-helpers.md
```diff
@@ -0,0 +1,442 @@
...
+{{#link-to 'profile'}}Profile{{/link-to}}
+```
+
+In the DOM you will have an `href` on the anchor that gets serializes as:
+
+```html
+<a href="/profile?someBool=true" class="active ember-view">Profile</a>
+```
+
+Looking at a template you would have no idea that rendering the `{{link-to}}` would result in the query params being serialized. From an implementation point of view, this is problematic as we are forced to `lookup` the `Route` and the associated `Controller` to grab the query params. This can add a non-trivial amount of overhead during rendering, especially if you have many `{{link-to}}`s on a route that link many different parts of your application. As a side-note, this is one of the things  that needs to be dealt with if we are ever to kill controllers.
```
locks: Does it have to be "kill controllers"? :X

```suggestion
Looking at a template you would have no idea that rendering the `{{link-to}}` would result in the query params being serialized. From an implementation point of view, this is problematic as we are forced to `lookup` the `Route` and the associated `Controller` to grab the query params. This can add a non-trivial amount of overhead during rendering, especially if you have many `{{link-to}}`s on a route that link many different parts of your application. As a side-note, this is one of the things that needs to be dealt with if we are ever to kill controllers.
```

text/0000-router-helpers.md
```diff
@@ -0,0 +1,442 @@
...
+
+### Does Not Work With Angle Bracket Invocation
+
+Since angle bracket invocation does not support positional params, `{{link-to}}` has to adapt it's public API.
```
locks:
```suggestion
Since angle bracket invocation does not support positional params, `{{link-to}}` has to adapt its public API.
```

text/0000-router-helpers.md
```diff
@@ -0,0 +1,442 @@
...
+
+### Costly API
+
+`{{link-to}}` has [a lot of functionality](#kitchen-sink), however this functionality does come at a cost for every instance of `{{link-to}}`. This is not ideal especially if you're just using `{{link-to}}` to generate a url that can be transitioned to. By providing fine grain control of the functionality, applications should see a performance boost.
```
locks:
```suggestion
`{{link-to}}` has [a lot of functionality](#kitchen-sink).
This functionality come at a cost for every instance of `{{link-to}}`.
This is not ideal, especially if you're just using `{{link-to}}` to generate a URL that can be transitioned to.
By providing fine-grained control of the functionality, applications should see a performance boost.
```

chadhietala:
After the core team call on Friday I have updated the RFC. These changes include:

- Removal of `{{transition-to}}` modifier.
- Removal of all deprecations
- Addition of transition attribution

The reason to remove the `{{transition-to}}` modifier largely dealt with the fact that we need `HTMLAchorElements` to have the `href` populated during server-side rendering. Modifier's do not run during SSR therefore all the anchor tags in the HTML output would not work if the JavaScript failed to load. 

We decided to remove all the deprecations that this RFC originally introduced because there is no way of passing the in memory model to the route you are transitioning to. This functionality requires a bit more thought and a solution may be introduced in a subsequent RFC. 

This RFC now introduces the notion of a transition attribution. This is simply a formalization of the `Transition.data` parts of the RFC. I have amended the RFC with 2 examples showing how this functionality could be used.

alexparker:
@sandstrom Sorry wasnt really clear i'd only seen proposals for breaking out the helpers, but I was thinking more in terms of reusable parts inside the component.js, but the parts could be reused in components just fine

jneurock:
I like these changes. Thank you!

But perhaps only tangentially related, they don't address the only reason I ever extended `LinkComponent`...

Is there a story for dynamically generated route names that allow for an arbitrary list of models? I.E., a named `models` (or similar) parameter where I can pass a list?

In some of my apps I have code like this (after extending `LinkComponent`):

```javascript
{{#each linkableThings as |linkableThing|}}
  {{link-to linkableThing.title linkableThing.routeName dynamicModels=linkableThing.models}}
{{/each}}
```

The implementation code was trivial and it made me wonder "why isn't this a thing?" It also made me think a spread operator in Handlebars would be pretty awesome.

rwjblue:
@jneurock - Great point! I totally agree that this should be possible with any new helpers that we add. One possible solution would be to make the various helpers take position arguments _or_ named arguments. So for example, we could do:

```js
Using positional arguments:

<a href="{{url-for "foo" someModel}}"></a>

OR using named arguments:

<a href="{{url-for routeName=linkableThing.routeName models=linkableThing.routeName}}"></a>
```

What do you think @chadhietala?

chadhietala:
@rwjblue I think it is useful for the `models` to be a positional or named argument, but I don't think it makes sense for the route name as there are never more than one passed.

rwjblue:
Derp, yes that makes sense thank you!

chadhietala:
After our Friday meeting we decided to move this into final comment period.

text/0000-router-helpers.md
```diff
@@ -0,0 +1,571 @@
...
+_Named Params_
+- **`models`** _Object|Array|Number|String_: Same as the positional parameter. It is a compiler error if the positional params and the named param is used.
+- **`queryParms`** _Object_:  Optionally pass key value pairs that will be used to determine if the route is transitioning out.
+
+_Returns_
+- _Boolean_: Determines if the route is transitioning out.
+
+### Event Dispatcher
+
+In the past, only `HTMLAnchorElement`s that were produced by `{{link-to}}`s would produce a transition when a user clicked on them. This RFC changes to the global `EventDispatcher` to allow for any `HTMLAnchorElement` with a valid root relative `href` to cause a transition. This will allow for us to not only allows us to support use cases like the ones described in the [motivation](#anchor-tags), it makes teaching easier since people who know HTML don't need know an Ember specific API to participate in routing transitions.
```
acorncom: @chadhietala A question for you here. Everything else on this RFC looks awesome and I'm quite eager to see it all land. When it comes to this change though, I'd like to raise a potential snag we should make sure we address.

Specifically, in a "switching into Ember" scenario for a primarily server-rendered app, will there be enough safety valves as part of this change to allow folks to safely drop Ember in on their site without it unexpectedly hijacking all `<a>` tags? If there weren't a wildcard route setup as specified below, theoretically the Ember router would still be in a `whitelist` scenario and would only intercept click events for `<a>` tags that it **has** been setup to handle. But thought it was worth clarifying so we don't inadvertently cause ourselves problems for our `npm install your way to Ember` story down the road ...

sandstrom: The event dispatcher is only hooked up on the ember application `rootElement`, so it's fine to drop in Ember on a server-rendered page, even after this change.

https://github.com/emberjs/ember.js/blob/aa838632/packages/%40ember/-internals/views/lib/system/event_dispatcher.js#L241

chadhietala: @acorncom Yes they should continue to work in an SSR'd page since the events are rooted on the `rootElement`. The app that is migrating needs to make sure the root element is not `body`. I think this is typically what people to do in this case.

text/0000-router-helpers.md
```diff
@@ -0,0 +1,571 @@
+- Start Date: (fill me in with today's date, YYYY-MM-DD)
+- RFC PR: (leave this empty)
```
xg-wang: PR and this RFC name should be updated

locks: done, thank you!

knownasilya:
Did these helpers ever get implemented? https://rfcs.emberjs.com/id/0391-router-helpers/

NullVoxPopuli:
They did not. I think we forgot about them. Want to pr their implementation? We probably want an update to this rfc adding their exports to the same place as LinkTo?

PR #395, opened 2018-11-01
@ember-data packages

runspired:
[rendered](https://github.com/runspired/rfcs/blob/ember-data-packages-rfc/text/0000-ember-data-packages.md)

### DIFF
--- a/text/0000-ember-data-packages.md
+++ b/text/0000-ember-data-packages.md
@@ -17,45 +17,83 @@ This documents presents the proposed **public** import path changes for `ember-d
 
 ## Detailed design
 
-### @ember-data/model
+<table>
+  <thead>
+    <tr>
+      <th colspan="2">Before</th>
+      <th colspan="2">After</th>
+    </tr>
+    <tr>
+        <th>`import DS from 'ember-data';`</th>
+        <th>Direct Import</th>
+        <th>After</th>
+        <th>Notes</th>
+    </tr>
+  </thead>
+  <tbody>
+    <tr>
+      <td colspan="4">`@ember-data/model`</td>
+    </tr>
+    <tr>
+      <td>`DS.Model`</td>
+      <td>`import Model from 'ember-data/model';`</td>
+      <td>`import { Model } from '@ember-data/model';`</td>
+      <td>Not the default export to make working with attr/belongsTo/hasMany even easier</td>
+    </tr>
+    <tr>
+      <td>`DS.attr`</td>
+      <td>`import attr from 'ember-data/attr';</td>
+      <td>`import { attr } from '@ember-data/model';`</td>
+      <td></td>
+    </tr>
+    <tr>
+      <td>`DS.belongsTo`</td>
+      <td>`import { belongsTo } from 'ember-data/relationships';`</td>
+      <td>`import { belongsTo } from '@ember-data/model';`</td>
+      <td></td>
+    </tr>
+    <tr>
+      <td>`DS.hasMany`</td>
+      <td>`import { hasMany } from 'ember-data/relationships';`</td>
+      <td>`import { hasMany } from '@ember-data/model';`</td>
+      <td></td>
+    </tr>
+    <tr>
+      <td>`DS.Errors`</td>
+      <td>none</td>
+      <td>`import Errors from '@ember-data/model/errors';`</td>
+      <td>We should also deprecate directly importing this class</td>
+    </tr>
+    <tr>
+      <td>`DS.PromiseManyArray`</td>
+      <td>none</td>
+      <td>`import PromiseManyArray from '@ember-data/model/promise-many-array';`</td>
+      <td>We should also deprecate directly importing this class</td>
+    </tr>
+    <tr>
+      <td>`DS.ManyArray`</td>
+      <td>none</td>
+      <td>`import ManyArray from '@ember-data/model/many-array';`</td>
+      <td>We should also deprecate directly importing this class</td>
+    </tr>
+    <tr>
+      <td>(`@private`) `DS.InternalModel`</td>
+      <td>none</td>
+      <td>none</td>
+      <td>Access to this private API will be removed</td>
+    </tr>
+    <tr>
+      <td>(`@private`) `DS.RootState`</td>
+      <td>none</td>
+      <td>none</td>
+      <td>Access to this private API will be removed</td>
+    </tr>
+  </tbody>
+</table>
 
-  * Model: `import { Model } from '@ember-data/model';`
-    * replaces: `import { Model } from 'ember-data';`
-    * replaces: `import Model from 'ember-data/model;`
-  * attr
-    * replaces: `import { attr } from 'ember-data'`
-    * replaces: `import attr from 'ember-data/attr';`
-    * `import { attr } from '@ember-data/model';`
-  * belongsTo
-    * replaces: `import { belongsTo } from 'ember-data'`
-    * replaces: `import belongsTo from 'ember-data/belongsTo';`
-    * `import { belongsTo } from '@ember-data/model';`
-  * hasMany
-    * replaces: `import { hasMany } from 'ember-data'`
-    * replaces: `import hasMany from 'ember-data/hasMany';`
-    * `import { hasMany } from '@ember-data/model';`
-  * Errors
-    * we will deprecate directly importing this class
-    * replaces: `import { Errors } from 'ember-data'`
-    * `import Errors from '@ember-data/model/errors';`
-  * PromiseManyArray
-    * we will deprecate directly importing this class
-    * replaces: `import { PromiseManyArray } from 'ember-data'`
-    * `import PromiseManyArray from '@ember-data/model/promise-many-array';`
-  * ManyArray
-    * we will deprecate directly importing this class
-    * replaces: `import { ManyArray } from 'ember-data'`
-    * `import ManyArray from '@ember-data/model/many-array';`
-  * `@private` InternalModel
-    * we will deprecate directly importing this class
-    * replaces: `import { ManyArray } from 'ember-data'`
-    * `import ManyArray from '@ember-data/model/-private/internal-model';`
-  * `@private` RootState
-    * we will deprecate directly importing this class
-    * replaces: `import { RootState } from 'ember-data'`
-    * `import RootState from '@ember-data/model/-private/root-state';`
+### Unresolved Questions
 
-**Unresolved Questions**
+### `@ember-data/model`
   
   1) `InternalModel` and `RootState` are tightly coupled to the store and to our provided `Model`
     implementation. Overtime we need to uncouple this, but given their coupling to `Model` and our
@@ -65,7 +103,7 @@ This documents presents the proposed **public** import path changes for `ember-d
   2) Do the following belong in `@ember-data/model`, or in `@ember-data/relationship-layer`?
     if not with relationships, does the package name "relationship-layer" become confusing?
     
-    The argument for them being here is they are far more related to the current `Model` ui
+   The argument for them being here is they are far more related to the current `Model` ui
     presentation layer than they are to the state book-keeping which could be used with
     alternative implementations.
 
--- a/text/0000-ember-data-packages.md
+++ b/text/0000-ember-data-packages.md
@@ -6,14 +6,84 @@
 
 ## Summary
 
-This documents presents the proposed **public** import path changes for `ember-data`.
- How *exactly* we re-align any internals is informed but not determined by this.
+This documents presents the proposed **public** import path changes for `ember-data`, and moving `ember-data`
+  into the `@ember-data` namespace.
 
 ## Motivation
 
- * Unify import location `import { Model } from 'ember-data'` or `import Model from 'ember-data/model'` ?
- * Drop the concept of a single `DS` namespace
- * Unlock the potential to enable end users to drop unneeded portions of ember-data
+**Reduce Confusion & Bike Shedding**
+
+Users of `ember-data` have often noted their confusion by the existence of both direct and "god object" (`DS.`) style
+ imports for modules from `ember-data`. The documentation currently uses primarily the `DS.` use style, and users have
+ expressed interest and confusion over why the documentation has not been updated to reflect direct imports.
+
+**Improve The TypeScript Experience**
+
+Presence of multiple import locations confuses `Typescript`'s autocomplete, symbol resolution, and type hinting.
+ 
+**Simplify The Mental Model**
+
+Users of `ember-data` complain about the large API surface area; however, a large portion of this surface area is
+ non-essential user-land APIs that the provided adapter and serializer implementations expose. This move to packages
+ helps us simplify the mental model in three ways.
+ 
+ First: it gives us a natural way of dividing the documentation and learning story such that key concepts
+   and APIs are more discoverable.
+ 
+ Second: it allows us specifically to isolate the API surface area explosion of the provided adapter and serializer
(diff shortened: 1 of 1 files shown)


text/0000-ember-data-packages.md
```diff
@@ -0,0 +1,396 @@
...
+  * `PromiseManyArray`
+  * `ManyArray`
+  
+#### `@ember-data/serializers`
+
+  1) We should move automatic registration of transforms into a more traditional
+    `app/` directory re-export for the package so that when the package is dropped they
+    cleanly drop as well.
+
+#### `@ember-data/store`
```
dfreeman: Is there specific reasoning for why some of the package names are singular (e.g. `model`, `store`) and some are plural (`adapters`, `serializers`)?

chriskrycho:
Thinking along the lines @lupestro raises, we'll want a story for the TypeScript side of things for types we need to <i>name</i> but which are not <i>importable</i>. I'll have more later, as I'm excited about this.

text/0000-ember-data-packages.md
runspired: Yes. `adapters/` and `serializers/` are going to each contain ALL of the current adapter and serializer implementations, not just a single one. It's not practical for these implementations to be sectioned off by format (`json-api` `rest` `json` `base`). I would expect alternatives to these packages to be available in the future.

runspired:
> Are the classes for which there will be no imports ever returned by public methods? 

@lupestro I think you are asking " are there public classes for which there will be no public imports", to which the answer is YES.

> is there any scenario where not having an export for the classes returned by public methods prevents doing something we'd reasonably want to do?

Generally speaking,  public classes that don't have public import paths are ones which require extensive setup and lifecycle management by the store and for which it would be infeasible for end users to generate one except via the `store`.

For the most part, public store interfaces are available for accessing instances of these things, where public methods are not available these gaps should be addressed. The common case folks point out is `snapshots`, which is something we have in our sights for addressing soon.

text/0000-ember-data-packages.md
```diff
@@ -0,0 +1,396 @@
...
+
+There are a few public classes that are not exposed at all via `export` today. Those classes will not be given
+  public export paths, but the package containing their documentation and implementation is shown here:
+
+  * `@ember-data/store`
+    * `Reference`
+    * `RecordReference`
+  * `@ember-data/relationship-layer`
+    * `BelongsToReference`
+    * `HasManyReference`
```
mehulkar: are these public in the sense that instances of these classes are exposed to end users? Or that they are meant to be imported and instantiated as standalone classes? If the former, I'm not sure a public export is required. 

For the record, the only time I've cared about these classes is to call `.load()` on them, and since there network fetch path is different from using `.get('relationshipName')`, I try not to use this API anymore.

runspired: They are public in that we hand users instances of these classes, but they need no public import paths as users should not be instantiating them on their own. The documentation however ought to live alongside the appropriate mental model (and be where the code should live too)

text/0000-ember-data-packages.md
```diff
@@ -0,0 +1,396 @@
...
+    </tr>
+    <tr>
+      <td>DS.RESTSerializer</td>
+      <td>import RESTSerializer from 'ember-data/serializers/rest';</td>
+      <td>import RESTSerializer from '@ember-data/serializers/rest';</td>
+    </tr>
+    <tr>
+      <td>DS.EmbeddedRecordsMixin</td>
+      <td>import EmbeddedRecordsMixin from 'ember-data/serializers/embedded-records-mixin';</td>
+      <td>import EmbeddedRecordsMixin from 'ember-data/serializers/rest/mixins';</td>
```
pangratz: 👋 Shouldn't this be either

- ` import EmbeddedRecordsMixin from 'ember-data/serializers/rest/mixins/embedded-records';`

or

- ` import { EmbeddedRecordsMixin } from 'ember-data/serializers/rest/mixins';`

? Where the first would be more consistent with how we would import the `BuildURLMixin`.

```suggestion
      <td>import EmbeddedRecordsMixin from 'ember-data/serializers/rest/mixins/embedded-records';</td>
```

runspired: Was a mistake, should have been

```js
import EmbeddedRecordsMixin from '@ember-data/serializers/rest/mixins/embedded-record';
```

igorT:
Need to add RecordData in

igorT:
Should add a discussion of moving the package into a separate github org

text/0000-ember-data-packages.md
```diff
@@ -0,0 +1,464 @@
...
+    </tr>
+  </thead>
+  <tbody>
+    <tr>
+      <td colspan="3"><h3>@ember-data/model</h3></td>
+    </tr>
+    <tr>
+      <td>DS.Model</td>
+      <td>import Model from 'ember-data/model';</td>
+      <td>import { Model } from '@ember-data/model';</td>
```
chadhietala: I feel like this should stay as the default export. This would mirror the pattern from #176

runspired: @chadhietala I could agree there, especially as the other named packages do have a default export. The reasoning for `Model` specifically to not have one was that the common case is

```ts
import { Model, attr, belongsTo, hasMany } from '@ember-data/model';
```

but this seems equally fine and less likely to be confusing since it holds the same pattern as elsewhere.

```ts
import Model, { attr, belongsTo, hasMany } from '@ember-data/model';
```

igorT:
Action item from an in person discussion: IM implementation should go into the store package and Model should reexport it

runspired:
@igorT I don't think it's necessary to make that a detail of the RFC, only to note it for the initial implementation.

igorT:
From the ED meeting:
- Move the org movement into a separate discussion
- Add ember-data package part
- Add migration path details/user impact
- Maybe add ember-data/adapter, ember-data/serializer
- @igorT to propose serializer/adapter splits
- constraint: import paths need to match the local paths on disks:
   - complex discussion about package.json and nested paths
   - ember object has an example
- bring in learning team

tomdale:
I don't think the naming rules for Ember Data should deviate from the rules used for Ember.js in [RFC #176](https://github.com/emberjs/rfcs/blob/master/text/0176-javascript-module-api.md).

chriskrycho:
FWIW, I strongly agree with @tomdale here and I think if we have some set of limitations imposed by Node then we should work *around* Node's limitations.

However, it's worth note that it's *certainly* possible to support non-Node-friendly lookup rules (even with the `"moduleResolution": "node"` value set in `tsconfig.json`) in an application – we're doing that in our app today with my [True Myth](https://github.com/true-myth/true-myth) library, which *does* have nested module exports, not just at the root level:

```ts
import Maybe, { map } from 'true-myth/maybe';
```

☝️ is *not* a legal Node import (and accordingly there are workarounds in the library and notes in its docs for Node-only consumers), but it *is* a resolvable import in contexts which understand ES modules, and for all the reasons that [RFC #176](
https://github.com/emberjs/rfcs/blob/master/text/0176-javascript-module-api.md) hashed out, it works extremely well to support the design of module APIs that express the actual intent and design of the library.

chriskrycho:
An addendum to the above after I chatted with @runspired elsewhere for a while: one of the goals here is to make it so that the custom shenanigans I do to make this "just work" in the current Ember CLI pipeline is not necessary. And the base Node algorithm, as noted in that RFC, doesn't support it. That doesn't mean we *shouldn't* do the workarounds necessary, but it would be a lot *nicer* not to have to.

The net of that conversation was me feeling 🤔 🧐 about the right way forward here.

runspired:
cc @chriskrycho 

> one of the goals here is to make it so that the custom shenanigans I do to make this "just work" in the current Ember CLI pipeline is not necessary.

Precisely. A key constraint is that both maintainers and consumers of `ember-data` should not need to implement work arounds to have first-class typescript and editor support. There are tradeoffs, and these concerns are being raised just to make sure we properly take them into account.

-------
cc @tomdale 

> I don't think the naming rules for Ember Data should deviate from the rules used for Ember.js in RFC #176.

Putting aside the conversation surrounding editor / typescript support, (e.g. in a perfect world in which those constraints were fully solved), `ember-data` still faces unique constraints that Ember did not have to consider when implementing the `Ember Modules RFC`. Here, I believe we can escape with merely following the _spirit_ of the law, and not implementing it to the rule.

_*The Spirit*_
> Guiding Learners
> We can group framework classes and utilities by functionality, making it clear what things are related > and how they should work together. People can feel confident they are getting only what they need
> at that moment, not an entire framework that they're not sure they're benefiting from.

> For that reason, package names should assist the developer in understanding what capabilities are added by bringing in that new package. 

[(source)](https://github.com/emberjs/rfcs/blob/master/text/0176-javascript-module-api.md#guiding-learners)

Furthermore, the original RFC acknowledges that [it may not account for all the necessary constraints](https://github.com/emberjs/rfcs/blob/master/text/0176-javascript-module-api.md#module-naming--organization)

> And keep in mind, this RFC specifies a baseline module API. Nothing here precludes adding additional models in the future, as we discover missing pieces.

Most importantly, [the mental model](https://github.com/emberjs/rfcs/blob/master/text/0176-javascript-module-api.md#organize-by-mental-model) was the driving principle behind the organization proposed by the RFC

> This proposal attempts to re-align module naming with that simplified mental model, placing everything into packages based on the chunk of functionality they provide:

This organization was also driven by two rules:

* [Classes are Default Exports](https://github.com/emberjs/rfcs/blob/master/text/0176-javascript-module-api.md#classes-are-default-exports)
* [One Level Deep](https://github.com/emberjs/rfcs/blob/master/text/0176-javascript-module-api.md#one-level-deep)

If we organize by mental model, these rules conflict when put into practice in `ember-data`.

Specifically, `AdapterError` and it's subclasses are classes meant for use with Adapters. `Transform` and it's subclasses are classes meant for use with the reference `Serializer` implementations.

Using the `default export` rule and completely ignoring the `one level deep` requirement, organizing by the mental model could result in the following:

_*note: mixins are treated as classes below, with the  `-mixin` addition to the name to avoid an extra `/`_

```js
import Adapter from '@ember-data/adapter';
import BuildURL from '@ember-data/adapter/build-url-mixin';
import Adapter from '@ember-data/adapter/rest';
import Adapter from '@ember-data/adapter/json-api';
import AdapterError, { errorsHashToArray, errorsArrayToHash } from '@ember-data/adapter/error';
import InvalidError from '@ember-data/adapter/error/invalid';
import TimeoutError from '@ember-data/adapter/error/timeout';
import AbortError from '@ember-data/adapter/error/abort';
import UnauthorizedError from '@ember-data/adapter/error/unauthorized';
import ForbiddenError from '@ember-data/adapter/error/forbidden';
import NotFoundError from '@ember-data/adapter/error/not-found';
import ConflictError from '@ember-data/adapter/error/conflict';
import ServerError from '@ember-data/adapter/error/server';
//...
import Serializer from '@ember-data/serializer';
import Serializer from '@ember-data/serializer/json';
import Serializer from '@ember-data/serializer/rest';
import EmbeddedRecords from '@ember-data/serializer/rest/embedded-records-mixin';
import Serializer from '@ember-data/serializer/json-api';
import Transform from '@ember-data/serializer/transform';
import DateTransform from '@ember-data/serializer/transform/date';
import BooleanTransform from '@ember-data/serializer/transform/boolean';
import NumberTransform from '@ember-data/serializer/transform/number';
import StringTransform from '@ember-data/serializer/transform/string';
```

What this shows is that a strict interpretation of the `classes are default exports` condition is incompatible with the `one level deep` condition when organizing according to the mental model.

**However**

There's also the following considerations here that `Ember` does not really have:

* sub-classes (multiple) of primary classes
* sub-classes that will almost never be directly imported (transforms) due to being automatically registered in a manner such that they may be "resolved"

With this in mind, we could potentially make the argument that _sub-classes_ could be "named exports". In the case of `Errors` this follows the intuitive naming of these classes anyway (`InvalidError` `TimeoutError` etc.), but it does mean we end up with `Transform` and `Adapter` in the names of the reference implementations of adapters, serializers, and transforms. This is a tradeoff we may just want to accept: there is an outlier in the case of `embedded-records` though, which results in a potential move it to it's own package.

```js
import Adapter, {
  RESTAdapter,
  JSONApiAdapter,
  BuildURLMixin
} from '@ember-data/adapter';
import AdapterError, {
  errorsHashToArray,
  errorsArrayToHash,
  InvalidError,
  TimeoutError,
  AbortError,
  UnauthorizedError,
  ForbiddenError,
  NotFoundError,
  ConflictError,
  ServerError
} from '@ember-data/adapter/error';
//...
import Serializer, {
  JSONSerializer,
  RESTSerializer,
  JSONAPISerializer
} from '@ember-data/serializer';
import EmbeddedRecords from '@ember-data/embedded-records-mixin';
import Transform, {
  DateTransform,
  BooleanTransform,
  NumberTransform,
  StringTransform
} from '@ember-data/serializer/transform';
```

In some cases, this is a nice cleanup, but in others it produces awkwardness.

There is still a third option, a middle ground, in which we recognize the increased importance of Adapter and Serializer and their relative Prominence in the mental model by keeping their subclasses as default exports. Personally, I find this the most intuitive, but it must be conceded that the "mixed" nature of this approach risks confusion... but likely no more confusion than "sub classes not being default exports" already risks.

```js
import Adapter, { BuildURL } from '@ember-data/adapter';
import Adapter from '@ember-data/adapter/rest';
import Adapter from '@ember-data/adapter/json-api';
import AdapterError, {
  errorsHashToArray,
  errorsArrayToHash,
  InvalidError,
  TimeoutError,
  AbortError,
  UnauthorizedError,
  ForbiddenError,
  NotFoundError,
  ConflictError,
  ServerError
} from '@ember-data/adapter/error';
//...
import Serializer from '@ember-data/serializer';
import Serializer from '@ember-data/serializer/json';
import Serializer, { EmbeddedRecords } from '@ember-data/serializer/rest';
import Serializer from '@ember-data/serializer/json-api';
import Transform, {
  DateTransform,
  BooleanTransform,
  NumberTransform,
  StringTransform
} from '@ember-data/serializer/transform';
```

It's been quite some time since I came to the realization that regardless of approach, we're weighing inescapable tradeoffs and making compromises when it comes to the `adapter` and `serializer` package organization. The main positive is that this split into packages is the first step towards providing a world in which these packages aren't necessary at all, and whatever compromises made here might end here too.

tomdale:
I don't understand the concern about IDE/editor integration. Can someone unpack "custom build shenanigans"?

The Node module resolution algorithm absolutely supports nested modules in npm packages, in the form `@scope/pkg/module`, natively, without any additional configuration needed. You can ship packages like this as static files to npm and they will work just fine with TypeScript, Node, webpack, Rollup, or anything that uses the Node module resolution algorithm under the hood.

See https://docs.google.com/document/d/1CZC2rcpxffTDfRDs6p1cfbmKNLA6x5O-NtkJglDaBVs/preview#heading=h.k0mh3o8u5hx for a description of how Angular handles structuring their packages to accomplish this natively.

text/0000-ember-data-packages.md
```diff
@@ -76,6 +76,13 @@ With the landing of `RecordData` and the merging of the `modelFactoryFor` RFC, i
 
 ## Detailed design
 
+This RFC proposes import paths following the guidelines established in [Ember Modules RFC #176](https://github.com/emberjs/rfcs/pull/176),
+  with two addendums to account for scenarios that weren't faced by `ember`:
+* `error` sub-classes are named exports
```
tomdale: s/error/Error/, right?

runspired:
After further discussion, we've decided to accept this RFC.

PR #398, opened 2018-11-02
RouteInfo Metadata

chadhietala:
[Rendered](https://github.com/emberjs/rfcs/blob/route-info-meta-data/text/0000-RouteInfo-Metadata.md)

### DIFF
--- a/text/0000-RouteInfo-Metadata.md
+++ b/text/0000-RouteInfo-Metadata.md
@@ -98,7 +98,7 @@ This field will also be added to `RouteInfoWithAttributes` as it is just a super
 
 We feel that this a low-level primitive that will allow existing tracking addons to encapsulate. That being said the concept here is pretty simple: What gets returned from `buildRouteInfoMetadata` becomes the value of `RouteInfo.metadata` for that `Route`.
 
-The guides and tutorial should be updated to encoporate an example on how these APIs could integrate with services like Google Analytics.
+The guides and tutorial should be updated to incorporate an example on how these APIs could integrate with services like Google Analytics.
 
 ## Drawbacks
 
--- a/text/0000-RouteInfo-Metadata.md
+++ b/text/0000-RouteInfo-Metadata.md
@@ -215,7 +215,7 @@ This could work but there are two things that are problematic here:
 1. What happens to the this data if you subclass it? Do you merge or clobber the field?
 2. This is a generic property name and may conflict in existing applications
 
-### Return MetaData From `activate`
+### Return Metadata From `activate`
 
 Today `activate` does not get called when the dynamic segments of the `Route` change, making it not well fit for this use case.
 
--- a/text/0000-RouteInfo-Metadata.md
+++ b/text/0000-RouteInfo-Metadata.md
@@ -82,7 +82,7 @@ interface Route {
 
 ### `RouteInfo.metadata`
 
-The `metadata` optional field on the `RouteInfo` will be populated with the return value of `buildRouteInfoMetadata`. If there is no metadata associated with the `Route`, the `metadata` field will be `null`.
+The `metadata` optional field on `RouteInfo` will be populated with the return value of `buildRouteInfoMetadata`. If there is no metadata associated with the `Route`, the `metadata` field will be `null`.
 
 ```ts
 interface RouteInfo {
--- a/text/0000-RouteInfo-Metadata.md
+++ b/text/0000-RouteInfo-Metadata.md
@@ -55,7 +55,8 @@ export default Service.extend({
 
 While the `RouteInfo` object is sufficient in providing developers metadata about the `Route` itself, it is not sufficient in layering on application specific metadata about the `Route`. This metadata could be anything from a more domain-specific name for a `Route`, e.g. `profile_page` vs `profile.index`, all the way to providing contextual data when the `Route` was visited.
 
-This metadata could be used for more pratical things like updating the `document.title`. Currently addons like [Ember CLI Head](https://github.com/ronco/ember-cli-head) and [Ember CLI Document Title](https://github.com/kimroen/ember-cli-document-title) require to supply special metadata fields on your `Route` that will be used to update the title. This API would be a formalized place to place that metadata.
+This metadata could be used for more pratical things like updating the `document.title`.
+Currently, addons like [Ember CLI Head](https://github.com/ronco/ember-cli-head) and [Ember CLI Document Title](https://github.com/kimroen/ember-cli-document-title) require the user to supply special metadata fields on your `Route` that will be used to update the title. This API would be a formalized place to place that metadata.
 
 See the [appendix](#appendix-a) for examples.
 
--- a/text/0000-RouteInfo-Metadata.md
+++ b/text/0000-RouteInfo-Metadata.md
@@ -121,7 +121,7 @@ export default Route.extend({
 
   init() {
     this._super(...arguments);
-    setRouteMetaData(this, {
+    setRouteMetadata(this, {
       trackingKey: 'page_profile',
       profile: {
         viewing: this.userId,
--- a/text/0000-RouteInfo-Metadata.md
+++ b/text/0000-RouteInfo-Metadata.md
@@ -147,7 +147,7 @@ import Service, { inject } from '@ember/service';
     this.router.on('routeDidUpdate', (transition) => {
       let { to, from } = transition;
       let { trackingKey: fromKey } = getRouteMetadata(from);
-      let { trackingKey: toKey } = getRouteMetaData(to);
+      let { trackingKey: toKey } = getRouteMetadata(to);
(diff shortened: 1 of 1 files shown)


text/0000-RouteInfo-Metadata.md
```diff
@@ -0,0 +1,305 @@
+- Start Date: 2018-11-02
+- RFC PR: (leave this empty)
+- Ember Issue: (leave this empty)
+
+# RouteInfo MetaData
+
+## Summary
+
+The RFC introduces the ability to associate application specific metadata with it's corresponding `RouteInfo` object. This also adds a `metadata` field to the `RouteInfo` which will be the return value of `buildRouteInfoMetadata` for it's corresponding `Route`.
```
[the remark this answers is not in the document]
sohara: The way this sentences is formulated is that the subject is this RFC itself, but presumably the `RouteInfo` object is associated with a specific route instance, not this RFC. Therefore, it's not quite sufficient to correct the misuse of `it's` with `its`.  How about:
```
The RFC introduces the ability to associate application specific metadata with a route's corresponding `RouteInfo` object. This also adds a `metadata` field to the `RouteInfo` which will be the return value of `buildRouteInfoMetadata` for its corresponding `Route`.
```

text/0000-RouteInfo-Metadata.md
```diff
@@ -0,0 +1,305 @@
...
+
+While the `RouteInfo` object is sufficient in providing developers metadata about the `Route` itself, it is not sufficient in layering on application specific metadata about the `Route`. This metadata could be anything from just a more domain specific name for a `Route` e.g. `profile_page` vs `profile.index`, all the way to providing contextual data when the `Route` was visited.
+
+This metadata could be used for more pratical things like updating the `document.title`. Currently addons like [Ember CLI Head](https://github.com/ronco/ember-cli-head) and [Ember CLI Document Title](https://github.com/kimroen/ember-cli-document-title) require to supply special metadata fields on your `Route` that will be used to update the title. This API would be a formalized place to place that metadata.
+
+See the [appendix](#appendix-a) for examples.
+
+## Detailed design
+
+### `buildRouteInfoMetadata`
```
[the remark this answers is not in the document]
chadhietala: Yes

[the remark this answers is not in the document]
acorncom: Would it be possible to name that something a bit shorter? i.e. `getRouteMetadata` or something like that? I imagine you're aiming to keep from clobbering userland methods that already exist, but it'd sure be nice to not have to type that whole thing on a regular basis ... 😉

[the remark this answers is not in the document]
rwjblue: I think that it’s not going to be super super common and verbosity also helps keep it clear what’s going on while also avoiding clobbering existing methods....

Leooo:
Something we found on our side is that it's very useful to track meta data (titles, breadcrumbs titles, tracking info etc.) In the router's route declaration itself, rather than looking at each indidual routes properties: https://github.com/britishgas-engineering/ember-router-meta

chadhietala:
After our meeting on Friday we decided to move this into final comment period.

@Leooo interesting. I think the reason why I opted into putting it directly on the `Route` was that you may need to derive the return value based on values only available at runtime.

lolmaus:
Hi! Got a number of questions regarding this.

1. Since the hook is fired before the `beforeModel`, how do I put model data into metadata?
1. How do I update metadata from a controller or a service?
1. How do I use computed properties in metadata? It would be nice to have a usage pattern that makes metadata update automatically when a CP recomputes.
1. How do I put metadata into a template? The init hook approach requires *wa-a-ay* too much boilerplate for a simple `{{metadata.pageTitle}}`.
1. How do I put metadata into `<head>`? It may be out of scope and there are addons covering this (which can start using this RFC), but still it would be nice for the RFC to provide some guidelines. Or maybe it could even supersede them entirely.

Leooo:
@chadhietala something that the ember-router-meta addon allows is to define most static properties in the router, and override only when needed with some logic in the route, using the [related service](https://github.com/britishgas-engineering/ember-router-meta#editrouteroute-key-val). We use it to set page names for example - where only a few of them are dynamic. It's pretty much perfect for our use case - it would have been so much a hassle to define page titles etc. in each route, while now we have the best of both worlds and most of the things are defined in the router like for example:

```javascript
this.route('dashboard', {pageName: 'My Account', pageType: 'Account management',
          section: 'My Account', path: '/accounts', breadCrumb: 'My account', intent: 'Manage your account'}, function () {
```

Nice job @Mikek2252 btw

chadhietala:
@lolmaus I have intentionally designed the API to dissuade stateful things to occur in the hook and it is meant to be read-only once it is created. This is to align with the other properties on the `RouteInfo` object as specified in in [RFC#95](https://github.com/emberjs/rfcs/blob/master/text/0095-router-service.md#routeinfo-type). The data can be passed into component's through the `readsRouteInfo` as specified by [RFC#95](https://github.com/emberjs/rfcs/blob/master/text/0095-router-service.md#dynamically-scoped-route-info). The data would also be accessed in the template through `{{get-route-info}}` as specified by [RFC#95](https://github.com/emberjs/rfcs/blob/master/text/0095-router-service.md#dynamically-scoped-route-info). As I showed in the Appendix of the RFC you can just do `document.title = routeInfo.metadata.title`

lolmaus:
@chadhietala Hmm, so since the data can only be static, the use of this RFC is quite limited?

Can it explain what it is intended for and what it isn't?

text/0000-RouteInfo-Metadata.md
```diff
@@ -0,0 +1,308 @@
...
+  },
+  // ...
+});
+```
+
+## Motivation
+
+While the `RouteInfo` object is sufficient in providing developers metadata about the `Route` itself, it is not sufficient in layering on application specific metadata about the `Route`. This metadata could be anything from a more domain-specific name for a `Route`, e.g. `profile_page` vs `profile.index`, all the way to providing contextual data when the `Route` was visited.
+
+This metadata could be used for more pratical things like updating the `document.title`.
```
ro0gr: I think `document.title` part in the Motivation section is misleading a bit.

While a proposed stateless `metadata` design allows you to cover titles for static routes. It doesn't seem possible to support `document.title` for routes with dynamic params. 

For example, let's say I have a route with a dynamic segment `/products/:id`. In this case I would expect to have a `product.title`(loaded in the model hook) included to the `document.title`.

ef4: Any time we lean on the global `router.currentRoute` in an event handler like this, we are implicitly making timing guarantees about the relative order of events.

Maybe that is OK and unavoidable, but it's worth thinking about alternatives that make the data dependency more explicit:

```ts
this.router.on('routeDidUpdate', async (transition) => {
  let model = await transition.to.withAttributes();
  document.title = transition.to.title(model);
})
```

(in which I'm imagining a new `withAttributes(): Promise<RouteInfoWithAttributes>` method on `RouteInfo`.)

ef4: There's a tradeoff here, because the simple thing introduces asynchrony, which then also invites a whole host of other issues.

ef4: Wait, is there any reason `routeDidUpdate` can't receive a `transition` that already has `RouteInfoWithAttributes`? It seems like by the time we're doing didUpdate we have the info.

chadhietala:
@ro0gr and @lolmaus To deal with shortcomings of not having access to the model data in a natural way we decided to change the type of the object that is passed on the `Transition.to` and `Transition.from` fields. They will now be `RouteInfoWithAttributes` which have the resolved model on them. This will allow you to do something like.

```js
export default Route.extend({
  buildRouteInfoMetadata() {
    return {
      title(model) {
        return `Product: ${model.name}`;
      }
    }
  }
  // ...
});
```

```js
this.router.on('routeDidUpdate', async (transition) => {
  let model = await transition.to.attributes;
  document.title = transition.to.title(model);
})
```

chadhietala:
@ro0gr Yes you are correct RE: `transition.to.metadata.title(model);`. Yea you can totally do something like that, but I think this may be a thing people may want to lint for.

text/0000-RouteInfo-Metadata.md
```diff
@@ -0,0 +1,308 @@
...
+# RouteInfo MetaData
+
+## Summary
+
+The RFC introduces the ability to associate application specific metadata with its corresponding `RouteInfo` object. This also adds a `metadata` field to `RouteInfo`, which will be the return value of `buildRouteInfoMetadata` for its corresponding `Route`.
+
+```js
+// app/route/profile.js
+import Route from '@ember/routing/route';
+import { inject as service } from '@ember/service';
```
jacobq: typo: importing as the name `service` but using `inject` below

PR #403, opened 2018-11-26
[DATA] Identifiers 

runspired:
[Rendered](https://github.com/runspired/rfcs/blob/ember-data-identifiers/text/0403-ember-data-identifiers.md)

### DIFF
--- a/text/0403-ember-data-identifiers.md
+++ b/text/0403-ember-data-identifiers.md
@@ -1,0 +1,657 @@
+- Start Date: 2018-11-25
+- RFC PR: #403
+
+# Ember Data | Identifiers 
+
+## Summary
+
+`Identifiers` provides infrastructure for handling `identity` within `ember-data` to
+ satisfy requirements around improved caching, serializability, replication, and handling
+ of remote data.
+
+This concept would parallel a similar structure proposed for `json-api` resource identifier
+ `lid` property [drafted for version `1.2` of the `json-api` spec](https://github.com/json-api/json-api/pull/1244).
+
+In doing so we provide a framework for future RFCs and/or addons to address many common
+ feature requests.
+
+----
+
+## Motivation
+
+This groundwork RFC represents the union of a diverse set of motivations, each of which 
+ is discussed below in no particular order of importance, outside of the first. **This 
+ RFC is not seeking to immediately address each of the motivations below, we are adding
+ infrastructure to make future RFCs possible in these spaces**
+
+### Unified concept of Identity
+
+Identity is a core concept to managing a cache and guaranteeing [atomicity, consistency,
+ isolation, and durability](https://en.wikipedia.org/wiki/ACID_(computer_science)). 
+ Currently, `ember-data` has no unified mechanism for Identity. This missing mechanism 
+ introduces errors in application code, makes `ember-data` internals needlessly complex, 
+ and complicates the method signatures of many public APIs.
+
+Creating a unified `Identity` concept will allow `ember-data` to expose `identity` as a 
+first-class primitive to our users, improving the mental model, improving application 
+resiliance to errors, and providing a clear language of communication between `ember-data`'s 
+various primitives.
+
+Today, we handle `identity` in a myriad of ad-hoc ways:
+
+* `InternalModel` instances as keys for most internal methods and the relationship layer.
+* `type+id` for serializing/deserializing a `record` or [`resource`](https://jsonapi.org/format/#document-resource-object-identification)
+* `type+clientId` for caching newly created records on the client and communicating 
+   about them with `RecordData`.
+* Various other non-serializable forms of identity for tracking requests, relationship 
+   membership and state.
+* In some cases we have no concept at all where one is needed (for instance, caching 
+  queries)
+
+We wish to simplify and codify our handling of identity.
+
+### Simplify StoreWrapper and RecordData APIs
+
+Today, to deal with the lack of a unified `identifier` concept, we overload many 
+ `StoreWrapper` and `RecordData` API method signatures with `modelName`, `id`, `clientId`
+ as arguments. This leads to method signatures being long and needlessly unwieldy.
+
+Note: We had initially intended to overload all of these classes method signatures in 
+this way, to ensure that `RecordData` implementations could be `singleton`s, but we failed
+to correctly implement the `RecordData` RFC in this regard and a near future RFC will look
+at rectifying this using the `Identity` APIs introduced here.
+
+Moving to a unified concept of `identity` opens up a path to clean up these method
+ signatures.
+
+### Operations
+
+Operations are a foundational concept for `acid` transactions. Without the ability to 
+ describe an operation and a clear mental model of what operations exist and achieve, it
+ is difficult to understand how an action affects state. Granularity and clarity is key.
+
+While `ember-data` does not yet have concepts of operations or transactions, mutations and
+ updates are applied directly, there are many areas we could improve upon by introducing
+them. As with data, operations upon data should be serializable so that local state can 
+be accurately cached on local clients.
+
+### Nested Saves / API Transactions / Websocket Support
+
+Many applications wish to create or update and save multiple records together. Achieving
+ this in `ember-data` today is unweildy and has many difficult edge cases: one of which
+ is correctly matching data received back from the API to the newly created records already
+ on the client.
+
+A similar edge case occurs when a newly created record is saved for the first time and 
+ prior to receiving the request response the same record is recieved via another means
+(background polling, websocket subscription, etc.). We have no means of matching the record
+returned by the alternative means to that of the request, leading to a second cache entry 
+being created and an error once the initial request completes.
+
+One solution has been to generate and assign `id`s for records on the client, but this is
+ not always desireable. These scenarios are a major motivation for `lid` in the `json-api`
+spec. Users wishing to solve these cases would be able to serialize the `lid` of the 
+`Identifier` for a newly created record and reflect that `lid` back in any payloads send 
+from their API for the session to correctly match the payloads to the record.
+
+### Better Cache Serialization & Improved Infra for Offline Support
+
+In order to enable users to achieve full offline support, or to serialize the store 
+ or transport across the wire (for example as an advanced fastboot rehydration mode)
+the entire state of the store needs to be serializable. This RFC introduces the 
+foundation for mechanisms through which this can be later achieved.
+
+----
+
+## Detailed design
+
+```typescript
+export interface Identifier {
+  lid: string;
+}
+
+export interface RecordIdentifier extends Identifier {
+  id?: string | null;
+  type: string;
+}
+```
+
+Note: the referential stability (object reference) of all identifiers created by the
+`store` is guaranteed. E.g. any data that results in the lookup of an identifier 
+producing the same `lid` token will return the same `Identifier` instance. This is 
+useful for being able to use identifiers for either `Map` or `WeakMap` cache solutions.
+
+### Buckets
+
+In an ideal world, the `lid` of each `Identifier` would be a `v4` [uuid](https://en.wikipedia.org/wiki/Universally_unique_identifier),
+ making it practically unique in all contexts. However, due to requirements around design
+ flexibility and performance we are only requiring that `Identifiers` be unique _within 
+ their bucket_ for the data they are intended to reference.
+
+Each underlying primitive will have its own bucket, as new primitives are formalized, new
+ buckets will emerge. Initially we expect only a `record` bucket which aligns with today's
+`IdentityMap` cache for `Record`s. Examples of future buckets may include a cache for 
+`queries`, `documents`, `transactions`, `operations`, `errors`, `meta` or any number of other
+ concepts that represent state required to be serializable.
+
+In our ideal world, the `lid` would then be a `uuid-v4` that is practically unique across
+ buckets and not just within. While we  will ship a minimal `uuid-v4` generator to be used
+ for generating identifiers when needed on the client, generating large quantities of `uuid`s
+ is cost prohibitive.  An early performance analysis suggests that a few thousand identifiers
+ would bring a cost in the tens of milliseconds on powerful machines. When generating
+ identifiers for data returned by an API, this cost would impede optimizations around
+ rendering.
+
+This cost is primarily due to the need to generate large quantities of random bytes: a cost
+ that is necessarily cpu intensive. Additionally, many forms of data (such as `json-api` 
+ resources) come with unique or nearly unique identifying information already (`type` + `id`,
+`href` etc.).  Some APIs already make use of `v4` `uuid`s as IDs, and for these APIs it would
+ make the most sense to implement a custom generation method to reuise these `id`s as `lid`s
+ when present.
+  
+To balance performance with the requirements of `identity`, we are choosing what we feel is a
+ *sensible default*.  Users for whom this default does not meet their requirements may override
+the appropriate hooks to generate identifiers that do.
+
+### Customizing Identifiers
+
+For users wishing to provide increased guarantees around uniqueness and serializability
+ of identifiers we provide the ability to configure how we generate and manage identifiers.
+
+Given the guarantees around uniqueness and serializability that are required, this configuration
+ applies to all `IdentifierCache` instances and meaning that it is shared across all buckets,
+ store instances. Ideally the configuration does not change between application instances but
+ for encapsulation purposes for fastboot and tests we allow and encourage repeated setup/teardown.
+ Specifically, this recommendation is that in order to provide a strong guarantee of uniqueness
+ and serializability, _identifiers generated by separate store or application instances but which
+ represent the same data should result in the generation of the same `lid`_.
+
+Indeed in this vein, we have not provided a mechanism for distinguishing what instance of a `store`
+ has asked for an `Identifier` when multiple stores are present, but the `initializer` pattern
+ recommended below does offer the ability to distinguish _per-application_. We do not recommend
+ using this availability to affect your generation method.
+
+Supplying custom `lid` generation can be done using `setIdentifierGenerationMethod`. Currently
+ there is only one bucket (`record`) as discussed above, but we reserve the ability to add 
+ additional buckets in the future.
+
+Users should do any identifier customization within an instance-initialize prior to making use
+ of the store. Given the more universal nature of this customization, we recommend ensuring that
+ you consider the mechanics of multiple applications in fastboot or test application instance
+ scenarios when instantiating and populating any secondary lookup tables or caches for identifiers.
+ Weakmapping these lookup tables and caches to the application instance will accomplish this.
+ An example is provided below.
+
+```typescript
+/*
+  A method which can expect to receive various data as its first argument
+  and the name of a bucket as its second argument. Currently the second
+  argument will always be `record` data should conform to a `json-api`
+  `Resource` interface, but will be the normalized json data for a single
+  resource that has been given to the store.
+
+  The method must return a unique (to at-least the given bucket) string identifier
+  for the given data as a string to be used as the `lid` of an `Identifier` token.
+
+  This method will only be called by either `getOrCreateIdentifier` or 
+  `createIdentifierForNewRecord` when an identifier for the supplied data
+  is not already known via `lid` or `type + id` combo and one needs to be
+  generated or retrieved from a proprietary cache.
+
+  `data` will be the same data argument provided to `getOrCreateIdentifier`
+  and in the `createIdentifierForNewRecord` case will be an object with
+  only `type` as a key.
+*/
+type GenerationMethod = (data: Object, bucket: string) => string;
+
+/*
+ A method which can expect to receive an existing `Identifier` alongside
+ some new data to consider as a second argument. This is an opportunity
+ for secondary lookup tables and caches associated with the identifier
+ to be amended.
+
+ This method is called everytime `updateRecordIdentifier` is called and
+  with the same arguments. It provides the opportunity to update secondary
+  lookup tables for existing identifiers.
+  
+ It will always be called after an identifier created with `createIdentifierForNewRecord`
+  has been committed, or after an update to the `record` a `RecordIdentifier`
+  is assigned to has been committed. Committed here meaning that the server
+  has acknowledged the update (for instance after a call to `.save()`)
+
+ If `id` has not previously existed, it will be assigned to the `Identifier`
+  prior to this `UpdateMethod` being called; however, calls to the parent method
+  `updateRecordIdentifier` that attempt to change the `id` or calling update
+  without providing an `id` when one is missing will throw an error.
+*/
+type UpdateMethod = (identifier: StableIdentifier, newData: Object, bucket: string) => void;
+
+/*
+A method which can expect to receive an existing `Identifier` that should be eliminated
+ from any secondary lookup tables or caches that the user has populated for it.
+*/
+type ForgetMethod = (identifier: StableIdentifier) => void;
+
+/*
+ A method which can expect to be called when the parent application is destroyed.
+
+ If you have properly used a WeakMap to encapsulate the state of your customization
+ to the application instance, you may not need to implement the `resetMethod`.
+*/
+type ResetMethod = () => void;
+
+export function setIdentifierGenerationMethod(method: GenerationMethod): void {}
+
+export function setIdentifierUpdateMethod(method: UpdateMethod): void {}
+
+export function setIdentifierForgetMethod(method: ForgetMethod): void {}
+
+export function setIdentifierResetMethod(method: ResetMethod): void {}
+```
+
+A simple custom generation method might be an increasing counter like below:
+
+```typescript
+import { setIdentifierGenerationMethod } form '@ember-data/store';
+
+export function initialize(applicationInstance) {
+  // note how `count` here is now scoped to the application instance
+  // for our generation method by being inside the closure provided
+  // by the initialize function
+  let count = 0;
+
+  setIdentifierGenerationMethod((resource: Resource) => {
+    return resource.lid || `my-key-${count++}`;
+  });
+}
+
+export default {
+  name: 'configure-ember-data-identifiers',
+  initialize
+};
+```
+
+### Identifiers for Records
+
+When discussing identifiers for records it is useful to be familiar with `json-api` 
+interfaces for [ResourceObjects](https://jsonapi.org/format/#document-resource-objects) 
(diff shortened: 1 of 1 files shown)


text/0000-ember-data-identifiers.md
```diff
@@ -0,0 +1,194 @@
...
+    * New Re-Arg: `notifyPropertyChange(identifier, propertyName)`
+  * `notifyHasManyChange(modelName, id, clientId, key)`
+    * deprecated (not in original RFC) in favor of `notifyPropertyChange`
+  * `notifyBelongsToChange(modelName, id, clientId, key)`
+    * deprecated (not in original RFC) in favor of `notifyPropertyChange`
+  * `attributesDefinitionFor(modelName)`
+    * New Re-Arg: `attributesDefinitionFor(identifier)`
+  * `relationshipsDefinitionFor(modelName)`
+    * New Re-Arg: `relationshipsDefinitionFor(identifier)`
+  * `setRecordId(modelName, id, clientId)`
```
igorT: Should we make this just `notifyPropertyChange` `id`, no reason to make set id special?

text/0000-ember-data-identifiers.md
```diff
@@ -0,0 +1,194 @@
...
+
+### New APIs.
+
+* **recordIdentifierFor**
+
+`import { recordIdentifierFor } from '@ember-data/store';`
+
+```typescript
+/*
+  Returns the RecordIdentifier for the given ResourceIdentifier.
```
igorT: Do we need a split between a `recordIdentifier` and a `resourceIdentifier` if we are not guaranteeing stability? If you don't have to pass the given `recordIdentifier`, this just seems like wasted roundtrip

text/0000-ember-data-identifiers.md
```diff
@@ -0,0 +1,194 @@
...
+
+  Specifically this means that we:
+
+  - validate the `id` `type` and `lid` combo against known identifiers
+  - return an object that is guaranteed to have the `lid` for the given `type` and `id`
+  - return an object that also contains the stored `meta`
```
igorT: This is interesting, haven't thought about it much. Whats the thinking behind attaching the `meta` to it?

text/0000-ember-data-identifiers.md
```diff
@@ -0,0 +1,194 @@
...
+  - this object is referentially stable
+*/
+export function recordIdentifierFor(store: Store, resourceIdentifier: ResourceIdentifier): RecordIdentifier {}
+```
+
+### Changes to existing APIs.
+
+**Store Methods**
+* `recordDataFor(modelName, id, clientId)`
+  * New Re-Arg: `recordDataFor(identifier)`
```
igorT: Typing the identifier here would help differentiate between the `resourceIdentifier` and `recordIdentifier`

text/0000-ember-data-identifiers.md
runspired: It is special as we use this to notify that the ID has been received from a save when no ID was present before and that we must now update the cache key / index.

The data flow is *whacked* here though, as `store._setRecordId` which this delegates to is going to call `internalModel.setId` which needs to be fixed to call `recordData.<updateId>` fpr the case where someone calls `store.createRecord('foo', {}).set('id', '<someid>');`

text/0000-ember-data-identifiers.md
runspired: I think you resolved the difference for yourself below :) one is non-normalized (things we accept from the API), the other is the stricter normalized version (e.g. MUST have `lid` etc.)

text/0000-ember-data-identifiers.md
runspired: Primarily because the `json-api` spec allows it.

text/0000-ember-data-identifiers.md
runspired: I will once we have resolved which it should be, this question is in `unresolved questions` atm.

text/0000-ember-data-identifiers.md
dgeb: I would be more comfortable with `ResourceIdentifier` being exported from a JSONAPI-specific package, and not at all part of ember-data core.

runspired: @dgeb I considered that but no available lib matches what we actually expect.

Specifically:

We allow null id, and lid is not in spec yet. We also will have to handle migrating off of clientId to favor lid.

I can see a future in which this doesn’t need to be provided by ember-data though.

jessica-jordan:
@runspired Looking at the WIP tag, is it alright for us - @emberjs/the-ember-times-editors - to feature this already in next week's newsletter?

runspired:
@jessica-jordan still too many rough edges, but I think we'll be in good shape by late Wednesday/Thursday for featuring it next week.

text/0000-ember-data-identifiers.md
```diff
@@ -0,0 +1,321 @@
...
+export interface ResourceIdentifier {
+  lid?: string;
+  id: string | null;
+  type: string;
+  meta?: Meta;
+}
+
+export interface RecordIdentifier {
+    lid: string;
+    id: string | null;
```
hjdivad: Doesn't `id?` make more sense here?

You may not have an id at the time you create a `RecordIdentifier`, in particular when you create the record locally, but that doesn't make the id `null` it makes it unknown.

text/0000-ember-data-identifiers.md
```diff
@@ -0,0 +1,321 @@
...
+    type: string;
+    meta: Meta | null;
```
hjdivad: similarly i think `meta?` is more appropriate.

text/0000-ember-data-identifiers.md
```diff
@@ -0,0 +1,321 @@
...
+}
+```
```
hjdivad: `ResourceIdentifier` and `RecordIdentifier` have extremely similar names that do not make clear their semantic differences (ie that a record identifier is referentially stable and a resource identifier may not be).

What do you think about renaming `RecordIdentifier` to `StableIdentifier`.

text/0000-ember-data-identifiers.md
hjdivad: @runspired @dgeb can we not make that json-api specific package as part of this change?  presumably with the intent to eventually (or perhaps immediately) move it into the `json-api` org?

text/0000-ember-data-identifiers.md
```diff
@@ -0,0 +1,321 @@
...
+### New Typings
+
+```typescript
+import { Value as JSONValue } from 'json-typescript';
+
+export type Meta = { [k: string]: JSONValue };
+
+export interface ResourceIdentifier {
+  lid?: string;
+  id: string | null;
```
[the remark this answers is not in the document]
runspired: optional (`?`) has a very different meaning from `string | null`. Here, we are guaranteeing that the key is present, but `null` if not a `string`.

text/0000-ember-data-identifiers.md
runspired: an unknown `id` is `null`.

text/0000-ember-data-identifiers.md
runspired: `?` would imply that the key is sometimes `undefined`. To maintain shape, we always have the key present.

text/0000-ember-data-identifiers.md
runspired: Stability is not the difference between the two.

A `ResourceIdentifier` is the `json-api` interface, which may or may not have `lid` and `meta` but MUST have `type` (and must have `id` if not for a serialized version of a new client created resource with an `lid`)
A `RecordIdentifier` is our own interface for internal use, which additionally MUST have `lid` and which always has `meta` (although it may be `null`).

runspired: cc @dgeb for thoughts, as we are paralleling a similar conceptual parity from orbit here.

text/0000-ember-data-identifiers.md
```diff
@@ -0,0 +1,321 @@
...
+
+### Singleton RecordData
+
+The original proposal for `RecordData` designed the APIs in such a manner that `RecordData` could be a `singleton`. Unfortunately, the final
+ implementation did not match the RFC proposal and limits implementors to `RecordData` that are `1:1` with a `record`. This diminishes the value
+ of `RecordData` being a `cache` manager, and makes integration with alternative cache strategies (such as `orbitjs`) more difficult.
+
+In order to rectify this, we would need to add either `identifier` or `modelName, id, clientId` to the front of every method signature provided by `RecordData` as well as many methods provided by `StoreWrapper`. Given our interest in reducing the complexity of method signatures and unifying concepts of `identity`, we feel the path of least churn is to use this opportunity to correct the `RecordData` RFC to reduce the overall level of churn.
+
+### Operations
```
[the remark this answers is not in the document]
runspired: It is the path by which we correct the mistakes of the original implementation

text/0000-ember-data-identifiers.md
runspired: > @runspired @dgeb can we not make that json-api specific package as part of this change? presumably with the intent to eventually (or perhaps immediately) move it into the json-api org?

Not sure why I can't reply to this comment on the comment, but the TL;DR is that it is unlikely we ever perfectly match the signature of jsonapi here for `RecordIdentifier`, but we may be able to match the official interface for `ResourceIdentifier` once the spec includes the `lid` addition.

text/0000-ember-data-identifiers.md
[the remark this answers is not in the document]
hjdivad: That may be, but we can switch to record identifiers without operations, and operations without record identifiers.  It's not a big deal, but tackling both at once makes the rfc less focused IMHO  (and subsequently less clear).

text/0000-ember-data-identifiers.md
hjdivad: For the separate package I was referring specifically to `ResourceIdentifier`.  I agree `RecordIdentifier` should be internal.

---

> Stability is not the difference between the two.

> Additionally, our identifiers used internally or given from ember-data to application code will be referentially stable, allowing them to be used as keys in either WeakMap or Map based caches.

Are `RecordIdentifier`s referentially stable?

text/0000-ember-data-identifiers.md
```diff
@@ -0,0 +1,250 @@
...
+let identifier2 = recordIdentifierFor({ type: 'foo' }); // => { lid: 'some-random-unique-key-123b' }
+let identifier3 = recordIdentifierFor({ type: 'bar' }); // => { lid: 'some-random-unique-key-123c' }
+```
+
+Supplying custom `lid` generation can be done using `setIdentifierGenerationMethod`. Currently there is
+ only one bucket (`resource`), but we reserve the ability to add additional buckets in the future. Buckets
+ will be discussed further below.
+
+```typescript
+type GenerationMethod = (data: Object, bucket?: string) => Identifier;
```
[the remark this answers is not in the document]
runspired: fixed

text/0000-ember-data-identifiers.md
```diff
@@ -0,0 +1,396 @@
...
+    return lid;
+  }
+
+  // handle everything else
+  lid =  `my-key-${count++}`;
+  typeid_cache[cacheKey] = lid;
+
+  return lid;
+});
+ ```
```
hjdivad: It would help to augment this example with store API to make it clear what the flow is.

eg if I'm doing

```js
setIdentifierGenerationMethod(exampleInRFC);

store.findRecord('user', 'jackson5').
  then(doSomething); // this makes a request

// later

store.findRecord('user', 'abc123').
  then(doSomethingWithUser); // this does not make a request
```

At what times does the store call the generation method and with what arguments?

text/0000-ember-data-identifiers.md
```diff
@@ -0,0 +1,396 @@
...
+Note: the referential stability (object reference) of all identifiers created by the `store` is guaranteed.
+ E.g. any data that results in the lookup of an identifier producing the same `lid` token will return the
+ same `Identifier` instance. This is useful for being able to use identifiers for either `Map` or `WeakMap`
+ cache solutions.
+
+### Buckets
+
+In an ideal world, the `lid` of each `Identifier` would be a `v4` [uuid](https://en.wikipedia.org/wiki/Universally_unique_identifier),
+ making it unique in all contexts. However, due to requirements around design flexibility and performance we are only requiring that 
+ `Identifiers` be unique _within their bucket_ for the data they are intended to reference.
```
hjdivad: What about this requirement + multiple stores?  It's something that's been thought about and maybe deserves a few words of clarity in the RFC.

text/0000-ember-data-identifiers.md
```diff
@@ -0,0 +1,396 @@
...
+  lid?: string;
+  id: string;
+  type: string;
+}
+
+export interface Identifier {
+  [IS_IDENTIFIER]: true;
+  [DEBUG_IDENTIFIER_BUCKET]: string;
+  lid: string;
+}
```
hjdivad: What came of the discussion around using only `lid: string` and storing additional information in a `WeakMap`?

text/0000-ember-data-identifiers.md
```diff
@@ -0,0 +1,396 @@
...
+ cache solutions.
+
+### Buckets
+
+In an ideal world, the `lid` of each `Identifier` would be a `v4` [uuid](https://en.wikipedia.org/wiki/Universally_unique_identifier),
+ making it unique in all contexts. However, due to requirements around design flexibility and performance we are only requiring that 
+ `Identifiers` be unique _within their bucket_ for the data they are intended to reference.
+
+Each underlying primitive will have it's own bucket, as new primitives are formalized, new buckets will emerge. Initially we expect
+ only a `resource` bucket which aligns with today's `IdentityMap` cache for `Record`s. Examples of future buckets may include a
```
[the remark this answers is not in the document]
runspired: `resource` is the `string` value that the bucket symbol would be set to in this case.

text/0000-ember-data-identifiers.md
```diff
@@ -0,0 +1,396 @@
...
+
+### Refreshing an `Identifier` When recycling an `id`
+
+Occasionally some APIs re-use the same `id` for different `data`. Common scenarios for this include reusing the
+  `id` of a previously deleted record for a new record, and less commonly a stable `id` to reference the "currently logged in user".
+
+When this occurs, the existing `lid` needs the chance to be forgotten and a new `lid` generated. This method eliminates
+  the identifier from our internal cache only. Caches associated with custom identifiier generation methods must be cleared
+  by the implementors of those custom methods. Any data associated with the original `lid` should be purged from caches prior
+  to calling this method. We leave it to follow up RFCs to provide infrastructure for safely and correctly eliminating records from caches.
```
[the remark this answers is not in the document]
runspired: We still must handle connecting data coming from the API which may or may not have an `lid` to the appropriate identifier, if the API gives us the same `type+id` combo that we have seen previously, but for a different piece of data we would introduce mistakes anywhere we re-used the same `lid`. 

Perhaps in some situations this is allowable or even preferable, and users implementing custom generation methods can determine this for themselves: however we need to give users a chance to purge their existing caches for the stale `lid`.

text/0000-ember-data-identifiers.md
```diff
@@ -0,0 +1,396 @@
...
+  /*
+   Returns a new Identifier for the supplied data. Call this method to generate
+   an identifier when  a new resource is being created local to the client and
+   potentially does not have an `id`.
+  */
+  createIdentifierForNewRecord(type: string): Identifier {}
+
+  updateRecordIdentifier(identifier: Identifier, data: Resource): void;
+
+  forgetRecordIdentifier(identifier: Identifier): void
```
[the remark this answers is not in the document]
runspired: `forget` differs significantly from `unload` in semantics and a new verb was chosen here in part to convey intent and partly to avoid the baggage that comes with `unload`.

`forget` is intended to fully purge a given identifier from the identifier cache. It differs in semantics from `unload` primarily in how we consider the existence (or lack thereof) of remote data.

What this means is that we treat the identifier as if we no longer know anything about it locally or remotely. E.g. it may or may not exist remotely still, but we remove all reference to it.

This differs from the semantics of `unload` in which if a record has not first had a deletion persisted we move to a state in which we assume no local data but that remote data does still exist.

text/0000-ember-data-identifiers.md
```diff
@@ -0,0 +1,396 @@
...
+
+Called when an identifier has been generated for resource data prior to `id` being available for that resource
+ and complete resource data is now available. `ember-data` will automatically call this with the resolved payload
+ after save for any newly created records. An `identifier` can only be updated once, and only when transitioning
+ the associated resource from a never-before-persisted to persisted state.
+
+```ts
+Store {
+  updateRecordIdentifier(identifier: Identifier, data: Resource): void;
+}
```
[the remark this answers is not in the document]
runspired: It provides the opportunity to update the primary and secondary lookup tables for the identifier.

Additionally, in the case of a `RecordIdentifier` that was created locally, it provides the ability to do a "one time only" upgrade of the identifier to assign an `id`.

text/0000-ember-data-identifiers.md
```diff
@@ -0,0 +1,396 @@
...
+A simple custom generation method might be an increasing counter like below:
+
+```typescript
+import { setIdentifierGenerationMethod } form '@ember-data/store';
+let count = 0;
+
+setIdentifierGenerationMethod((resource: Resource) => {
+  return resource.lid || `my-key-${count++}`;
+});
+```
```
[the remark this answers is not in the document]
runspired: Great suggestion 👍

text/0000-ember-data-identifiers.md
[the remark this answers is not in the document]
runspired: (and it should have been the string `record`, fixed)

text/0000-ember-data-identifiers.md
runspired: Users should do any identifier generation configuration within an instance initializer prior to making use of the store.

For `RecordIdentifiers` the method is called once a record (remote or local) becomes known about. In the `findRecord` case this will be greedy (prior to network request being sent), so as long as the first request for `jackson5` in the example had populated a secondary lookup table for `abc123` to point at the same identifier prior to the second request being fired no request would be made.

In cases such as `findRecord` or async relationships where `type+id` is known but full resources are not available, `updateRecordIdentifier` will be called once the resource is loaded to ensure that we populate these secondary lookup tables.

I'll add clarifying text around this.

text/0000-ember-data-identifiers.md
runspired: Great suggestion, will add 👍

text/0000-ember-data-identifiers.md
runspired: As it was discussed further with other folks (including @stefanpenner) the realization was had that we do not need referential stability for WeakMap purposes (both because `lid` is stable and because ember-data has need of controlling most of the object lifecycle).

However, continuing to use an object wrapper, especially a stable object wrapper, comes with a multitude of benefits, including:

- **debugging:** enhanced debugging by associating additional information with the identifier in dev builds
- **debugging:** enhanced debugging by users being able to see `type` and `id` at a glance when
- **bug prevention:** enforcement of the use of the identifier generation process (to ensure lookup tables are properly populated)
- **debugging:** ability to tell at a glance that an identifier was properly processed
- **ergonomics:** closer alignment to jsonapi that makes it true that all `RecordIdentifiers` are also `ResourceIdentifierObjects`

text/0000-ember-data-identifiers.md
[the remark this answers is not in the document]
hjdivad: > Additionally, in the case of a RecordIdentifier that was created locally, it provides the ability to do a "one time only" upgrade of the identifier to assign an id.

Are the semantics then that it mutates the `identifier` parameter?

text/0000-ember-data-identifiers.md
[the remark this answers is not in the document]
hjdivad: ok.  I'm not a fan of the name.  I would recommend `removeRecordIdentifier` mainly b/c `remove` is more standard terminology for caches than `forget`.

I don't plan on being a 1-man bikeshedder though if others feel strongly about `forget`.

text/0000-ember-data-identifiers.md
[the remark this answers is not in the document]
hjdivad: fair enough

text/0000-ember-data-identifiers.md
hjdivad: > In the findRecord case this will be greedy (prior to network request being sent), so as long as the first request for jackson5 in the example had populated a secondary lookup table for abc123 to point at the same identifier prior to the second request being fired no request would be made.

Indeed.  What causes that secondary cache population to happen?  Is your intent in the example above for that to happen out of band from a userland adapter?  If so, adding that detail will go a long way to making it clear how this example solves the issue of secondary caches.

text/0000-ember-data-identifiers.md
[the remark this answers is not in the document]
hjdivad: Ok so per OOB discussion we mutate the argument here.  That makes sense; I'd recommend making that explicit here.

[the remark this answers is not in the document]
runspired: Yes. This is now made clear in the RFC

text/0000-ember-data-identifiers.md
runspired: Added more details around secondary caches and expanded examples.

text/0000-ember-data-identifiers.md
runspired: Added

text/0000-ember-data-identifiers.md
runspired: I added this reasoning into a discussion in `Alternatives`

text/0000-ember-data-identifiers.md
```diff
@@ -0,0 +1,651 @@
...
+// provided for additional debuggability
+const DEBUG_CLIENT_ORIGINATED = Symbol('record-originated-on-client');
+const DEBUG_IDENTIFIER_BUCKET = Symbol('identifier-bucket');
+
+// json-api payload version
+export interface ResourceIdentifierObject {
+  lid?: string;
+  id: string;
+  type: string;
+}
```
dgeb: Is there a reason to include `ResourceIdentifierObject` here? This should be an adapter/serailizer concern that seems to muddy the clarity of this proposal, which deals with records and their identity.

text/0000-ember-data-identifiers.md
```diff
@@ -0,0 +1,658 @@
...
+
+## Detailed design
+
+```typescript
+export interface Identifier {
+  lid: string;
+}
+
+export interface RecordIdentifier extends Identifier {
+  id: string | null;
```
dgeb: I believe this is more properly `id?: string` since `id` may be completely missing (which seems preferable to `null`).

text/0000-ember-data-identifiers.md
```diff
@@ -0,0 +1,658 @@
...
+const DEBUG_CLIENT_ORIGINATED = Symbol('record-originated-on-client');
+const DEBUG_IDENTIFIER_BUCKET = Symbol('identifier-bucket');
+
+export interface StableIdentifier {
+  lid: string;
+  [IS_IDENTIFIER]: true;
+  [DEBUG_IDENTIFIER_BUCKET]: string;
+}
+
+export interface RecordIdentifier extends StableIdentifier {
```
dgeb: Shouldn't this be `StableRecordIdentifier` ?

courajs:
I know I'm pretty late to the party here, but are we sure we don't want to take the opportunity to generalize away from string-only ids? Sometimes it is useful to use compound identifiers. I'm actually using them in an (experimental, without ED) app right now - `{site_id: '<a uuid>', index: <an integer>}`.

Although these can be serialized to/from strings, it places restrictions on the individual sub-keys (no colons, or must remain a fixed length, etc).

Maybe the cost is too great (need to define an equality protocol, makes maps/caches/lookups harder & more expensive, etc), but I'd love to hear any thoughts on the subject.

text/0000-ember-data-identifiers.md
runspired: Thanks for the catch! Over eager refactor 😂

runspired:
@courajs 

> I know I'm pretty late to the party here, but are we sure we don't want to take the opportunity to generalize away from string-only ids? Sometimes it is useful to use compound identifiers. I'm actually using them in an (experimental, without ED) app right now - {site_id: '<a uuid>', index: <an integer>}.

Compound identifiers still need to be serializable and (more importantly) would not provide a path for resolving a large number of the edge cases this RFC tackles (mainly around polymorphism, alternate indexes, new record creation, and race conditions during save).

Moreover (and not to nitpick, but I feel this is important) in the example compound key you gave (`{site_id: '<a uuid>', index: <an integer>}`), only one of the two properties is actually required. If `site_id` is truly a `uuid`, then it is unique across types and not merely within the space that `index` represents. Obviously, the more common case here is that `site_id` is not a `uuid`.

> Although these can be serialized to/from strings, it places restrictions on the individual sub-keys (no colons, or must remain a fixed length, etc).

I feel just the opposite is true here.

First: we don't attach an arbitrary length to the `lid` requirement, nor does it propose any meaning around delimiters within the string (such as colons). All we ask is that given some data, you tell us how to identify. The specifics around what guarantees you need to provide on key-uniqueness are up to you.

Second: While you _could_ attempt to parse your compound properties out of the `lid`, you very much _should not_, and we intend to obfuscate the `lid` we generate a bit to help folks avoid trying to parse `type` and `id` information out of our default strategy (vs just using `type` and `id` which are also available for instance).

> Maybe the cost is too great (need to define an equality protocol, makes maps/caches/lookups harder & more expensive, etc), but I'd love to hear any thoughts on the subject.

It would most definitely make a lot of costs greater, lookups harder, equality checks difficult. More importantly it would make serializing the state of things unnecessarily difficult.

Consider the case where the compound key used to generate the `lid` is not 1 or 2 keys but N keys. Ultimately for us to ask you if your object is arbitrarily equal to another object that may not have all N keys would be impossible. However, asking you if an object with a serialized form of identity is equal to an object with all N keys but where the serialized form is not present is very achievable.

text/0000-ember-data-identifiers.md
runspired: `null` is correct as that is what new records will have; however, a missing id when `lid` is available is allowable, so updated to allow not having `id` in this case.

sandstrom:
Very excited about this! As someone who has followed the `lid` discussion in JSON-API for a while, I think this would be a great improvement to Ember Data! 🏅

runspired:
At our meeting yesterday we voted to merge this RFC. Having squashed the commits and renamed the file with the RFC # and updated the appropriate links, I am now merging :)

PR #416, opened 2018-12-14
Glimmer Components

pzuraq:
[rendered](https://github.com/pzuraq/emberjs-rfcs/blob/glimmer-components/text/0000-glimmer-components.md)

Huge thanks to @tomdale, @rwjblue, @krisselden, @mike-north, @stefanpenner, @chadhietala, @MelSumner, and everyone else who helped with feedback and design in this RFC! This has been a long time coming, and represents a massive community effort that I have been very glad to be a part of!

### DIFF
--- a/text/0000-glimmer-components.md
+++ b/text/0000-glimmer-components.md
@@ -436,11 +436,15 @@ interface GlimmerComponent<T = object> {
   isDestroying: boolean;
   isDestroyed: boolean;
 
-  constructor(owner: Owner, args: T): void;
-
-  didUpdateArgs(): void;
+  constructor(owner: Opaque, args: T): void;
   willDestroy(): void;
 }
+```
+
+This class will be importable from `@glimmer/component`;
+
+```js
+import Component from '@glimmer/component';
 ```
 
 In addition, two new element modifiers will fill in the gaps left by the
@@ -518,9 +522,8 @@ Glimmer components have 3 properties: `args`, `isDestroying`, and `isDestroyed`.
 
 As discussed in the motivation section, `args` is an object with the values of
 the named arguments passed to the component. This property will be updated
-whenever the arguments change, with the `didUpdateArgs` hook being called after
-any updates. It will be shallow-frozen in development mode to prevent users
-from setting values on it.
+whenever the arguments change. It will be shallow-frozen in development mode to
+prevent users from setting values on it.
 
 #### `isDestroying`
 
@@ -582,16 +585,18 @@ cycle in a number of ways:
   `didUpdateAttrs` do _not_ have opposing bookends is inconsistent with this
   pattern.
 
+* Hooks that are used to update derived state, such as `didUpdate` and
+  `didUpdateAttrs`, can be generally be replaced with tracked or computed
+  properties that pull the required values as they are used, rather than eagerly
+  as they are updated. This is more inline with Glimmer's pull-based change
+  tracking system, and encourages better practices that are easier to optimize.
+
 * Hooks which are used to manipulate elements or the DOM in general can be
   removed in favor of element modifiers, which are discussed in detail in the
   next section.
 
-We can also encourage users to remove code which can be placed in tracked
-properties from hooks in general, lessening dependence on them, and gaining
-benefits from moving toward a pull-based model where computation happens as a
-side-effect of a property's use. Based on these considerations, we can reduce
-these hooks to one per major phase of the component's lifecycle: the
-`constructor`, `didUpdateArgs`, and `willDestroy`.
+Based on these considerations, we can reduce these hooks to just a setup and
+teardown method: `constructor` and `willDestroy`.
 
 #### `constructor`
 
@@ -626,24 +631,6 @@ class Person {
 Class fields are assigned _after_ the call to `super` in the constructor, but
 _before_ any of the user's code runs, allowing their values to be accessed by
 users as well.
-
-#### `didUpdateArgs`
-
-This hook runs after the component's `args` have been updated, allowing users to
-run any code they need in order to react to changes external to the component.
-This is particularly useful for running asynchronous code based on the value of
-an argument, such as `fetch` requests or ember-concurrency style tasks. It has
-the following timing semantics:
-
-* **Always**
-  * called when `args` has changed
-  * called _before_ any child component `didUpdateArgs` hooks
-  * called _before_ any `{{did-render}}` modifiers in the component's template
-* **May or May Not**
-  * be called if `args` contents are stable, i.e. the `args` object itself has
-    changed, but all the values on `args` are the same.
-  * be called in a stable order relative to sibling component `didUpdateArgs`
-    hooks
 
 #### `willDestroy`
 
@@ -789,31 +776,103 @@ using `didRender` or `didUpdate`. The only instance we found of this was in
 where it was used detect when an
 [overlay](https://ember-google-maps.sandydoo.me/docs/overlays) component has
 rendered and needs to be repositioned. For this rare case, we believe a
-[ResizeObserver](https://wicg.github.io/ResizeObserver/) may be more
-appropriate. For the more general case where a user needs to detect any and all
-rerenders in a subtree, we believe a component defined with a custom component
-manager, which still retains this ability, is an appropriate escape hatch.
+[MutationObserver](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver)
+set to detect mutations to the DOM subtreemay be more appropriate.
+Alternatively, a component can be defined with a custom component manager, which
+still retains this ability.
 
 The usages from the audit and their equivalent solution in Glimmer components
 have been included in this RFC in an appendix.
+
+### Actions
+
+In classic components, actions are defined on the `actions` hash, and can be
+referenced in templates using strings passed to the `{{action}}` helper:
+
+```js
+export default Component.extend({
+  actions: {
+    buttonPressed() {
+      // ...
+    }
+  }
+})
+```
+```hbs
+<button onclick={{action 'buttonPressed'}}>Press Me!</button>
+```
+
+This form of action sending is based on the `ActionHandler` mixin and requires
+that the component class have a `send` method. Glimmer components will _not_
+implement this API, and as such will not support string based action helpers.
+In development mode a special error will be thrown instead, informing users of
+alternatives.
+
+Instead, users should use helpers or decorators to bind functions to the
+component instance. The `action` helper and modifier do this in templates, as
+does the `bind` helper provided by the [ember-bind-helper](https://github.com/Serabe/ember-bind-helper) addon:
+
+```js
+export default class ButtonComponent extends Component {
+  buttonPressed() {
+    // ...
+  }
+}
+```
+```hbs
+<button onclick={{action this.buttonPressed}}>Press Me!</button>
+```
+
+Alternatively, a decorator could be used to bind the helper to the instance.
+This helper could bind _lazily_ as it is consumed, rather than eagerly when the
+instance is created, to prevent performance overhead:
+
+```js
+export default class ButtonComponent extends Component {
+  @bind
+  buttonPressed() {
+    // ...
+  }
+}
+```
+```hbs
+<button onclick={{this.buttonPressed}}>Press Me!</button>
+```
+
+However, one method for binding methods which should be discouraged is assigning
+an arrow function to class fields:
+
+```js
+export default class ButtonComponent extends Component {
+  buttonPressed = () => {
+    // ...
+  }
+}
+```
+
+This is messy for a few reasons:
+
+* The method is no longer available on the prototype, making it difficult to
+  mock
+* It breaks `super` and inheritance, meaning subclasses have no way to override
+  the arrow function
+* Values such as `arguments` will not be set since it is an arrow function
+
+For more details, see [this document explaining the rationale for decorators](https://github.com/tc39/proposal-decorators/blob/master/bound-decorator-rationale.md)
+over class fields for binding.
 
 ## Dependencies
 
 In it's current form this RFC is dependent on 2 of 3 open RFCs being accepted:
 
-* EITHER the [Decorators RFC](https://github.com/emberjs/rfcs/pull/408) or
-  [Tracked Properties RFC](https://github.com/emberjs/rfcs/pull/410) must be
+* The [Decorators RFC](https://github.com/emberjs/rfcs/pull/408) must be
   accepted, because Glimmer components cannot be defined using classic class
-  syntax. We only need one - it is fine to say that you can only use computed
-  properties or tracked properties for now, while work out the kinks in the
-  other.
+  syntax. If it is not accepted this RFC will have to be amended to add a way
+  for users to define Glimmer components with classic classes.
 
-  If neither is accepted, this RFC will have to be amended to add a way for
-  users to define Glimmer components with classic classes.
-
-* The [Render Element Modifiers RFC](https://github.com/emberjs/rfcs/pull/415) must be accepted, since Glimmer
-  components currently do not have any render lifecycle hooks or ways to
-  interact with the DOM.
+* The [Render Element Modifiers RFC](https://github.com/emberjs/rfcs/pull/415)
+  must be accepted, since Glimmer components currently do not have any render
+  lifecycle hooks or ways to interact with the DOM.
 
   If it is not accepted, this RFC will have to explore some of the alternatives
   listed below (`{{capture-element}}`, `bounds`, and render hooks).
@@ -940,9 +999,6 @@ bullet points here are:
   argument changes for any values that can be computed directly via getters.
   Ideally, most logic for derived state is conventionally in tracked or computed
   properties.
-* `didUpdateArgs` is the place to react to argument changes that don't fit into
-  getters. This includes asynchronous calls, conditional changes (e.g. if an arg
-  changed, do X, else do nothing), etc.
 * `willDestroy` is the correct place for all teardown code, like in classic
   components.
 * `didReceiveAttrs`, `willRender`, and `didRender` code should be extracted into
@@ -987,6 +1043,19 @@ concept in Ember, and as such will likely be unfamiliar to users and require
 more learning than normal to get used to. This also means that users will not
 be able to rely on well established patterns, and will have to develop new ones
 for dealing with element manipulation.
+
+### Lack of positional parameter support
+
+Glimmer components are meant to cover _most_ common use cases, but are also
+meant to be as minimal as possible. As such, they do _not_ have support for
+positional parameters. Positional parameters are already unusable with any
+component invoked using angle bracket syntax, but Glimmer components will also
+not support them even when using curly bracket invocation.
+
+The use cases for positional parameters are very uncommon, so it doesn't make
+sense to add them to the main component class as an option. Instead, we should
+make alternative component classes which support positional parameters, perhaps
+exclusively (e.g. asserting if positional parameters are _not_ defined).
 
 ## Alternatives
 
@@ -1149,8 +1218,8 @@ enabling typed injections.
 
 |Usage|Use Case|Converts To|
 |-----|--------|-----------|
-|[link][ep-did-update-attrs-1]|Setup component state based on incoming arguments|`didUpdateArgs`|
-|[link][ep-did-update-attrs-2]|Setup component state based on incoming arguments|`didUpdateArgs`|
+|[link][ep-did-update-attrs-1]|Setup component state based on incoming arguments|Tracked properties|
+|[link][ep-did-update-attrs-2]|Setup component state based on incoming arguments|Tracked properties|
 |[link][ep-did-update-attrs-3]|Element setup/update code based on incoming arguments|`{{did-render}}` with arguments|
 |[link][ep-did-update-attrs-4]|Element setup/update code based on incoming arguments|`{{did-render}}` with arguments|
 |[link][ep-did-update-attrs-5]|Element setup/update code based on incoming arguments|`{{did-render}}` with arguments|
@@ -1165,13 +1234,13 @@ enabling typed injections.
 
 |Usage|Use Case|Converts To|
 |-----|--------|-----------|
-|[link][ep-did-receive-attrs-1]|Setup component state based on incoming arguments|`didUpdateArgs` and `constructor`|
-|[link][ep-did-receive-attrs-2]|Setup component state based on incoming arguments|`didUpdateArgs` and `constructor`|
-|[link][ep-did-receive-attrs-3]|Setup component state based on incoming arguments, validate incoming arguments|`didUpdateArgs` and `constructor`|
+|[link][ep-did-receive-attrs-1]|Setup component state based on incoming arguments|Tracked properties and `constructor`|
+|[link][ep-did-receive-attrs-2]|Setup component state based on incoming arguments|Tracked properties and `constructor`|
+|[link][ep-did-receive-attrs-3]|Setup component state based on incoming arguments, validate incoming arguments|Tracked properties and `constructor`|
 |[link][ep-did-receive-attrs-4]|Animate based on incoming arguments|`{{did-render}}` with arguments|
-|[link][ep-did-receive-attrs-5]|Trigger validations|`didUpdateArgs` and `constructor`|
-|[link][ep-did-receive-attrs-6]|Element update code and sending an action|`didUpdateArgs` and `{{did-render}}` with args|
-|[link][ep-did-receive-attrs-7]|Updating logic and element update code|`didUpdateArgs` and `{{did-render}}` with args|
+|[link][ep-did-receive-attrs-5]|Trigger validations|Tracked properties and `constructor`|
+|[link][ep-did-receive-attrs-6]|Element update code and sending an action|Tracked properties and `{{did-render}}` with args|
+|[link][ep-did-receive-attrs-7]|Updating logic and element update code|Tracked properties and `{{did-render}}` with args|
 
 [ep-did-receive-attrs-1]: https://github.com/miguelcobain/ember-paper/blob/b36f7a7b5e19c60fa71e945590178d86274bd49d/addon/components/paper-autocomplete.js#L93
 [ep-did-receive-attrs-2]: https://github.com/miguelcobain/ember-paper/blob/b36f7a7b5e19c60fa71e945590178d86274bd49d/addon/components/paper-card-actions.js#L17
@@ -1283,12 +1352,12 @@ enabling typed injections.
 
 |Usage|Use Case|Converts To|
 |-----|--------|-----------|
-|[link][ep-did-render-1]|Resize component based on changes to args|`{{did-render}}` with arguments, or [ResizeObserver](https://wicg.github.io/ResizeObserver/)|
+|[link][ep-did-render-1]|Resize component based on changes to args|`{{did-render}}` with arguments, or [MutationObserver](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver)|
 |[link][ep-did-render-2]|Animate component based on changes to arguments|`{{did-render}}` with arguments|
 |[link][ep-did-render-3]|Set `elementDidRender` boolean on instance|`{{did-render}}`|
-|[link][ep-did-render-4]|Measure component on render|`{{did-render}}` with arguments, or [ResizeObserver](https://wicg.github.io/ResizeObserver/)|
-|[link][ep-did-render-5]|Resize component based on changes to size|`{{did-render}}` with args, or [ResizeObserver](https://wicg.github.io/ResizeObserver/)|
-|[link][ep-did-render-6]|Measure element sizes based on changes to args|`{{did-render}}` with args, or schedule `afterRender` from `didUpdateArgs`|
+|[link][ep-did-render-4]|Measure component on render|`{{did-render}}` with arguments, or [MutationObserver](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver)|
+|[link][ep-did-render-5]|Resize component based on changes to size|`{{did-render}}` with args, or [MutationObserver](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver)|
+|[link][ep-did-render-6]|Measure element sizes based on changes to args|`{{did-render}}` with args|
 
 [ep-did-render-1]: https://github.com/miguelcobain/ember-paper/blob/b36f7a7b5e19c60fa71e945590178d86274bd49d/addon/components/paper-input.js#L97
 [ep-did-render-2]: https://github.com/miguelcobain/ember-paper/blob/b36f7a7b5e19c60fa71e945590178d86274bd49d/addon/components/paper-speed-dial-action-action.js#L34
@@ -1303,8 +1372,8 @@ enabling typed injections.
 
 |Usage|Use Case|Converts To|
 |-----|--------|-----------|
-|[link][egm-did-update-attrs-1]|Synchronize options with Google maps|`didUpdateArgs`|
-|[link][egm-did-update-attrs-2]|Update component based on changes to arguments|`didUpdateArgs`|
+|[link][egm-did-update-attrs-1]|Synchronize options with Google maps|Refactor to use actions to modify data, or use a modifier|
+|[link][egm-did-update-attrs-2]|Update component based on changes to arguments|Refactor to use actions to modify data, or use a modifier|
 
 [egm-did-update-attrs-1]: https://github.com/sandydoo/ember-google-maps/blob/3f846751eda7fff08a02367c04407cf545741f00/addon/components/g-map.js#L103
 [egm-did-update-attrs-2]: https://github.com/sandydoo/ember-google-maps/blob/3f846751eda7fff08a02367c04407cf545741f00/addon/components/g-map/map-component.js#L54
@@ -1343,7 +1412,7 @@ enabling typed injections.
 
 |Usage|Use Case|Converts To|
 |-----|--------|-----------|
-|[link][egm-did-render-1]|Detect changes to subtree and reposition overlay|[ResizeObserver](https://wicg.github.io/ResizeObserver/) or custom component that can trigger actions on subtree rerenders|
+|[link][egm-did-render-1]|Detect changes to subtree and reposition overlay|[MutationObserver](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver) or custom component that can trigger actions on subtree rerenders|
 
 [egm-did-render-1]: https://github.com/sandydoo/ember-google-maps/blob/master/addon/templates/components/g-map/overlay.hbs#L6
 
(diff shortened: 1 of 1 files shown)


lennyburdette:
This is fantastic. Nice write-up!

I think that mentioning event handlers is warranted. While `<button onclick={{action ...}}>` is probably sufficient in most cases, losing the event methods on the component class has at least two drawbacks.

1. The classic component event methods hooked into Ember's event delegation system and therefore are potentially more efficient. 

    `{{#each (times 100000)}} <button onclick={{action ...}}></button> {{/each}}` assigns 100k event attributes in the DOM.

    On the other hand, `{{#each (times 100000)}} <XButton @click={{action ...}} /> {{/each}}` does not assign any event attributes, instead using the "click" event handler on added to the application's root element.

   I don't know the memory and performance implications of DOM event attributes in modern browsers, but AFAIK event delegation still has performance wins.
2. The `focusIn` and `focusOut` class component event methods don't have DOM attribute equivalents! This is important because the `onfocus` and `onblur` DOM attributes don't bubble. [Here's a twiddle](https://ember-twiddle.com/33e5d8c305a2d23fd1b161ad69b3f7c3?openFiles=templates.application.hbs%2Ctemplates.components.my-component.hbs) demonstrating the issue. (It [doesn't look like](https://github.com/facebook/react/issues/6410) react has solved this either.)

I'm guessing that element modifiers are the way forward and I would be perfectly happy with that. Could you add something to the RFC clarifying this? Maybe just a section in the "How we teach this".

pzuraq:
Great points! I think we should definitely add a section mentioning the reasons why we are not including event handler hooks on the components, and we can also discuss the possibilities for solving those problems. Addressing them individually:

1. There may be some wins there, particularly in the laziness of "pulling" on the  event handler function and binding it. Either way though, I believe an event handler function would have to be made _somewhere_ per item, and redirected to at some point. I'm not sure about this at all, but I would _imagine_ that a global jquery style event handler (how handlers on classic components appear to be routed, even post-Jquery) would be somewhat more expensive than letting the browser do the handling using standard event dispatching. Really unsure about that though, would have to benchmark to find out I think 😄 

2. I definitely think this is something we would need to add using element modifiers somehow, it seems like `addEventListener` has the behavior we're after.

I do think outlining specific APIs is out of scope for this RFC, but we should definitely mention current alternatives for users who are used to classic components in "How we teach this", specifically use either `{{action}}` or `onclick=` style event handlers.

text/0000-glimmer-components.md
```diff
@@ -0,0 +1,1386 @@
...
+initial setup of the component. Service injections and args being available in
+this way also makes them available to class field initializers, which run
+immediately _after_ the call to `super`:
+
+```js
+class PersonComponent extends GlimmerComponent {
+  @service time;
+
+  // Use the values of args in an initializer
+  fullName = `${this.args.firstName} ${this.args.lastName}`;
```
scottmessinger: This is really neat. Just to clarify, this would not be tracked, right? So, if the args change, `fullName` wouldn't update, correct?

pzuraq: Yes, for two reasons:

1. The property is not marked as tracked, so it will not trigger a rerender even if it is changed (unless you use `set`)

2. The property is not a getter, so it doesn't know _how_ to recalculate even if it knows it's dependent on the values of the args. In that sense, this would just be a convenience for initial setup.

text/0000-glimmer-components.md
```diff
@@ -0,0 +1,1386 @@
...
+      title={{@post.imageTitle}}
+    />
+  {{/if}}
+
+  {{@post.text}}
+</section>
+```
+```js
+// components/post.js
+export default class PostComponent extends GlimmerComponent {
```
dfreeman: This might be outside the scope of the RFC, but in (and post-)Octane, do you expect `GlimmerComponent` to be the commonly-used identifier for this base class? Of course users can import it however they like, but I imagine the guides and blueprints will all standardize on one particular name.

In the near term I can see the value in distinguishing between Glimmer components and classic Ember ones (though in many codebases I imagine the ES class syntax will also be a differentiator), but in the long term do you see `GlimmerComponent` as the standard name for these things?

rwjblue: I’d personally think the common thing would be to `import Component from ‘special-place’` (and not have each import duplicate the base class type/name), and rely on the import path to be the signifier. In most modules I would expect the import to be pretty close to the class extend line so I don’t really see this as an issue...

dfreeman: I agree, and I think it's telling that `sparkles-component` users (at least in the cases I've checked) write `import Component` rather than `import SparklesComponent`.

pzuraq: I’ll update the examples accordingly, including imports so folks see the whole picture

rtablada: From a learning and context switching perspective it may make sense to have the blueprint (and taught) import class name be `GlimmerComponent` this makes it easier to distinguish and understand the differences for existing users.

The flip side of that is that there's some legacy baggage for new users or Greenfield projects that only use Glimmer style components.

pzuraq: One distinguishing factor will be that classic components will _probably_ still be defined using classic class syntax in the official guides. Even if the decorators RFC is merged, it does not include legacy decorators (as discussed in the RFC) and they will only be available via addons like `ember-decorators`, so the official guides will not be able to use them (or we would have to make the addons official).

mike-north:
Is there any concept of [positional params](https://guides.emberjs.com/release/components/passing-properties-to-a-component/#toc_positional-params) in the angled/glimmer component world?

```hbs
<form onSubmit={{action foo}}>
  <CustomInput @firstName> First Name </CustomInput>
  <CustomInput @lastName > Last Name  </CustomInput>
</form>

```

pzuraq:
In the [Angle Bracket Invocation RFC](https://github.com/emberjs/rfcs/blob/master/text/0311-angle-bracket-invocation.md#positional-arguments) positional parameters were decided against for angle bracket syntax. I believe the reason this came to be was the difficulty parsing positional parameters vs. attributes, for instance:

```js
<option selected></option>
```

It's difficult to tell if `selected` is meant to be a positional parameter or an attribute.

However, positional parameters are still legal in curly syntax and will continue to be, which was part of the reason for having both invocation styles. Glimmer components do not support positional parameters in the current proposal, and I believe the use cases are rare enough that a custom component manager (that does things like _enforce_ that they are defined, and asserts against users using angle bracket invocation for them) would be a good alternative. 

This should definitely be mentioned in the RFC either way, thanks for bringing it up!

rwjblue:
> Is there any concept of positional params in the angled/glimmer component world?

Just to reiterate @pzuraq's point, we need to be very careful to not infer that the _invocation_ style is related to the base class used (mostly because that _was_ true in past revisions of these features)...

michaelrkn:
What's the reason for having a new class for Glimmer components? Why not add these features incrementally to the existing components API?:

* Add the `.args` property and deprecate accessing properties otherwise.
* Let users opt into outer HTML semantics, either through `ember feature:enable template-only-glimmer-components` or maybe even on a component-by-component basis, eg: 

```
  outerHtmlSemantics: true
});
```

* Deprecate the properties and hooks that aren't part of Glimmer components.
* Users can already use ES6 classes as of Ember 3.6.

It seems like this would make it easier for existing users to move to this new approach to components, and easier for new users to not have two component classes to consider. (And Ember 4.0 can drop the deprecated the features, and what's left is Glimmer components.)

pzuraq:
@michaelrkn that's a great point! This has been considered over time but I think has been seen as undesirable and problematic for a number of reasons. We should definitely add a section addressing this in Alternatives, but to answer directly:

### It's still a single massive shift

The biggest one to me is that we have such a large shift, that no matter what we effectively end up with two "classes". As you pointed out, outer HTML semantics, which are perhaps the biggest shift for Glimmer Components, would still require users to opt-in via a feature flag or per-component flag. Being able to transition one component at a time is important, so a per-component flag would definitely be necessary.

This flag would essentially disable a whole bunch of hooks and properties (all of the event hooks, `tagName`, `classNames`, `element`, etc), which would basically make the component behave like a completely different class. Instead of actually having two classes though, we would be doing this dynamically. Two class APIs, if you will.

There is no middle state for most of these features. You can't start using outer HTML and then get rid of `tagName` and `classNames` and event hooks one PR at a time. You'll have to rewrite your component altogether as soon as you turn it on.

### Typings and performance

Typescript is starting to become more commonly used in Ember apps, and even users who don't use it are able to benefit from typings. It's a very nice DX to be able to import the component class and immediately see what properties are part of its public API. This is possible today with Ember components, but there's a lot of noise (over 50 props and methods) so this makes it less helpful. Combine this with the outer-html feature flag, where you are effectively _changing_ the public API based on a flag, and it becomes even less helpful.

Additionally, Glimmer components having a much smaller API will be immediately better for performance, as will them not inheriting from `EmberObject`. This will help both new apps and larger apps that are performance sensitive as soon as they begin to transition to Glimmer components, instead of waiting another year or two for the extra hooks to be removed and even longer for the full removal of `EmberObject` from classic components.

### Confusing conventions

Part of the issue with attempting to add these APIs incrementally is it may make it much less clear what the Ember Way is, for any given combination of APIs. For instance, what are best practices before you enable outer-html, but after you switch to using `.args`? What are best practices before you enable outer-html, but after lifecycle hooks have been deprecated? What are best practices when using classic class syntax with all these new features?

Doing this incrementally, feature by feature, necessarily increases the possible permutations of in-between states, and makes it harder to define conventions for each one of them. New users may get lost in the variety of features (and permutations of them) they need to learn, and existing users may be frustrated by the lack of clarity.

### Interoperability

One of the stated goals of Glimmer components in the RFC is to interoperate with Glimmer.js, and allow users to begin using components with both frameworks. On its own, this is definitely not a compelling enough reason to make the decision to have a new default component class - Ember should do what's best for Ember. However, with the other reasons pushing us toward a new base class anyways, it is a nice win that will enable more experimentation, and increase Ember's appeal to users who care about byte-size. Without a new base class, we definitely would not be able to do this for some time, since it would require removing all Ember specific parts of the component class (including extending from `EmberObject`).

text/0000-glimmer-components.md
```diff
@@ -0,0 +1,1386 @@
...
+as valuable data points for the viability of a simpler component API, and inform
+the design accordingly.
+
+## Detailed design
+
+Glimmer components have the following interface:
+
+```ts
+interface GlimmerComponent<T = object> {
+  args: T;
```
scottmessinger: I've always wanted to have something like React's PropTypes in Ember. More than that, I really wanted to be able to specify the attributes a component should receive with Typescript. By specifying the shape of the `args` using Typescript, could we get typed arguments in Glimmer components?

michaelrkn:
> We should definitely add a section addressing this in Alternatives, but to answer directly

Yeah - you could even just copy your comment over!

> There is no middle state for most of these features.

Can't ES6 classes, `.args`, and outer HTML be implemented independently from each other? 

> This flag would essentially disable a whole bunch of hooks and properties (all of the event hooks...

Why would all of the event hooks need to be disabled in order to use outer HTML? With `ember feature:enable template-only-glimmer-components`, can't users get outer HTML with the current event hooks?

> Glimmer components having a much smaller API will be immediately better for performance, as will them not inheriting from EmberObject.

What is the performance impact? And isn't a major version release a better place to focus on removing APIs?

text/0000-glimmer-components.md
pzuraq: So, the answer is _partially_. You definitely will be able to use TS to specify the shape of args when defining classes, and internally within the class these types will work. However, we don't currently have a way to integrate _templates_ with the Typescript compiler, so there's no way for us to validate that the types are correct when a component is used in a template. Long term, I think this is the ideal way to have argument typings.

In the meantime, runtime validations like React's PropTypes are possible, but extracting them from TS would be tricky. There's been some discussion of this on Discord, but nothing concrete as of yet. Libaries like [@ember-decorators/argument](https://github.com/ember-decorators/argument) and [ember-prop-types](https://github.com/ciena-blueplanet/ember-prop-types) can continue to fill the gap, but I definitely think we can do more here!

pzuraq:
> Can't ES6 classes, .args, and outer HTML be implemented independently from each other?

Yes, the absolutely can. What I was referring to here was the set of features that are tied together by Outer HTML, so:

* `tagName`
* `classNames`
* `classNameBindings`
* `attributeBindings`
* Event hooks such as `click`, `focus`, `hover`, etc.
* Default bound properties such as `isVisible`, `ariaRole`, etc.
* The `element` property, and any usages of it

There may be some more I'm forgetting at the moment, but the point is that as soon as you toggle that switch, you have to rewrite _all_ of the above to the new world. That's a fairly big switch, and really points to two separate class APIs.

> Why would all of the event hooks need to be disabled in order to use outer HTML?

I was specifically referring to the `click`, `focus`, `hover` etc. style event hooks here, not the lifecycle hooks like `didInsertElement` and `didRender`. These _would_ be able to continue functioning, but would likely need to be rewritten since you can't use `element` anymore.

> What is the performance impact? 

It's hard to say without actual benchmarking, and it's hard to benchmark without rewriting a real world application 😄 We do have a minimal component API in the form of `sparkles-component` now so we can start to do this to get a real world estimation.

However, the fact is that we _are_ doing much more work than is necessary right now, between the usage of `EmberObject` and the many, many lifecycle and event hooks of classic components. Components are one of the most commonly instantiated objects in an Ember app, so even the smallest wins here should move the needle.

> And isn't a major version release a better place to focus on removing APIs?

Absolutely! This is _not_ a deprecation RFC, and classic components _will_ remain part of the Ember API for the forseeable future. They still have functionality which is not available in Glimmer components (such as positional parameters), and they are very widespread in usage so it will take a long time to transition away from them. 

This was part of the reason for creating component managers in the first place, and the painstaking design that was done to ensure that all forms of component defined using them would interoperate at all levels - to ensure that transitions could be done incrementally, without removing or deprecating component APIs that are still in common use.

NullVoxPopuli:
@michaelrkn that'd be a good thing to do in a deprecation RFC - for helping out with the transition and such

text/0000-glimmer-components.md
```diff
@@ -0,0 +1,1386 @@
...
+The advantage in this is that named arguments are fully transparent. When seen
+in a template, users can know without a doubt that the named argument was a
+value passed in from the invocation. Likewise, when they see a standard binding
+to a value like `{{this.firstName}}` or `{{firstName}}`, they know this is a
+value defined on the component - it could a computed property, it could come
+from a service, it could be a random value, but it is _not_ an argument.
+
+Glimmer components align their argument access with named args by making
+arguments available exclusively on a (shallow) frozen object, `this.args`.
+Attempting to modify `this.args` will hard error, meaning that like templates,
```
voorEddie: Apologies for this question may be too small. What's 

> Attempting to modify `this.args` will hard error, ....

I guess you are trying to say something like 'will throw error' right?

text/0000-glimmer-components.md
scottmessinger: So, if understanding you correctly, this RFCs is does some the work to support typed args but it's not all the work that needs to be done to validate that a component is being passed the right arguments. In other words, this RFCs lets us specify the types a component should receive but we can't actually validate we're sending those types because the templates aren't integrated into the typescript compiler. Is that right?

rwjblue: Yes, **but** this is a nice incremental step forward. It also allows the usage of `this.args` _inside_ the component to be properly typed which is a nice win too...

text/0000-glimmer-components.md
rwjblue: Yes, exactly!

scottmessinger:
Does this RFC for Glimmer Component prevent introducing single file components in the future? React pioneered this and since then, it's been picked up by [Svelte](https://svelte.technology/guide#computed-properties) and [Vue.js](https://vuejs.org/v2/guide/single-file-components.html) (among others). Personally, I only use single file components with Ember and it's glorious:

```js
import hbs from "htmlbars-inline-precompile"

    // ....

   layout: hbs`
     <h1>Template here</h1>
   `
})

```

For me, it makes the app way less confusing to navigate as I'm no longer juggling 2 files for every one component. My feelings aside, the adoption of single file components in both React, Vue.js, and Svetle.js suggests other developers really love the simplicity this provides.

With the removal of `layout`, the way I do things now won't work and a new way of integrating the template and component definition would have to be developed. I know this RFC isn't addressing single file components, but does anything in the RFC seem like it would prevent templates and components from being in the same file in the future?

cibernox:
@scottmessinger I believe it doesn't prevent single-file components. It could just be implemented with babel transform that takes out the `template: hbs...` bit and treats it like a different file.

As I read it what this DOES prevent is something like reusing a single template file for several components, is that so? Maybe the same thing could be achieved creating a symlink?

rwjblue:
> Does this RFC for Glimmer Component prevent introducing single file components in the future?

No, this RFC definitely does not foreclose on the concept of single file components (which is definitely something that we would like to push forward in the near term). 

> 
> ```js
> import hbs from "htmlbars-inline-precompile"
> 
> export default Ember.Component.extend({
>     // ....
> 
>    layout: hbs`
>      <h1>Template here</h1>
>    `
> })
> ```

It does however make the _specific_ syntax that you highlighted not work (because there would no longer be a `layout` property). The single file component world would use a more customized syntax so that both the JS/TS and template can be properly highlighted...

rwjblue:
We reviewed this RFC at the core team meeting today, and believe that this is ready for final comment period.

courajs:
I _love_ all the detail here, especially in the teaching section.

What's the rationale on needing both `isDestroying` and `isDestroyed`? Is there a good motivating example to share?

mike-north:
Could we add a bit to "how we teach this" that addresses `@glimmer/component` templates having no backing object, where `@ember/component` and `sparkles-component` templates having one?

One snag I've already seen experienced devs hit as they adjust to the "octane mental model" is the use of `{{action (mut x) }}` with _implicit state_.

The following (template-only) `Ember.Component` component will work as you'd expect, allowing the user to toggle whether content is hidden/shown with respective buttons.

```hbs
{{#if showSomething}}
  [Here's your something]

  <button onclick={{action (mut showSomething) false}}>Hide</button>
{{else}}
  <button onclick={{action (mut showSomething) true}}>Show</button>
{{/if}}
```

This will _not_ work the same way with glimmer components, since with no backing object, `showSomething` would no longer be an implicit value that's initialized to `undefined`.

related: https://github.com/emberjs/ember.js/issues/17202

pzuraq:
@mike-north Glimmer Component's _do_ have a backing object, and patterns with `mut` and such will continue to work the same as they always have. Template-only components, which were proposed in #278, do not. I think there may be some confusion here since template-only-components are pure Glimmer_-VM_ components, without a state bucket or backing class, and I don't believe the guides have been updated to cover them since they are an optional feature that will remain optional until Ember@v4.

mike-north:
@pzuraq I'd argue that "Glimmer Components" vs "Glimmer VM Components" is an implementation detail. 

In the GC world, deleting my .js or .ts file would result in one of two things, depending on whether the [flag that's currently enabled in the blueprint, remains so](https://github.com/ember-cli/ember-octane-blueprint/blob/64c863537c38e11f1f404318024b8f392a5a82c5/files/config/optional-features.json#L4).

**A (`template-only-glimmer-components = false`)** - the template is treated as a template-only Ember.Component. Lifecycles will work differently, `{{action 'foo'}}` is fair game, inner-html mode, etc...

**B (`template-only-glimmer-components = true`)** - the template is treated as a template-only Glimmer VM component. No more state bucket or backing object.

Both of these involve the component's nature changing more than developers may be used to (compare to the classic mental model of "you can safely delete empty JS modules" for components, routes, controllers, adapters, serializers). This is something that needs special teaching attention to set devs up for success.

pzuraq:
True, and I understand your concern - this change in behavior _needs_ to be documented. However, this was discussed and decided upon in a previous RFC. If it were not for #278, we could for instance have decided to continue creating a backing component for template-only-components, and just changed it to be an instance of `GlimmerComponent`. I think the most appropriate place to discuss the impact on guides and learning for removing the backing class from template-only-components would have been #278.

That said, Octane (and more generally, a new [Edition](https://github.com/emberjs/rfcs/pull/371)) is about rolling up the various changes since the last edition and making sure they are all fully covered in the guides. This definitely includes TO-components, along with Angle Bracket invocation, named args, _and_ Glimmer Components. I think ideally we would have done a better job discussing the learning impact when TO-components were first proposed, but they will most certainly be covered by the effort to update the docs in this new edition.

richard-viney:
This RFC is looking great! Handling of component argument defaults is the one thing I'm still a little uncertain about.

My thoguhts are:

1. Using the `{{or}}` helper buries the default in an unknown place in the template, and requires it to be duplicated if the arg is used multiple times.
2. A defaulting getter or an `@arg` decorator are better, though my take is that this neither are technically an argument default, because they're no longer referenced using the `@` syntax in the template.

Has any consideration been given to some kind of `defaultArgs()` method? E.g.

```ts
interface Args {
  enabled: boolean
};

export default class extends GlimmerComponent<Args> {
  static defaultArgs(): Partial<Args> {
    return { enabled: true };
  }
};
```

Any arguments that aren't provided to the component, or have the value `undefined`, would have the default applied. The `@` syntax in the template can be used to access an argument that had a default value applied. This could be limited to primitive `number/string/boolean/null` values to avoid component instances ending up with references to the same JS object in their args.

I realise the above may cut against the intended role of `@` arguments somewhat.

pzuraq:
@richard-viney that's a great point, and while I do believe it has been discussed _somewhere_ I can't seem to find it in this PR so I think it'll be good add the reasoning behind why we didn't add `defaultArgs` here.

From a language design perspective, it comes down to being able to look at the template and know, with _confidence_, what things are. Currently, with all of the new features, we can know that:

* `{{@arg}}` refers to the value passed into the component, `<Foo @arg="value">`
* `{{this.field}}` refers to a field/property _on_ the component
* `{{item}}` refers to either a global value (helper, modifier, etc) _or_ a local variable (`{{#each items as |item}}`)

> Note: `item` still can technically also refer to the property on the component instance, but not using `this` has been deprecated.

We can know these things about these values without even looking at the rest of the template, which is pretty cool when you consider how dynamic Handlebars templates used to be! It used to be that you had no idea whether or not something was a property, an argument, a local variable, a global, etc. without reading not only the _entire_ template, but _also_ the `component.js`, since arguments _were_ properties.

With `{{@arg}}` syntax as it is currently, this is even better. We can actually skip around and read templates without having to read _any_ javascript, because we know for certain that whenever we see `{{@arg}}`, it's referring _directly_ to the value passed in on template invocation. There is no way the `component.js` could have affected the value, so we don't even need to check. Personally, I like to be able to read templates when getting to know a new application and get a general understanding of how things work, so this constraint helps a _lot_. 

The fallback is that we use an alias with a default, so something like `{{this.argWithDefault}}` let's say. DX-wise, this is no better or worse than `defaultArgs`. We still have to go to `component.js` to see what the value is, and we now _know_ that from a hint in the template. So, keeping the constraint doesn't really harm DX, it just feels a little strange.

One final note is that one frequent use case for default arg values is for subclassing different types of components and overriding default values:

```hbs
<!-- button.hbs --> 
<button class="{{this.buttonClass}}">{{yield}}</button>
```
```js
// button.js
  buttonClass: 'default',
});
```
```js
// success-button.js
export default Button.extend({
  buttonClass: 'success',
});
```

We believe this is actually better done (and probably more performant) "functionally", via templates:

```hbs
<!-- button.hbs --> 
<button class="{{or @buttonClass 'default'}}">{{yield}}</button>
```
```hbs
<!-- success-button.hbs --> 
<Button @buttonClass="success">{{yield}}</Button>
```

This can quickly become a maintenance nightmare however, as soon as you add a new arg to `Button` it needs to be passed through each type of button. We're currently discussing adding a `...arguments` syntax, which would mirror the `...attributes` syntax that currently exists, and would cover that edge case (and in general make forwarding arguments much easier).

That covers the language design part of this, but the other part is that the `{{@arg}}` syntax currently completely bypasses the component instance, and goes directly to the arguments in the template layer. Via the [Custom Components RFC](https://github.com/emberjs/rfcs/blob/master/text/0213-custom-components.md#createcomponent), the component manager also _cannot_ mutate the arguments object in the template layer, so there really isn't a way to do it, and it would be a pretty significant engineering effort to make it possible (if it could even be done a backwards-compatible way).

rwjblue:
@richard-viney - Thanks for chiming in! Unfortunately, I think you hit the nail on the head here:

> I realise the above may cut against the intended role of @ arguments somewhat.

😺 

Specifically, a named argument in the template tells you explicitly that that value is _unaffected_ by the backing JS/TS file for the template. If we made it possible to modify/mutate the value of the named argument in the JS/TS file that pretty squarely violates the original goal. The [motivation section of the Named Arguments RFC](https://emberjs.github.io/rfcs/0276-named-args.html#motivation) has more explanation and details here.

> Using the {{or}} helper buries the default in an unknown place in the template, and requires it to be duplicated if the arg is used multiple times.

In lots of cases where you have a simple default, this is a perfectly good solution when you don't also need the defaulted value in the JS/TS side. RE: the caveats: 

* Not all templates will need to default multiple times (and you could use a `{{#let` to share a single value if you do).
* I don't think putting template defaults in the template is burying it 😜. Contrarily, I think putting a defaulting mechanism in the JS/TS file when its only needed in the template is worse (barring some exceptional cases where the defaulting is needed in the JS/TS for other reasons, the logic is complicated and having it in JS/TS makes it easier to reason about and test, etc).

> A defaulting getter or an @arg decorator are better, though my take is that this neither are technically an argument default, because they're no longer referenced using the @ syntax in the template.

Yes! I think this is absolutely the right thing! Seeing `{{this.foo}}` in the template **tells** the reader of the template to "go look in the JS/TS file backing this template". That is _precisely_ the message you should be sending when you need to use the JS/TS file to gather a default value.

---

In a future world (when [Property Lookup Fallback Deprecation](https://emberjs.github.io/rfcs/0308-deprecate-property-lookup-fallback.html) and "template imports" are done) there will be three different ways of referencing values in a template:

* `{{foo}}` -- this will **only** mean a local value defined in the template itself (ala block param)
* `{{this.foo}}` -- this will **only** mean "lookup on the backing context" (aka look in the component's JS/TS file)
* `{{@foo}}` -- this will **only** mean "look at the callsite to track down what is passed in"

IMHO, this **massively** clarifies templates for future readers, and reduces a number of super common pitfalls in the current system...

rwjblue:
@pzuraq haha, our replies crossed in the mail! I think you probably said it better though 😃

richard-viney:
Thanks for the responses!

Yes I've read the Named Args RFC carefully, have been using them for some time in production, and generally agree with the rationales put forward there.

> * Not all templates will need to default multiple times (and you could use a {{#let to share a single value if you do).

I agree that for arguments which only need a default applied once in the template it's reasonable to apply the default in the template directly. Is using `{{or}}` actually going to be recommended practice though? Won't this cause surprises when the argument value passed in is falsey/blank? A new (very simple) helper may be needed instead of `{{or}}` for this template-based defaulting behaviour.

Wrapping a template in `{{#let}}` in order to apply argument defaults feels a tad awkward to me - is this being suggested as a best practice? Such defaults obviously aren't available in the component JS/TS, and a developer can still easily use the `@` version of the argument without realising they actually needed the defaulted version as available in the `{{#let}}` block.

> * I don't think putting template defaults in the template is burying it 😜.

By 'burying' what I meant was: how does a developer coming to a component know where to find the default value for an argument if it has one? i.e. they're interested in the component's interface, rather than diving into its internals. The most immediately useful thing to them is the `Args` interface (TS-specific, sorry) as it defines the component's primary public interface. Default values of non-optional args are relevant information here too.

As an aside I've found that documentation of a component's public interface (including defaults) is complicated a little by template-only components. The team needs a different convention for documenting its available args (and their defaults). For backed components this is handled fairly naturally in the backing TS/JS file. Are there other documentation conventions for components that I'm not aware of?

> Contrarily, I think putting a defaulting mechanism in the JS/TS file when its only needed in the template is worse (barring some exceptional cases where the defaulting is needed in the JS/TS for other reasons, the logic is complicated and having it in JS/TS makes it easier to reason about and test, etc).

My experience is that it's common for a component to have some args that are only consumed in the template, some args that are only consumed in the JS/TS, and some args that are consumed in both places. Splitting responsibility for defaults such that args only referenced in the template are defaulted there, and args that are referenced in the JS/TS are defaulted there in some other manner seems sub-optimal, ideally it would all be handled in a unified way. Also args may switch where they're being consumed (e.g. when refactoring) which complicates matters if defaults aren't handled via a single system.

Some other possible discussion points:

* Using `this.args` inside the JS/TS to access an argument value becomes possibly the wrong thing to do, as there may be some other property that should be used to get a necessary default value applied.
* Similarly the `@` syntax becomes possibly the wrong thing to do in the template, and it's quite hard for a dev to know that.
* The code example I gave with an arg called `enabled` that defaults to true seems like a plausible situation developers may implement - so what do we recommend to them? It seems a shame to have to backtrack on great conventions such as use `@`/`this.args` when accessing argument values just because a simple default value of `true` is needed.

I can't speak directly to the implementation/architectural complexity here - though I recognise it is an important consideration. Would it still be a significant challenge if the 'defaults' object was immutable was provided once on registration of a component type? The main point here is that there would be no need to access the component instance when falling back to a default argument value. I wasn't intending to suggest that individual component instances be able to customise the defaults in any way, they would apply globally to all instances.

Sorry for the exposition! I've been following the project for a while and felt compelled enough by this to chime in.

pzuraq:
I think you make some great points here. A few thoughts here:

> Is using {{or}} actually going to be recommended practice though? Won't this cause surprises when the argument value passed in is falsey/blank?

FWIW, I think that this behavior is actually desired in many cases, specifically when a value is null or undefined. I do think that this should be handled by a `{{nullish}}` helper, but my point there is that defaulting is not always as simple as "use this value if I wasn't passed anything", and while I think providing some default logic may be convenient, I also believe we will likely see workarounds for many of these cases (as we do today).

> The most immediately useful thing to them is the Args interface (TS-specific, sorry) as it defines the component's primary public interface. Default values of non-optional args are relevant information here too.

Exposing the expected interface of the component in a unified way definitely is important, agree 100% here. I do think that this is a somewhat separate problem from the question of defaults though, specifically because this problem is mostly related to TO-components (since components with a class do have a way to define this interface, as you pointed out), and one of the major benefits of TO-components is that they _do not instantiate a class_, making them more performant than components with a backing class. I think whatever mechanism we use to define this interface, it shouldn't encourage/require creation of a class just to define it.

> Splitting responsibility for defaults such that args only referenced in the template are defaulted there, and args that are referenced in the JS/TS are defaulted there in some other manner seems sub-optimal

I'm not sure I agree here. I think splitting the _interface_ definition is definitely sub-optimal, but the defaulting logic could actually be better being split because it means that the default values are shown where they are relevant, and don't require context switching to find. 

A few final thoughts here:

* I expect most defaulting to occur in a standard way in the component definition, except in very simple cases. `{{this.argOrDefault}}` will be pretty standard in most apps, and there are already standardized solutions popping up for this. From a language perspective, this is not any worse than `{{@arg}}` if `{{@arg}}` could be defaulted. We can see that `{{this}}` is referenced, so we must go to the component definition, and upon doing so we see that it is an argument alias with a default.

  Conversely, if `{{@arg}}` can be defaulted, we _do_ lose the guarantees we talked about before. We must now always check the component definition to know if there was a default applied. So, allowing `{{@arg}}` syntax to be defaulted is strictly worse for `{{@arg}}` syntax, while the alternative of using aliases is pretty comparable.

* A defaulting interface could be added at a later point if we find that it is very difficult for developers to work without one. This would give us more time to figure out design details, such as how it would work for TO-components (maybe it's a function or exported `const` instead of a method on a class?) and would allow us to experiment with different approaches, and figure out how to best integrate it into Glimmer-VM. I think this RFC can move forward as it stands, and we can continue to discuss defaults and develop solutions in the community. If we rush it, we'll be stuck with it.

* I think there could be more novel approaches for defaulting in templates that could help here, for instance maybe a `default-args` helper:

  ```hbs
  {{default-args
    @arg1="default text"
    @arg2=true
  }}

  <div selected={{@arg2}}>{{@arg1}}</div>
  ```

  This could be accomplished today with template transforms, similar to [ember-let](https://github.com/thefrontside/ember-let), allowing us to experiment without committing to a particular design.

richard-viney:
One point to note is that having a way to specify defaults in the JS/TS file doesn't have to mandate a backing component class/context, along with its potential performance drawbacks. Building on what you suggested, we could envision a backing file such as

```ts
export interface Args {
  enabled: boolean
}

export const defaultArgs: Partial<Args> = {
  enabled: true
}
```

Here, the JS/TS file doesn't export a component class at all. This would allow defaults to be specified in a unified way across both normal and TO components, and has other benefits such as a unified place to define/document the component interface that works for TO components too, which I think is highly valuable. It also allows for proper type checking in TO components in future when type-safety in templates is a thing, which wouldn't be possible in current TO components as there'd be no `Args` interface to type against.

Another observation is that the issues raised re. avoiding context switching between the template file and the backing file when reading the code are already going to reduced: first by MU, and then entirely by single-file components (when available). If we take both of these as broadly desirable changes then the concern re. context switching (which is mentioned frequently) would go away. At this point it would be hard to champion any defaulting constructs that were either specific to the use of arguments in templates (which could cause confusion due to divergence from the values held in `this.args`), or that required a backing class for the component (which has performance implications). Hence why the `export const defaultArgs` solution may have some merit.

An additional nicety here is that adding a backing class to a previously TO component becomes less potentially error-prone, as there'd be no figuring out of what arg defaults may be being applied in the template that now need to be moved to the backing class to be used in the JS/TS. Instead you just export the new backing class and continue with `this.args.` and `@`.

I considered suggesting a `{{default-args}}` style construct in the template like you mention, but it being template-specific made it less appealing (similar to the `{{#let}}` idea) as the end goal I'm shooting for here is a unified defaulting approach across both the template and the backing class (if one exists).

> FWIW, I think that this behavior is actually desired in many cases, specifically when a value is null or undefined. I do think that this should be handled by a {{nullish}} helper, but my point there is that defaulting is not always as simple as "use this value if I wasn't passed anything", and while I think providing some default logic may be convenient, I also believe we will likely see workarounds for many of these cases (as we do today).

If we follow JS convention for functions with default arguments then an argument with a value of `undefined` would use its default value if one was provided. An alternative would be to apply the default only when the component invocation does not specify a value for the argument, which I assume Glimmer can detect. Personally I'd vote for the former for parity with JS function argument defaults.

> A defaulting interface could be added at a later point if we find that it is very difficult for developers to work without one. This would give us more time to figure out design details, such as how it would work for TO-components (maybe it's a function or exported const instead of a method on a class?) and would allow us to experiment with different approaches, and figure out how to best integrate it into Glimmer-VM. I think this RFC can move forward as it stands, and we can continue to discuss defaults and develop solutions in the community. If we rush it, we'll be stuck with it.

Agreed, most (all?) of what's being suggested around handling defaults is an addition rather than a fundamental change to the overall Glimmer components RFC, and could be handled in a follow-up RFC based on broader experience and feedback on pain points. We can certainly make do in the meantime with `@arg` and co.

rwjblue:
Thank you @richard-viney and @pzuraq for hammering through the ideas around defaulting and concluding that it is severable into a follow up RFC!

rwjblue:
This RFC was moved into Final Comment Period a little over a week and a half ago, all of the concerns that were raised have been addressed (or will be layered on top of this one in a new RFC): its time has come. 😸 

Thank you to everyone who participated in the process, we absolutely would not have been able to do it without y'all!

mike-north:
Will `@glimmer/component` continue to provide type information as part of its public API?

Issue #376, opened 2018-09-18
Deprecate Ember.Namespace


runspired:
Largely a holdover from the globals era.

http://www.emberjs.com.cn/api/classes/Ember.Namespace.html

I believe it still has some use by the Ember Inspector for registering versions of libraries, but alternative ways of accessing library version information exist.

Gaurav0:
Is DS from Ember Data an Ember.Namespace?

rwjblue:
@Gaurav0 yes

rwwagner90:
@runspired we're using it to get all the app containers I believe. https://github.com/emberjs/ember-inspector/blob/master/ember_debug/vendor/startup-wrapper.js#L161

Is there a better way to do this? As we try to add engines support, I'm also curious if we could add logic here to add in the different engines too?

hjdivad:
Following on to @rwwagner90's comments, `Ember.Namespace.NAMESPACES_BY_ID` is *extremely* useful when debugging, as it gives you devtools console access to apps (and therefore the container, store &c.).

Any deprecation/removal here should retain some happy path for debugging IMHO.

rwwagner90:
Agreed with @hjdivad, I think @runspired was saying this will be replaced with some new way to get the apps/containers, and as long as we retain that functionality, I am good with it 👍

rwjblue:
While I completely agree with the overall point that @hjdivad made (we **must** provide public apis for console debugging), the actual mechanism that @hjdivad mentioned is 1) private api and 2) has a bunch of pitfalls. Accessing the “container” in this way exposes a very rough API that was never intended for access outside of the framework itself, we should instead provide a convienient public API for access the public owner APIs that we are familiar with.

mixonic:
Howdy! We discussed this today at the framework meeting and think this issue is definitely ready to move into an RFC. Namespace should no longer be providing any unique functionality, not that we know of.

rwwagner90:
@mixonic what are the public APIs that replace it? Is there a better way to do things like this? `Ember.Namespace.NAMESPACES[1].__container__.lookup('service:foo')`

nightire:
> @mixonic what are the public APIs that replace it? Is there a better way to do things like this? `Ember.Namespace.NAMESPACES[1].__container__.lookup('service:foo')`

`getOwner(this).lookup('service:foo');` -> https://api.emberjs.com/ember/4.4/functions/@ember%2Fapplication/getOwner

webark:
[...]

PR #417, opened 2018-12-18
Extensible inspector

thorsteinsson:
[Rendered](https://github.com/thorsteinsson/rfcs/blob/extensible-inspector/text/0000-extensible-inspector.md)

I'd like to get feedback before diving deeper into the detailed design.

Big thanks to @rwwagner90 and @acorncom for the help and direction.

### DIFF
--- a/text/0000-extensible-inspector.md
+++ b/text/0000-extensible-inspector.md
@@ -1,0 +1,72 @@
+- Start Date: 2018-09-12
+- RFC PR: (leave this empty)
+- Ember Issue: (leave this empty)
+
+# Extensible Inspector
+
+## Summary
+
+Unlock experimentation for the Ember Inspector by allowing addons to extend it.
+
+## Motivation
+
+Ember itself has increasingly shifted toward a stable core with experimentation through addons and the Inspector could benefit from a similar philosophical shift. It would move maintenance of debugging functionality away from Inspector maintainers. Instead of Inspector maintainers having to carefully weigh ongoing maintenance costs within the Inspector for each new feature, we could move experiments to addons and co-locate debugging functionality within the same addon that provides the functionality.
+
+I first got excited about this after talking to David Baker and the learning team at EmberConf and David wrote a good explanation at
+https://discuss.emberjs.com/t/ember-inspector-call-for-feature-requests-pain-points-and-contributors/15187/20.
+
+### Use cases
+
+*Data layer*
+Allow other data layers to implement devtools, for example Redux.
+
+*Ember addons*
+Increase the debuggability of Ember addons by providing devtools for them.
+
+Few examples:
+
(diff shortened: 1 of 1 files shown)


Gaurav0:
I like the idea of using sandboxed iframes for security; my largest concern would have been addons that would use this capability to do the sorts of things only chrome extensions normally could do like getting around the application's content security policy and accessing any 3rd party site.

However, people will chafe under many of the restrictions that sandboxed iframes have, like they do today with Ember Twiddle.

lukemelia:
This approach seems good from a security and architecture perspective. My initial thought is that extracting ember-data to run in this way could be a good path into this architecture. However, I am struck that not having the ember-data functionality of inspector in production would be a big loss.

Perhaps there could be a hybrid approach where all addons can take the serve devtools package from ember-cli in development, and blessed addons can also publish packages that are consumed by the ember-inspector build process.

Additional thought, what API compatibility issues/constraints does this introduce? Right now, inspector has to worry about backward compatibility with ember apps running at different versions of ember. Inspector would now also have to worry about addon inspector plugins coded to different versions of the addon inspector API.

rwwagner90:
@lukemelia I agree that attempting to do this with ember-data could be a good testbed for figuring out the path for other addons.

> Additional thought, what API compatibility issues/constraints does this introduce? Right now, inspector has to worry about backward compatibility with ember apps running at different versions of ember. Inspector would now also have to worry about addon inspector plugins coded to different versions of the addon inspector API.

Inspector's master branch only supports newer Ember, and uses the new public APIs that @chancancode helped expose for us 🎉. As long as we can get everything in Inspector using public API, I think it should be relatively stable and we wouldn't have to worry about ever changing APIs.

acorncom:
> Inspector would now also have to worry about addon inspector plugins coded to different versions of the addon inspector API.

@rwwagner90 I believe Luke was concerned more with Inspector APIs changing over time and needing to support different approaches there. I wonder if we could explore a capabilities approach akin to managers in Ember and work to be strictly additive in how we increase plugin APIs over time?

rwwagner90:
@acorncom I think whether we are talking about Inspector consuming APIs or Inspector exposing APIs, we should be safe as long as we make sure we always use public, supported APIs. Then if things had to change from the existing public API, we would have to do a major version bump.

wagenet:
@thorsteinsson, @rwwagner90, @acorncom is there still interest in moving this forward? If so, I'll make sure that it gets a more proper review.

webark:
this kind of ties into some of the discussion around #376

acorncom:
@wagenet I think there's still real benefit to the ecosystem over time to doing something like this. Yehuda was saying something on Twitter recently about separately deployable engines being an important part of Polaris. The RFC here mentions that being a possible way to address the idea, so solving deployable engines might in turn unlock this as well ...

Issue #304, opened 2018-02-05
Deprecate `Route#render`/`Route#renderTemplate`


locks:
Part of the [deprecations initiative](https://github.com/emberjs/rfcs/projects/1).

PR #418, opened 2018-12-19
Deprecate Route Render APIs

chadhietala:
[Rendered](https://github.com/emberjs/rfcs/blob/01af231ad8c1275424eba457f1241469d7b33b77/text/0418-deprecate-route-render-methods.md)

Closes https://github.com/emberjs/rfcs/issues/304.

### DIFF
--- a/text/0418-deprecate-route-render-methods.md
+++ b/text/0418-deprecate-route-render-methods.md
@@ -20,7 +20,7 @@ The usage of `render` is deprecated. Please see the following deprecation guide 
 The following will be compile time deprecation for named outlets:
 
 ```
-Please refatctor `{{outlet <NAME>}}` to a component <SOURCE_INFO>.
+Please refactor `{{outlet <NAME>}}` to a component <SOURCE_INFO>.
 ```
 
 ## Motivation
--- a/text/0418-deprecate-route-render-methods.md
+++ b/text/0418-deprecate-route-render-methods.md
@@ -6,7 +6,22 @@
 
 ## Summary
 
-This RFC proposses the deprecation of [`Route#render`](https://emberjs.com/api/ember/3.6/classes/Route/methods/render?anchor=render), [`Route#renderTemplate`](https://emberjs.com/api/ember/3.6/classes/Route/methods/render?anchor=renderTemplate) and named `{{outlet}}` APIs.
+This RFC proposses the deprecation of [`Route#render`](https://emberjs.com/api/ember/3.6/classes/Route/methods/render?anchor=render), [`Route#renderTemplate`](https://emberjs.com/api/ember/3.6/classes/Route/methods/render?anchor=renderTemplate) and named `{{outlet}}` APIs. The following deprecation message will be emitted upon usage of `render` or `renderTemplate`:
+
+```
+The usage of `renderTemplate` is deprecated. Please see the following deprecation guide to migrate.
+```
+and
+
+```
+The usage of `render` is deprecated. Please see the following deprecation guide to migrate.
+```
+
+The following will be compile time deprecation for named outlets:
+
+```
+Please refatctor `{{outlet <NAME>}}` to a component <SOURCE_INFO>.
+```
 
 ## Motivation
 
@@ -232,7 +247,16 @@ export default Component.extend({
 
 # How We Teach This
 
-This has not been a mainline API for quite some time now. The guides do not mention this functionality at all. It is likely that the majority of Ember applications do not use these APIs.
+These APIs not been a mainline API for quite some time now. The guides briefly mention this functionality. In those cases we should mirgate the guides should link to the `{{in-element}}` documentation and the component documentation. The above "Transition Path" will serve as the deprecation guide.
+
+# Role Out Plan
+
+Prior to adding the deprecation we must first do the following items
+
+* [ ] The `{{in-element}}` helper implementation remains incomplete. It should be completed.
+  * [ ] The [small changes](https://github.com/emberjs/rfcs/blob/master/text/0287-promote-in-element-to-public-api.md#small-proposed-changes) section of the `{{in-element}}` RFC needs to be implemented. Specifically the helper should "replace all the content of the destination".
+  * [ ] The `{{in-element}}` helper should be documented in the [API docs](https://emberjs.com/api/ember/release/classes/Ember.Templates.helpers).
+  * [ ] Adding `{{in-element}}` usage to the guides can be considered.
 
 ## Drawbacks
 
--- a/text/0418-deprecate-route-render-methods.md
+++ b/text/0418-deprecate-route-render-methods.md
@@ -10,11 +10,15 @@ This RFC proposses the deprecation of [`Route#render`](https://emberjs.com/api/e
 
 ## Motivation
 
-These APIs are largely holdovers from a time where components where not as prominent in your typical Ember application. While they are still documented, these APIs created an interesting coupling between the `Route` and the template. These APIs are also prone to breaking conventional naming conventions, which can lead to confusion for developers.
+These APIs are largely holdovers from a time where components where not as prominent in your typical Ember application. While they are still documented, these APIs created an interesting coupling between the `Route` and the template. These APIs are also prone to breaking conventional naming conventions, which can lead to confusion for developers. Another issue is that it is unclear how something like this works with route based tree shaking, as there are no strong conventions or direct imports as to what is actually being used.
 
 ## Transition Path
 
-The migration plan here is to just move to components for all the use cases you would use theses APIs for. For example, there was a time when you would do something like the following if you wanted to have a master-detail view for a shopping app.
+The migration plan here is going to be somewhat situational based on the UI that was being constructed. For cases where named outlets were being used it is likely that they should just be moved to components. For cases where you were escaping the existing DOM hierarchy to render a template somewhere else in the DOM, one should use the built-in [`{{in-element}}`](https://github.com/emberjs/rfcs/blob/master/text/0287-promote-in-element-to-public-api.md) helper or an addon like [ember-elsewhere](https://github.com/ef4/ember-elsewhere). Below are some example of how a migration would look.
+
+__Migrating Named Outlets__
+
+Given:
 
 ```js
 Ember.Route.extend({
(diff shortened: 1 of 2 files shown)


rtablada:
I'm a HUGE 👍 on this (though we do have some legacy features that rely on route rendering).

This really simplifies the route/template and rendering relationship across the broader ecosystem. Especially when advising older projects that rely heavily on `Route.prototype.render` IMO having a dedicated migration guide and deprecating is better than our current silence (and support burden) in documentation.

---

From a migration strategy I think some migration guides around common scenarios which from experience route rendering can usually be replaced by one of a few strategies:

1. Match the template name to the route (for the simple case where these got decoupled for some reason)
2. Create a component and pass in attrs (in the case where a template may have been shared across multiple routes using the `render` method to populate
3. Use a parent route and template (often when all children routes have the same UI populated but failed to take advantage of parent routing and outlets)
4. Use components in a parent route (usually application) with a service tied to it to facilitate state population, changes, and communication
5. Use wormholes to push components or UI fragments into what is a common use case for render + named outlets

jgwhite:
I quickly grep’d the Heroku Dashboard code base to see how we’d be impacted by this. Turns out we use `Route#render` quite extensively — 38 calls in total.

The vast majority deal with rendering ancillary UI into top-level outlets (e.g. top-nav, action-bar). Using wormholes seems like an appropriate migration path for these.

The other usages are to switch between different templates for an error route. Presumably the appropriate migration path for this is to turn those templates into components.

chadhietala:
@jgwhite yea this was typically the case. 

@Gaurav0 I need to check. If not I need to update this RFC and call out that you can still have different route templates.

knownasilya:
Here's a scenario that uses `render`/`renderTemplate` https://github.com/embermap/ember-cli-tailwind/blob/master/app/instance-initializers/ember-cli-tailwind.js#L8, although this might be the wrong tool for the job, but I'm unsure what the right tool would be.

benedikt:
In one of my projects I use `renderTemplate` and `render` to replace a sidebar layout of a parent route for certain sub-routes. How would I achieve a similar setup without those methods?

I tried to visualize the situation in the image below. I hope it makes sense in combination with the code snippets.

```js
// router
this.route('authenticated', { path: '' }, function() {
  this.route('campaigns', { resetNamespace: true }, function() {
    this.route('campaign', { path: ':campaign_id' }, function() {
      this.route('messages', function() {
        this.route('message', { path: ':message_id' });
      });
    });
  });
});
```

```js
// campaign.messages 
renderTemplate() {
  this.render({ into: 'authenticated' });
}
```

layout

lifeart:
I use `renderTemplate` for route templates hot-reloading in `ember-ast-hot-loader` - https://github.com/lifeart/ember-ast-hot-load/blob/master/addon/services/hot-loader.js#L143
If we drop it, we need some way to trigger route template rerendering.

rmachielse:
We make a lot of use of render and renderTemplate as well, to render templates in different places. For example dialogs and fullscreen components that can’t be part of the html tree but have to be below it instead. We have a few named outlets for this in application.hbs. I don’t see how this can be replaced with components, as there isn’t just a single one for each outlet but there is a number of components that has to be rendered there depending on the route. The only way for us to solve this would be to allow these things to be rendered in the tree css-wise I guess. In any case this would leave us with a huge architectural question.

knownasilya:
Yeah maybe that is true about it being a type of work around, although things like ember-elsewhere are using public API https://github.com/emberjs/rfcs/pull/287 so they are very lightweight.

I do like the how ember-component-routes addon handles this, feels very clean: https://wongpeiyi.github.io/ember-component-routes/#/rendering. e.g.

```js
import { ComponentRoute } from 'ember-component-routes';

export default ComponentRoute.extend({
  renderComponents() {
    this.renderComponent('app-menu', { outlet: 'sidebar' });
    this.renderComponent('post-page', { into: 'parent-page' });
  }
});
```

So maybe an alternative API should be proposed, to at least allow experimentation like above.

rmachielse:
That one looks cleaner indeed. However, as long as ember has controllers and routes, `renderTemplate` looks like a very valid need and I don't see any big advantages in getting rid of it. I do see the complexity argument and I do recognize that this is probably true for smaller applications. However for larger applications like ours, `renderTemplate` reduces complexity big time, as most developers can keep on developing templates and controllers as always, while the template may be rendered out of the tree. So yes, maybe this is an advanced feature that shouldn't distract beginning developers too much, but it is one that makes our life much easier.

rmachielse:
@LevelbossMike the separation of concerns is the most valuable thing to me. I think the decision of where to render a template should not be a responsibility of the template itself. Also in our case it allows us to reuse templates, without having to change anything in the template itself.

rmachielse:
If this functionality is considered to be too far away from core principles, could it be extracted to an addon? Not one that extends support temporarily but one that can be maintained for longer time? Or does it depend on internal api's too much?

knownasilya:
Deprecation doesn't mean it will be removed, so it should be around until Ember 4.0, which is still probably a year away most likely (imo).

rmachielse:
Understood, but I would like a future-proof solution.

chadhietala:
I have updated the RFC with clearer examples on how to migrate.

@rmachielse: #287 introduced a public API to do what ember-elsewhere / ember-wormhole do. It was created specifically for the cases where you do need to escape the existing DOM hierarchy and render somewhere else. This is great for things like modals, adding to navigation depending on downstream logic, etc. In the cases where you do need access a controller for a specific template and don't want to refactor to a component you can [inject controllers](https://emberjs.com/api/ember/3.6/functions/@ember%2Fcontroller/inject) onto other objects. The opinion of the core team is that components is the way templates should be reused.

@Gaurav0 `templateName` requires another RFC and likely some investigation of migration strategy.

lifeart:
@chadhietala any way to force  route template rerender without `renderTemplate` call?

chadhietala:
@lifeart have you seen https://github.com/emberjs/ember.js/pull/17000 and https://github.com/emberjs/ember.js/pull/17001 ?

lifeart:
@chadhietala of cource! I made [ember-ast-hot-reload addon](https://github.com/lifeart/ember-ast-hot-load). And it allow users to hot-reload route templates, using `renderTemplate` method. How I can replace it if it will be deprecated?

chadhietala:
We decided to move this to FCP. 

@lifeart, @wycats said we should find a way to expose an API in the glimmer-vm to better support hot-reloading.

Kerrick:
We use these APIs in dozens of places across our very large Ember application. For context, our router.js file has 162 instances of `this.route(`, and nests up to 7 routes deep.

Our use case is that a large portion of our routes render `into: 'application'` in order to _not_ maintain navigation etc. from the higher-level routes. We use Routes + Controllers + Route Templates nested as normal so we can read our parent routes' model data, but we render those Route Templates into the top level of the DOM. The code examples provided in the RFC did not make the migration path clear to us.

To illustrate, here are three of our screens, with the URL visible so you can see the route nesting coming into play. We simply want the user to feel like "I'm in a completely new application now" when they navigate into a child route. It's the quintessential use case for using the proposed-to-be-deprecated methods, and it doesn't seem to have a replacement suggested in the RFC (unless I'm missing something).

screen shot 2019-01-30 at 3 18 06 pm
screen shot 2019-01-30 at 3 18 25 pm
screen shot 2019-01-30 at 3 18 50 pm

Kilowhisky:
Just to throw my 2c in here i use named outlets to provide a hook for sub-routes to render into the main menu system on both mobile and desktop. This allows for main navigation to be at a 'portal' level and for sub routes to not be concerned about item placement. 

For instance my 'portal' template has this in it.
```
<div class="nav>
                   [..... some other nav stuff]
                   {{outlet 'tool-bar'}}
</div>
```
At another location in the template it has the mobile menu stuff.
```
<div class="nav-mobile">
                [.... mobile specific nav stuff]
                {{outlet 'tool-bar-mobile'}}
</div>
```

A sub-route then outputs into the parent template by doing this.
```
    /**
     * Renders the default template and the tool-bar in the tool-bar outlet
     */
    renderTemplate: function() {
        this.render();
        this.render('portal.track.tool-bar', {
            outlet: 'tool-bar'
        });
        this.render('portal.track.tool-bar-mobile', {
            outlet: 'tool-bar-mobile'
        });
    },
```

This results in a sub route that can switch menu items out as the user navigates around. For instance the highlighted items in the picture below are specific to the current route the user is in.

image

Here is what that same view looks like on mobile screen. 

image

Using named outlets lets me alter the overall structure based on things like resolution. These templates are then full fledged members of the parent view and that means i can share data and actions at the sub-route level instead of having to orchestrate everything at the portal level.

chadhietala:
@Kerrick and @Kilowhisky please take a look at the follow example. I believe both of the concerns are around escaping the hierarchy and I believe this can be solved with `{{in-element}}` or an addon like ember-eslewhere.

rwjblue:
> It works well for us, it is simple to reason about and every other alternative pattern we have tried is a PITA to reason about and has worse rendering performance.

I’d like to understand this better. I find the alternatives here **massively** easier to reason about, what makes them seem overly complicated in your scenarios? I’m also curious how the proposed alternatives would have any different performance at all.

nullrocket:
Maybe I should have tried the in-element alternative before I posted the above. We had not tried that pattern, we have been using ember for so long that we settled into our current patterns a couple years ago.  I am still finding renderTemplate easier to keep a mental model of but that could be because I am used to our projects being organized that way. 

I built two twiddles of the same dummy app, one using renderTemplate one using the suggested `{{in-element}}` helper.  I notice a couple differences so far, using renderTemplate  I have a component at the application level template that handles layout, with named outlets I can render just the sections that change when navigating between routes leaving the unchanged things like a common sidebar untouched. I can also use the file system to keep a mental model of where those changed templates are like `/route-one/sidebar/template.hbs ` With the `{{in-element}} `helper I have to re-render the same sections on each route change.  

Maybe if I understood why the change was necessary it would help me be more excited about it.  I'm not seeing many other comments saying they want the change, of course the core team probably has more channels than these comments to base a decision on.

**Using renderTemplate**
https://ember-twiddle.com/c4119b26904fe3f2af7ac04a7e36a6d4?openFiles=application.template.hbs%2Crender-to-element.template.hbs&route=%2Froute-two

**Using `{{in-element}}`**
https://ember-twiddle.com/b257d76afae9ed6f04b663d6360cf04a?openFiles=application.template.hbs%2Crender-to-element.template.hbs

benedikt:
Following @nullrocket's example, I tried to implement my `renderTemplate` use-case with both `renderTemplate` and `{{in-element}}`. I didn't quite get it to work in the same way. What am I missing (or doing wrong)? 

* Using `renderTemplate`: https://ember-twiddle.com/7d32412e6a88b0e837b7a01c06d6e310
* Using `{{in-element}}`: https://ember-twiddle.com/2a87d0ebb8e151cf1b6919c7db241cdb

benedikt:
@LevelbossMike Aah, thanks! That explains my issue. Looks like this will work for my use-case indeed. 👍

text/0418-deprecate-route-render-methods.md
```diff
@@ -0,0 +1,249 @@
...
+  init() {
+    this._super(...arguments);
+    this.accountPlaceholder = document.getElementById('account-placeholder');
+  }
+});
+```
+
+# How We Teach This
+
+This has not been a mainline API for quite some time now. The guides do not mention this functionality at all. It is likely that the majority of Ember applications do not use these APIs.
```
mixonic: > This has not been a mainline API for quite some time now.

I think this was supposed to read "This API has not existed for very long"?

chadhietala: I'm referring to the fact that `renderTemplate` and the associated APIs do not and have not played a predominate role in the programming model for quite some time now.

mixonic:
I'd like to suggest a first draft of the steps we need to take to implement this RFC (mostly cribbed from best practice, I don't think we have any of this written down):

* [ ] The `{{in-element}}` helper implementation remains incomplete. It should be completed.
  * [ ] The [small changes](https://github.com/emberjs/rfcs/blob/master/text/0287-promote-in-element-to-public-api.md#small-proposed-changes) section of the `{{in-element}}` RFC needs to be implemented. Specifically the helper should "replace all the content of the destination".
  * [ ] The `{{in-element}}` helper should be documented in the [API docs](https://emberjs.com/api/ember/release/classes/Ember.Templates.helpers).
  * [ ] Adding `{{in-element}}` usage to the guides can be considered.
* [ ] The **Transition Path** section of this RFC should be considered a first draft for the [deprecation guide](https://emberjs.com/deprecations/v3.x).
* [ ] Content for the deprecation message itself needs to be drafted in a PR to Ember.

text/0418-deprecate-route-render-methods.md
Gaurav0: This API existed in Ember 1.0, so it definitely "has existed for very long". :)

chadhietala:
@mixonic Can you review the latest changes and confirm that they address your process concerns?

text/0418-deprecate-route-render-methods.md
```diff
@@ -0,0 +1,273 @@
...
+  init() {
+    this._super(...arguments);
+    this.accountPlaceholder = document.getElementById('account-placeholder');
+  }
+});
+```
+
+# How We Teach This
+
+These APIs not been a mainline API for quite some time now. The guides briefly mention this functionality. In those cases we should mirgate the guides should link to the `{{in-element}}` documentation and the component documentation. The above "Transition Path" will serve as the deprecation guide.
```
locks:
```suggestion
These APIs not been a mainline API for quite some time now. The guides briefly mention this functionality. In those cases we should migrate the guides should link to the `{{in-element}}` documentation and the component documentation. The above "Transition Path" will serve as the deprecation guide.
```

text/0418-deprecate-route-render-methods.md
```diff
@@ -0,0 +1,273 @@
...
+
+# Role Out Plan
+
+Prior to adding the deprecation we must first do the following items
```
locks:
```suggestion
We must first do the following items prior to adding the deprecation:
```

text/0418-deprecate-route-render-methods.md
```diff
@@ -0,0 +1,273 @@
...
+
+The drawback of this is that it is churn for applications that are relying heavily of these imperative APIs to construct their UI. They will need to encapsulate and use the existing declarative APIs.
+
+## Alternatives
+
+No real alternatives as we want to move away from these style of imperative APIs in favor of declarative ones.
+
+## Unresolved questions
+
+Optional, but suggested for first drafts. What parts of the design are still
```
locks:
```suggestion
```

text/0418-deprecate-route-render-methods.md
```diff
@@ -0,0 +1,273 @@
...
+TBD?
```
locks:
```suggestion
```

text/0418-deprecate-route-render-methods.md
mixonic: IMO it would be helpful for this sentence to state its subject. I think @chadhietala's wording in his response to be above was just better:

> `renderTemplate` and its associated APIs do not and have not played a predominate role in Ember's common programming model for quite some time. The guides only briefly mention this functionality. In those guides cases that are documented (maybe should link to them here) the guides should link to the `{{in-element}}` documentation and component documentation as is appropriate.
>
> The "Transition Path" section above will serve as a first draft of the deprecation guide.

mixonic:
@chadhietala Thank you for the continued cleanup on this after it had theoretically already been approved for merge :-/ Please consider my previous comment about next steps addressed by your changes.

I think you already plan to raise the confusing state of the `{{in-element}}` public API (`nextSibling` vs `insertBefore`) at today's framework core meeting, and IMO resolving that could be an additional item in the TODO list you included in this RFC.

Thanks Chad.

rwjblue:
> I think you already plan to raise the confusing state of the `{{in-element}}` public API (`nextSibling` vs `insertBefore`) at today's framework core meeting

Yep, the confusion is cleared up on that front.

rwjblue:
Thanks for iterating on this @chadhietala / @mixonic! 

I think this is in a good place to land (only a few weeks after we meant to 😸)...

mixonic:
> Yep, the confusion is cleared up on that front.

Haha @rwjblue can you drop a link or summarize to break the suspense?

rwjblue:
@mixonic hmmm, I don't follow, you laid out the specifics in https://github.com/emberjs/rfcs/pull/418#issuecomment-461939795:


> * [ ]  The `{{in-element}}` helper implementation remains incomplete. It should be completed.
>   
>   * [ ]  The [small changes](https://github.com/emberjs/rfcs/blob/master/text/0287-promote-in-element-to-public-api.md#small-proposed-changes) section of the `{{in-element}}` RFC needs to be implemented. Specifically the helper should "replace all the content of the destination".
>   * [ ]  The `{{in-element}}` helper should be documented in the [API docs](https://emberjs.com/api/ember/release/classes/Ember.Templates.helpers).
>   * [ ]  Adding `{{in-element}}` usage to the guides can be considered.

You were totally right, `{{in-element` will have replace by default semantics, and to get "append" (currently what `{{-in-element` does) you will need to specify `insertBefore=null`.

rwjblue:
FWIW, @chadhietala is working on those changes over in https://github.com/glimmerjs/glimmer-vm/pull/918.

mixonic:
@rwjblue right, when I say:

> the confusing state of the `{{in-element}}` public API (`nextSibling` vs `insertBefore`)

I'm definitely talking about that `nextSibling` is/was the actual implemented API in Glimmer and Ember, but the RFC to promote `{{in-element}}` only mentioned `insertBefore`. I see in today's core team notes:

> RJ: Concretely, forget about next-sibling, only support insert-before. It’s a better semantic naming.

So the item I was suggesting could be added or at the least should be tracked is that https://github.com/glimmerjs/glimmer-vm/pull/918 and the sibling PR to Ember should be migrating to `insertBefore` and dropping `nextSibling` (I'm not sure if it needs intimate API deprecation or not).

IBue:
To me, the `renderTemplate/render` combo inside a Route actually has a very declarative character. It enables VERY easily mounting a nested details-route's template-subtree somewhere up in the template hierarchy (as in @benedikt 's case) or even into the application template via main `{{outlet}}`s for some sort of intermediate 'fullscreen' view of e.g. editing forms without any specific CSS intervention or concern about data context.
Hopefully, the lately specified replace-innerHtml behaviour of `{{in-element}}` will provide an almost 1:1 transition path by the following pattern:

target route template: `<div id="targetSlotX">{{outlet}}</div>`
source route template: `{{#in-element targetSlotX}}…{{outlet}}…{{/in-element}}`

NullVoxPopuli:
@ibue, you may be interested in: https://github.com/emberjs/rfcs/issues/499

Issue #420, opened 2018-12-19
Deprecate Application controller special properties:


chadhietala:
- currentRouteName
- currentURL
- ...

PR #421, opened 2018-12-19
Deprecate Application Controller Router Properties

chadhietala:
[Rendered](https://github.com/emberjs/rfcs/blob/deprecate-application-controller-props/text/0000-deprecate-application-controller-props.md)
Closes https://github.com/emberjs/rfcs/issues/420


lholmquist:
The rendered link still points to the old name of the RFC,  i think this is the updated link https://github.com/emberjs/rfcs/blob/deprecate-application-controller-props/text/0421-deprecate-application-controller-props.md

PR #425, opened 2018-12-22
Website Redesign RFC

wifelette:
[Rendered](https://github.com/wifelette/rfcs/blob/master/text/0425-website-redesign.md).

PR #431, opened 2019-01-13
Restructuring the Guides Table of Contents

jenweber:
[Rendered](https://github.com/jenweber/rfcs/blob/guides-restructure/text/0431-guides-restructure.md)

### DIFF
--- a/text/0000-guides-restructure.md
+++ b/text/0000-guides-restructure.md
@@ -72,7 +72,7 @@ The following Table of Contents will be applied iteratively over the course of m
 - Testing
 - Configuration
 - Deploying
-- Upgrading
+- Upgrading (will contain links to upgrade resources and syntax conversion guides)
 - Debugging (currently named Ember Inspector)
 
 ### Topic naming
--- a/text/0000-guides-restructure.md
+++ b/text/0000-guides-restructure.md
@@ -188,23 +188,8 @@ Community buy-in is important to reduce perception of churn and encourage percep
 We will also test an early beta restructuring with beginner-level developers and developers who don’t know Ember. Although the rollout for the live site will take months, a rough cut, undeployed, could be completed in 1-2 weeks. It could serve as a North Star for the work to be done.
 
 ### Implementation plan
+
 This work will need to be done over many months/the next year, and will be done incrementally in the deployed guides. The plan prioritizes the refactors that will make Octane Guides writing easier. It will be communicated in the form of Quest issues, with help requested via the Ember Times, Discord, Discuss, etc.
-
-1. Settle on a Table of Contents layout*
-2. Make the refactors necessary for Components to come before Templates*
-3. Switch the order in the live Guides*
-4. Work Computed Properties into the Components section*
-5. Create an Upgrading section (this will house the Octane conversion guide)*
-6. Write What is Ember. This will be where we give a light intro into how the pieces fit together, and link to CLI guides resources on the file structure*
-7. Make the refactors necessary to add Controllers into the Routes explanations
-8. Group Controllers with Routes in the live Guides
-9. Make Services its own topic
-10. Rename “Glossary” to “JavaScript concepts” and move the JS Primer there
-11. Create Deploying section. Low on content, links to CLI Guides resources.
-12. Create Styles section. Low on content, links to CLI Guides resources.
-13. Rename Models to “Connecting to an API”
-
-*must be done before the first Octane edition release
 
 ## Drawbacks
 
--- a/text/0000-guides-restructure.md
+++ b/text/0000-guides-restructure.md
@@ -177,7 +177,11 @@ Notably missing is "Ember Object Model." This is on purpose. It will be pulled i
 
 ### Why doesn’t this RFC include rewriting content?
 
-Individual pages have already been refactored over the past two years by many contributors. Examples include the Ember Objects page, Controllers, using third party libraries, and explanations of data management. Many of the pain points that current Ember devs remember from their early days have been fixed. For example, it’s clear that Ember Data/JSONAPI aren’t mandatory, that you *can* use things like fetch, that Computed Properties need to be consumed for them to fire… we’re in a pretty good place! If we choose a good structure for the Table of Contents, it will make it much easier to write/rewrite individual sections.
+Individual pages have already been refactored over the past two years by many contributors. Examples include the Ember Objects page, Controllers, using third party libraries, and explanations of data management. Many of the pain points that current Ember devs remember from their early days have been fixed. For example, it’s clear that Ember Data/JSONAPI aren’t mandatory, that you *can* use things like fetch, that Computed Properties need to be consumed for them to fire… we’re in a pretty good place! Those improvements of individual topics may continue without the need for an RFC.
+
+If we choose a good structure for the Table of Contents, it will make it much easier to write/rewrite individual sections.
+
+It is also important to note here that that Ember does not have a foundation or funding arm, so although it would be great to have a dedicated writer tackle this from the ground up, we must choose an approach that would be realistic for a group of volunteers to achieve.
 
 ### Technical approach
 
@@ -386,6 +390,7 @@ All three of the libraries above cover forms in their own dedicated section. The
 - What possible pain points does the community see?
 - Are there any areas missing from the Table of Contents?
 - What do people think of removing the “Ember Object Model” section?
+- Are the goals of the Ember Data and Ember Inspector teams supported by this new layout?
 
 ### One last note
 
--- a/text/0000-guides-restructure.md
+++ b/text/0000-guides-restructure.md
@@ -191,6 +191,29 @@ We will also test an early beta restructuring with beginner-level developers and
 
 This work will need to be done over many months/the next year, and will be done incrementally in the deployed guides. The plan prioritizes the refactors that will make Octane Guides writing easier. It will be communicated in the form of Quest issues, with help requested via the Ember Times, Discord, Discuss, etc.
 
+Community members will be asked to participate in PR review. A diversity of technical experience levels, English-language proficiency level, and perspectives will create a stronger output.
+
+The intial steps will aim for the quick wins and the urgent changes that aid in Octane documentation. 
+
+Here's what we could expect a minimal first pass to look like:
+
+- What is Ember?
+- Getting Started
+- Tutorial (already in progress of being split out)
+- Templates
+- Components
+- Routing (includes Routes, Router, and Controllers)
+- State management (aka computed properties)
+- Services
+- Ember Data (renamed from "Models")
+- Addons and dependencies
+- Testing
+- Configuration
+- Application Concerns
+- Ember Inspector
+
+The groupings like "Fundamentals" require architectural work on the guides app. They can be done in parallel depending on volunteer capacity and interest, or delayed until the end.
+
 ## Drawbacks
 
 This refactor is biased towards new user experience, so existing Ember users could experience the most drawbacks.
@@ -356,3 +379,7 @@ All three of the libraries above cover forms in their own dedicated section. The
 - What possible pain points does the community see?
 - Are there any areas missing from the Table of Contents?
 - What do people think of removing the “Ember Object Model” section?
+
+### One last note
+
+The Guides are one of those things where everyone has an opinion, and that's ok! However, as a reminder, please be kind and constructive in your comments. The Guides and the tools they cover are the work of many dedicated community members. They are authored and maintained through great volunteered and donated effort. Although we always know there is room for improvement, we are proud of where we came from and excited for where we're going next.
--- a/text/0000-guides-restructure.md
+++ b/text/0000-guides-restructure.md
@@ -10,9 +10,9 @@ As our favorite framework has grown and changed a lot over the past few years, s
 
 ## Motivation
 
-The current structure and flow of the guides reflects the past, not the present experience for Ember developers. It’s time to fix that. With Octane coming down the line, and an influx of new Ember developers, it’s more important than ever to consider the overall learning experience.
+With Octane coming down the line, and an influx of fresh attention, it’s more important than ever to consider Ember's overall learning experience. It's time for a shift. The current structure and flow of the guides reflects the past, not the present experience for Ember. It’s time to fix that.
 
-At the time the Guides were originally written, things like object-oriented programming in JavaScript, components, routing, and SPAs in general were new ideas. If you look at our guides today, they start by teaching the “Ember Objects” programming concept, in a way that few beginners will use in their apps. However, this made sense at the time, because it would have been so unfamiliar to developers. Now, the learning hierarchy needs have shifted from “learn these new concepts” to “learn how the familiar pieces fit together so you can build quickly.”
+When the Guides were originally written, things like object-oriented programming in JavaScript, components, routing, and SPAs in general were new ideas. For example, they start by teaching the “Ember Objects” programming concept, necessary background information at the time. It would have been an unfamiliar pattern at the time. Now, the learning hierarchy needs have shifted from “learn these unique web development concepts” to “learn how familiar pieces fit together so you can build great apps quickly.”
 
 ## Detailed design
 
@@ -20,17 +20,18 @@ The focus of this RFC is not to write new content, but make better use of what i
 
 The overall learning strategy is to establish a common core of sequential knowledge, and then later topics can be read standalone, skipping around. What this means is that if beginners read through to routes, they could skip around the topics lower down the chain successfully.
 
-Finally, it is important to acknowlegde that the Guides are intended to represent the "Happy Path" of using Ember. It is not possible to have something that is a perfect fit for everyone's app needs, nor should it cover our entire API surface.
+Finally, it is important to acknowlegde that the Guides are intended to represent the "Happy Path" of using Ember. It is not possible to have something that is a perfect fit for everyone's app needs, nor should they cover our entire API surface.
 
 ### Process and preparation
 
 This Table of Contents plan was developed over the course of many months, meetings, and writing sessions. It includes input from many people in the community. Here are the sources of inspiration, information, and planning:
 
-- Chris Garrett's ([@pzuraq](https://github.com/pzuraq)) 2018 Roadmap Blog Post, [Ember as a Component-Service Framework](https://medium.com/@pzuraq/emberjs-2018-ember-as-a-component-service-framework-2e49492734f1), May 2018
-- My own Roadmap blog post reflections, [Be loud and be ready](https://gist.github.com/jenweber/a9fbea98478fc3841fb8b24f7dc961c8), May 2018
+- Chris Garrett's ([@pzuraq](https://github.com/pzuraq)) 2018 Roadmap Blog Post, [Ember as a Component-Service Framework](https://medium.com/@pzuraq/emberjs-2018-ember-as-a-component-service-framework-2e49492734f1)
+- My own Roadmap blog post reflections, [Be loud and be ready](https://gist.github.com/jenweber/a9fbea98478fc3841fb8b24f7dc961c8)
 - Extensive discussion in Ember Learning Team Meetings, which are held weekly and open to the public
 - The [2018 Roadmap RFC](https://github.com/emberjs/rfcs/blob/26c4d83fb66568e1087a05818fb39a307ebf8da8/text/0000-roadmap-2018.md) by [Tom Dale](https://github.com/tomdale)
 - The Octane Strike Team meetings and followup conversations
+- Hands-on writing of a mini guide for Octane
 - Ember.js Framework Core Team meetings
 - A [Twitter poll](https://twitter.com/jwwweber/status/1081352702083452928), "Name two things you think the guides should do well/better"
 - Discussion on Discord of the initial Table of Contents draft with some key contributors and content creators
@@ -42,9 +43,11 @@ This Table of Contents plan was developed over the course of many months, meetin
 - The successful restructure and rewrite of the [Ember CLI Guides](https://cli.emberjs.com) (a far more drastic project than this RFC aims to be)
 - Attempting to reorder the Guides in a branch, as an experiment
 
+We are confident that this is possible thanks to the incredible response and effort shared by the community for the CLI Guides work.
+
 ### Table of Contents
 
-This is the aspirational version. There will be many intermediate states between here and there, outlined later in this RFC.
+The following Table of Contents will be applied iteratively over the course of many months. The quick wins and urgent Octane refactors will be applied ASAP, and are described in the Implementation section of this RFC. Justification for the ordering is provided following the topics list.
 
 **Core Concepts**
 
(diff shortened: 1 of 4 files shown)


lifeart:
Is it possible to add kinda `ember-auto-import` into "Working with JavaScript", to get users easy way to include any npm installed module?

In components section it will be nice to have different components types overview. (https://www.chriskrycho.com/2018/higher-order-components-in-emberjs.html, https://medium.com/macsour/components-patterns-in-ember-js-5e6fc6eea28f)

Also, we should point to end-users template-only components it's a kinda component without owned context, not just template.

jenweber:
@lifeart good idea, I agree that we should definitely tell devs about it in "Working with JavaScript." It's covered in Addons & Dependencies, and we can link to it. I also agree that we should have something about CSS. This RFC aims to do the bare minimum to rearrange what we have and prepare for Octane, so CSS is somewhat out of scope.

As we consider the Table of Contents, it's likely that there will be many excellent "out of scope" ideas/reflections that can be turned into action items, so I just created a Gist here that anyone can comment on, and I'll add to it as we go along: [Community Ideas for the Ember.js Guides](https://gist.github.com/jenweber/5d9083433deddaa63c835a4ff1901803)

pzuraq:
This RFC is absolutely excellent! I especially like the clear and concise rules for defining the "Fundamentals" and "Leveling Up" sections, I think that they will be great guides for reducing bikeshedding in the future about where something should go. 

Overall I'm very :+1: on this RFC, I like the groupings, and I think it also makes excellent use of our existing content and minimizes the changes that will be required for Octane. One thing that has been concerning for folks reviewing the various RFCs (Glimmer Components, Modifiers, Tracked Props, etc.) is just how much of the guides will need to change, and I think this does an excellent job of outlining and calling out those things specifically. 

The one thing I might add is a section on Template-Only Components under Components (cc @mike-north on that) but I _think_ based on the language of the RFC, the exact content for each section is not what is being proposed, so much as the overall groupings and the rules for those groupings. That means we would have plenty of flexibility to change those individual subsections as we begin the refactors, if we realize that we needed another section or a section was actually not needed.

jenweber:
@pzuraq you are right that the focus is on the "rules" and high level topics. That said, for the "Creating a component" section, it has a callout to "Types of Components" in order to cover that. I have also noted it in the Gist! Thanks!

CodingItWrong:
Tiny bit of feedback; I may have more later. I feel like other frameworks refer to "dev tools" rather than debugging, so that may be a better option for the Table of Contents item. It might not be as good for those who don't know frontend dev, though.

text/0000-guides-restructure.md
```diff
@@ -0,0 +1,397 @@
...
+
+The most common points of confusion for new devs have become things like “how does everything fit together? Why can't I call a sibling component's actions directly?” The faster we can get the “Mental Model” of Ember across, the better. The Super Rentals tutorial covers the bases well for using the CLI and adding basic interactions. Anecdotally, thanks to the Tutorial, it seems that the missing pieces in the guides are mostly architectural. "Anatomy of an app" is the main area of new content, where we attempt to give an overview of the topics a web developer has to care about, the parts of Ember that address them, and the file structure. Current Ember developers will reference this section while learning new features.
+
+#### Templates & Template helpers
+
+Templates come before components, since it is possible for someone to not know JavaScript/Handlebars and still contribute to Ember. For example, designers could effectively work in only templates, using plain HTML. To make Templates work as the opening topic, additions are necessary in the introduction.
+
+Template Syntax will cover Handlebars basics and built-in features like `{{input}}`, `{{#each}}`, and `{{#if}}`.
```
NullVoxPopuli: are we still calling them handlebars long term? 
I remember a discord discussion about this some time back -- due to divergence of all the mental model things that ember's templates fixed when compared to vanilla handlebars. My discord search-fu isn't so great. seems to not be as fuzzy as I need it to be

text/0000-guides-restructure.md
```diff
@@ -0,0 +1,397 @@
...
+
+Lastly, another reason to split out Template syntax from Template Helpers is because the syntax section is expected to expand with the addition of Octane content.
+
+An alternative to this split is to reduce the documentation of helpers themselves in the Guides, and lean on the API docs instead, however great care should be taken in removing documentation. Without a strong case for it, we lean towards leaving them in the Guides.
+
+#### Working with JavaScript
+
+A common concern of new learners is, what is JavaScript and what is special to Ember? The Working with JavaScript section incorporates the existing JavaScript Primer and new content that will help make the distinctions clearer. Ember.js has procatively adopted JavaScript APIs as they become available, such as Classes and Decorators, in some cases before the larger JavaScript community becomes familiar with them. Key people in the Ember Community are also involved in the development of JavaScript itself through TC39. To an extent, it is our responsibility to at least suggest to developers which JavaScript concepts they must learn in order to feel comfortable with Ember.
```
NullVoxPopuli: so much :+1:

text/0000-guides-restructure.md
```diff
@@ -0,0 +1,397 @@
...
+Migrating
+Migration from Vue 1.x
+Migration from Vue Router 0.7.x
+Migration from Vuex 0.6.x to 1.0
+Meta
+Comparison with Other Frameworks
+Join the Vue.js Community!
+Meet the Team
+
+One possible lesson here is that we could split up Components like Vue did with Component Basics and Components In-Depth. Their dedicated section on Computed Properties inspired the inclusion in our new Table of Contents.
```
NullVoxPopuli: I'm a fan of this, components can do _a lot_, but all that extra stuff would go in "Leveling Up", which I think is good. starting out, people don't need to manage remote data / queries in their components, but that would be a good topic for a standalone guide rather than in the intro guides. :+1:

text/0000-guides-restructure.md
```diff
@@ -0,0 +1,397 @@
...
+- Making an Interactive Component
+- Developer Tools
+
+Nowhere on our current site do we have a highly visible explanation of what Ember is, beyond snippets. In light of this glaring omission, we have added a “What is Ember” section to the Guides Table of Contents above. It is not meant to replace the ongoing “Why Ember” and marketing-focused descriptions that are underway.
+
+### Vue
+
+As a fully-featured framework, Vue is an easier comparison for possible Table of Contents listings. Keep in mind that much of this type of content is present in our CLI docs instead, so this list will look longer than what we are aiming for.
+
+Introduction
```
efx: noticed these should use the same list syntax, PR here: https://github.com/jenweber/rfcs/pull/2/files

efx:
My high level and brief feedback:
I am heartily in favor of restructuring the guides as proposed here. I think it fits well with the broader JavaScript community and thus would leverage developer's existing knowledge.

Gaurav0:
How are we going to teach both classic and glimmer components? I would like to propose all content specifically pertaining to classic components be moved to Leveling Up following the Octane release.

text/0000-guides-restructure.md
gossi: I would suggest to differentiate between control structures (loops and conditionals), block and inline styles and then helpers and components syntax (where each is linking to the dedicated section of the docs).

Second: the `{{input}}` helper is a bad example and should be treated as *classic* or *legacy*, since it is a two-way-binding that'll go away with octane/glimmer arriving, where people are held to use regular `<input>` elements with proper bindings on them. Put them into a deprecated/classic section or better throw them out (but this may be to early 😇)

text/0000-guides-restructure.md
```diff
@@ -0,0 +1,397 @@
...
+Components will need the most refactor work and new content for Octane, so here’s the subtopic breakdown:
+
+- Creating a component (includes Component Types post-Octane)
+- Displaying data in a Template
+- Adding actions
+- Nesting components (includes suggestions of when to break things into components)
+- Passing arguments
+- Using computed properties (extremely basic example, link to the Computed Property section)
+- Working with arrays of data
+- Lifecycle hooks
```
gossi: I think the outline for this section needs to be revisted again in terms of general patterns, e.g. adding actions, passing arguments or state management with computeds/tracked vs. component-type related topics. There is ember-components, sparkles/glimmer components, hooked components, template only components ... and all of them have different APIs (template only has none! :D) but lifecycle hooks are different for those types of components. The guides shall not spread the idea of ambiguity amongst them.

text/0000-guides-restructure.md
```diff
@@ -0,0 +1,397 @@
...
+
+#### State management
+
+The State Management section is important because whether "computed properties" are available in their current state or as Decorators, the nunances must be within easy reach for existing Ember developers and future learners. We rename this topic in alignment with the Topic Naming convention mentioned above.
```
gossi: + `@tracked` ?

MelSumner:
@jenweber can we add an accessibility section? I know it's not Ember-specific but it is something we should make sure every Ember developer is aware of. 

I'll take on getting the content created for it, and I think it's important for our guides to address.

jenweber:
@MelSumner This PR is mostly about the bare minimum to do Octane, but since Accessibility is such a glaring omission in the Guides, I'll include it. It is one brand new topic that I think merits being added!

text/0000-guides-restructure.md
jenweber: I will remove `{{input}}` from this RFC explanation. I am not sure if there is any consensus around whether `{{input}}` is good or bad, but it doesn't need to show up in this explanation.

I will also add a note to this RFC regarding the naming of Handlebars.

text/0000-guides-restructure.md
```diff
@@ -0,0 +1,397 @@
...
+
+**Core Concepts**
+
+- What is Ember?
+- Getting Started
+- Anatomy of an app (future new content)
+
+**Fundamentals**
+
+- Template syntax
```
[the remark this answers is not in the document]
jenweber: Template Syntax currently exists as a subsampling of `Templates.` It would need a new intro on the first topic.

I can imagine that it would include a little bit of why. Do you have any material we could draw from? A good followup here would be to create an issue in the Guides. Nothing stops us from adding a little bit of "why" in today.

text/0000-guides-restructure.md
```diff
@@ -0,0 +1,397 @@
...
+
+## Detailed design
+
+The focus of this RFC is not to write new content, but make better use of what is already there by presenting it in a different order. We aim to rearrange the table of contents without breaking any URLs, with refactors necessary to fix the transitions between topics, and with the minimal _new_ content needed to teach Octane. We take care to not make it look like there is “more to learn.”
+
+The overall learning strategy is to establish a common core of sequential knowledge, and then later topics can be read standalone, skipping around. What this means is that if beginners read through to routes, they could skip around the topics lower down the chain successfully.
+
+Finally, it is important to acknowlegde that the Guides are intended to represent the "Happy Path" of using Ember. It is not possible to have something that is a perfect fit for everyone's app needs, nor should they cover our entire API surface.
```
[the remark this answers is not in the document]
jenweber: I think there was discussion of this in the past. However in our current implementation, if I remember right, indexing guides and blog posts together has technical hurdles such that time would be better spent elsewhere.

I would rather that we look carefully at what things in the blogs really ought to be better explained in the Guides or API docs. If you can think of any examples, please open an issue in the Guides. Thanks!

jenweber:
@Gaurav0 your suggestion and one by @NullVoxPopuli point towards splitting up components. I will reflect on this some and see if I can think of a solid rule to apply here. There are some edge cases where hbs components still need to be used, I think, but since they are edge cases that you don't need to know to learn other things, it makes them potentially "Leveling Up".

It's kinda tough to figure out how to do this without rewriting `Components` entirely, so maybe we should try to come up with what a rewrite would look like, and that could possibly inform the interim steps.

Thanks for your feedback!

jenweber:
@CodingItWrong thanks for the idea. Renamed Developer Tools. We don't want to give people the idea that Ember is buggy, haha. I think even new people might get the idea, and we can link to it in useful places like the Tutorial (I think that is already happening)

mikkopaderes:
A little bit off topic since I haven't seen any info on it yet.

Is the guides getting a total refresh as layed out in the RFC a blocker for Octane release? I hope it is.. (and also #425)

jenweber:
@mikkopaderes, good question. It is not my intention for this work to be a blocker. It is meant to be incremental by design, however I do think some pieces of it will be absolutely necessary for Octane. For example, we will need to do major work on the Templates and Components introductions, which would necessitate a reordering.

My understanding of #425 is that it is also not a blocker, however that's a good question to put in that RFC.

Octane marks an inflection point where we have MVP features together. As more features are included in later releases, so too will our Guides keep progressing towards this goal.

Let me know if you have any other questions! I will add the non-blocker statement to this RFC.

text/0431-guides-restructure.md
```diff
@@ -0,0 +1,400 @@
...
+One possible lesson here is that we could split up Components like Vue did with Component Basics and Components In-Depth. Their dedicated section on Computed Properties inspired the inclusion in our new Table of Contents.
+
+### Angular
+
+Angular is also a full-featured framework that has a lot in common with Ember.
+
+FUNDAMENTALS
+Architecture
+
+  Architecture Overview
```
toddjordan: when rendered these don't show up on separate lines because there are no markdown bullet points used with them

toddjordan:
Some perspective on remove the object model section...

When I came to Ember I had a lot of years in Object Oriented programming, as well as a good amount JavaScript experience.

The Object Model of Ember was still a bit of a learning curve for me because of the following:
- I wasn't used to having to use get and set... Always just `this.myProp = myVal` or `let myVal = this.myProp`.  
- I also was trolled at first for having to use things like pushObject/removeObject for arrays
- I didn't understand the need for reOpen, and didn't realize at first that it was updating the prototype.  Also wasn't sure when you needed to do this outside of initial class declarations.
- Computeds were a new concept for me at the time.

I just wanted to give the perspective that the objects section wasn't just for people coming from places not familiar with OO or JavaScript, but that Ember did have some unique things to its Object structures that were important to know for anyone new to use it at a basic level. 

That all said, I think if we deemphasize reopen, and make sure we mention some of the special object apis in a state management section, then think we'll be covered.  I can't remember off the top of my head though if we don't call set, will a rerender not happen or assert be thrown?  If so then we may want to consider moving discussion of that earlier, like in the initial component section.

text/0000-guides-restructure.md
gossi: Technically it would make sense to start with embers component manager and explain there can be different types of components ... and this will be the point where the guides will state ember as suuuper difficult. Not a good way to get learners onboarded but technically not wrong :D

Instead, what may be more helpful to pick a little example that states:

- how to create a component
- how attributes and arguments work (and how to handle defaults for `@arguments` - this WILL come up)
- how to bring properties from component to template (in combination with state managemet)
- how modifiers interact with components at the example of action
- ..... where we'll come to lifecycle hooks stating different types of components

From your list "Displaying data in a Template" and "Working with arrays of data" seems like repetitive (only given the headlines). Nesting components is an advanced topic, which would better interact with given patterns (not only for ember) that exist for components and how to apply them with ember, which opens another section for the guides (what I just send you over at discord).

The list I outlined might still be technically driven but I'm quite happy with my grouping now make it approachable to beginners =)

lupestro:
Can the "Developer Tools" section truly be broadened into "Debugging". It occurs to me that there are a handful of areas where debugging an application requires following change into the "black box" of the Ember code. Often, you can trap at specific points to determine if the "signal is lost", with only a "theory of operation" explanation at your back, without having to become intimate with the details of the code. Where this is possible, it would be really helpful to developers who dread finding themselves thrust into the steep learning curve of things that are only peripheral context to the problem they are really trying to solve, when they can least afford to delve there. 

Based on my experience of the past eighteen months on my latest project, this would be especially helpful when things go wrong with deeply nested module loading, computed properties, async and runloop behavior, and the interaction of Ember and DOM events. A section on chasing sources of memory leaks would also be helpful.

jenweber:
I have made some small changes to this RFC thanks to feedback by @mike-north!

1. Reunite Template syntax and helpers. Custom helpers are much more powerful than I realized for showing off the capabilities of templates and unit tests, in a way that is welcoming to people coming in from other frameworks and total beginners
2. Add a section on who our audience is

Then from the Learning team:
1. Upgrading explicitly includes editions conversion guides.

jenweber:
@lupestro thanks for the input! Initially I had called it Debugging, but based on feedback from others, changed it to Developer Tools because it was suggested as a broader term. 

I think I would like the Ember Inspector team to weigh in on what to name the section that would help people discover Inspector. It's the most powerful tool we have. @rwwagner90 or others, any thoughts?

jenweber:
The Learning Core team has decided to move this to FCP, "Final Comment Period." To learn more about what FCP means, see the RFC [README](https://github.com/emberjs/rfcs). In short, more comments are welcome but the days are numbered.

Thanks to everyone has participated so far, both in this PR and chat!

rwwagner90:
@jenweber React has a `Tools` section, and a subsection called `Debugging` under it. https://reactjs.org/community/debugging-tools.html

Vue calls it `Tooling` with a `Devtools` section under that. I think the common theme is have a section for tools, then have something related to Debugging/DevTools under that.

text/0431-guides-restructure.md
jenweber: thank you!

text/0431-guides-restructure.md
```diff
@@ -0,0 +1,414 @@
...
+
+**Core Concepts**
+
+- What is Ember?
+- Getting Started
+- Anatomy of an app (future new content)
+
+**Fundamentals**
+
+- Templates
```
chilicoder: @jenweber please think about this order in Fundamentals

* Components
My guess is that most of people who will read the guides will be familiar with React and I suggest to start with this section cause it would be most familiar section for them. 
* Working with JavaScript / Templates 
I don't have strong opinion what should go first, it should depend on how Components will be written and what will be the bridge. One thing I really would appreciate to see in Template section is that they automatically resolve promise-values. When I wrote a React app some time ago I was stunned not to have it there.

* State management (aka computed properties)
Here maybe we can mention some similarities with Vue.js. In case the guides will be read by those who tried Vue.

* Routing (includes Routes, Router, and Controllers)
Here would be nice to explain semilarities with React router (since it was a port from the Ember one). And in the end we can make a bridge to services (since Router is a service)
* Services

text/0431-guides-restructure.md
```diff
@@ -0,0 +1,414 @@
...
+
+### Table of Contents
+
+The following Table of Contents will be applied iteratively over the course of many months. It is not considered a blocker for the release of the Octane edition. The quick wins and urgent Octane refactors will be applied ASAP, and are described in the Implementation section of this RFC. Justification for the ordering is provided following the topics list.
+
+**Core Concepts**
+
+- What is Ember?
+- Getting Started
+- Anatomy of an app (future new content)
```
chilicoder: Very valuable point. And just in the right place. 

I would like to see here detailed explanation about ember's folder structure and how this convention enables reusability of addons.

chilicoder: Also, since we discuss here folder structure, would be nice to mention here CLI which takes care about boilercode, naming and folders.

text/0431-guides-restructure.md
mike-north: I love that you're thinking about optimizations that could be made around visitors who have context from other technologies, but there are a lot of different places people could be coming from. 

For example, an _early_ version of React Router [credited Ember's router as a source of inspiration](https://github.com/ReactTraining/react-router/blame/1.0.x/README.md#L134), but the current version is very different. Even for just this one library, which versions(s) should we assume?

Kicking off right into components is an interesting idea! The main difference between React's component model and Ember/Vue/Angular is the use of declarative templates instead of a `Component#render` function -- wouldn't this suggest that templates should be introduced first?

These points you're focusing on are valuable, but really tough to get right in a "throughout the guides" kind of way. Maybe we could channel these ideas into a new section like "...So you're coming from React..." that aims to establish a bridge between a reader's understanding in one world to the ember world

text/0431-guides-restructure.md
MelSumner: @chilicoder we could probably add a callout that links to the CLI guide. We purposefully designed this component for just this sort of use case. 👍

MelSumner:
@jenweber - 

> We propose that the Guides should focus on the following audience - newcomers who have done the tutorial, and want to build real apps. It will show the "happy path" and how the parts of Ember interact to form good, useful patterns, in a way that would be out of scope for an API documentation block.

Since this is the audience that the guides already focus on, should we phrase this to reflect that? Maybe "We propose that the Guides should continue to focus on the following audience"?

Maybe not necessary, but food for thought. 

Excellent work on this RFC!

text/0431-guides-restructure.md
locks:
```suggestion
- Templating
```

My justification: Just like "routing" is phrased as the concern and not the concrete implementation (route), and is meant to include routes, the router, and controllers, so should "templating".
The idea is to give developers a good mental model of the templating language and how to rationalize about its behaviour.

text/0431-guides-restructure.md
```diff
@@ -0,0 +1,414 @@
...
+
+## Summary
+
+As our favorite framework has grown and changed a lot over the past few years, so have our [Ember.js Guides](https://guides.emberjs.com)! This project aims to use the excellent work that has been done by hundreds of contributors and arrange it in a way that provides a natural learning flow for today’s Ember.js developers.
+
+## Motivation
+
+With Octane coming down the line, and an influx of fresh attention, it’s more important than ever to consider Ember's overall learning experience. It's time for a shift. The current structure and flow of the guides reflects the past, not the present experience for Ember. It’s time to fix that.
+
+When the Guides were originally written, things like object-oriented programming in JavaScript, components, routing, and SPAs in general were new ideas. For example, they start by teaching the “Ember Objects” programming concept, necessary background information at the time. It would have been an unfamiliar pattern at the time. Now, the learning hierarchy needs have shifted from “learn these unique web development concepts” to “learn how familiar pieces fit together so you can build great apps quickly.”
```
locks:
```suggestion
When the Guides were originally written, things like object-oriented programming in JavaScript, components, routing, and Single-Page Applications ([SPA](https://en.wikipedia.org/wiki/Single-page_application)) in general were new ideas. For example, they start by teaching the “Ember Objects” programming concept, necessary background information at the time. It would have been an unfamiliar pattern at the time. Now, the learning hierarchy needs have shifted from “learn these unique web development concepts” to “learn how familiar pieces fit together so you can build great apps quickly.”
```

text/0431-guides-restructure.md
```diff
@@ -0,0 +1,414 @@
...
+
+Lastly, another reason to split out Template syntax from Template Helpers is because the syntax section is expected to expand with the addition of Octane content.
+
+An alternative to this split is to reduce the documentation of helpers themselves in the Guides, and lean on the API docs instead, however great care should be taken in removing documentation. Without a strong case for it, we lean towards leaving them in the Guides.
+
+#### Working with JavaScript
+
+A common concern of new learners is, what is JavaScript and what is special to Ember? The Working with JavaScript section incorporates the existing JavaScript Primer and new content that will help make the distinctions clearer. Ember.js has procatively adopted JavaScript APIs as they become available, such as Classes and Decorators, in some cases before the larger JavaScript community becomes familiar with them. Key people in the Ember Community are also involved in the development of JavaScript itself through TC39. To an extent, it is our responsibility to at least suggest to developers which JavaScript concepts they must learn in order to feel comfortable with Ember.
```
locks:
```suggestion
A common concern of new learners is, what is JavaScript and what is special to Ember? The Working with JavaScript section incorporates the existing JavaScript Primer and new content that will help make the distinctions clearer. Ember.js has proactively adopted JavaScript APIs as they become available, such as Classes and Decorators, in some cases before the larger JavaScript community becomes familiar with them. Key people in the Ember Community are also involved in the development of JavaScript itself through TC39. To an extent, it is our responsibility to at least suggest to developers which JavaScript concepts they must learn in order to feel comfortable with Ember.
```

text/0431-guides-restructure.md
mike-north: @locks do you feel that "Templating" is more easily identifiable and relatable to most of our audience than "Templates"?

locks: That's a good question! I don't think it would impact identifiability and discoverability and might aid nudge into the mental model we want, "templating", "routing", "working with javascript".
Out of curiosity I checked other projects, Vuejs and Angular seem to go with "Template syntax", Aurelia and Polymer with "Templating" and "DOM templating".

text/0431-guides-restructure.md
```diff
@@ -0,0 +1,414 @@
...
+- Adding actions
+- Nesting components (includes suggestions of when to break things into components)
+- Passing arguments
+- Using computed properties (extremely basic example, link to the Computed Property section)
+- Working with arrays of data
+- Lifecycle hooks
+
+#### State management
+
+The State Management section is important because whether "computed properties" are available in their current state or as `@tracked` Decorators, the nunances must be within easy reach for existing Ember developers and future learners. We rename this topic in alignment with the Topic Naming convention mentioned above.
```
locks:
```suggestion
The State Management section is important because whether "computed properties" are available in their current state or as `@tracked` Decorators, the nuances must be within easy reach for existing Ember developers and future learners. We rename this topic in alignment with the Topic Naming convention mentioned above.
```

jenweber:
Hello everyone! Thank you so much for all your thoughtful comments and suggestions. The Learning Team has reached consensus to merge this RFC.  I'm excited that we have this opportunity to improve the learning story, alongside the improvements that come with Octane.

Considering this is just the "MVP" plan, the future is looking good :) We welcome your continued feedback and ideas along the way.

PR #433, opened 2019-01-18
Accessible Routing 

MelSumner:
[Rendered](https://github.com/MelSumner/rfcs/blob/MelSumner-a11y-routing/text/0000-a11y-routing.md)

Updated on Jan 12, 2020 to reflect conversations during framework f2f weekend.

### DIFF
--- a/text/0000-a11y-routing.md
+++ b/text/0000-a11y-routing.md
@@ -1,4 +1,4 @@
-- Start Date: (2019-01-18, updated on 2019-05-13)
+- Start Date: (2019-01-18, updated on 2020-01-12)
 - RFC PR: (leave this empty)
 - Ember Issue: (leave this empty)
 
@@ -24,17 +24,13 @@ One of the challenges is that screen reader technology is all closed-source exce
 
 This RFC proposes that we take the first step toward accessible routing by adding an announcement message when the URL has changed. While this will not address the second part of accessible routing, focus, it will move the framework closer to the goal. 
 
-- A component is created that provides a screen-reader only message, informing the user that they have navigated to a new URL. There will be a default message but developers can also set a custom message (those with internationalized apps may likely prefer to do).
+We will implement a live region in the page inside of the body element. The message will be updated when the user interacts with the page in such a way that a new URL is needed (i.e., navigating to a new page, or interacting with an element such as a button that submits a form and moves the user to a confirmation page or similar). We will listen for `routeDidChange` and present the navigation event message to the user at that time. This message will inform the user that they have navigated to a new URL. There will be a default message but developers will also be able to create a custom message (those with internationalized apps will likely prefer to do so).
 
-We would implement a live region in the page inside of the body element. The message will be updated when the user interacts with the page in such a way that a new URL is needed (i.e., navigating to a new page, or interacting with an element such as a button that submits a form and moves the user to a confirmation page or similar). We will listen for `routeDidChange` and present the navigation event message to the user at that time. 
-
-The addon would attempt to provide a sensible resolution for all involved:
-- the performance gains from `pushState` remain in place
-- users with assistive technology are informed that a page transition has occurred
+The addon would attempt to provide a sensible resolution for all involved, as the performance gains from `pushState` remain in place but users with assistive technology will also be informed that a page transition has occurred.
 
 Some important details: 
 - This would be added to the default Ember blueprint and included with all new Ember apps. 
-- The developer would need to opt out of this option entirely to turn it off (but it will not be recommended to do so).
+- The developer would need to opt out of this entirely to turn it off (but it will not be recommended to do so).
 - This will not address focus being reset in the way that users with assistive technology expect. This still needs to be addressed and we will continue to work toward a reasonable solution for this in the future. 
 
 ## How we teach this
--- a/text/0000-a11y-routing.md
+++ b/text/0000-a11y-routing.md
@@ -36,7 +36,7 @@ One potential solution is that we don't overwhelm the screen reader user with th
 The component template: 
 
 ```hbs
-<div tabindex="-1" class="ember-sr-only" id="nav-message">
+<div tabindex="-1" class="ember-sr-only" id="ember-a11y-refocus-nav-message">
   Page transition is complete. You may now navigate as you wish.
 </div>
 ```
--- a/text/0000-a11y-routing.md
+++ b/text/0000-a11y-routing.md
@@ -1,4 +1,4 @@
-- Start Date: (2019-01-18, updated on 2019-05-13)
+- Start Date: (2019-01-18)
 - RFC PR: (leave this empty)
 - Ember Issue: (leave this empty)
 
--- a/text/0000-a11y-routing.md
+++ b/text/0000-a11y-routing.md
@@ -1,0 +1,91 @@
+- Start Date: (fill me in with today's date, YYYY-MM-DD)
+- RFC PR: (leave this empty)
+- Ember Issue: (leave this empty)
+
+# Accessible Routing in Ember
+
+## Summary
+
+When the user navigates to a new route within an Ember application, screen readers do not read out the new content or appropriately move focus. This makes Ember applications not natively accessible for users with assistive technology.
+
+Ember needs to provide a machine-readable way for assistive technology (like screen readers) to know that the route transition has occurred.
+
+## Motivation
+
+In a world where accessible websites are increasingly the norm, it becomes good business to ensure that Ember applications are able to reach basic accessibility standards. The first initiative in this area is to ensure that Ember can provide the same level of accessibility as a static website- and the first task is to tackle basic navigation within a web app.
+
+In a static website, checking `document.activeElement` in the console upon navigation to a new page should return `[object HTMLBodyElement]`. However, in an Ember app this is not the case. For example, if you navigate to one of the pages via the navbar, the focus remains on the element in the navbar. This is a problem for accessibility; the screen reader doesn’t know that a route transition has occurred, and the screen reader user doesn’t know that the link to the new page actually worked (they hear nothing). 
+
+So we have a puzzle to figure out: why don't screen readers read out the new content or appropriately move focus when the user navigates to a new route within an Ember application. In researching the issue more thoroughly, we discovered that NVDA doesn't seem to have a way to handle the history API. (Which makes sense- screen readers were released before the history API was released.)
+
+One of the challenges is that screen reader technology is all closed-source except for [NVDA](https://www.nvaccess.org/), the open source screen reader for Windows (and typically used with Firefox). The source code for NVDA is available on [GitHub](https://github.com/nvaccess/nvda/) and is written in Python. It seems useful, at least initially, to try to solve for SPAs + NVDA, since both can be done in open source. (Related: [NVDA Issue #6606](https://github.com/nvaccess/nvda/issues/6606)) 
+
+## Detailed design
+
+After reviewing possible approaches to this problem, it seems like the very first step should be an implementation that, at the very _least_, returns Ember back to the starting line. We could then build on that by providing a more machine-intelligent iteration for an upscaled experience. This iterative approach would give us a better way to tackle the more difficult edge-case scenarios.
+
+This RFC proposes that we could either implement a two-phase approach (outlined below) OR decide to move directly to the approach outlined in phase two. 
+
+### Phase One - Return to the starting line
+
+This solution proposes the simplest possible solution- implement in Ember what already exists in other traditional websites:
+1. When a browser loads a website, the screen reader starts to read the content of that page
+2. When the user navigates to a new page in that website, the screen reader starts to read the content of that page
+3. (repeat)
+
+Perhaps we could achieve this by adding a function to set focus on the body element when a route transition has occurred. 
+
+In order to not break similar solutions that may already exists in other apps or addons, I propose that the solution be implemented as an optional feature/with a flag. The default could be set to on, and if folks already have a solution in their addons/apps, they could then opt-out of this feature. 
+
+### Phase Two - Intelligent content focus
+
+It would be even better to have a more intelligent solution, since there are several ways new page content could be introduced in an Ember application.
+
+[Ember-self-focused](https://github.com/linkedin/self-focused/tree/master/packages/ember-self-focused) and [ember-a11y](https://github.com/ember-a11y/ember-a11y) are both examples of community-produced addons that attempt to provide a contextual focus, based on new page content. 
+
+From the [Ember-self-focused](https://github.com/linkedin/self-focused/tree/master/packages/ember-self-focused) readme: 
+> When UI transitions happen in a SPA (or in any dynamic web application), there is visual feedback; however, for users of screen reading software, there is no spoken feedback by default. Traditionally, screen reading software automatically reads out the content of a web page during page load or page refresh. In single page applications, the UI transitions happen by rewriting the current page, rather than loading entirely new pages from a server; this makes it difficult for screen reading software users to be aware of UI changes (e.g., clicking on the navigation bar to load a new route).
+
+> If the corresponding HTML node of the dynamic content can be focused programmatically, screen reading software will start speaking the textual content of that node. Focusing the corresponding HTML node of the dynamic content can be considered guided focus management. Not only will it facilitate the announcement of the textual content of the focused HTML node, but it will also serve as a guided context switch. Any subsequent “tab” key press will focus the next focusable element within/after this context. 
+
+The tedious, error-prone task of keeping track of HTML nodes that can/could/would/should have focus is something that Ember could do for you. 
+
+A comparable from another JS Ecosystem: [Reach Router](https://reach.tech/router) is available for React apps. 
+
+## How we teach this
+
+For phase one, the guides should include information about skip links, since the application author will need to add those in themselves (just as they would for a static site). For phase two, we would explain that skip links are no longer needed because we have integrated that functionality in a more machine intelligent way. 
+
+### Skip Links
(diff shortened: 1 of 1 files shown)


text/0000-a11y-routing.md
```diff
@@ -0,0 +1,91 @@
...
+This allows the user with assistive technology to skip directly to the main content and is useful for not needing to repeatedly navigate through a navbar, for example.  
+
+
+It's possible that this solution is a [spherical cow](https://en.wikipedia.org/wiki/Spherical_cow), and the complexity of route transitions in a framework deserves an accessibility solution that is is equally complex. 
+
+It's possible that enterprise Ember users have already implemented their own solution, since governments in many countries require this by law. An out of the box solution would need to be well-advertised and documented to avoid any potential conflicts for users. 
+
+## Alternatives
+1. We could try to implement something like Apple’s first responder pattern: https://developer.apple.com/documentation/uikit/uiresponder/1621113-becomefirstresponder 
+2. Wait for a SPA solution to be implemented in assistive technology. I would note that this option seems less than ideal, since this specific issue has been reported to the different assistive technologies available, and has been an on-going issue for many years now.
```
Robdel12: This, and I bet AT vendors will try and push the burden onto the browsers since AT operates at the OS level.

text/0000-a11y-routing.md
```diff
@@ -0,0 +1,91 @@
...
+
+> If the corresponding HTML node of the dynamic content can be focused programmatically, screen reading software will start speaking the textual content of that node. Focusing the corresponding HTML node of the dynamic content can be considered guided focus management. Not only will it facilitate the announcement of the textual content of the focused HTML node, but it will also serve as a guided context switch. Any subsequent “tab” key press will focus the next focusable element within/after this context. 
+
+The tedious, error-prone task of keeping track of HTML nodes that can/could/would/should have focus is something that Ember could do for you. 
+
+A comparable from another JS Ecosystem: [Reach Router](https://reach.tech/router) is available for React apps. 
+
+## How we teach this
+
+For phase one, the guides should include information about skip links, since the application author will need to add those in themselves (just as they would for a static site). For phase two, we would explain that skip links are no longer needed because we have integrated that functionality in a more machine intelligent way.
```
Robdel12: > For phase two, we would explain that skip links are no longer needed because we have integrated that functionality in a more machine intelligent way. 

Is this true? I usually still keep nav skip links in my SPAs because it's helpful to move past the giant headers / nav bars web apps have.

text/0000-a11y-routing.md
```diff
@@ -0,0 +1,91 @@
...
+
+So we have a puzzle to figure out: why don't screen readers read out the new content or appropriately move focus when the user navigates to a new route within an Ember application. In researching the issue more thoroughly, we discovered that NVDA doesn't seem to have a way to handle the history API. (Which makes sense- screen readers were released before the history API was released.)
+
+One of the challenges is that screen reader technology is all closed-source except for [NVDA](https://www.nvaccess.org/), the open source screen reader for Windows (and typically used with Firefox). The source code for NVDA is available on [GitHub](https://github.com/nvaccess/nvda/) and is written in Python. It seems useful, at least initially, to try to solve for SPAs + NVDA, since both can be done in open source. (Related: [NVDA Issue #6606](https://github.com/nvaccess/nvda/issues/6606)) 
+
```
Robdel12: I've long had a dream to take NVDA (since it's OSS) and turn make a "headless" screen reader version of it. _Just_ like we have a headless Chrome we push around in tests, I'd like a headless AT to send commands to and asset the right thing has happened. 

It's a neat idea, but I have no idea if its even possible.

text/0000-a11y-routing.md
MelSumner: @Robdel12 the idea for phase 2 is that we'd make it so AT would automatically implement skip link functionality so it was a cohesive experience- I think of it as an upgrade to skip links.

sandstrom:
Perhaps this is an unpopular point of view, but I think there are other areas that should be prioritized over accessibility. Here are three examples:

1. One of the many things that the community is already [eagerly waiting for](https://www.emberjs.com/statusboard/).

2. Better localization support. There are ~6.5 billion non-native english speakers, and better support for handling multiple languages, currency formatting, etc. would be better in terms of "accessibility" (millions of additional people that would "get access").

3. Things that makes it easier to run Ember on low-cpu devices, such as phone (i.e. performance improvements). Again, billions of people don't have desktop computers with Intel chips. Would "enable access" for a lot more people than improving screenreader support will ever do.

*Obviously one thing doesn't prohibit another, one can work on both screen reader support and localization functionality at the same time. But in a world of limited resources, I'd still like to point out that I don't think this should be prioritized.*

text/0000-a11y-routing.md
```diff
@@ -0,0 +1,91 @@
...
+## Appendix
+
+### Related RFCs
+- [Route Helpers](https://emberjs.github.io/rfcs/0391-router-helpers.html) by Chad Hietala
+
+### Related Reading
+
+- [NVDA Developer Guide](https://www.nvaccess.org/files/nvda/documentation/developerGuide.html)
+- [Microsoft Active Accessibility (MSAA)](https://docs.microsoft.com/en-us/windows/desktop/winauto/microsoft-active-accessibility) MSAA is an Application Programming Interface (API) for user interface accessibility. MSAA was introduced as a platform add-on to Microsoft Windows 95 in 1997. MSAA is designed to help Assistive Technology (AT) products interact with standard and custom user interface (UI) elements of an application (or the operating system), as well as to access, identify, and manipulate an application's UI elements. The current and latest specification of MSAA is found in part of [Microsoft UI Automation](https://docs.microsoft.com/en-us/windows/desktop/WinAuto/uiauto-specandcommunitypromise) Community Promise Specification.
+- [Structured Negtotiation: A Winning Alternative to Lawsuits](https://www.lflegal.com/book/)
```
pangratz:
```suggestion
- [Structured Negotiation: A Winning Alternative to Lawsuits](https://www.lflegal.com/book/)
```

MelSumner:
> This is probably an unpopular point of view, but I think there are other areas that should be prioritized over accessibility.

Thankfully, it's not a zero sum game! Since my primary area of expertise is accessibility, this is where I can propose specific improvements to Ember. 

If your area of expertise is in any of the things you suggested, please do feel free to write RFCs to address those issues- there's room here for everyone! :+1:

Robdel12:
@sandstrom I have to wholly disagree with that statement based on one thing: every single new ember app out there right now is broken to AT. There's **nothing** the user can do to try and work around or fix it. I know it's not ideal, but if you need to translate something to a different language, you have the ability to. AT users are simply stuck high and dry. If they're an _advanced_ AT user they might be able to figure out what's going on in a SPA, but probably won't bother. 

I think **every** SPA framework / lib needs to take accessibility seriously & prioritize the work now that everyone is building SPAs. Until those framework / lib authors take it seriously, [all SPAs will be broken to AT.](https://medium.com/@robdel12/single-page-apps-routers-are-broken-255daa310cf) 

And, as @MelSumner said, it's not zero sum. You can get the ball rolling on first class i18n support if you'd like. Maybe learn some lessons from Angular when they took on that effort. 

[Lastly, this work might save everyone's companies from being sued.](https://www.lflegal.com/2019/01/dominos-ninth-circuit/) for not being accessible.

text/0000-a11y-routing.md
drewlee: +1

text/0000-a11y-routing.md
drewlee: Skip links are also readily utilized by keyboard-only users, so it would still be a concern outside of ATs. Additionally, as an author, I would expect a means of disabling the generated skip link in case where we want to implement something much more custom and robust.

nathanhammond:
This is roughly a duplicate of #66. Having already gone through the process I believe that Ember should be a thin core with extension points for the functionality that we need for accessibility which would then be provided by an addon. If (and _only_ if) we can't implement things in user space we should figure out how to enable those things by way of changing core. Those changes should still be minimal and not focused on accessibility so much as exposing the internal Ember.js state required to accomplish the goal.

Benefits to being external:
- Enables backwards compatibility. If you bundle something into Ember.js itself the user cannot get that functionality prior to that version of Ember. If distributed as an addon it can easily work all the way back to Ember 1.10, like `ember-a11y`.
- Enables configuration by way of interactions between multiple addons and their build steps and ordering. This allows integration and coordination between multiple addons who may care about using a single extension point.
- Enables shipping independently from Ember. This _alone_ is a big enough reason for me to encourage never including accessibility-related functionality in core. We used jQuery in the early days of Ember to wallpaper over browser differences. AT is a massive target and not something consistent enough to address once and walk away from. For example: https://github.com/ember-a11y/ember-a11y/issues/72

My proposed path forward:
- Implement the set of things that you would like to have default-enabled in Ember in one or more addons.
- Once those addons are considered "done" open an RFC to have those be included in the default Ember blueprint.
- Document their use in the Ember Guides.

This will be far less painful for maintenance over a longer time scale, allow us to better serve AT users, and enable more people to contribute to the success of the effort.

As a person who is incredibly familiar with this space (and the author of one of the referenced examples) I am _strongly_ opposed to landing this RFC.

text/0000-a11y-routing.md
```diff
@@ -0,0 +1,91 @@
...
+
+So we have a puzzle to figure out: why don't screen readers read out the new content or appropriately move focus when the user navigates to a new route within an Ember application. In researching the issue more thoroughly, we discovered that NVDA doesn't seem to have a way to handle the history API. (Which makes sense- screen readers were released before the history API was released.)
+
+One of the challenges is that screen reader technology is all closed-source except for [NVDA](https://www.nvaccess.org/), the open source screen reader for Windows (and typically used with Firefox). The source code for NVDA is available on [GitHub](https://github.com/nvaccess/nvda/) and is written in Python. It seems useful, at least initially, to try to solve for SPAs + NVDA, since both can be done in open source. (Related: [NVDA Issue #6606](https://github.com/nvaccess/nvda/issues/6606)) 
+
+## Detailed design
```
nathanhammond: This design should be explicit. As it stands this is not a concrete proposal; it cannot be implemented using the RFC as a guide.

text/0000-a11y-routing.md
```diff
@@ -0,0 +1,91 @@
...
+This RFC proposes that we could either implement a two-phase approach (outlined below) OR decide to move directly to the approach outlined in phase two. 
+
+### Phase One - Return to the starting line
+
+This solution proposes the simplest possible solution- implement in Ember what already exists in other traditional websites:
+1. When a browser loads a website, the screen reader starts to read the content of that page
+2. When the user navigates to a new page in that website, the screen reader starts to read the content of that page
+3. (repeat)
+
+Perhaps we could achieve this by adding a function to set focus on the body element when a route transition has occurred.
```
nathanhammond: This doesn't account for modals as well as non-`body`-hosted applications.

text/0000-a11y-routing.md
```diff
@@ -0,0 +1,91 @@
...
+
+This allows the user with assistive technology to skip directly to the main content and is useful for not needing to repeatedly navigate through a navbar, for example.  
+
+
+It's possible that this solution is a [spherical cow](https://en.wikipedia.org/wiki/Spherical_cow), and the complexity of route transitions in a framework deserves an accessibility solution that is is equally complex. 
+
+It's possible that enterprise Ember users have already implemented their own solution, since governments in many countries require this by law. An out of the box solution would need to be well-advertised and documented to avoid any potential conflicts for users. 
+
+## Alternatives
+1. We could try to implement something like Apple’s first responder pattern: https://developer.apple.com/documentation/uikit/uiresponder/1621113-becomefirstresponder
```
nathanhammond: It is possible to implement this, but it's a really wonky experience on the Web. To do it right you would likely need `inert`.

text/0000-a11y-routing.md
```diff
@@ -0,0 +1,91 @@
...
+It's possible that this solution is a [spherical cow](https://en.wikipedia.org/wiki/Spherical_cow), and the complexity of route transitions in a framework deserves an accessibility solution that is is equally complex. 
+
+It's possible that enterprise Ember users have already implemented their own solution, since governments in many countries require this by law. An out of the box solution would need to be well-advertised and documented to avoid any potential conflicts for users. 
+
+## Alternatives
+1. We could try to implement something like Apple’s first responder pattern: https://developer.apple.com/documentation/uikit/uiresponder/1621113-becomefirstresponder 
+2. Wait for a SPA solution to be implemented in assistive technology. I would note that this option seems less than ideal, since this specific issue has been reported to the different assistive technologies available, and has been an on-going issue for many years now. 
+
+## Unresolved Questions
+- how would we handle loading states?
```
nathanhammond: `ember-a11y` works with loading states out of the box with exactly one caveat: [when the loading state is *below* the pivot route](https://github.com/ember-a11y/ember-a11y/issues/61#issuecomment-451288140).

text/0000-a11y-routing.md
```diff
@@ -0,0 +1,91 @@
...
+- how does/could/would this fit in with the router helpers work that Chad did?
```
nathanhammond: This is likely just an implementation detail and doesn't really need to be addressed in this RFC. This would clean up maybe a dozen lines of code in `ember-a11y` except that I want for `ember-a11y` to remain backwards compatible, so it won't adopt those until the old APIs are removed.

Given that, ideally any solution would _not_ leverage those helpers unless it had to.

MelSumner:
@nathanhammond I believe I have attempted to address your concerns with the updates I have made to this RFC.

chadhietala:
@MelSumner thanks for taking the time doing the investigation and providing paths forward. I think I would prefer that we do something with `{{outlet}}`s, either making an `{{application-outlet}}` or just making outlet smarter.

What I mean by making outlets smarter is that [RFC#93](https://github.com/emberjs/rfcs/blob/master/text/0095-router-service.md) talked about exposing the `{{get-route-info}}` which is the data structure that knows what the [logical hierarchy of the routes](https://api.emberjs.com/ember/3.10/classes/RouteInfo/) is. We would also like for this to be the data structure that is used for setting outlets as we have something that is similar today for setting the outlets. Another aspect this system is that we compute a `from` and `to` version of the RouteInfos, so technically we know the pivot point of the UI that we would need to focus.

I'm a bit cautious about exposing ember specific IDs that we use to bring focus to elements. The reason is that those IDs effectively become public API, but public API with no transition path if we decided to change the name or remove it. We have seen this with `.ember-transitioning-in`, `.ember-view` etc. Instead I think we could use a weakmap to look up the element that needs `focus()`.

Overall though I think this is heading in the right direction.

text/0000-a11y-routing.md
fuzzbomb: Indeed, AT is a red herring as far as skip-links are concerned.  The main group who _rely_ on skip links are keyboard users _without_ AT.  Screen reader users have a variety of alternative navigation mechanisms available.

Skip-links will still be necessary, until browsers provide at least one built-in mechanism for reaching the main content (e.g. heading navigation, landmark navigation, or spatial navigation).

text/0000-a11y-routing.md
```diff
@@ -22,66 +22,134 @@ One of the challenges is that screen reader technology is all closed-source exce
...
+  {{outlet}}
+</main>
+```
 
-    <a href="#maincontent">Skip to main content</a> 
+2. change the application.hbs outlet so it is different from the usual `{{outlet}}`, and have it already include what we need:
 
-This allows the user with assistive technology to skip directly to the main content and is useful for not needing to repeatedly navigate through a navbar, for example.  
+```hbs
+{{application-outlet}}
```
nathanhammond: I much prefer having a single `outlet` type. @ef4 and I have created a few over time:
- `liquid-outlet` ... from `liquid-fire`.
- `focusing-outlet` ... from `ember-a11y` which lifts the outlet stuff from `liquid-fire`.
- `animated-outlet` ... in Apple Music which is a few dozen lines of code for hardcoded CSS fade transitions.

The problem with this strategy is that none of those solutions by default play nicely together: they're fighting over the same extension point. To wire them together you have to make one dependent upon another and wrap things. You end up (approximately) with:

```hbs
{{! custom-outlet.hbs }}
{{#animated-outlet}}
  {{#focusing-outlet}}
    {{outlet}}
  {{/focusing-outlet}}
{{/animated-outlet}}
```

I feel like the API for the `outlet` itself should be extensible instead of needing to wrap it. I'm thinking something like element modifiers. It's a little hard to figure out what it should do since and `outlet` doesn't exactly have the same lifecycle hooks that a component has.

text/0000-a11y-routing.md
```diff
@@ -22,66 +22,134 @@ One of the challenges is that screen reader technology is all closed-source exce
...
+```
 
+In this way, if an application author did not want to use the `{{application-outlet}}`, for whatever reason (maybe they are nesting apps, or maybe they already have a focus solution) they could remove the `application-` and be left with a classic outlet.
```
nathanhammond: I feel like the behavior in this scenario should be opt out, not opt in. Start with `outlet` that does the accessible thing, opting out requires changing things in the application. Presumably the person who knows how to opt out will also be more familiar with what they're doing.

wagenet:
This still seems very important. I'm curious whether the routing spikes around Polaris would address this.

MelSumner:
👋 
The Ember Framework Core team recognizes accessible routing as an integral part of the future router work that is now being designed and planned. We are also excited about the [Navigation API](https://wicg.github.io/navigation-api/) work in progress that will lend itself to this work. 

As such, I'm going to close this RFC. 

For the time being, I encourage developers to check out [ember-a11y-refocus](https://github.com/ember-a11y/ember-a11y-refocus), an addon that provides accessible routing for Ember apps.

PR #442, opened 2019-02-07
Add `descriptor()`

pzuraq:
[Rendered](https://github.com/emberjs/rfcs/blob/11adcbb2cf35fd52cbdf8f3d5639521154e94967/text/0000-tracked-property-classic-class-descriptor.md)

### DIFF
--- a/text/0000-tracked-property-classic-class-descriptor.md
+++ b/text/0000-tracked-property-classic-class-descriptor.md
@@ -1,0 +1,107 @@
+- Start Date: 2018-02-07
+- Relevant Team(s): Ember.js, Learning
+- RFC PR:
+- Tracking: (leave this empty)
+
+# Tracked Property Classic Class Descriptor
+
+## Summary
+
+Add a `descriptor` decorator for classic classes only which allows users to
+define native getters and setters (and other properties):
+
+```js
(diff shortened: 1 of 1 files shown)


text/0000-tracked-property-classic-class-descriptor.md
```diff
@@ -0,0 +1,107 @@
...
+
+`descriptor` should be taught during the classic class introduction in the
+guides, and should primarily be taught as a way to add _getters_ and _setters_
+to classic classes. Other uses are covered in other ways, so they shouldn't be
+highlighted.
+
+All examples including tracked properties in the guides which have a classic
+class equivalent will use `descriptor` instead of native getters, demonstrating
+how it works by example.
+
```
MelSumner: This is already in the guides, yes?

pzuraq: It is in the Octane branch, yes. We can remove it if this RFC is not accepted though.

pzuraq:
We were able to make actual native getters work in `EmberObject.extend`:

```js
EmberObject.extend({
  get foo() {}
});
```

So this feature isn't necessary anymore.

Issue #316, opened 2018-03-23
Deprecate {{with}} helper


locks:
With the introduction of the [block `{{let}}` helper](https://github.com/emberjs/rfcs/blob/master/text/0286-block-let-template-helper.md), it is now possible to replace usages of `{{with}}`.
The reason to deprecate `{{with}}` is that the conditional behaviour of the block not being rendered when the first parameter to `{{with}}` is falsey is confusing to users.

To give an example, before:

```
{{#with parent.child as |child|}}
  {{child}}
{{else}}
  No children!
{{/let}}
```

After:

```
{{#let parent.child as |child|}}
  {{#if child}}
    {{child}}
  {{else}}
    No children!
  {{/if}}
{{/let}}
```

For convenience, an `if-let` helper could be proposed at a later stage.

balinterdi:
What would this be replaced by? The `{{let}}` helper? Or would it be moved to an add-on?

Panman8201:
@balinterdi Yep https://github.com/emberjs/rfcs/blob/master/text/0286-block-let-template-helper.md#deprecating-with

kellyselden:
For someone using this effectively, what is their alternative?

```hbs
{{#let parent.child as |child|}}
  {{#if child}}
    {{child}}
  {{/if}}
{{/let}}
```

The above seems very ugly for someone who was happy with how `{{with}}` functioned.

rwjblue:
@kellyselden - Conceptually, that is better referred to as `if-let`, and is absolutely a fine thing. `{{#with` is just a really bad name for that pattern...

kellyselden:
@rwjblue My apologies, I didn't know that was a thing. That is a good replacement.

rwjblue:
This just came up again at a core team call, I think we are all 👍 to adding `if-let` (with today's `{{with}}` semantics) and deprecating `{{with}}` itself. 

We would _love_ help with authoring the RFC from folks with time (and I am happy to champion with the broader group as needed)...

GavinJoyce:
@rwjblue LMK if this is still available, I can author the draft RFC

locks:
@GavinJoyce go for it! Thanks for the help 🙏

PR #445, opened 2019-02-13
Deprecate {{with}}

GavinJoyce:
closes https://github.com/emberjs/rfcs/issues/316

[Rendered](https://github.com/gavinjoyce/rfcs/blob/gj/deprecate-with/text/0000-template.md)

### DIFF
--- a/text/0445-deprecate-with.md
+++ b/text/0445-deprecate-with.md
@@ -1,0 +1,151 @@
+- Start Date: 2019-02-13
+- Relevant Teams: Ember.js, Learning
+- RFC PR: https://github.com/emberjs/rfcs/pull/445
+- Tracking: (leave this empty)
+
+# Deprecate `{{with}}`
+
+## Summary
+
+The `{{with}}` helper has always had slightly confusing conditional semantics, this was one of the motivators for [introducing](https://github.com/emberjs/rfcs/blob/master/text/0286-block-let-template-helper.md) the easier to understand `{{let}}` helper. Now that `{{let}}` exists, the remaining use case for using `{{with}}` is for its unique conditional semantics. These conditional semantics can be cleanly represented with a combination of `{{let}}` and `{{if}}` statements so we should deprecate `{{with}}`.
+
+## Motivation
+
+The difference between `{{let}}` and `{{with}}` is with how they handle conditional arguments. The `{{let}}` helper's block content is always rendered, regardless of its parameters. In contrast, `{{with}}` only renders its main block when the first position parameter is truthy. For example:
+
+```hbs
+{{#with "Alex" as |value|}}
+  {{value}} is truthy
+{{else}}
+  The first positional param was falsy
+{{/with}}
+```
+
+Will render "[Alex] is truthy".
+
+```hbs
+{{#with false as |value|}}
+  {{value}} is truthy
+{{else}}
+  The first positional param was falsy
+{{/with}}
+```
+
+Will render "The first positional param was falsy".
+
+The conditional arguments behavior of `{{with}}` can easily be replicated using a combination of `{{let}}` and `{{if}}` in a way that's easily readable:
+
+```hbs
+{{#let "Alex" as |value|}}
+  {{#if value}}
+    {{value}} is truthy
+  {{else}}
+    The first positional param was falsy
+  {{/if}}
+{{/let}}
+```
+
+## Detailed design
+
+We'll create an AST transform in [`packages/ember-template-compiler`](https://github.com/emberjs/ember.js/tree/master/packages/ember-template-compiler) which will emit a deprecation warning for all uses of `{{with}}`. The deprecation warning will be:
+
+```
+Using `{{with}}` is deprecated, please use `{{let}}` instead.
+```
+
+This message will link to the following deprecation details which aim to give clear guidance on how to migrate to using `{{let}}`, `{{if}}` and `{{else}}` in different combinations:
+
+----
+
+The use of `{{with}}` has been deprecated, you should replace it with either `{{let}}` or a combination of `{{let}}`, `{{if}}` and `{{else}}`:
+
+**If you always want the block to render, replace `{{with}}` with `{{let}}` directly:**
+
+Before:
+
+```hbs
+{{#with (hash name="Ben" age=4) as |person|}}
+  Hi {{person.name}}, you are {{person.age}} years old.
+{{/with}}
+```
+
+After:
+
+```hbs
+{{#let (hash name="Ben" age=4) as |person|}}
+  Hi {{person.name}}, you are {{person.age}} years old.
+{{/let}}
+```
+
+**If you want to render a block conditionally, use a combination of `{{let}}` and `{{if}}`:**
+
+Before:
+
+```hbs
+{{#with user.posts as |blogPosts|}}
+  There are {{blogPosts.length}} blog posts
+{{/with}}
+```
+
+After:
+
+```hbs
+{{#let user.posts as |blogPosts|}}
+  {{#if blogPosts}}
+    There are {{blogPosts.length}} blog posts
+  {{/if}}
+{{/let}}
+```
+
+**If you want to render a block conditionally, and otherwise render an alternative block, use a combination of `{{let}}`, `{{if}}` and `{{else}}`:**
+
+Before:
+
+```hbs
+{{#with user.posts as |blogPosts|}}
+  There are {{blogPosts.length}} blog posts
+{{else}}
+  There are no blog posts
+{{/with}}
+```
+
+After:
+
+```hbs
+{{#let user.posts as |blogPosts|}}
+  {{#if blogPosts}}
+    There are {{blogPosts.length}} blog posts
+  {{else}}
+    There are no blog posts
+  {{/if}}
+{{/let}}
+```
+
+---
+
+For people on older versions of Ember that support `{{let}}`, we'll create an `ember-template-lint` rule that they can use to prevent the use of `{{with}}`.
+
+We'll also create a codemod which will assist people when migrating from `{{with}}` to `{{let}}`.
+
+## How we teach this
+
+We'll mentiton the deprecation in an Ember point release blog post.
+
+As mentioned above, the deprecation message will contain a link to clear guidelines on how to migrate to using `{{let}}`.
+
+There is nothing to remove from the Ember.js Guides since we already teach only the use of `{{let}}`.
+## Drawbacks
+
+This adds a little churn to Ember's API.
+
+## Alternatives
+
+We could leave `{{with}}` as is. I don't believe that this is a good option as the name `{{with}}` is confusing.
+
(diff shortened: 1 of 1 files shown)


rwjblue:
Definitely in favor of migrating `{{with` -> `{{if-let`, thanks for picking this up @GavinJoyce!

text/0000-template.md
```diff
@@ -0,0 +1,90 @@
...
+
+We'll mentiton the new `{{if-let}}` helper in an Ember point release blog post.
+
+The deprecation message will link to good documentation which describe why `{{with}}` was deprecated and gives clear guidelines on how to migrate to using either `{{let}}` or `{{if-let}}`.
+
+## Drawbacks
+
+This adds a litte API and documentation churn.
+
+## Alternatives
```
patocallaghan: There's probably a very good reason for not doing it but just throwing this out there. Rather than increase the template API surface area could we enhance `{{let}}` with the `{{else}}` functionality?

GavinJoyce: @patocallaghan see the [motivation section](https://github.com/emberjs/rfcs/blob/master/text/0286-block-let-template-helper.md#motivation) from the `{{let}}` RFC:

> While the conditional semantics of with are coherent with the other built-in helpers like each and if, users often find this unexpected. The fact that only the first positional parameter of with controls whether the block is rendered might also add to the confusion.

`{{let}}` was purposefully designed to always render its block.

tcjr:
Deprecating `{{with}}` is a good idea.  But `{{if-let}}` seems completely superfluous since you can get the desired behavior with `{{let}}` and `{{if}}`.  

```
{{#let "Alex" as |value|}}
  {{#if value}}
    {{value}} is truthy
  {{else}}
    The first positional param was falsy
  {{/if}}
{{/let}}
```

If the legacy `{{with}}` didn’t exist, would we even consider adding `{{if-let}}` to core?  

What about `{{unless-let}}`?

hemlok:
> But `{{if-let}}` seems completely superfluous since you can get the desired behavior with `{{let}}` and `{{if}}`.

Another benefit of composing {{#let}} and nested {{#if}} is that you could handle multiple params and react to different positional params being false-y in different ways in the UI 🤔

GavinJoyce:
@tcjr Thanks, I'll add your suggestion to the list of alternatives. I crafted this RFC based on the comments in https://github.com/emberjs/rfcs/issues/316#issuecomment-449489686, it seemed to indicate the the core team where in favor of adding `{{if-let}}`.

I actually prefer your example of combining `{{let}}` and `{{if}}`, I think it reads better. Also, if we were to add `{{if-let}}`, I don't believe that it would be widely used. I can't think of a place in our own app where we would us it.

> This just came up again at a core team call, I think we are all 👍 to adding `if-let` (with today's `{{with}}` semantics) and deprecating `{{with}}` itself.

@rwjblue do you remember why there was support from the core team for adding `{{if-let}}`. Was a combination of `if` and `let` also considered?

GavinJoyce:
Thinking about it more, I strongly agree that we should just deprecate `{{with}}` and not introduce `{{let-if}}`. I'll update the RFC to reflect this now. If some good reasons emerge to introduce `{{let-if}}` and there is consensus, I can add it back to the RFC.

kellyselden:
@GavinJoyce Perhaps you could also alter the RFC to ensure public API exists so that and addon can implement `if-let` without hacks (might already be possible).

GavinJoyce:
@kellyselden here's a simple `{{if-let}}` component implementation:

https://ember-twiddle.com/9af662e85c5f06b6e4cb7f58ab08c954?openFiles=templates.application.hbs%2Ctemplates.components.if-let.hbs

screen shot 2019-02-14 at 13 27 26

GavinJoyce:
😜I got the naming wrong in the twiddle above, used `{{if-let}}` instead of `{{let-if}}`. I think that's another good reason not to introduce `{{let-if}}` in ember core.

rwjblue:
I think that @tcjr / @hemlok's intuition here is good. Manually writing out the nested version of if-let is easy enough, is not much longer than the shorthand `{{if-let` version, and is _probably_ more intuitive for folks that don't have experience with languages that themselves have a built in `if-let` (e.g. rust or some lisps).

tldr; 👍 on just making this a straight up deprecation RFC...

text/0000-template.md
```diff
@@ -0,0 +1,78 @@
...
+  {{/if}}
+{{/let}}
+```
+
+## Detailed design
+
+We'll create an AST transform in [`packages/ember-template-compiler`](https://github.com/emberjs/ember.js/tree/master/packages/ember-template-compiler) which will emit a deprecation warning for all uses of `{{with}}`. The deprecation warning will be something like:
+
+```
+Using `{{with}}` is deprecated, please use a combination of `{{let}}` and `{{if}}` instead.
```
rwjblue: I think that the message should only suggest `if` _if_ there is an `{{else`, what do you think?

text/0000-template.md
```diff
@@ -0,0 +1,78 @@
...
+```
+
+This new deprecation warning should link to some documentation which clearly describes when you should use `{{let}}` and when you should use `{{let}}` and `{{if}}` in combination.
+
+## How we teach this
+
+We'll mentiton the deprecation in an Ember point release blog post.
+
+The deprecation message will link to good documentation which describe why `{{with}}` was deprecated and gives clear guidelines on how to migrate to using either `{{let}}` or `{{let}}` and `{{it}}` in combination.
```
rwjblue: We should include the deprecation guide prose in this section directly. The deprecation guide entry will be linked to directly from the deprecation itself, and should show before/after examples for the various styles of `{{with` usage.

text/0000-template.md
```diff
@@ -0,0 +1,78 @@
...
+
+We could leave `{{with}}` as is. I don't believe that this is a good option as the name `{{with}}` is confusing.
+
+We could deprecate `{{with}}` and introduce `{{if-let}}` in Ember core. This RFC originally made that exact proposal, I was strongly persuaded of the [lack of need for `{{if-let}}` by @tcjr](https://github.com/emberjs/rfcs/pull/445#issuecomment-463594185).
+
+We could deprecate `{{with}}` and introduce `{{if-let}}` in an addon instead of Ember core.
+
+## Unresolved questions
+
+This [comment](https://github.com/emberjs/rfcs/issues/316#issuecomment-449489686) seems to indicate some support for `{{if-let}}` from the core team. Perhaps there are some compelling reasons for add it that are not currently included in this RFC?
```
rwjblue: I _personally_ like `if-let`, but I think that this is just as easy to understand and doesn't rely on the reader to grok extra concepts....

text/0000-template.md
GavinJoyce: Perhaps some people rely on the main `{{with}}` block not rendering when the first argument in falsy? If so, I believe that they would need to migrate to using `{{let}}` and `{{if}}`.

dgeb:
Good discussion. I also favor deprecating `{{with}}` without introducing a new replacement simultaneously.

text/0000-template.md
rwjblue: Ya, I totally agree. My _guess_ (and anecdotally what I've found when teaching/pairing) is that most folks are **surprised** by the current behavior and actually do want "render this block all the time".

I think I'd propose the following "choose your own adventure" for the deprecation guide on migrating away from `{{with` usage:

* if you always want to render a given block unconditionally: use `{{let`
* if you want to render a given block conditionally: use `{{let` + `{{if`
* if you want to render a given block conditionally, otherwise rendering an alternate block: use `{{let` + `{{if` + `{{else`

rwjblue:
I'd also like to suggest creating a linting rule (to aid folks who are not yet on newer versions with the deprecation but can use `{{let`, to avoid `{{with`) and a codemod (leveraging `codemod-cli` + `ember-template-recast` _should_ make this straight forward).

text/0000-template.md
GavinJoyce: 👍 sounds good, thanks

GavinJoyce: @rwjblue I've added the deprecation guide content to this RFC

GavinJoyce:
> I'd also like to suggest creating a linting rule

✅ 

> and a codemod (leveraging `codemod-cli` + `ember-template-recast` _should_ make this straight forward).

@rwjblue for the same reasons as why we need to provide three options in the deprecation guide, I'm not sure that it will be possible to fully automate the migration from `{{with}}` to `{{let}}`

rwjblue:
> @rwjblue for the same reasons as why we need to provide three options in the deprecation guide, I'm not sure that it will be possible to fully automate the migration from `{{with}}` to `{{let}}`

Firstly, there _is_ a fully correct codemod that we could write. It would just look hideous in most cases when you _didn't_ want the conditional behaviors.

For example:

```hbs
{{#with (concat first ' ' last) as |name|}}
  <h2>Hello, {{name}}!</h2>
{{/with}}
```

Would have to be written as:

```hbs
{{#let (concat first ' ' last) as |name|}}
  {{#if name}}
    <h2>Hello, {{name}}!</h2>
  {{/if}}
{{/let}}
```

Even though we _mostly_ know that we don't need that conditional.

---

I think the codemod could actually have some pretty nice heuristics to be _pretty_ close to what you would do by hand:

* if the `{{#with` has an `{{else`, always emit the nested `{{#if`
* if the argument to `{{#with` is a subexpression, assume that it is non-falsey (and therefore do not emit the nested `{{#if` inside the block)
* otherwise use a "simple" `{{#let` without nested if

In addition, we could allow a flag to the codemod so that the person running it can opt in to "100% correctness, but super verbose" mode. The [Ember.K](https://emberjs.github.io/rfcs/0178-deprecate-ember-k.html) codemod did essentially the same thing by _requiring_ a flag to assume "chaining" or not ([codemod docs here](https://github.com/abuiles/ember-watson#remove-usages-of-emberk)).

GavinJoyce:
> I think the codemod could actually have some pretty nice heuristics to be _pretty_ close to what you would do by hand

👍thanks, I've added a short note on adding a codemod to assist the migration

text/0000-template.md
```diff
@@ -0,0 +1,150 @@
...
+For people on older versions of Ember that support `{{let}}`, we'll create an `ember-template-lint` rule that they can use to prevent the use of `{{with}}`.
+
+We'll also create a codemod which will assist people when migrating from `{{with}}` to `{{let}}`.
+
+## How we teach this
+
+We'll mentiton the deprecation in an Ember point release blog post.
+
+As mentioned above, the deprecation message will contain a link to clear guidelines on how to migrate to using `{{let}}`.
+
```
jenweber:
```suggestion

There is nothing to remove from the Ember.js Guides since we already teach only the use of `{{let}}`.
```

chancancode:
We discussed this today at the framework core team meeting and decided to move this RFC into Final Comment Period

text/0000-template.md

chancancode:
We decided to accept this RFC at the framework core team meeting today. Thanks @GavinJoyce for working on this and everyone else for the discussion 🎉

chancancode:
@GavinJoyce do you mind renaming the file to match the RFC number?

GavinJoyce:
@chancancode updated, thanks

Issue #390, opened 2018-10-16
Pre-RFC: Deprecate partials


tomdale:
Partials are widely believed to be a bad idea, but we don't do enough to help people understand that they should *always* prefer a component to a partial.

We should deprecate partials in Ember, ideally as part of Octane. I believe the only thing blocking this currently is the lack of a deprecation RFC. We may also want a code mod that can automatically migrate the most common partial cases, although I'm not sure if I'd want to block deprecation on that.

rwjblue:
>  We may also want a code mod that can automatically migrate the most common partial cases, although I'm not sure if I'd want to block deprecation on that.

I would like to see a codemod mentioned in the deprecation RFC, but I don't think we need to hold up work on implementing the deprecation for any codemod work...

knownasilya:
Would https://github.com/dyfactor/dyfactor be the tool for that? I know it's still young, but I think it could handle all partial migrations if done right.

ctcpip:
@lbeauvilliers it's happening

lbeauvilliers:
@ctcpip 😱

I mean ...  🎉

webark:
parsing the partial for all variables, and passing all those into a component seems like a harry task for a codemod..

rwjblue:
You say hairy, I say fun 😜

Alonski:
So just chiming in that after upgrading to Ember v3.7 I am seeing a regression related to Partials that is causing an error in my app.
I am all for deprecating partials!

rwjblue:
@Alonski - You might want to report that over on emberjs/ember.js, we haven't intentionally broken anything related to partials...

GavinJoyce:
Is there anything blocking the initial RFC from being created? If not, and it suits, I'd like to volunteer to author the RFC and then help implement the changes needed if it lands.

rwjblue:
Nope, nothing blocking other than needing someone to work on it. 😸

GavinJoyce:
👍 thanks, I'll draft an RFC soon

GavinJoyce:
Here's the RFC to deprecate `{{partial}}`s: 

https://github.com/emberjs/rfcs/pull/449

ngouy:
the idea of a "kind of component" having the same context as the parent was sometimes very helpfull : 
https://blog.nathangouy.com/emberjs-full-component-inheritance/

[...]

PR #449, opened 2019-02-17
Deprecate {{partial}}

GavinJoyce:
closes https://github.com/emberjs/rfcs/issues/390

[Rendered](https://github.com/gavinjoyce/rfcs/blob/gj/deprecate-partials/text/0000-template.md)

### DIFF
--- a/text/0000-template.md
+++ b/text/0000-template.md
@@ -1,0 +1,102 @@
+- Start Date: 2019-02-17
+- Relevant Teams: Ember.js, Learning
+- RFC PR: https://github.com/emberjs/rfcs/pull/449
+- Tracking: (leave this empty)
+
+# Deprecate `{{partial}}`
+
+## Summary
+
+Partials are an old Ember construct that have no benefits and many downsides when compared to modern Ember components. We should deprecate them in favor of components.
+
+## Motivation
+
+Partials have a number of downsides when compared with components: 
+
+ - They are hard to reason about as they inherit the scope of the calling template
+ - They perform poorly in comparison to components
+
+Partials should have no place in modern Ember applications, components should always be preferred.
+
+Once removed, Ember's API will become smaller and more consistent.
+
+## Detailed design
+
+We'll create an AST transform in [`packages/ember-template-compiler`](https://github.com/emberjs/ember.js/tree/master/packages/ember-template-compiler) which will emit a deprecation warning for all uses of `{{partial}}`. The deprecation warning will be:
+
+```
+Using `{{partial}}` is deprecated, please use a component instead.
+```
+
+This message will link to the following deprecation details which aim to give clear guidance on how to migrate to component:
+
+---
+
+The use of `{{partial}}` has been deprecated, you should replace the partial with a component as follows:
+
+Let's consider a simple example of a `partials/quick-tip` partial being invoked from the `application.hbs` template:
+
+`app/templates/application.hbs`
+
+```hbs
+{{#let (hash title="Don't use partials" body="Components are always better") as |tip|}}
+  {{partial "partials/quick-tip"}}
+{{/let}}
+```
+
+`app/templates/partials/quick-tip.hbs`
+
+```hbs
+<h1>Tip: {{tip.title}}</h1>
+<p>{{tip.body}}</p>
+<button {{action 'dismissTip'}}>OK</button>
+```
+
+Here's the same template code after migrating the `partials/quick-tip` partial to be a component.
+
+`app/templates/application.hbs`
+
+```hbs
+{{#let (hash title="Don't use partials" body="Components are always better") as |tip|}}
+  <QuickTip @tip={{tip}} @onDismiss={{action 'dismissTip'}} />
+{{/let}}
+```
+
+`app/templates/components/quick-tip.hbs`
+
+```hbs
+<h1>Tip: {{@tip.title}}</h1>
+<p>{{@tip.body}}</p>
+<button {{action @onDismiss}}>OK</button>
+```
+
+---
+
+A codemod, while not necessary for this RFC to land, would greatly simplify the migration of partials to components. We should endeavor to create this codemod as part of efforts to implement this RFC. The codemod might work as follows:
+
+ * Examine each partial template to recognize which properties originate from the caller scope
+ * Generate a component to replace each partial
+ * Replace each `{{partial "foo-bar"}}` invocation with a component invocation, passing arguments as required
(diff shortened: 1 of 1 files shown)


andremalan:
This is an RFC that I'd assumed had happened already! We've been treating partials as deprecated for years!

text/0000-template.md
```diff
@@ -0,0 +1,98 @@
+- Start Date: 2019-02-17
+- Relevant Teams: Ember.js, Learning
+- RFC PR: https://github.com/emberjs/rfcs/pull/TODO
```
galfert:
```suggestion
- RFC PR: https://github.com/emberjs/rfcs/pull/449
```

text/0000-template.md
```diff
@@ -0,0 +1,98 @@
...
+## Motivation
+
+Partials have a number of downsides when compared with components: 
+
+ - They are hard to reason about as they inherit the scope of the calling template
+ - They perform poorly in comparison to components
+
+Partials should have no place in modern Ember applications, components should always be preferred.
+
+Once removed, Ember's API will become smaller and more consistetnt.
```
galfert:
```suggestion
Once removed, Ember's API will become smaller and more consistent.
```

text/0000-template.md

text/0000-template.md
GavinJoyce: 👍

ngouy:
I guess it's more appropriate to post it here :

- `They are hard to reason about as they inherit the scope of the calling template` -> i'm not sure to understand. It's it their reason to be ?

the idea of a "kind of component" having the same context as the caller was sometimes very helpful :
https://blog.nathangouy.com/emberjs-full-component-inheritance/

I understand the reason of deprecating this for simplifications and performence reasons, I just wonder how it will be possible to do something like that later on... passing all attributes with `a=a b=b c=c`. (Not sure but did an RFC or smt opened for passing argument with `...` notation ? I have the impression I saw something like this)

rwjblue:
> Not sure but did an RFC or smt opened for passing argument with `...` notation ? I have the impression I saw something like thi

There is `...attributes`, but it does not work for arguments. It would _probably_ be possible to allow something akin to `...arguments` to angle bracket invocation, but it hasn't been done (or designed) just yet.

ngouy:
Ok my bad I thought I saw something like this in a RFC. It would have been a great tool to have a smooth transition on this one, but giant thing to implement due to bunch of corner cases I guess.

tomdale:
Great writeup, @GavinJoyce. This is long overdue, and perhaps one of the least controversial deprecations RFCs to ever come across the core team desk. 😉 We're moving this one to Final Comment Period, so please chime in now if you've got concerns with the proposal as written.

text/0000-template.md
```diff
@@ -0,0 +1,98 @@
...
+ * Replace each `{{partial "foo-bar"}}` invocation with a component invocation, passing arguments as required
+ * Delete the partial handlebars files
+
+ If we do implement a codemod, we should mention it in the deprecation details above.
+
+## How we teach this
+
+We'll mentiton the deprecation in an Ember point release blog post.
+
+The deprecation message will contain a link to clear guidelines on how to migrate from a `{{partial}}` to a component.
```
jenweber:
```suggestion
The deprecation message will contain a link to clear guidelines on how to migrate from a `{{partial}}` to a component. See the section above beginning with "The use of `{{partial}}` has been deprecated..." for the content.
```

Just a small pointer for our possible future contributors. Thanks for composing it ahead of time!

mixonic:
Regarding the migration script: Mostly the use cases considered are looking at downward data flow. We should consider `<button {{action 'foo'}}>` inside a partial- I presume once you move that into a glimmer component the action won't behave the same way it did in the partial.

rwjblue:
We discussed this at todays Ember framework core team meeting, the team decided to move this forward into final comment period.

chancancode:
@mixonic yep that would break because this form of action implicitly closes over `this`, we should make sure the migration guide mentions that (and accounted for in the codemod, if there is one?), and suggest passing it in as an argument:

```hbs
{{!--outside--}}
<ConvertedFromPartial ... @action={{action 'foo'}} .../>

{{!--inside--}}
<button {{action @action}}>
```

omairvaiyani:
I'm fully in favour of this deprecation, but I don't think a codemod that fully automates the migration will land. There is too much implicit two-day data flow going on with partials to be safely handled by static analysis. If this [library](https://github.com/dyfactor/dyfactor) was still active, it may have been possible, but probably beyond difficult for smaller teams to manage.

I think a codemod that superficially converts partials to components (and generates the placeholder component files), followed by a list of items migrated, that would greatly help with the migration process, but a large part of it will most likely be arduous and manual. Still worth it as we've noted our partials being a point of incredible confusion for our junior web developers.

rwjblue:
> There is too much implicit two-day data flow going on with partials to be safely handled by static analysis.

@omairvaiyani agree to disagree 😸, I believe the codemod steps discussed in this RFC would work fine and the number of cases that were not possible would be limited to those where you have partial usage _inside_ of a partial.

However, I _also_ agree with you that **even without** codemod support, this RFC is still worthwhile.

chancancode:
@GavinJoyce can you update the RFC to include the `{{action "name"}}` example? Today at the core team meeting, we agreed to accept the RFC with that minor modification, so I'll merge when that's sorted.

GavinJoyce:
@chancancode I've updated the deprecation details to include an example action:

---

`app/templates/application.hbs`

```hbs
{{#let (hash title="Don't use partials" body="Components are always better") as |tip|}}
  {{partial "partials/quick-tip"}}
{{/let}}
```

`app/templates/partials/quick-tip.hbs`

```hbs
<h1>Tip: {{tip.title}}</h1>
<p>{{tip.body}}</p>
<button {{action 'dismissTip'}}>OK</button>
```

Here's the same template code after migrating the `partials/quick-tip` partial to be a component.

`app/templates/application.hbs`

```hbs
{{#let (hash title="Don't use partials" body="Components are always better") as |tip|}}
  <QuickTip @tip={{tip}} @onDismiss={{action 'dismissTip'}} />
{{/let}}
```

`app/templates/components/quick-tip.hbs`

```hbs
<h1>Tip: {{@tip.title}}</h1>
<p>{{@tip.body}}</p>
<button {{action @onDismiss}}>OK</button>
```

chancancode:
🎉 Thanks @GavinJoyce!

PR #452, opened 2019-02-21
[Data] Medium Term Plan

igorT:
[Rendered](https://github.com/igorT/rfcs/blob/igor/ember-data-medium-plan/text/0452-ember-data-medium-term-plan.md)

### DIFF
--- a/text/0452-ember-data-medium-term-plan.md
+++ b/text/0452-ember-data-medium-term-plan.md
@@ -1,0 +1,145 @@
+- Start Date: 2019-02-20
+- Relevant Team(s): Ember Data
+- RFC PR: [452](https://github.com/emberjs/rfcs/pull/452)
+- Tracking: [Issue](https://github.com/emberjs/rfc-tracking/issues/50)
+
+# **Summary**
+
+This RFC sets the plan for Ember Data's medium term development, on the path towards modularizing the library. We expect the following immediate design and implementation work to take between one and two quarters. The main goals are:
+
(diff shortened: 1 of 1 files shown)


mikkopaderes:
Not trying to be rude or anything and maybe it's just me not being able to fully absorb everything in this RFC. ✌️ 

Basically, my take is shouldn't the community just be concerned with the public APIs that'll be exposed to us? I feel the core team should be the one to decide how they want to design and structure their code base so long as they commit to the promised APIs and not break existing ones. There would be less overhead for the community to take in the RFC details and will let you guys (and other community members) to just dev immediately without waiting for our feedback.

pete-the-pete:
Sounds good, and thanks for the diagrams!

The list of RFCs looks like a commitment to some ideas, which sounds good, but what does that mean for other RFCs? Can some be closed? Will others be folded in?

igorT:
@mikkopaderes Your perspective makes a lot of sense, and I think in general internal refactors usually don't go through an RFC process. However given the size of the change, and the long term nature of the work, we felt that it made sense to have a place to lay out the plan and propose and debate the direction and align the contributors.. While not perfect, an RFC made sense for that.

igorT:
@pete-the-pete Great point. This RFC spawned several new ones, and with regards to existing ones, we are going through them and addressing them, though most of them aren't impacted by these specific changes, other than the fact it should be easier to implement some of the RFC functionality in addons as opposed to ED core

PR #454, opened 2019-02-23
SFC & Template Import Primitives

tomdale:
## Summary

Expose low-level primitives for associating templates with component classes and customizing a template's ambient scope.

These primitives unlock experimentation, allowing addons to provide highly-requested features (such as single-file components and template imports) via stable, public API.

[Rendered](https://github.com/emberjs/rfcs/blob/tomdale/template-primitives/text/000-sfc-and-template-import-primitives.md)

### DIFF
--- a/text/000-sfc-and-template-import-primitives.md
+++ b/text/000-sfc-and-template-import-primitives.md
@@ -205,15 +205,14 @@ import BlogPost from './components/blog-post';
 
 import { createTemplateFactory, setComponentTemplate } from '@ember/template-factory';
 
-const templateJSON$1 = {
+const wireFormat$1 = {
   "id": "ANJ73B7b",
   "block": "{\"statements\":[\"...\"]}",
-  "meta": { "moduleName": "src/ui/components/MyComponent.js" }
+  "meta": { "moduleName": "src/ui/components/MyComponent.js" },
+  "scope": () => [BlogPost, titleize]
 };
 
-const template$1 = createTemplateFactory(templateJSON$1, {
-  scope: () => [BlogPost, titleize]
-});
+const template$1 = createTemplateFactory(wireFormat$1);
 
 export default setComponentTemplate(template$1, class MyComponent extends Component {
   @service session;
@@ -224,7 +223,7 @@ Note that the file formats shown above are hypothetical and used as
 illustrative examples only. This RFC does *not* propose a recommended file
 format. Rather, the goal is to expose public JavaScript API for:
 
-1. Compiling a template into JSON.
+1. Compiling a template into a JavaScript wire format.
 3. Associating a compiled template with its backing JavaScript class.
 2. Specifying which values are in scope for a given template at runtime.
 
@@ -240,8 +239,8 @@ This RFC proposes the following new modules and exports:
 
 Name | Description  
 ---------|----------
-`import { precompile } from '@ember/template-compiler'` |  Compiles template source code into a JSON wire format.
-`import { createTemplateFactory } from '@ember/template-factory'` | Creates a template factory from wire format JSON, provides scope values.
+`import { precompile } from '@ember/template-compiler'` |  Compiles template source code into a JavaScript wire format.
+`import { createTemplateFactory } from '@ember/template-factory'` | Creates a template factory wrapper around the wire format.
 `import { setComponentTemplate } from '@ember/template-factory'` |  Associates a template factory with a JavaScript component class.
 
 Detailed descriptions and rationales for these APIs are provided below.
@@ -260,7 +259,7 @@ the first time a template is rendered.
 Ember uses Glimmer's JIT mode, so templates are sent to the browser as an
 optimized JSON data structure we call the _wire format_. The
 `@glimmer/compiler` package provides a helper function called `precompile`
-that turns raw template source code into this "pre-compiled" JSON wire format.
+that turns raw template source code into this "pre-compiled" wire format.
 
 ```js
 import { precompile } from '@glimmer/compiler';
@@ -279,28 +278,28 @@ const json = precompile(`<h1>{{this.firstName}}</h1>`);
 
 The exact structure of the wire format returned from `precompile` is not
 specified, is not considered public API, and is likely to change across minor
-versions. Other than a guarantee that it is a string that can be safely
-parsed via `JSON.stringify`, users should treat the value returned by
-`precompile` as completely opaque.
+versions. Other than a guarantee that it is a string that can be embedded in
+JavaScript and is a valid JavaScript expression, users should treat the value
+returned by `precompile` as completely opaque.
 
 ### Template Factories
 
 Once we have our template compiled into the wire format, we need a way to
-annotate it so that Ember knows that the JSON value represents a compiled
-template. The wrapper object around the wire format JSON that does this is
-called a _template factory_.
+annotate it so that Ember knows that the value represents a compiled
+template. The wrapper object around the wire format that does this is called
+a _template factory_.
 
 This RFC proposes exposing a function for creating template factories called
 `createTemplateFactory`, exported from the `@ember/template-factory` package.
 
-Users would use this function to turn the raw, wire format JSON into a template
+Users would use this function to turn the raw wire format into a template
 factory object at runtime:
 
 ```js
 import { createTemplateFactory } from '@ember/template-factory';
 
-const json = /* wire format JSON here */
-const templateFactory = createTemplateFactory(json);
+const wireFormat = /* embed wire format from precompile here */
+const templateFactory = createTemplateFactory(wireFormat);
 ```
 
 ### Associating Template Factories and Classes
@@ -352,8 +351,8 @@ arguments:
 import Component from '@glimmer/component';
 import { createTemplateFactory, setComponentTemplate } from '@ember/template-factory';
 
-const json = /* wire format JSON here */
-const templateFactory = createTemplateFactory(json);
+const wireFormat = /* wire format here */
+const templateFactory = createTemplateFactory(wireFormat);
 
 class MyComponent extends Component {
   /* ... */
@@ -403,87 +402,149 @@ superclass.
 In this context, _scope_ refers to the set of names that can be used to refer
 to values, helpers, and other components from within your template. For
 example, when you type `{{@firstName}}`, `{{this.count}}`, or `<UserAvatar
-/>`, how does Ember (or even you, as the programmer) figure out what each of these
-refers to? The answer to that is determined by the template's scope.
+/>`, how does Ember know what each of these refers to? The answer to that is
+determined by the template's scope.
 
 Today, the names available inside a particular template are determined by a
 few different rules:
 
-1. `this` always refers to the component instance (if the template has a backing class).
-2. Arguments (like `@firstName`) are placed into scope when you invoke the component.
-3. Block parameters (like `item` in `{{#each @items as |item|}}`) are in scope, but only inside the block.
-4. Anything else must be looked up dynamically through Ember's container.
-   This lookup process is somewhat complex and not always as fast as we'd like.
+1. `this` always refers to the component instance (if the template has a
+   backing class).
+2. Arguments (like `@firstName`) are placed into scope when you invoke the
+   component.
+3. Block parameters (like `item` in `{{#each @items as |item|}}`) are in
+   scope, but only inside the block.
+4. Anything else, like components and helper names, must be looked up
+   dynamically through Ember's container. This lookup process is somewhat
+   complex and not always as fast as we'd like.
 
-Because the rules of how components, helpers, and other values are looked up
-can be confusing for new learners, we'd like to explore alternate APIs that
-allow users to explicitly import them and refer to them from templates.
+Because the rules for how components and helpers are looked up are implicit
+and can be confusing for new learners, we'd like to explore alternate APIs
+that allow users to explicitly import values and refer to them from
+templates.
 
-Today, it's not easy for addons to add additional names to a template's scope
-that take precedence over dynamic resolution.
+This RFC proposes that templates compiled via the `precompile` function
+resolve unknown identifiers from the ambient JavaScript scope, rather than
+performing dynamic resolution through the container.
 
-This RFC proposes that the `precompile` function, described earlier, also
-accepts a `scope` option that specifies additional identifiers (as strings)
-that are available in the template's scope:
+To do this, `precompile` emits JavaScript code that contains references to
+the same identifiers as those used in the template:
 
 ```js
 import { precompile } from '@ember/template-compiler';
 
-const json = precompile(`{{t "Hello!"}} <User @user={{this.currentuser}} />`, {
-  scope: ['User', 't']  
-});
+const wireFormat = precompile(`{{t "Hello!"}} <User @user={{this.currentUser}} />`);
+/* => `{
+  "id": "ANJ73B7b",
+  "block": "{\"statements\":[\"...\"]}",
+  "scope": () => [t, User]
+}`                ^---- template identifiers emitted as JavaScript references
+*/
 ```
 
-The `scope` option is an array of zero or more strings. In this example, the
-identifiers `User` and `t` are added to the template scope. Identifiers
-provided this way are in scope for the entire template (unless shadowed by a
-block argument).
-
-This RFC further proposes that the `createTemplateFactory` function,
-described earlier, also accepts a `scope` option that returns the reified
-scope values at runtime:
+If a template references an identifier that is not in lexical
+scope, a `ReferenceError` will be produced at runtime:
 
 ```js
-import { createTemplateFactory } from '@ember/template-factory';
-import User from './OtherComponent';
-import t from '../helpers/i18n/t';
+import { precompileTemplate } from '@ember/template-compilation';
 
-const json = /* wire format JSON here */
-const template = createTemplateFactory(json, {
-  scope: () => [User, t]
-});
+// Compiles successfully but the generated code will throw a ReferenceError when
+// evaluated.
+const wireFormat = precompile(`<MissingComponent />`);
+/* => `{
+  "id": "ANJ73B7b",
+  "block": "{\"statements\":[\"...\"]}",
+  "scope": () => [MissingComponent]
+}`                ^------ will produce ReferenceError
+*/
 ```
 
-Note that, unlike with `precompile`, the `scope` option here is not an array
-but a function that _returns_ an array. This allows the array to be created
-after the JavaScript module has finished evaluating, ensuring the bindings
-included in the array have had time to be fully initialized.
+Optionally, you can provide an array of strings specifying the ambient
+identifiers in scope. If a `scope` array is provided and the template
+contains an identifier that is not present in the list, that reference will
+fall back to today's dynamic resolution lookup behavior:
 
-The order of the array is unimportant, other than that the same order must be
-preserved between the identifiers passed to `precompile` and the
-corresponding values passed to `createTemplateFactory`.
+```js
+// While `t` comes from ambient JavaScript scope, `User` will be looked up
+// through the container.
+const wireFormat = precompile(`{{t "Hello!"}} <User @user={{this.currentUser}} />`, {
+  mode: 2,
+  scope: ['t']
+}); /* => `{
+  "id": "ANJ73B7b",
+  "block": "{\"statements\":[\"...\"]}",
+  "scope": () => [t]
+}`                ^----- only includes identifiers specified in `scope`
+*/
+```
 
-Values provided to the template scope must be constant. Once the `scope` array has been created,
-mutation of those values is unsupported and may cause unexpected errors
-during render. Future RFCs may introduce support for scope values that change
-over time.
+```js
+import { precompile } from '@ember/template-compilation';
 
-Scope values are also limited to component classes and template helpers (i.e.
-subclasses of `Helper` or functions wrapped with `helper()` from
-`@ember/component/helper`). However, support for additional kinds of scope
-values may be expanded in future RFCs.
+const wireFormat = precompile(`<MissingComponent />`, {
+  scope: ['BlogPost', 'OtherComponent']
+})
+/* => Throws an exception: "MissingComponent is not defined."   */
+```
 
-If any of the following conditions are met, a runtime exception will be thrown:
+Addons that need to support existing templates that rely on container lookup
+to discover components and helpers should provide the list of valid
+identifiers via `scope`. Omitting this list causes all lookups to happen via
+ambient scope rather than through the container.
 
-* The `scope` function returns a value that is not an array.
-* The `scope` function returns an array containing a value that is not a
-  component class, helper, or `undefined`.
-* The `scope` function returns an array whose length is different than
-  the `scope` array provided to `precompile`.
-* A `scope` was provided when calling `precompile`, but is not provided when
-  calling `createTemplateFactory` with the resulting wire format.
-* A `scope` was _not_ provided when calling `precompile`, and _is_ provided when
-  calling `createTemplateFactory` with the resulting wire format.
+As of this RFC, scope values resolved via ambient scope are limited to
+component classes and template helpers (i.e. subclasses of `Helper` or
+functions wrapped with `helper()` from `@ember/component/helper`). However,
+support for additional kinds of scope values may be expanded in future RFCs.
+
+### Compilation Modes
+
+To support future evolution of template syntax and semantics, this RFC
+defines a `mode` option that can specify different rules for template
+compilation.
+
+This RFC defines a single mode:
+
+   Mode  | ID | Description
+---------|--------------|------------
+**Classic** | 1 | All deprecated and non-deprecated syntax supported. Mix of dynamic resolution and JavaScript scope supported.
+
+Modes are specified as an integer ID to reduce the size of compiled output
+and to allow additional modes to be added in the future. To specify the mode,
+pass a `mode` option to `precompile`:
+
+```js
+const wireFormat = precompile(`{{t "Hello!"}} <User @user={{this.currentUser}} />`, {
+  mode: 1,
+  scope: ['t']
+}); /* => `{
+  "id": "ANJ73B7b",
+  "block": "{\"statements\":[\"...\"]}",
+  "scope": () => [t]
+}`
+*/
+```
+
+If the mode is not specified, classic mode (mode `1`) is assumed. Future RFCs
+may specify additional modes.
+
+In strict mode, templates that references names that are not present in a supplied
+`scope` array produce a compile-time error. In sloppy mode, references to names
+that are not specified in the `scope` array will fall back to dynamic
+resolution:
+
+```js
+// `t` comes from ambient scope, `User` will be looked up.
+const wireFormat = precompile(`{{t "Hello!"}} <User @user={{this.currentUser}} />`, {
+  mode: 2,
+  scope: ['t']
+}); /* => `{
+  "id": "ANJ73B7b",
+  "block": "{\"statements\":[\"...\"]}",
+  "scope": () => [t]
+}`                ^----- only includes identifiers specified in `scope`
+*/
+```
 
 ## How we teach this
 
@@ -507,22 +568,34 @@ import { precompile } from '@ember/template-compiler';
 ```
 
 Compiles a string of Glimmer template source code into an intermediate "wire
-format" and returns it as a string. The exact value returned from
-`precompile` is not specified, is not considered public API, and is likely to
-change across minor versions. Other than the fact that it is guaranteed to be
-a string that may be safely parsed by `JSON.parse()`, you should treat the
-value returned by `precompile` as completely opaque.
+format" and returns it as a string of JavaScript code. The exact value
+returned from `precompile` is not specified, is not considered public API,
+and is likely to change across minor versions. The string is guaranteed to be
+a valid JavaScript expression. Other than that, you should treat the value
+returned by `precompile` as completely opaque.
 
-Optionally, an array of additional identifiers can be specified that will be
-included in the template's ambient scope. Each identifier must have a
-corresponding value provided at run time when the template's template factory
-is created.
+By default, components and helpers referenced in the passed template will be
+looked up from the JavaScript scope where the wire format string is inserted.
+To enable legacy container resolution behavior, a `scope` option can be
+provided containing a list of identifiers that should be considered to come
+from ambient JavaScript scope. Any identifiers in the template that are not
+present in the `scope` list will fall back to dynamic container resolution at
+runtime.
+
+An optional `mode` option may also be provided. Currently, only one mode is
+supported, which is the classic mode with ID of `1`. If no mode is specified,
+the default mode of `1` is assumed. Future RFCs may specify additional modes.
 
 ```ts
 function precompile(templateSource: string, options?: PrecompileOptions): string; 
 
 interface PrecompileOptions {
   scope?: string[];
+  mode?: PrecompileMode;
+}
+
+enum PrecompileMode {
+  Classic = 1;
 }
 ```
 
@@ -539,13 +612,7 @@ provided when calling `createTemplateFactory` that returns an array of
 reified values corresponding to each identifier.
 
 ```ts
-function createTemplateFactory(wf: WireFormatJSON, options?: CreateTemplateFactoryOptions): TemplateFactory;
-
-interface CreateTemplateFactoryOptions {
-  scope?: () => ScopeValue[];
-}
-
-type ScopeValue = ComponentFactory | HelperFactory | HelperFunction | undefined;
+function createTemplateFactory(wf: WireFormat): TemplateFactory;
 ```
 
 #### setComponentTemplate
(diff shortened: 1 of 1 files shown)


lifeart:
Regarding conversation about hot-module replacement. Is it possible to `reassign` component's template?

kellyselden:
One of the big selling points of Ember is being able to jump into a random project and know where everything is and how everything works because of the common conventions.

Would you consider this a drawback of the proposal, potentially encouraging drastically different component experiments?

text/000-sfc-and-template-import-primitives.md
```diff
@@ -0,0 +1,679 @@
...
+```
+
+Note that, unlike with `precompile`, the `scope` option here is not an array
+but a function that _returns_ an array. This allows the array to be created
+after the JavaScript module has finished evaluating, ensuring the bindings
+included in the array have had time to be fully initialized.
+
+The order of the array is unimportant, other than that the same order must be
+preserved between the identifiers passed to `precompile` and the
+corresponding values passed to `createTemplateFactory`.
```
gossi: This reads super risky to me. The order at two "independent" locations must be the same. Couldn't it be mapped by name?

gossi:
I like the experimentational idea of this RFC. Like "hey community, here is your toolbelt, now let's see what SFC you'll come up with and what might be the final spec we can compromise and align on".
However, I also do see that there will be plenty of SFC specs out there and ember can face a wild-wild-west situation here, which can end up problematic 😈. Each time creating a new component, you'll need to watch out for the correct addon to match your SFC-style.

I dunno if left out, but template, script and what about styles? Think about css modules or blocks. Do they need an API, too?

ef4:
I would like us strongly consider keeping the template part as HBS within this new standardized format. Not precompiling it to the unstable wire format.

The benefit would be that addons can apply whatever SFC-to-standard transformations they want _at their own publication time_. That is not possible with the current proposal because the template wire format is tied to one exact ember version.

This is the pattern we are already working toward in [embroider](https://github.com/embroider-build/embroider), in that I want addons to publish HBS to NPM, but with any custom preprocessors (including AST transforms) pre-applied. This greatly simplifies the final app build because all templates can share one standardized configuration, rather than setting up special preprocessors per addon.

luxferresum:
When doing `precompile` with a `scope`, how could I provide curly components? Would ```
precompile(`{{my-component}}`, {
  scope: ['my-component']  
});
```
work as expected? Shouldn't we specify components and helpers in separate arrays? And if I want a component to be available for curly and angle-bracket style, do I need to specify it twice?

File: text/000-sfc-and-template-import-primitives.md (hunk 1 above)
Comment (pzuraq) [reply text/000-sfc-and-template-import-primitives.md]: I don't think it matters later on either way, once the scope is correctly associated, either by name or order, you should be able to pull the values off the object/array and use them in a minifiable way. The issue is with the `createTemplateFactory()` in the first place - creating that object and assigning the keys must be shipped to the client, and _that_ part cannot be minified.

If there are a fair number of imports per file, say 5, this could add up in large apps. It's not a _massive_ amount, yes, and if this were an API expected to be used by your average user I would totally agree, but the idea here is that this will be _compiled output_ from a preprocessor, so it's much less likely that it'll get messed up IMO

**Comment by chriskrycho** [comment]
General read so far: 👍 on the overall direction, but I have serious (if as of yet somewhat inchoate) concerns about the `precompile` and scope API. The rationale in **Alternatives** makes sense to me, but I'm going to mull over the rest of the weekend and early next week and try to make those concerns concrete and potentially outline some alternatives.

This combination of things is what makes me 😬:

- the bag-o'-names definition for `scope` (which I don't see _any_ good way to write types for – I'm sure we could come up with something, but it won't be pretty)
- the corresponding *need* for a bunch of runtime checks (with errors thrown for them)
- the drawbacks noted around runtime (vs. AOT) behavior, which is closely related to the runtime semantics of the previous two points

All of those point to the need for a more statically- than dynamically-checked API here. (IMO they point *very strongly* to that, but I'm definitely biased toward static/build-time checks over runtime checks.)

More to come—hopefully with some API ideas—after I've had some further time to think on this.

**Comment by pzuraq** [comment]
@kellyselden totally agree that the conventional file layout is a huge benefit to us (once you learn it)! I think we definitely should keep the resolution rules for things like routes, controllers, and services long term. These are the backbone of an app's structure IMO.

However, for component's and helper functions I feel like we've been struggling against this system for quite some time. Some components are meant to be reused everywhere, and some are meant for specific use cases, but _all_ components get put in the same place, and the lack of granularity and the ability to organize and sort these different types of components can get overwhelming quickly!

This is why "local lookup" became a thing in MU, it was to allow some more flexibility. But the rules for local lookup meant:

1. Things are no longer simple, you can't just look at the `components` folder, you need to look at several different folders to find the thing you're looking for.
2. You now have a large amount of implicit state about the app to reason about what a particular thing resolves to, _especially_ for private collections and name collisions.

I think explicit imports solves these problems a bit more nicely, it gives each app more flexibility for organization, and it means you can easily figure out where a given value is coming from very quickly. The downside is app file structures could become less conventional, as you pointed out, but I think a better way to approach that would be with linting rules, which allow us to guide the ecosystem toward a convention, without _forcing_ every app to follow it if it doesn't meet their needs.

**Comment by tomdale** [comment]
@lifeart:

> Regarding conversation about hot-module replacement. Is it possible to `reassign` component's template?
> Kinda `hey, component manager, use this layout for this component, instead of early assigned`

No, this is not supported at this time. Making templates static allows for the template compiler to emit faster bytecode that does not need to insert guards around every component invocation that checks if the template has changed. It also means component invocations can be compiled to bytecode that jumps directly to a fixed offset in the heap, rather than having to first look up the offset in the heap.

To support hot module reloading, a follow up RFC could specify that dynamic template replacement is supported in development mode only, by putting the VM in a less optimized mode.

@kellyselden:

> One of the big selling points of Ember is being able to jump into a random project and know where everything is and how everything works because of the common conventions.
> 
> Would you consider this a drawback of the proposal, potentially encouraging drastically different component experiments?

Yes, I tried to capture this in the Drawbacks section.

In reality, the framework core team has been discussing template imports and SFC formats for some time, and we are interested in exploring those designs via an "official" addon. The benefit of stabilizing the primitives first is that, like with Sparkles/Glimmer components, we can iterate quickly on an "official" addon outside of the core framework, and that addon will be compatible with a wide range of Ember versions. The primitives also make it possible for the entire community to explore the solution space in parallel and contribute back good ideas.

But, yes, as with component manager and Glimmer components, having the ability to experiment and extend is important, but Ember's values mean we eventually need a recommended default.

@gossi 

> I like the experimentational idea of this RFC. Like "hey community, here is your toolbelt, now let's see what SFC you'll come up with and what might be the final spec we can compromise and align on".
> However, I also do see that there will be plenty of SFC specs out there and ember can face a wild-wild-west situation here, which can end up problematic 😈. Each time creating a new component, you'll need to watch out for the correct addon to match your SFC-style.

Agreed, see above reply.

> I dunno if left out, but template, script and what about styles? Think about css modules or blocks. Do they need an API, too?

We've thought about this a bit, but considered it out of scope for this RFC. The key insight here is that it's important for the template to inherit its lexical scope from JavaScript, so the API has been designed around making that case possible. It's not obvious to me whether other kinds of files, like CSS, benefit from being in the same scope as JavaScript, and if that is even a coherent thing to express given the differences between the two languages. We are definitely interested in exploring ways to make CSS more productive in Ember in the future.

@ef4:

> I would like us tp strongly consider keeping the template part as HBS within this new standardized format. Not precompiling it to the unstable wire format.

+1, will update the RFC to support this constraint.

@luxferresum:

> When doing `precompile` with a `scope`, how could I provide curly components? … Shouldn't we specify components and helpers in separate arrays? And if I want a component to be available for curly and angle-bracket style, do I need to specify it twice?

Names in the template's ambient scope must be valid JavaScript identifiers, so names with dashes are not supported. That said, explicit imports means that the previous naming constraints are no longer in place, so you can invoke components with curlies like this:

```js
import OtherComponent from './OtherComponent';

  <template>
    {{OtherComponent foo=bar}}
  </template>
}
```

That said, it is recommended that invocations of components be changed from curly to angle bracket syntax, simply for the visual disambiguation between components and helpers that it provides.

Components and helpers don't need to be provided separately because we can determine their type at runtime from the value provided. So long as it is given a valid JavaScript identifier, an imported component can be invoked via `{{OtherComponent}}` and `<OtherComponent />` in the same template just fine and does not need to be specified twice.

luxferresum:
> so names with dashes are not supported.
so no helpers with dashes? What is the reason for this limitation?

---

And shouldn't be `precompile` gets exported in `ember-source` so we can import it during build from node?

jenweber:
Is this content important enough that some subset of the explanations belong in the "advanced" section of the Ember CLI guides? There's a long way to go to fill in more basic info for addon authors, but perhaps a callout to the APIs would be appropriate. What do you think?

paxer:
What I really like in Ember is a clear separation of templates (HTML, hbs) and JavaScript, this concept allows our designers to work with the templates utilisiIg their HTML and css knowledge without involvement JavaScript engineers most of the time. Our engineering team do not write HTML/css at all, it is all up to designers. So I am not a big fan of the concept which React use, it is a real issue for us because designers don’t want to mess with JSX.  VueJS single file component looks a lot better in terms of clear separation between templates and JS, however I think Ember conventions solve the problem where to find a template and with the new UI file structure I think it will be even easier since both as I remember will be in the same folder. So I hope if Ember still go single file component way - it will be more Vue-like and not React.

sukima:
> What I really like in Ember is a clear separation of templates (HTML, hbs) and JavaScript, this concept allows our designers to work with the templates utilisiIg their HTML and css knowledge without involvement JavaScript engineers most of the time. 

As much as I personally agreed with this. The devil’s advocate is that the separation is the core reason Ember gets so much apathy in the larger JS communities. The common theme I see when asking React developers and React shops is the illusion that cowboy coding and unstructured spaghetti code is supirior in most cases. Many devs seem to prefer JSX or single component files in Vue.

Again as much as I agree with Ember’s conventions I worry about how competitive it is in a market which glorifies low level code. Could exposing more flexibility in lower level primitives offer more interest to the larger JS communities?

davewasmer:
@sukima @paxer to your concerns: I think it's important to keep in mind that this RFC is merely proposing that we _unlock_ the _possibility_ to experiment. Nothing here says that single-file components are necessarily the way of the future for Ember. All this does is give us a chance to iterate and try to discover potential solutions (including ones that might continue to allow designers to jump into the templates in some fashion).

text/000-sfc-and-template-import-primitives.md
```diff
@@ -0,0 +1,679 @@
...
+<script>
+export default {
+  data() {
+    return {
+      name: 'World'
+    }
+  }
+}
+</script>
+```
```
kovalchik: Is it worth also calling out how Angular utilizes template literals for inline templates?

```js
    import { Component } from '@angular/core';
     
    @Component({
      selector: 'app-root',
      template: `
        <h1>Hello {{name}}</h1>
      `
    })
    export class AppComponent {
      name = 'World';
    }
```

text/000-sfc-and-template-import-primitives.md
```diff
@@ -0,0 +1,679 @@
...
+import Component from '@glimmer/component';
+import { inject as service } from '@ember/service';
+import titleize from './helpers/titleize';
+import BlogPost from './components/blog-post';
+
+export default class MyComponent extends Component {
+  @service session;
+}
+</script>
+```
```
kovalchik: I think it might be worth adding an example which shows how a syntax could facilitate both inline and separate files. Although this RFC is only intended to facilitate experimentation, separation of concerns has long been a tenant of Ember's and is an important use case for many users. An example showing how such an API could unlock easy inline templates, while not excluding dedicated template files, would probably go a long way to making people feel more comfortable about exploring this direction.

#### For example (this is completely contrived, but hopefully it gets the point across):

`my-component/component.js`
```js
import titleize from './helpers/titleize';
import BlogPost from './components/blog-post';

  @service session;

  template() {
    return `
      {{#let this.session.currentUser as |user|}}
        <BlogPost @title={{titleize @model.title}} @body={{@model.body}} @author={{user}} />
      {{/let}}
    `;
  };
}
```

#### Could also be written as:

`my-component/template.hbs`
```hbs
{{#let this.session.currentUser as |user|}}
  <BlogPost @title={{titleize @model.title}} @body={{@model.body}} @author={{user}} />
{{/let}}
```

`my-component/component.js`
```js
import titleize from './helpers/titleize';
import BlogPost from './components/blog-post';
import Template from './template.hbs';

  @service session;
  template() { return Template };
}
```

I'm aware there's some problems with the second example, such as `BlogPost` and `titleize` not being utilized within the component directly, problems which this RFC aims to help address by opening up experimentation. However, I think it would be helpful to demonstrate that such experimentation doesn't necessarily lead us to a place where we can't support both inline and separate file templates.

pzuraq: FWIW I believe that these would be the corresponding examples for the variations. I do agree having them be available in this RFC would be nice, since this is also about how to move forward with importing in non-SFC components in general

### Frontmatter

`my-component/template.hbs`

```hbs
---
import titleize from './helpers/titleize';
import BlogPost from './components/blog-post';
---

{{#let this.session.currentUser as |user|}}
  <BlogPost @title={{titleize @model.title}} @body={{@model.body}} @author={{user}} />
{{/let}}
```

`my-component/component.js`
```js

  @service session;
}
```

### Inline

`my-component/template.js`
```js
import titleize from './helpers/titleize';
import BlogPost from './components/blog-post';

<template>
  {{#let this.session.currentUser as |user|}}
    <BlogPost @title={{titleize @model.title}} @body={{@model.body}} @author={{user}} />
  {{/let}}
</template>
```

`my-component/component.js`
```js

  @service session;
}
```

### Vue Style

`my-component/template.hbs`
```hbs
<script type="module">
  import titleize from './helpers/titleize';
  import BlogPost from './components/blog-post';
</script>
<template>
  {{#let this.session.currentUser as |user|}}
    <BlogPost @title={{titleize @model.title}} @body={{@model.body}} @author={{user}} />
  {{/let}}
</template>
```

`my-component/component.js`
```js

  @service session;
}
```

chriskrycho:
The updates made here address my concerns. 👍 👏

text/000-sfc-and-template-import-primitives.md
kovalchik: Thanks for the better examples @pzuraq!

text/000-sfc-and-template-import-primitives.md
```diff
@@ -0,0 +1,875 @@
...
+
+export default {
+  components: {
+    OtherComponent
+  }
+}
+</script>
+```
+
+Neither of these approaches is exactly right for Ember. Ideally, we'd find a
```
joefraley: Why are neither of these exactly right for Ember? Do you mean that neither of them is exactly aligned with your ideal solution in a general sense? Or do you mean that there is something specifically different about Ember that renders the Vue and React strategies undesirable?

text/000-sfc-and-template-import-primitives.md
```diff
@@ -0,0 +1,875 @@
...
+   backing class).
+2. Arguments (like `@firstName`) are placed into scope when you invoke the
+   component.
+3. Block parameters (like `item` in `{{#each @items as |item|}}`) are in
+   scope, but only inside the block.
+4. Anything else, like components and helper names, must be looked up
+   dynamically through Ember's container. This lookup process is somewhat
+   complex and not always as fast as we'd like.
+
+Because the rules for how components and helpers are looked up are implicit
```
joefraley: I find this ember's least attractive property. I'm glad it's called out explicitly here

text/000-sfc-and-template-import-primitives.md
```diff
@@ -0,0 +1,875 @@
...
+
+```js
+import { t } from '../helpers/t';
+import User from './user';
+import { templateFactory } from '@ember/template-factory';
+
+const template$1 = templateFactory({
+  "id": "ANJ73B7b",
+  "block": "{\"statements\":[\"...\"]}",
+  "scope": () => [User, t]
```
joefraley: are there memory implications of maintaining explicit references to these values? I don't know, I only mean this is explicitly different from ember's current architecture. consider a page with hundreds or thousands of the same component (like a table cell, for example). will that page have noticeably different memory constraints on this RFC than it does in ember today?

MelSumner:
> but Ember's values mean we eventually need a recommended default.

What would prevent us from starting out with a recommended default?

text/000-sfc-and-template-import-primitives.md
```diff
@@ -0,0 +1,875 @@
...
+
+Although exploring these features is the primary motivation, an important
+benefit of stabilizing low-level APIs is that they enable the Ember community
+to experiment with new, unexpected ideas.
+
+### Single-File Components
+
+In Ember components today, JavaScript code lives in a `.js` file and template
+code lives in a separate `.hbs` file. Juggling between these two files adds
+friction to the developer experience.
```
MelSumner: > Juggling between these two files adds friction to the developer experience.

Can we word this to be more relative? It is not true for everyone.

sukima: I usually split my editor window to show the JS on the top and HBS on the bottom. I don't experience this _Juggling_ problem. Is it possible that the tools a developer uses contributes more to the problem then the file structure itself?

sukima: If both the JS and HBS were to be in one file I'd have to split the file anyway and it would make the hot swap more difficult as I'd have to make two parts of a file and jump between them. Not only that but having two syntaxs and linting rules and compilers in one file is a massive undertaking in tooling and comprehension. I like that I can see a `.js` file and know it is JavaScript. I don't have to worry it is a Frankenstein experiment of non-JavaScript. I can `ember-template-lint *.hbs` and not have to also conciser it needs to be integrated into eslint. Doing any kind of find  find or `git grep call-site -- *.hbs` would compound the problem. I could go on but I hope my point is made. JSX is a fanatical nightmare in maintenance, usability, and comprehension. It may look nice for a really small module but I'd ask since when have all your components been that small? Real world apps are huge and complicated and we as devs need all the help we can get making sense of it all--computer tools too.

chriskrycho: I think @MelSumner's note on this is probably the most apt: developer experiences around this are *mixed*. I have lots of friends who simply don't experience the kind of "nightmare in maintenance, usability, and comprehension" that @sukima does – and I absolutely believe *both* reports.

These kinds of things tend to be very specific to individual tastes (in much the same way that many people love each of Python and Ruby but very few people love both), and they are also profoundly impacted by the patterns used in specific apps and teams.

I also think it's worth remembering that the point here is to *unlock different approaches*, which may allow us to find something that works *better* than today's story *or* the alternatives currently used in other frameworks. Regardless, Ember will certainly continue to have a *good default*.

By way of analogy: we have first-tier supported experience with Ember Data—but don't in the least discourage people from using Apollo.

chriskrycho:
@MelSumner I'd suggest we *have* a recommended default already: the current JS + HBS files situation. By the same token, we don't yet *need* a recommended default for SFC-style approaches, because the current default will continue to work well unless or until we find something generally recognized to be better.

luxferresum:
After exploring the component managers I've got curious why this API is not added to the component managers but completely separate? Wouldn't it be nicer to ask the component manager for the template? We currently ask the component manager for the `this` context, why not also ask it for the template (some kind of embeddable template) and the scope?

text/000-sfc-and-template-import-primitives.md
```diff
@@ -0,0 +1,875 @@
...
+
+const template$1 = templateFactory({
+  "id": "ANJ73B7b",
+  "block": "{\"statements\":[\"...\"]}",
+  "scope": () => [MissingComponent]
+});           /*  ^------- will produce a ReferenceError */
+```
+
+As of this RFC, scope values resolved via ambient scope are limited to
+component classes and template helpers (i.e. subclasses of `Helper` or
```
luxferresum: did we miss modifiers here? 🤔

text/000-sfc-and-template-import-primitives.md
MelSumner: only if we're using it as an example of something we'd never do...In all seriousness though, It may be better to leave it out.

thec0keman:
This may be out of scope for this RFC, but it would be great if this set of API's included a way to plug in a template preprocessor such as Emblem.  Emblem currently plugs into Ember CLI hooks and converts its `.embl` files into `.hbs` before handlebars compilation kicks in.  If individual addons are now controlling template compilation then it seems like this won't be possible.  (At least without adding customization to each addon?)

ef4:
@thec0keman I don't think this precludes template preprocessors. The format described here is what those preprocessors would target as output.

knownasilya:
Anyone know when this will enter FCP?

rwjblue:
This RFC replaces #204.

text/000-sfc-and-template-import-primitives.md
```diff
@@ -0,0 +1,875 @@
...
+Or imagine this syntax, where a component's template is located syntactically
+within the class body:
+
+```js
+import Component from '@glimmer/component';
+import { inject as service } from '@ember/service';
+import titleize from './helpers/titleize';
+import BlogPost from './components/blog-post';
+
+export default class MyComponent extends Component {
```
neborn: One of things that I really like about this example, and one of the things that I think/hope the SFC feature will enable is the ability to define multiple components within a single file. (I know, already consolidating from two files to one, and now I am asking to be able to consolidate from n files to one.)

chadhietala: Yea, I want this world too. I'm actually a bit more excited about the possibility to have multiple helpers along side components. I think helpers are under utilized because the process of creating a new one is so heavy.

joefraley:
> Neither of these approaches is exactly right for Ember. Ideally, we'd find a way to "bend the curve" of tradeoffs: maintaining the performance benefits of templates, while gaining productivity and learnability by having those templates seamlessly participate in JavaScript's scoping rules.

i would trade any degree of performance for any degree of productivity/transparency. i would pay 1000% performance slow down for a 1% improvement in legibility/ease of use. i have almost never been in a situation in my career where 2x or 5x performance was make-or-break, but almost every codebase i have ever worked in/organization i have ever been part of has been brutalized by onboarding/general productivity issues

joefraley:
> Again as much as I agree with Ember’s conventions I worry about how competitive it is in a market which glorifies low level code. Could exposing more flexibility in lower level primitives offer more interest to the larger JS communities?

why does this matter? is it important for ember to be competitive in adoption with other approaches? i would naïvely think that the only important thing about a tool is whether it helps people accomplish their aims. if more people have other tools or other aims than there are people that find ember useful for their aims, does it matter? i appreciate that more popular tools have intrinsically broader and more active communities, which one can expect to furnish more and better solutions to a wider variety of problems (because numbers). 

at the same time, i rarely see people talk this way about the design of elm-lang or cycle.js. the attitude of those communities seems largely to be "we'll make tools that solve our problems and we won't worry about the rest". is the community growth stuff the shortcoming with that attitude, or is it also something different? why should market adoption bear on technical guidance?

for example, ember could be great for a small number of very specific people. those people might not be people attracted to using react. should converting-react-users be part of technical design of ember?

fwiw i would not choose to use ember for some of the reasons alluded to here and others, but having explicit control over template scope would be helpful to me in making good use of ember where i must.

pzuraq:
> i would trade any degree of performance for any degree of productivity/transparency. i would pay 1000% performance slow down for a 1% improvement in legibility/ease of use

That may be true for some applications, but regressing in performance too much has actually impacted applications in the community (Discourse, for instance, couldn't upgrade to Ember 2 for quite some time because of the 1.13 performance regressions). Performance does become a concern for many applications as they grow, and templates are one area we can leverage a performance advantage as it stands, so it makes sense to continue to try to keep that advantage.

joefraley:
> > i would trade any degree of performance for any degree of productivity/transparency. i would pay 1000% performance slow down for a 1% improvement in legibility/ease of use
> 
> That may be true for some applications, but regressing in performance too much has actually impacted applications in the community (Discourse, for instance, couldn't upgrade to Ember 2 for quite some time because of the 1.13 performance regressions). Performance does become a concern for many applications as they grow, and templates are one area we can leverage a performance advantage as it stands, so it makes sense to continue to try to keep that advantage.

I think that is totally fair and valid. I don't mean to say that performance is never important or valuable to anyone. I only meant that I have never valued even large performance gains against even marginal productivity gains.

my fundamental point is: "it makes sense to continue to try to keep that advantage." is reasonable, and i gladly accept performance gains _all things being equal_. but i do not want to buy it if at prices labeled in productivity. in the same way, i would not trade security or privacy for faster web experiences. i would rather wait 20s for the thing to load than 200ms at the expense of vulnerabilities.

i'm saying that i think balancing performance against productivity is a zero-sum losing game.

pzuraq:
Totally get your point. I think the main thing @tomdale is pointing out here is that we _can_ bend the curve here, because the most valuable benefits from a DX perspective, template imports and being able to use normal JS scoping rules, do not have a major impact on performance. We can do this and maintain all of our benefits from template syntax, without needing a custom re-export process like in Vue or other frameworks. This is the low hanging fruit.

Where it gets harder is if we wanted to go further down the path of JSX, and have a much more dynamic templating syntax that can interleave with JavaScript seamlessly. There are advantages to it, but there are disadvantages too, and it would be a much larger lift to get there even if we decided those advantages were worth it.

sdhull:
Is this the sort of thing (SFC) that could be enabled by an addon so we can experiment with it? Google [surfaced this](https://github.com/urbany/ember-cli-single-file-components), but it's 2 years old 🤔

MelSumner:
> i would trade any degree of performance for any degree of productivity/transparency.

I don't think we should do things that put developers before end users.

lifeart:
This is how we have `jsx` support https://github.com/tomdale/glimmer-lite-donut/pull/3

lifeart:
I think we can adopt `svelte` parser https://astexplorer.net/#/gist/f3a958bf1aec337fa6c3e7636ce578d4/809ae5f1ca13fa50a4a3e3a19373e2fddcfc610a

ming-codes:
JavaScript as template compilation target also paves way for typed template. My concern in this is that how are we going to embed type signatures when the template is compiled into byte code. Maybe a different wire format for non-production builds?

lifeart:
@ming-codes, see https://github.com/lifeart/els-addon-typed-templates

Devoter:
What do you think about this type of single-file components?
```js
import { hbs } from 'ember-cli-htmlbars';

const layout = hbs`
<div>
  <span>Hello, Ember's world!</span>
</div>
`;

  constructor(layout = layout) {
    super(...arguments);
  }
}
```

Sure, `layout` property should be supported by `@glimmer/component` and the components manager.

pzuraq:
@Devoter there are two issues with that exact design:

1. The constructor parameters for Glimmer components are already defined and public API, so we can't change them as is

2. It would be fairly misleading to users, since templates _cannot_ be dynamic. They _must_ be defined statically, ahead of time. Passing it via a constructor argument implies that it can change per-instance of the component.

gabrielcsapo:
Any movement on FCP? I think this is going to add a lot of value for developers and the community at large.

pzuraq:
I believe this RFC is superceded by #496, which is under active development. The primitives provided by #496, along with the existing `setComponentTemplate` API, would provide the same functionality as this RFC proposes.

Gaurav0:
> Where it gets harder is if we wanted to go further down the path of JSX, and have a much more dynamic templating syntax that can interleave with JavaScript seamlessly. There are advantages to it, but there are disadvantages too, and it would be a much larger lift to get there even if we decided those advantages were worth it.

I would prefer we did not do this. It would further introduce the ability to do arbitrary non performant things with arbitrary side effects during render.

pzuraq:
It would not be possible to have the same level of dynamism as JSX with the Glimmer VM. It would require a massive rewrite of the rendering engine, and I don't think it's anywhere near in scope for this feature.

sukima:
I'm curious what the drive to merge languages are. Templates are a LISP language, why not lean on LISP for this and import as a procedure

```hbs
{{import Foo from "…"}}
```

I think one of the bigist reasons I really don't like JSX is that it mixes contexts. My JS mingles with my HTML in a weird-have-to-understand-the-implementation way.

Is it possible to have template imports inside the `<template>`-context-switching addition to the JavaScript spec?

NullVoxPopuli:
> {{import Foo from "…"}}

I'd like to see something like that implemented  -- even if just for comparison. Implementation wise, I have a *hunch* it's going to be a similar type of transform as gjs :thinking:

pzuraq:
> Templates are a LISP language, why not lean on LISP for this and import as a procedure

This actually goes back a bit in the decision making process, all the way to Module Unification. We actually _were_ planning on doing that, it just looked very different:

```hbs
<some-addon-name::some-component>
</some-addon-name::some-component>
```

A few things changed after this syntax was accepted:

1. We decided to make `<AngleBracket/>` invocation capital case. This was not fatal on its own, but it would mean there was an awkward mix between snake-case package names and CapitalCase component names, _or_ that we would have to translate addon names to capital case as well. Then...
2. NPM added named scopes to package names, making them much longer, and conceptually colliding with `@` in args:

The result would have been something like

```hbs
<@my-scope/some-addon-name::SomeComponent>
</@my-scope/some-addon-name::SomeComponent>
```

This was clearly not going to work, it was way too verbose overall. So, we went back to the drawing board.

The first draft actually _was_ something similar to what you proposed with `{{import Foo from "…"}}`. However, there was immediately an issue: how do handle named imports?

```hbs
{{import DefaultValue, ( Foo as bar, baz ) from "…"}}
```

Every combination we came up with felt wrong in some way. Then, another point was made: The whole reason we were having to make this decision in the first place was because the wider ecosystem - in this case NPM - decided to make a fundamental change to the way that imports worked. It's is 100% possible that other changes could happen in the future that could _also_ collide with this new custom import syntax. Changes such as additional import features, changes to build tools that would make it impossible for us to use a custom import syntax for bundling, etc.

This is why in the end we decided that we would use JavaScript import syntax for imports. In the end, developing a custom, parallel syntax would require a lot of work, and it would _always_ have the potential to conflict with future changes in the wider ecosystem. This had already happened to us once, and we decided that we didn't want it to happen again.

> I think one of the biggest reasons I really don't like JSX is that it mixes contexts. My JS mingles with my HTML in a weird-have-to-understand-the-implementation way.

FWIW, I empathize with this concern, as I actually feel this pain today with the current split between template files and JavaScript. For me, having to context switch constantly between two different files is really hard on my development flow, and I find it really hard to reason about what's going on. 

I think everyone is a bit different here, and while I've definitely heard from plenty of folks who prefer the separation, I have also heard from just as many, if not more, people who would prefer to have the component and template in the same file. So I think whatever the standard/default setting is, we may not end up satisfying everyone.

What I keep coming back to is tests. Today, we write templates in tests, and I believe we want to keep doing that as a framework:

```js
test('MyComponent works', function() {
  await render(hbs`
    <MyComponent/>
  `);
});

test('MyComponent works with args', function () {
  await render(hbs`
    <MyComponent @arg={{this.foo}}/>
  `);
});
```

In order for this work with template imports, we either need to repeat the imports in every single template:

```js
test('MyComponent works', function() {
  await render(hbs`
    ---
    import MyComponent from './my-component';
    ---
  
    <MyComponent/>
  `);
});

test('MyComponent works with args', function () {
  await render(hbs`
    ---
    import MyComponent from './my-component';
    ---
   
    <MyComponent @arg={{this.foo}}/>
  `);
});
```

Or we need to say that the component has access to the outer JS scope:

```js
import MyComponent from './my-component';

test('MyComponent works', function() {
  await render(hbs`
    <MyComponent/>
  `);
});

test('MyComponent works with args', function () {
  await render(hbs`
    <MyComponent @arg={{this.foo}}/>
  `);
});
```

Out of these two options, the later seems much better. It also means we can finally solve the problem of testing contexts with local tracked state, and you can add local helpers as well:

```js
test('MyComponent works with args', () => {
  let state = new TrackedObject({ a: 1, b: 2 });

  let add = helper(([a, b]) => a + b);

  await render(hbs`
    <MyComponent @arg={{add state.a state.b}}/>
  `);

  state.a = 4;

  await settled();
});
```

These features would definitely take some getting used to, but to me this feels like it would really help with the expressiveness of tests and the ability to write great tests. We are actually the only framework besides React I think that writes templates in tests like this, so I think it's good to maximize our capabilities and leverage here 😄 

So if we went that direction in tests, the question then becomes: why have two ways of doing the same thing? The learning curve of two things is always higher than one, as is the infrastructure to be built, etc.

So yeah, this is where I am currently. This is not the state of the core team as a whole, there is no consensus yet, but personally I find these capabilities really motivating. Having one unifying story for writing templates in tests and app-code seems like it would really reduce tooling cost and learning cost, and help us to accelerate our development and progress as a community. Definitely open to more discussion here, and proposals for alternative syntaxes and solutions, etc (feel free to PR them to `ember-template-imports`!).

knownasilya:
Have written an app that used `{{import ...}}` syntax exclusively, I can whole heartedly agree on having one way to do imports, and the confusion of `{{import ...}}` in some of the situations mentioned.

text/000-sfc-and-template-import-primitives.md
fredericrous: @MelSumner I would argue that this is true for everyone that only has one screen or is working from a laptop (outside, at a coffee shop, or traveling).
Grouping css, html & js together would significantly reduce the amount of files I have opened. My screen would still need to be split to display either the child or parent component

sdhull: One analogy that comes to mind here is _unnecessary abstractions_. If you've ever traced a method call through 2 or 3 or 10 other methods that are all a single-line and thought to yourself "y tho?" you know what I'm talking about.

When a component could fit in its entirety (js + template) on a single (laptop) screen, opening two files, I find myself thinking something similar: "y tho?" The more you split functionality among different files (don't forget CSS!) the more you have to go hunting to get a complete picture of said functionality. And the more you risk inadvertently breaking something when you make changes. Yes, there's a bit of a learning curve here (iykyk) and once you get over the hump, it's not as much of a pain point. But there are certainly circumstances where it's a tool I'd reach for were it available.

@MelSumner I think @chriskrycho's comment is excellent wording:
> the point here is to unlock different approaches, which may allow us to find something that works better than today's story or the alternatives currently used in other frameworks

One use-case that comes to mind is platforms. If I'm building a platform (RIP Boundless Labs 😢), it's more ergonomic (IMO) to ask 3rd party developers to submit components as a single file.

Also, FWIW @fredericrous I am personally one of the rare(?) laptop workers, regardless of where I'm working from.

chriskrycho:
Should this be merged? We shipped the primitives a *long* time ago, with roughly (albeit slightly tweaked) versions of the API described here, and `ember-template-imports` *uses* them to implement user-facing features! Also, given I’m in the middle of writing an RFC proposing we stabilize on one specific format, it’d be good if we did something with the RFC for the primitives it’s built on!

ef4:
The primitives in this RFC got broken out separately: 

 - setComponentTemplate landed via https://github.com/emberjs/rfcs/blob/master/text/0481-component-templates-co-location.md
 - precompileTemplate landed via https://github.com/emberjs/rfcs/blob/master/text/0496-handlebars-strict-mode.md

Together those achieve the thing this RFC was trying to do.

PR #460, opened 2019-03-07
Yieldable named blocks

wycats:
[Rendered](https://github.com/emberjs/rfcs/blob/2f4c4e2135173b2f8bec72a570a9986a69ef8edd/text/0460-yieldable-named-blocks.md)

### DIFF
--- a/text/0460-yieldable-named-blocks.md
+++ b/text/0460-yieldable-named-blocks.md
@@ -1,0 +1,300 @@
+- Start Date: 2019-03-06
+- Relevant Team(s): Ember.js
+- RFC PR: [#460](https://github.com/emberjs/rfcs/pull/460)
+- Tracking: (leave this empty)
+
+# Yieldable Named Blocks
+
+## Summary
+
+This RFC amends [#226 (named blocks)][named-blocks] to finalize the syntax of named blocks and reduce the scope of the feature.
+
+[named-blocks]: https://emberjs.github.io/rfcs/0226-named-blocks.html
+
+## Motivation
+
+The original named blocks RFC (#226) defined a mechanism for passing multiple blocks to a component, as well as semantics for capturing the block in JavaScript.
+
+Since #226, RFC [#311 (angle bracket invocation)][angle-bracket] repurposed the original syntax for invocation.
+
+```hbs
+<Tab>
+  <@header></@header>
+</Tab>
+```
+
+In #226, the above syntax passed a named block called `@header` to the `Tab` component. RFC #311 repurposed that syntax to mean "invoke the component located at `@header` in the default block of `Tab`".
+
+Curly bracket components already support two blocks (the default block and the `else` block), while angle bracket components, as implemented today in Ember, only support a single default block.
+
+While RFC #226 predates a formal RFC about angle bracket components, it speculates significantly about usage in angle bracket components, and is heavily motivated by the desire to support more than a single block in that context.
+
+That said, #226 included significant additional scope beyond allowing angle bracket invocations to pass more than one block, including:
+
+- using `<@name>` as a syntax for passing named blocks to curly components
+- passing named blocks as `@name` variables, accessible in the receiving component as `@blockName`
+- capturing named blocks in JavaScript using `this.blockName`, as well as capturing named blocks in JavaScript by forwarding them into JavaScript as a parameter to a helper or another named argument.
+
+While versions of those features are all desirable, the RFC, in total, had a fairly large scope and also speculated a lot about the future of angle bracket components which would not be approved until RFC #311.
+
+This RFC proposes a version of named blocks with a smaller scope, with a plan to reintroduce the other goals of the original named blocks RFC in a subsequent RFC.
+
+> Note: The version of named blocks described in this RFC has already been implemented in Glimmer VM.
+
+[angle-bracket]: https://emberjs.github.io/rfcs/0311-angle-bracket-invocation.html
+
+## Detailed design
+
+> The detailed design of this RFC assumes a starting point without RFC #226, and describes the functionality in terms of RFC #311.
+
+When invoking a component with angle bracket syntax, the content between the starting tag and the ending tag is passed as a block to the receiving component.
+
+This means that the receiving component can invoke the block using `{{yield}}`. The block is not exposed to JavaScript at all, and it may only be invoked in the body of the receiving component's template.
+
+On their own, blocks provide useful functionality. For example, consider an `Article` component that takes a `@body` argument:
+
+```hbs
+<Article @body={{this.body}} />
+```
+
+The `Article` component could be implemented this way:
+
+```hbs
+<article>
+  <section>{{@body}}</section>
+</article>
+```
+
+This works, but what if you want to include HTML functionality or use helpers. In that case, you could use a block:
+
+```hbs
+<Article>
+  <div class='byline'>{{byline this.author}}</div>
+  <div class='body'>{{this.body}}</div>
+</Article>
+```
+
+The `Article` component could then be implemented by using `{{yield}}` to invoke the block:
+
+```hbs
+<article>
+  <section>{{yield}}</section>
+</article>
+```
+
+But what if we also want the ability to pass a title to the component. We could use a `@title` argument:
+
+```hbs
+<Article @title={{this.title}}>
+  <div class='byline'>{{byline this.author}}</div>
+  <div class='body'>{{this.body}}</div>
+</Article>
+```
+
+Implemented this way:
+
+```hbs
+<article>
+  <header>{{@title}}</header>
+  <section>{{yield}}</section>
+</article>
+```
+
+Again, that works, but if you want to use any of the block features, such as the ability to use HTML, you can't pass a second block.
+
+This RFC introduces the ability to pass any number of blocks to a component. With that capability, you could call `Article` this way:
+
+```hbs
+<Article>
+  <:title>
+    <h1>{{this.title}}</h1>
+  </:title>
+  <:body>
+    <div class='byline'>{{byline this.author}}</div>
+    <div class='body'>{{this.body}}</div>
+  </:body>
+</Article>
+```
+
+And implement it this way:
+
+```hbs
+<article>
+  <header>{{yield to='title'}}</header>
+  <section>{{yield to='body'}}</section>
+</article>
+```
+
+### Block Parameters
+
+Just like normal blocks, named blocks can take block parameter, which are passed via `yield`.
+
+```hbs
+<Article @article={{this.article}}>
+  <:header as |title|>
+    <h1>{{title}}</h1>
+  </:header>
+  <:body as |body|>
+    <div>{{body}}</div>
+  </:body>
+</Article>
+```
+
+```hbs
+<article>
+  <header>{{yield @article.title to='header'}}</header>
+  <section>{{yield @article.body to='body'}}</section>
+</article>
+```
+
+To summarize, today's Ember allows all components to pass a default block, and curly components to pass a default block and an `inverse` block. Component templates can invoke default templates by using `{{yield}}` and `else` templates by using `{{yield to='inverse'}}`.
+
+This RFC introduces a new syntax for passing arbitrary blocks to angle bracket components that can be invoked by the receiving component using the same `yield` syntax.
+
+### Syntax
+
+This RFC proposes an extension to the angle bracket invocation syntax.
+
+From:
+
+```
+AngleBracketInvocation :
+  | AngleBracketWithBlock
+  | AngleBracketWithoutBlock
+
+AngleBracketWithBlock :
+  "<" ComponentTag ComponentArgs? BlockParams? ">"
+  Block
+  "</" ComponentTag ">"
+
+AngleBracketWithoutBlock :
+  "<" ComponentTag ComponentArgs? "/>"
+```
+
+To:
+
+```
+AngleBracketInvocation :
+  | AngleBracketWithBlocks
+  | AngleBracketWithBlock
+  | AngleBracketWithoutBlock
+
+AngleBracketWithBlock :
+  "<" ComponentTag ComponentArgs? BlockParams? ">"
+  BlockBody
+  "</" ComponentTag ">"
+
+AngleBracketWithBlocks :
+  "<" ComponentTag ComponentArgs? BlockParams? ">"
+  NamedBlock+
+  "</" ComponentTag ">"
+
+NamedBlock :
+  | "<:" Identifier "/>"
+  | "<:" Identifier BlockParams? ">" BlockBody "</:" Identifier ">"
+
+AngleBracketWithoutBlock :
+  "<" ComponentTag ComponentArgs? "/>"
+
+ComponentTag :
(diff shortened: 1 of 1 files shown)


rwjblue:
@wycats it is very exciting to see this feature moving forward, thank you for working on it!

I noticed that this RFC specifically does not make any changes to `yield`, which leaves one of the nice features in the named block RFC out: the concept of unified renderables. While I definitely understand the desire to cut scope in order to actually ship something, I think that this RFC should at least mention that the specific concept is either rejected completely or is ripe for a future RFC.

Specifically, as a component author I want to allow a caller to pass any number of things (plain string, closure component, or a named block). In order to do that (unless I've misunderstood things in this RFC) is via something akin to the following snippet in the component's template:

```hbs
<article>
  <header>
    {{#if (has-block 'title')}}
      {{yield to="title"}}
    {{else}}
      {{@title}}
    {{/if}}
  </header>
  <section>
    {{if (has-block 'body')}}
      {{yield to='body'}}
    {{else}}
      {{yield}}
    {{/if}}
  </section>
</article>
```

This allows the callsite to use any of the following invocations:

```hbs
<Article @title={{@article.title}}>
  {{! block contents here }}
</Article>
<Article @title={{component 'special-article-title'}}>
  {{! block contents here }}
</Article>
<Article>
  <:title><b>{{@article.title}}</b></:title>
  <:body>
    {{! block contents here }}
  </:body>
</Article>
```

While thats totally fine its a bit gnarly to type out each time, and since blocks themselves can't be passed (mentioned in the RFC as capturing) it can't be abstracted away into a shared control flow component. 

What is the plan for this? Just keep the verbosity here, and come up with some clever AST "macros"? 

I'm also curious how the longer term plan to be able to pass these blocks around would work, have you thought much about that? Since it seems that this design intentionally does **not** limit named blocks to exist in the same conceptual space as named arguments, naming conflicts are likely (see API I demo'ed just above :P ), that seems _vaguely_ concerning unless we have a relatively good idea how we plan to reference a passed in block (and that that isn't as a named argument).

wycats:
@rwjblue The concept of "unified renderables" as envisioned in #226 was described (and carefully specified) in more detail in #432, so I left it out of this RFC.

From the perspective of this RFC, if you have a component that takes a block for `title` and you want to pass a string, you would still use the block syntax to pass it:

```hbs
<Article>
  <:title>{{@article.title}}</:title>
  <:body>
    {{! block contents here }}
  </:body>
</Article>
```

It's slightly more verbose than `@title=`, but only slightly, and it maintains a consistent mental model for what you're passing along.

It also makes upgrading from a simple block to a block that takes named arguments easier to teach, implement (and use), which was a thorny aspect of the original proposal.

That said, when we do decide on a strategy for reifying blocks in a future RFC, we will have to decide what happens when you put one inside of `{{...}}`. I can easily believe that we would make that work, and I think it's a good topic of discussion for that RFC.

wycats:
> I'm also curious how the longer term plan to be able to pass these blocks around would work, have you thought much about that? Since it seems that this design intentionally does **not** limit named blocks to exist in the same conceptual space as named arguments, naming conflicts are likely (see API I demo'ed just above :P ), that seems _vaguely_ concerning unless we have a relatively good idea how we plan to reference a passed in block (and that that isn't as a named argument).

My thinking is that we'd use `:foo` (not inside of a string) as the name for reification.

```hbs
<Article :title=:headline />
```

In this example, the `:headline` block in the current template would be passed to `Article` as the `:title` block. This is purely speculative, but it's an example of how we could maintain the namespace separation going forward.

rwjblue:
> It also makes upgrading from a simple block to a block that takes named arguments easier to teach, implement (and use), which was a thorny aspect of the original proposal.

Sure, I see what you mean here but until the broader ecosystem can assume newer Ember versions (which support this new named block syntax) addon components would have to continue to use the snippet I demoed above (which kinda sucks). It would be nice to have something nicer for them to use, but again maybe we could do it via an AST transform ala a macro.

pzuraq:
Really like this proposal overall! I think this is well-scoped, and minimal, without blocking us from iterating on it in the future. Most of my concerns are for future functionality actually, I just want to note them down here so they're available when we get around to it:

* I like the idea of being able to assign blocks using `:foo=`, but I don't think it makes much sense to be able to _refer_ to blocks within a template with `:foo` on the right hand side. If we compare to named arguments, it's inconsistent:

  ```hbs
  <!-- @bar refers to the argument -->
  <div @foo={{@bar}}></div>

  <!-- @bar refers to the argument, and `<>` invokes the argument as a renderable -->
  <@bar></@bar>

  <!-- :bar refers to the argument -->
  <div :foo=:bar></div>

  <!-- :bar captures a block -->
  <:bar></:bar>
  ```

  I think it could get very confusing when sometimes the identifier is used to refer to the block value, and sometimes it is used to capture/assign the value. Instead, I think we could use the `yield` helper, and allow it to be used inline:

  ```hbs
  <div @foo={{yield to="bar"}}></div>
  ```

  This means that invocation and capturing are always consistent.

* In designing Ember Table, one use case that came up consistently was wanting named blocks, and wanting them to be able to receive their own arguments:

  ```hbs
  <EmberTable as |t|>
    <t.head @columns={{columns}} />

    <t.body @rows={{rows}} />
  </EmberTable>
  ```

  I think that this will also be something we want if/when we get to block capture, but it's definitely outside of the scope of this RFC. It's also something that other frameworks can accomplish today, and I think would be very convenient.

  One possibility would be if blocks were captured on something like `this.blocks`, then we could refer to block arguments via `this.blocks.NAME.args`:

  ```hbs
  <EmberTable>
    <:head @columns={{columns}}/>
    <:body @rows={{rows}}/>
  </EmberTable>
  ```

  ```js
  this.blocks.head.args.columns;
  this.blocks.body.args.rows;
  ```

  It's a little verbose, but not too bad - I'd personally be ok with that style of access. I just want to make sure we don't foreclose on the possibility of adding block arguments down the road, since like I said, I think they'd be very useful (along with generic block capturing)

rtablada:
@wycats really exciting to see this feature moving forward and coming to fruition.

While after reading for a bit I was able to parse the example in the "block parameters" section it was a bit unclear what was the actual result. On first glance it looked like the positional params were fallbacks if the block was not invoked.

It took some head scratching to put together that it worked like normal positional params to the `yield` helper.

I'm not sure how to best improve this (maybe different variable/param names since `title`/`header`/`body` get repetitive and possibly confusing?).

rtablada:
Playing around with the HBS a little I came up with this possible example (slightly modified from the existing example:

```hbs
<Article @article={{this.article}}>
  <:header as |title author|>
    <h1>{{title}} by {{author.firstName}}</h1>
  </:header>
  <:body as |post|>
    <div>{{post.body}}</div>
  </:body>
</Article>

<article>
  <header>{{yield @article.title @article.author to='header'}}</header>
  <section>{{yield @article.post to='body'}}</section>
</article>
```

This not only decouples some of the more repetitive naming (especially `@article.body to="body"`) but it also shows yielding multiple values.

jgwhite:
Does this RFC propose that block names MUST be unique? i.e. would the following cause a compile-time error?

```hbs
<Article>
  <:byline>Alice</:byline>
  <:byline>Bob</byline>
</Article>
```

Exelord:
I think we should allow on duplications. It may be useful to display the same content in different places with different format.

Eg you may want to display article's created At in header and footer. So you will be able to configure one block to see the result in both places.

jgwhite:
@exelord interesting, could you provide a code snippet to go with that example?

Exelord:
Here it is:

```hbs
<div class="article">
  <div class="article-header">
    <h1>{{this.article.title}}</h1>
    <p>{{yield to="credits"}}</p>
  </div>

  <p>
    {{this.article.body}}
  </p>

  <div class="article-footer">
    <p>{{yield to="credits"}}</p>
  </div>
</div>

<Article @article={{this.article}}>
  <:credits>
    <i class="fa-book" />
    ~{{this.article.author}}
  </:credits>
</Article>
```

In the same way we are not blocking usage of the same property call twice in a template (eg you are able to call `{{this.article.body}}` multiple time) I dont see a reason why we should limit blocks calls.
___
The other importnat question (I think) that need to be clarified is block name format, as it not clear from the example.

eg, it is allowed to name blocks like `custom-block`, `custom_block`, `customBlock` and then call them respectively `<:custom-block>`, `<:custom_block>`, `<:customBlock>`?

From angle bracket convention I assume the corect syntax in that case is `{{yield to="customBlock"}}` and `<:customBlock>`. 
But what in case if someone will use `{{yield to="custom-block"}}`? Should the value be transformed or block will be not yielded properly?

Exelord:
And I think we should also consider restricting `else` block calls to point always to inverse
```hbs
<MyArticle>
  <:else>
    No article :(
  </else>
</MyArticle>
```
will equal
```hbs
{{#my-article}}
{{else}}
  No article :(
{{/my-article}}
```

jgwhite:
@exelord that makes sense, though I think we’re talking about different (but related) things. My comment refers to the presence of multiple blocks with the same name, whereas yours refers to yielding to a single block multiple times. Both worth clarifying, for sure.

rwjblue:
We discussed this in todays Ember, and we are moving this into final comment period!

rwjblue:
We discussed this in today's core team meeting, and are moving it into final comment period.

rwjblue:
OK, y'all it's that time, let's do this!

Kerrick:
Does this have a tracking issue yet?

rwjblue:
Yep, over in https://github.com/emberjs/rfc-tracking/issues/44

NullVoxPopuli:
It's been some time since this has landed in ember, and I'd say an absolute critical missing feature is the ability to forward blocks without needing to explore if/else has-block conditionals.

For example, say you are wrapping a component that has named blocks.
If the wrapped component has no markup then this is easy, you just:
```hbs
{{!-- wrapper.hbs --}}
<Wrapped>
  <:block-a>
    {{yield to="block-a"}}
  </:block-a>
  <:block-b as |any number of args|>
    {{yield any number of args to="block-b"}}
  </:block-b>
</Wrapped>
```
But now let's pretend the act of passing a block at all adds additional markup (which is handy in many cases!), but for your wrapper component you need to ensure that you don't force the rendering of that additional markup when the consumer of your wrapper component does not pass those named blocks -- the wrapper.hbs now has to look like this:
```hbs
{{!-- wrapper.hbs --}}
{{#if (has-block "block-a")}}
  <Wrapped>
    <:block-a>
      {{yield to="block-a"}}
    <:block-a>
  </Wrapped>
{{else if (has-block "block-b")}}
  <Wrapped>
    <:block-b as |any number of args|>
      {{yield any number of args to="block-b"}}
    </:block-b>
  </Wrapped>
{{else if (and (has-block 'block-a') (has-block 'block-b'))}}
  {{!-- NOTE: at the time of writing `and` is not a built in helper -- implementation is in progress tho --}}
  <Wrapped>
    <:block-a>
      {{yield to="block-a"}}
    </:block-a>
    <:block-b as |any number of args|>
      {{yield any number of args to="block-b"}}
    </:block-b>
  </Wrapped>
{{else}}
  <Wrapped />
{{/if  
```

This is _especially_ deadly to productivity as the number of named blocks increases (which is very common in layout-style components).

The number of conditionals you need is the square of the total number of blocks.

So, if you have a max of 6 named blocks, you need 36 if/else conditionals to properly wrap that component with only 5 named blocks.
Now imagine if your component also takes a number of arguments... Your template files becomes _huge_.

NullVoxPopuli:
maybe a `<Wrapped ...:blocks />` would be useful for this as well :)

NullVoxPopuli:
Here is an additional (maybe more concrete) demonstration and use case that came up with the need for this feature: 
[Using named blocks with component / dynamic content in i18n](https://limber.glimdown.com/?t=%60%60%60gjs+live+preview%0Aimport+%7B+compileHBS+%7D+from+%27ember-repl%27%3B%0Aimport+Component+from+%27%40glimmer%2Fcomponent%27%3B%0A%0Aconst+translations+%3D+%7B%0A++%27test-key%27%3A+%60before+%7Blabel%7D+text+%5B%5B%5Btext%5D%5D%5D+and+then+%5B%5B%5Blink%5D%5D%5D%60%2C%0A%7D%0A%0Aclass+T+extends+Component+%7B%0A++get+compileResult%28%29+%7B%0A++++const+translation+%3D+translations%5Bthis.args.key%5D%3B%0A%0A++++if+%28%21translation%29+return%3B%0A%0A++++%2F%2F+https%3A%2F%2Fgithub.com%2FRuslanZavacky%2Fember-intl-component-string%2Fblob%2Fmain%2Faddon%2Fcomponents%2Fi18n%2Findex.js%23L44%0A++++const+asYieldable+%3D+translation.replace%28%2F%5C%5B%5C%5B%5C%5B%28.*%3F%29%5C%5D%5C%5D%5C%5D%2Fg%2C+%27%7B%7Byield+to%3D%22%241%22%7D%7D%27%29%3B%0A++++%2F%2F+need+i18n+service+from+ember-intl+to+interpret+the+%60%7B%7D%60+placeholders%0A++++%2F%2F+translated+%3D+this.i18n.t%28asYieldable%2C+this.args%29%3B%0A++++%2F%2F+return+compileHBS%28translated%29%3B%0A++++return+compileHBS%28asYieldable%29%3B%0A++%7D%0A%0A++%3Ctemplate%3E%0A++++%7B%7B%23if+this.compileResult.component%7D%7D%0A++++++%7B%7Byield+this.compileResult.component%7D%7D%0A++++%7B%7Belse%7D%7D%0A++++++missing+translation%2C+likely%0A++++%7B%7B%2Fif%7D%7D%0A++%3C%2Ftemplate%3E%0A%7D%0A%0Aconst+wooo+%3D+%27hi%27%3B%0A%0Aexport+default%0A++%3Ctemplate%3E%0A++++%3CT+%40key%3D%22test-key%22+%40label%3D%22my+label%22+as+%7Cwrapper%7C%3E%0A++++++%3Cwrapper%3E%0A++++++++%3C%3Atext%3E%0A++++++++++some+text+that+is+dynamic+%7B%7Bwooo%7D%7D%0A++++++++%3C%2F%3Atext%3E%0A%0A++++++++%3C%3Alink%3E%0A++++++++++%3Ca+href%3D%22%23%22%3Elink%3C%2Fa%3E%0A++++++++%3C%2F%3Alink%3E%0A++++++%3C%2Fwrapper%3E%0A++++%3C%2FT%3E%0A++%3C%2Ftemplate%3E%0A%0A%60%60%60)

(was working on this with @RuslanZavacky )
tl;dr: we want to be able to use components in translations.
So, today, we need to have this extraneous wrapper component:

```hbs
<T @key="test-key" @label="my label" as |wrapper|>
  <wrapper>
     <:text>
       some text that is dynamic {{wooo}}
     </:text>

     <:link>
       <a href="#">link</a>
     </:link>
  </wrapper>
</T>
```    
But it'd be ideal if we didn't need it at all:
```hbs
<T @key="test-key" @label="my label" as |wrapper|>
  <:text>
    some text that is dynamic {{wooo}}
  </:text>

 <:link>
   <a href="#">link</a>
 </:link>
</T>
```
but we can't do that unless we can forward all blocks passed to `T` to the underlying component.

PR #461, opened 2019-03-07
[DATA] Singleton RecordData

runspired:
[Rendered](https://github.com/runspired/rfcs/blob/ember-data-singleton-record-data/text/0000-ember-data-singleton-record-data.md#singleton-record-data)

### DIFF
--- a/text/0461-ember-data-singleton-record-data.md
+++ b/text/0461-ember-data-singleton-record-data.md
@@ -1,0 +1,444 @@
+- Start Date: 2019-03-06
+- RFC PR: 461
+- [RFC Tracking Issue #51](https://github.com/emberjs/rfc-tracking/issues/51)
+
+# Singleton Record Data
+
+## Summary
+
+Ensures that `RecordData` can be implemented as a singleton, eliminates several redundant APIs
+that creeped into the original implementation, and simplifies the method signatures of
+RecordData APIs by using [`Identifiers`](https://github.com/emberjs/rfcs/pull/403).
+
+## Motivation
+
+`RecordData` is the data-cache primitive for information given to `ember-data`, and while
+a cache-per-entity setup is sometimes desired, a singleton cache offers opportunities for
+performance optimization and improved feature sets. Current default `RecordData`, and
+`InternalModel` which it replaced are examples of cache-per-entity strategies.
+
+Our original intent [when discussing `RecordData`](https://github.com/emberjs/rfcs/pull/293)
+was for it to be possible to be implemented as a singleton; however, this intent was not
+captured well in that RFC and while the APIs presented there would have enabled it, the
+actual implementation differed in ways that prevent `singleton` implementations.
+
+The introduction of `Identifiers` presents us with a good opportunity to refactor the API
+surface of our cache-primitive, simplifying and streamlining how it works while solidifying
+and codifying its ability to be a `singleton`.
+
+### Reduce Overloading
+
+Previously, to deal with the lack of a unified identifier concept, we overloaded many
+`StoreWrapper` and `RecordData` method signatures with `modelName, id, clientId` as
+`arguments`. The original intent was to overload **all** of their method signatures
+in this way to ensure `singleton` implementations could be built, but we failed to
+correctly implement the original `RecordData RFC` in this regard.
+
+The introduction of `Identifiers` provides us a cleaner interface for communicating identity
+in these APIs. For uniformity, methods will always take an `identifier` as their **first
+argument**.
+
+## Detailed design
+
+Because `RecordData` is a public userland interface we
+must rely on capabilities reporting to handle the
+deprecations for it introduced by this RFC.
+
+We will use this opportunity to improve the encapsulation
+of `RecordData` via introduction of a sandbox.
+
+This sandbox will be responsible for handling interop,
+deprecations and other book-keeping tasks we require as needed
+while ensuring that `RecordData` implementations are properly
+encapsulated (e.g. implementations can only talk to other
+implementations via public API).
+
+RecordData implementations must now specify a `version`.
+The implicit version when unspecified is `"1"` (the version
+which this RFC supplants and deprecates). Implementations
+without a version or with a version equal to `"1"` will
+receive a deprecation notice the first time an instance of
+a previously unseen `RecordData` class is encountered.
+
+In keeping with the support policy of `Ember.js`, future
+releases of `ember-data` will support the versions of `RecordData`
+current at the time of the last two `ember-data` `LTS` releases.
+
+The APIs proposed by this RFC constitute version `"2"`
+and will be used by `ember-data` when the following is
+`true`.
+
+```js
+recordData.version === "2";
+```
+
+It is possible that once this upgrade to `identifiers` has occurred that we never need to
+increment this `version` again. If so, a future RFC may choose to deprecate the `version`
+property altogether once "version 1" is no longer supported.
+
+### Sandboxing
+
+The sandbox implementation would guarantee the following
+
+Beginning with this RFC, `StoreWrapper.recordDataFor` will no longer directly return the `RecordData`
+instance provided by the `createRecordDataFor` hook, instead returning a delegate.
+
+This encapsulates the `RecordData` feature ensuring that communication is via public API
+and provides us the ability to manage the three concerns listed above and described below.
+
+**Managing Deprecations**
+
+When a `RecordData` method or method signature is deprecated, we will detect calls to deprecated
+methods or calls using deprecated method signatures and print the appropriate deprecation messages.
+
+**Version Interop**
+
+When a `RecordData` method or method signature is deprecated the instance provided by `createRecordDataFor`
+may not have implemented the deprecated method or signature. Using the supplied version, calling context,
+and other information available we will transform calls to deprecated methods or signatures
+into their supported equivalent.
+
+**Updates to Store Methods**
+
+<table>
+<thead>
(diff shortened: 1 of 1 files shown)


igorT:
I think this is a great overall approach. Just a couple questions/suggestions:

What happens If you pass in an identifier that RD hasn't seen before to the methods? 

Also, it might be helpful if you reiterated what the final RD interface looks like with these changes applied.

text/0000-ember-data-singleton-record-data.md
```diff
@@ -0,0 +1,450 @@
...
+RecordData implementations must now specify a `version`.
+The implicit version when unspecified is `"1"` (the version
+which this RFC supplants and deprecates). Implementations
+without a version or with a version equal to `"1"` will
+receive a deprecation notice the first time an instance of
+a previously unseen `RecordData` class is encountered.
+
+Note that to comply with `SemVer`, any given release of `ember-data` will
+only have support for a maximum of two versions of `RecordData`, with
+deprecated versions being dropped following LTS releases.
```
hjdivad: This is a good policy and standard in the ember ecosystem, but it doesn't really have much to do with semver per se.

Also, it's somewhat implied, but we should be explicit that we only intend to increment these versions at LTS, so that "we support 2 versions of RecordData" means in practice that we support the versions that existed at LTS versions of ember data.

hjdivad: Maybe something like


> In keeping with the support policy of Ember.js, future releases of `ember-data` will support the versions of RecordData current at the time of the last two `ember-data` LTS releases

text/0000-ember-data-singleton-record-data.md
```diff
@@ -0,0 +1,450 @@
...
+
+When a `RecordData` method or method signature is deprecated the `target` may not have
+implemented the deprecated method or signature. The `RecordDataManager` would use the
+combination of calling context, presence of the method on the `target` and the declared
+support `version` of the `target` to convert calls to deprecated methods or signatures
+into their supported modern equivalent.
+
+**Internal Book Keeping**
+
+In some situations `ember-data` will need to internally track of specific kinds of requests
```
hjdivad: s/to internally track of/internally keep track of

text/0000-ember-data-singleton-record-data.md
```diff
@@ -0,0 +1,450 @@
...
+or operations sent to a `RecordData` instance. The `Manager` is our "intercept" point at
+which such tracking would be handled. Notably `RequestManager` will never fail to call
+the appropriate method with the appropriate arguments on the `target`, it will merely
+make notes about what has been called. E.g. `RequestManager` allows for `ember-data` to
+introduce `stateless side-effects` for some method calls, but it will not alter the original call
+or response beyond translating calls for deprecations. `Stateless` means that these `side-effects`
+will not alter any `application state` and have largely invisible effects on the data layer.
+From the perspective of a `RecordData` author these methods should be considered `stateless`.
```
hjdivad: Is this paragraph intended to be about `RecordDataManager` and not `RequestManager`?

text/0000-ember-data-singleton-record-data.md
```diff
@@ -0,0 +1,450 @@
...
+**Friendly API Accesss**
+
+Some `RecordData` authors may wish to allow their `RecordData` instances to share information
+with each other via private API. While this is heavily discouraged and patterns exist that
+allow for this to be unnecessary, in the **extreme** event that it is required we allow
+access to the `RecordData` class via `Manager.target()`.
+
+Using this `API` comes with consequences as it decreases the serializability of some forms
+of state and allows for unexpected sideeffects.
+-->
```
hjdivad: Remove This?

text/0000-ember-data-singleton-record-data.md
```diff
@@ -0,0 +1,450 @@
...
+
+**Updates to Store Methods**
```
hjdivad: might be a pain in ghfm but vertically aligning these cells top makes this an easier read IMHO

text/0000-ember-data-singleton-record-data.md
```diff
@@ -0,0 +1,450 @@
...
+    key: string
+  ): void {}
+
+  @deprecated
+  notifyBelongsToChange(
+    modelName: string,
+    id: string | null,
+    clientId: string | null,
+    key: string
+  ): void {}
```
hjdivad: these are deprecated in favour of what?

text/0000-ember-data-singleton-record-data.md
```diff
@@ -0,0 +1,450 @@
...
+  unloadRecord(identifier: Identifier);
+  isRecordInUse(identifier: Identifier): boolean;
+  isEmpty(identifier: Identifier): boolean;
+  isNew(identifier: Identifier): boolean;
+
+  getAttr(identifier: Identifier, propertyName: string): any;
+  isAttrDirty(identifier: Identifier, propertyName: string): boolean;
+  changedAttributes(identifier: Identifier): AttributesChanges;
+  hasChangedAttributes(identifier: Identifier): boolean;
+  rollbackAttributes(identifier: Identifier): void;
```
hjdivad: If we're changing the API anyway, let's be consistent about whether they're `Attr`s or `Attributes`

text/0000-ember-data-singleton-record-data.md
```diff
@@ -0,0 +1,450 @@
...
+
+</td>
+<td>
+
+```ts
+interface RecordDataV2 {
+  unloadRecord(identifier: Identifier);
+  isRecordInUse(identifier: Identifier): boolean;
+  isEmpty(identifier: Identifier): boolean;
+  isNew(identifier: Identifier): boolean;
```
hjdivad: Why do we have record in `isRecordInUse` but not in `isNew`?  No doubt it's a holdover from the original state flags, but for a new api does this naming distinction make sense?

text/0000-ember-data-singleton-record-data.md
runspired: I like that wording

text/0000-ember-data-singleton-record-data.md
runspired: 👍

text/0000-ember-data-singleton-record-data.md
runspired: Yeah I noticed that, will try

text/0000-ember-data-singleton-record-data.md
runspired: I cleaned that up elsewhere but agree, made this edit to improve the consistency

text/0000-ember-data-singleton-record-data.md
runspired: None of the updates for RecordData state use `Record` in the names, so at first I was tempted to say "it doesn't make sense" and "we should drop it".

https://github.com/igort/rfcs/blob/igor/record-state/text/0000-record-data-state.md

But then I decided to investigate a bit more about why `isRecordInUse` exists:

`RecordData.isRecordInUse` calls `StoreWrapper.isRecordInUse` which calls `InternalModel.isRecordInUse`.

It is only used in one place (by `_cleanupOrphanedRecordDatas`)

And it was unspecified by the original RFC for both the StoreWrapper and the RecordData location: https://github.com/emberjs/rfcs/blob/master/text/0293-record-data.md

I think we should take this opportunity to drop it entirely as an internal concern. It essentially was something we leaked due to the relationship layer using RecordData instances as identifiers, but once we refactor that layer to use identifiers instead the need for it here dies.

text/0000-ember-data-singleton-record-data.md
runspired: 🤦‍♂

text/0000-ember-data-singleton-record-data.md
runspired: Ah, I meant to discuss this and forgot. We may want to add `target()` to `RecordDataManager` to allow this access, but in my investigations I do not believe that it is actually necessary and that we would be better off not exposing the target unless strongly compelled to later.

text/0000-ember-data-singleton-record-data.md
hjdivad: That sounds good to me.  We should make sure this behaviour is tested of course; but I agree that the "unload" story with relationships is quite specific to the relationship layer code.

text/0000-ember-data-singleton-record-data.md
runspired: `notifyPropertyChange`

text/0000-ember-data-singleton-record-data.md
```diff
@@ -0,0 +1,463 @@
...
+  @deprecated // this RFC deprecates this method entirely and without replacement
+  removeFromInverseRelationships(isNew: boolean): void;
+}
+```
+
+</td>
+<td valign="top">
+
+```ts
+interface RecordDataV2 {
```
hjdivad: We should add `version` here or at least somewhere in the RFC fully describe the `RecordDataV2` interface.

hjdivad:
Per weekly discussion; action item to merge:

1. the version issue listed inline
2. drop the discussion of `*Manager` as it's an implementation detail and distracts from the public API

igorT:
@igorT to look into `isRecordInUse`

chriskrycho:
Would someone please be so kind as update the `Rendered` link to point to the final location of the text? 🙏 Thanks!

hjdivad:
updated; thanks @chriskrycho

PR #462, opened 2019-03-10
Configuring addon modules in Module Unification layout

ppcano:
[Rendered](https://github.com/ppcano/rfcs/blob/addon-module-config/text/0000-addon-module-config.md)

### DIFF
--- a/text/0000-addon-module-config.md
+++ b/text/0000-addon-module-config.md
@@ -1,17 +1,17 @@
 - Start Date: 2019-03-10
 - Relevant Team(s): Ember CLI
-- RFC PR: (after opening the RFC PR, update this with a link to it and update the file name)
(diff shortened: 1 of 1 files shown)


mixonic:
@ppcano Thanks for the thought on this!

As you may already be aware in https://blog.emberjs.com/2019/03/11/update-on-module-unification-and-octane.html and https://github.com/emberjs/rfcs/pull/367#issuecomment-468930941 the module unification design was largely set aside in favor of designs which layer more transparently on the ES modules system. This means most of the implementation details, even those lacking design like the addon modules configuration API, are effectively dead ends.

With that in mind (and after raising this with framework core at https://github.com/emberjs/core-notes/blob/master/ember.js/2019-09/Meeting_Ember_Core_Team_20190920.md) I'm going to flag this RFC as FCP to close. I appreciate the time you put in here, and it would be great to work with you an another contribution or design.

PR #463, opened 2019-03-13
Record Data State

igorT:
[Rendered](https://github.com/igort/rfcs/blob/igor/record-state/text/0463-record-data-state.md)

### DIFF
--- a/text/0463-record-data-state.md
+++ b/text/0463-record-data-state.md
@@ -1,0 +1,83 @@
+- Start Date: 2019-03-13
+- Relevant Team(s): data
+- RFC PR: https://github.com/emberjs/rfcs/pull/463
+- Tracking: https://github.com/emberjs/rfc-tracking/issues/47
+    
+# Record State on Record Data RFC
+  
+
+## Summary
+
+This RFC is a follow-up RFC for #293 RecordData.
+
+Add `isNew` , `isDeleted` and `isDeletionPersisted` methods on Record Data that expose information about newly created and deleted states on the record
+
+## Motivation
+
+RecordData handles managing local and server state for records. While the current API surface of RecordData handles most data wrangling needs, as we have used RD internally in ED and in addons we have repeatedly run into the need of knowing whether a record is in a new or a deleted state.  The initial version of Record Data made the choice to not expose state information in order to limit the scope of the RFC and to give us time to come up with a design.
+
+## Detailed design
+
+Add the following methods on the RecordData interface:
+
+```ts    
+interface RecordData {
+    
+  // To be added in this RFC
+  isNew(identifier: RecordIdentifier): boolean
+
+  isDeleted(identifier: RecordIdentifier): boolean
+
+  isDeletionCommitted(identifier: RecordIdentifier): boolean
+    
+  setIsDeleted(identifier: RecordIdentifier, boolean: isDeleted): void
+}
+```
+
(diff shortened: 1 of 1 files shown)


mikkopaderes:
Are there any plans for an `isUnloaded` state? Current use-case is for an adapter that supports realtime changes (e.g. Firebase). Whenever someone loses permission or a record was deleted some adapters might want to unload those records in realtime. However, I don't think there's anything that exists to indicate this state.

Another related property may be `unloadReason` where it'll contain the reason why the record is in `isUnloaded: true` state.

runspired:
@mikkopaderes the short answer is that there is not. We did explore whether `recordData` should consider `unload` as an operation on `RecordState`, and the clear answer was **`"no"`**.

A TL;DR of what to do instead is ______________

If we start with a bit of context on what `unloadRecord` is _meant_ to achieve and how that fits in both the pre-record-data and post-record-data worlds helps to understand ______________

In pre-record-data `ember-data`, `InternalModel` fulfilled many roles including:

* being the identifier for a resource that may or may not have been loaded
   Read the [Identifiers RFC](https://github.com/emberjs/rfcs/pull/403) for more context and insight on how we are replacing and improving that role.

* being the cache for attributes/relationships for a resource (now handled by `RecordData`)

* being the cache for loading & network state: see [Request State RFC](https://github.com/emberjs/rfcs/pull/466)

* being the cache for `RecordState` (this RFC)

* being the cache for `Error State` (but not errors themselves which were stashed on record instances)
  Read the [RecordData Errors RFC](https://github.com/emberjs/rfcs/pull/465)

An unfortunate design within `InternalModel` was that `loading|network|error|record` state were mixed into a single state machine. You can see (from the RFCs above) that this is something we are working to rectify.

With the `InternalModel` design, `unloadRecord` played a _poorly fit_ role of needing to change the `state` of the `InternalModel` from wherever it was back into `empty`, needing to notify any relate records that it would need to be loaded again if re-accessed, and of actually emptying the cache.

In the `RecordData` design, unloadedRecords are simply absent from the cache, with knowledge of their existence maintained if needed via an `Identifier`. By separating the identity from the data, we can seamlessly release the data, and use the absence of it to know we need to retrieve it when asked.

text/0000-record-data-state.md
```diff
@@ -0,0 +1,113 @@
...
+
+    
+    interface RecordData {
+    
+    	// To be added in this RFC
+    	isNew(identifier: RecordIdentifier): boolean
+    
+    	isDeleted(identifier: RecordIdentifier): boolean
+    
+      isDeletionPersisted(identifier: RecordIdentifier): boolean
```
rwjblue: Bizarre spacing here, maybe tabs vs spaces or something?

text/0000-record-data-state.md
```diff
@@ -0,0 +1,117 @@
...
+  unloadRecord(identifier: RecordIdentifier);
+
+  rollbackAttributes(identifier: RecordIdentifier);
+  changedAttributes(identifier: RecordIdentifier): ChangedAttributesHash;
+  hasChangedAttributes(identifier: RecordIdentifier): boolean;
+
+  getAttr(identifier: RecordIdentifier, key: string): any;
+  getRelationship(identifier: RecordIdentifier): JsonApiRelationship;
+
+  performMutation(operation: Operation): void
```
igorT: Remove from RFC before merging

PR #465, opened 2019-03-14
Record Data Errors RFC

igorT:
[Rendered](https://github.com/igorT/rfcs/blob/igor/record-data-errors/text/0465-record-data-errors.md)

### DIFF
--- a/text/0465-record-data-errors.md
+++ b/text/0465-record-data-errors.md
@@ -1,0 +1,76 @@
+- Start Date: 2019-03-13
+- Relevant Team(s): data
+- RFC PR: https://github.com/emberjs/rfcs/pull/465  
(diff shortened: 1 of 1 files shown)


text/0465-record-data-errors.md
```diff
@@ -0,0 +1,76 @@
...
+```
+
+`RecordValidationError` follows the subset of the JSON api errors spec. For example, if the record being saved was rejected because the attribute `password` was empty, the `RecordValidationError` could look like: 
+
+```ts
+{
+  title: 'Field cant be empty',
+  detail: 'Field must be at least 8 characters long',
+  source: {
+    pointer: 'attributes/password'
```
Turbo87, line 53: shouldn't this be `/data/attributes/password`? 🤔

PR #468, opened 2019-03-15
`@classic` Decorator

pzuraq:
[Rendered](https://github.com/pzuraq/emberjs-rfcs/blob/classic-decorator/text/0000-classic-decorator.md)

### DIFF
--- a/text/0000-classic-decorator.md
+++ b/text/0000-classic-decorator.md
@@ -43,6 +43,8 @@ This RFC proposes adding a `@classic` decorator which can be used on classes
 that still use `EmberObject` APIs:
 
 ```js
+import classic from 'ember-classic-decorator';
+
 @classic
 export default class ApplicationController extends Controller {
   init() {
@@ -109,12 +111,13 @@ considered `EmberObject` APIs, since they could be implemented purely in native
 classes, and aren't tied to the way `EmberObject` does inheritance.
 
 If a class, or any of its parent classes, uses one of these methods _without_
-being decorated with `@classic`, a warning will be logged to let users know that
-they are using classic object APIs, and they should either refactor away from
-using those APIs, or use the `@classic` decorator to opt-out of warnings.
+being decorated with `@classic`, a lint rule will log a warning to let users
+know that they are using classic object APIs, and they should either refactor
+away from using those APIs, or use the `@classic` decorator to opt-out of
+warnings.
 
-The `@classic` decorator itself will brand the class it is applied to with a
-symbol, disabling warnings on that instance of the class.
+The `@classic` decorator itself will brand the class it is applied to, disabling
+warnings on that instance of the class.
 
 ### Non-transitionable Base Classes
 
@@ -134,10 +137,10 @@ entirely.
 
 ### Implementation
 
-The decorator and the warnings would be added by an official Ember addon. This
-would allow us to keep the implementation details separate from the main Ember
-codebase, especially specifics like babel transforms for stripping out the
-`@classic` decorator in production builds.
+The decorator and the warnings would be added by an official Ember addon,
+`ember-classic-decorator`. This would allow us to keep the implementation
+details separate from the main Ember codebase, especially specifics like babel
+transforms for stripping out the `@classic` decorator in production builds.
 
 The `@classic` decorator should only be applied in DEBUG mode, and should be
 stripped entirely from production builds. Other than that, the details of the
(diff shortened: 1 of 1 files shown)


text/0000-classic-decorator.md
```diff
@@ -0,0 +1,196 @@
...
+Non-transitionable base classes _cannot_ be converted to native classes without
+relying on behaviors that are specific to `EmberObject`. Classic components, for
+instance, rely deeply on the way that `EmberObject.create` sets up its instance,
+and any utility class that extends from `EmberObject` _must_ use `create` to
+create instances. While this is also true of `Route`s and `Service`s, the
+important distinction is that _users_ don't have to ever use `create` in their
+own code, or are not passed any arguments other than injections, so for those
+classes the usage of `create` is an implementation detail.
+
+These classes will _always_ warn users if they are not decorated with
```
NullVoxPopuli: is this a linter warning or a deprecation style warning?

pzuraq: This would be a deprecation style warning, since it’s not information we can actually gather through static analysis. We need to actually run/construct the class to get this information

text/0000-classic-decorator.md
```diff
@@ -0,0 +1,196 @@
...
+   `EmberObject` specific functionality, and to be aware that they are not
+   working with "just" native classes.
+2. Provide a hint to _linters_ so that they can use a different set of rules for
+   "classic" classes, such as always use `init`, etc. and native classes, such
+   as to never use `.extend()` or mixins.
+3. Disable a number of warnings that will alert users when they have encountered
+   an edge case, such as using `init` in a parent class and `constructor` in a
+   child class.
+
+This should allow users who wish to adopt native class syntax to do so
```
NullVoxPopuli: should this also be a code mod? to add the @classic decorator to each EmberObject class?

pzuraq: Absolutely! I think a codemod for this would be very helpful

Gaurav0: I'm not so sure. Adding @classic is fairly quick and it might behoove developers to evaluate whether each class needs it manually.

piotrpalek:
I think it would be wise to separate `EmberObject` and native class syntax. There are too many edge cases and weird behaviors. It would also lead to an easier way of teaching people: "look, this is how Ember is build, but we have this new world where we use native classes and which has different behavior, but it doesn't work for all things that _classic Ember_ supports". Worked great for functional components in the React world.

pzuraq:
@piotrpalek that's fair, in fact our recommended upgrade path will be:

* Keep classic components as classic classes until they can be converted to Glimmer components
* Keep utility classes that extend `EmberObject` as they are, using classic syntax, until you can convert them directly to native classes.

However for Routes/Controllers/Services, there are already far fewer differences between native class and classic class syntax. Really, the only major ones are using mixins, and using `init` vs `constructor`. Given we don't have any native class alternatives for these yet, and that the difference is much much smaller between the two, it makes sense to recommend users to convert over to the newer syntax and incrementally remove these differences. In most cases, I think it will be a pretty straightforward conversion.

There are also some users who _must_ use native class syntax everywhere, such as users who wish to adopt TypeScript. This is different from the React conversion, because in that case, React was converting from one TS compatible form (native classes) to another (functional components). Since those users already exist, and have to deal with these problems anyways, this would provide better tooling for them to detect possible bugs.

piotrpalek:
Right, I didn't consider TS users. I can understand the reasons, I've also re-read the RFC and understand it a little better now. My comment was probably less enthusiastic about this because I both didn't understand it enough, as well as a drive-by comment after #467 :) Tl;dr I now consider it a good solution to a tough problem.

Sorry if it came out overly aggressive/negative!  🙏

pzuraq:
It's really no problem! This is going to be a tricky transition, and we're trying to strike the right balance - feedback that this could be too complicated is _exactly_ what led to this RFC, and the simplifications in #467 / #451. It's always good to get feedback, thanks for giving us your thoughts 😄

text/0000-classic-decorator.md
Gaurav0: Will it actually be a deprecation, with an identifier and an `until` field? Or just use warn from 'ember-debug'?

mixonic:
I believe this RFC needs to identify a package that `@classic` is imported from

rwjblue:
> I believe this RFC needs to identify a package that @classic is imported from

Ya, totally agree. The current implementation is done in an addon, and I don't see any reason not to keep it that way...

pzuraq:
Added the import path and addon name to the RFC, and updated some of the language. We found during implementation that it wasn't possible to restrict all usage of classic APIs in non-classic classes, so we used lint rules for these instead. The general restrictions are the same as before though.

rwjblue:
👍 from the core team at today's meeting, moving this into final comment period...

MelSumner:
At today's core team meeting, we have decided to move forward with this RFC. 👍

PR #318, opened 2018-03-24
`array` helper

mmun:
[Rendered](https://github.com/emberjs/rfcs/blob/array-helper/text/0318-array-helper.md)

Closes #128.

PR #470, opened 2019-03-21
`{{fn}}` Helper

rtablada:
[rendered](
https://github.com/rtablada/rfcs/blob/bind-helper/text/0000-bind-helper.md)


rtablada:
A possible speedbump for this is the addon https://github.com/Serabe/ember-bind-helper.

This addon provides an incompatible API to this RFC and is implemented by performing an AST transform to the private API of the `action` helper and modifier.

The API for this RFC provides a clearer and cleaner mental model and a "just code" implementation, but we should understand that there is potential collisions for teams that are using this addon in their apps today.

Serabe:
I'm not using any private API.

I've already talked to some users of the addon about providing an upgrade path to this RFC.

text/0000-bind-helper.md
```diff
@@ -0,0 +1,124 @@
...
+```
+
+### Setting the `this` context and arguments
+
+```hbs
+{{bind this.car.drive 1000 'mile' context=this.honda}}
+```
+
+```js
+this.car.logout.bind(this.car, 1000, 'mile')
```
Serabe: This should be `this.car.logout.bind(this.honda, 1000, 'mile')`.

Serabe:
I'm very much in favor of this landing. I'll try to get a beta version of ember-bind-helper that facilitates current users stop using it.

text/0000-bind-helper.md
```diff
@@ -0,0 +1,124 @@
...
+
+```hbs
+{{{bind this.session.logout context=this.anotherSession}}}
+```
+
+```js
+this.session.logout.bind(this.session)
+```
+
+### Setting the `this` context and arguments
```
chrisrng: imo this is a little confusing since its the opposite of what the JavaScript bind function argument order is (wrt the context)

Can we make use of the [array helper](https://github.com/emberjs/rfcs/pull/318) instead? Similar to how Angle Bracket does positional params (https://guides.emberjs.com/release/reference/syntax-conversion-guide/#toc_params-array)

```hbs
{{bind this.car.drive context=this.honda params=(array 1000 'mile')}}
```

text/0000-bind-helper.md
```diff
@@ -0,0 +1,124 @@
...
+```
+
+### Setting the `this` context
+
+```hbs
+{{{bind this.session.logout context=this.anotherSession}}}
+```
+
+```js
+this.session.logout.bind(this.session)
```
mehulkar: should be: `this.session.logout.bind(this.anotherSession)` i think?

text/0000-bind-helper.md
```diff
@@ -0,0 +1,124 @@
...
+Because `function.prototype.bind` always binds the `this` context the bind helper would require functions to be already bound or the `this` named argument would need to be explicitly set.
+
+## How we teach this
+
+For guides we would switch to recommending the `bind` helper to pass functions into components args and modifiers.
+In guides we would no longer recommend using the `action` helper based on the reasons listed in motivations.
+
+## Drawbacks
+
+Since this recommended design sets the `this` context to `null` using `function.prototype.bind` this could be confusing to junior developers if the function was not already bound (using something like an `@bound` decorator, JS binding, arrow functions, or the `bind` helper earlier in the stack).
```
mehulkar: s/junior developers/new ember developers. IMO this context binding seems like a basic thing a framework should provide when template/JS class are already closely coupled. it's confusing if it's not out-of-box behavior.

text/0000-bind-helper.md
Serabe: Doing this would be making the use of this helper too verbose and a step backwards with respect to the `action` helper.

text/0000-bind-helper.md
Serabe: :+1: to s/junior developers/new ember developers/

While it might look like a good idea to have automagic context binding, this behaviour is closes to how JS and other libs work. Reducing the difference between JS and HBS is a great goal, IMHO.

jonnii: When calling an action on a component don't you almost always want `this` to be the component instance? would an `@action` called with `bind` be automatically have the `this` context set to the component instance?

pzuraq: Correct, `@action` provides context binding. The magic that we’re discussing that is problematic is the fact that a _helper_ or _modifier_ can access this context implicitly.

text/0000-bind-helper.md
pzuraq: Yeah, we debated about this a bit, it comes down to “naming is hard”. What we want is a partial application helper. We’ve discussed:

* `partial` - used by aome other languages, but we can’t because it means something in hbs
* `apply` - also used by other languages, but already used in JS, and means something very different
* `curry` - could work, but then we have to explain the term. Also, may not be _technically_ correct (currying produces a function that recieves exactly 1 argument, or something like that)
* `papply` - easy to misinterpet if you don’t know what the term “partial application”
* `partial-apply` - fairly verbose for a common task

`bind` is the _least_ bad choice IMO, it’s like if you made a `bind` function with a slightly better API, but I also agree it’s not ideal and could be confusing. I think `curry` is the second best out of those options, and am open to suggestions!

text/0000-bind-helper.md
jonnii: In which case the "how we teach this" section is misleading:

> For guides we would switch to recommending the bind helper to pass functions into components args and modifiers. In guides we would no longer recommend using the action helper based on the reasons listed in motivations.

This suggests you'd always want to use `bind` over `action` but I'd think that bind is only useful for the cases where you'd want a different context, which is (in my experience) never been a problem I've faced. Not to say that it's not a problem, but `targetObject` was how this was done in the past and that wasn't very pleasant.

pzuraq: I think the missing piece is that we would recommend _against_ the `{{action}}` helper, and would recommend using the `@action` decorator for all actions. The `{{action}}` helper would likely be deprecated at some point in the future.

rtablada: For clarification (and something I will add to the RFC) is that we should recommend `bind` over `action` for many reasons but mainly because of the implicit `target=this` which is applied at build time.

For guides we would recommend `bind` without any `context=` because as @pzuraq says we'd recommend always using JS to bind context using the `@action` decorator or some other mechanism (like an `@autobind` decorator or even a language feature in the future)

mehulkar: So for a button event listener to work, a dev would need to use {{bind}} in the template and @action in the JS. That seems like a weird thing to remember, IMO. @autobind or @bindcontext in the JS seems a little better IMO, but I know that’s a big departure from the current “action” terminology.

text/0000-bind-helper.md
```diff
@@ -0,0 +1,124 @@
...
+```
+
+```js
+this.add.bind(null, 1, 2);
+```
+
+### Setting the `this` context
+
+```hbs
+{{{bind this.session.logout context=this.anotherSession}}}
```
luxferresum: I hope the tiple braces are by accident?

text/0000-bind-helper.md
luxferresum: but isnt is confusing considered that in 90% of the cases `{{bind` will *not* be used to bind context but `@action` will bind context?

text/0000-bind-helper.md
jonnii: So the way this will work would be that you'd decorate a method with `@action` and then call it with `{{bind}}` like:

```
@action
function increment(amount) {
  this.value = this.value + amount;
}

<button onclick={{bind this.increment 5}} />
```

The `@action` decorator in this case would be responsible for setting the correct context (e.g. setting the `this` to the component). Is that right?

text/0000-bind-helper.md
bendemboski: I think `context` is a pretty overloaded word that doesn't really convey a lot of meaning. Why don't we use `thisArg` to make it less ambiguous and match the [EMCA spec](https://www.ecma-international.org/ecma-262/5.1/#sec-15.3.4.5)'s naming?

text/0000-bind-helper.md
richard-viney: @jonnii That's my understanding, yes. Note also that with the RFC for the `{{on}}` modifier (#471)  you would now write `<button {{on "click" (bind this.increment 5)}} />`. v0.8 of the `ember-on-modifier` addon currently also allows you to omit the `bind` and just write `<button {{on "click" this.increment 5}} />`, but that isn't currently in the RFC.

viniciussbs: @mehulkar we already have to remember to declare actions inside `actions`. Using `@action` is a 1-1, so it would not be a problem, I guess.

viniciussbs:
I use `ember-bind-helper` today to call service methods. Using the proposed `bind` template helper and something like a `@bound` decorator, I could use the same sintaxe I already use, I guess:

```js
@bound
method() {
  // method on a service
}
```

```handlebars
<button onclick={{bind myService.method}}>
  Call a method on a service
</button>
```

text/0000-bind-helper.md
mehulkar: @rtablada, A little confused, what is the purpose of the default `target=this` being applied at build time? 

@viniciussbs Yeah, actions were a weird idiosyncrasy of the framework. bringing in plain JS bind, but keeping the idiosyncrasy makes things more weird for me, and just seems like an unnecessary hat tip to backwards compatibility. Wouldn’t it make more sense to just default to `context=this`?

Serabe:
@viniciussbs in that case you don't even need the `bind` helper, as you are not passing any extra arguments.

text/0000-bind-helper.md
Serabe: @mehulkar defaulting to `context=this` is more harmful than helpful. You might not be passing a function defined in the template context, but one coming from a service or a model, like `{{bind this.model.save}}`. That is the reason in `ember-bind-helper` we default to the everything but the last part of the call invocation (in the example `this.model`).

text/0000-bind-helper.md
luxferresum: +1, or just `this`?

rtablada:
@viniciussbs for some clarification we would recommend `@action` in the service and then using the `on` modifier on the button (no bind needed since you are not partially applying args or setting `this`).

```js
@action
method() {
  // method on a service
}
```

```hbs
<button {{on "click" this.myService.method}}>
  Call a method on a service
</button>
```

text/0000-bind-helper.md
jonnii: When using `bind` with a context and an `@action` decorator which one wins?

Serabe: `@action`

Serabe:
Quick update: if you are using `ember-bind-helper`, see the [deprecation path](https://github.com/Serabe/ember-bind-helper#deprecation-path) and let me know if I can do something else to ease it.

text/0000-bind-helper.md
mehulkar: > That is the reason in ember-bind-helper we default to the everything but the last part of the call invocation

yeah this would be fine (and better) also. Defaulting to `null` is what I'm objecting to.

text/0000-bind-helper.md
```diff
@@ -0,0 +1,124 @@
...
+<button onclick={{bind (mut showModal) true}}>Click</button>
+<button {{on "click" (bind (mut showModal) true)}}>Click</button>
+```
+
+Because `function.prototype.bind` always binds the `this` context the bind helper would require functions to be already bound or the `this` named argument would need to be explicitly set.
+
+## How we teach this
+
+For guides we would switch to recommending the `bind` helper to pass functions into components args and modifiers.
+In guides we would no longer recommend using the `action` helper based on the reasons listed in motivations.
```
karthiicksiva: <button onclick={{action this.increment}}>Click</button> => Not recommended
<button {{action "increment"}}>Click</button> => ?

@rtablada What about action modifer? Is that still recommended approach? Or action helper/modifier will be deprecated soon and avoid the usage?

viniciussbs:
@rtablada So, if using `@action` inside a service is OK, the `bind` helper will be - most of the time - used to partially apply args. Like @luxferresum said, `bind` will _not_ be used to bind context; we will bind context with `@action` and partially apply args with `bind`. That sounds weird and makes like more the other names shared by @pzuraq. I'll comment there.

text/0000-bind-helper.md
viniciussbs: I got confused by that too, @luxferresum. If `@action` will be the recommended way to bind context, I prefer a helper called `curry`, `apply-args` or something on that direction to partially apply arguments.

jonnii:
> @rtablada Like @luxferresum said, bind will not be used to bind context; we will bind context with @action and partially apply args with bind.

Except when you supply your own context - this means you need to look at both the template and the code to understand what the context of the ultimate action is. Is there a solution where the context is always defined in one place or the other?

pzuraq:
We discussed this RFC at the face-to-face last week, and with @rtablada's permission I've updated it to reflect the feedback from that meeting and the current state of things. Our main feeling is that the mismatch between the semantics of `bind` in JS and templates would be confusing to users (and there was some feedback to that effect), so it would be better to not have context binding at all for this helper, relying entirely on the `@action` decorator for binding.

We discussed _many, many_ names for this new helper. I outlined some of the alternatives we went over in the RFC, along with the reasoning and pros/cons for them. There is a decent amount of support for `with-args`, moreso than `bind`, which is why I updated the RFC to use it, but I definitely want to be clear that we are still open to suggestions! If anyone has ideas for the helper name, please feel free to comment and add them here 😄

jonnii:
@pzuraq the examples still have the `context` argument, is this intentional? One more question. Would something like this be possible?

```
<button onclick={{this.increment()}}>Click</button>
<button {{on "click" this.add(1, 2)}}>Click</button>
```

This is such a common operation that maybe there's a way to express it with a more "native" feeling syntax. Under the covers this would be compiled to use `with-args`.

pzuraq:
@jonnii thanks for catching those! That was unintentional, I'll update soon.

So, the thing is, you examples will be possible soon (if they aren't already). They just have a bit of a different syntax:

```hbs
<button onclick={{this.increment}}>Click</button>
<button {{on "click" (this.add 1 2)}}>Click</button>
```

Glimmer templates have their own way of invoking functions (helpers), which is different from JavaScript. I don't think it makes much sense for us to try to _embed_ JS in our templates, that would definitely be going the JSX route.

The issue is though, the above runs those helpers _immediately_, because it's an inline function call. What we want is to have a syntax where we return a _new_ function that is called _later_. In React/JSX they have this same problem, and they encourage arrow functions in templates:

```jsx
<button onClick={(e) => this.deleteRow(id, e)}>Delete Row</button>
```

We briefly discussed adding something like arrow functions to Glimmer templates, something like:

```hbs
<button {{on "click" (|e| => (this.add 1 2))}}>Click</button>
```

But you can see this is actually still pretty verbose, and it opens up a whole can of worms. I think it's a feature we can definitely consider in the future, but in the meantime it makes more sense to have a helper which passes back a bound function.

text/0000-bind-helper.md
```diff
@@ -0,0 +1,114 @@
...
+
+### Multiple Argument Partial Application
+
+```hbs
+{{with-args this.add 1 2}}
+```
+
+```js
+return function() {
+  this.log.call(this, 1, 2);
```
chriskrycho:
```suggestion
  this.add.call(this, 1, 2);
```

rtablada:
Thanks @pzuraq for the updates! After a lot of discussion I agree that the primary goal of this RFC was to allow for argument application and moving away from `bind` in this case was the right choice.

It also descopes a lot of the feedback and confusion of `context` binding. It is clearer to bind in JS since it allows dev tools to better understand the value of `this` and it is cleaner to read. There may be a small subset where binding context in the template is the right thing to do and if so we can look at unified solutions if there is enough demand.

Going forward I think the happy path is to use `with-args` when you need to apply args, otherwise the raw function can be passed.

text/0000-bind-helper.md
```diff
@@ -0,0 +1,114 @@
...
+
+At build time the `action` helper currently is passed through an AST transform to explicitly bind the `this` to be deterministic at runtime. This is a private API where the outputted Glimmer is not a 1-1 to the template. Also, the `action` helper is confused and has overlap with the `action` modifier which has similar but slightly different behavior.
+
+Instead of this confusing and overloaded behavior, a new `with-args` helper would be introduced to do partial application (with no need for build time private APIs), and context binding will be done instead using the `@action` decorator in classes.
+
+## Detailed design
+
+The `with-args` helper will take in a function and then the set of arguments that will be partially applied to the function.
+
+Here are some examples of the `with-args` helper and the equivalent JS:
```
mixonic: The `call` examples here are not equivalent, as they bind `this` to the method. `{{with-args` specifically does not do this.

luxferresum:
I'm also very happy with this change. The only thing thats not so elegant is that `with-args` is a bit a long name. Have we considered just `args` as the name?

rtablada:
@luxferresum yeah there's been a bunch of bike shedding around name. I've come around to `with-args` since the pros outweigh the cons

Pros:

1. Descriptive
2. Opens up `args` to be used for other helpers

Cons:

1. `with-args` is a bit long to type

With template/sfc imports in the future pipeline you'll be able to alias to match your teams naming schemes.

text/0000-bind-helper.md
pzuraq: We're assuming that the wrapped function will have the correct context wherever it is called, e.g. if you were to pass that outer function to `addEventListener`, `this` would be the correct context (the element of the event handler) and binding it would result in the correct `this` being passed to `call`. Not binding it means it would default to whatever the context of `this` was when the function was defined, I _believe_.

richard-viney:
Would the name `args` possibly be a point of confusion due to the existence of `this.args` on the component?

pzuraq:
After more discussion, a few more options have come up for us to consider:

* `fn`, short for "function". There is prior art here from Lisp dialects which use `fn` for declaring an anonymous function, we would be using it for a similar purpose, but slightly different. It's much less verbose than `with-args`, which is a plus, but may be a little less clear:

```hbs
<div {{on "click" (fn this.save model)}}></div>
```

* A new custom syntax for partial application, something like `{{> }}` or `{{=> }}`. This could be nice for brevity, but may be less searchable and clear for newcomers, and may be more difficult to learn in general:

```hbs
<div {{on "click" (> this.save model)}}></div>
```

balinterdi:
I was at peace with `with-args` but I think `fn` is even better as it more accurately describes what it does (since the first argument to the helper is the function itself, only the other arguments are the arguments to that function).

`>` and other symbols are just too cryptic, it makes teaching/learning harder for no good reason. I also don't mind typing, so `func` or even `function` are fine but `fn` is easy to search for, so I think it's perfect.

chriskrycho:
Alignment with Rust can only be a good thing. We've got `::`, now `fn`? ***I'M SO IN.***

😆

rwjblue:
The RFC has been updated to use `fn` (instead of `{{with-args}}`), and with that change the core team believes this is ready to be moved into the final comment period.

viniciussbs:
I still prefer `with-args`. I don't think it's too verbose. And verbosity is not a problem for me when it's for clarity sake - like the `did-insert-element` modifier. But `fn` is a good alternative to `bind`, too.

cibernox:
Fwiw, I also prefer `bind` or `with-args`. I don't find the former too imperative, but I do find `fn` a bit cryptic.

bendemboski:
I still prefer `curry` because, according to it's common (not mathematical) usage, that's *exactly* what the helper does. But maybe that's not a widely-used/understood enough term?

chriskrycho:
@bendemboski heh, well, strictly speaking, it *partially applies* a function rather than *currying* it. And (at least in my view) we probably shouldn't encourage the confusion of the two terms, as they are both independently useful concepts!

bendemboski:
@chriskrycho yeah, that's a fair point. I was sorta relying on the fact that I believe most people kinda blur or confuse the two notions, so calling it currying would immediately invoke the idea of partial application, and since there's no sensible way (that I can think of) that `{{curry this.doSomething this.model}}` could describe a currying operation, it would end up being pretty intuitive. But maybe relying on a general lack of clarity around a distinction between terms isn't a good way to go...

chriskrycho:
@bendemboski yeah, I actually *can* think of ways that you might use it, but they're used together often enough to cause exactly that confusion. (I edited in an example for folks that are following along via email, which should hopefully show why they're often used together *and* why they're worth keeping conceptually distinct!) I definitely think keeping it *correct* is a worthwhile goal: it's almost never a good idea to *intentionally* name something in a way that runs contrary to established conventions or definitions.

text/0000-bind-helper.md
```diff
@@ -0,0 +1,113 @@
...
+
+```hbs
+{{fn this.add 1 2}}
+```
+
+```js
+return function() {
+  this.add.call(this, 1, 2);
+}
+```
```
simonihmig: I got a bit confused by looking at these examples, as they make it look like you have to pass *all* expected arguments, rather than partially applying them. Wouldn't the correct JS equivalent of `{{fn this.add 1 2}}` look more like this?

```js
return function(...args) {
  this.add.call(this, 1, 2, ...args);
}
```

pzuraq: Yeah actually that’s totally correct, we will update them!

TRMW:
Another vote for `with-args` over `fn` here, since it makes it a little clearer what this helper does. Either way I think these changes will be way less confusing than `action`, which I've seen many new Ember devs get tripped up by.

sheriffderek:
Great discourse! Maybe we can ask a non-ember person what they think 

`{{on "click" (fn (mut showModal) true)}}` might mean. 

`fn` seems like a pretty good place to land / but things like `mut` were always a big turn off to me - and I'd be curious how terrifying this looks to someone deciding to use the framework for the first time.

pzuraq:
@sheriffderek I think these examples are for folks who do use `mut`. We don't really recommend it directly in the guides, and it's not going to be more (or less) recommended as a result of this RFC. It's a different helper that can be combined with this helper, so it's a bit of a different concern.

text/0000-bind-helper.md
chriskrycho: @pzuraq it looks like this change didn't make it into the final version of the RFC text?

jacobq:
updated [link to RFC470](https://github.com/emberjs/rfcs/blob/master/text/0470-fn-helper.md)

PR #477, opened 2019-04-17
Blueprints Update

kellyselden:
[Rendered](https://github.com/kellyselden/rfcs/blob/blueprints-update/text/0477-blueprints-update.md)

### DIFF
--- a/text/0477-blueprints-update.md
+++ b/text/0477-blueprints-update.md
@@ -19,15 +19,20 @@ Projects like [libkit](https://github.com/tildeio/libkit), [@glimmer/blueprint](
 
 Up until now, ember-cli-update could update apps without any configuration, but in order to handle all blueprints and their options, some state has to be stored in the project.
 
-The code is already completed [here](https://github.com/ember-cli/ember-cli-update/pull/552), but I wanted to publish this RFC before releasing it into the wild. I want to make sure the format of these files are agreed upon, because if this is successful, it will probably propagate throughout the ecosystem.
+The code is already completed but opt-in. I wanted to publish this RFC before releasing it into the wild. I want to make sure the format of these files are agreed upon, because if this is successful, it will probably propagate throughout the ecosystem.
 
 ```json
 // config/ember-cli-update.json
 {
-  "blueprints": [
+  "packages": [
     {
       "name": "my-custom-blueprint",
-      "version": "0.0.1"
+      "version": "0.0.1",
+      "blueprints": [
+        {
+          "name": "my-custom-blueprint"
+        }
+      ]
     }
   ]
 }
@@ -37,38 +42,77 @@ The code is already completed [here](https://github.com/ember-cli/ember-cli-upda
 
 How should the state be structured?
 
-```json
-{
-  "blueprints": [
-    {
-      "name": "my-custom-blueprint",
-      "version": "0.0.1"
-    }
-  ]
-}
-```
+```ts
+/**
+  This lives at config/ember-cli-update.json by default.
+*/
+interface BlueprintConfig {
+  /**
+    Keep track of incompatible schema changes.
+  */
+  schemaVersion: number;
 
-```json
-{
-  "blueprints": {
-    "my-custom-blueprint": {
-      "version": "0.0.1"
-    }
-  }
-}
-```
+  /**
+    A package is a collection of blueprints from a single source.
+  */
+  packages: {
+    /**
+      The name of the package.
+    */
+    name: string;
 
-Future-proofing for more options
+    /**
+      Target a blueprint on your local filesystem, monorepo or otherwise.
+    */
+    location?: string;
 
-```json
-{
-  "additionalFutureOptions": null,
-  "blueprints": {
-    "my-custom-blueprint": {
-      "version": "0.0.1",
-      "additionalFutureOptions": null,
-    }
-  }
+    /**
+      The version of the package used for blueprint diffing.
+    */
+    version: string;
+
+    /**
+      The blueprints in the package. Can be a default blueprint having
+      the same name as the package, or a named blueprint like component
+      generators.
+    */
+    blueprints: {
+      /**
+        A base blueprint is the blueprint you started your project with.
+        This could be the ember app blueprint, or a custom blueprint like libkit.
+        If this is missing or false, it is a partial blueprint, like the changes
+        you get when installing an ember addon like ember-cli-mirage.
+        There must be one base blueprint at all times.
+      */
+      isBaseBlueprint?: boolean;
+
+      /**
+        The name of the blueprint to be run.
+      */
+      name: string;
+
+      /**
+        The path inside your project to run in. Think deeply nested
+        component generators. Missing means project root.
+      */
+      cwd?: string;
+
+      /**
+        The location of a codemods manifest if there are codemods
+        associated with this blueprint.
+      */
+      codemodsUrl?: string;
+
+      /**
+        All options should be normalized to match any codemod requirements.
+        The work to normalise them might not be done at first, so users
+        writing the CLI commands are expected to do it. For example,
+        `--no-welcome` should be expanded to `--welcome=false` so that
+        a codemod with a requirement of `[ "--welcome=false" ]` can match it.
+      */
+      options?: string[];
+    }[]
+  }[]
 }
 ```
 
@@ -111,22 +155,6 @@ This is a helper for saving the state of an old blueprint's run without ember-cl
 
 **Methods of delivery**
 
-Blueprints could be responsible for injecting the state into projects
-
-```json
-// my-custom-blueprint/blueprints/my-custom-blueprint/files/config/ember-cli-update.json
-{
-  "blueprints": [
-    {
-      "name": "my-custom-blueprint",
-      "version": "<%= blueprintVersion %>"
-    }
-  ]
-}
-```
-
-or ember-cli-update could manage it itself if the blueprint does not provide this info. You would have to provide the initial blueprint version to start from.
-
 The goal is for when you type `ember-cli-update` in your project, you will get a prompt like
 
 ```
@@ -148,43 +176,27 @@ ember new my-app -b my-blueprint --option1 --option2=foo
 ```
 
 ```json
-// my-blueprint/files/config/ember-cli-update.json
+// my-app/config/ember-cli-update.json
 {
-  "blueprints": [
+  "packages": [
     {
       "name": "my-blueprint",
-      "version": "<%= blueprintVersion %>",
-      "options": [
-        "option0=<%= option0 %>",
-        "option1=<%= option1 %>",
-        "option2=\"<%= option2 %>\""
+      "version": "0.0.1",
+      "blueprints": [
+        {
+          "name": "my-blueprint",
+          "options": [
+            "option1=true",
+            "option2=\"foo\""
+          ]
+        }
       ]
     }
   ]
 }
 ```
 
-```json
-// my-app/config/ember-cli-update.json
-{
-  "blueprints": {
-    "my-blueprint": {
-      "version": "0.0.1",
-      "options": [
-        "option0=false",
-        "option1=true",
-        "option2=\"foo\""
-      ]
-    }
-  }
-}
-```
-
 The updater can now generate a project with the correct options.
-
-```
-ember new my-app -b my-blueprint --option0=false --option1=true --option2="foo"
-```
 
 **Codemods**
 
@@ -192,16 +204,32 @@ There is no reason why you couldn't provide your own codemods with this system. 
 
 ```json
 {
-  "blueprints": {
-    "my-blueprint": {
+  "packages": [
+    {
+      "name": "my-blueprint",
       "version": "0.0.1",
-      "codemods": "some-server.com/my-blueprint-codemods-manifest.json"
+      "blueprints": [
+        {
+          "name": "my-blueprint",
+          "codemodsUrl": "some-server.com/my-blueprint-codemods-manifest.json"
+        }
+      ]
     }
-  }
+  ]
 }
 ```
 
-TODO: The options need to be normalized somehow to be used in the codemods system.
+## How we teach this
+
+I’m not sure. It could be a section in the ember-cli-update README, or leave it up to the blueprints that want to support this to document. If this feature takes off in the ecosystem, then it might warrant a guides section on keeping your blueprints up-to-date.
+
+## Drawbacks
+
+A drawback could be that blueprints start writing this state, but the consumer doesn’t want to use ember-cli-update to keep it in sync. In that case, it’s another file the user has to delete.
+
+## Unresolved questions
+
+The options need to be normalized somehow to be used in the codemods system.
 
 `--custom-option foo` vs `--custom-option=foo` vs `--custom-option="foo"`
 
@@ -221,10 +249,4 @@ TODO: The options need to be normalized somehow to be used in the codemods syste
 ]
 ```
 
-## How we teach this
-
-I’m not sure. It could be a section in the ember-cli-update README, or leave it up to the blueprints that want to support this to document. If this feature takes off in the ecosystem, then it might warrant a guides section on keeping your blueprints up-to-date.
-
-## Drawbacks
-
-A drawback could be that blueprints start writing this state, but the consumer doesn’t want to use ember-cli-update to keep it in sync. In that case, it’s another file the user has to delete.
+For now, we should leave it up to the end-user to normalize the options to match any codemod manifest. This is too hard a problem to nail down on the first iteration.
--- a/text/0477-blueprints-update.md
+++ b/text/0477-blueprints-update.md
@@ -72,6 +72,43 @@ Future-proofing for more options
 }
 ```
 
+**Commands**
+
+***bootstrap***
+
+```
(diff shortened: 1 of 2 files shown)


meirish:
Looks like you need the canonical link in the description - https://github.com/emberjs/rfcs/blob/464855635effeda3ea183d9438bbc8e86d99b4b0/text/0477-blueprints-update.md

kellyselden:
@meirish Can you try again? I updated the link.

mehulkar:
Awesome! I think this would be fantastic. Personally, I am perpetually confused by the various state files in Ember projects... add to that our own build and CI tools, the top level of a project directory quickly becomes intimidating for new users. I would love to see this go in an existing file or even package.json to avoid Yet Another File. It would also be super important for the file to state if it can be changed by end users or not and what effect it would have, so JSON format (sans comments) would be less than ideal, IMO.

kellyselden:
@mehulkar I've found via further testing that mixing a file that is tracked by the default ember-cli blueprint (.ember-cli, package.json) and modified by ember-cli-update (to update the blueprint metadata), it gets cumbersome. Both processes try to edit the same file. So in theory, ember-cli could alter the .ember-cli file in a way that conflicts with what you have in yours. This would give you a git conflict, then the updater can't read it anymore because it is in a state of invalid JS. I think this further supports a separate file.

mehulkar:
That makes sense

kellyselden:
Added sections about complete vs partial blueprints, saving options, and supporting custom blueprint codemods.

text/0477-blueprints-update.md
```diff
@@ -0,0 +1,248 @@
...
+      {
+        "name": "my-custom-blueprint",
+        "version": "0.0.1"
+      }
+    ]
+  }
+}
+```
+
+I vote “ember-cli-update.json”. Since the code will be modified by the updater, it’s easier to reformat JSON. “.ember-cli” I think supports JS, so it would be hard to modify. “package.json” seems like an abuse to put more metadata in there.
```
rwjblue: Agreed, lets drop the other examples and focus specifically on the stand alone `ember-cli-update.json`.

text/0477-blueprints-update.md
```diff
@@ -0,0 +1,248 @@
...
+  }
+}
+```
+
+**Methods of delivery**
+
+Blueprints could be responsible for injecting the state into projects
+
+```js
+// my-custom-blueprint/blueprints/my-custom-blueprint/files/.ember-cli
```
rwjblue: Let's consolidate all examples in this RFC on the `ember-cli-update.json` format. I find it pretty confusing to jump amongst 2 or 3 different formats while reviewing.

text/0477-blueprints-update.md
```diff
@@ -0,0 +1,248 @@
...
+ * ember-cli
+ * ember-cli-mirage
+ * ember-ci
+```
+
+The reason this is so powerful is any org could create their own _partial_ project blueprint (public or private). This blueprint can make any slight (or massive) tweaks to the official blueprints, and ember-cli-update can keep any project in sync with both the official blueprint and your org's blueprints, at the same time.
+
+**Complete vs partial**
+
+We need a way to denote a project replacement blueprint and a supplemental blueprint. A project replacement removes any sort of default ember-cli blueprint tracking, even if it still looks like a normal Ember project (has ember-cli, ember-addon, etc.). A supplemental blueprint is one that piggy-backs on a complete blueprint and makes certain tweaks to files. Think of ember-cli-mirage's extra project configs, or your company's custom settings on top of a normal Ember app.
```
rwjblue: Why do we need to know the difference between complete and partial? I think you are basically saying that we need to know the difference between the _original_ project blueprint (aka the "root" blueprint) and the others so that we know how to build up the comparison diff, is that right?

rwjblue:
A few things here:

* We should go with a dedicated file for `ember-cli-update.json` (not merging with `.ember-cli` file)
* The file should be in the configured config directory (from `package.json`'s `ember-addon.configPath` falling back to `config/`)
* We should include a version value for the file format itself (a top level key for `configVersion` or some such), so that we aren't absolutely completely locked down on the final format out of the gates
* Need to update the RFC to be more explicit about how blueprint options are stored (for example, if I `ember new foo --yarn --no-welcome` where are those flags stored in this format?)
* We should workshop the exact format for the blueprints a bit more, I'll try to propose an alternate schema in a separate comment as a way to kick start that convo

rwjblue:
RE: format for the blueprints array within `config/ember-cli-update.json`, I think we should come up with a detailed interface for that format. 

Here is what I was thinking (based on reading through this RFC, and looking at some of what was implemented already):

```ts
interface BlueprintConfig {
  /**
    Is this the initial project blueprint (e.g. `ember new` or `ember addon`)?
  */
  isRoot: boolean;
  /**
    The name of the blueprint to be ran.
  */
  name: string;

  /**
    The package the blueprint comes from.
  */
  package: string;

  /**
    The version of the package above to use. 
  */
  version: string;

  /**
    The path the blueprint is ran from.
  */
  rootPath: string;

  /**
    The blueprint command line options.
  */
  options?: {
    /**
      The positional arguments to pass to the blueprint.
    */
    positional?: string[];
    /**
      The named arguments to pass into the blueprint.
    */
    named?: {
      [key: string]: string;
    };
  }
}
```

text/0477-blueprints-update.md
kellyselden: At the time of writing this, I thought I would need to know the difference between the two to do the correct thing, but after implementing most the functionality, I don't think that's true anymore.

kellyselden:
> The file should be in the configured config directory (from package.json's ember-addon.configPath falling back to config/)

Good idea. Is there precedence for how a third party could use the same logic to detect this? Filed https://github.com/ember-cli/ember-cli-update/issues/675.

> We should include a version value for the file format itself (a top level key for configVersion or some such), so that we aren't absolutely completely locked down on the final format out of the gates

Great idea. Filed https://github.com/ember-cli/ember-cli-update/issues/676.

As far as the blueprint schema:

`isRoot` is probably not necessary.
Are `name` and `package` ever going to diverge? Or are you thinking a pretty name to show up in the inquirer prompt instead of package name?
`rootPath` is for running local file path blueprints right? If so, I'm currently using `location`, which could be expanded to `localFilePath` or `locationOnDisk`.
After experimenting, I don't think `options` needs to be anything more complicated than an array of strings to be `.join(' ')`ed, or maybe just a string that is passed as-is.
We need a property that is only used for default ember-cli blueprints, `type` or `defaultType` that is either `app` or `addon`. It's the only special property needed to tell ember-cli to use `ember new` or `ember addon`. Every other blueprint always goes through `ember new`.

rwjblue:
> Are name and package ever going to diverge?

Yes, there are many packages that provide more than one blueprint. For example, `ember-cli` provides quite a few (including the most popular ones like `app`, `addon`, and `in-repo-addon`).

> rootPath is for running local file path blueprints right?

No, the thing I was thinking was to record the location the blueprint was ran from. For example, you could imagine running `ember g component foo-bar` from within an in-repo-addon. I would expect that most often `rootPath` would be `.`.

> After experimenting, I don't think options needs to be anything more complicated than an array of strings to be .join(' ')ed, or maybe just a string that is passed as-is.

I think it should at least be an array of strings (not a single string). Most mechanisms to exec prefer accepting an array of arguments (otherwise you have to boot a shell to interpret the single string, and booting a shell will not be as portable across win/*nix/etc).

> We need a property that is only used for default ember-cli blueprints, type or defaultType that is either app or addon. It's the only special property needed to tell ember-cli to use ember new or ember addon. Every other blueprint always goes through ember new.

This is what I meant by `isRoot`. The entry in the blueprints array that has this as `true` is the main project blueprint. Every other blueprint update would be ran _after_ this one.

kellyselden:
If someone runs ember-cli-update and the list says 3 blueprints need updates, do you envision the user could select them all to run sequentially, or one at a time? I currently have the list single item selectable to eliminate ordering issues, but that won't be very ergonomic if they have tons...

kellyselden:
For `ember-addon.configPath`, does that have an equivalent in an app? Can an app configure its config/ folder?

rwjblue:
Yes, the same config location applies (like its still `ember-addon.configPath` for an app).

kellyselden:
`isRoot` is confusing to me. Unlike `rootPath`, it doesn't have anything to do with dirs (to my understanding). It is instead similar to `isPartial` then? The reason I say it doesn't have anything to do with dirs is the idea of partial, but project-wide blueprints. You could store some .travis.yml and .eslintrc.js tweaks in a blueprint. Would `isRoot` be true because it is processed at the root dir, or false because it is not a complete project blueprint?

What sort of uses does `isRoot` solve? Is it an ordering problem? Would blueprints array order also be important? Is it to prevent having two `isRoot`s in the list overwriting each other? What would happen if both an ember app and an ember addon blueprint showed up in the list at the same time?

rwjblue:
In order to generate a valid diff that we can apply while updating, we need to know the order in which to run the blueprints. I was using `isRoot` to signify the "main project blueprint" (e.g. the one that actually creates the project folder), but the name is less important than the overall point here.

---

Imagine you have a set of blueprints that this new system is managing: 

```js
{
  blueprints: [
    {
      isRoot: true,
      name: 'app',
      package: 'ember-cli',
      version: '3.10.0',
      options: ['my-app-name', '--yarn', '--no-welcome'],
    },
    {
      name: 'liquid-fire',
      package: 'liquid-fire',
      version: '0.30.0',
    },
    {
      name: 'ember-cli-typescript',
      package: 'ember-cli-typescript',
      version: '2.0.0',
    }
  ]
}
```

We then need to run the following in order to create the "from" state (to set up the diff):

```
npx ember-cli@3.10.0 new my-app-name --yarn --no-welcome --blueprint app
cd my-app-name
yarn add -D liquid-fire@0.30.0 ember-cli-typescript@2.0.0
ember generate liquid-fire
ember generate ember-cli-typescript
```

Then to gather the "to" state:

```
npx ember-cli@latest new my-app-name --yarn --no-welcome --blueprint app
cd my-app-name
yarn add -D liquid-fire ember-cli-typescript
ember generate liquid-fire
ember generate ember-cli-typescript
```

---

All of the blueprints **except** the top level (aka "root" one) are treated effectively the same. You install a package, then run its generator. The root blueprint is _special_: it creates the project folder, it must be ran first, etc.

I'm 100% happy to find a better name than "root" here, but I do think the distinction of which blueprint is the root one is important and should be labeled in the config file format.

kellyselden:
@rwjblue That's a pretty good approach. How would we handle partial app blueprints? These would follow the libkit or octane blueprint conventions. They are not addons. They only have a couple files and are applied via `init`. Would it go like

```
npx ember-cli@latest new my-app-name --yarn --no-welcome --blueprint app
cd my-app-name
ember init -b my-custom-blueprint --random-option
cd my-app-name
yarn add -D liquid-fire ember-cli-typescript
ember generate liquid-fire
ember generate ember-cli-typescript
git add -A
git commit -m"to"
git tag to
```

In this case, you need to know if it is root to run it first, and you also need a flag to know whether it's an addon or not.

Another thought, if they have an ember-cli update and an addon update, and they only choose the addon update, should we use lastest ember-cli or their version to use as the base?

Third thought: a potential pitfall of doing them all at once. If two addons give you a completely new .eslintrc.js file for example. If you were to do the updates one at a time, you would get an idea what was happening in either case. If we ran them both at once for you, last one wins would be all you would get. I don't think this is something we can solve though.

rwjblue:
> In this case, you need to know if it is root to run it first, and you also need a flag to know whether it's an addon or not.

Why do we need to know if it’s an addon? The only difference (AFAIK) between `ember new` and `ember addon` is the blueprint option that is used. See [here](https://github.com/ember-cli/ember-cli/blob/2959aef3f66dfe63435d2e2be11051dd81c517b5/lib/commands/addon.js#L12). 

In other words, instead of `ember addon` you can do:

```
ember new name-here --blueprint addon <options here>
```

rwjblue:
> If two addons give you a completely new .eslintrc.js file for example. If you were to do the updates one at a time, you would get an idea what was happening in either case. If we ran them both at once for you, last one wins would be all you would get. I don't think this is something we can solve though.

Ya, agree. Though I think we could solve it by offering a way to run a single blueprint at a time. The default would be the thing we’ve been talking about (run them all in sequence, then diff, then apply), but we would also be able to have a command to run just one.

kellyselden:
> ember new name-here --blueprint addon <options here>

woah good to know!

runspired:
What I want is for `ember-cli` to provide a file that states whether for a specific `package+version` whether an install command has been run.

It would then be up to addons to use that in a standard `npm/yarn` `post-install` hook to decide whether or not to re-run a blueprint/installation/upgrade process. Basically this is "default blueprints" but more universal. We could even provide a little bit of tooling for bootstrapping just such a command into an addon.

I like that this RFC introduces a place in a project under version control for tracking just this!

What I dislike about this RFC is it seems to get us further away from the broader community's tooling expectations.  I shouldn't need to know that for project-x I need to install a new dependency with `ember install` or upgrade it with `ember upgrade` but for others `npm/yarn install` or `yarn upgrade` is correct.

kellyselden:
I'm struggling to figure out how best to store the options. The goal is to both regenerate CLI commands and match codemod requirements. For example, you want to be able to generate

```
ember new my-app --no-welcome
```

but you also want to be able to target a codemod with requirements like

```json
  {
    "versions": {
      "ember-cli": "2.16.0-beta.1"
    },
    "projectOptions": ["app", "addon", "welcome"],
    "nodeVersion": "4.0.0",
    "commands": ["ember-modules-codemod"]
  }
```

The codemod schema is going to need some work, but the idea is you want to gate the codemod on the presence of options.

If we store CLI arguments verbatim

```json
    {
      "name": "ember-cli",
      "version": "3.2.0-beta.1",
      "options": {
        "positional": [
          "--no-welcome"
        ]
      }
    }
```

we are not going to match any codemod requirements that don't know about the `--no-welcome = --welcome=false` rule. This is also a problem for aliases.

We could either mirror ember-cli's argument parsing, which is going to be brittle, or nudge people somehow to store normalised options in the config so that codemods can match them.

kellyselden:
I'm starting to flesh out more ideas here https://github.com/ember-cli/ember-cli-update/wiki/Config-Schema

kellyselden:
I added the latest config schema that I've been iterating on [here](https://github.com/ember-cli/ember-cli-update/wiki/Config-Schema). I think this RFC is now ready to progress. I implemented all the features behind opt-in commands, and I've extensively tested them in the real world. The config schema now feels solid with room to grow.

rwjblue:
@kellyselden - Awesome! Would you mind opening an issue to track feedback (and include instructions on how to leverage the new commands and whatnot)? I'd love to kick the tires 😸

kellyselden:
@rwjblue Here you go https://github.com/ember-cli/ember-cli-update/issues/812.

kellyselden:
@rwjblue you mentioned storing the config in `ember-addon.configPath`. Is it expected that an addon would have it's blueprint config stored in `tests/dummy/config/ember-cli-update.json`? That is not where I would expect because it has nothing to do with the dummy app or testing.

rwjblue:
Yes, the top level `config` is generally for stuff that is used when consumed. `tests/dummy/config` is where the addon's own configuration should go. I'm totally game to change that around a bit (e.g. add an `addon-config/` or something in the top level or remove the top level `config/environment.js` and saying that `config/` _is_ for the addon's own config) but thats a bit of a separate issue.

kellyselden:
Now that Octane is out and the Holidays are over, I'd love to make progress on this. Anything remaining to be addressed? This is already extremely useful in my various projects:

// existing ecosystem support (default addon blueprints)
https://github.com/ember-vr/ember-aframe/blob/master/tests/dummy/config/ember-cli-update.json
https://github.com/ember-vr/ember-aframe-shim/blob/master/tests/dummy/config/ember-cli-update.json

// partial ember blueprints
https://github.com/kellyselden/ember-extra-blueprint
https://github.com/kellyselden/ember-a11y-blueprint

// tracking custom blueprints (libkit)
https://github.com/tildeio/ts-std/pull/10
https://github.com/cross-check/cross-check/pull/32

// used as node templates (non-ember)
https://github.com/kellyselden/standard-node-template
https://github.com/kellyselden/node-template
https://github.com/kellyselden/ember-cli-test-server/blob/master/config/ember-cli-update.json

cc @rwjblue @wycats

NullVoxPopuli:
just saw a demo of this. I'm a fan :+1:

rajasegar:
@NullVoxPopuli Can you please share the demo link here

NullVoxPopuli:
I cannot -- it was an internal / work demo

kellyselden:
I think it's about time I put this into FCP. I'll ask someone to merge it in a week, or else just close it.

rwjblue:
Sorry @kellyselden, I forgot to add the label when you pinged.

FYI - We are moving this into final comment period, and barring new issues being discovered will merge it in about a weeks time.

kellyselden:
@rwjblue Do you mind merging this?

kellyselden:
The feature was released, so can someone please merge or close this.

locks:
Whoops! Merging it :)

locks:
As always, thanks to everyone that gave us their input to improve the proposal!

PR #478, opened 2019-04-18
Tracked Properties Updates

pzuraq:
[Rendered](https://github.com/emberjs/rfcs/blob/tracked-properties-updates/text/0478-tracked-properties-updates.md)

### DIFF
--- a/text/0478-tracked-properties-updates.md
+++ b/text/0478-tracked-properties-updates.md
@@ -1112,19 +1112,18 @@ set(maybeProxy, 'firstName', 'Amy');
 
 ## Drawbacks
 
-> Why should we _not_ do this? Please consider the impact on teaching Ember,
-> on the integration of this feature with other existing and planned features,
-> on the impact of the API churn on existing apps, etc.
-
-> There are tradeoffs to choosing any path, please attempt to identify them here.
+- The interop story here may a bit confusing for users at first. `@watchable`
+  should only be used in some cases, and it could unclear when it should be
+  used. Documentation should help alleviate this, along with clear examples.
+- We're introducing a decorator that will eventually be deprecated and removed
+  as part of this process, which is essentially some tech debt we're taking on.
+  However, we know that this has a timeline for removal, and it is purely a
+  temporary measure for interop, so it's not a significant amount of debt to
+  take on in the meantime.
 
 ## Alternatives
 
-> What other designs have been considered? What is the impact of not doing this?
-
-> This section could also include prior art, that is, how other frameworks in the same domain have solved this problem.
-
-## Unresolved questions
-
-> Optional, but suggested for first drafts. What parts of the design are still
-> TBD?
+- We could not provide `@watchable` instead. This would mean there isn't really
+  an interop path for users who want to depend on native getters from CPs and
+  observers, leaving a large gap that could prevent users from updating
+  altogether.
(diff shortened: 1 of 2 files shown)


jonnii:
> It's arguable that computed properties that are relying on these caching semantics are problematic in general. After all, it's strange to setup state like this during construction, usually you would use a CP instead, and if CPs are trying to use a value, it generally means that value should be a dependency. 

I understand what's happening here but my issue with the fix is it seems like it requires you to put the `@watchable` attribute on the class which is doing _the right thing_. If the behavior is fixed there would be no feedback that this attribute is now no longer required. Perhaps using `@watchable` should also come with a deprecation warning or a linting hint to encourage changing the flow to be a cp/dependency and/or migrate to native classes?

pzuraq:
We tend to do deprecations as separate RFCs from introducing new features. Computed properties _aren't_ officially deprecated yet, though they likely will be sometime in the future. Until then, I don't think we should provide a deprecation warning for users of `computed`.

A linting rule may make more sense here. There is precedent with the "avoid-observers" eslint rules, for instance. I don't believe we need an RFC to create a rule like that though.

rwjblue:
> I don't believe we need an RFC to create a rule like that though

Agree, we can create the rule without any RFC type things. We'd only need to RFC enabling it by default.

text/0478-tracked-properties-updates.md
```diff
@@ -0,0 +1,1130 @@
...
+
+const Profile = EmberObject.extend({
+  // provided on create
+  person: null,
+
+  username: alias('person.fullName'),
+});
+```
+
+`@watchable` would be imported from `@glimmer/tracking` alongside `@tracked`.
```
rwjblue: Seems odd to import from `@glimmer` since its only about interop with Ember features (computed and observers).

MelSumner:
Since this is important to the interop story, we've decided to move this into FCP.

luxferresum:
Basically a `@tracked` is always `@watchable`, right? Why would I then ever need to use `@watchable` and *not* `@tracked`?
From my perspective `@tracked` does no harm if used too much, right?

pzuraq:
`@tracked` cannot and should not be used on native getters and setters. In the long run, all getters and setters should be native, without any decoration.

`@watchable` can _only_ be used on native getters and setters, by contrast. It is only for compatibility, and eventually should be removed.

If we made `@tracked` work on both, it would quickly get confusing - users might think they need to mark _everything_ as `@tracked`, which they do not, and they will find it much harder to quickly search for code that should be updated and refactored.

text/0478-tracked-properties-updates.md
```diff
@@ -0,0 +1,1131 @@
...
+In classic Ember, all property access had to go through these two methods. Over
+time, these rules have become less strict, and now they have been minimized to
+just a few cases. In general, in a modern Ember app, you shouldn't need to use
+them all that much. As long as you are marking your properties as `@tracked`,
+autotracking should automatically figure out what needs to change, and when.
+
+However, there still are two cases where you _will_ need to use them:
+
+- When accessing and updating plain, undecorated properties on objects
+- When using Ember's `ObjectProxy` class
```
hjdivad: Missing a third case: manual proxies that implement `unknownProperty` and `setUnknownProperty`

mixonic:
As we've discussed a `@classic` decorator in [RFC 468](https://github.com/emberjs/rfcs/pull/468) and this is also viewed as a somewhat temporary compatibility/migration path item, I'll float `@classicDependency`

PR #481, opened 2019-04-23
Component Templates Co-location

chancancode:
[Rendered](https://github.com/emberjs/rfcs/blob/co-location/text/0481-component-templates-co-location.md)

### DIFF
--- a/text/0481-component-templates-co-location.md
+++ b/text/0481-component-templates-co-location.md
@@ -1,0 +1,647 @@
+- Start Date: 2019-04-12
+- Relevant Team(s): Ember.js, Ember CLI
+- RFC PR: (after opening the RFC PR, update this with a link to it and update the file name)
+- Tracking: (leave this empty)
+
+# Component Templates Co-location
+
+## Summary
+
+We propose to place a component's JavaScript file (the class) and its template
+in the same directory on the file system.
+
+Today:
+
+```
+app
+├── components
+│   ├── just-class.js
+│   ├── my-widget.js
+│   ├── nested
+│   │   └── another-widget.js
+│   └── ...
+├── models
+│   └── ...
+├── routes
+│   └── ...
+├── services
+│   └── ...
+├── templates
+│   ├── components
+│   │   ├── just-template.hbs
+│   │   ├── my-widget.hbs
+│   │   ├── nested
+│   │   │   └── another-widget.hbs
+│   │   └── ...
+│   ├── some-route.hbs
+│   └── ...
+└── ...
+```
+
+Proposed:
+
+```
+app
+├── components
+│   ├── just-class.js
+│   ├── just-template.hbs
+│   ├── my-widget.hbs
+│   ├── my-widget.js
+│   ├── nested
+│   │   ├── another-widget.hbs
+│   │   └── another-widget.js
+│   └── ...
+├── models
+│   └── ...
+├── routes
+│   └── ...
+├── services
+│   └── ...
+├── templates
+│   ├── some-route.hbs
+│   └── ...
+└── ...
+```
+
+## Motivation
+
+Today, a component's JavaScript file is located in the `app/components`
+directory. On the other hand, it's template is located in the
+`app/templates/components` directory.
+
+This design decision is a relic of Ember's pre-1.0 days. In those early days,
+before components were even conceived, "views" were the primary building blocks
+in Ember's UI programming model. Like components, they consist of a view class
+located in `app/views/$NAME.js` (subclassing `Ember.View`), and/or a template
+located in `app/templates/$NAME.hbs`. In other words, every template in Ember
+could optionally have an associated view class, and vice versa.
+
+Similar to components, they can be "invoked" in other templates using either
+the `{{view "$NAME"}}` or `{{#view "$NAME"}}...{{/view}}` syntax. At the time,
+this was how Ember developers create reusable pieces of UI in their app.
+
+Unlike components, when a view is "invoked", the view instance does not become
+the `this` context of its template. Instead, the `this` from the calling side
+will persist (similar to partials), while properties on the view instance are
+accessible from within the template with `{{view.someProperty}}` syntax.
+
+Over time, this programming model has gone through several iterations of
+refinements (remember the short-lived "controls"? 😀), eventually evolved into
+the component-based programming model as we know it today.
+
+When components were introduced, it was decided that they would live in a
+separate "namespace" than views: their JavaScript files are located in
+`app/components` instead of `app/views`, and it uses the now-familiar curly
+invocation syntax instead of the `view` keyword.
+
+This decision causes two problems. First, given a template file, such as
+`app/templates/foo-bar.hbs`, how do we know if it belongs to a component or a
+view? Furthermore, since they live in different namespaces, it is totally legal
+to have unrelated components and views with the same names, which could result
+in naming collisions if their templates were both located in `app/templates`.
+To solve this problem, component templates were also given a separate namespace
+`app/templates/components`, as opposed to sharing `app/templates` with views.
+
+Over the course of the Ember 1.x release series, components have taken over as
+the primary UI programming primitive. Views were eventually deprecated and
+removed from Ember's API entirely in Ember 2.0. Despite that, the ramifications
+of that era are still around us.
+
+We argue that the current file system layout for component files is one such
+example. Given how much Ember has evolved since its early days, most of the
+design constraints we had at the time no longer apply. On the other hand, new
+features in the framework – such as the introduction of Ember CLI and the addon
+ecosystem – has brought new capabilities and constraints to the table. As a
+result, this design has caused a lot of "ergonomics papercuts", which we have
+enumerated below. While each of these may feel inconsequential on their own,
+together they cause a lot of unnecessary friction and feel out-of-place in the
+modern Ember experience.
+
+### 1. Component class/template coupling
+
+The JavaScript class and template of a component are not just _related_ but
+fundamentally _coupled_. Together, they provide the whole implementation of
+the component. It is impossible to understand one of them without the other.
+When one file is changed, the other likely need to be updated as well. Placing
+these files physically far apart on the file system fails to communicate this
+tight coupling, in addition to causing a lot of inconveniences when navigating
+a project.
+
+### 2. No single enumeration of components
+
+Given the importance of components in modern Ember apps, when navigating a
+large-scale or otherwise unfamiliar project, some of the most common task for
+Ember developers is to enumerate all components from the given app or to
+determine whether there exists a component with a specific name. The former is
+useful for getting a general sense of the scope, scale and coding style, and
+the latter is sometimes necessary to disambiguate components from helpers or
+to determine whether a component is part of the app or provided by an addon.
+
+Unfortunately, under the current filesystem layout, neither `app/components` or
+`app/templates/components` provide a complete view for those purposes. With
+template-only components, it is possible to define a component without a class,
+therefore a casual scan of `app/components` will not discover those components.
+With the release of Octane, we expect template-only components to become more
+common. Likewise, albeit less common, it is also possible to define a component
+with *only* a class, making `app/templates/components` potentially incomplete
+also.
+
+Therefore, the "correct" way to explore the components in an app is to mentally
+merge the contents of both folders, which is difficult and counterproductive.
+
+### 3. Deviation from route template conventions
+
+In the view-centric days of Ember, `app/templates` was a "kitchen-sink" folder
+without much internal structure – it mixes route templates with reusable view
+templates, as well as component templates, each of which could be grouped into
+their own deeper nesting structure.
+
+With Ember 2.0 removing views, and therefore view templates, `app/templates`
+now has a clear purpose and internal structure: each template file corresponds
+to a route, and each folder corresponding to a nested route. For example,
+`app/templates/post.hbs` is the template for the top-level `post` route, and
+`app/templates/admin/authors/new.hbs` is the template for `admin.authors.new`.
+
+Of course, there is one prominent exception to this: `app/templates/components`
+breaks the otherwise clear convention. Unlike other folders in `app/templates`,
+`app/templates/components` does not represent a nested route. Furthermore, its
+internal structure (if any) also differs from its siblings: folders represent
+arbitrary logical/domain groupings, rather than sub-routes.
+
+Since `app/templates/components` did not start with a special character (e.g.
+`app/templates/-components`), it shows up in the middle between its routes
+siblings in most apps, due to lexicographical sorting of the filenames.
+
+This inconsistency causes confusion and prevents us from teaching the otherwise
+clear convention crisply.
+
+### 4. Lack of single import
+
+Even though the JavaScript class and template of a component are fundamentally
+coupled and inseparable pieces, this relationship is only very loosely encoded
+in Ember today. In fact – the only thing that ties them together is their name:
+when Ember tries to render a component, it will separately lookup the class and
+template by name through the resolver. These elements can both be present at
+the same time, or in the cases of template-only or class-only components, one
+of the resolutions will return a negative result. Only when both elements are
+missing would Ember declare the component missing and raise an error.
+
+This setup falls short when designing components meant for subclassing. For
+example:
+
+```js
+// app/components/foo-bar.js
+
+import MyParentComponent from './my-parent';
+
+export default class FooBarComponent extends MyParentComponent {
+  // ...
+}
+```
+
+As far as Ember can tell, this file defines a class-only component by the name
+of `foo-bar`. The fact that it inherits from `my-parent` is an irrelevant and
+unobservable detail. While the developer may expect this to also inherit the
+parent component's template (`app/templates/components/my-parent.hbs`), Ember
+has no way to know that the two are related.
+
+Because of this problem, addons typically use the `layout` property from the
+classic component API to get around this problem. When generating a component
+in an addon, this is the default output from the blueprint:
+
+```js
+// addon/components/foo-bar.js
+
+import Component from '@ember/component';
+import layout from '../templates/components/foo-bar';
+
+export default Component.extend({
+  layout
+});
+```
+
+```hbs
+{{!-- addon/templates/components/foo-bar.hbs --}}
+
+{{yield}}
+```
+
+```js
+// app/components/foo-bar.js
+
+export { default } from 'my-addon/components/foo-bar';
+```
+
+First, within the `addon` folder, there is the component class and template.
+Having these files in the `addon` folder allow them to be imported from
+JavaScript (e.g. `import FooBar from 'my-addon/components/foo-bar';`) for
+subclassing. However, as mentioned above, this does not also bring along the
+component's template.
+
+To avoid developers having to also import the template, the addon component
+blueprint sets the `layout` property on the component's class. At runtime,
+when Ember fail to find a template associated with the component, it will
+then fall back to this property.
+
+Finally, it re-exports the component in the `app` folder, which is merged with
+the app's `app` folder, allowing the component to be resolved by the resolver
+and invoked from Handlebars.
+
+This workaround gets the job done, but has several major drawbacks.
+
+First, it adds a significant amount of boilerplate to addon components, which
+could be puzzling even to seasoned addon developers. In fact, the problem it
+solves is unique to addons – there is nothing wrong with using inheritance in
+apps – we just didn't think we could justify adding this amount of boilerplate
+to every app and putting them in Ember's "happy path".
+
+Second, the `layout` property in classic component uses a feature in Glimmer VM
+called "late-bound layout", in which we must assume a different template for
+each _instance_ of the component (the `layout` property can be set in `init`).
+This is fundamentally more difficult to optimize and excludes the component
+from key optimizations (and opportunities for future optimizations).
+
+Most importantly, Glimmer components do not have the equivalent of the `layout`
+property, meaning that addons (or components designed for inheritance in mind)
+cannot be written using Glimmer components today.
+
+To solve these problems once and for all, we need to change things such that
+component templates are automatically associated with their JavaScript class at
+build time, and importing the JavaScript class should result in a value that
+has this metadata associated with it, which needs to survive subclassing as
+well. Solving this problem is also essential for unblocking template imports.
+
+We believe the technical design for allowing co-location will solve this
+problem nicely.
+
+## Detailed design
+
+### High-level design
+
+We propose to allow placing a component's template adjacent to its JavaScript
+file in `app/components`. For example, for a component named `foo-bar`, it will
+be `app/components/foo-bar.js` and `app/components/foo-bar.hbs`.
+
+In addition, per the node resolution convention, we propose to allow `index`
+files inside a directory to have the equivalent semantics. In the example
+above, it could also be structured as `app/components/foo-bar/index.js` and
+`app/components/foo-bar/index.hbs`. This allows additional files related to the
+component (such as a `README.md` file) to be co-located on the filesystem.
+
+For template-only components, they can be either `app/components/foo-bar.hbs`
+or `app/components/foo-bar/index.hbs` without a corresponding JavaScript file.
+
+Similarly, for addons, templates can be placed inside `addon/components` with
+the same rules laid out above.
+
+In all of these case, if a template file is present in `app/components` or
+`addon/components`, it will take precedence over any corresponding template
+files in `app/templates`, the `layout` property on classic components, or a
+template with the same name that is made available with the resolver API.
+Instead of being resolved at runtime, a template in `app/components` will be
+associated with the component's JavaScript class at build time.
+
+### Low-level primitives
+
+We propose to introduce the following low-level APIs:
+
+* The `setComponentTemplate` function takes two arguments, the first being the
+  pre-compiled (wire-format) template, the second being the component class. It
+  transparently associates the given template with the component class in a way
+  can be retrieved later with the `getComponentTemplate` function described
+  below. For convenience, `setComponentTemplate` will return the component
+  class (the second argument).
+
+* The `getComponentTemplate` function takes a component class and returns the
+  template associated with the given component class, if any, or one of its
+  superclasses, if any, or `undefined` if no template association was found.
+
+This is one possible way to implement these functions:
+
+```js
+function setComponentTemplate(template, componentClass) {
+  Object.defineProperty(componentClass, "__template__", {
+    configurable: true,
+    enumerable: false,
+    writable: false,
+    value: template
+  });
+
+  return componentClass;
+}
+
+function getComponentTemplate(componentClass) {
+  return componentClass.__template__;
+}
+
+class Foo {}
+class Bar extends Foo {}
+class Baz extends Bar {}
+
+class Bat {}
+
+
+// USAGE
+
+setComponentTemplate(precompile("foo template"), Foo);
+setComponentTemplate(precompile("baz template"), Baz);
+
+getComponentTemplate(Foo); // => foo tempalte
+getComponentTemplate(Bar); // => foo tempalte (inherited)
+getComponentTemplate(Baz); // => baz tempalte (overridden)
+
+getComponentTemplate(Bat); // => undefined
+```
+
+In an actual implementation, we would probably want to avoid polluting the
+component class with a string key (`__template__` in this example) and use a
+`Symbol` or `WeakMap` based strategery.
+
+For performance reason, changing the template is not allowed once set. It is
+also illegal to call `setComponentTemplate` on a component class that has
+already been rendered, or once `getComponentTemplate` has been called on it.
+Together, these rules ensure the results from calling `getComponentTemplate`
+can be reliably cached, either by the internals of `getComponentTemplate`
+itself or by one of its callers.
+
+While the default export of a component's JavaScript file is usually a class,
+it is not a strict requirement with custom component managers. To accommodate
+this, `setComponentTemplate` can be passed any JavaScript `Object`.
+
+### Build-time transformations
+
+The low-level `setComponentTemplate` and `getComponentTemplate` APIs are not
+intended to be called by end-users directly, even though they will be public
+and part of our semver stability guarantee. Instead, they are meant to be used
+primarily in the output emitted by build tools. At build time, any template
+files found in `{app,addon}/components` will be inlined into the component's
+JavaScript file and removed from the build output. For example, given these
+files on disk:
+
+```js
+// app/components/foo-bar.js
+
+import Component from '@ember/component';
+
+export default Component.extend({
+  // ...
+});
+```
+
+```hbs
+{{!-- app/components/foo-bar.hbs --}}
+
+foo bar!
+```
+
+The build output will be something to the effect of this JavaScript file:
+
+```js
+// app/components/foo-bar.js
+
+import Component, { setComponentTemplate } from '@ember/component';
+
+// output of precompile("foo bar!")
+const TEMPLATE = {
+  id: "...",
+  block: "...",
+  meta: { moduleName: "app/components/foo-bar" }
+};
+
+const CLASS = Component.extend({
+  // ...
+});
+
+export default setComponentTemplate(TEMPLATE, CLASS);
+```
+
+The variables are named here for clarity, but the actual build output would
+be careful to avoid introducing hygiene issues and other observable semantic
+changes to the JavaScript file.
+
+One caveat here is that each component JavaScript file should export a value
+that is unique to that file. For example, this should be **avoided**:
+
+```js
+// app/components/foo-bar.js
+
+import MyParentComponent from './my-parent';
+
+// BAD: don't do this!
+export default MyParentComponent;
+```
+
+This is problematic because `setComponentTemplate` will be called on
+`MyParentComponent` directly, affecting the parent component and all of its
+descendants. This can be avoided by subclassing, even when no customization is
+required:
+
+```js
+// app/components/foo-bar.js
+
+import MyParentComponent from './my-parent';
+
+// GOOD: do this instead!
+export default class extends MyParentComponent {}
+```
+
+Most cases of this problem can be linted against easily.
+
+### Template-only components
+
+For template-only components, we propose to introduce the following low-level
+API:
+
+* The `templateOnlyComponent` function takes no arguments and produces a unique
+  value that can be used to represent a template-only component.
+
+Again, this function, though public, is not intended to be called by users
+directly. It is primarily used in the output emitted by build tools. When a
+template is found in `{app,addon}/components` but without a corresponding
+JavaScript file, the build output will be something similar to the following:
+
+```js
+// app/components/foo-bar.js
+
+import Component, { setComponentTemplate } from '@ember/component';
+import templateOnlyComponent from '@ember/component/template-only';
+
+// output of precompile("foo bar!")
+const TEMPLATE = {
+  id: "...",
+  block: "...",
+  meta: { moduleName: "app/components/foo-bar" }
+};
+
+const CLASS = templateOnlyComponent();
+
+export default setComponentTemplate(TEMPLATE, CLASS);
+```
+
+In addition to build tooling, addon authors may also find this function
+useful. Currently, the only way to create a "true" template-only component
+is by enabling an optional feature in the app. Since addons cannot assume the
+value of that flag in the consuming app, they currently cannot take advantage
+of the feature. By providing this function, addon authors can work around this
+problem by explicitly defining a JavaScript file with `templateOnlyComponent()`
+as the default export.
+
+On the other hand, app developers, should _not_ have any reason to use this
+function directly in their app, since they could just as easily enable the
+optional feature across the board.
+
+### Codemod
+
+A codemod will be provided to seamlessly migrate component templates.
+
+Such a "codemod" will essentially just merges `app/templates/components` into
+`app/components`. For a quick "taste test" of what the resulting tree will look
+like in your app, the following commands give a close approximation:
+
+```bash
+$ rsync --archive --remove-source-files app/templates/components/ app/components/
+$ rm -rf app/templates/components/
+```
+
+Of course, the resulting output won't "work", but it can be useful for getting
+a sense of what it's like to work with the new layout on editors, Github, etc.
+
+### Generator
+
+We propose to make some updates to the components generator to accompany this
+change.
+
+1. It should accept a `--component-class` option. This controls which base
+   class is used for that component and whether native classes are used. The
+   legal values for this option are `@glimmer/component` (aliased as `-gc`),
+   `@ember/component` (aliased as `-ec`), `@ember/component/template-only`
+   (aliased as `-tc`) or an empty string (aliased as `--no-component-class` and
+   `-nc`).
+
+   The latter two differ in that `@ember/component/template-only` would
+   generate an explicit JavaScript file with `templateOnlyComponent()` as the
+   default export, which is useful for addons, whereas `--no-component-class`
+   would skip generating a JavaScript file altogether.
+
+   This option may be extended in the future to allow other custom components
+   to provide their own blueprints, but for now, passing anything other than
+   the allowed values will result in an error.
+
+2. It should accept a `--component-structure` option.
+
+   When this option is set to `flat` (aliased as `-fs`), the component's JavaScript and template files will both be generated at the root of
+   `{app,addon}/components`.
+
+   When this option is set to `nested` (aliased as `-ns`), a folder will be
+   generated for the component in `{app,addon}/components`, and the component's
+   JavaScript and template files will be generated as `index.{js,hbs}` inside
+   the folder.
+
+   When this option is set to `classic` (aliased as `-cs`), the component's
+   template file will be generated in `{app,addon}/templates/components`. For
+   addons, when used with `--component-class=@ember/component`, this will also
+   emit the `layout` property workaround.
+
+3. Thses options will default to `--component-class=@ember/component` and
+   `--component-structure=classic` for backwards compatibility.
+
+   However, the default values can be overridden in `.ember-cli` as usual, and
+   teams are encouraged to do so as they see fit. Due to a limitation in how
+   the system works, the names for these options are chosen such that they are
+   unlikely to conflict with options on other `ember` commands, which is why
+   they are a bit verbose.
+
+   For Octane apps, the default app blueprint will include a `.ember-cli` file
+   that defaults to `--no-component-class` and `--component-structure=flat`.
+   The guides and documentation will assume these settings going forward.
+
+### Timing
+
+We propose to include this change as part of the Octane edition. Admittedly,
+this is a rather late addition. However, there are a few compelling reasons for
+including it:
+
+1. Module Unification was a much-anticipated feature that
+   [did not make it](https://blog.emberjs.com/2019/03/11/update-on-module-unification-and-octane.html)
+   into the Octane edition. Even though it tries to solve a lot of other
+   problems, component co-location is one of the main benefits. This proposal
+   allows us to incrementally deliver a solution one of the biggest pain points
+   while we work out the other issues.
+
+2. The Octane edition will enable template-only components by default, which
+   will encourage its use and intensifies some of the issues enumerated in the
+   motivation section.
+
+3. The Octane edition will highlight Glimmer Components as the main programming
+   model. Without the `setComponentTemplate` API proposed in this RFC, addons
+   authors will not be able to adopt Glimmer Components. This will, over time,
+   significantly increase the barrier to authoring addons, as most Ember
+   developers will no longer be familiar with the classic components API.
+
+4. The Octane edition presents a good opportunity to update the documentation
+   and get the whole community on board with adopting a feature like this. It
+   is important that build tools, linters, codemods, etc take this feature into
+   account, and shipping this as part of the Octane edition is a good way to
+   signal that.x
+
+Even though it is pretty late in the Octane timeframe, we believe this feature
+is relatively limited in scope (compared to, say, `@tracked`) and can be
+implemented relatively quickly.
+
+## How we teach this
+
+As mentioned above, the learning team will update the guides and API docs to
+assume the "flat" co-located layout.
+
+With this layout, it should be much easier for new users to form a mental model
+around components. We can start by teaching that the most basic component is
+just a reusable piece of markup (template-only components), but can "upgraded"
+to have dynamic content by taking arguments using the `@name` syntax, and
+further "upgraded" to keep internal states by creating a JavaScript file next
+to the template and finally add interactivity with element modifiers.
+
+Because template-only component is such a light-weight concept, it is arguably
+not necessary to separate out the topic of "templates" from "components" in the
+guides, but this can be addressed in a future revision of the guides.
+
+We will teach that the `templates` folder is used for route templates. This can
+be introduced at the same time as the `routes` and `controllers` folder, which
+would come at a later time than components.
+
+## Drawbacks
+
+Unlike the rest of the Octane features, this is not as battle-tested. Also, as
+mentioned above, this addition would be quite last-minute. However, it is quite
+normal to discover minor problems (e.g. issues with using Glimmer components in
+addons today) through real-world usage, and that is precisely the point of the
+Octane preview period. Since Octane is all about polish and cohesion, we think
+it is worth addressing these issues before stamping on the quality seal.
+
+Another drawback is that it only address the co-location issue for components,
+not other related types like [[route, controller, route template]] and [[model,
+adapter, serializer]], or even co-location of tests. However, we believe the
+situation with components is unique enough (see the motivation section) that
+they are not merely *related*, but *coupled*. That, along with the fact that
+components are much more common, sets them apart from the rest and justifies
+solving the problem first.
+
+There is a small risk that we will subject the community to another migration
+when we finalize the replacement of Module Unification (the "New File Layout").
+However, we feel pretty confident that regardless of where the collection of
+components will end up on disk (e.g. `src/ui/components`), the internal
+structure of that collection will closely match what is proposed in this RFC.
+Ultimately, we expect there to be automatic migrators for these kinds of
+changes anyway, so the cost of the possible churn is contained.
+
+## Alternatives
+
+We can forgo the opportunity to address these problems as part of Octane, and
+instead land this proposal (or something like it) after the edition has
+shipped. That would come at a cost of addon authors not being able to adopt
+some of the Octane features until we find alternative solutions.
+
+## Unresolved questions
+
+* Where would the build tool changes be implemented? And how?
+* How should the build tool changes interact with "pods" and `template.hbs`?
+* How should the new generator options interact with "pods"?
--- a/text/0481-component-templates-co-location.md
+++ b/text/0481-component-templates-co-location.md
@@ -556,8 +556,19 @@ change.
 
 ## How we teach this
 
-As mentioned above, the learning team will update the guides and API docs to
-assume the "flat" co-located layout.
+As mentioned above, we will update the learning resources to assume the "flat"
+co-located layout. Throughout the Ember Guides, the Tutorials, and the CLI
+Addon Tutorial, we would update the file paths in all component examples. Since
+we assume the classic layout today, in most cases only the template paths would
+need to be updated. The prose describing the location of files would also need
+to change.
+
+The API documentation will describe the full set of file layout options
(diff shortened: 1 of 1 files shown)


text/0481-component-templates-co-location.md
```diff
@@ -0,0 +1,647 @@
...
+
+We propose to make some updates to the components generator to accompany this
+change.
+
+1. It should accept a `--component-class` option. This controls which base
+   class is used for that component and whether native classes are used. The
+   legal values for this option are `@glimmer/component` (aliased as `-gc`),
+   `@ember/component` (aliased as `-ec`), `@ember/component/template-only`
+   (aliased as `-tc`) or an empty string (aliased as `--no-component-class` and
+   `-nc`).
```
GavinJoyce: Will we be referring to `@ember/component`s as "classic components" once Octane lands? If so, perhaps we should use `-cc` instead of `-ec` for `@ember/component`? This would also mirror the `-cs` classic structure below, "classic" would then consistently mean the old way of doing things.

GavinJoyce:
I'm very excited to see this RFC. 

A constant pain point for new people joining our company and working on our app and with Ember for this first time has been the relative difficulty and mental load of navigating between component `.hbs` and `.js` files. This RFC will significantly improve that pain point. It will also make teaching Ember easier as the file layout rules will be much simpler and will no longer require a detailed explanation.

This also seems like a nice incremental step towards module unification.

SolPier:
Its not really this RFCs subject, but could it open a way for a controller/template collocation as well ?

As @GavinJoyce stated, these files distance in the project tree is a relative painpoint of learning and practicing ember. As the coupling of controller/template is conceptually the same as components js/hbs, and as the technical solution may be similar (from afar), they may be an opportunity to hit two birds in one shot ? I mean, benefiting of the simplification of the two template+js couplings.

lifeart:
Hm, we use `ember g component component-name --pod` and get same structure (without podModulePrefix)

ggayowsky:
Will this RFC affect applications using pod structured components?

It was unclear to me, but reading the RFC I got the feeling that an application using the pod structure:

```
app
|-- components
|    |-- component-1
|    |    |-- component.js
|    |    |-- template.hbs
|    |-- component-2
|    |    |-- component.js
|    |    |-- template.hbs
...
```

Would need to rename everything to use `index.js` and `index.hbs` i.e.

```
app
|-- components
|    |-- component-1
|    |    |-- index.js
|    |    |-- index.hbs
|    |-- component-2
|    |    |-- index.js
|    |    |-- index.hbs
...
```

SolPier:
Here the js+hbs aren't nested in a directory int the PR.

pods:
```
app/
|-- components/
|    |-- component-1/
|    |    |-- component.js
|    |    |-- template.hbs
```

this PR:
```
app/
|-- components/
|    |-- component-1.js
|    |-- component-1.hbs
```

This is less of a change from the current structure (the component.js doesn't changes at all by example, nor its name nor its location).
Only the template location changes.

text/0481-component-templates-co-location.md
```diff
@@ -0,0 +1,647 @@
...
+
+We propose to allow placing a component's template adjacent to its JavaScript
+file in `app/components`. For example, for a component named `foo-bar`, it will
+be `app/components/foo-bar.js` and `app/components/foo-bar.hbs`.
+
+In addition, per the node resolution convention, we propose to allow `index`
+files inside a directory to have the equivalent semantics. In the example
+above, it could also be structured as `app/components/foo-bar/index.js` and
+`app/components/foo-bar/index.hbs`. This allows additional files related to the
+component (such as a `README.md` file) to be co-located on the filesystem.
```
kovalchik: For the folder based layout, I understand why `app/components/foo-bar/index.js` represents the component’s class file; this generally makes sense from a module import perspective. I’m not sure I agree with `app/components/foo-bar/index.hbs` representing the template. This *kind of* makes semantic sense for template only components, but for non-template only components it creates two identically named files (with the exception of the extension) that both represent the component. This seems like a confusing convention to me, as JS modules typically have one entry point. Once tests are introduced to the folder structure and users update addons that provide functionality like CSS modules or Storybook tests, the index naming convention seems like it will make even less sense. If I were a new user, I'm not sure I would understand why there are two index files in this folder.

```
- foo-bar
  - index.js
  - index.hbs
  - styles.css  // index.css ??
  - tests.js
  - other.js
```

Many other component folder structures I’ve seen that utilize an `index.js` file, do so in the following way. Since Ember hooks up the component `index` for us, I don’t feel that we particularly need to adhere to an `index` naming convention, especially since it creates duplication with the template file name. That being said, I'm not totally opposed to the `index.js`/`index.hbs` proposal; I'm mostly nitpicking at this point. :)

```
- foo-bar
  - index.js  // re-exports component.js
  - component.js
  - template.hbs
```

text/0481-component-templates-co-location.md
```diff
@@ -0,0 +1,647 @@
...
+We believe the technical design for allowing co-location will solve this
+problem nicely.
+
+## Detailed design
+
+### High-level design
+
+We propose to allow placing a component's template adjacent to its JavaScript
+file in `app/components`. For example, for a component named `foo-bar`, it will
+be `app/components/foo-bar.js` and `app/components/foo-bar.hbs`.
```
kovalchik: I have some concerns about supporting components in a folder (`app/components/foo-bar/index.js`) and also components at the top level of the components directory (`app/components/foo-bar.js`).

Which of these structures will the CLI default to? Will there be a config setting to enforce usage of one form over the other on a project-wide basis?

For my own projects, I would personally prefer the folder based approach, so I do appreciate the flexibility but it seems like having two file structure options reduces the consistency and could lead to developer confusion.

In particular, the non-folder based structure seems like it could quickly create a mess of the components directory, especially if other file types are introduced, via further folder structure refinements or addons. Although the following component files are located together, it's not particularly easy to parse what belongs to what.

```
- components
  - first-component.js
  - first-component.hbs
  - first-component.css
  - first-component-tests.js
  - first-component-other.js
  - second-component.js
  - second-component.hbs
  - second-component.css
  - second-component-tests.js
  - second-component-other.js
```

davewasmer:
Thanks for this @chancancode! I really appreciated the historical context in this writeup.

I'm a big fan of co-location, but I'm concerned about this proposal. While I don't disagree with any of the design or implementation, I worry that the larger picture isn't addressed.

### Understated cost of churn

The RFC readily admits this structure might change once again if / when some future form of Module Unification lands:

> There is a small risk that we will subject the community to another migration when we finalize the replacement of Module Unification 

I think this downside is understated here. I'm less worried about the time it takes to convert code between formats, and more worried about the mental churn. Ember already has two (ish) supported filesystem layouts - this RFC adds a third (but for components only!) which we are acknowledging will likely be obsoleted.

If / when Module Unification lands, it seems like a painful story to explain to a new dev coming onto an Ember project: "well, this new app uses MU, which is this whole set of rules, and this other app is still on pods, but not the original pods, the new colocated pods, and this addon is kinda like classic layout ...".

### Not comprehensive

The Drawbacks section mentions that this RFC is exclusively focused on components, and that the narrow focus is justified by the relative prevalence of components and the fact that templates & JS files are coupled:

> That, along with the fact that components are much more common, sets them apart from the rest and justifies solving the problem first.

I worry that this is the wrong framing. The potential drawback here is not which order to solve these in (i.e. components first vs. something else), but rather, solving them without considering the interdependencies. Module Unification, on the other hand, featured a thoroughly designed solution to filesystem layout that treated the project and the developer experience as a whole. I don't see a similar treatment of the broader problem in this RFC.

But I'm not suggesting that we need to go back to top-down, fully designed solutions. I'm a big fan of the direction that the template imports RFC goes: open up primitives that allow community experimentation, and once a stable, battle tested solution emerges, adopt it as the new standard. For the folks that like experimentation, they can play around on the bleeding edge with its associated papercuts. For folks wanting more stability, they can stick with tried and true patterns until the consensus emerges.

This pattern seems even more reasonable when we consider that we already support co-location: pods. The proposed design here certainly feels more ergonomic & performant on some margins, but Ember devs can already get most of the benefits of co-location by simply adopting pods.

This RFC unfortunately feels like an "all of the downsides, none of the upsides" compromise: we don't get the experimentation from the community, but we also don't get the comprehensive solution from top-down designs.

### Delaying

> We can forgo the opportunity to address these problems as part of Octane, and instead land this proposal (or something like it) after the edition has shipped. That would come at a cost of addon authors not being able to adopt some of the Octane features until we find alternative solutions.

This seems like a misunderstanding of editions (perhaps my own)? There's no hard feature lines with Octane - you can adopt the independent features piecemeal. Octane simply marks the "low point of incoherence". Waiting until after it "ships" simply means addon authors simply upgrading their Ember version like normal when the next feature lands.

rwjblue:
While I like the overall design proposed, there are some specific parts that seem problematic to me. The main thrust of my objection is that I believe that the template and component class are inextricably coupled (this RFC seems to agree), but if we accepted this RFC we are left in a fairly incoherent state from the perspective of component manager authors. Specifically, [emberjs/rfcs#213](https://emberjs.github.io/rfcs/0213-custom-components.html) makes it very clear that a component manager should be the only source of information about a given component, but if implemented as proposed the template retrieval system is _also_ making assumptions and retrieving information from the component class.

I believe the correct change to the design here is to:

* Expose a capability for `providesTemplate` (or similar but better name) to the component manager's optional features which allows managers to implement `getTemplate` (in line with the existing `getContext` hook)
* Expose `setTemplate` / `getTemplate` utility functions to do the association (and assertions) proposed in this RFC so that arbitrary component manager authors can leverage them in their own `getTemplate` hooks
* Use the `setTemplate` / `getTemplate` system _by default_ when the `providesTemplate` optional feature is not enabled

chriskrycho:
Two high-level thoughts here:

1.  I share the concerns articulated by both @davewasmer and @rwjblue. Related, the similarity between this proposed structure and pods is unsurprising, but there is no discussion of the cost of a situation where users need to understand possibly all three mental models *in the same project*: it is often difficult to migrate everything at once. The discussion of the codemod (and indeed of the whole proposal) completely elides the existence of pods; and while pods are not a "first-class" feature in some ways, they are in widespread use and are supported output from Ember CLI.

2.  The combination of the way browser module imports work and the *default* (which I think is probably right) for how [Node 12 supports ES Module imports](https://medium.com/@nodejs/announcing-a-new-experimental-modules-1be8d2d6c2ff) (see discussion [here](https://github.com/nodejs/modules/issues/268)) makes me wonder if simply using explicit imports (including extension names!) might be an important design consideration here. It's a difference from what Ember users do and expect today, but there are also a number of (potentially substantial) upsides to explicitly importing with file extension. Given as well the work toward static resolution with Embroider, I'd appreciate there being at minimum a *discussion* of those considerations here.

text/0481-component-templates-co-location.md
buschtoens: I personally also much prefer the "pods" approach, exactly for the reason that it is easier to extend, e.g. `styles.css` files or even further "private" files that are imported by the `component.js`.

simonihmig:
I would second the concerns raised here, especially by @kovalchik and @davewasmer. 

From what I understand this RFC proposes several (in my view different) things:
1. associating a template at build time rather than resolving at runtime (performance)
2. ability to extend a component including its template
3. colocation

What's not obvious to me (and I might very well miss essential things here) is why these things are interdependent. Specifically couldn't there be ways to solve 1. and/or 2. without introducing a new file layout?

Regarding 3., I am certainly a big supporter for colocation. Fwiw, there was just one app we built in the past 5+ years that wasn't using pods. And that was using "MU" 🤷‍♂️😉

But I share the previously voiced concerns (mental churn, learning effort) that we are introducing yet another file layout. And this while we have an existing (pods) and a future (formerly known as MU) solution. And while those feel quite similar, the new proposal differs substantially IMHO: the collocated but flat layout is a new thing, and the nested one (awkwardly IMHO) introduces that `index` convention.

So *assuming* we do have to introduce a new (temporary) way of having colocated components to get the other benefits (see above), I would...
* question why we need to actually introduce two new file layouts (flat and nested). Shouldn't we find agreement on one opinionated way? (hey, convention over configuration)
* strongly opt for the nested version, as 
  * it allows colocating other files easily and ergonomically as mentioned before (styles, tests)
  * prevents having an even longer list of files in your `components/` folder
* change the nested/folder version to the following layout, rather than introducing this `index` thing. Which is *exactly* what we had before with pods, and presumably (according to the MU RFC) what we get with the new file layout (ignoring the `app/components/` base path)
```
├── components
│   └── foo-bar
|      ├── component.js
│      └── template.hbs
```

But at the end I would rather try to focus efforts on getting that thing formerly known as MU to land sooner than later, rather than binding resources (for implementation, documentation, codemods etc.) to a just temporarily used (AFAIU) new file layout.

lifeart:
@rwjblue 
> * Expose `setTemplate` / `getTemplate` utility functions to do the association (and assertions) proposed in this RFC so that arbitrary component manager authors can leverage them in their own `getTemplate` hooks
> * Use the `setTemplate` / `getTemplate` system _by default_ when the `providesTemplate` optional feature is not enabled

Is it possible to get template asynchronously?

`getTemplate -> promise -> someBinaryDataDownloaded from server / compiled in WebWorker .. etc

joukevandermaas:
I don't really understand why Octane has to include file-system changes. I get that the Ember team is trying to ship a really nice new version of the framework that can be marketed and seems "modern", but imo Octane *already* does that. Just because module unification is now out of scope, does not mean it has to be replaced. 

I think doing a last-minute change that has not gone through the same user-testing that the rest of Octane has, is more likely to hurt than to help. My feeling reading this RFC was that it will only confuse things more, because now there are three ways to lay out files, *all* of which have their own confusing inconsistencies. For example, I think routes and especially controllers are also "inextricably coupled" to their templates.

For what it's worth our app has used pods for a few years, and it has never really been a problem for us. If file-system changes are absolutely necessary, perhaps pods can be the default for Octane apps?

text/0481-component-templates-co-location.md
Alonski: For a project with 100 components. This means another 100 files on disk that take up space and need to be parsed during build time even if it is a one liner. 
I actually like having the JS and HBS both be named `index.ext` 
It shows the intent that they are used *together*. :)

chriskrycho:
@rwjblue @chancancode @wycats can one of you elaborate that for those of us who weren't able to be on that conversation? ❤️

rwjblue:
@chriskrycho - Sure, I can give it a shot, sorry if this turns out to be rambly.

My concerns centered around the conceptual encapsulation between the exported custom component (literally the export from the `app/components/foo.js` file) and the rest of the system. It was my position that nothing in the system should make assumptions about what that exported thing is and how to interact with it _other than_ the component manager (note: I still _largely_ agree with that position).

During the conversation I realized/discovered/whatever a few important points:

* In order to support the "bundle compiler" mode (also know as AOT, "binary templates", and likely other names) we need to have a conceptually different "thing" handle compilation than the "browser runtime" component manager. In other words, there are essentially _two_ component manager concepts: one compile time (where is this things template) and one for runtime concerns (give me this things context, update its arguments, etc).
* The system in use by Ember (and soon to be migrated _back_ into Glimmer.js) **already** requires that the manager is associated to the exported custom component (this is what `setComponentManager` does in fact), and therefore there is a required fundamental lookup just to _get_ the manager. 

Those two realizations (the first I just wasn't aware of, and the second one I knew but did think much about) made it clear that the proposal in this RFC for associating the template with the class is no different than the mechanism we already must use to associate the manager. Eventually, we may decide to merge the two things (e.g. :wave: :wave: `setComponentManager` taking both a runtime and compile time factory :wave: :wave:), but _right now_ the only thing needed from the compile time manager is "get this things template". In other words, `setComponentTemplate` is _effectively_ the same thing as a theoretical `setComponentCompileTimeManager` (or `setComponentManager(buildCompileManager, buildRuntimeManager, klass)`).

The benefit of the existing proposal is that it lets move forward addressing the immediate concerns, and then we can iterate easily in the RFC that begins considering bringing "bundle compiler" into Ember.

Hope that mess of thoughts helped..

wycats:
In addition to an overwhelming amount of positive feedback, I'm seeing a few different sources of objections on this thread. I cite @davewasmer not to pick on him (his concerns are shared by several others), but rather because his articulation of the concerns is quite eloquent.

> Ember devs can already get most of the benefits of co-location by simply adopting pods.

I understand the for a lot of Ember developers, pods "already exist", but I want to emphatically state that the details of pods are not the future of Ember, that there are many discrepancies between how pods work today and the form that any future direction for the framework will take, and that ecosystem tooling and library support for pods is spotty and inconsistent *today*.

Because pods are so popular, we absolutely will need a transition path from pods to the updated *official* file layout, of course.

> The RFC readily admits this structure might change once again if / when some future form of Module Unification lands

Module Unification was the last feature that we attempted to land as a "big bang" update that came along with significant changes to the programming model. Features that we have attempted to land in this way have uniformly struggled to cross the finish line (routable components painfully comes to mind).

Going forward, I think it makes sense to improve the file layout as we identify standalone, useful improvements, rather than waiting for a wholesale update that will solve all of our problems.

This does **not** mean that each improvement will result in breaking changes, but it does mean that we might get stuck with the consequences of earlier decisions that we regret later on. In general, attempting to avoid regrets is a recipe for accomplishing nothing, and the long history of substantial Ember changes shows that Ember is no exception to this reality.

This also doesn't mean that we should make wild changes with nary a thought about what might come next. To the extent that we can reasonably predict our next technical steps, we should try to align our current work with those steps. To the extent that we can identify significant goals ahead of time, we should take steps along the way to increase our chance of accomplishing those goals.

But we shouldn't avoid making useful progress on the grounds that we can't perfectly prove that we won't regret the work we do today.

kellyselden:
When RFCing a new feature, we ask to not mentioning any new deprecations. Those are to be debated on their own merit in a separate RFC. I think we should treat Edition RFCs the same way. Discuss the new feature by itself, and save any edition addition in a separate RFC. The reason I say this is because I fear we are anxious about "putting a bow" on the Octane edition. There will always be one more thing we can put in it. I would like to avoid any Ember 1.13 mistakes and allow this to land, but not stress about getting it in the existing edition.

simonihmig:
> I understand the for a lot of Ember developers, pods "already exist", but I want to emphatically state that the details of pods are not the future of Ember, that there are many discrepancies between how pods work today and the form that any future direction for the framework will take

Not sure what the future of Ember means in that specific context, but I would *assume* it's related to a more deterministic/compiler/build time centric way of doing things? If so, my question stands why this is coupled to introducing a new file layout? 

*Note: I am very well aware that I most certainly don't have enough context and knowledge to see the whole picture here, but if so it would be nice to get me and other non-core folks on the same page here!*

> and that ecosystem tooling and library support for pods is spotty and inconsistent today.

But how would introducing yet another file layout, that will predictably be superceded by the real thing (MU), help in that regard, rather than do additional "harm"?

> This also doesn't mean that we should make wild changes with nary a thought about what might come next. To the extent that we can reasonably predict our next technical steps, we should try to align our current work with those steps.

All of this makes sense. But as we already have a concrete idea about MU, I still wonder [as said before](https://github.com/emberjs/rfcs/pull/481#issuecomment-486383770) why the proposed form of colocation (which is good by itself) is actually *not* aligned with MU (and existing pods), as both the flat and the nested forms are quite different, while aligning it perfectly with both pods and MU could be easily achieved AFAICT?

tschoartschi:
I also really like the idea of Co-location. Beside all the arguments mentioned I also like Co-location for Merge Requests (Pull Requests in Github jargon). It makes reasoning about what's going on in PRs much easier.

Nevertheless I share the concerns of the others. I'm also not sure if a new file system layout does more good than harm in the current state. Especially because this layout also seems to be an intermediate step. In my point of view, it would make sense to default to pods or to stick with the current file layout and maybe make the build tool smarter (but I have to less knowledge if this is feasible). 

Maybe we could save resources of the core team if we are not introducing a new file system layout and help them focus on MU instead. When MU lands we could write some code-mods or migrate our apps by hand but we would have to do it only once. 

Maybe someone with more insight could elaborate on why pods are not an option. I don't see too many differences in having a folder with a `compent.js` and `template.hbs` to `index.js` and `index.hbs`. 

If there are massive benefits with `index` over `component/template` I could also live with this. I wouldn't go without folders for the components. As others pointed out this could get messy very quickly especially if you want to store something like `style.css` along with a component.

wycats:
Since pods continue to come up, it's worth digging a little bit deeper into how pods work in Ember.

First of all, there is no sharp "pods" mode in Ember:

image

Instead, the Ember resolver looks for components and routes in their "pods" location first. Failing that, it looks in the "classic" location. The Ember CLI docs [on this topic](https://cli.emberjs.com/release/advanced-use/project-layouts/) are useful.

Pods layout in Ember means roughly two things:

1. storing component JavaScript files in `app/components` alongside their Handlebars templates (as `component.js` and `template.hbs`
2. storing the collection of routes-related files in `app/${routeName}`

> Aside: Pods was added to Ember before we firmed up the RFC process, and much of the implementation work was done in Ember CLI before Ember CLI participated in RFCs. It was confusingly billed as an experimental feature despite not being behind a flag or going through maturity stages. Nonetheless, the core team has historically considered Pods an experiment, and resisted recommending their use or changing the generators to default to creating pods. We always expected that pods would eventually be replaced with something similar but not identical.

This RFC is an attempt to clean up and standardize (1) in order to make it the default behavior. This RFC does not attempt to clean up and standardize (2), but it anticipates a future RFC that would do so.

From this perspective, it's natural to ask (as @tschoartschi did):

> Maybe someone with more insight could elaborate on why pods are not an option. I don't see too many differences in having a folder with a component.js and template.hbs to index.js and index.hbs.

In other words, why use `index.js` and `index.hbs` rather than sticking with the names used in the pods experiment?

The primary reason for this is to align with the precedent of the Node community, which allows free refactoring from `name.js` to `name/index.js`. In practice, this means not only that `require()` in Node supports this refactoring, but also that tooling based on Node precedent supports it as well. For example, [TypeScript's module resolution](https://www.typescriptlang.org/docs/handbook/module-resolution.html) looks for `index.ts` as part of its algorithm:

image

In practice, this means that when you say `import { some, stuff } from "./name"` in TypeScript, it finds `./name/index.ts`. In addition to the important effects on TypeScript tooling like Jump to Definition, it also means that VSCode's JavaScript support will jump to `./name/index.js` when you invoke "jump to definition".

When a `name.hbs` is included alongside `name.js`, it works similarly to the TypeScript behavior when a `name.d.ts` is included alongside `name.js`. It's providing built-time information, but the `.js` file is the primary location of the imported module at runtime. Because `name.js` and `name.d.ts` works, and because of how Node module resolution works, it follows that `name/index.js` and `name/index.d.ts` would work the same way.

In fact, it would be very surprising if `name/index.d.ts` only worked to provide additional typing information if it was literally called `index.d.ts` and it would similarly be very unidiomatic if the only way to add typings to a module was `name/module.js` and `name/typings.d.ts`. There are other programming languages that have idioms along these lines, but JavaScript is not one of them.

This RFC, therefore, diverges a bit semantically from pods. In the pods system, if you wanted to import `component.js`, you would have to say `import './name/component'`, even though referring to the same component in a template does not include the filename `component`. This RFC aligns the name used to refer to a component in a template with its JavaScript module name, according to JavaScript idioms.

With this context in mind, it would be very strange for us to support `component-name/index.{js,hbs}` but not `component-name.{js,hbs}`. The `.hbs` file has the same name as the `.js` file, and is located next to it, because it's providing build-time data for that module.

> Interestingly, if we chose to be more restrictive than this, it would mean that `component-name.{js,d.ts}` would work, but you would be forced to use `component-name/index.{js,d.ts,hbs}` when adding a Handlebars file. This might be a good idea for ergonomic reasons, but going with the grain on .d.ts precedent feels right to me.

wycats:
I should also mention that `name.js` and `name.hbs` share the same name because they are semantically coupling the template's contents with the JavaScript file's contents at build-time. References to `this` inside the template will always refer to an instance of the component class in the associated `.js` file.

This RFC does not imply that you should write `index.md` instead of `README.md`, any more than you would call the README in the root of a node package `index.md`.

wycats:
Finally, I think it's important to mention that this RFC is not intended to be a stopgap that will be superseded by yet another tweaked file system layout some time in near future. Instead, it is an attempt to specify the long-term way in which component JavaScript files and their associated templates will be laid out next to each other on disk.

This RFC is saying that when we eventually tackle route layout, to finish the job of standardizing the features in today's pods layout, they should be `name.{js,hbs}`.

None of that precludes the possibility of a future single-file layout, but it would have to be sugar for this layout, and the introduction of a future single-file layout would not come along with a deprecation of the separate-file layout.

lifeart:
can we auto generate index.js/ts/d.ts and reexport component.js/ts inside it?

index.js
index.d.ts
component.js
template.js
style.css

(Component, Template, Style - optional files)

chancancode:
I've removed the Octane timeframe from the RFC to allow the feature to be discussed on its own merits. If accepted, I still think it would probably land within the Octane timeframe anyway, but it doesn't have to be a blocker for the edition.

tschoartschi:
@wycats thank you for the detailed clarification. I think it makes sense to align  with the wider JavaScript eco-system. For me the only question is what's a better default for the blueprint. I would opt for folder structure but maybe I'm biased because we use pods for components (we do not use it for routes)

GavinJoyce:
I want to add some detail to why I'm very keen on this RFC landing and to address some feedback from people who are already using pods and don't necessarily see the value in this.

As much as possible, we try to stay on the Ember happy path in Intercom. It's valuable to be able to avoid bikeshedding and short-cut technical discussions when there are clear conventions in place. The happy path helps onboard new engineers onto our product teams, the vast majority of people who work on our Ember app were new to Ember when they joined Intercom. We've been very successful sticking to Ember's happy path and we'll continue to follow it where possible.

While the improved developer ergonomics of the pods layout were always attractive, pods was never the Ember happy path. As long as I remember pods being available, I remember clear guidance from the core team that pods do not represent the future. The guides have always reflected this too. For the last few years at least, a version of module unification was being actively worked on and constantly seemed just around the corner. We have been patiently waiting, and contributing where we can, for a better happy path file layout for years.

We obviously now have the benefit of hindsight, big projects often are a [costly accumulation of hard work and assumptions](https://www.intercom.com/blog/shipping-is-your-companys-heartbeat/) and often fail without ever shipping. MU was one of these projects and we should learn from it: we need to improve Ember's file layout happy path incrementally.

This RFC is an incremental improvement to Ember's happy path file layout, it will deliver an immediate benefit to existing apps that don't use pods and to all new apps. It will likely be the first of many steps which will bring us to a comprehensive MU like file layout. By shipping this incremental change we will learn from real-world usage and future RFCs can leverage these learnings. Although these changes might not be directly applicable to apps using pods, they will ultimately benefit all apps as with this we move a step closer to having a single file layout for all Ember apps.

If you use pods, you can continue using them until a comprehensive MU like solution has landed. I strongly believe that that solution should be designed and shipped incrementally and that this RFC is a strong first step.

simonihmig:
Thanks @wycats for taking the time to go into details here and address the concerns. I think that was extremely helpful!

> When a name.hbs is included alongside name.js, it works similarly to the TypeScript behavior when a name.d.ts is included alongside name.js. It's providing built-time information, but the .js file is the primary location of the imported module at runtime.

I feel the TS analogy is a bit misleading, as the typings file is just meta data, and cannot reasonably exist without the js file. While a template can exist without a component.js file and vice versa. So they are related, but different things. 

But maybe that's really part of the (mental) problem I and others have, thinking of a component and its template as different things (`types` in MU nomenclature). Maybe the better mindset would be to think of them more as two sides of the same coin (even when one can optionally be omitted), which only when combined (compiled) together form that *one* thing called a component. This mindset would clearly demand not only to have those file co-located but also to give them the same name! (as they are just one thing, or MU `type`).

> This RFC, therefore, diverges a bit semantically from pods. In the pods system, if you wanted to import component.js, you would have to say import './name/component', even though referring to the same component in a template does not include the filename component. This RFC aligns the name used to refer to a component in a template with its JavaScript module name, according to JavaScript idioms.

Very good point! Which made me reconsider my initial concerns, and which seemed to be missing in the RFC's motivation, or at least I didn't really get it before. I can see now how this can help pave the way for template imports...

> Finally, I think it's important to mention that this RFC is not intended to be a stopgap that will be superseded by yet another tweaked file system layout some time in near future. Instead, it is an attempt to specify the long-term way in which component JavaScript files and their associated templates will be laid out next to each other on disk.

Ok, that seems to answer my other concern. So any future MU-successor RFC will take the precedent set by this RFC into account, and *presumably* not try to reinvent component file layout again (i.e. revert to `component.js`/ `template.hbs`), right?

text/0481-component-templates-co-location.md
```diff
@@ -0,0 +1,599 @@
...
+to have dynamic content by taking arguments using the `@name` syntax, and
+further "upgraded" to keep internal states by creating a JavaScript file next
+to the template and finally add interactivity with element modifiers.
+
+Because template-only component is such a light-weight concept, it is arguably
+not necessary to separate out the topic of "templates" from "components" in the
+guides, but this can be addressed in a future revision of the guides.
+
+We will teach that the `templates` folder is used for route templates. This can
+be introduced at the same time as the `routes` and `controllers` folder, which
```
jenweber:
```suggestion
be introduced in the Guides at the same time as the `routes` and `controllers` topics, which
```

text/0481-component-templates-co-location.md
```diff
@@ -0,0 +1,599 @@
...
+would come at a later time than components.
```
jenweber:
```suggestion
come later in the Table of Contents.
```

text/0481-component-templates-co-location.md
```diff
@@ -0,0 +1,599 @@
...
+   unlikely to conflict with options on other `ember` commands, which is why
+   they are a bit verbose.
+
+   For Octane apps, the default app blueprint will include a `.ember-cli` file
+   that defaults to `--no-component-class` and `--component-structure=flat`.
+   The guides and documentation will assume these settings going forward.
+
+## How we teach this
+
+As mentioned above, the learning team will update the guides and API docs to
```
jenweber: Suggested change to this first paragraph:

> As mentioned above, we will update the learning resources to assume the "flat" co-located layout. Throughout the Ember Guides, the Tutorials, and the CLI Addon Tutorial, we would update all Component Handlebars code sample filenames. The prose describing the location of files would also need to change. The API documentation will describe the full set of file layout options supported in the major version of Ember. A section should also be added to the CLI guides, which describes how to properly import components from addons that have a mix of layout types. We will not cover blending file layouts in the Ember Guides, since they represent the happy path for an app, but we could link to the CLI Guides explanation.

jessica-jordan:
This reads great, thank you for writing up this RFC :sparkles: One thing I'm wondering about from a learning perspective is if there should be a commonly used term for the old location style of components. We could use that term to refer back to that pattern in e.g. the Guides or the Super Rentals tutorial where needed. I think it would also be helpful to have that common phrasing for not only the new location style (`co-located components`), but also for the old location style, to be used in learning resources like blog posts, talks or discussions by the community.

What are everyone's thoughts on this and if it makes sense to introduce a term for it, what could be a good name?

rwjblue:
Thank you @GavinJoyce, https://github.com/emberjs/rfcs/pull/481#issuecomment-487266284 very closely matches my perspective as well.

I wanted to take what @wycats said in https://github.com/emberjs/rfcs/pull/481#issuecomment-487237048 a bit further:

> Finally, I think it's important to mention that this RFC is not intended to be a stopgap that will be superseded by yet another tweaked file system layout some time in near future. Instead, it is an attempt to specify the long-term way in which component JavaScript files and their associated templates will be laid out next to each other on disk.

It is my opinion (shared by many if not all core team members) that the original redesign for the filesystem in Ember apps (aka Module Unification / #143) is **very very** unlikely to _actually_ land at this point. There are too many new factors that the original system doesn't account for (e.g. with template imports, why do we need local lookup?), @tomdale sums things up pretty well in [this blog post](https://blog.emberjs.com/2019/03/11/update-on-module-unification-and-octane.html). While I absolutely stand by the design work we did in #143 (and still quite like the final file layout that we landed on) we simply can't land it. 

Instead of following through with the original MU design, we should instead shift our focus to making small iterative improvements to the file system layout. IMHO, these smallish bitesized improvements are the only way to _actually_ ship a better layout. I see this RFC as the first in a series of RFCs, each to be digested on its own, but the culmination of which is a layout very much like the one proposed in #143.

_tldr; Instead of attempting to implement all of the features of MU across the entire ecosystem at once, this RFC focuses on one aspect: component and template colocation. **Let's get this done folks!**_

Alonski:
@jessica-jordan What about classic for old and Co located for this?

Panman8201:
> @jessica-jordan What about classic for old and Co located for this?

While I think "classic" is a good name for the original, "co-located" might be a bit too generic if there will be other incremental updates that do a similar thing (eg: controllers w/ their associated template). Instead, maybe be a bit more specific like "co-located components". Or perhaps those working on the file layout changes have a better suggestion based on other ideas for the incremental changes.

rtablada:
To start I'm very 👍 on moving to `app/components` with node style naming/resolution and even with shipping this as the blueprint default for Octane or soon after.

However, I don't share that thought on the technical implementations and other concerns raised (at best they deserve or need a separate RFC).

The other issues raised in the RFC are:

1. extends bringing along templates from parent components (where applicable)
2. Avoiding late template binding for the 80% use case for performance
3. Adding public APIs for build time to associate templates with components (mostly to solve the above questions and give an implementation for how to support the new file layout without too many resolver or component manager changes)

These are both good goals and would fulfill a story for components but I think they are separate issues than colocation (though they do become more painful for sure).
My worry around these other two issues is that they seem to be in competition, conflict, or confusion with other efforts and RFCs (specifically #454).

---

To solve the main issue that this RFC puts forward (colocation of templates to match node naming conventions) this could be achieved with a broccoli build time move step that moves modules from the node path to the traditional path (with some consideration as well around import statements).

Is this ideal? No, but it doesn't muddy co-location and folder structure with other concerns.

In my mind this is a polyfill of sorts as the public primitives proposed like Embroider and/or #454 gain further consensus and implementation.
In that form as the primitives land, then the "mover" build step then gets replaced with a build step that uses APIs such as `setComponentTemplate`.

oligriffiths:
This is excellent and I fully support colocation of component and template files. 

A lot of the comments cover my overall concerns but I wanted to share some thoughts.

1. If the intention of this RFC is to replace pods and iteratively improve upon file system layout in order to get towards MU (or its descendent architecture) then 👍. Can this be added in part of the preamble as to these intentions.

2. Scope: This RFC seems to be covering multiple scopes:
  a. file system and developer ergonomics
  b. addon developer API
  c. runtime/compiler optimizations.
  d. proposing an official core-team approved co-location strategy (bye bye pods)

I would like the RFC to be very clear about exactly what the intentions are, and if at all possible, separate the scopes into separate complimentary RFCs if at all possible

3. Pods: given that 2.a. is already technically possible with pods, explanation about why pods can't/shouldn't be used/improved to achieve the desired goal should be included. This should also be addressed in how we teach this, as there are valid concerns about a 3rd way of file system layout.

Over all this is awesome, and I look forward to seeing it progress. Thanks.

PrinceCornNM:
We have been doing this with pods for years. It's a must for our team. It saves time and would benefit the whole community.

Panman8201:
> 3. Pods: given that 2.a. is already technically possible with pods, explanation about why pods can't/shouldn't be used/improved to achieve the desired goal should be included. This should also be addressed in how we teach this, as there are valid concerns about a 3rd way of file system layout.

"can't/shouldn't" is a bit strong since pods are used by a lot of projects (as proven by all the comments here). Instead of telling everyone to "switch back to classic layout", let the incremental improvements shake-out, at which point I think there will be a "natural" migration path and draw for projects to change. IMO, this RFC should have no impact on pods layout. Let pods be what it is, and continue to push classic layout forward.

MelSumner:
> Another drawback is that it only address the co-location issue for components, not other related types like [[route, controller, route template]] and [[model, adapter, serializer]], or even co-location of tests. However, we believe the situation with components is unique enough (see the motivation section) that they are not merely related, but coupled. That, along with the fact that components are much more common, sets them apart from the rest and justifies solving the problem first.

While this is listed in the drawbacks section, I don't see this as a drawback. I think this is a benefit. I strongly agree with the justification that the component .js & .hbs files are coupled, not merely related- and I would be opposed to a solution that wanted to put the related files there too (i.e., tests). 

I don't personally find it bothersome to navigate to the different folders (it never occurred to me that this was something I could be irritated about), it seems like a nice improvement and a salve for those who _do_ find this DX troublesome. 

For community members in this thread who have been mentioning or wondering about pods and/or module unification- please remember to read https://blog.emberjs.com/2019/03/11/update-on-module-unification-and-octane.html. While I think the position of the blog post solidifies that it's not part of Octane, it's also important for us all to understand that there ended up being significant issues with the approach overall, and it needed to be re-thought.

oligriffiths:
@Panman8201 I only say that about pods to draw reasoning in the “alternatives” section as to why pods can’t be used for this, or why they shouldn’t be updated to support this use case, as opposed to outlawing their use. However as @wycats says, pods have never been an officially supported feature and it seems their inclusion flew in under the radar. They certainly don’t feature in the Ember “happy path”, however I personally love the collocation feature of components in pods. It may be worth exploring allowing pods to be enabled more granularly for specific things like components. Just a thought.

text/0481-component-templates-co-location.md
chancancode: 👍 I applied the suggestions manually.

I continue to have doubts that it would be appropriate to cover these kind of things in the API docs, although I understand the motivation. Specifically, there is no natural place to cover abstract things like these in API docs – which must be attached to a concrete item, like a class or a function. Things like this (and angle/curly invocation styles, which was the last time it came up) are not a good fit for that because they apply to the abstract concept of components in general, not a specific class like `Ember.Component` (which is where they currently live as a compromise). I believe @wycats shares my concerns too.

However, we can stick with the current policy for the purpose of this RFC and discuss options outside of this thread.

chancancode:
>> When a name.hbs is included alongside name.js, it works similarly to the TypeScript behavior when a name.d.ts is included alongside name.js. It's providing built-time information, but the .js file is the primary location of the imported module at runtime.

> I feel the TS analogy is a bit misleading, as the typings file is just meta data, and cannot reasonably exist without the js file. While a template can exist without a component.js file and vice versa. So they are related, but different things.

I don't think this was intended to be a very important point, but you can definitely have a `d.ts` file without a corresponding js file.

Here is one example:

```ts
// enums.d.ts
export const enum Color {
  Red = "#ff0000",
  Green = "#00ff00",
  Blue = "#0000ff",
}

// index.ts
import { Color } from "./enums";

console.log(`Red: ${Color.Red}`);
console.log(`Green: ${Color.Green}`);
console.log(`Blue: ${Color.Blue}`);

// output
console.log("Red: " + "#ff0000" /* Red */);
console.log("Green: " + "#00ff00" /* Green */);
console.log("Blue: " + "#0000ff" /* Blue */);
```

> But maybe that's really part of the (mental) problem I and others have, thinking of a component and its template as different things (types in MU nomenclature). Maybe the better mindset would be to think of them more as two sides of the same coin (even when one can optionally be omitted), which only when combined (compiled) together form that one thing called a component. This mindset would clearly demand not only to have those file co-located but also to give them the same name! (as they are just one thing, or MU type).

That's exactly right.

chancancode:
> This reads great, thank you for writing up this RFC ✨ One thing I'm wondering about from a learning perspective is if there should be a commonly used term for the old location style of components. We could use that term to refer back to that pattern in e.g. the Guides or the Super Rentals tutorial where needed. I think it would also be helpful to have that common phrasing for not only the new location style (co-located components), but also for the old location style, to be used in learning resources like blog posts, talks or discussions by the community.
> 
> What are everyone's thoughts on this and if it makes sense to introduce a term for it, what could be a good name?

I think classic is fine for what we have now (matches the generator's `--component-structure=classic`), but I'm not sure if the "new" one needs a name more than we need a name for what we have today before this RFC.

That is to say, when the guides are updated, what is proposed here would just implicitly be the "current" file layout.

chancancode:
@rtablada @oligriffiths This scope/intention of this RFC is to presents a solution that solves *all four problems listed in the motivation section*. I don't think it would make sense to split it up. Co-location is a solution, not a problem – a solution to the four problems listed in the motivation section. Of course, we could split up the four problems into different RFCs, but it would then make sense to merge them since they have the same proposed solution. Do you have any concrete objections to the proposed solutions? If so, let's discuss the concrete objections.

This RFC does not intend to conflict, replace or otherwise interfere with other RFCs like #454 (or ["strict mode"](https://github.com/emberjs/rfcs/tree/strict-mode), etc). They will all be built on top of the `setComponentTemplate` primitive here. If there are any specific concerns let's discuss the concrete issues.

chriskrycho:
(Minor aside as a point of clarification for any ember-cli-typescript v2 users: the example @chancancode shared about `const enum` types is perfectly valid TypeScript, but doesn't work in ember-cli-typescript b/c we use Babel for transpiling TS and it doesn't support `const enum`. The point stands in general: you *can* have type-only exports in `.d.ts` files and those are fine. The example provided works in regular TS because a `const enum` *is* a type-only export; a non-`const` `enum` would not work because it actually generates runtime code.)

MelSumner:
> We have been doing this with pods for years. It's a must for our team. It saves time and would benefit the whole community.

Can you clarify- are you saying that you agree with this RFC, because the component js/hbs co-location is similar to pods? Or something else?

rwjblue:
We discussed this at the Ember core team meeting this past Friday, and are moving this into the final comment period.

davewasmer:
> Instead of following through with the original MU design, we should instead shift our focus to making small iterative improvements to the file system layout. IMHO, these smallish bitesized improvements are the only way to actually ship a better layout - @rwjblue 

This is the core of my concern with this RFC. I'm strongly in favor of co-location overall, and have no opinion on the implementation details described here. But I'm worried that this approach to incrementally tweaking the filesystem layout will lead to confusion, and ultimately, end up thwarting a core benefit of Ember:

> As much as possible, we try to stay on the Ember happy path in Intercom. It's valuable to be able to avoid bikeshedding and short-cut technical discussions when there are clear conventions in place. - @GavinJoyce 

---

Let's assume this RFC lands, and eventually a couple more tweaks to the filesystem layout land as well (let's say route folders and ... something with helpers - I'm just making this up as an example). When a new dev joins an Ember team and starts to work on their app, how will they know where to find things on disk? 

So far, the answer has been: _"pods or classic"_.

With this incremental approach, it seems like the answer will be: _"well, we use Ember's classic structure, except we use the co-located component templates and the new route folders, but not the new helper layout"_.

Each tweak would seem to introduce a new mental model that must be layered on top of all the previous mental models to produce the filesystem structure for a given app. For something so basic and core to the experience of developing an app as "where does the code I care about live?", that's a lot of cognitive load.

> Finally, I think it's important to mention that this RFC is not intended to be a stopgap that will be superseded by yet another tweaked file system layout some time in near future. ... This RFC is saying that when we eventually tackle route layout, to finish the job of standardizing the features in today's pods layout, they should be name.{js,hbs}. - @wycats 

My worry is that this is not the way Ember users will experience this RFC. Yes, on some level, the RFC is aimed at uniting the component's JS and template (a worthy goal). But in practical terms, the way it changes dev experience is a new filesystem layout model (now a 3rd, after classic and pods).

When, say, a route layout RFC lands as well - that's another (4th!) layout model. I don't think we can treat this RFC as a different level of abstraction that just combines nicely with some future route layout RFC - the practical effect is that a route layout RFC adds another layer of rules to be mentally evaluated with all other previous sets of rules.

The end result of this RFC plus a routing layout RFC would be 4 different ways to structure an Ember app. To make it even worse, there will likely be some mixture of them allowed to coexist in the same app at the same time. That seems like a significant mental burden.

---

I'm curious why we can't expose lower level primitives that allow this to be experimented with in addon-land, like the template imports RFC, as I suggested above. I'll admit I haven't fully grokked the underlying implementation details outlined here so far - is that fundamentally not possible, or simply not a goal?

wycats:
> Let's assume this RFC lands, and eventually a couple more tweaks to the filesystem layout land as well (let's say route folders and ... something with helpers - I'm just making this up as an example). When a new dev joins an Ember team and starts to work on their app, how will they know where to find things on disk?

> So far, the answer has been: "pods or classic".

I think this is a little bit misleading. At minimum, the current answer is [pods, classic or mixed](https://cli.emberjs.com/release/advanced-use/project-layouts/).

image

As much as people are convinced that they love Pods and we should just have adopted them wholesale, Pods were always an experimental approach with a number of edge-cases. It's hard to catalog the edge-cases precisely because Pods was never incorporated into Ember proper.

We aren't going to eliminate Pods over night because, like or not, Pods were included as an option in the Ember blueprints, and at various points were recommended strongly by important members of the Ember ecosystem. But they were never a consensus position, had known edge-cases that many people don't understand, and interact poorly with any number of addons.

The goal of this RFC is to migrate us, over the next year or so (my guess) to a single answer to the question. A single answer is better than two answers, but it does require us to have three answers in the meantime. We're hopeful that strong codemods will reduce the impact of the change, and don't intend to force people to schedule their teams to migrate to co-location any time soon (certainly not before 4.0).

What's missing in this analysis (both the original comment and my response so far) is the question a developer asks about a new app: "which file layout should I use". Today, the answer is: _"the default, but maybe pods"_ (depending on who you're talking to). The goal of this RFC is to significantly increase the number of people for whom the right advice will be: "the default, duh". We believe that since co-location addresses a significant enough part of the reason people advocate for pods, and has the benefit of being built-in, that more people will advocate for the default layout.

This is speculation, of course, but so is everything else we've been talking about. I don't know for sure that this speculation is right, but getting to the point where Ember has one default file layout that is supported by the vast majority of the community is an extremely important goal that I think needs to take precedence here.

davewasmer:
> getting to the point where Ember has one default file layout that is supported by the vast majority of the community is an extremely important goal that I think needs to take precedence here.

I agree wholeheartedly 💯

> I think this is a little bit misleading. At minimum, the current answer is pods, classic or mixed.

Fair point. Allow me to rephrase my objection below - I think @wycats response here has helped me clarify my own thinking on this.

> As much as people are convinced that they love Pods

I should perhaps clarify - I'm not suggesting we adopt pods wholesale as the solution. I'm personally in favor of some kind of co-location, but fully acknowledge the shortcomings of pods. I'd love a fully support co-located filesystem layout eventually.

> A single answer is better than two answers, but it does require us to have three answers in the meantime.

But it seems like (and perhaps I'm misunderstanding here?) that's not the end of the story. We'll have 3 answers once this RFC lands. Then 4 if/when a routing layout RFC lands. Possibly more if we do more tweaking beyond that.

In the long run, sure, we'll settle on one solution, but I worry that the multiplicity of solutions in the meantime is unnecessarily painful.

I've asked about an alternative approach a couple times so far, but didn't see anyone address it yet. I wonder if we could try to expose lower level primitives that allow addon-land code to experiment. Once the community begins to find consensus on a holistic solution, we RFC that for core adoption.

I'm sure such an approach has already been considered, so I'd love to understand what its downsides might be.

GavinJoyce:
> We'll have 3 answers once this RFC lands. Then 4 if/when a routing layout RFC lands. Possibly more if we do more tweaking beyond that. (@davewasmer)

> In the long run, sure, we'll settle on one solution, but I worry that the multiplicity of solutions in the meantime is unnecessarily painful. (@davewasmer)

I don't view these incremental steps as necessarily introducing new solutions nor that they will be particularly painful. The rules for where files live in any particular app at any one time isn't that complicated and I don't believe that the rules changing, either incrementally or all at once, will be any more of a burden than the usual set of other incremental changes that upgrading an Ember app brings.

As I see it, there are a number of scenarios that apps will be in during this incremental and transitional phase:

### Apps using pods:

**During the transitional phase:** Do nothing, continue to use pods.
**Once complete:** Use provided codemods to migrate away from pods to new happy path.

### Apps using the current file layout and wishing to migrate incrementally:

**During the transitional phase:** Migrate step by step using provided codemods.
**Once complete:** Nothing to do, already using new happy path.

### Apps using the current file layout and wishing to wait until the new happy path lands:

**During the transitional phase:** Do nothing, continue to use the current file layout
**Once complete:** Use provided codemods to migrate away to new happy path

GavinJoyce:
In additional to [my reasons why we're excited about this RFC in Intercom](https://github.com/emberjs/rfcs/pull/481#issuecomment-487266284), I'm also keen on landing this RFC as I believe that it will significantly improve our chances of winning the next set of ember developers.

I'm currently working on what will hopefully be a high quality and highly polished comprehensive Ember tutorial for the [freeCodeCamp youtube channel](https://www.youtube.com/channel/UC8butISFwT-Wl7EV0hUK0BQ) (~1 Million subscribers !!). 

Octane brings an opportunity like we've never had before to win the hearts and minds of the next generation of people building web apps. One of the first things I usually do in tutorials like this is to create some components. Glimmer components in Octane are an early opportunity to show why modern Ember is exciting, elegant and in many ways better than other, more popular alternatives. Having to explain why the component javascript and template are located in two different places puts Ember at a significant disadvantage to other competing frameworks like React though. For this reason, I've always have this little internal battle over whether I should use pods or not when planning a new tutorial. I've never used pods in tutorials as they themselves require some explanation at the beginning, I know that EmberMap sometimes do use pods for tutorials though.

Octane is a significant milestone for Ember, we'll have a likely moment of significant interest from the wider Javascript community. If this RFC lands in a similar time frame as Octane, we'll be able to create much more polished and concise tutorials that will maximise that chances that new people will consider investing more time in exploring, learning and adopting Ember.

wycats:
> I've asked about an alternative approach a couple times so far, but didn't see anyone address it yet. I wonder if we could try to expose lower level primitives that allow addon-land code to experiment. Once the community begins to find consensus on a holistic solution, we RFC that for core adoption.

This RFC actually *does* create a primitive for addons to use, but I don't see how that helps with the problem of creating ecosystem fragmentation. Since we expect co-location to be extremely popular, having no opinion yet about the future will almost certainly result in:

- pods continuing to be perceived as a semi-official solution
- even more churn in the ecosystem as people work out which co-location approach to use

Since the thing people are worried about is that they worry that this change will fragment what an Ember project actually is when working on a new app, I don't think that 🙈 🙊 🙉  and punting to the ecosystem will help with churn or reduce confusion.  To the contrary, I think it will increase the amount of practical fragmentation, reduce our ability to communicate clearly to new users, and make it harder to migrate everyone once the experimentation phase dies down.

In many cases, those costs are worth the benefits of exploring a space and finding a good solution. In this case, we have been running a long-term experiment with pods that already has significant costs, and the benefits of further experimentation don't appear to buy us enough to justify taking more time to give Ember users a clear, default, colocation solution.

wycats:
The core team discussed this RFC and believe that enough time has passed without new arguments to bring this FCP to a close.

MichalBryxi:
Just got bitten by an assumption that folder:
`my-ember-addon/addon/templates` is not needed after this migration. The folder is literally empty. I deleted it and `ember server` started shouting:

```
(node:4009) UnhandledPromiseRejectionWarning: Error: ENOENT: no such file or directory
    at Object.statSync (fs.js:855:3)
    at SourceNodeWrapper.build (my-ember-addonnode_modules/broccoli/dist/wrappers/source-node.js:10:21)
    at Promise.resolve.then.then.then (my-ember-addonnode_modules/broccoli/dist/builder.js:99:36)
    at process._tickCallback (internal/process/next_tick.js:68:7)
(node:4009) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 10)
```

The fix is quite simple:
```
mkdir my-ember-addon/addon/templates
touch my-ember-addon/addon/templates/.gitkeep # to make sure fresh `git clone` will work
```

Is there a reason for this to happen? Or could we maybe get rid of that folder completely without the failures?

rwjblue:
@MichalBryxi - No, that is not expected (the `addon/templates` folder should not be required). Can you please open an issue on ember-cli/ember-cli with steps to reproduce?

MichalBryxi:
Ok, sorry. Tried to reproduce this again. But now there is no problem with `my-ember-addon/addon/templates` folder missing. I have the exact commit where the problem occurred, but can't reproduce now 😞. So my best guess would be that it was some `node_modules`/tmp cache? Or maybe node/yarn version mis-match (due to volta adoption)? Sorry for the noise.

Just for historical documentation:
 - I'm using monorepo with yarn workspaces
 - The problematic folder was in an addon
 - The problem occurred when running `ember s` in app in that monorepo that consumes given addon
 - I went through volta adoption in the process - pinned (different) yarn and node versions

buschtoens:
@MichalBryxi Was `ember s` already running when you deleted the `addon/templates` directory and then threw the error?

MichalBryxi:
@buschtoens That was also my thinking at the point. Pretty sure I restarted the ember server when I saw this. I also tried to play around with this scenario now, no luck.
The app is working for me now without that directory, so 👍  🤷‍♂

Issue #473, opened 2019-04-05
RFRFC remove `hash` from the Ember lexicon


chancancode:
This is specifically talking about calling "map/dictionary" objects as "hash", not about "hash location" in the router.

Problems:

* `hash` is probably a Ruby-ism that we inherited, which makes little sense in our JS context
  * aside: `hash` in ruby refers to the fact that it is implemented as a hash table, however, I believe the more common name for the abstract data type (i.e. decoupled from the implementation) is a map, dictionary or associative array..?
* In JavaScript, it is probably more common to refer them to just as "objects" (the "o" in "POJOs") 
  * a `Map` is also a thing in modern JavaScript, but it's not what the `hash` helper produces (maybe it should? probably not? unknown.)
* A straw-man proposal is to rename the `hash` helper to `obj` or `object`, and (maybe later?) deprecate `hash`
  * We should provide a codemod
* What should we call the second argument passed to helpers? The current signature is commonly referred to as "params and hash", and that's the name/identifier we generate for the function parameters in the blueprints. "options"? What about "params", do we want to rename that too?

rwjblue:
> What should we call the second argument passed to helpers?

IMHO, we should call them `positional` and `named`...

dgeb:
I would favor transitioning from `hash` to `obj` if the `fn` helper lands in #470 (since they seem analogous).

bertdeblock:
Does renaming `params` and `hash` in the helper blueprint need to be part of an RFC or is that something we can just do as is? In the [Helper Functions](https://guides.emberjs.com/release/components/helper-functions/#toc_writing-a-helper-function) guides we also use the term `args` instead of `params`.

rwjblue:
I think that we can change the blueprint argument names independently of a discussion around the "hash" helper.

wagenet:
I'm closing this due to inactivity. This doesn't mean that the idea presented here is invalid, but that, unfortunately, nobody has taken the effort to spearhead it and bring it to completion. Please feel free to advocate for it if you believe that this is still worth pursuing. Thanks!

PR #482, opened 2019-04-23
Deprecate hash helper

GreatWizard:
[Rendered](https://github.com/GreatWizard/rfcs/blob/hash-helper/text/0482-deprecate-hash.md)

### DIFF
--- a/text/0482-deprecate-hash.md
+++ b/text/0482-deprecate-hash.md
@@ -1,0 +1,77 @@
+- Start Date: 2019-04-23
+- Relevant Team(s): Ember.js
+- RFC PR: https://github.com/emberjs/rfcs/pull/482
+- Tracking: (leave this empty)
+
+# Deprecate `hash` helper
+
+## Summary
+
+This RFC intends to rename the `hash` helper into `obj`.
+
+## Motivation
+
+The name of the helper can be confusing depending of the background of the developer.
+
+JavaScript developers are not used to the hash terminology because they already have literal object notation like `let options = { url: 'http://www.example.com', method: 'GET' }`.
+
+In addition, hash terminology can means different things depending on the context.
+In another context, it can means cryptographic hash or a list of objects.
+
+## Transition Path
+
+The new `obj` helper should be implemented in Ember.js with the same behavior as `hash` helper which means it should returns an immutable object without prototype.
+And this helper can't be override by a custom helper in your application.
+
+Polyfill will helps to use more recent code without upgrading your version of Ember.
+The codemod will helps to migrate deprecated code into stable code.
+Since it's a strict renaming, there is no need to deprecate `hash` right away.
+
+1. Implement `obj` helper in Ember.js
+1. Polyfill `obj`
+1. Write the codemod
+1. Deprecate `hash`
+
+## How We Teach This
+
+The new `obj` helper maintains the same behavior as `hash`, so it is not necessary to explain it.
+
+The term "object" and its short-name "obj" should be familiar to JavaScript developers since they are used frequently in code and when talking about code, for example when mentioning POJOs (Plain Old JavaScript Objects).
+
+To use the new helper, you should change your code from:
+
+```hbs
+{{#let (hash name='Guillaume' job='Ember developer') as |person|}}
+  <p>{{person.name}}</p>
+{{/let}}
+```
+
+to:
+
+```hbs
+{{#let (obj name='Guillaume' job='Ember developer') as |person|}}
+  <p>{{person.name}}</p>
+{{/let}}
+```
+
+Alternatively, you can also run the codemod that will be provided.
+This codemod should also be automatically run by `ember-cli-update` for the relevant versions.
+
+For addon developers, you must use the polyfill as dependency to be compatible with all versions of Ember.
+
+## Drawbacks
+
+### Reserved name
+
+If this RFC is implemented and you already declared a `obj` helper, it will throw an exception that you can't override the built-in helper.
+
+## Alternatives
+
+We could do nothing and leave things as is.
+
+## Unresolved questions
+
+### Obj or Object?
+
+We could use the long name `object` for the helper.
+This one is longer but more understandable.


simonihmig:
I am not opposed to this, but given the vast amount of RFCs and changes and new concepts currently and over the past amount of months introduced as part of Octane and beyond, I am not sure this is the right time to introduce even more churn, in this case with relatively low value.

buschtoens:
Good point. Can we maybe add `{{obj}}` without deprecating `{{hash}}` so that these who are "in on it" can write future-proof code already? Alternative would be writing a polyfill / add-on.

chriskrycho:
Alternative perspective there: this is a relatively small (and trivially codemod-able) change, and landing it at the same time as we're updating/improving other helpers (`fn`, `on`, etc.) seems reasonable and even helpful to me.

bendemboski:
Speaking for myself, I find `hash` more clear than `obj`. I don't doubt that `obj` will immediately make people think `object`, but `object` is such a generic term that means so many things in so many different contexts. Is it an EmberObject? A Javascript object? Some other type of object in the Ember/glimmer ecosystem that I haven't heard of?

By contrast, I think the term `hash` (or some variant of `hash table`) is pretty widely used across many programming languages to mean a set of key-value pairs. So while it can also mean a cryptographic hash, a hash function, a piece of a URL, or a culinary dish consisting of diced or chopped meat, potatoes and spices that are mixed together and cooked, I think it does a much better job of communicating what it does in context (i.e. when immediately followed by a list of assignments that are easily understood to be key/value pairs) than `obj` does.

So my personal vote would be for `hash` over `obj`.

However, perhaps we could consider `pojo` instead? That should immediately communicate what it's doing to pretty much any Javascript developer, and also has the advantage of being google-able, which `obj` (or `object`) and `hash` do too, but it's much harder to get to a clear/likely definition since they both mean multiple things.

rwwagner90:
I personally do not think changing `hash` to `obj` is going to make things immediately clear to a developer who has never seen it before. With that in mind, we could assume the developer will go check the docs to see what the helper is doing, where they could read up on what `hash` is.

I would prefer to stick with `hash`, as it is familiar to Ember devs already, unless there is a strong case for changing it.

jonnii:
I think if this was changed (which I wouldn't be for) it would also make sense to change the `RSVP.hash` method to match.

chancancode:
See also: https://github.com/emberjs/rfcs/issues/473

@GreatWizard can you update the RFC to include renaming the helper parameters and the terminology in documentation

joukevandermaas:
I agree with @rwwagner90 that `obj` is not clearer on the surface, but I could easily be convinced otherwise by some people who have trained Ember developers before and have some data/anecdotes.

In general I think just changing the name of something, without changing semantics, is not worth the effort. A bad name (and it is debatable whether `hash` is a bad name) will take a few minutes to get used to, but updating our apps may take many hours. A codemod would help but that will also take time and resources away from other Ember work.

andreisebastianc-ccc:
> I agree with @rwwagner90 that `obj` is not clearer on the surface, but I could easily be convinced otherwise by some people who have trained Ember developers before and have some data/anecdotes.

I helped around 10 developers become productive with Ember and the `hash` helper was always straightforward. `yield` on the other hand, that's another story.

GavinJoyce:
I don't think `hash` is a particularly a bad name, I personally haven't seen evidence of it being a confusing name with people new to Ember in my company. 

I also don't believe that `obj` is a good name, if we were to change `hash` I would prefer that we use `object` instead which I think it cleaner and avoids an unnecessary contraction.

ondrejsevcik:
If this goes through, it should probably use longer name `object` rather than `obj` as there is already `array` helper and not `arr`.

jenweber:
I do like the semantics of `obj` and I had no idea what `hash` was when I was first learning Ember. 

However, to add to some of the comments above, I think the google-ability is very important for new learners. If there are no other changes besides naming, I believe we should keep it as `hash`.

locks:
@jonnii `RSVP` is a third-party library

pzuraq:
> Speaking for myself, I find hash more clear than obj. I don't doubt that obj will immediately make people think object, but object is such a generic term that means so many things in so many different contexts. Is it an EmberObject? A Javascript object? Some other type of object in the Ember/glimmer ecosystem that I haven't heard of?

@bendemboski I think part of the confusion here is the idea that there could be any other type of object than a JavaScript object. There _shouldn't_ be more than one, and it should just be that - a POJO 😄 

I personally think this change would help to bring clarity as we start to move the ecosystem more in this direction. Once we can get rid of `EmberObject` and drop all the custom trappings of the old class model, the helper would _just_ be a direct analog to the only type of objects that we have to deal with - JavaScript objects.

That said, I do think the google-ability of hash is much higher, and `hash` is a term used for key-value stores frequently in other languages, so it's not a terrible term. I just think it actually adds more confusion for folks in the long run, since it'll be another term to teach, and an Ember-ism in general.

pzuraq:
Another potential option here would be a true object literal syntax (alongside an array literal syntax). Curlies are out for that option, but some languages use double squares to represent objects:

```hbs
<button {{on 'click' (set this 'options' [[ key=value ]])}}></button>
```

bendemboski:
@pzuraq I'd say yes and no. Even if EmberObjects go away, I still kinda doubt most non-advanced Javascript developers know that

```javascript
class Foo {}
typeof new Foo() === typeof {}
```

Even though I know that in the abstract, I tend to think of the two as very different use-case wise, even though they both end up being objects. Perhaps this is because I became experienced in several other languages before Javascript, and in most other languages whatever dictionary/associative array/hash thingy they provide is different from the thing you get when you instantiate a class.

I'm not a fan of the object literal syntax either -- more stuff that isn't google-able, and that would also force the consistency question of how to express arrays, causing some scope creep here.

As an aside, but one I've love to hear a response to, I'm still not understanding why we think this is a problem we need to solve. Do we have evidence that it's actually confusing people to any significant degree? I haven't heard an argument that this will make things better outside of what I judge to be speculative arguments based on prescriptive nomenclature, and with no evidence behind them, I have trouble seeing how they could win out over the google-ability argument.

pzuraq:
I think the main thing is we're trying to remove Ember-isms and get as close to Just JavaScript™️ as we possibly can. `{{hash}}` sticks out because, well, there's nothing in JavaScript that is a hash. There are objects/POJOs, and there are `Map`s and `WeakMap`s and `Set`s and many other constructs, but vanilla JS doesn't have the concept of a hash, so we are definitely layering something on top here.

Personally I've actually typed out `obj` instead of `hash` before and been surprised, then remembered "oh right, we named it something completely different", more than once. I wouldn't be surprised if that doesn't come up more often than not, especially when `{{array}}` exists and _is_ a direct analog to a real thing in JavaScript.

bendemboski:
@pzuraq what do you think of pojo? Seems Javascript-y, accurate, and google-able...

pzuraq:
`{{pojo}}` feels a little bit more like JS, though it is industry slang so it could be hard for newcomers. It also has overlap/confusion with Java (Plain Old Java Object) which is where the term actually came from - I've gotten _very_ confused looks from Java devs when I start talking about POJOs in JavaScript. I think `{{hash}}` would be better than `{{pojo}}` tbh.
"Object" is a very overloaded term in general, but it _does_ have a specific meaning in each language it exists in. Every object-oriented language _must_ define what the term "object" means, usually quite explicitly, and it's the first thing new learners have to become familiar with. I'm personally not as worried about the use of the term because of this, even if it's overloaded in general.

bendemboski:
> It also has overlap/confusion with Java (Plain Old Java Object)

True, but also [this](https://www.google.com/search?q=pojo+javascript) and [this](https://www.google.com/search?q=pojo+ember) (and of course the second would have Ember documentation hits after the change was made). And I don't think confused looks from *Java* devs are what we're worried about 😄 I suspect most developers that switched from Java to Javascript have encountered the term POJO and are quite comfortable with it in a Javascript context.

Anyway, I think maybe I'm starting to argue just to be right now, so I'm going to start disengaging.

pzuraq:
Yeah, TBQH this is definitely major bikeshed territory (clearly 😛). One idea might be to include a couple different options for a while to actually see what works better here. When it comes to language/semantics design, I think at some point there are questions that just can't be answered directly, they need to be experimented with at large so that the users of the language can see what works most naturally (like how spoken languages evolve through slang and colloquialisms and such).

bendemboski:
Agreed. Also, I think this conversation has gone on way too long without referencing [this](https://twitter.com/codinghorror/status/506010907021828096?lang=en) 😺

0xadada:
👎 bikeshedding only introduces more churn. Is the work to update the code, docs and asking developers to update hundreds/thousands of apps worth it, i think no.

rtablada:
I'm personally not a big fan of renaming to obj.

While shorter and arguably cleaner I think it's a disservice for future to move to obj.
In terms of naming, JS objects are hashes (key value stores) but this is not the case in all languages.
With a future eye on WASM, Rust, etc the way we pass keys and references is much closer to `hash` in all languages vs `object` which can vary from language to language.

rtablada:
For a clarifiction. IMO it would be a lot better to try to get a language level hashmap syntax as @pzuraq mentioned rather than shifting helpers around to try to address a bigger design issue.

pzuraq:
As another data point here, I was introducing a friend who doesn't have a formal education and is new in general to programming to Ember. He currently uses a modified version of Dojo.js at his company, and he knows Vue and a bit of React. 

His first question when he saw the hash helper was "what's a hash?", and as I started explaining hash maps and dictionaries his eyes started glazing over. I said "it's an object" and he immediately got it, and we moved on. In my experience, this isn't uncommon for people who aren't familiar with programming terminology.

mixonic:
In recent framework core conversations this problem space has come up. For reference

* https://github.com/emberjs/core-notes/blob/master/ember.js/2019-09/Meeting_Ember_Core_Team_20190920.md has a quick mention of my todo for this comment (but sorry no notes, it is hard to talk about a topic and be taking notes :-p)
* https://github.com/emberjs/core-notes/blob/master/ember.js/2019-09/F2F_Ember_Core_Team_20190914.md#literal-objectarray-notation from the recent F2F discusses a proposal for object and array literals in Ember templates. The gist is something like `{{hello (a=b)}}` and `{{hello [a.b c "hello"]}}`. There are a few interesting edge cases around exactly when the `(` for an object literal is required. This work is not in an RFC yet, but we certainly have consensus on core that some kind of literal is appropriate.

Given that we would like to introduce an syntax for objects, it doesn't seem very high value to change the name of the `(hash` helper. However there wasn't any push to FCP this RFC to close since there is agreement that the `(hash` helper *should* be deprecated (but for a literal syntax) and in general it seems like bad form to close RFCs which are superseded by something still only a pre-RFC proposal.

Thanks for raising this and for the discussion on this thread so far!

GreatWizard:
Let's close it for now.

PR #485, opened 2019-04-26
Set up CI with Azure Pipelines

igorT:
CI will fail if the RFC still has the default number


kategengler:
Seems like we already have Travis setup? Should we be using both?

igorT:
@kategengler (c/p from discord) I'm using this as one of the test ground for experimenting with Azure pipelines and looking at moving to them longer term so this seemed like a nice addition to test the workflow with. If we end up not doing that, or decide to, it seems easy enough to unify the workflow later

rwjblue:
Going to close this for now (I'm still keen on doing the general change, but it seems like we should use GH Actions now instead of Azure Pipelines directly).

PR #486, opened 2019-04-27
Deprecate mouseEnter/Leave/Move Ember events

simonihmig:
[Rendered](https://github.com/simonihmig/rfcs/blob/deprecate-mouseenter/text/0486-deprecate-mouseenter.md)

### DIFF
--- a/text/0486-deprecate-mouseenter.md
+++ b/text/0486-deprecate-mouseenter.md
@@ -3,11 +3,11 @@
 - RFC PR: https://github.com/emberjs/rfcs/pull/486
 - Tracking: 
 
-# Deprecate support for mouseEnter/Leave Ember events
+# Deprecate support for mouseEnter/Leave/Move Ember events
 
 ## Summary
 
-Deprecate support for `mouseenter`/`mouseleave` events in Ember's EventDispatcher. This affects
+Deprecate support for `mouseenter`, `mouseleave` and `mousemove` events in Ember's EventDispatcher. This affects
 the corresponding event handler methods (`mouseEnter() {}`) in Ember components and 
 `{{action "some" on="mouseenter"}}`. 
 
@@ -40,6 +40,10 @@ the implementation has some performance drawbacks, as it has to process every `m
 *any* element, create fake `mouseenter`/`mouseleave` events and try to dispatch them, even when 
 not a single component/action needs them.  
 
+Deprecating support for `mousemove` is also proposed, which is a (bubbling) event that does not have the higher 
+implementation cost as `mouseenter`/`mouseleave`, but nevertheless requires the EventDispatcher to optimistically handle
+these extremely high-volume events.
+
 While efforts to make this more "pay as you go" are [possible](https://github.com/emberjs/ember.js/pull/17911), 
 the trade-off of keeping support around still seems unfavorable, as these events fire so
 frequently, while they are (most certainly) very rarely used.
@@ -50,7 +54,7 @@ work with event handler methods, and `{{action}}` will eventually fade away in f
 
 ## Transition Path
 
-When a `mouseenter`/`mouseleave` event is dispatched by the EventDispatcher to a component
+When a `mouseenter`/`mouseleave`/`mousemove` event is dispatched by the EventDispatcher to a component
 or `{{action}}` listening to that event type, a deprecation warning will be issued.
 
 The linked migration guide will cover these use cases:
@@ -110,7 +114,7 @@ After (based on [RFC471](https://github.com/emberjs/rfcs/blob/master/text/0471-o
 The deprecation guide should explain the transition path as shown above.
 
 The Guide's [Handling Events](https://guides.emberjs.com/release/components/handling-events/#toc_event-names)
-section must remove the reference to `mouseenter`/`mouseleave`.
+section must remove the reference to `mouseenter`, `mouseleave` and `mousemove.
 
 Other than that, no changes are required, as the replacement APIs are all available and
 well established, and this RFC does not introduce anything new. Also having certain event
@@ -125,13 +129,8 @@ should not be a major issue.
 
 ## Alternatives
 
-* we could keep support in place, and eventually work on optimizations that minimize the 
-performance impact
-
-* we could even go one step further instead, and deprecate support also for `mousemove`, which is a
-(bubbling) event that does not have the higher implementation cost as `mouseenter`/`mouseleave`, 
-but nevertheless requires the EventDispatcher to optimistically handle all those super high-volume
-events which are very rarely needed/dispatched.
+We could keep support in place, and eventually work on optimizations that minimize the 
+performance impact.
 
 ## Unresolved questions
 
--- a/text/0486-deprecate-mouseenter.md
+++ b/text/0486-deprecate-mouseenter.md
@@ -82,7 +82,7 @@ export default Component.extend({
   didInsertElement() {
     this._super(...arguments);
     this._mouseenter = bind(this, this.mouseenter);
-    this.element.addEventListener('mouseenter', handler);
+    this.element.addEventListener('mouseenter', this._mouseenter);
   },
   willDestroyElement() {
     this._super(...arguments);
--- a/text/0486-deprecate-mouseenter.md
+++ b/text/0486-deprecate-mouseenter.md
@@ -139,7 +139,7 @@ After (based on [RFC471](https://github.com/emberjs/rfcs/blob/master/text/0471-o
 The deprecation guide should explain the transition path as shown above.
 
 The Guide's [Handling Events](https://guides.emberjs.com/release/components/handling-events/#toc_event-names)
-section must remove the reference to `mouseenter`, `mouseleave` and `mousemove.
+section must remove the reference to `mouseenter`, `mouseleave` and `mousemove`.
 
 Other than that, no changes are required, as the replacement APIs are all available and
 well established, and this RFC does not introduce anything new. Also having certain event
--- a/text/0486-deprecate-mouseenter.md
+++ b/text/0486-deprecate-mouseenter.md
@@ -66,33 +66,58 @@ Before:
 ```js
 import Component from '@ember/component';
 
-export default Component.extend({
+export default class MyComponent extends Component {
   mouseEnter(e) {
     // do something
   }
-});
+}
 ```
 
 After:
 
 ```js
 import Component from '@ember/component';
-import { bind } from '@ember/runloop';
+import { action } from '@ember/object';
 
-export default Component.extend({
-  mouseenter(e) {
+export default class MyComponent extends Component {
+  @action
+  handleMouseEnter(e) {
     // do something
-  },
-  didInsertElement() {
-    this._super(...arguments);
-    this._mouseenter = bind(this, this.mouseenter);
-    this.element.addEventListener('mouseenter', this._mouseenter);
-  },
-  willDestroyElement() {
-    this._super(...arguments);
-    this.element.removeEventListener('mouseenter', this._mouseenter);
   }
-});
+  
+  didInsertElement() {
+    super.didInsertElement(...arguments);
+    this.element.addEventListener('mouseenter', this.handleMouseEnter);
+  }
+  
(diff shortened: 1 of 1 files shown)


text/0486-deprecate-mouseenter.md
```diff
@@ -0,0 +1,144 @@
...
+Before:
+
+```hbs
+<button {{action "handleMouseEnter" on="mouseenter"}}>Hover</button>
+```
+
+After:
+
+```hbs
+<button onmouseenter={{action "handleMouseEnter"}}>Hover</button>
```
locks: This example should be removed and only the `on` modifier suggested!

text/0486-deprecate-mouseenter.md
```diff
@@ -0,0 +1,144 @@
...
+
+or (based on [RFC471](https://github.com/emberjs/rfcs/blob/master/text/0471-on-modifier.md))
+
+```hbs
+<button {{on "mouseenter" this.handleMouseEnter}}>Hover</button>
+```
+
+## How We Teach This
+
+The deprecation guide should explain the transition path as shown above.
```
locks: Could you write out the explanation instead? The goal for this section is that a contributor can write the deprecation guide and update existing documentation by reading this section alone.

simonihmig: This would mean just duplicating the before/after examples. Does that make sense really?

Fwiw, I looked at this deprecation RFC when writing this, which does a similar thing: https://github.com/emberjs/rfcs/blob/master/text/0449-deprecate-partials.md

rwjblue:
Thoughts on also adding `mousemove` to the list to be deprecated? I realize it doesn't have all of the same issues (since it bubbles), but it definitely has the same negative performance impact.

text/0486-deprecate-mouseenter.md
rwjblue: Ya, I don't really think it makes sense to duplicate the same content here. If the content needs to be moved into this section instead of transition path I guess thats fine, but it means that the transition path would just say "read the deprecation guide in the "how we teach this" below" which is a bit odd.

rwjblue:
Thanks for putting this together @simonihmig!

simonihmig:
> Thoughts on also adding mousemove to the list to be deprecated? 

Yes, I mentioned this under "Alternatives", and personally would be inclined to also include this!

Here's a profiler view when just moving the mouse over an Ember app: 

image

The first two scripting executions (mouseout + mouseover) are for mouseenter/leave support, the third is just the normal event listener for mousemove. All these take just a fraction of a millisecond, though at large scale it looks like this:

image

text/0486-deprecate-mouseenter.md
```diff
@@ -0,0 +1,144 @@
...
+import { bind } from '@ember/runloop';
+
+export default Component.extend({
+  mouseenter(e) {
+    // do something
+  },
+  didInsertElement() {
+    this._super(...arguments);
+    this._mouseenter = bind(this, this.mouseenter);
+    this.element.addEventListener('mouseenter', handler);
```
jessica-jordan: I'm wondering if it'd be

```suggestion
    this.element.addEventListener('mouseenter', this._mouseenter);
```

simonihmig: Uh, indeed! Thanks @jessica-jordan!

rwjblue:
@simonihmig - Would you mind updating to include `mousemove`? I'm adding this to an upcoming core team meeting to see about advancing...

simonihmig:
> Would you mind updating to include mouseover and mousemove?

@rwjblue updated this to include `mousemove`! 

But `mouseover` is not supported anyway [here](https://github.com/emberjs/ember.js/blob/master/packages/%40ember/-internals/views/lib/system/event_dispatcher.js#L71-L97). It was used for the `mouseEnter` implementation internally, but does not get dispatched itself, AFAICT.

rwjblue:
Ah, sorry for the confusion thank you @simonihmig!

text/0486-deprecate-mouseenter.md
```diff
@@ -0,0 +1,137 @@
...
+import Component from '@ember/component';
+
+export default Component.extend({
+  mouseEnter(e) {
+    // do something
+  }
+});
+```
+
+After:
```
rwjblue: This "after" example is pretty confusing to read/review. Here are a couple concrete things to be improved:

* The nuance that it is using `mouseenter` vs `mouseEnter` is **very** subtle. I'd like to see this tweaked to be something a bit less confusing (like `handleMouseEnter`) and use `@action` instead of manually binding in `didInsertElement`
* Would love to see two "After" examples, one similar to the existing one (with the tweaks suggested above) and anther for folks willing to migrate to "outer HTML" in their template (and then they could directly use `{{on` in the template, instead of `didInsertElement`).

text/0486-deprecate-mouseenter.md
```diff
@@ -0,0 +1,137 @@
...
+  }
+});
+```
+
+#### `{{action}}` modifier
+
+Before:
+
+```hbs
+<button {{action "handleMouseEnter" on="mouseenter"}}>Hover</button>
```
rwjblue: Is this correct? I thought we used `mouseEnter` for the `on` value here also?

text/0486-deprecate-mouseenter.md
simonihmig: @rwjblue by using `@action` you imply changing that to a native class? I am fine with that, just thought that as the Guides are still showing "legacy" class examples, I should stick to them as well for now?

text/0486-deprecate-mouseenter.md
simonihmig: Right! 🤦‍♂️ Fixed.

text/0486-deprecate-mouseenter.md
simonihmig: @rwjblue updated the examples according to your feedback! Using native classes now.

rwjblue: Sorry I missed this yesterday! You can use `action` in classic components:

```js
  handleMouseEnter: action(function(event) {

  })
})
```

simonihmig: Ah, makes sense. Though I haven't seen that yet in the wild...

@rwjblue do you want me to revert my changes back to classic components? 

I just wonder if we shouldn't stick with native classes in the examples, as native classes and decorators support have already landed, and `{{on}}` hasn't yet AFAIK. And given that it is going to take some time until the deprecation hits a stable version, we might already have landed Octane and native class examples in the Guides?

text/0486-deprecate-mouseenter.md
```diff
@@ -0,0 +1,162 @@
...
+```hbs
+<button {{on "mouseenter" this.handleMouseEnter}}>Hover</button>
+```
+
+## How We Teach This
+
+The deprecation guide should explain the transition path as shown above.
+
+The Guide's [Handling Events](https://guides.emberjs.com/release/components/handling-events/#toc_event-names)
+section must remove the reference to `mouseenter`, `mouseleave` and `mousemove.
```
pangratz:
```suggestion
section must remove the reference to `mouseenter`, `mouseleave` and `mousemove`.
```

text/0486-deprecate-mouseenter.md
rwjblue: Ya, this seems reasonable. It's an easy change to move back if we want to, and I don't think its a massive difference to "should we move forward".

text/0486-deprecate-mouseenter.md
```diff
@@ -0,0 +1,162 @@
...
+The deprecation guide should explain the transition path as shown above.
+
+The Guide's [Handling Events](https://guides.emberjs.com/release/components/handling-events/#toc_event-names)
+section must remove the reference to `mouseenter`, `mouseleave` and `mousemove`.
+
+Other than that, no changes are required, as the replacement APIs are all available and
+well established, and this RFC does not introduce anything new. Also having certain event
+types not be supported by the EventDispatcher is not something new, as other high-volume
+events like `mouseover` or `scroll` are also not supported (by default).
+
```
jenweber: Some additional details:

```suggestion
There will also be one deprecation entry for each event in the [deprecations guide](https://github.com/ember-learn/deprecation-app).

```

text/0486-deprecate-mouseenter.md
```diff
@@ -0,0 +1,162 @@
...
+```hbs
+<button {{on "mouseenter" this.handleMouseEnter}}>Hover</button>
+```
+
+## How We Teach This
+
+The deprecation guide should explain the transition path as shown above.
+
+The Guide's [Handling Events](https://guides.emberjs.com/release/components/handling-events/#toc_event-names)
+section must remove the reference to `mouseenter`, `mouseleave` and `mousemove`.
```
jenweber:
```suggestion
section must remove the reference to `mouseenter`, `mouseleave` and `mousemove`. It should also be removed from the API docs for [Components events](https://api.emberjs.com/ember/release/classes/Component) and [event delegation](https://api.emberjs.com/ember/release/classes/Application)
```

Please double check whether these actually need to change. I found them by searching the ember.js codebase for `mouseenter`. A similar search should be done for mouseleave and mousemove in case they are used in examples in other parts of the API docs.

rwjblue:
Discussed this at today's team meeting, and are 👍 on moving this into final comment period.

text/0486-deprecate-mouseenter.md
simonihmig: Good catch, thanks! The references for `Component` should be removed, while the example for `Application` shows a way to opt out from listening to those events, so that looks valid and could maybe stay in place until 4.0?

I haven't found any other references in the API docs...

text/0486-deprecate-mouseenter.md
simonihmig: @jenweber You think there should be separate `.md` content files, i.e. each deprecated event has its own anchor link? And so also different deprecation IDs (in the `deprecate()` call)? 

No problem for me, but why do you think this is useful? The migration examples will be quite repetitive then...

jenweber: I might have misunderstood something from the last framework team meeting. It was described to me that they would be separate depreciations, but if you think it makes sense to combine them, that’s fine by me.

simonihmig: Maybe separate deprecation entries for `Ember.Component` vs. `{{action}}`? That would make sense IMHO!

text/0486-deprecate-mouseenter.md
```diff
@@ -0,0 +1,162 @@
...
+frequently, while they are (most certainly) very rarely used.
+
+This is even more so given that Glimmer Components with their outerHTML semantics do not 
+work with event handler methods, and `{{action}}` will eventually fade away in favor of
+`{{on}}` using native `addListener()`.
+
+## Transition Path
+
+When a `mouseenter`/`mouseleave`/`mousemove` event is dispatched by the EventDispatcher to a component
+or `{{action}}` listening to that event type, a deprecation warning will be issued.
```
simonihmig: I already started implementing this, which made me think about this again. I stated that the deprecation is triggered when the event is dispatched. But this can be problematic, as it relies on a specific user interaction. So there might be cases where this deprecation fires *very* frequently, while in others it might fire very *rarely* (so the developer does not even see it). 

Maybe it's better to trigger that at *creation* time (i.e. in the `CurlyComponentManager` / `ActionModifierManager`). Thoughts?

rwjblue: Yes, indeed. We should fire on component creation (_basically_ the same way we assert when you use event methods in tagless Ember.Component), and for {{action}} we’d do it at template compilation (when it’s a string) or initial render (when the event name is dynamic).

simonihmig:
Updated this to
* show deprecations at creation time rather than when dispatching
* refer to separate deprecation guides for `Ember.Component` vs. `{{action}}`
* include `Component` API docs for references to events that need to be removed

pzuraq:
At today's core team meeting, we have decided to move forward with this RFC. 👍

PR #487, opened 2019-05-09
Custom Model Classes

igorT:
[Rendered](https://github.com/igorT/rfcs/blob/igor/custom-model-classes/text/0487-custom-model-classes.md)

### DIFF
--- a/text/0487-custom-model-classes.md
+++ b/text/0487-custom-model-classes.md
@@ -1,0 +1,243 @@
+- Start Date: 2019-05-09
+- Relevant Team(s): data 
+- RFC PR: [PR](https://github.com/emberjs/rfcs/pull/487)
+- Tracking: [Tracking](https://github.com/emberjs/rfc-tracking/issues/53)
+
+# Custom Model Class RFC
+
+## Summary
+
+This RFC is a follow-up RFC for #293 RecordData and replaces [https://github.com/runspired/rfcs/blob/ember-data-custom-records/text/0000-ember-data-model-factory-for.md#ember-data--modelfactoryfor](https://github.com/runspired/rfcs/blob/ember-data-custom-records/text/0000-ember-data-model-factory-for.md#ember-data--modelfactoryfor)
+
+Create a way for addons and user to define their own implementation of a model class. Adds an `instantiateRecord` method to `Store` which would allow addons and ED itself to offer a clean replacement the default DS.Model with a custom Model class. 
+
+## Motivation
+
+- Allowing Addons to experiment with replacing the default DS.Model implementation ([https://github.com/runspired/rfcs/blob/ember-data-custom-records/text/0000-ember-data-model-factory-for.md#ember-data--modelfactoryfor](https://github.com/runspired/rfcs/blob/ember-data-custom-records/text/0000-ember-data-model-factory-for.md#ember-data--modelfactoryfor))
+- Currently Ember Data conflates schema information and record implementation together as part of properties defined on `DS.Model`.
+DS.Model both defines the schema, that is the attributes and relationships present on a type, and the model implementation, that is the
+actual class which is exposed to the app containing those properties. This has several disadvantages:
+- it makes it hard to optimize the schema information, and ties to a definition of a JS class
+- it makes it hard to statically analyze the schema
+- it requires a 1-1 correspondence between the number of backing types in the system and the number of model classes, resulting
+in an app sometimes having to ship a much higher number of models than it otherwise would have.
+
+This RFC represents the first step in separating the record implementation from schema information.
+
+## Detailed design
+
+After the work in [Record Data Errors RFC](https://github.com/emberjs/rfcs/pull/465),  [Record Data State RFC](https://github.com/emberjs/rfcs/pull/463), and [Request State Service RFC](https://github.com/emberjs/rfcs/pull/466) we have enough coverage in public apis that we can implement the entire DS.Model class with just a few changes to Store APIs. 
+
+We will need five separate changes: 
+
+- A way to instantiate custom Records and for them to be able to access underlying record data
+- A way for those records to be notified of changes in Record Data
+- A way to access schema information from the store and not from the DS.Model static properties
+- Replacement of a few store methods that take DS.Models
+- Deprecate passing DS.Model classes to adapter/serializers/exposing them on Snapshots
+
+## Instantiating and destroying custom Records
+
+We need to make the following changes:
+
+* expose a method on the store for instantiating a record
+* add a hook to be notified when a record is being destroyed for potential cleanup
+* a way for the record to get notified when RecordData values change
+
+The following interface shows the interface of these methods:
+
+```ts
+interface RecordDataWrapper { // proxies a subset of RD methods and hides the rest
+  getAttr(identifier: RecordIdentifier, key: string)
+  isAttrDirty(identifier: RecordIdentifier, key)
+    
+  changedAttributes(identifier: RecordIdentifier)
+  hasChangedAttributes(identifier: RecordIdentifier)
+  rollbackAttributes(identifier: RecordIdentifier)
+  
+  getRelationship(identifier: RecordIdentifier, key: string)
+  
+  setRecordId(identifier: RecordIdentifier, id: string): void;
+  
+  getErrors(identifier: RecordIdentifier)
+  getMeta(identifier: RecordIdentifier)
+  
+  isNew(identifier: RecordIdentifier): boolean;
+  isDeleted(identifier: RecordIdentifier): boolean;
+
+  setDirtyAttribute(identifier, key: string, value: any): void;
+
+  addToHasMany(key: string, identifiers: Identifier[], idx?: number): void;
+  removeFromHasMany(key: string, identifiers: Identifier[]): void;
+  setDirtyHasMany(key: string, identifiers: Identifier[]): void;
+
+  setDirtyBelongsTo(name: string, identifier: Identifier | null): void;
+}
+  
+interface RecordDataFor {
+  (identifier: RecordIdentifier): RecordDataWrapper
+}
+    
+class Store {
+  instantiateRecord(
+    identifier: RecordIdentifier, 
+    createRecordArgs: { [key: string]: any }, // args passed in to store.createRecord() and processed by recordData to be set on creation
+    recordDataFor: RecordDataFor, 
+    notificationManager: NotificationManager): unknown
+    
+  teardownRecord(record): void
+}
+```
+
+Instead of passing the entire RecordData Class to `instantiateRecord`, we pass in a lookup method that returns a record data wrapper which exposes only the local facing methods and hide the server/adapter facing methods that the Model should not have access to.
+
+We also expose `teardownRecord`, which will get called whenever a record is getting unloaded, or otherwise disposed of (if we add future ways of destroying records that are uncoupled from unloading) from the identity map, thus giving addons an opportunity to perform cleanup. 
+
+## Change Notification
+
+In order for the record class to learn about changes to the underlying data, it will be passed a `NotificationManager` as a constructor argument, which will allow it to
+subscribe and unsubscribe to notifications of underlying changes to the data. Once `RecordData` calls one of the notification methods, the notification manager will call 
+any registered callback for the given identifier, and pass in the type of the notification, allowing the record the opportunity to repull the data if needed. There are no guarantees around the timing of the notification callback being called after `RecordData` informs the store of changes. We expect that in a modern Ember app with tracked properties, this wouldn't be the common path for tracking changes. 
+
+```ts
+function unsubscribe(token: UnsubscribeToken)
+
+interface NotificationCallback {
+  (identifier: Identifier, notificationType: 'attributes' | 'relationships' | 'errors'  | 'meta'  | 'unload'): void;
+}
+    
+interface NotificationManager {
+  subscribe(identifier, NotificationCallback): UnsubscribeToken
+ }
+```
+
+## Exposing schema information
+
+We currently keep schema information on `DS.Model` class. In order to allow for custom Model implementations we need to allow lookup of schema info from the store. We already have specified a schema api that RecordData consumes: `attributesDefinitionFor` and `relationshipDefinitionFor`. We would define on a schema interface that the store would expose and addons could use. **The schema methods are not ergonomic on purpose.** They match the current Record Data apis and are designed as a stepping stone on the way of having a better, user facing schema APIs. Addons could provide their own `SchemaDefinitionService` by calling `registerSchemaDefinitionService`. We would initially not allow calling `registerSchemaDefinitionService` more than once, but this constraint could potentially be relaxed in the future. The schema info is currently primarily geared towards being used by the internal relationship handling layer, the serializer/adapter layers and the DebugAdapter. Schema methods support both static and dynamic schema computation. For static schemas, the method can always respond with a schema definition based on the type passed, and for dynamic changing schemas, it can look up the underlying data by the identifer which is passed in. We would also add a method called `doesTypeExist`, which would return `true` if ED knew that a given string is a model type and `false` otherwise.
+  
+
+```ts
+export interface RelationshipDefinition {
+  kind: 'hasMany'| 'belongsTo';
+  type: string;
+  options: { [key: string]: any } ;
+  name: string;
+}
+    
+interface RelationshipsDefinition {
+  [key: string]: RelationshipDefinition
+}
+    
+interface AttributeDefinition {
+  name: string;
+  options: { [key: string]: any };
+  type: string;
+}
+
+interface AttributesDefinition {
+  [key: string]: AttributeDefinition
+}
+    
+interface SchemaDefinitionService {
+  // Following the existing RD implementation 
+  attributesDefinitonFor(identifier: RecordIdentifier | type: string): AttributesDefiniton
+  
+  // Following the existing RD implementation
+  relationshipsDefinitionFor(identifier: RecordIdentifier | type: string): RelationshipsDefinition
+
+  doesTypeExist(type: string): boolean
+}
+
+class Store {
+  registerSchemaDefinitionService(schema: SchemaDefinitionService): void
+
+  getSchemaDefinitionService(): SchemaDefinitionService
+}
+```
+
+## Adding store methods for manipulating records 
+
+Currently there exist methods on DS.Model that call into `internalModel` for it's functionality. In order for a parallel implementation
+to be possible, we need to expose that functionality through public methods on the store.
+
+```ts
+class Store {
+  saveRecord(record): Promise // equivalent of currently doing record.save()
+  serializeRecord(record): any // equivalent of currently doing record.serialize()
+  relationshipReferenceFor(identifier: RecordIdentifier, key: string): RelationshipReference
+}
+```
+
+this would allow you to have a custom model class like this: 
+
+```ts
+class CustomModel {
+  save() {
+    return this._store.saveRecord(this);
+  }
+}
+```
+
+### Record Arrays
+
+Currently Ember Data manages live Record Arrays which are returned as a response to query methods such as `findAll` or `queryRecords`. Because ED can track records which are returned from `instantiateRecord`, it will be able to seamlessly manage custom models which are part of Ember Data record arrays and clean them up correctly. 
+
+If an addon implements it's own record array like structures, it will be able to manage membership of Ember Data default records by subclassing `teardownRecord`, which gives it a convinient place to listen for a record being destroyed.
+
+
+## Deprecating DS.Model being passed to serializers and adapters and store.modelFor
+
+Currently some adapter/serializer methods get the underlying class passed in. For example:
+
+```ts
+// Adapter
+createRecord(store, type, snapshot)
+findRecord(store, type, id, snapshot)
+
+// Serializer
+normalizeResponse(store, primaryModelClass, payload, id, requestType)
+
+// Store
+modelFor(modelName)
+```
+
+### Serializing
+
+When serializing we already have a `Snapshot` passed in as an argument which has all of the information that a `ModelClass` provides. We would deprecate the class argument being passed in, and instruct the user to refactor towards using the `Snapshot`.
+
+For example: 
+```ts
+createRecord(store, type, snapshot) {
+  let url = `/api/${type.modelName}`;
+}
+```
+
+would become
+
+```ts
+createRecord(store, type, snapshot) {
+  let url = `/api/${snapshot.modelName}`;
+}
+```
+
+ For backwards compatibility, we would still lookup the class from the registry, and if we found a class we would return it but would no longer error if null was returned. If we did not find the class, we would create a shim class that exposed a deprecated `modelName`. We would deprecate accessing the properties on the class when passed to serializer/adapter by wrapping it in a proxy in dev mode. Currently Snapshots contain all of the data that is available on the Model class which would be needed for serializing/normalizing.
+
+### Normalizing
+
+While `Snapshot` roughly corresponds to a `Request` like object to be used while serializing, and we intend to in the future refactor it to be a request object, we don't have a similar construct when normalizing. In the future, normalizing will have a corresponding `Response` like object exposed, but for the time being we can still deprecate using the `primaryModelClass` argument for anything other than `modelName`, `eachAttribute`, `eachRelationship`. If a user accesses any other property on the `primaryModelClass` they will receive a deprecation.
+
+If there was a custom model instance provided, and we had no corresponding class to pass in to `normalizeResponse` we would pass in a shim class that only exposed the `modelName` property, and `eachAttribute` and `eachRelationship` methods which would proxy to the underlying schema methods, thus allowing the normalization layers to continue working with custom model classes.
+ 
+```ts
+interface ClassSchemaShim {
+  modelName: string;
+  eachAttribute( (name: string, attr: AttributeDefinition): void ): void;
+  eachRelationship( (name: string, attr: AttributeDefinition): void ): void;
+}
+```
+
+
+## How we teach this
+
+This is a very addon/very power user specific api, and would be inappropriate for the default guides. We would document it in the Api docs and potentially if there was a guide for Ember Data addon developers.
+


tchak:
`AttributeSchema` -> `AttributeDefinition`

text/0000-custom-model-classes.md
```diff
@@ -0,0 +1,158 @@
...
+- Replacement of few store methods that take DS.Models
+- Deprecate passing DS.Model classes to adapter/serializers/exposing them on Snapshots
+
+## Instantiating custom Records
+
+First we need to expose a method on the store for instantiating a record, and way for the record to get notified when RecordData values change:
+
+```ts
+interface RecordDataWrapper { // proxies a subset of RD methods and hides the rest
+  performMutation(operation: Operation): void
```
HeroicEric: Should be removed since we're moving away from this

text/0000-custom-model-classes.md
```diff
@@ -0,0 +1,158 @@
...
+- RFC PR: (after opening the RFC PR, update this with a link to it and update the file name)
+- Tracking: (leave this empty)
+
+# Custom Model Class RFC
+
+## Summary
+
+This RFC is a follow-up RFC for #293 RecordData and replaces [https://github.com/runspired/rfcs/blob/ember-data-custom-records/text/0000-ember-data-model-factory-for.md#ember-data--modelfactoryfor](https://github.com/runspired/rfcs/blob/ember-data-custom-records/text/0000-ember-data-model-factory-for.md#ember-data--modelfactoryfor)
+
+Create a pluggable method for instantiating model classes. Add an `instantiateRecord` method to `Store` which would allow addons and ED itself to offer a clean replacement the default DS.Model with a custom Model class.
```
HeroicEric: What is meant by "pluggable" here?

text/0000-custom-model-classes.md
```diff
@@ -0,0 +1,158 @@
...
+## Exposing schema information
+
+We currently keep schema information on `DS.Model` class. In order to allow for custom Model implementations we need to allow lookup of schema info from the store. We already have specified a schema api that RecordData consumes: `attributesDefinitionFor` and `relationshipDefinitionFor`. We would define these methods publicly on the store for addons to use. **These methods are not ergonomic on purpose.** They match the current Record Data apis and are designed as a stepping stone on the way of having a better, user facing schema APIs. The schema info is currently primarily geared towards being used by the internal relationship handling layer and the serializer/adapter layers.
+
+```ts
+export interface RelationshipSchema {
+  kind: 'hasMany'| 'belongsTo';
+  type: string;
+  options: { [key: string]: any } ;
+  name: string;
```
runspired: we should validate that this is in fact what we currently expose / want to document as info concerning `inversePropertyName` `inverseType` and `isPolymorphic` is typically available as well.

text/0000-custom-model-classes.md
```diff
@@ -0,0 +1,158 @@
...
+  subscribeUnload(identifier, NotificationCallback)
+  unsubscribe(identifier, NotificationCallback)
+ }
+    
+class Store {
+  instantiateRecord(
+    identifier: RecordIdentifier, 
+    createArgs: CreateArgs, 
+    recordDataFor: RecordDataFor, 
+    notificationManager: NotificationManager): unknown)
```
HeroicEric:
```suggestion
    notificationManager: NotificationManager): unknown
```

text/0000-custom-model-classes.md
```diff
@@ -0,0 +1,158 @@
...
+}
+```
+    
+
+## Replacing Store methods that deal with DS.Model
+
+```ts
+class Store {
+  // deprecate
+  scheduleSave(model: DS.Model):Promise
```
HeroicEric:
```suggestion
  scheduleSave(model: DS.Model): Promise
```

text/0000-custom-model-classes.md
```diff
@@ -0,0 +1,158 @@
...
+}
+```
+
+## Deprecating DS.Model being passed to serializers and adapters and store.modelFor
+
+Currently some adapter/serializer methods get the underlying class passed in. For example:
+
+```ts
+// Adapter
+createRecord(store, type, snapshot)
```
HeroicEric: I think it would be helpful to provide before & after examples here to show how to implement the changes required

text/0000-custom-model-classes.md
```diff
@@ -0,0 +1,158 @@
...
+  subscribeErrorsChanges(identifier, NotificationCallback)
+  subscribeMetaChanges(identifier, NotificationCallback)
+  subscribeUnload(identifier, NotificationCallback)
+  unsubscribe(identifier, NotificationCallback)
+ }
+    
+class Store {
+  instantiateRecord(
+    identifier: RecordIdentifier, 
+    createArgs: CreateArgs,
```
HeroicEric: the details of `CreateArgs` need to be added

text/0000-custom-model-classes.md
```diff
@@ -0,0 +1,158 @@
...
+## Summary
+
+This RFC is a follow-up RFC for #293 RecordData and replaces [https://github.com/runspired/rfcs/blob/ember-data-custom-records/text/0000-ember-data-model-factory-for.md#ember-data--modelfactoryfor](https://github.com/runspired/rfcs/blob/ember-data-custom-records/text/0000-ember-data-model-factory-for.md#ember-data--modelfactoryfor)
+
+Create a pluggable method for instantiating model classes. Add an `instantiateRecord` method to `Store` which would allow addons and ED itself to offer a clean replacement the default DS.Model with a custom Model class. 
+
+## Motivation
+
+- Allowing Addons to experiment with replacing the default DS.Model implementation ([https://github.com/runspired/rfcs/blob/ember-data-custom-records/text/0000-ember-data-model-factory-for.md#ember-data--modelfactoryfor](https://github.com/runspired/rfcs/blob/ember-data-custom-records/text/0000-ember-data-model-factory-for.md#ember-data--modelfactoryfor))
+- First step in separating the record implementation from schema information
```
hjdivad: This makes a lot of sense to folks that regularly attend the weekly meetings, but I don't think it's enough context to motivate the various sets of changes for typical Ember Data users who might not have run into the sorts of issues for which `DS.Model` causes problems.

Can we add something here or link to something else to add context for other Ember Data users?

text/0000-custom-model-classes.md
```diff
@@ -0,0 +1,158 @@
...
+
+## Detailed design
+
+After the work in [https://www.notion.so/Errors-RFC-8e9e1fd4e2474ed5a8539ed343047784](https://www.notion.so/Errors-RFC-8e9e1fd4e2474ed5a8539ed343047784),  [https://www.notion.so/Request-State-RFC-b994c09fe3094f2b92e4ff1dd36fa882](https://www.notion.so/Request-State-RFC-b994c09fe3094f2b92e4ff1dd36fa882), [https://www.notion.so/Record-State-on-Record-Data-RFC-3fd05bcd2a83439f96fbcf2bc0f3a581](https://www.notion.so/Record-State-on-Record-Data-RFC-3fd05bcd2a83439f96fbcf2bc0f3a581) we have enough coverage in public apis that we can implement the entire DS.Model class with just a few changes to Store APIs. 
+
+We will need four separate changes: 
+
+- A way to instantiate custom Records and have those records be notified of changes in Record Data
+- A way to access schema information not from the Record Class
```
hjdivad: A way for *what* part(s) fo the system to access schema information?

hjdivad: We should call out the specifics here

text/0000-custom-model-classes.md
```diff
@@ -0,0 +1,158 @@
...
+- Replacement of few store methods that take DS.Models
+- Deprecate passing DS.Model classes to adapter/serializers/exposing them on Snapshots
+
+## Instantiating custom Records
+
+First we need to expose a method on the store for instantiating a record, and way for the record to get notified when RecordData values change:
+
+```ts
+interface RecordDataWrapper { // proxies a subset of RD methods and hides the rest
```
hjdivad: What is the conceptual boundary between those methods exposed and those methods hidden?

text/0000-custom-model-classes.md
```diff
@@ -0,0 +1,158 @@
...
+    identifier: RecordIdentifier, 
+    createArgs: CreateArgs, 
+    recordDataFor: RecordDataFor, 
+    notificationManager: NotificationManager): unknown)
+}
+```
+
+Instead of exposing the entire RecordData Class, we expose only the local facing methods and hide the server/adapter facing methods that the Model should not have access to, by giving a RecordData lookup method to `instantiateRecord`
+
+## Exposing schema information
```
hjdivad: There is no discussion about dynamic/static expectations for these schemas.  It is somewhat implied that they can be dynamic as the function signatures take identifiers, but we should be explicit about this.

text/0000-custom-model-classes.md
```diff
@@ -0,0 +1,158 @@
...
+## Detailed design
+
+After the work in [https://www.notion.so/Errors-RFC-8e9e1fd4e2474ed5a8539ed343047784](https://www.notion.so/Errors-RFC-8e9e1fd4e2474ed5a8539ed343047784),  [https://www.notion.so/Request-State-RFC-b994c09fe3094f2b92e4ff1dd36fa882](https://www.notion.so/Request-State-RFC-b994c09fe3094f2b92e4ff1dd36fa882), [https://www.notion.so/Record-State-on-Record-Data-RFC-3fd05bcd2a83439f96fbcf2bc0f3a581](https://www.notion.so/Record-State-on-Record-Data-RFC-3fd05bcd2a83439f96fbcf2bc0f3a581) we have enough coverage in public apis that we can implement the entire DS.Model class with just a few changes to Store APIs. 
+
+We will need four separate changes: 
+
+- A way to instantiate custom Records and have those records be notified of changes in Record Data
+- A way to access schema information not from the Record Class
+- Replacement of few store methods that take DS.Models
+- Deprecate passing DS.Model classes to adapter/serializers/exposing them on Snapshots
```
hjdivad: Presumably we will depend on a feature flag for this?  We should call it out to make it clearer to custom model implementers what they need to do

text/0000-custom-model-classes.md
```diff
@@ -0,0 +1,158 @@
...
+}
+    
+interface NotificationManager {
+  subscribeAttributeChanges(identifier, NotificationCallback)
+  subscribeRelationshipChanges(identifier, NotificationCallback)
+  subscribeErrorsChanges(identifier, NotificationCallback)
+  subscribeMetaChanges(identifier, NotificationCallback)
+  subscribeUnload(identifier, NotificationCallback)
+  unsubscribe(identifier, NotificationCallback)
+ }
```
hjdivad: Why the asymmetry between the subscription and notification pattern here?  We subscribe with fine-grained methods but notify in a single coarse-grained callback.

hjdivad: There does not seem to be any discussion about the flow for triggering these callbacks.  Presumably the root API is from userland's `storeWrapper.notifyX` methods but in any case we should call out the flow whatever it is.

text/0000-custom-model-classes.md
```diff
@@ -0,0 +1,158 @@
...
+    recordDataFor: RecordDataFor, 
+    notificationManager: NotificationManager): unknown)
+}
+```
+
+Instead of exposing the entire RecordData Class, we expose only the local facing methods and hide the server/adapter facing methods that the Model should not have access to, by giving a RecordData lookup method to `instantiateRecord`
+
+## Exposing schema information
+
+We currently keep schema information on `DS.Model` class. In order to allow for custom Model implementations we need to allow lookup of schema info from the store. We already have specified a schema api that RecordData consumes: `attributesDefinitionFor` and `relationshipDefinitionFor`. We would define these methods publicly on the store for addons to use. **These methods are not ergonomic on purpose.** They match the current Record Data apis and are designed as a stepping stone on the way of having a better, user facing schema APIs. The schema info is currently primarily geared towards being used by the internal relationship handling layer and the serializer/adapter layers.
```
hjdivad: > The schema info is currently primarily geared towards being used by the internal relationship handling layer and the serializer/adapter layers.

What else depends on the schema information?

text/0000-custom-model-classes.md
hjdivad: (moved this comment downstream where it makes more sense)

text/0000-custom-model-classes.md
hjdivad: Also since the classes being given are different, it would be good to have a sense what, if anything is expected to happen between `storeWrapper.notifyWhatever` and `notificationCallback(...)`

igorT: Changed to coarse grained after feedback from @runspired

text/0000-custom-model-classes.md
igorT: Added

igorT:
Action item:
- Define path forward for snapshots after this rfc

text/0000-custom-model-classes.md
igorT: Fixed

text/0000-custom-model-classes.md
igorT: I rewrote the paragraph

text/0000-custom-model-classes.md
igorT: Removed

text/0000-custom-model-classes.md
igorT: Rewritten

igorT:
Maybe instead of passing null as a model class, we pass in a shim object for custom model classes to adapter/serializers.

text/0000-custom-model-classes.md
igorT: Thats a very good point. `isPolymorphic` should be available from the options. We currently in practice expose a bunch of underscored properties for inverses, but all of that data is calculable given the exposed public data. I would prefer to keep it private, and figure out in the schema work if we need to expose it

text/0000-custom-model-classes.md
igorT: Added

text/0000-custom-model-classes.md
igorT: `DebugAdapter` does.

text/0000-custom-model-classes.md
```diff
@@ -92,9 +84,26 @@ class Store {
...
+We currently keep schema information on `DS.Model` class. In order to allow for custom Model implementations we need to allow lookup of schema info from the store. We already have specified a schema api that RecordData consumes: `attributesDefinitionFor` and `relationshipDefinitionFor`. We would define these methods publicly on the store for addons to use. **These methods are not ergonomic on purpose.** They match the current Record Data apis and are designed as a stepping stone on the way of having a better, user facing schema APIs. The schema info is currently primarily geared towards being used by the internal relationship handling layer, the serializer/adapter layers and the DebugAdapter. Schema methods support both static and dynamic schema computation. For static schemas, the method can always respond with a schema definition based on the type passed, and for dynamic changing schemas, it can look up the underlying data by the identifer which is passed in.
```
runspired: It is slightly concerning to me that we have to introduce schemas in this way here vs through a more complete design (which we want to achieve soon). I do wonder if we shouldn't have targeted designing a schema service first.

text/0000-custom-model-classes.md
```diff
@@ -0,0 +1,225 @@
...
+- it makes it hard to optimize the schema information, and ties to a definition of a JS class
+- it makes it hard to statically analyze the schema
+- it requires a 1-1 correspondence between the number of backing types in the system and the number of JS classes, resulting
+in an app sometimes having to ship a much higher number of classes than it otherwise would have.
+
+This RFC represents the first step in separating the record implementation from schema information.
+
+## Detailed design
+
+After the work in [Record Data Errors RFC](https://github.com/emberjs/rfcs/pull/465),  [Record Data State RFC](https://github.com/emberjs/rfcs/pull/463), and [Request State Service RFC](https://github.com/emberjs/rfcs/pull/466) we have enough coverage in public apis that we can implement the entire DS.Model class with just a few changes to Store APIs.
```
hjdivad: Do we have a proof of concept for this in eg something like ember-m3 to make sure we have actually provided enough API to make those use cases possible?

text/0000-custom-model-classes.md
```diff
@@ -170,8 +173,49 @@ normalizeResponse(store, primaryModelClass, payload, id, requestType)
...
+```ts
+normalizeResponse(store, primaryModelClass, payload, id, requestType) {
+   let attrs = store.attributesDefitionFor(primaryModelClass.modelName);
+   Object.keys(attrs).forEach((attr) => ...do something);
+}
+```
 
+If there was a custom model instance provided, and we had no corresponding class to pass in to `normalizeResponse` we would pass in a shim class that only exposed `modelName` property, thus allowing the normalization layers to continue working with custom model classes.
+ 
 ## How we teach this
```
runspired: I'm a bit concerned about the learning story as we push folks to use Snapshots more and more. Currently they are marked as `private` API and there is no public way to generate one (for doing things like testing a serializer/adapter). We may want to add a public `store.createSnapshot` API and work with the learning team on how to properly document classes that are non-user-instantiated but are public API.

igorT: :+1: on documenting, but don't think we need to expose a creation method in order to document. I'm happy to have better snapshot docs as part of this rfc

text/0000-custom-model-classes.md
```diff
@@ -0,0 +1,225 @@
...
+ }
+```
+
+## Exposing schema information
+
+We currently keep schema information on `DS.Model` class. In order to allow for custom Model implementations we need to allow lookup of schema info from the store. We already have specified a schema api that RecordData consumes: `attributesDefinitionFor` and `relationshipDefinitionFor`. We would define these methods publicly on the store for addons to use. **These methods are not ergonomic on purpose.** They match the current Record Data apis and are designed as a stepping stone on the way of having a better, user facing schema APIs. The schema info is currently primarily geared towards being used by the internal relationship handling layer, the serializer/adapter layers and the DebugAdapter. Schema methods support both static and dynamic schema computation. For static schemas, the method can always respond with a schema definition based on the type passed, and for dynamic changing schemas, it can look up the underlying data by the identifer which is passed in. We would also add a method called `isModelType`, which would return `true` if ED knew that a given string is a model type and `false` otherwise.
```
hjdivad: > `isModelType`, which would return `true` if ED knew that a given string is a model type and `false` otherwise.

How would this be determined?  Is this intended to be a user extension point?

text/0000-custom-model-classes.md
igorT: That's why I designed them in a clunky way so we can replace them later and that most of the usage would be hidden by other public APIs. 

The only place you would have to manually deal with them would be in normalization methods. 
A medium term solution would be to not deprecate eachAttribute and eachRelationship being passed to normalize, and implement those in the shim class until a full featy

text/0000-custom-model-classes.md
```diff
@@ -0,0 +1,225 @@
...
+interface AttributesDefinition {
+  [key: string]: AttributeDefinition
+}
+    
+class Store {
+  // Following the existing RD implementation 
+  attributesDefitionFor(identifier: RecordIdentifier | type: string): AttributesDefiniton
+  
+  // Following the existing RD implementation
+  relationshipsDefinitionFor(identifier: RecordIdentifier | type: string): RelationshipsDefinition
```
hjdivad: What does it actually look like for a userland model class to provide schema information this way?  Are users expected to overwrite these methods in their own `store`?

The current implementation here has the store knowing something intimate about `DS.Model`. (in particular that it can read `attributes` and `relationshipsObject` from the model class).  Should we split the store into the "core" store and "default model userland" store?

text/0000-custom-model-classes.md
igorT: Yup, have a spike of both default DS.Model and m3

text/0000-custom-model-classes.md
igorT: Yup

text/0000-custom-model-classes.md
igorT: You would overwrite the methods. If you were doing that, you wouldn't be reading off of the DS.Model presumably. Not quite sure I understand the store split proposal, let's discuss it at the meeting

text/0000-custom-model-classes.md
runspired: I think shim class implementation is worth exploring

text/0000-custom-model-classes.md
hjdivad: @igorT can we post the spike somewhere public as an example people reviewing it can look at to see what this looks like "in the real world"?

text/0000-custom-model-classes.md
runspired: I would like to split the store.

runspired: (That discussion was rapidly approaching and having it now seems a good time)

text/0000-custom-model-classes.md
igorT: Sure thing, I'm in the process of extracting the merged rfcs out of it as well which will make it more readable as well

text/0000-custom-model-classes.md
hjdivad: The suggestion is for there to be a `store.ts` file that does not know anything about `DS.Model`, in part as a way to dogfood that `DS.Model` is actually something that can be swapped out.

text/0000-custom-model-classes.md
hjdivad: I don't see `RecordArrays`s mentioned.  Let's be explicit about the intent here.  Can userland model classes participate in `RecordArray`s? (eg if they're added to record arrays will they automatically be removed when unloaded?  If so does their corresponding `RecordData` have any control over whether or when this happens?)

If we'd like to see `RecordArray`s tackled separately (and I think that's reasonable) we should call that out in this RFC

igorT:
Talk about Record Arrays

text/0000-custom-model-classes.md
```diff
@@ -0,0 +1,218 @@
...
+
+would become
+
+```ts
+createRecord(store, type, snapshot) {
+  let url = `/api/${snapshot.modelName}`;
+}
+```
+
+ For backwards compatibility, we would still lookup the class from the registry, and if we found a class we would return it with a deprecation, but would no longer error if null was returned. If we did not find the class, we would create a shim class that exposed a deprecated `modelName`. We would deprecate accessing the class when passed to serializer/adapter by wrapping it in a proxy in dev mode. Currently Snapshots contain all of the data that is available on the Model class which would be needed for serializing/normalizing.
```
igorT: Misleading, fix deprecation prose

text/0000-custom-model-classes.md
```diff
@@ -0,0 +1,218 @@
...
+  type: string;
+}
+
+interface AttributesDefinition {
+  [key: string]: AttributeDefinition
+}
+    
+class Store {
+  // Following the existing RD implementation 
+  attributesDefitionFor(identifier: RecordIdentifier | type: string): AttributesDefiniton
```
rwjblue:
```suggestion
  attributesDefinitionFor(identifier: RecordIdentifier | type: string): AttributesDefiniton
```

igorT:
Add the overall store signature

text/0000-custom-model-classes.md
hjdivad: Per the discussion at the weekly meeting, this is implied.

dgeb:
I discussed this with @igorT today. 

One concern I've had here is that `type` is used inconsistently between `RelationshipDefinition` and `AttributeDefinition`, but Igor said that this is a legacy inconsistency and that changing this now would lead to breakages.

Also, some method return values on `RecordDataWrapper` are untyped (e.g. `getErrors`, `getMeta`).

I'm fine moving this to FCP if others are. I will give it another pass over the weekend.

igorT:
I have implemented a slightly older version of this api(pending some cleanup) at https://github.com/hjdivad/ember-m3/pull/294. It passes all the tests without any usage of `_internalModel` or other private apis

text/0000-custom-model-classes.md
```diff
@@ -0,0 +1,229 @@
...
+- RFC PR: (after opening the RFC PR, update this with a link to it and update the file name)
+- Tracking: (leave this empty)
+
+# Custom Model Class RFC
+
+## Summary
+
+This RFC is a follow-up RFC for #293 RecordData and replaces [https://github.com/runspired/rfcs/blob/ember-data-custom-records/text/0000-ember-data-model-factory-for.md#ember-data--modelfactoryfor](https://github.com/runspired/rfcs/blob/ember-data-custom-records/text/0000-ember-data-model-factory-for.md#ember-data--modelfactoryfor)
+
+Create a way for addons and user to define their own implemnentation of a model clas. Adds an `instantiateRecord` method to `Store` which would allow addons and ED itself to offer a clean replacement the default DS.Model with a custom Model class.
```
rwjblue:
```suggestion
Create a way for addons and user to define their own implementation of a model class. Adds an `instantiateRecord` method to `Store` which would allow addons and ED itself to offer a clean replacement the default DS.Model with a custom Model class. 
```

text/0000-custom-model-classes.md
```diff
@@ -0,0 +1,229 @@
...
+
+## Motivation
+
+- Allowing Addons to experiment with replacing the default DS.Model implementation ([https://github.com/runspired/rfcs/blob/ember-data-custom-records/text/0000-ember-data-model-factory-for.md#ember-data--modelfactoryfor](https://github.com/runspired/rfcs/blob/ember-data-custom-records/text/0000-ember-data-model-factory-for.md#ember-data--modelfactoryfor))
+- Currently Ember Data conflates schema information and record implementation together as part of properties defined on `DS.Model`.
+DS.Model both defines the schema, that is the attributes and relationships present on a type, and the model implementation, that is the
+actual class which is exposed to the app containing those properties. This has several disadvantages:
+- it makes it hard to optimize the schema information, and ties to a definition of a JS class
+- it makes it hard to statically analyze the schema
+- it requires a 1-1 correspondence between the number of backing types in the system and the number of JS classes, resulting
```
rwjblue:
```suggestion
- it requires a 1-1 correspondence between the number of backing types in the system and the number of model classes, resulting
```

text/0000-custom-model-classes.md
```diff
@@ -0,0 +1,229 @@
...
+in an app sometimes having to ship a much higher number of classes than it otherwise would have.
```
rwjblue:
```suggestion
in an app sometimes having to ship a much higher number of models than it otherwise would have.
```

text/0000-custom-model-classes.md
rwjblue: Was this addressed?

text/0000-custom-model-classes.md
```diff
@@ -0,0 +1,229 @@
...
+This RFC represents the first step in separating the record implementation from schema information.
+
+## Detailed design
+
+After the work in [Record Data Errors RFC](https://github.com/emberjs/rfcs/pull/465),  [Record Data State RFC](https://github.com/emberjs/rfcs/pull/463), and [Request State Service RFC](https://github.com/emberjs/rfcs/pull/466) we have enough coverage in public apis that we can implement the entire DS.Model class with just a few changes to Store APIs. 
+
+We will need five separate changes: 
+
+- A way to instantiate custom Records and for them to be able to access underlying record data
+- A way for those records be notified of changes in Record Data
```
rwjblue:
```suggestion
- A way for those records to be notified of changes in Record Data
```

text/0000-custom-model-classes.md
```diff
@@ -0,0 +1,229 @@
...
+- A way to access schema information not from the Record Class
```
rwjblue: This phrasing is confusing to me, what specifically does "not from the Record class" mean here?

text/0000-custom-model-classes.md
```diff
@@ -0,0 +1,229 @@
...
+- Replacement of few store methods that take DS.Models
```
rwjblue:
```suggestion
- Replacement of a few store methods that take DS.Models
```

text/0000-custom-model-classes.md
igorT: https://github.com/hjdivad/ember-m3/pull/294 and https://github.com/emberjs/data/tree/igor/schema-temp

text/0000-custom-model-classes.md
```diff
@@ -0,0 +1,229 @@
...
+- Deprecate passing DS.Model classes to adapter/serializers/exposing them on Snapshots
+
+## Instantiating and destroying custom Records
+
+First we need to expose a method on the store for instantiating a record, a hook to be notified when a record is being destroyed for potential cleanup and a way for the record to get notified when RecordData values change:
```
rwjblue: This would probably be easier to grok as a list:

```suggestion
We need to make the following changes:

* expose a method on the store for instantiating a record
* add a hook to be notified when a record is being destroyed for potential cleanup
* a way for the record to get notified when RecordData values change

The following interface shows the interface of these methods:
```

text/0000-custom-model-classes.md
```diff
@@ -0,0 +1,229 @@
...
+    store: Store,
+    createRecordArgs: { [key: string]: any }, // args passed in to store.createRecord() and processed by recordData to be set on creation
+    recordDataFor: RecordDataFor, 
+    notificationManager: NotificationManager): unknown
+    
+  willDestroyRecord(record): void
+}
+```
+
+Instead of exposing the entire RecordData Class, we expose only the local facing methods and hide the server/adapter facing methods that the Model should not have access to, by giving a RecordData lookup method to `instantiateRecord`
```
rwjblue: > Instead of exposing the entire RecordData Class

On what? Needs a slight prose tweak here

text/0000-custom-model-classes.md
igorT: Added a RecordArrays paragraph

text/0000-custom-model-classes.md
```diff
@@ -0,0 +1,242 @@
...
+}
+  
+interface RecordDataFor {
+  (identifier: RecordIdentifier): RecordDataWrapper
+}
+    
+class Store {
+  instantiateRecord(
+    identifier: RecordIdentifier, 
+    store: Store,
```
dgeb:
```suggestion
```

text/0000-custom-model-classes.md
```diff
@@ -0,0 +1,242 @@
...
+- A way to access schema information not from the Record Class
+- Replacement of a few store methods that take DS.Models
+- Deprecate passing DS.Model classes to adapter/serializers/exposing them on Snapshots
+
+## Instantiating and destroying custom Records
+
+We need to make the following changes:
+
+* expose a method on the store for instantiating a record
+* add a hook to be notified when a record is being destroyed for potential cleanup
```
dgeb: Clarify interaction with `unloadRecord`

text/0000-custom-model-classes.md
```diff
@@ -0,0 +1,242 @@
...
+}
+    
+interface SchemaDefinitionService {
+  // Following the existing RD implementation 
+  attributesDefinitonFor(identifier: RecordIdentifier | type: string): AttributesDefiniton
+  
+  // Following the existing RD implementation
+  relationshipsDefinitionFor(identifier: RecordIdentifier | type: string): RelationshipsDefinition
+
+  isModelType(type: string): boolean
```
dgeb: I think `isModelType` is pretty awkward. Something along the lines of `doesTypeExist` or `typeExists` seems a little more clear (imo).

dgeb:
I've left a couple suggestions but don't consider anything blocking.

In addition to the comments I left, I'd like to eventually iron out inconsistencies between usage of `type` in the relationship and attribute definitions, but I understand those interfaces are not being introduced by this RFC.

PR #489, opened 2019-05-16
Use Empress to render the Ember RFCs

mansona:
Howdy folks 👋 

This is not really an RFC, it is a proposed implementation change to how we currently "render" our RFCs. 

As it constitutes a reasonable change to the infrastructure, and will end up changing some URLs, I will format this PR in a similar way to a normal RFC so that I explain my motivations and why I think this is a good way forward.

## Summary

Currently we are using a technology called `mdbook` to render a website for reading Ember RFCs in a slightly better format than just reading Markdown files directly. This website https://emberjs.github.io/rfcs/ is linked as part of the description of this repo and I know that a few Core Team Members prefer to use these rendered pages when linking to RFCs from external sources.

I propose that we stop using `mdbook` and use some of the [Empress](https://github.com/empress) techniques to render these pages. As part of this move, I would like to also see the **official** URL of an RFC to move from https://emberjs.github.io/rfcs/0425-website-redesign.html to https://rfcs.emberjs.com/0425-website-redesign/. This would then be considered a stable URL and will require us to maintain redirects if we ever change them.

## Motivation

I think there are a number of benefits to this move: 

- we will have the official link of an RFC under an Ember-owned subdomain
- we will be using Ember to host important Ember infrastructure
- the user experience of this new implementation is significantly smoother (because it is a SPA with prember)
- embedded images will actually work in the new setup
- PRs to the RFC repo will come with a Netlify demo url, finally fixing the age-old problem of needing to link to a "rendered" Markdown 🎉
- when the Ember Website Redesign is complete we can also move to create a new design for the RFCs that uses the [ember-styleguide](https://github.com/ember-learn/ember-styleguide) because the implementation is in Ember
- we have the ability to create more structured data that essentially provides a consumable API for Ember RFCs with little to no effort

I have already deployed the contents of this PR to Netlify which you can see here: https://empress-rfcs.netlify.com/0425-website-redesign/. You will note that the embedded images are working (where they are not on the [corresponding page on the current setup](https://emberjs.github.io/rfcs/0425-website-redesign.html)) and as you navigate around the page you will see the experience is much smoother because you don't get the full page refresh.

## Detailed design

I will not go into an overly detailed design here as the work is already done (and you can check the repos for implementation details) and anyone who has been following my adventures in Empress will find this approach very familiar. I will lay out a brief architectural overview just so you can follow along. 

[Empress](https://github.com/empress) is a set of technologies and a particular approach for building content-based websites in Ember. It is not a single installable addon that does all this work, even though there are common elements. 

Empress is currently the technology that is powering the Ember Guides through [Guidemaker](https://github.com/empress/guidemaker) and [guidemaker-ember-template](https://github.com/ember-learn/guidemaker-ember-template). Currently the most used Empress project is [empress-blog](https://github.com/empress/empress-blog) which will eventually replace the technology powering the current Ember Blog with the [empress-blog-ember-template](https://github.com/ember-learn/empress-blog-ember-template). If you are interested in that work you can follow along on the [WIP PR on the ember-blog repo](https://github.com/ember-learn/ember-blog/pull/7)

The main benefit of Empress is that it allows you to effectively separate the content generation (through Markdown) and the implementation of the Ember application, and the implementation of templates and styles. As you will see from this PR there is next to no change to any of the existing files, you will see that I am adding an essentially empty Ember app with two key dependencies: 

- [rfc-process](https://github.com/empress/rfc-process)
- [rfc-process-mdbook-template](https://github.com/empress/rfc-process-mdbook-template)

these two addons work together to build the Markdown into the relevant JSON:API structure required for the app, construct the app routes required, and provide the templates and styling you see whey you visit the Netlify example url. I have opted to implement essentially a copy of the mdbook style for the first iteration so that we don't have too much change at once 😉 

When it comes to redesigning this RFC website it is a matter of creating a new template addon (`rfc-process-ember-template`) and then replacing the dependency in this repo once it is ready. This is a very well proven technique that is common across all Empress "products".

## Drawbacks

There is at least one drawback that I can see from this approach: 

### Existing links

As this will be changing the canonical link of each of the RFCs, any existing inbound links to the https://emberjs.github.io/rfcs/ website will be broken. 

Although this is clearly an important drawback it also supports my belief that it is paramount that we are in full control over the URLs and domains that we use for important projects like the Ember RFCs. We cannot just simply create a redirect rule because GitHub does not support server redirects (like a 301 HTTP redirect). 

#### Mitigation

The best way to mitigate this problem is to do a once-off job to generate html files that do client-side redirects in all of the same locations of existing links. For example, the https://emberjs.github.io/rfcs/0425-website-redesign.html page would be replaced with a html file that contains something like: 

```html
<p> this document has moved to https://rfcs.emberjs.com/0425-website-redesign/</p>
<p> you will get automatically redirected in 5 seconds</p>
```

There is an added benefit to this approach because we can also add a `<canonical>` meta tag pointing to the same location, which means we will be able to redirect at least a fraction of the SEO value of inbound links to the new location.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,17 +1,27 @@
-language: rust
+---
+language: node_js
+node_js:
+  - 12
+
+sudo: false
+dist: trusty
+
+addons:
+  chrome: stable
+
+cache:
+  directories:
+    - $HOME/.npm
+
+env:
+  global:
+    # See https://git.io/vdao3 for details.
+    - JOBS=1
+
+before_install:
+  - npm config set spin false
+
 script:
-  - rev=$(git rev-parse --short HEAD)
-  - (cargo install mdbook --vers 0.0.26 --force || true)
-  - ./generate-book.sh
-
-branches:
-  only:
-    - master
-
-deploy:
-  provider: pages
-  skip_cleanup: true
-  github_token: $GITHUB_TOKEN
-  local_dir: ./book/
-  on:
-    branch: master
+  - npm run lint:hbs
+  - npm run lint:js
+  - npm test
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,20 @@
+# EditorConfig helps developers define and maintain consistent
+# coding styles between different editors and IDEs
+# editorconfig.org
+
+root = true
+
+
+[*]
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+indent_style = space
+indent_size = 2
+
+[*.hbs]
+insert_final_newline = false
+
+[*.{diff,md}]
+trim_trailing_whitespace = false
--- a/.ember-cli
+++ b/.ember-cli
@@ -1,0 +1,9 @@
+{
+  /**
+    Ember CLI sends analytics information by default. The data is completely
+    anonymous, but there are times when you might want to disable this behavior.
+
+    Setting `disableAnalytics` to true will prevent any data from being sent.
+  */
+  "disableAnalytics": false
+}
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,4 +1,6 @@
-# See https://help.github.com/ignore-files/ for more about ignoring files.
+# unconventional js
+/blueprints/*/files/
+/vendor/
 
 # compiled output
 /dist/
@@ -9,15 +11,8 @@
 /node_modules/
 
 # misc
-/.env*
-/.pnp*
-/.sass-cache
-/connect.lock
 /coverage/
-/libpeerconnection.log
-/npm-debug.log*
-/testem.log
-/yarn-error.log
+!.*
 
 # ember-try
 /.node_modules.ember-try/
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,0 +1,42 @@
+module.exports = {
+  root: true,
+  parserOptions: {
+    ecmaVersion: 2017,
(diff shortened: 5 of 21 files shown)


rwjblue:
Thanks for working on this @mansona! I love it!!

.travis.yml
```diff
@@ -1,17 +1,27 @@
...
-  on:
-    branch: master
+env:
+  global:
+    # See https://git.io/vdao3 for details.
+    - JOBS=1
+
+before_install:
+  - npm config set spin false
+
```
kategengler: How does this deploy?

rwjblue:
We discussed this in todays core team call, and in general are absolutely in favor of the main goals/design here. The only concern was around colocating the Ember app in this repo along with the markdown content. Specifically, we are worried that the increased "PR noise" (dependabot or manual updates) will reduce the effectiveness of this repo as a way for our users to "watch" the discussion on future APIs.

@mansona - Have you considered this point already? It seems like there could be a few other ways to handle this to avoid that issue (e.g. separate repo for the app + git submodules, etc), what are the downsides there?

wagenet:
@mansona this seems like this needs a bit more work from you. Are you still interested in moving it forwards?

mansona:
Hey sorry I forgot this was still open. We have moved this in a new direction so it doesn't require anything to be committed to the RFC repo.

In fact "empress" is already being used to deploy the new https://rfcs.emberjs.com/ site 😉

PR #492, opened 2019-05-21
Fix a typo and list formatting in 0369-deprecate-computed-clobberability.md

chriskrycho:
- makes list items properly indented below their parent item
- a case of "User's" to "Users" where it's not possessive


rwjblue:
Thanks!

PR #494, opened 2019-05-30
Async Observers

pzuraq:
[Rendered](https://github.com/emberjs/rfcs/blob/async-observers/text/0494-async-observers.md)


GavinJoyce:
Thanks for working on this @pzuraq.

I like that this RFC allows apps that currently use some observers to incrementally move away from them while incrementally moving to the Octane programming model at the same time.

Am I correct in saying that this RFC will allow addons like `ember-composable-helpers` migrate away from [detecting argument changes using synchronous observers](https://github.com/DockYard/ember-composable-helpers/issues/318) to using arguments which are tracked properties and specifically that it should be possible for helpers like [`sort-by`](https://github.com/DockYard/ember-composable-helpers/blob/07db325d76f0952ddd715de189f8332612585b9b/addon/helpers/sort-by.js) to detect the type of arguments received and only use observers when the inputs are not tracked properties?

pzuraq:
I would take it a step further actually, I believe helpers like `sort-by` will be able to convert and remove observers even if the properties they are passed are normal CPs or properties updated using `set`, since autotracking works with those too 😄 helpers and computed properties like Ember's own `sortBy` helper or the ones included in `ember-macro-helpers` should be able to update too. I believe the majority of, in not all, cases where sync observers are absolutely necessary can be solved with autotracking.

text/0000-async-observers.md
```diff
@@ -0,0 +1,167 @@
+- Start Date: 2019-05-30
+- Relevant Team(s): Ember.js, Learning
+- RFC PR:
```
rwjblue:
```suggestion
- RFC PR: https://github.com/emberjs/rfcs/pull/494
```

text/0000-async-observers.md
```diff
@@ -0,0 +1,167 @@
+- Start Date: 2019-05-30
```
rwjblue: Can you rename the file to `text/0494-async-observers.md`?

snewcomer:
Pleasure reading this RFC!  Would you mind explaining `autotracking`?  Any links to some code I can look at?  I look to help updating some of these packages in the future.

pzuraq:
I think the best explanation for the moment is in the [Tracked Properties RFC](https://github.com/emberjs/rfcs/blob/master/text/0410-tracked-properties.md). At some point I want to do a bit of a deeper dive in a blog post, but it's going to take a while to write, so I'm unsure when I'll get around to it. Hopefully sooner rather than later!

richard-viney:
The "Rendered" link needs updating as it's pointing to https://github.com/emberjs/rfcs/blob/async-observers/text/0000-async-observers.md but the `0000` here needs to be `0494`.

rwjblue:
We discussed this in todays core team meeting, and are 👍 on moving into final comment period.

rwjblue:
FYI - Updated the `rendered` link above.

PR #499, opened 2019-06-10
An alternative to Controllers

NullVoxPopuli:
[rendered](https://github.com/NullVoxPopuli/rfcs/blob/controller-migration-path/text/0000-controller-migration-path.md)

depends on: 
   - ~[RFC 496: Handlebars Strict Mode](https://github.com/emberjs/rfcs/pull/496)~
   - [RFC 454: SFC & Template Import Primitives](https://github.com/emberjs/rfcs/pull/454)
   - ~~[RFC 380: Add queryParams to the router service](https://github.com/emberjs/rfcs/pull/380)~~
     - or: [RFC 712: Query Params as Derived Data](https://github.com/emberjs/rfcs/pull/712) 
             and [RFC 715: Arbitrary Query Params](https://github.com/emberjs/rfcs/pull/715) 

Update (2020-02-08):
 - I need to rewrite some stuff to account for:
   - RFCs #454 and #496 aren't actually a requirement, but more just fit along with my ideal framework semantics
   - ~~[RFC 570: URL Manager](https://github.com/emberjs/rfcs/pull/570) replaced #380 (and is required, because without controllers, we need a way to handle query params)~~
   
 Update (2020-03-03):
 - I need to rewrite some stuff to account for:
    - new derived data approach to query params and arbitrary query params
    - Need to discuss ideas presented in Ember Discord after strict-mode / template imports started being experimenting with and folks were trying to figure out how to make route templates less awkward: https://discord.com/channels/480462759797063690/518154533143183377/815728311841062944

### DIFF
--- a/text/0000-controller-migration-path.md
+++ b/text/0000-controller-migration-path.md
@@ -271,19 +271,6 @@ export function parseError(error: any): ParsedError {
 ```
 
 
-## How we teach this
-
-> What names and terminology work best for these concepts and why? How is this
-idea best presented? As a continuation of existing Ember patterns, or as a
-wholly new one?
-
-> Would the acceptance of this proposal mean the Ember guides must be
-re-organized or altered? Does it change how Ember is taught to new users
-at any level?
-
-> How should this feature be introduced and taught to existing Ember
-users?
-
 ## Drawbacks
 
 These changes would require many changes to existing apps, and there likely won't be an ability to codemod people's code to the new way of doing things. This could greatly slow down the migration, or have people opt to not migrate at all. It's very important that every use case for controllers today _can_ be implemented using the aforementioned techniques. If people are willing to share their controller scenarios, we can provide a library of examples of translation so that others may migrate more quickly.
--- a/text/0000-controller-migration-path.md
+++ b/text/0000-controller-migration-path.md
@@ -236,6 +236,7 @@ Given a visit to `/posts`
 2. lookup a component named "posts", `component:posts` and `template:components/posts`.   
   NOTE: if the `template:posts` template was found, this step does not happen.  
   a. this component is invoked _as if_ `template:posts` was defined as `<Posts @model={{this.model}} />`
+3. default to `{{outlet}}` for sub routes
 
 ## Drawbacks
 
--- a/text/0000-controller-migration-path.md
+++ b/text/0000-controller-migration-path.md
@@ -236,7 +236,7 @@ Given a visit to `/posts`
 2. lookup a component named "posts", `component:posts` and `template:components/posts`.   
   NOTE: if the `template:posts` template was found, this step does not happen.  
   a. this component is invoked _as if_ `template:posts` was defined as `<Posts @model={{this.model}} />`
-3. default to `{{outlet}}` for sub routes
+3. default to `{{outlet}}` or `{{yield}}` for sub routes
 
 ## Drawbacks
 
--- a/text/0000-controller-migration-path.md
+++ b/text/0000-controller-migration-path.md
@@ -7,23 +7,23 @@
 
 ## Summary
 
+ - Goal: For components to eventually make up the entirety of the view layer
  - Provide alternative APIs that encompass existing behavior currently implemented by controllers.
    1. query params
    2. passing state and handling actions to/from the template
 
  - Out of scope:
-   - Deprecating and eliminating controllers -- but the goal of this RFC is to provide a path to remove controllers from Ember altogether. 
+   - Deprecating and eliminating controllers -- but the goal of this RFC is to provide a _path_ to remove controllers from Ember altogether. 
 
  - Not Proposing
-   - Adding the ability to access more properties than `model` from the route template
    - Adding the ability to invoke actions on the route.
-   - Any changes to the `Route`
+   - Any changes to the `Route` or `Controller` as they exist today
    
 This'll explore how controllers are used and how viable alternatives can be for migrating away from controllers so that they may eventually be removed.
 
 ## Motivation
 
- - Many were excited about [Routable Components](https://github.com/ef4/rfcs/blob/routeable-components/active/0000-routeable-components.md).
+ - Many were excited about [Routable Components](https://github.com/ef4/rfcs/blob/routeable-components/active/0000-routeable-components.md) (This RFC is _not_ "Routable Components").
  - Many agree that controllers are awkward, and should be considered for removal (as mentioned in many #EmberJS2019 Roadmap blogposts).
  - There is, without a doubt, some confusion with respect to controllers:
    - should actions live on controllers?
@@ -48,85 +48,19 @@ All of this behavior can be implemented elsewhere, and most of what controllers 
 
 ### 1. Query Params
 
-In this other RFC that [proposes adding a @queryParam decorator and service to manage query params](https://github.com/emberjs/rfcs/pull/380), query-param management would be pulled out of the controller entirely.
-
-Presently, query params live on controllers, which are singletons, allowing query param values to be set, and unset, when navigation to and away from a route. Query params can have the same behavior implemented on a service, which can be tested out on an addon, [ember-query-params-service](https://github.com/NullVoxPopuli/ember-query-params-service).
-
-Behavior like `replaceState` and `refreshModel: true`, would still live on the route as that behavior is more of a route concern than it is a concern of the value of the query params.
-
-- **Mapping a Query Param to state**
-
-  Old: query params _must_ come from a controller.
-  ```ts
-  import Controller from '@ember/controller';
-
-  export default class SomeController extends Controller {
-    queryParams = ['foo'];
-    foo = null;
-  }
-  ```
-  ```hbs
-  <SomeComponent @foo={{this.foo}} />
-  ```
-
-  New: query params can be used anywhere via dependency injection.
-  ```ts
-  import Component from "@glimmer/component";
-  import { queryParam } from "ember-query-params-service";
-
-  export default class SomeComponent extends Component {
-    @queryParam foo;
-
-    addToFoo() {
-      this.foo = (this.foo || 0) + 1;
-    }
-  }
-  ```
-
-- **Mapping a Query Param to a state of a different name**
-
-  Old: controller must exist to define query params
-  ```ts
-  import Controller from '@ember/controller';
-
-  export default class SomeController extends Controller {
-    queryParams = {foo: 'bar' };
-    bar = null;
-  }
-  ```
-
-  New: No controller exists
-  ```ts
-  import Component from "@glimmer/component";
-  import { queryParam } from "ember-query-params-service";
-
-  export default class SomeComponent extends Component {
-    @queryParam('foo') bar;
-
-    addToFoo() {
-      this.bar = (this.bar || 0) + 1;
-    }
-  }
-  ```
+See this RFC for [adding query params to the router service](https://github.com/emberjs/rfcs/pull/380), query-param management would be pulled out of the controller entirely.
 
 ### 2. Passing State and Handling Actions to/from the Template
 
 In order to fully eliminate the controller, there _must_ be a way to handle computed properties or state along with responding to actions from the template. Components can cover this already, and would be an intuitive way to interact with the view and behavioral layer.
 
 Currently, controllers + templates must be used in the app by having:
- - for classic apps:
-   - `app/controllers/{my-route-name/or-path}.js`
-   - `app/templates/{my-route-name/or-path}.hbs`
- - for pods apps:
-   - `{pod-namespace}/{my-route-name/or-path}/controller.js`
-   - `{pod-namespace}/{my-route-name/or-path}/template.hbs`
- - for MU or (future) MU-inspired layouts:
-   - `src/ui/routes/{my-route-name/or-path}/controller.js`
-   - `src/ui/routes/{my-route-name/or-path}/template.hbs`
+  - `app/controllers/{my-route-name/or-path}.js`
+  - `app/templates/{my-route-name/or-path}.hbs`
 
-The proposed new behavior would be that when a controller is not present, and there is only a template, that template becomes a template-only/stateless component. The template-only/stateless component has a number of advantages as it forces the developer to separate UI layout and to think about the semantic breakdown of the template. Here are some existing examples of template-only/stateless route templates in an app made today:
+The proposed new behavior would be that when a controller is not present, and there is only a route template, that template becomes a template-only/stateless component (no backing js file). The template-only/stateless component has a number of advantages as it forces the developer to separate UI layout and to think about the semantic breakdown of the template. Here are some existing examples of template-only/stateless route templates in an app made today:
 
-from emberclear.io
+from emberclear.io (using module unification)
  - [routes/application/template.hbs](https://github.com/NullVoxPopuli/emberclear/blob/master/packages/frontend/src/ui/routes/application/template.hbs)
  - [routes/contacts/template.hbs](https://github.com/NullVoxPopuli/emberclear/blob/master/packages/frontend/src/ui/routes/contacts/template.hbs)
 
@@ -144,28 +78,23 @@ Sample:
 </section>
 ```
 
-Everything is encapsulated in components so that intent is clear as to which part of the template is responsible for what behavior. However, this is only the simplest use case, where all derived state is handled by the components rendered by the existing template.
+Everything is encapsulated in components so that intent is clear as to which part of the template is responsible for what behavior. 
 
-For the case where we _need_ derived state at the root of our route's tree.
- - for classic apps:
-   - `app/components/{my-route-name/or-path}.js`
-   - `app/templates/{my-route-name/or-path}.hbs`
- - for pods apps:
-   - `{pod-namespace}/{my-route-name/or-path}/component.js`
-   - `{pod-namespace}/{my-route-name/or-path}/template.hbs`
- - for classic apps after [RFC #481](https://github.com/emberjs/rfcs/pull/481) (Colocated Component and Template files)
-   - `app/components/{my-route-name/or-path}.js`
-   - `app/components/{my-route-name/or-path}.hbs`
- - for MU or (future) MU-inspired layouts:
-   - `src/ui/routes/{my-route-name/or-path}/component.js`
-   - `src/ui/routes/{my-route-name/or-path}/template.hbs`
+The reason for explicitly calling out this behavior of template-only components at the route-level is because the template already exists there today, and if we make it a component, that makes future transitions easier.  Making the route template a component _only_ means that instead of only existing a `this.model`, there would be a `@model` argument.
 
-For current classic apps, this may feel a bit like resolution magic, but it's just the same lookup rule as for controllers, just in a different folder. This changes makes more contextual sense with pods and module unification (or some future derivitive of module unification) where the backing context to the template is co-located alongside the template.
+However, this is only the simplest use case, where all derived state is handled by the components rendered by the existing route template.
+
+For the case where we _need_ derived state at the root of our route's tree, we _only_ add a resolution to a component of the same name..
+  - `app/components/{my-route-name/or-path}.js`
+  - `app/templates/{my-route-name/or-path}.hbs`
+ 
+It's just the same lookup rule as for controllers, just in a different folder. If people end up having a ton of top-level state-based components per route, their component sturcture will end up matching their routing structure. This is fine, as it means that certain components are _only_ used for particular routes, which may make the unification of to a new layout more straight forward.
 
 These are not "routable components", but components that are invoked after the `model` hook resolves in the route, recieving only a single property `@model`.
 
  - Before: with controllers
     ```ts
+    // app/routes/some.js
     import Route from '@ember/routing/route';
 
     export default class SomeRoute extends Route {
@@ -177,6 +106,7 @@ These are not "routable components", but components that are invoked after the `
     }
     ```
     ```ts
+    // app/controllers/some.js
     import Controller from '@ember/controller';
 
     export default class SomeController extends Controller {
@@ -188,6 +118,7 @@ These are not "routable components", but components that are invoked after the `
     }
     ```
     ```hbs
+    {{!-- app/templates/some.hbs --}}
     {{#each this.onlineContacts as |contact|}}
       {{contact.name}}
     {{/each}}
@@ -195,6 +126,7 @@ These are not "routable components", but components that are invoked after the `
 
  - After: the controller is swapped for a component 
      ```ts
+    // app/routes/some.js
     import Route from '@ember/routing/route';
 
     export default class SomeRoute extends Route {
@@ -206,6 +138,7 @@ These are not "routable components", but components that are invoked after the `
     }
     ```
     ```ts
+    // app/components/some.js
     import Component from "@glimmer/component";
 
     export default class SomeComponent extends Component {
@@ -219,6 +152,7 @@ These are not "routable components", but components that are invoked after the `
     }
     ```
     ```hbs
+    // app/templates/components/some.js
     {{#each this.onlineContacts as |contact|}}
       {{contact.name}}
     {{/each}}
@@ -239,24 +173,24 @@ Given a visit to `/posts`
   - this component is invoked _as if_ `template:posts` was defined as `<Posts @model={{this.model}} />`
 - default to `{{outlet}}` or `{{yield}}` for sub routes
 
-**NOTE** as a side effect of treating the default template in this manner, it would become a requirement, in order to maintain cohesion, that we treat loading and error templates similar to the above lookup rules.
+**NOTE** as a side effect of treating the default template in this manner, it would become a requirement, in order to maintain cohesion, that we treat loading and error templates similar to the above lookup rules. This means that loading and error components can be used with service injections, and other state.
 
 ## How we teach this
 
 Since this is a new programming paradigm, we'll want to update the guides and tutorials to reflect a new "happy path".
  - Remove most mentions of controllers in the guides / tutorials. maybe just leave the bare informational "what controllers are", at least until a deprecation can be figured out.
- - Swap out the controllers page in the guides for a page that shows examples of how a component is chosen to be the entry point / outlet of the route. Additionally on this page, it should mention that there can be an optional template (like we have today) that receives the `@model` argument that takes priority over any component with the same name as the route and without any template at all, it's just an `{{outlet}}` for subroutes. 
+ 
+ - Swap out the controllers page in the guides for a page that shows examples of how a component is chosen to be the entry point / outlet of the route. Additionally on this page, it should mention that there is a default template (like we have today) that receives the `@model` argument that takes priority over any component with the same name as the route and without any template at all, it's just an `{{outlet}}` for subroutes.  
  
     The controllers page for octane has already moved under routing, so talking about just rendering behavior should make things feel simpler.
     
  - There is already a separate guides page for query params. Query params should remain on their own page, but just be updated to use the `@queryParam` decorator, as described in [RFC 380](https://github.com/emberjs/rfcs/pull/380)
 
-
 ## Drawbacks
 
-It's very important that every use case for controllers today _can_ be implemented using the aforementioned techniques. If people are willing to share their controller scenarios, we can provide a library of examples of translation so that others may migrate more quickly.
+It's important that common use cases for controllers today _can_ be implemented using the aforementioned techniques. If people are willing to share their controller scenarios, we can provide a library of examples of translation so that others may migrate more quickly.
 
-It may be possible to write a codemod to help with the migration as the biggest difference, at leaste in this RFC's initial samples is that `model` is accessed on `this.args` instead of just `this`.  The tool and maybe even a linter rule could look for pre-existing components that have any of the same names as routes, and assert that they handle the `model` argument.
+It may be possible to write a codemod to help with the migration as the biggest difference, at least in this RFC's initial samples is that `model` is accessed on `this.args` instead of just `this`.  The tool and maybe even a linter rule could look for pre-existing components that have any of the same names as routes, and assert that they handle the `model` argument.
 
 
 ## Unresolved questions
--- a/text/0000-controller-migration-path.md
+++ b/text/0000-controller-migration-path.md
@@ -29,6 +29,13 @@ This'll explore how controllers are used and how viable alternatives can be for 
 
 ## Detailed design
 
+### Query Params
+
+In this other RFC that [proposes adding a @queryParam decorator and service to manage query params](https://github.com/emberjs/rfcs/pull/380), query-param management would be pulled out of the controller entirely.
+
+### Error and Loading States
+
+
 > This is the bulk of the RFC.
 
 > Explain the design in enough detail for somebody
--- a/text/0000-controller-migration-path.md
+++ b/text/0000-controller-migration-path.md
@@ -239,6 +239,17 @@ Given a visit to `/posts`
   - this component is invoked _as if_ `template:posts` was defined as `<Posts @model={{this.model}} />`
 - default to `{{outlet}}` or `{{yield}}` for sub routes
 
+## How we teach this
+
+Since this is a new programming paradigm, we'll want to update the guides and tutorials to reflect a new "happy path".
+ - Remove most mentions of controllers in the guides / tutorials. maybe just leave the bare informational "what controllers are", at least until a deprecation can be figured out.
+ - Swap out the controllers page in the guides for a page that shows examples of how a component is chosen to be the entry point / outlet of the route. Additionally on this page, it should mention that there can be an optional template (like we have today) that receives the `@model` argument that takes priority over any component with the same name as the route and without any template at all, it's just an `{{outlet}}` for subroutes. 
+ 
+    The controllers page for octane has already moved under routing, so talking about just rendering behavior should make things feel simpler.
+    
+ - There is already a separate guides page for query params. Query params should remain on their own page, but just be updated to use the `@queryParam` decorator, as described in [RFC 380](https://github.com/emberjs/rfcs/pull/380)
+
+
 ## Drawbacks
 
 It's very important that every use case for controllers today _can_ be implemented using the aforementioned techniques. If people are willing to share their controller scenarios, we can provide a library of examples of translation so that others may migrate more quickly.
--- a/text/0000-controller-migration-path.md
+++ b/text/0000-controller-migration-path.md
@@ -8,11 +8,16 @@
 ## Summary
 
  - Provide alternative APIs that encompass existing behavior currently implemented by controllers.
+   1. query params
+   2. passing state and handling actions to/from the template
+
  - Out of scope:
    - Deprecating and eliminating controllers -- but the goal of this RFC is to provide a path to remove controllers from Ember altogether. 
+
  - Not Proposing
    - Adding the ability to access more properties than `model` from the route template
    - Adding the ability to invoke actions on the route.
+   - Any changes to the `Route`
    
 This'll explore how controllers are used and how viable alternatives can be for migrating away from controllers so that they may eventually be removed.
 
@@ -31,42 +36,17 @@ This'll explore how controllers are used and how viable alternatives can be for 
 
 ## Detailed design
 
-### Rendering
+First, what would motivate someone to use a Controller today? Borrowing from [this blog post](https://medium.com/ember-ish/what-are-controllers-used-for-in-ember-js-cba712bf3b3e) by [Jen Weber](https://twitter.com/jwwweber): 
 
-The templates used for routes should remain the same. Any context that would go in controllers can be split out to components. This has the advantage of components defining the semantic intent behind what they are rendering, such as in this example application template: [routes/application/template.hbs from emberclear](https://github.com/NullVoxPopuli/emberclear/blob/master/packages/frontend/src/ui/routes/application/template.hbs)
-```hbs
-<TopNav />
+1. You need to use query parameters in the URL that the user sees (such as filters that you use for display or that you need to send with back end requests)  
 
-<NotificationContainer @position='top-right' />
+2. There are actions or variables that you want to share with child components  
 
-<OffCanvasContainer class='no-overflow'>
-  <UpdateChecker />
+3. (2a) You have a computed property that depends on the results of the model hook, such as filtering a list
 
-  {{outlet}}
+All of this behavior can be implemented elsewhere, and most of what controllers _can_ do is already more naturally covered by components.
 
-</OffCanvasContainer>
-
-<Modals />
-<AppShellRemover />
-```
-
-This route-level template only defines the layout. Everything is encapsulated in components so that intent is clear as to which part of the template is responsible for what behavior.
-
-In some cases, the route template may be as small as only having some wrapping styles ([routes/contacts/template.hbs from emberclear](https://github.com/NullVoxPopuli/emberclear/blob/master/packages/frontend/src/ui/routes/contacts/template.hbs))
-```hbs
-<section class='section'>
-  <div class='container'>
-    <Header @contacts={{this.model.contacts}} />
-
-    <div class='content'>
-      <ContactTable />
-    </div>
-  </div>
-</section>
-```
-Here, model is the only property that the template has access to. Because this template has a more narrow and specific purpose, it should be referred to explicitly as a "route template".
-
-### Query Params
+### 1. Query Params
 
 In this other RFC that [proposes adding a @queryParam decorator and service to manage query params](https://github.com/emberjs/rfcs/pull/380), query-param management would be pulled out of the controller entirely.
 
@@ -105,17 +85,17 @@ Behavior like `replaceState` and `refreshModel: true`, would still live on the r
 
 - **Mapping a Query Param to a state of a different name**
 
-  Old:
+  Old: controller must exist to define query params
   ```ts
   import Controller from '@ember/controller';
 
   export default class SomeController extends Controller {
-    queryParams = ['foo'];
-    foo = null;
+    queryParams = {foo: 'bar' };
+    bar = null;
   }
   ```
 
-  New:
+  New: No controller exists
   ```ts
   import Component from "@glimmer/component";
   import { queryParam } from "ember-query-params-service";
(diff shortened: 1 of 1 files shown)


lifeart:
In current queryParams implementation qp-values binded to model, it's possible to have different states for one page for different models, how we can handle it using new approach?a

NullVoxPopuli:
[RFC 380](https://github.com/emberjs/rfcs/pull/380) proposes a way to store query params per sub-route :)
certainly more tests can be added, but I think this covers the basics: https://github.com/NullVoxPopuli/ember-query-params-service/blob/master/tests/acceptance/navigation-test.ts#L56

lifeart:
@NullVoxPopuli my comment about this behavour https://ember-twiddle.com/b9e78f44f0a63a1a5b0f75ca28b9ebef - try type some search and change models

NullVoxPopuli:
Yup. that works with [RFC 380](https://github.com/emberjs/rfcs/pull/380) :)

ef4:
This is on a good track but needs more detail. Can you expand upon the lookup priority rules?

For example, I think if `app/templates/your-route.hbs` or `app/controllers/your-route.js` exist, they probably need to keep doing what they do now. But if they don't, we can add the new behavior to use `app/components/your-route.js` and/or `app/templates/your-route.hbs`.

That would be safe as long as we gave you an upgrade step that generated explicit template files for any of your classic routes that were previously relying on `{{outlet}}` being auto-generated for them.

I also think we end up in an inconsistent state if we don't also allow `app/templates/slow-model-loading.hbs` to become `app/components/templates/slow-model-loading.hbs`. As far as I can tell that doesn't require any design other than saying what parameters it will see (maybe `@error`).

jrjohnson:
The title doesn't really match the RFC @NullVoxPopuli maybe `Alternative Controller APIs`? Or something less final since you're not actually proposing the removal of controllers

NullVoxPopuli:
@jrjohnson That's fair, removing controllers would require a deprecation RFC, which this is not

jasonmit:
Should the RFC recommend alternative approaches to storing long lived state?  I imagine the rouleable components will not be singleton's - correct?

NullVoxPopuli:
@ef4 I've added some details around look priority :)

> I also think we end up in an inconsistent state if we don't also allow...

I think addressing loading/error states should happen in a separate RFC. As far as I understand it, error and loading stuff is _just_ templates. I'd like to keep this RFC focused on controllers, unless the community or the core teams say it absolutely needs to be done. idk :)

@jasonmit correct. any component used for routing should be considered ephemeral. Services should be used for long-term state. :)

Panman8201:
FWIW, I feel like the move away from controllers can be aligned with router refactors. In that, once folks move to (potentially) new route semantics, controllers will no longer be used/loaded. Instead, some form of a component will become the route template, such as a template-only component or maybe the route defines which component to render, etc. More thoughts coming as soon as I get to my ember2019 post.

stefanpenner:
> @jasonmit correct. any component used for routing should be considered ephemeral. Services should be used for long-term state. :)

 bingo

jasonmit:
I imagined the store(s) would be backed by a service but will there be a recommended approach regarding strategy or common API that can be shared?  Or is this an opportunity for Ember as a the framework to step out of the way and lean more on community solutions?

NullVoxPopuli:
> I imagined the store(s) would be backed by a service but will there be a recommended approach regarding strategy or common API that can be shared? 

service injections cover this :)

```ts
import { task } from 'ember-concurrency';

  @service store;
  @service toast;

  @task(function*() {
    const record = this.store.createRecord('post', { ...whateverAttributes });

    try {
      yield record.save();
      toast.success('yay!');
    } catch (e) {
      toast.error(e);
    }
  }); create();
  }
}
```

jasonmit:
I'm aware of the service inject API but feel it doesn't quite layer over what we have now.  Currently there's a 1:1 mapping of route backed state vs what sounds like a recommendation for arbitrary service(s) used to hang state from.  Uniformity across codebases breaks down here and may be worth documenting recommended paths in the migration docs.  Perhaps something for the Ember learning team.

NullVoxPopuli:
>  what sounds like a recommendation for arbitrary service(s)

hmm, maybe I misunderstand? can you provide some concrete examples of how you use controllers today, that components wouldn't be able to handle?

>  Uniformity across codebases breaks down here  

can you expand on this?

willviles:
I understand that it's more a concern that it's a move from a highly conventional approach (FooRoute => FooController), to a more ad-hoc approach (FooRoute => ???Service/s).

Can someone explain why routes themselves can't be singletons and store their own long-lived state? Like a service that just automatically has hooks invoked on route changes?

jasonmit:
> hmm, maybe I misunderstand? can you provide some concrete examples of how you use controllers today, that components wouldn't be able to handle?

Any state that isn't queryparam state.  One example could be an email client that keeps draft state for emails in memory over the lifetime of the app instance.  Today, you might throw it in a controller or a service.  If you're in a large codebase then there may be many instances like this and you'd end up with really small, oddly shaped, services when migrating an app over to a controller-less world.

So the question I've been failing to pose (sorry, limited by typing on my phone) is are we recommending moving everyone to services for this state or is there a common API that can be shared for mapping route backed state without one off services where a dedicated services may be a bit too heavy.

NullVoxPopuli:
> Can someone explain why routes themselves can't be singletons and store their own long-lived state?

I think because services already fill that space.

> are we recommending moving everyone to services for this state 

I am :)

> API that can be shared for mapping route backed state without one off services where a dedicated services may be a bit too heavy.

I agree

@willviles and @jasonmit 
I guess it depends on how common your long-lived controller state is.  Personally, I haven't run in any such scenario, but am more than happy to read about other's experiences :)

like, for the example of having a draft-state for emails or something, I'd `ember g service emails/email-builder-form` and then inject that into however many components need it.

I think the act of injecting a service for this use case keeps cohesion with the rest of the framework :)

flashios09:
if i understood correctly
for a `posts` route(http://localhosts:4200/posts)

**BEFORE**, we have:
* posts/route.js(fetch data, init the model)
* posts/controller.js(decorate the model)
* posts/template.hbs

**AFTER**, we will have:
* posts/route.js(same behavior, fetch data, init the model)
* posts/component.js(get the `model` from the route as an **arg**, then decorate it)
* posts/template.hbs

this is like renaming `controller` file with `component` ?

NullVoxPopuli:
@flashios09 almost

If using a component, the after would be:

- posts/route.js (same behavior, fetch data, init the model)
- app/components/posts.js (get the model from the route as an arg, `@model`)
- app/templates/components/posts.hbs

flashios09:
@NullVoxPopuli 
> * app/templates/components/posts.hbs

so we will not have `app/templates/posts.hbs` any more ?
the whole `view` layer will be replaced by `components` ?

NullVoxPopuli:
> so we will not have app/templates/posts.hbs any more?

you have the option to not have, but it may exist as a template-only component.
It may be _easier_ for it to remain as-is, but I think keeping it consistent with component invocation would be good for adding a `@model` arg, as well as keeping a `this.model` like it has today. That way any existing templates would still work.

>  the whole view layer will be replaced by components?

yes ;)

sandstrom:
Your RFC states:

> [I do not propose] Adding the ability to invoke actions on the route

So how would one invoke actions on the route, should this RFC become reality? Or do you propose that all actions on routes are moved elsewhere?

NullVoxPopuli:
> So how would one invoke actions on the route, should this RFC become reality? Or do you propose that all actions on routes are moved elsewhere? 

imo, it's most maintainable to have routes strictly be fore gating access and loading minimally required data. actions should either be in components or services :)

sandstrom:
@NullVoxPopuli Alright, thanks for clarifying!

- Perhaps this RFC could mention that "the goal is to eventually remove actions from routes".
- If we do want to remove actions from routes, I think something like 'routable components' (where these actions could live) would make more sense that what's proposed in this RFC. Otherwise the migration burden will be larger.
- If we want to go in the direction of only passing the result of the model hook to the view-layer (controllers/templates), wouldn't another name be better? Perhaps `state`, `package` or `payload`? Right now there is some ambiguity in that `model` may be an instance of an Ember Data Model, or it may be an array, or a hash. By using another name we'd get rid of that ambiguity.

I'm all for getting rid of controllers, but I think something like routable components would make more sense than simply axing controllers without thinking how related things (routes, etc) should change to accommodate. Basically, I understand the allure of writing up a simple RFC that changes as little as possible, but in complex systems, if you change one thing, there is often a need to adjust adjacent pieces to make the whole system work well after the change. 

In its current state it feels like this RFC want's to rip out some stuff that everyone wants to get rid of, without providing comprehensive answers to how other parts of the system should change to accommodate.

NullVoxPopuli:
> Perhaps this RFC could mention that "the goal is to eventually remove actions from routes".

I don't think I can make that claim, at least not in _this_ RFC.
Actions for loading and error handling will still be useful even if this RFC is implemented.

>  I think something like 'routable components'

Not to be finicky here, but these are components-that-happen-to-be-used-for-the-route-entry-point :)
The term "routable components" has so much baggage. haha

>  (where these actions could live) would make more sense than what's proposed in this RFC. 

I think the RFC does mention that actions that otherwise would have been on the controller should go in the targeted component. I'll double check though! :)

> wouldn't another name be better? Perhaps state, package or payload? Right now there is some ambiguity in that model may be an instance of an Ember Data Model, or it may be an array, or a hash. By using another name we'd get rid of that ambiguity.

To reduce the amount of changes this RFC introduces, I want to keep the naming of that the same. 
fwiw, imo, the only ambiguity is when ember-data is present. But, in an ideal world, I'd like to propose renaming all 3 of the route lifecycle hooks. <3

>  it may be an array, or a hash.

I think community conventions should either recommend that the model result be a hash, or maybe eventually I'll RFC a deprecation for non-hash model values so that people can be forced (not sure if this is a good strategy) to have semantically meaningful model values? idk

>  without providing comprehensive answers to how other parts of the system should change to accommodate.

This RFC recommends resolving components of the same name when no controller or route template is present :)

pzuraq:
So, I think there may be some confusion around the fate of the original Routable Components RFC, and that this may be limiting the design space here. This proposal is solid, and definitely a way we could approach removing controllers from Ember in the short term with minimal changes, but I think it would still end up wanting something a bit more cohesive in the long run to completely replace controllers.

When @wycats [closed the original RFC](https://github.com/emberjs/rfcs/pull/38#issuecomment-355800759) a little over a year ago, he detailed the reasons for doing so, and it really was mostly about the fact that the RFC was seriously outdated (it had been open for 3 years) and that there were other prerequisites for nailing down the design of routable components, such as implementing Glimmer components. The general idea was solid, but the RFC was a bit premature.

In the discussions I've had with core team members, I do believe the core team is generally in favor of a much simplified "Routable Components Lite" design. In fact, @chadhietala proposed such a design in his [Ember.js 2019 roadmap post](https://gist.github.com/chadhietala/50b977a7d3476069892d351c65af418c#routing) which I think would be a great starting point.

That said, if it turns out that it would take too long to implement this new design, I do think the path laid out in this RFC could be a good stepping stone. It's clear from the many 2019 roadmap blogposts that the community wants to get rid of controllers ASAP, so we shouldn't block on the perfect design when we have a good enough path forward

ef4:
> I think addressing loading/error states should happen in a separate RFC. As far as I understand it, error and loading stuff is just templates.

Today they are backed by the controller. So if we don't address them, we haven't really done the work to design an alternative for controllers.

And IMO it makes this RFC simpler instead of more complex, but the whole thing boils down to a single rule: **any time you looked up route template X and didn't find it, look up component X too and if it exists, use that**.

If we don't include the error and loading templates, the rule is actually more complicated, because there would be caveats for "except loading templates, which act differently, and error templates, which act differently".

To me the whole point of this RFC is making it possible to have 100% of the templates in your application be Just Components. So nobody needs to learn about any other kind of template. The introduction of glimmer components is what makes this tenable. Classic components didn't work for this because they have a wrapping div, whereas route template and glimmer components do not.

NullVoxPopuli:
> Today they are backed by the controller. 

I did not know that. Good to know :)
Then yeah, let's apply the same lookup logic to error/loading templates then :)

ef4:
> Can someone explain why routes themselves can't be singletons and store their own long-lived state? 

Today they are singletons. But I have been trying to get routes to have *less* state (ideally none) for years. We have basically succeeded as of RFC #418.

Statefulness in routes defeats the original point of having routes, which is that they are supposed to be mapping between state in the URL and state in the application. To succeed in that mission they're much easier to understand if they are stateless transformations between those two stateful things.

State in the Route itself is not reflected in the URL and therefore easy to lose and harder to test and reason about.

If people want long-lived state that is scoped per route, they can say `this.thingService.things[routeName] = rememberThisThing`.

stefanpenner:
> Today they are singletons. But I have been trying to get routes to have less state (ideally none) for years. We have basically succeeded as of RFC #418.

I can get behind this 100%.

NullVoxPopuli:
> If people want long-lived state that is scoped per route, they can say this.thingService.things[routeName] = rememberThisThing.

this is how the query-params RFC is _currently_ implemented in the ember-query-params-service addon :)

basicBrogrammer:
Not sure if we can pull ideas from this post https://emberway.io/skeleton-screen-loading-in-ember-js-2f7ac2384d63 but it feels like it might have some parallels 

After working with react a tiny amount, controllers started to click for me. They are very similar to the "container" component pattern in react. 

Seems like the 2 major milestones for this RFC would be what do we do with actions and what to do about the loading state. 

Happy to spike with you on it! (first time joining in on the RFC convos so bare with me)

webark:
Perhaps we could move to utilizing the “loading” and “error” actions on the route rather then relying on the implicit “-loading” intermediary routes?

NullVoxPopuli:
@webark 

> Perhaps we could move to utilizing the “loading” and “error” actions on the route

those actions are on the route: https://guides.emberjs.com/release/routing/loading-and-error-substates/#toc_the-loading-event

>  rather then relying on the implicit “-loading” intermediary routes?

I have a plan for a follow RFC to make some changes to loading and error handling in general to make it more explicit

basicBrogrammer:
Not 💯 familiar with the loading/error states ... would it make sense to offload these to the router service ? Then in the routed component you could set the loading state `router.loaded()` or `router.failed()` after loading data 🤔

knownasilya:
A similar approach for long lived state could utilize an addon like https://github.com/stefanpenner/ember-state-services, and use the route name as the key that stores the state.

andrewcallahan:
This is really well thought out. Many of us have complained about Controllers, but you've actually _done_ something about it!

I have to agree with @flashios09 though, that this approach feels like simply replacing Controller js files with Component js files.

Assuming Module Unification isn't coming before this, I'm hesitant to add hierarchy to the `/components` folder and to in any way tie the `/components` folder structure to the URL structure. If Module Unification lands before this, I'm also hesitant to have Component .js and .hbs files in the new `/routes` folder. That seems confusing to me and is begging for something like Routable Components.

Speaking of Routable Components, I'm pretty excited to hear @pzuraq mention that the Core Team is interested in taking another swing at them! If that is indeed the case, then whatever Ember does to remove Controllers should have this endgame in sight. 

The approach you outline seems to require a fair amount of work and imposes additional architectural decisions that _could_ need to be undone depending on how Routable Components ultimately work.

If Ember is really going to get rid of Controllers, and move to Routable Components, here is how I envision the transition happening with the least amount of work to merge an RFC, the easiest transition path for developers, and maximum architectural flexibility for the Core Team.

## Phase 1: Remove Controllers De Facto
1. The Core Team makes a commitment to remove Controllers and replace them with Routable Components when they announce the new Roadmap.
2. In the next Ember release, everything besides Query Params on Controllers (state, actions, etc) get deprecation warnings.
3. Developers then begin to move over most of the logic away from Controllers. This would entail rendering a single Component in any stateful Template and involve mostly copying and pasting.
4. Query Params service is launched.
5. Apps that used Query Params on their Controllers move to use the new Query Params Service.

## Phase 2: Remove Controllers De Jure
Hopefully, many developers would have done this work by now, and have no more Controllers in their apps. Now Controllers can officially be removed and the deprecation warnings now become actual errors. For apps that have been keeping up, it's a non-event upgrade. For apps that haven't, this is a kind of a line in the sand/wake up call that it's time they need to upgrade. What we are left with now is a world with no Controllers, stateless Templates, and many apps that use Templates simply to render a single Component. In other words, we are perfectly primed for Routable Components!

## Phase 3: Routable Components
It took a bit of time to get here, which is good in two aspects. First, it gives developers time to update their apps without causing too much churn. Second, it gives the Core Team time to get the interface for Routable Components right. Routable Components would be a big change in Ember and it's vital to avoid what Sandy Metz would call the "_wrong abstraction_". I have no clue as to the final design of Routable Components, but my suspicion is that we would get to the best design if we give the Core Team time and imposed no more constraints during the transition phase away from Controllers.

This outline is a bit handwavy, and I'm sure there are details I'm overlooking, etc - so please consider this a thought exercise to build and iterate upon!

NullVoxPopuli:
@andrewcallahan 

> I have to agree with @flashios09 though, that this approach feels like simply replacing Controller js files with Component js files.

This is not true -- or at least not what I'm trying to convey. In my opinion, I _believe_ the happiest path is to use the template-only scenario. Which is just what we have today, but without a backing controller.
Additionally, this is how I see it working for each of the types of layouts we have:

** classic **

 - My personal happy path
    - app/routes/my-route.js
    - app/templates/my-route.hbs

 - For the case where you really do have a single root entry point
    - app/routes/my-route.js
    - app/components/my-route.js
    - app/templates/components/my-route.hbs

** MU or MU-inspired, (eventually) **
    - src/ui/routes/my-route/route.js
    - src/ui/routes/my-route/-components/{component.js, template.hbs} (exact path TBD, but something co-located like this -- could be `index`, `entry`, `render`, whatever. we aren't at that point yet)

> , I'm hesitant to add hierarchy to the /components folder and to in any way tie the /components folder structure to the URL structure. 

oh, I def wouldn't. I would most certainly encourage the use of the template-only rendering context, as I have shown in some of the rendering examples from emberclear 😉 

> If Module Unification lands before this, I'm also hesitant to have Component .js and .hbs files in the new /routes folder. That seems confusing to me and is begging for something like Routable Components.

yeah, I wouldn't want that either. components are core to frontend, and it'd be weird to give them any sort of special treatment apart from other components (aside from resolver lookup conventions in the classic world)
maybe what would be better for an eventual convention in a MU-like app is this:
```ts
// for the path `my-route`
// this file exists at `src/ui/routes/my-route/index.js`
import MyComponent from './-components/my-component';

// named export must match name of route
export class MyRoute extends Route {
  async model() { ... }
}

// the default export is the rendered thing that get's passed the `@model` arg.
export default MyComponent;
```

> The approach you outline seems to require a fair amount of work and imposes additional architectural decisions that could need to be undone depending on how Routable Components ultimately work.

can you explain these thoughts more? (cause I either don't see it, or don't understand 😢  )

> If Ember is really going to get rid of Controllers, and move to Routable Components, 

can we not use the term routable components? this term has some much baggage and frustrating with it. 🤣 

> Phase 1, Phase 2, Phase 3

These phases feel a little backwards to me. In order to be totally backward compatible, we'll need to do something like the following:

- query param alternative should happen first as it's a pre-req, and already a huge pain point in existing apps. So the query params RFC, https://github.com/emberjs/rfcs/pull/380, can help people out long before anything happens with controllers :)
- we need to first have an alternative, see how the community adopts it, and if it really is going to work.
   - this is where components via this RFC or some derivative or something completely different comes in to play.
   - documentation would need to be updated to recommend the new happy path. the guides and api documentation either minimizes its mention of controllers or has a side-note somewhere mentioning the historical baggage that they will be at this point
   - this also means that we'll probably have a long period of time where people are migrating and have both controllers and whatever the controller-alternative is. Migrating takes a while :)
- next we can decide to finally deprecate controllers altogether. (you import or instantiate a controller, you get deprecation warning)
   - hopefully migration speed picks up here, but we'll need a deprecation RFC to make this official
- finally we can remove controllers as part of a major version bump as it'd be breaking change (whether that be ember 4 or 5 or further out is totally up in the air!) 😄

andrewcallahan:
Thanks for the quick reply @NullVoxPopuli! Let's go one-by-one with the issues you brought up to see if we can converge in our understanding ;)

### 1. Does this approach simply replace Controller js files with Component js files?

Here is what I understand you are proposing
>** Classic **
>app/routes/my-route.js
>app/components/my-route.js
>app/templates/components/my-route.hbs

>** MU **
>src/ui/routes/my-route/route.js
>src/ui/routes/my-route/-components/component.js
>src/ui/routes/my-route/-components/template.hbs

For the classic style, it looks like `app/controllers/my-route.js` is just replaced with `app/components/my-route.js`. I get that it's a bit more complicated than that, but fundamentally it's the same thing - some resolver looks through the files and hooks up a .hbs file in the `/templates` folder with a .js file in another folder (this time `/components` instead of `/controllers`). That is something I'd like to move away from. I have comments on the MU style later on.

### 2. Should Templates have .js files at all?
>In my opinion, I believe the happiest path is to use the template-only scenario. Which is just what we have today, but without a backing controller.

That is what I think would be optimal too. What I don't understand is why you hook up the Template file to a .js file in the `/components` folder (Classic) or have a Component .js file in the new `/routes` folder (MU) at all. Both of these would lead to stateful Templates, and not a "_template-only scenario_".

### 3. Should the `/components` folder have hierarchy tied to the URL
I'm confused here. On the one hand, you say "_oh, I def wouldn't_", but in the classic scenario you laid out `app/components/my-route.js` is tied to the template `app/templates/my-route.hbs`. For nested routes, I would presume this would mean `app/components/rentals/edit.js` would similarly be tied to the  `app/templates/rentals/edit.hbs` file. If so, the `/components` folder structure would be directly tied to the route structure - just like the `/templates` folder is today.

### 4. Should Component .js and .hbs files be in the new `/routes` folder?
I'm again confused here. On the one hand, you say "_yeah, I wouldn't want that either_", but in the MU scenario you have the Component .js and .hbs files in it:

>- src/ui/routes/my-route/-components/component.js
>- src/ui/routes/my-route/-components/template.hbs

Now, your suggestion (below) of specifying a Component from the Route file is interesting and would avoid the pitfall of having two places for Component .js files. I could see something like this being one possible way to get Routable Components but am not sure it's needed before then.
>maybe what would be better for an eventual convention in a MU-like app is this:

>// for the path `my-route`
>// this file exists at `src/ui/routes/my-route/index.js`
>import Route from '@ember/routing/route';
>import MyComponent from './-components/my-component';
>
>// named export must match name of route
>export class MyRoute extends Route {
>  async model() { ... }
>}
>
>// the default export is the rendered thing that get's passed the `@model` arg.
>export default MyComponent;

### 5. Does this RFC, as it stands, take a fair amount of work?
I'm not sure exactly, but I would pose that question back to you or Core Team members lingering around here. Setting aside the work with the Query Params Service, it seems like more work to do *anything* other than just put in deprecation warnings for Controllers, which is what I'm proposing to do until Routable Components land.

### 6. Would this approach make additional architectural decisions before Routable Components land?
It seems so to me. This may not be a bad thing if they are the right ones. However, if for instance  Components are specified in a Route in your above example, that is a decision that could potentially be unmade in the future, if say Routable Components occur somewhere else such as `router.js`. I really don't have too much insight here - just trying to think things through and "measure twice, cut once".

### 7. Should we call Routable Components "Routable Components"?
Of course :wink:! In all seriousness though, if there is a better name I'm all ears! The final implementation could also help with naming - perhaps it's more along the lines of "Route Specified Component" as your example above outlines.

### 8. What is the proper order to move from Controllers to Routable Components?
> These phases feel a little backward to me.

Good points.

Rethinking this a bit, here is an alternative proposal based on your feedback.
### Phase 1: Query Params Service
1. Merge #380
2. Update guides on how to use Query Params
3. Deprecate the use of legacy Query Params in Controllers
4. Eliminate the use of legacy Query Params in Controllers

### Phase 2: Remove Controllers
I still think that removing Controllers, even if Routable Components _never_ materialize is worthwhile in and of itself. Controllers are (to me) so confusing, hard to teach, and unnecessary that I'd be happy just to see them gone. I'm certainly sympathetic to the idea of having a replacement for something before you eliminate it. In the case of Controllers though, I don't think this is needed and in fact I am proposing that the replacement to Controllers at first is...nothing! Just make Templates stateless and require developers to render a single top-level component if they want state. I'm hoping this is _temporary_, and that Routable Components land ASAP. However, the point of this RFC is to deliver something quick to deal with Controllers that is independent of Routable Components, which I believe simply removing Controllers fulfills.

1. Put deprecation warnings on Controller actions and properties
2. Update guides to remove references to Controllers, and add in references to using a single Component in a Template in order to use state and actions.
3. Eliminate Controllers after enough time has passed.

### Phase 3: Routable Components (or some other name)
I love your idea of having an alternative to test out to gauge how the community responds. Perhaps this could live as an addon at first? Part of the dream of modularization is to make these experiments easier - but I'm not sure if it's feasible for this or not. As you say "_Migrating takes a while_", so I'm not sure when this phase will happen, but to me, I'm in less of a rush to _get_ Routable Components than I am to _get rid of_ Controllers, hence why I'm okay with it being last. Also, I really hope that this time the feature actually lands :joy:.

Hopefully, I helped clear up some of your questions. I would love to hear your thoughts on this and get to a place of mutual understanding!

NullVoxPopuli:
@andrewcallahan I think we're misunderstanding each other on parts, but I think you've convinced me of something, which I wasn't expecting, but I'll get to that later.

> 1. Does this approach simply replace Controller js files with Component js files?
> I get that it's a bit more complicated than that, but fundamentally it's the same thing - some resolver looks through the files and hooks up a .hbs file in the /templates folder with a .js file in another folder (this time /components instead of /controllers). 

I think you're over thinking it a bit.

when you do `ember g component my-route-name`, it generates 2 files and a test. I'm just saying that for classic style, those files would be used. Nothing magic here -- just a component. :)

> 2. Should Templates have .js files at all?
> What I don't understand is why you hook up the Template file to a .js file in the /components folder 

I was trying to convey that the template file would take priority. if you have route template (like you would today), there would be no component lookup (like it works today).

>  For nested routes, I would presume this would mean app/components/rentals/edit.js would similarly be tied to the app/templates/rentals/edit.hbs file. If so, the /components folder structure would be directly tied to the route structure - just like the /templates folder is today.

Yeah, I def don't want that. Which as you bring up some of these issues, I'm thinking **we should close this RFC** until we have Template Imports, because I think this is kind of too hard, and we need to soley talk about it from the aspect of co-location.


> 4. Should Component .js and .hbs files be in the new /routes folder?
>  but in the MU scenario you have the Component .js and .hbs files in it:

In module unification, this is call a private collection. Today, there is a bunch of resolution magic to make this work, but I want to encourage this pattern for all components that are solely for a route and nothing else. I've even used in React projects, and it's worked well. :)

> I could see something like this being one possible way

My biggest concern with it would be that it would feel configuration heavy and maybe error-prone due to misspellings. 

> 5. Does this RFC, as it stands, take a fair amount of work?

it shouldn't. Of course I haven't looked into the rendering/resolution logic, but I _assume_ it's adding _one_ additional lookup rule, and then making sure the correct arg is passed to the component.
This may be a over simplification. 

> Phase 2: Remove Controllers

I think this is too disruptive and would require a major version bump. :-\
We should be resistant to major changes.

I understand that controllers are hard to teach, but we _must_ opt for easy migration paths when available, and not forcing people to jump ship. Remember when the "Routable Components" RFC was getting hyped, people started trying to avoid controllers at all cost? that was a bad idea. :wink: 

At the moment, I think this is the easiest path:

 - wait for the query params RFC
 - wait for template imports
 - wait for route/template co-location (or new project structure)
 - encourage stateless templates at the top level of a route, utilizing self-managed private collections per route to import whatever is needed at the route-template
 - with all of this, and probably a bunch of other features, cut a new edition to bring back cohesion with all the new features / layout stuff / patterns etc
 - after the edition, deprecate controllers as well as deprecate whatever else needs deprecating post-edition

So, yeah:

TL;DR: I'm not even sure controllers need to be replaced anymore... But maybe just deprecated after whatever the edition after Octane is.

andrewcallahan:
Thanks again for thinking this through with me @NullVoxPopuli - I now have a much better understanding of some of the aspects of your RFC that were confusing me!

I'm sad to see this RFC close (if only for now), but I think exploring how exactly Controllers would be removed, and what the implications would be has been super illuminating.

I'm hoping that as the Core Team makes the Roadmap this year if Routable Components replacing Controllers are indeed on the table, they should definitely look through this conversation for context around pre-requisites, timing, and roll out strategy.

ef4:
I agree that this is probably nicer if we base it off [co-location](https://github.com/emberjs/rfcs/pull/481), which is already a merged RFC and not a big lift in terms of implementation. I don't think that means this RFC is premature, we have that merged RFC to build off of. 

Some of the discussion I'm seeing is IMO repeating the mistakes that caused the previous thing called "routable components" to never ship. If this RFC succeeds, it will be because it keeps the scope reasonable.

I can get behind this RFC if it results in a simple rule that's easier to learn than current Ember. Something as simple as what I wrote above:

> Any time you looked up route template X and didn't find it, look up component X too and if it exists, use that.

Notice that says nothing about how components are looked up, where they go, etc. All that is both outside the scope of this RFC and necessarily *should work the same as the rest of Ember*, so there's less to learn. If route templates are looked up in a special way, nope nope nope. If we still end up with two different kinds of templates in two different folders hierarchies, I will be a strong :-1:.

I see no downside to having all your routes backed by components that are in their regular componenty directories. Yes, that means the route hierarchy can appear under `app/components`. I disagree that that's bad. You still have one true map to where all the route entrypoints are, which is `router.js`. That is the place to browse to learn every Ember app's structure. It's already the most reliable place to look, especially given classic vs pods vs aborted-MU, and eventually classic vs template-colocation.

NullVoxPopuli:
Yeah, maybe some of my explicitness here has shot me in the foot.

I need to update the RFC anyway. I'll try to comb over the confusion parts and address/simplify.

Thanks, @ef4!

NullVoxPopuli:
I've updated this RFC as well as updating a bunch of my comments

michaelrkn:
I love this RFC! Two thoughts:

1. After this lands, would it make sense for `ember new route` to create `app/templates/components/route-name.hbs` instead of `app/templates/route-name.hbs` (or, assuming co-location lands first, app/components/some.hbs`) by default?

2. I can't find it now, but I remember seeing a blog post suggesting an API that looks something like:

```
export default class SomeRoute extends Route {
  data() {
    return RSVP.hash({
      songs: this.store.findAll('song'),
      albums: this.store.findAll('album')
    });
  }
}
```

And when this `data` hook is used instead of `model`, then instead of loading a component as if it were invoked as `<SomeRoute @model={{this.model}} />`, the component is loaded as if it were invoked as `<SomeRoute @songs={{this.model.songs}} @albums={{this.model.albums}} />`.

One thing I really like about the RFC as it is now is how minimal the proposed change is. But it leaves us with an API that has this notion of a singular "model" that doesn't always make sense in a component-based world (and that actually often didn't make sense in our old controller-based world, either).

NullVoxPopuli:
> . But it leaves us with an API that has this notion of a singular "model" that doesn't always make sense in a component-based world (and that actually often didn't make sense in our old controller-based world, either).

yeah, I don't think this is a feature people would want to use right away. I think there are a few follow up RFCs that would be needed to make things _noice_.

I do think renaming the route's lifecycle hooks would be beneficial, as _model_ is **somewhat** of a backend data term, and is def confusing to people who aren't familiar with the nomenclature.

knownasilya:
@michaelrkn a future splarguments feature would cover number 2.

michaelrkn:
@NullVoxPopuli I'd be interested in what you think the follow-up RFCs would be - it might be good to add under "unresolved questions", or at least as a comment here. Thanks!

pzuraq:
@knownasilya one thing to note about that, I think the likely path for splarguments is to first introduce _just_ the ability to spread arguments themselves, just like `...attributes`. Something like `...@arguments`.

The ability to arbitrarily spread _any_ value, especially dynamic values, is much more complicated, both on an implementation level and on a design level. I'd like to see us get there, but I'm not sure how long it'll take, whereas the much more scoped ability to spread _all_ arguments will probably be something that can ship much sooner.

simonihmig:
Thanks @NullVoxPopuli for working on this, I think this is long overdue! :)

Though I do have some questions and concerns:

* I assume this should be a non-breaking change, as new Ember features are supposed to be introduced in that way. But as @ef4 already touched mentioning the automagically generated `{{outlet}}` templates when no explicit route template exists, having a component with a name that by coincidence matches the name of a template-less route will make the route render that component (that is not meant to render that route at all) under the new lookup rules. Fixing this is easy by adding that explicit `{{outlet}}` template, but this still would be a breaking change I think!
* Can you add some example how component-based route rendering is supposed to work with nested routes? So `/foo/bar` will render the `<Foo::Bar>` component. But how should the template for `<Foo>` look like? Will it have an `{{outlet}}`? Does that even work now?
* What I do not like so much about the proposal is the ambiguity added through having *either* a template-only route *or* a component-based one. 
    * So when you want to have a look at what is rendered for a given route, where do you look first, at `app/templates/my-route.hbs` or at `app/templates/components/my-route.hbs`? I believe this could be confusing, especially for new devs that aren't aware of the lookup rules, and wonder what kind of magic is happening there.
    * Given that named args ([RFC 276](https://emberjs.github.io/rfcs/0276-named-args.html)) and `{{this.arg}}` ([RFC308](https://emberjs.github.io/rfcs/0308-deprecate-property-lookup-fallback.html)) have introduced more determinism and reduced ambiguity, this feels like a step backward IMHO
    * it introduces another Emberism, that build tools (Embroider) AFAIU have to understand and resolve, before other tools like Webpack would be able to do their work, e.g. route-based code splitting. 
* Given the lookup rules will *implicitly* try to lookup a component with the same name as the route, this opens the door for naming conflicts with existing components.
  * Say you have a `/user` route, that you want to refactor to not use a backing controller anymore after upgrading to the new Ember version shipping the changes here. But you already have a `<User>` component, to e.g. render a simple user avatar, so not related to the `/user` route at all, but instead spread over many dozen places all over the app. 
  * That would require you to rename the existing component to make space for the new route-component, which could be a major PITA
  * It feels like you have to fight the framework, as the assumption of having a component with the same name as the route is baked into the framework

Given this, and that you already stated you would favor template-only routes, I wonder what the reason is to not propose just that, and not introduce the component lookup change at all?

Like when you do have some state/action on the controller now, you can still move that to a component (with either the same name as the route of a totally different one btw), and render that component explicitly in the route template:

```hbs
<MyComponent @model={{this.model}}>
```

AFAICT this would solve all the problems mentioned above. In fact it wouldn't even require changes on the framework itself, it would "just" be an (officially) recommended pattern and would require "only" changes to the guides. And pave the way for a future deprecation of controllers. Or do I miss something?

chadhietala:
I see this has become a lively discussion... 

I believe #454 #496 are critical for explaining how the component is "looked up". The TL;DR is that the component should not be looked up at all. The SFC and template import primitives are all about by-passing the container for resolution and relying directly on JS imports. For instance the below example is a direction I would welcome.

```js
import FooComponent from '../components/Foo'

@render(FooComponent)
export default class extends Route {
  arguments() {
    return { name: 'Chad' }
  }  
}
```

pzuraq:
Personally I would vote for static class fields, so we could avoid the extra decorator import and assign multiple components for different states:

```js
import FooComponent, { LoadingComponent, ErrorComponent } from '../components/Foo';

export default class extends Route {
  static render = FooComponent;
  static loading = LoadingComponent;
  static error = ErrorComponent;

  arguments() {
    return { name: 'Chad' }
  }  
}
```

chriskrycho:
Seconding/thirding @chadhietala's sentiment here. My biggest hesitation with the proposal as I understand it is that it's adding another magic rule around routes/controllers/components which new Ember devs would have to learn. If it can be built instead on other primitives that we're aiming to land, which eliminate the implicitness and give tools (IDEs, static analysis, etc.) a better point to hook into, that seems like a substantial win.

I don't want to suggest getting stuck in a "design it all up front" mode, but something like the static class fields approach given the ability to actually import components would be a very, very big win over "oh, you can use a controller *or* you can have a component named the same thing and that just works, but you'll have pain when migrating if you have an existing component which has the same name as a route and delete the controller…"

Another upside to the form @pzuraq proposed is that you *could* use the same component for all three states if you had one that composed a hypothetical `<Loader>` component with the actual layout for each of the custom states within it.

---

Given that the component colocation RFC was accepted, the text here should also be updated to reflect that in terms of the default locations of the component backing class and template files!

NullVoxPopuli:
> I believe #454 #496 are critical for explaining how the component is "looked up". The TL;DR is that the component should not be looked up at all.

I'd be a _huge_ fan if we just had a js-less template per route that either defaults to outlet, or just _is_ what we have today (but eventually with no controller)

> For instance the below example is a direction I would welcome. (`@render`)

that's interesting, but if we also need to follow the same pattern for loading and error, we'd wind up with something like this: 

```ts
@render(FooComponent)
@whenLoading(LoadingComponent)
@onError(ErrorComponent)
export default class extends Route {
  arguments() {
    return { name: 'Chad' }
  }  
}
```

> Personally I would vote for static class fields, so we could avoid the extra decorator import and assign multiple components for different states

yaaaaaas (I mentioned this in my #EmberJS2019 blogpost) :)

> Given that the component colocation RFC was accepted, the text here should also be updated to reflect that in terms of the default locations of the component backing class and template files!

legit :+1:


ok, so... My favorite options so far are the static fields, where:

> Another upside to the form @pzuraq proposed is that you could use the same component for all three states if you had one that composed a hypothetical <Loader> component with the actual layout for each of the custom states within it.

I think that's awesome, and would give a way for people who want to _heavily_ use components for everything a way to do so while also still being semantic/ergonomic for those of us that want to use routes :)

So, next updates to this:
 - rephrase everything in to be in a post SFC/Template-import world
 - static fields
 - ditch resolving? 
   - this _resolves_ @simonihmig's concerns over accidental breaking changes, but I think @ef4 was in favor of the resolutions? idk. (sorry!)

NullVoxPopuli:
Updated some things, mostly trimmed away a lot things that are now out of date.
 - I've added static fields
 - I mostly don't mention resolving
 - Add links to dependent RFCs

What I still have to do is flesh out examples, learning materials, and more motivation and transition for how migrating away from controllers would work. (I mostly just wanted to remove more of the outdated stuff so conversation can continue (in case people read the RFC before the comments))

andrewcallahan:
Love the changes to this RFC you've made @NullVoxPopuli! This feels so much simpler now and is a very intuitive way to connect a Route to a Component.

As this RFC stands now I see a lot of big wins:
1. Allows for the elimination of the `/controllers` folder since there is now an "alternative"
2. Allows for the elimination of the `/templates` folder since Components are now imported and declared in the Route (and this RFC assumes co-location is merged)
3. No more resolving .hbs files "magically" based on file name or location - everything is explicitly imported and declared in the Route (this will especially be helpful for loading and error states)
4. The `/components` folder hierarchy remains decoupled from the URL structure (which was my biggest point of confusion and disagreement before)

These are all so important that I don't know which is my favorite. It's all very exciting! Also, this feels awfully close to "Routable Components", whatever that may turn out to be :wink:

knownasilya:
Would love to see a migration section (https://mobile.twitter.com/scalvert/status/1140667889122336770) in this RFC.

NullVoxPopuli:
It'll be added in the "How we teach this section" :)

NullVoxPopuli:
though, if someone understands enough of what's going on here, and presently have the time to work on it, feel free to PR to my branch :P

ef4:
I'm OK with explicit import of components and I agree with that general direction, but I do want to point out that it doesn't give us any static analyzability that we don't already have, so long as people are sticking to the conventional names for everything (which they'll now be forced to do since `Route#render` is deprecated).

Embroider can already statically determine which Route, Controller, and template files go with each route. Having a convention means it's not ambiguous.

So design explicit imports if that's the developer experience we want. But don't design them that way assuming it's needed for static analysis, because it's not.

chriskrycho:
@ef4 I should clarify that my own emphasis on static analyzability is not for the *build* pipeline, where I recognize that, as you note, "convention means it's not ambiguous", but for all the other tooling. There, the explicitness of imports is a huge win, as it ends up being a thing *we don't have to build ourselves*. The fewer things I have to teach a language server to understand/etc. and the more I can just rely on tools' existing support, the better.

chriskrycho:
@NullVoxPopuli, the final unanswered question is [answered unambiguously in the guides](https://guides.emberjs.com/v3.10.0/routing/loading-and-error-substates/#toc_error-substates):

> The "reason" for the error (i.e. the exception thrown or the promise reject value) will be passed to that error state as its `model`.

josemarluedke:
I love the direction this is going! 

I do like a lot the ability to import a component to the route to render. It does address some of my concerns from my #EmberJS2019 blog post.

For the loading and error states, I believe it would be useful to pass an argument to these components to clarify what context they have been rendered. For example, if I would like to render the same component for all the three states, it would be difficult to know if it is loading or not. If your model returns a promise, you would need to get a helper for seeing if it is pending or not. Maybe Ember could pass a flag to these components, something equivalent to this: `@isLoading={{true}}` and `@isError={{true}}`. Or  `@isRouteLoading={{true}}` and `@isRouteError={{true}}`.

chriskrycho:
Even better would be to pass a tagged value, which is *nicer* for JS users, and *much* nicer from the perspective of TS users. Given these backing definitions (in TS, but they could equally just be "what we do" as long as it's well-defined):

```ts
enum State {
  Loading = "loading",
  Resolved = "resolved",
  Error = "error"
}

type Model<T, E> =
  | { state: State.Loading }
  | { state: State.Resolved, data: T }
  | { state: State.Error, error: E };
```

Assuming that whole item is passed as the model to the component, equivalent to this:

```hbs
<SomeComponent @model={{model}} />
```

From the JS perspective, you'd be able to check it like this:

```js

class SomeComponent extends Component {
  constructor(owner, args) {
    super(owner, args);
    switch (args.model.state) {
      case 'resolved':
        // normal handling...
        break;
      case 'loading':
        // loading handling...
        break;
      case 'error':
        // error handling...
        break;
      default:
        // can't actually get here
        break;
    }
  }
}
```

If we designed it to "spread" the arguments instead, as if this were the invocation:

```hbs
<SomeComponent @state={{model.state}} @data={{model.data}} />
```

Then on the JS side, it'd be the same, except with one fewer layers of access: just `args.state` instead of `args.model.state`.

The upshot for TS is that you could write something like this (assuming the original invocation style):

```ts

class SomeComponent extends Component<Model<String, Error>> {
  constructor(owner, args) {
    super(owner, args);
    switch (args.model.state) {
      case State.Resolved:
        // We can't see the `error` value here; it's a type error to access it.
        transformData(args.model.data);
        break;
      case State.Loading:
        // We can't see the `error` or `data` values here; it's a type error to
        // access either.
        break;
      case State.Error:
        // We can't see the `data` value here; it's a type error to access it.
        transformErrorForDisplay(args.model.error);
        break;
      default:
        // special helper that throws a runtime error, but more importantly
        // *type-checks for exhaustiveness*.
        assertNever(args.model.state);
        break;
    }
  }
}
```

You *can* do similar with the booleans, but in general this pattern is *much* more robust and reliable in my experience—even when you're just using JS, because you can never accidentally end up with `isLoading === true` *and* `isError === true` this way.

NullVoxPopuli:
I just added some examples and incorporated @chriskrycho's idea with the route state type args and such :) <3

michaelrkn:
Love where this is at!

When new routes are generated, I imagine we want to use this format. It would probably make sense to generate a component with the same name. Does that make sense? If so, should that be included under How We Teach This?

When a route is generated, what should be the default value for the loading and error components? Should an empty component for Loading and Error be included in new apps and the default values set to those?

michaelrkn:
Also, does this really depend on #496 and #454? It seems like this is complementary, not dependent.

NullVoxPopuli:
> Also, does this really depend on #496 and #454? It seems like this is complementary, not dependent.

It's more of a conceptual requirement in order to make teaching easier.

> we no longer need controllers, but they can be worked on in parallel, not depending on each other.

correct ;)

michaelrkn:
Are we missing anything here around the `index` and `application` routes? I assume they will continue working with controllers as-is, unless you generate the route explicitly. 

Maybe then in a 4.0 release they switch behavior to use components in their "implicit" form? Or maybe they will need to be explicit? I think these concerns are for a future RFC, but I just want to make sure we aren't missing anything for this RFC.

tleperou:
> When new routes are generated, I imagine we want to use this format. It would probably make sense to generate a component with the same name.

That means for the route `users` generated via the cli; we get **also** a component `users` automatically generated explicitly (in the file structure) or implicitly (maybe during the run time)?


> Love where this is at!

200% agree. I used to create a wrapper component to avoid using controllers and routers -when possible- for the application's business logic. I can't forward to using this RFC proposal :dancing_women: 

As @michaelrkn mentioned, the #380 dependency of the RFC may not be required for implementing it. However, is the proposal here really relevant without #380 ? Query parameters are the cornerstone of controllers :sweat_smile:

michaelrkn:
> That means for the route users generated via the cli; we get also a component users automatically generated explicitly (in the file structure) or implicitly (maybe during the run time)?

I would suggest explicitly, in line with how much of Ember is moving now.

> However, is the proposal here really relevant without #380 ? Query parameters are the cornerstone of controllers 😅

I think they are independent. You could have #380 and still have a route render a template, no component. And you could have this RFC, and have a route render a component, no query params. The two together cover all use cases, but neither blocks the other.

NullVoxPopuli:
They are technically independent, but as independent RFCs, they don't tell the story I want to tell. :-\

michaelrkn:
I had another thought today, although I don't know if it's any good. One possible disadvantage to the approach we have here and in #510 is that there's still some hidden/implicit behavior. Here in #499, we have some static properties that define what loading, error, and render components to use, and we just have to know the framework calls them in order from somewhere. In #510, we have an object that's returned from a hook that we have to know the framework will pass into the component as arguments. 

What if we took a much more explicit approach, like:

```js
import hash from 'rsvp';
import ProfileComponent from './components/Profile';
import LoadingComponent from './components/Loading';
import ErrorComponent from './components/Error';

export default class ProfileRoute extends Route {
  @service store;
  onEnter({ profile_id }) {
    renderComponent(LoadingComponent);
    hash({
      profile: this.store.findRecord('profile', profile_id),
      comments: this.store.query('comments', { profile_id })
    }).then((data) => {
      renderComponent(ProfileComponent, data);
    }).catch((error) => {
      renderComponent(ErrorComponent, error);
    });
  }
}
```

Then it would be very obvious what gets called, at what point, and what data is getting passed in. 

On the other hand, maybe this is overly verbose, and maybe it gives too much freedom for people to stray away from the happy path.

Just a thought!

tleperou:
That direction is interesting. The scope of the #499 has to be updated to go further (check out "Not Proposing" bullet point in Summary). I do agree that hidden/implicit behaviors are dangerous - not only for newbies.

On reading your thoughts, I see two topics:
* the lifecycle hooks: `onEnter` (would make sense to get `onLeave`?)
* the rendering method of a component from a Route

The verbosity of a hook, like this one, seems reasonable:
```js
async onEnter(params) {
  renderComponent(LoadingComponent);

  try {
    const routeDate = await this.loadRouteData(params);
    renderComponent(ProfileComponent, routeDate );
  }
  catch (error) {
    renderComponent(ErrorComponent, error);
  }
}

loadRouteData({ profile_id }) {
  return hash({
    profile: this.store.findRecord('profile', profile_id),
    comments: this.store.query('comments', { profile_id })
  });
}
```

### Confusing?

Do you feel too that manipulating the rendering from the route is a too-easy option for mistakes, and difficulties for teaching:
* render many components; why only one?
* render no component; is that an error?
* do we render a component or a route?
* if it's a component, why it's a route?
* where `renderComponent` injects the DOM? into an `outlet` from the route template, or in place of the entire route template?
* `renderComponent`, but `renderTemplate` and `render` from a route [had been deprecated](https://github.com/emberjs/rfcs/blob/master/text/0418-deprecate-route-render-methods.md)


### What is the story?

Some parts of the story we want to tell here #499 are common to #510 . I might be wrong, but in both cases, we want to get a more explicit way to what Ember already does -slightly different, basically using Component substitutes:
* `activate` / `deactivate` route
* `actions.error` / `actions.loading`
* `beforeModel`, `model` and `afterModel`, in the case of #510 

Using the Component's API as a substitute of some Controller and Route's behaviours sounds like telling the _that_ story:

```js
// routes/profile.js
import { State } from '@ember/route';
 
export default class ProfileRoute extends Route {
  // magic or explicit?
  // let's say, explicit
  state = new State();

  // route's lifecycle
  async model({ profile_id }) {
    // or `load` #510 
  }
  async onError(error) {}
}
```

```hbs
{{! routes/templates/profile.hbs }}
<h2>Welcome to you profile dashboard</h2>

{{#if this.state.loading}}
  <LoadingComponent />
{{/if}}

{{#if this.state.error}}
  <ProfileErrorComponent @error={{this.state.error}} />
{{else}}
  <ProfileSubComponentAndCo @profile={{this.profile}} />
{{/if}}
```

It provides a way to propagate route's error to child components

###  What about TS

TypeScript would let us explicitly implement interfaces, like:
```js
export default class ProfileRoute extends Route implements OnLoading, OnError {
  onLoading() {}
  onError() {}
}
```
But this is currently impossible because Ember does not rely on TS, right? It looks like Angular, though.

tleperou:
Does Route need hooks? If so, it would be great to explicitly implement them:
```js
import { OnRender, OnLoading, OnError } from '@ember/routing/route';
export default class ProfileRoute extends Route implements OnRender, OnLoading, OnError {}
```
But this is currently impossible because Ember does not rely on TS, right? It looks like Angular, though.


@michaelrkn `renderComponent` sounds like the [deprecated](https://github.com/emberjs/rfcs/blob/master/text/0418-deprecate-route-render-methods.md) `render` and `renderTemplate` 🤔 .

What about getting, also, a function instead of a static field? In the case, we want to handle states differently:

```js
import FooComponent, { LoadingComponent, ErrorComponent } from './components/Foo';

export default class extends Route {
  @service router;
  @service remoteLogger;

  static render = FooComponent;
  static loading = LoadingComponent;
  error(error) {
    if(error instanceof ErrorAPI) {
      this.remoteLogger.log(error, new Date().getTime());
    }
    return ErrorComponent; // how to pass `error` to that component?
  }
}
```

NullVoxPopuli:
I Updated the description with the status of this RFC -- which links to something @pzuraq spit balled in Discord ;)

NullVoxPopuli:
Closing in favor of:
 - [setRouteComponent](https://github.com/emberjs/rfcs/pull/731)
 - [Query Params as Derived Data](https://github.com/emberjs/rfcs/pull/712) and [Arbitrary Query Params](https://github.com/emberjs/rfcs/pull/715)
 
 Thanks for moving the ball with `setRouteComponent`, @pzuraq <3 🎉

PR #501, opened 2019-06-12
Deprecate inline link-to

NullVoxPopuli:
[rendered](https://github.com/NullVoxPopuli/rfcs/blob/deprecate-inline-link-to/text/0000-deprecation-inline-link-to.md)

### DIFF
--- a/text/0000-deprecation-inline-link-to.md
+++ b/text/0000-deprecation-inline-link-to.md
@@ -1,0 +1,34 @@
+- Start Date: (fill me in with today's date, YYYY-MM-DD)
+- Relevant Team(s): (fill this in with the [team(s)](README.md#relevant-teams) to which this RFC applies)
+- RFC PR: (after opening the RFC PR, update this with a link to it and update the file name)
+- Tracking: (leave this empty)
+
+# Deprecate Inline {{link-to}}
+
+## Summary & Motivation
+
+`<LinkTo>` does not support inline invocation, and while `{{link-to}}` does, deprecating the inline form will not only help people migrate but may reduce misuse of the inline arg (what can happen instead of converting to block-from).
+
+Because `{{link-to}}` uses a variety of positional params, it has been inconsistent with itself with respect to the meaning of each position of the params. Sometimes the first param is a route name, sometimes it is the content of the anchor tag.
+
+## Transition Path
+
+Convert all inline `{{link-to}}`s to block-form, or use the `<LinkTo>` component.
+
+## How We Teach This
+
+The ember guides would need to remove mentions of the inline-link form, such as:
+
+https://guides.emberjs.com/release/templates/links/#toc_adding-additional-attributes-on-a-link
+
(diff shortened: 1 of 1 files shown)


Panman8201:
I use inline `{{link-to}}` fairly often because I never have anything special within the block `{{#link-to "route.name"}}Nothing Special{{/link-to}}` so inline just shortened it up `{{link-to "Nothing Special" "route.name"}}`

I don't have a strong feeling one way or the other so meh. 🤷‍♂ But since Glimmer Components don't have positional parameters, and if `{{link-to}}` will eventually become a Glimmer Component then this RFC make sense.

To the alternate suggestion, I've created components that `{{#if (has-block)}}{{yield}}{{else}}{{@text}}{{/if}}` as a way to support both options.

NullVoxPopuli:
I think in a post-curly world, the options would be:

```hbs
<LinkTo @route='route.name'>
  Nothing Special
</LinkTo>

for block or

<LinkTo @route='route.name' @text='Nothing Special' />

for inline
```

mehulkar:
Is a lint rule, build time transform, or codemod possible to handle existing inline link-tos? (I personally don't like build time transforms of my code, but codemod/lint rule would be awesome. Either way, I don't have a horse in this race, since I do not use inline link-to!

rwjblue:
It's definitely easy to write a lint rule for it, and it _may_ be possible to write a simple codemod that would handle the majority of cases.

robclancy:
we made our own component for this which works like this...

`<Link @to={{route 'users.show' user.id}}>`

The benefit of this is that we can build the route anywhere and have it sent to the link. Whereas if you have separate attributes it gets a bit harder. Obvioulsy we shouldn't be using `<Link` since it is already used by head stuff. 

Just thought I'd mention it, either way I 100 percent want to see this change made.

lougreenwood:
I'm confused, the PR says:

> `<LinkTo>` does not support inline invocation

but above @NullVoxPopuli says:

> I think in a post-curly world, the options would be:
```
<LinkTo @route='route.name' @text='Nothing Special' />

for inline
```

So does that mean if this RFC is merged, we will need to also add "inline angle bracket link-to" to fill in the gap?

robclancy:
Personally I think the inline should just never be supported. It isn't how HTML works.

paulyoder:
I use the inline link-to all the time. I would love to see a codemod available if it's deprecated.

tleperou:
Even if I use inline `{{link-to}}` in my apps, I do agree with @robclancy to limit only to one option.

That makes sense to provide an HTML-DOM-like `<Link-to></Link-to>`:
* Developers feel more comfortable with a `<a>`-like component
* Emberistas love the convention over configuration philosophy, hence to stick with one opinion
* Easier to teach. In contrast, it avoids explaining why there are 2 different manners

NullVoxPopuli:
Support for positional link-to args has been removed in v4

NullVoxPopuli:
More information: https://deprecations.emberjs.com/v3.x#toc_ember-glimmer-link-to-positional-arguments

PR #502, opened 2019-06-15
Explicit Service Injection

NullVoxPopuli:
[rendered](https://github.com/nullvoxpopuli/rfcs/blob/explicit-dependency-injection/text/0000-explicit-dependency-injection.md)


webark:
@NullVoxPopuli Where does the class name come from? Is it just the path camel cased with the type tacked on? Would it be something you import? Is it the class name you define in the individual file that would get pulled out?

flashios09:
hi @NullVoxPopuli ,
i like the idea :clap:
this will add some imports but it will make the code more clear/readable
i wish if we can do the same with a model
so instead if this:
```js
this.store.findAll('post', '...')
``` 
we have:
```js
@service(StoreService) store;
// i prefer the pluralized form, `posts` not `post` for model name
@model(PostsModel) posts;
// ...

doSomething() {
    this.store.findAll(this.posts, '...');
    // or
    this.posts.somePostsModelMethod(...);
   // maybe
   this.posts.findAll(...);
}
```

buschtoens:
A few tangents regarding TypeScript.

Currently I (we?) do this for TypeScript:

```ts
import Service, { inject as service } from '@ember/service';
import BarService from './bar';

class FooService extends Service {
  @service bar!: BarService;
}
```

Alternatively you can get around the extra import of `BarService`, if you provide the service name explicitly and use the old Ember Object Model:

```ts
import Service, { inject as service } from '@ember/service';

class FooService extends Service.extends({
  bar: service('bar')
}) {
}
```

This works because of the clever [registry pattern](https://github.com/typed-ember/ember-cli-typescript-blueprints/blob/master/blueprints/service/files/__root__/__path__/__name__.ts#L9-L14). The reason you have to use the Ember Object Model is that decorators still can't yet change the type signature: https://github.com/microsoft/TypeScript/issues/4881

Once they could, the following should type-check and infer the service class automatically:

```ts
import Service, { inject as service } from '@ember/service';

class FooService extends Service {
  @service bar;
}
```

For the time being, if you don't want to or can't use the Ember Object Model, but dislike explicitly importing the injtectee class, you could use the trick that we discussed in https://github.com/machty/ember-concurrency-decorators/pull/50: Use a Babel transform to convert class property assignments to decorated properties.

```ts
import Service, { inject as service } from '@ember/service';

class FooService extends Service {
  bar = service('bar');
}

// gets transformed into

class FooService extends Service {
  @service('bar') bar;
}
```

I have a transform for it ready, but I need to battle-test and optimize it further:
[`babel-plugin-transform-class-property-assignment-to-decorator`](https://github.com/buschtoens/babel-plugin-transform-class-property-assignment-to-decorator)

gossi:
@buschtoens

> Alternatively you can get around the extra import of `BarService`

if I understand this RFC correctly, it is about explicitely importing the respective class and exchange the string lookup for the class definition.

Second is: As much as we love typescript, this RFC must work for JS and as such the registry pattern from e-c-ts isn't available in pure js land. So with this RFC TS code will look like this:

```ts
import Service form '@ember/serivce';
import NotificationsService from 'my-project/services/notifications';

class Foo extends Service {
  @service(NotificationsService) notifications: NotificationsService;
}
```

at least until decorators can mutate the type definition for properties (then we can get rid of the _doubled_ written class name).

buschtoens:
@gossi This is exactly why I raised these points. While Ember is committed to not pushing TypeScript onto anybody and offering first-class support for JS, we are equally committed to offering first-class support for TypeScript as well.

I find it important that, as long as decorators cannot change types in TS or using a Babel transform is not a community-accepted and agreed upon opinion, we do not deprecate string lookups, as this could worsen ergonomics for some TypeScript users.

buschtoens:
One further observation: The Ember Resolver / Container system is a bit of magic in the background. You don't _always_ know where the backing injectee class is actually located or it might not be accessible to the code you are authoring.

There's also nothing preventing you from registering the same class with multiple names or even generating the names at runtime.

Admittedly these are uncommon edge cases.

lougreenwood:
Also, by requiring explicit class name use, we're coupling the class to a specific service, but this shouldn't be a concern of the class:

```
Dependency injection is one form of the broader technique of inversion of control. The client delegates the responsibility of providing its dependencies to external code (the injector). The client is not allowed to call the injector code;[2] it is the injecting code that constructs the services and calls the client to inject them. This means the client code does not need to know about the injecting code, how to construct the services **or even which actual services it is using; the client only needs to know about the intrinsic interfaces of the services because these define how the client may use the services**. This separates the responsibilities of use and construction.
```
https://en.wikipedia.org/wiki/Dependency_injection

I mean, if we're going to import the `ClassName`, why not instantiate the service singleton and skip the `service()` container lookup all together.... But we all know that's a bad idea as it removes IoC.

So whilst I like the look of this as purely a nicer way to write TS classes in Ember right now - it's not really DI any more since the explicit class name is required (no more IoC).

So it seems that if the goal is to improve typing in TS, proper TS support for decorators will give us that.

For newbies that get confused by strings, maybe we need to better teach the newbies about DI and why a string name is about as explicit & accurate a DI lookup name should be to maintain proper separation.

buschtoens:
@lougreenwood I totally agree with the point your making regarding isolation of concerns.

However, I think @NullVoxPopuli was not aiming at better TypeScript support, but instead better "cmd+clickability" support for generic JS IDEs / users that _don't_ use TypeScript.

<sup>Please correct me, if I am wrong.</sup>

lougreenwood:
yeah - you're right, I was getting ahead of myself - sorry :D

But IMO, we also shouldn't start hacking apart established patterns for nicer IDE support - our solution should "work with the patterns" as well as "using the platform".

So if I understand JS & IDE's correctly... Since JS is not typed, and by it's nature DI de-couples - then there's no "platform" mechanism which is true to DI other than typing (and decorators changing function signature) which allows nice IDE integration?

NullVoxPopuli:
@webark the class would be an import of the service itself :)

```ts
import Component from '@glimmer/components;
import { inject as service } from '@ember/services;
import MyService from 'appmame/services/my-service';

  @service(MyService) myService;
}

```

:)

NullVoxPopuli:
@buschtoens the registry pattern could be used with classes as keys, couldn't it?

NullVoxPopuli:
> then there's no "platform" mechanism which is true to DI 

If only we had interfaces we could use instead :)

NullVoxPopuli:
Renamed RFC due to technicality in DI definition

buschtoens:
> the registry pattern could be used with classes as keys, couldn't it?

@NullVoxPopuli Yes, of course it could. But that's not the point I was trying to make. 🙂

Switching to a class-based registry / decorator and deprecating the string-based one, means that the future TypeScript ergonomics will be much worse, because you can't any more use the property name to infer the injection, as show in https://github.com/emberjs/rfcs/pull/502#issuecomment-502341690.

NullVoxPopuli:
Gotchya, so string lookup stays :)

webark:
> the class would be an import of the service itself

Thanks for clearing that up!

So there’s a pattern where you extend a dependencies service, to either overwrite or extend the dependent service. 

When this is done, all of your existing references are updated.

With direct imports, would you always import a dependents service from the addons merged “app” space, even if it doesn’t exist? Or would you need to go and update all of your imports where you had imported them from the depended addon?

NullVoxPopuli:
>  would you always import a dependents service from the addons merged “app” space, even if it doesn’t exist? Or would you need to go and update all of your imports where you had imported them from the depended addon?

it should resolve to the same thing in both scenarios, afaik

buschtoens:
> > With direct imports, would you always import a dependents service from the addons merged “app” space, even if it doesn’t exist? Or would you need to go and update all of your imports where you had imported them from the depended addon?

> it should resolve to the same thing in both scenarios, afaik

An addon that has a service re-export in its `app` tree would make the service available in the host app, via the resolver, however the file would not be physically "there", meaning that "cmd+click" would fail, while the import would work at runtime.

---

I don't want to explode the scope here, but this segways into another interesting problem: What is the "official" way (pre and post this RFC) to override / clobber an addon's service (from another addon possibly)?

One way is using `ember-addon.after` in the `package.json` of the overriding addon to ensure that its `app` tree clobbers the `app` tree of the overridden addon. You could do the same from the host app as well.

One major hazard of this approach is https://github.com/babel/ember-cli-babel/issues/240: If you use a file extension other than `js`, the clobbering might fail non-deterministically.

NullVoxPopuli:
> I don't want to explode the scope here, but this segways into another interesting problem: What is the "official" way (pre and post this RFC) to override / clobber an addon's service (from another addon possibly)?

maybe this RFC should be held off until embroider ships ;)

but, I don't think we should continue to have clobbering of services. It's hard to debug.
if I import a service form an addon, I want that to be the actual location so I can ctrl/cmd click it and see what it is.

webark:
> An addon that has a service re-export in its app tree would make the service available in the host app, via the resolver, however the file would not be physically "there", meaning that "cmd+click" would fail, while the import would work at runtime.

this just goes against the main motivation of “enable "go to definition" support from service definitions so developers can more easily discover the where and how their service is defined.” 

For using services in addons, it seems just as magical also, and more confusing around “is this a singleton” especially if you could import the same service from the invisible “app” space, and where it lives in the addon’s “addon” space in the same app.

mehulkar:
This RFC should also probably provide some text about how to register/inject services in component rendering tests.

chriskrycho:
Regarding TypeScript lookups: FWIW, the string-keyed lookup is *way* more of a pain there than the proposed API here, in some hypothetical world where `@service bar` is allowed to change the type of `bar`. Today, we do some serious shenanigans to make `bar = service('bar')` work, and the proposed API here would entirely sidestep those. We wouldn't *need* the type registry anymore, which would be a huge win; the types would Just Work™ pretty much automatically with any conceivable implementation of `@service` in a world where decorators have support for modifying types. TL;DR: the second we can burn the string-key-lookup/type-registry pattern to the ground, I will. 😆 

As regards the proposal overall, @ef4 and I actually batted around an idea in this vague direction sometime last year on Twitter, and I'm strongly in favor of doing something in this general direction. I'd like to see a proposed API here for how you provide a mock for a given service in e.g. test contexts, though, @NullVoxPopuli, same as @mehulkar suggested! Without a clear way to do that, this proposal is quite thoroughly hobbled.

lougreenwood:
@chriskrycho but isn't what your describing not really DI (which is what we describe the service injector as in the docs) - granted our current system isn't DI either... since currently and as this RFC proposes, it requires the class to know the exact device to be injected, instead of a common service interface.

>The client should have no concrete knowledge of the specific implementation of its dependencies. It should only know the interface's name and API. As a result, the client won't need to change even if what is behind the interface changes. However, if the interface is refactored from being a class to an interface type (or vice versa) the client will need to be recompiled.[12] This is significant if the client and services are published separately. This unfortunate coupling is one that dependency injection cannot resolve.

> The injector introduces the services into the client. Often, it also constructs the client. An injector may connect together a very complex object graph by treating an object like a client and later as a service for another client. The injector may actually be many objects working together but may not be the client. The injector may be referred to by other names such as: assembler, provider, container, factory, builder, spring, construction code, or main. 

https://en.wikipedia.org/wiki/Dependency_injection

Assuming we want to have "true" DI for services, and we also want to address the points of testability (@mehulkar) and clobbering (@buschtoens) we could consider moving to a DI system where we continue to use strings to represent interface names, but those strings lookup against some new DI/service config, where we map the service interface name to a specific service implementation and the decoupling occurs in this config.

Perhaps there's also some way this config could export the mapped services as class names, so the intent of this RFC can also be delivered and we get as close to DI as JS-without-interfaces allows us?

lougreenwood:
Or maybe I'm just taking this whole "injector" stuff too literally, and what we call the "service injector" is just the Emberism for retrieving the service singleton instance and actually has nothing to do with DI, but over the years has become confused with DI?

Which kinda makes sense when considering that only services & controllers have these features, and both are singletons...

NullVoxPopuli:
@lougreenwood & @chriskrycho 
I think it would be _truer_ D.I. as the lookup could _act_ like an interface.

For example, to enable clobbering, to show an example of registration, and stub services for testing:

(also pardon the naming, as it's for demonstration)
```ts
// This behaves as the base implementation
// This doesn't have to be interface-like, but I'm going to say it is... 
// _semantics_..
class MyInterface extends Service {
  @tracked foo = 0;

  add() {
    this.foo++;
  }
}

// registration would look like it is today
appInstance.register(MyInterface, MyInterface);
// where we register MyInterface on the *key* MyInterface, just as today, it would look like
appInstance.register('service:my-interface', MyInterface);
```
Now, where this _IS_ Dependency Injection, and how we aren't just using the concrete class all the time is where you can do things like this

```ts
class MyImplementation extends MyInterface {
  add() {
    this.foo += 2;
  }
}

// both stubbing (in a test), or clobbering (for whatever reason), would look the same
appInstance.register(MyInterface, MyImplementation);
```

ef4:
1. The actual RFC text doesn't show the import of the names, which is pretty critical to the design.

2. We shouldn't confuse DI with modules. They have different semantics. As others have pointed out, if all we cared about is taking an exact module and instantiating a per-app singleton of it, we wouldn't need DI at all. The existing system *really is* intended as a DI system, which is why it's possible to swap out the implementation in a particular test. It's also why we can swap out things like the DOM abstraction in order to run Ember on top of the browser vs fastboot. That kind of swapping is not legal under the ES module spec.

   Now, there may be a way to interpret this proposal that straddles these concerns. Maybe the class that you pass to `service` is interpreted only as an *interface*, and not an *implementation*. But that is a pretty subtle distinction to teach. 

3. You're implying in the `register` examples that services still have names, but I don't think that's true in general. We can't derive the name from the class, unless we introduce additional verbosity in the Service definition side or additional build system magic.

4. There are several things that feel like they *should* work if the API is shaped this way, that could not be allowed to work. For example:

   ```js
   // this would be bad style, plus how would anybody ever override this with `register`?
   import SomeService from './random/path/not/under/services/some';
   @service(SomeService) someService;

   // also, this "should" work, but completely defeats the point of DI
   @service(class extends Service { customStuff() {} }) inlineService;
   ```

NullVoxPopuli:
Added examples of registering and clobbering services

pzuraq:
One thing that I think might be pretty valuable with explicit dependency injections like this is the ability to register services in addons own components. Currently, if you want to have a service for a particular addon, you _must_ reexport the service in the `/app` folder. There's no other way to expose the service to the consuming application.

I'm not sure if Emroider has a solution to this yet, but AFAICT, if we ever want to remove the `/app` folder entirely, we'll either need users to register services themselves, or come up with an alternative. This seems like it would solve the problem, if we can work out the details. It'd definitely be worth calling that out in the motivation.

NullVoxPopuli:
@pzuraq since we're heading in the direction of explicitness over resolutions, I think making the apps register services from addons would be good. it would remove a lot of whimsy as well

chriskrycho:
@ef4 point of clarification for my benefit:

> 4. There are several things that feel like they _should_ work if the API is shaped this way, that could not be allowed to work. For example:
>     ```js
>     // this would be bad style, plus how would anybody ever override this with `register`?
>     import SomeService from './random/path/not/under/services/some';
>     @service(SomeService) someService;
>     
>     // also, this "should" work, but completely defeats the point of DI
>     @service(class extends Service { customStuff() {} }) inlineService;
>     ```

Largely agreed on the second example there. On the first, however, is the a reason we should forbid that? Obviously services live in a conventional location today; however, it's not clear to me why they *must* live in a conventional location going forward. (Not, I think, that this proposal should cover changing that behavior, but this suggests—as do several other points in the thread—that a revision to the DI system may be a prerequisite for a change of this magnitude.)

NullVoxPopuli:
```ts
// also, this "should" work, but completely defeats the point of DI
@service(class extends Service { customStuff() {} }) inlineService;
```
@ef4 / @chriskrycho I don't think this would work, because `@service` is still doing a lookup, and if an anonymous class is passed to `@service` there is nothing to lookup... _unless someone registered a service under the `Service` key..._ (I would declare "how dare they!")

chriskrycho:
@NullVoxPopuli there's no difference between that and the import from the module whatsoever. (While TS differentiates between the type and the runtime object, JS only knows of the runtime object.) You can do all sorts of fun things as a result (like `const foo = class Bar {}`), and the default export is no different in that regard from any other.

pzuraq:
@NullVoxPopuli I'm not sure that's true, in the case of `ember-animated` for instance the service is meant to be fully private. Exposing implementation details to users is definitely not ideal in this case, it's not just being explicit.

I allowing addons to inject their services this way I think is still explicit, and solves the problem of privacy and setup for addons.

NullVoxPopuli:
right, but _whatever_ the runtime object is -- is still lookup-able
image


in the picture, imagine we registered via
```ts
appInstance.register(B, C);
```

when we `appInstance.lookup(B)`, we get an instanceof C.

that's all we need afaik

lifeart:
ctrl+click work in `unstable language server` )

NullVoxPopuli:
> ctrl+click work in unstable language server )

no need if TSServer handles it :)

lupestro:
I'm a little concerned about supporting the true DI case where early during application startup we choose between independent implementations of the same (duck-typed) interface to register based on deployment-time data. You set it up in the registry once and no longer care which class was supplied. The examples all seem to assume there is a single "true" implementation and all the DI is for test stubs.

NullVoxPopuli:
@lupestro, it'd work if the same class key was used.

The main difference is that a class is used instead of a string. There technically doesn't have to be any correlation. But I'm recommending that at the very least the concrete class is a subclass if not the OG implementation

mehulkar:
The main problem being solved here is traceability. “Why does inject with this string mean that this Singleton is available?”

An alternative solution (and less disruptive I think), would be to instead remove the inject import, and require that the service is looked up and set as a property in the constructor. As a parallel to that, the registration of the Singleton into the container should also be explicit in a app boot hook. 

This would solve the traceability problem, reduce the surface area of the framework, and also retain “true DI” as others have pointed out. 

If nothing else, maybe this can be added to the Alternatives section of the RFC?

NullVoxPopuli:
> An alternative solution (and less disruptive I think), would be to instead remove the inject import, and require that the service is looked up and set as a property in the constructor. As a parallel to that, the registration of the Singleton into the container should also be explicit in a app boot hook.

Can you provide some examples?

mehulkar:
I'm imagining something like in the component

```js
  init() {
    this.set('foo', getOwner(this).lookup('service:foo'));
  }
});
```

and this in app/app.js in the default blueprint

```js
Application.extend({
  onBoot(appInstance) {
    const serviceClasses = discoverServices();
    serviceClasses.forEach(s => appInstance.register(`service:${service.name}`);

    // similar things for components and everything else that can be looked up from the container
  }
});
```

Two things:

- I don't know if the semantics of setting the service in init/constructor are the same as the `inject` import
- The `onBoot` hook is completely made up and any resemblance to existing mechanism is purely coincidental. HOWEVER, I think "explicit" registration of things in the container as part of the default blueprint from `ember new` would solve many of the "magic" problems (and would also unlock more experimentation for many things!)

pzuraq:
@mehulkar I'm not sure that solution would end up being much more traceable, and it would incur a lot of extra overhead. You would essentially always have to go to the `app/app.js` file to find the definition of a service, since it could be coming from anywhere. The language server wouldn't be able to pick up on it without a lot of custom rules, which would probably be brittle.

I think the value of explicit, import based DI is it makes it very clear what the system is doing. You are injecting an _instance of this class_ - that can be inferred pretty easily, even by someone who has no knowledge of an Ember app. They can also find that class pretty easily, given the definition. They don't have to know any implicit file lookup rules, or the way that a particular application has configured its service registrations (which, if it's done manually, will likely become bespoke per app).

Re: @lupestro's concerns, I think that while this is uncommon, it's definitely a problem we need to consider. I really like the idea that if you want to inject a service class, then the actual implementation must either be the class, or a subclass. This would allow us to implement an abstract class for the definition, with concrete implementations that extend from it:

```ts
// Typescript
abstract class CookieService {
  abstract getValue(key: string): string {}

  abstract setValue(key: string, value: string): void {}
} 

// Component definition
  @service(CookieService) cookie;
}

// Impl
class FastbootCookieService extends CookieService {
  // ...
}

class BrowserCookieService extends CookieService {
  // ...
}

// Initializer
export function initialize(appInstance) {
  if (isFastboot()) {
    appInstance.register(CookieService, FastbootCookieService);
  } else {
    appInstance.register(CookieService, BrowserCookieService);
  }
}
```

mehulkar:
> I think the value of explicit, import based DI is it makes it very clear what the system is doing.

Just to be clear, [I would prefer this also](https://mehulkar.com/blog/2019/06/emberjs2019-roadmap-response/) 😄, as long as there continue to be a way to inject a replacement (in tests or at build time as @lupestro brought up).

> The language server wouldn't be able to pick up on it without a lot of custom rules, which would probably be brittle.

Yep, same situation as now, I presume.

> I'm not sure that solution would end up being much more traceable, 

Maybe not today, since app.js is a largely ignored file. But I think `app/app.js` could be a much more useful entry point for Ember projects in general. Preparing the container and "discovering" core classes like components and services would be a very useful part of the default blueprint and give devs a clearer understanding of Ember's mental model. But I agree that that's a much larger discussion and could have other side effects that can't be explored as part of this RFC.

> and it would incur a lot of extra overhead.

For learning or for performance? From a learning standpoint, I don't think it's unreasonable to:

1. have a mental model of a container
1. know that things get put into the container at some point
1. know that other parts of your app look these things up in that container

We already understand 1 and 3. It's just 2 that's invisible and is the part that isn't traceable.

If the blueprint continues register the default export from `app/services/foo.js` as `service:foo`, the convention promise is met, *and* the dev gets more freedom and better understanding of what's happening (which is what this RFC is all about, IMO).

You still never have to look at `app.js` though. Injections can work the exact same way.

Anyway, I'm all for injecting imported classes! This alternative makes a lot of sense to me and also paves the way for other things to become more explicit (like initializer and instance initializer functions). Requiring a base class for the weakmap also makes sense. Services are already required to be extended from `Ember.Service` (pre-native classes, anyway), so this shouldn't preclude any existing use cases that I can think of.

Thanks for hearing me out!

mehulkar:
Actually, one other point about the common base class as a weakmap key...

@pzuraq your example of having an abstract base class is great, but I worry that people will extend the *real* class to create the stub, rather than first extract an abstract class.

For example, if you have a service with 100 methods in it and you want to write an isolated test that will make use of one of those methods. If you extend from the real service and stub out that one method, you'll still end up with 99 other methods that will eventually leak and break your test. This would be either be a very bad pattern to set or be asking for a lot of overhead for devs to create an abstract class before writing a test.

pzuraq:
> > and it would incur a lot of extra overhead.
>
> For learning or for performance?

I actually meant for:

1. Tooling
2. Day-to-day app development

If we were to force manual registration in `app/app.js`, like you're suggesting, even if the blueprints were to manually add the registrations, the fact is that in time and in very large applications, they will get out of sync with the actual folder structure, import paths, etc. I've had to do this for statically typed languages, and even when you have the type system helping you, it becomes a mess pretty quickly.

It essentially will mean that for every single service you see in an app, if you want to find it you will have to:

1. Open up `app/app.js`
2. Look for the service that is registered
3. Look for the import path of that service
4. Open up the actual file

Sometimes you'll be able to skip some of these steps based on conventions, yes, but we won't be able to writing general purpose tooling that could take you through it with a click, since any custom registration could occur at any point.

FWIW, this is actually closer to where we _used_ to be, before we had the `service()` injector/decorator. You would have to manually register values in the container in app instance initializers (which is likely where this type of wiring would go) and you would have to manually look up registrations. I think `service()` was a valuable iteration on that period, it really cleaned up Ember and made it much easier to understand. It did this, IMO, because it didn't violate the "as-above-so-below" rule as I call it - you should never have to traverse to the other side of an application/repository/library/ecosystem to understand what is going on locally.

pzuraq:
> but I worry that people will extend the real class to create the stub, rather than  first extract an abstract class.
>
> For example, if you have a service with 100 methods in it and you want to write an isolated test that will make use of one of those methods. If you extend from the real service and stub out that one method, you'll still end up with 99 other methods that will eventually leak and break your test. This would be either be a very bad pattern to set or be asking for a lot of overhead for devs to create an abstract class before writing a test.

Honestly I think if you have a service that has 100 _public_ API methods, you probably have too big of a service 😄 There are also plenty of cases where you _don't_ want to override all of a services implementation methods, so I think it's a totally valid pattern to extend a service without creating an abstract base class. 

The point on stubbing is to create a service that behaves more or less the same externally, but doesn't have certain side-effects. It's usually better to keep most of the service's logic the same, and isolate the parts that need to be changed for tests. This way, you tightly close over the domain in a way that is transparent to the test suite. A great example is Pretender - it completely hides the fact that it's mocking XHR requests.

kimroen:
Having worked a bit with Angular in the past few months, looking at how [dependency injection and services works there](https://angular.io/guide/architecture-services) was pretty interesting:

```ts
import MyService from '../services/my-service';

  constructor(
    private myService: MyService
  ) {}
}
```

`this.myService` now works inside this component.

As far as I can tell this though, this is entirely based on it being TS so it's not necessarily something we could use directly, but maybe there's something to draw inspiration from?

NullVoxPopuli:
@kimroen, that's how it works in asp.net as well, and was def part of the inspiration that went into this :)

text/0000-explicit-dependency-injection.md
```diff
@@ -0,0 +1,234 @@
...
+```
+
+A new map can exist on the registry that can appropriately be wired up to register, unregister, etc to handle the lookup-by-class-definition at the same time as the current registry behaves, allowing for backward compatibility with todays injection usage.
+
+Examples: 
+
+(also pardon the naming, as it's for demonstration of the _roles_ that each class definition can have)
+```ts
+// This behaves as the base implementation / abstract class
+class MyInterface extends Service {
```
kellyselden: In terms of teachability, confusing the difference between interface and implementation is confusing.

NullVoxPopuli: In what ways do you think it could be made more clear? Do you think maybe I should remove all the interface terminology and use Abstract Class instead? 
(I still might do that, cause interfaces don't have implementations)

kellyselden: Maybe instead of `MyInterface` and `MyImplementation`, `MyRealService` and `MyReplacementService`. In this case, naming it like how someone would actually use it would go further in the understanding, instead of trying to shoehorn CS concepts in IMO.

NullVoxPopuli: done! thanks!

chriskrycho:
> Having worked a bit with Angular in the past few months, looking at how [dependency injection and services works there](https://angular.io/guide/architecture-services) was pretty interesting:
> 
> ```ts
> import MyService from '../services/my-service';
> 
> class MyComponent extends Component {
>   constructor(
>     private myService: MyService
>   ) {}
> }
> ```
> 
> `this.myService` now works inside this component.
> 
> As far as I can tell this though, this is entirely based on it being TS so it's not necessarily something we could use directly, but maybe there's something to draw inspiration from?

That’s correct; they only make their DI system work this way by emitting runtime metadata from TS (including decorators in particular). That’s obviously a non-starter for Ember.

pzuraq:
Something we have discussed is in the future, if decorators land and eventually, _parameter_ decorators are proposed and land (could take a while), we could do something like:

```js
import MyService from '../services/my-service';

  constructor(
    private myService: MyService
  ) {}
}

**Comment by pzuraq** [comment]
Over the weekend I hacked out an addon that implements this RFC using Ember's public APIs, [`ember-totally-not-module-based-services`](https://github.com/pzuraq/ember-totally-not-module-based-services) (the name is meant to be humorous, see the README 😛). I figured it'd be a good idea for us to have some reference for how this RFC would actually work, and a place to experiment with the API to see if it's ergonomic or not, and what it's shortcomings might be. I also made an [`@abstract`](https://github.com/pzuraq/abstract-decorator) decorator for defining abstract classes, which could be handy for following the patterns that have been described in this RFC.

After more thought and playing around with the API, I think the biggest concern I have so far is that we don't have a simple, declarative way to override a service at the application level. Today, if you need to override a service `foo` to add some functionality, you can create the `services/foo.js` file and you're done with it. For cases with multiple implementations, like fastboot, this was never so simple, since you would have to have some branching logic somewhere such as an instance initializer, but it does seem like adding the step of creating an initializer _just_ to override a service would be a bit of unnecessary boilerplate.

I've been toying around with the idea of a `@register` or `@override` decorator, something that we could use the _declaratively_ tell the DI system that this class is meant to be registered at a particular key:

```js
@override({ key: Store })
class MyCustomStore extends Store {}
```

It could even have conditionals:

```js
@override({ key: CookieService, if: !isFastboot() })
class BrowserCookieService extends CookieService {}
```

The duplication of the key in these cases wouldn't be ideal, since you're both extending and registering it, but the declarative nature of a class decorator would be nice because it actually is statically analyzable (unlike initializers), and would also get us away from relying implicitly on the ordering of the build system. I do think it gets tricky if you try to allow _addons_ to use a decorator like this as well - if both the app _and_ an addon (or two addons) attempt to override the same service, we have an ordering problem once again. This could also come later on, or in parallel to this RFC - it would work just as well with string based service keys, and would have the same benefits of getting us off using the build system for resolution if we can nail down the semantics.

chriskrycho:
FWIW, on this point:

>  Today, if you need to override a service foo to add some functionality, you can create the `services/foo.js` file and you're done with it.

I very much understand _why_ people reach for this, but every time I've seen it, it's an anti-pattern—the amped up version of the failure which "prefer composition over inheritance" is meant to address. It may be a thing we want to support as an escape hatch, because sometimes addon's services are badly designed, but even there, simply injecting it into one's *own* service and providing a wrapper API around it seems to me to be a much better pattern than the override-and-it-happens-to-have-the-sam-ename pattens we currently use.

The trick, of course, is for the scenario where you want it to be resolved back in the original addon's code or similar, but that *also* seems like a very, very smelly smell to me.

I don't have more fully-formed thoughts than that, but I figured I'd raise it as a point against the importance of that kind of "overridability".

Semi-related: perhaps the design of a new system here should not *assume* it, so as to be backwards compatible, but it seems to me that accounting for a future where addons' exports are no longer merged into the consuming app's namespace may illuminate a better design here. If you don't have namespace merging, you also don't have the *ability* to override in the same way, and new patterns will need to emerge to account for the kind of escape-hatch work referenced here. Again, we may not want to defer *this* work until that time, but leaving it as an intentional hole in the design—acknowledged and unanswered, designed to be filled in later when we have more of an idea how that system will play out—may be a reasonable option.

simonihmig:
I have some concerns with this, mostly along the lines of what @lougreenwood and @lupestro said before.

While the RFC is mostly concerned about basically what the key for a lookup operation is (a string or a class), this small-ish change can only properly be assessed when we are clear about what the scope of our DI system is, and where it is eventually heading to:

**1. Singleton container with test mocking**
In this scenario the assumption would be to have one single implementation for a given service, and you already now where it is (in `/app/services/name.js`), so you know where to import it from. But as you don't want to create separate instances, our "DI" system's main purpose is to manage singleton instances. And additionally have some means to switch implementations mainly during tests.

This is probably how we use Ember's DI in most cases, but effectively that is a very limited form of DI. And given the dynamic nature of JS, we could mock services in tests even without any dynamic registering/lookup.

**2. True DI**
In "True DI"(TM) the system would not make any assumptions of where a dependency comes from, and how it is implemented. The explicit goal of Inversion of Control (IoC) is to foster decoupling. So having to import a concrete *implementation* of a service just to declare a dependency on *any* implementation that satisfies the needed *interface* is counter-productive, as it *increases* coupling. Rather neither the dependency nor the consumer should be aware of each other (other than through the interface). The consumer ideally should just declare a dependency on some interface, which the DI needs to inject an implementation for.

The obvious problem we have here is that our platform does not know the concept of an interface (even TS has that only at build-time). So IMHO our string-based lookup actually comes closer to that than any injection based on a concrete implementing class. Even an abstract class seems wrong to me, as to my understanding that is still concerned on an (incomplete, sharable in a class hierarchy) *implementation* rather than defining an *interface*.

Let's look at an example: 

We want to have a central `logger` service, but with different implementations based on context:
* `ConsoleLogger` for development
* `SentryLogger` for production
* `WinstonFileLogger` in FastBoot

Now we have an `auth` service that needs to log things. What logger implementation should that import for the injection declaration? This get's even worse when an addon's service needs a `logger` injection, how would it be able to decide which class to import?

Having an abstract class for that was a suggestion, but as said before, I don't see an abstract class as a valid replacement for some other way to declare a dependency on an interface. 

And I also see some practical problems with such an (abstract-class based) approach. Ideally services wouldn't need to extend from `Ember.Service` or any other common (abstract) class anymore. Instead two possible different implementations could come directly from two different npm packages, not extending `Ember.Service` or having any other relation to Ember at all, as long as they satisfy the same interface.

**TL;DR**

While our current DI is commonly used in a way as described under 1., maybe that is because it is IMO quite limited still. At least I would love to see it evolve to be more flexible in a "true DI" sense. And in that regard the proposed changes here seem counter-productive, as they increase coupling while IoC is supposed to decrease it. And as such it would make teaching the benefits of IoC and loose coupling more difficult.

NullVoxPopuli:
> At least I would love to see it evolve to be more flexible in a "true DI" sense. And in that regard the proposed changes here seem counter-productive, as they increase coupling while IoC is supposed to decrease it. And as such it would make teaching the benefits of IoC and loose coupling more difficult.

Kinda throwing this out there, but we could use decorators to fake a runtime interface -- as long as that object that declared the interface was only define once, I think the "references" / "usages" aspect of today's tooling would still work.

chriskrycho:
Much of this discussion has me wondering if it's not worth taking a step back and reconsidering the shape of this entirely. It's worth remembering that dependency injection and particularly DI *containers* are a specific way of implementing the much more general *pattern* of inversion of control. However, a commitment to this specific kind of DI container is *not* a hard necessity.

Functional languages with currying and partial application as native language constructs (F♯, Elm, etc.) tend to solve this simply by passing arguments which must conform to a given interface. And there are similar patterns available even in more traditional "OO" languages where the arguments are passed to the constructor.

That's not to say that we should switch to an FP style, _per se_—that would be a fairly radical shift in the programming model, and while I'd like it personally, it may or may not be the best migration path here! My point is simpler: if we're evaluating what the future of our inversion-of-control pattern should look like at this level, perhaps it's worth taking a more fundamental look at the options on the table for idiomatic JavaScript and thinking about how we might make a great developer experience there. A good *migration path* is not the same as an API that is effectively just a superset of today's API—that's just *one* possible path (if often a good one)!

pzuraq:
I had a long conversation with @runspired about this RFC and the future of Ember Data last week, and his concerns mirrored @simonihmig's concerns. We also came to the conclusion that what this RFC is after, ultimately, is to formalize services around interfaces. This is also how most major DI frameworks work in typed languages, so it makes sense that everyone kind of wants to be able to do this. Since we don't have interfaces in JS, classes are the next best thing, and abstract classes are the closest real pattern that we could use to emulate interfaces.

Another important point is that Ember Data is internally going to be moving a direction where True DI (@simonihmig's use case 2) is much more common, even required. Data will define interfaces rather than concrete services in the future, and various _other_ addons will implement those interfaces in their own services. The ability to swap out implementations like this for a particular service is very valuable for an _ecosystem_, since it means that other addons and libraries can rely on the existence of a particular service for a particular key, declaratively.

The biggest concern we ended up having with using a class as the key is, even if it were an abstract class, tooling would link users to _that_ class even if a different class were the selected implementation, and that could be very confusing. The Ember language server might be able to solve this with some clever redirects, but we also have to consider environments that won't be able to run the language server, such as Github (since it recently implemented go-to-definition).

We did land on a couple of deficiencies of the current system though, which explicit service injections would solve:

1. The current system is intimately tied to the build process. If a service is provided by an addon, there's no easy way to determine where it is coming from. If multiple services are provided for the same name, there's no easy way to determine which one will "win". You have to know intimate details of the build to figure these problems out, and so does the language server if we want to have tooling for it.

2. The most common and recommended way for addons to provide _private, internal_ services is through this merging and build logic. It's not uncommon for addons to need a private service, and it's not uncommon for _another_ addon to use this to override a those "private" services and extend them. 

Both of these problems could be addressed, even if we stuck with string based keys. Our strawman solution was to:

1. Require that all _global_ services be re-exported in the main consuming application's `/app/services` folder (or the equivalent in any future file layout). This would give users a single location to look for all of their app's services, along with a place to select which implementation they want for a particular piece (e.g. using a GraphQL implementation of a particular Ember Data service interface).

2. Have a method for addons to declare and use local, private services. This could be the existing syntax described in my previous post, or it could be a new syntax, but the main point is that these would not be leaked into the global namespace.

If we made these changes, then I think the main difference between strings and classes as keys would be DX around and support for go-to-definition. Strings would be supported via the language server, and simple to understand, but would not work in any environment by default (no Github support). Classes would work in most places, but could be confusing in cases where they're used to emulate interfaces. I still like the idea of being able to inject classes directly in the common case, but I can also see how it could add more complexity (or percieved complexity). I think ultimately, we should try to address these two concerns separately in different RFCs.

One last thought here: JavaScript doesn't have interfaces, but _TypeScript_ does. It may be possible for `ember-cli-typescript` to implement a babel transform that takes something like:

```ts
@service foo: MyServiceInterface;
```

And transforms it into:

```ts
@service('my-service') foo;
```

Making the ability to "inject interfaces" a progressive enhancement of TS that doesn't impact JS APIs directly. How we would associate a string name with an interface would be tricky, but maybe this is a way we could get the best of both worlds.

runspired:
@simonihmig and @pzuraq collectively have done a very good job summarizing my positions.

I fully agree with @pzuraq that there are things we should do to tighten up the existing expectations such that knowing where a service in use was declared is quickly discoverable, and that doing so is separate from changing the nature of the DI system (which this RFC would do largely by accident) or the syntax of using the DI system.

One thing I'd note on the value of strings-v-classes as keys that I think got glossed over is that were we to cleanup the expectations around where services can originate from then when you don't have a language server it is still quick to find the source (look in the services directory for the service of that name, if it's a re-export you can go-to-source from there). E.g. if we need a language server for either pattern (which we would) I feel strings come with the additional benefit of conveying the location to go look for the concrete implementation when that language server is not present.

NullVoxPopuli:
I mean, really all _I_ want is control+click. With strings, we just need to make up our own tooling, which sounds like more work. :-\

runspired:
I had this discussion with @NullVoxPopuli out of band, but wish to summarize it here.

> I mean, really all I want is control+click. With strings, we need to make up our own tooling, which sounds like more work. :-\

Moving to using classes as keys would not prevent us from using our own tooling nor save us any work.  In either case we end up needing a language-server to ensure that control+click works.

For strings, we have to map strings to a class.
For classes, we have to map classes to other classes.

This remapping from class A to class B is likely to cause users a large amount of confusion. (Just imagine a new user being told that to "use service B" you must "import service A").

**Why would this be?**

Let's run through an example of some pitfalls with classes as keys.

Imagine the addon `GoodCommunicator` provides the service called `central-hub`, and in several of it's own other services or components it consumes that service.

```ts
export default class CentralHubService {
  dispatch() {}
  subscribe() {}
}
```

Let's say that `GoodCommunicator` is such a great addon that many other addons also try to coordinate some behaviors via the `central-hub`.

```ts
import CentralHubService from 'good-communicator/services/central-hub';

export default MyComponent extends Component {
  @service(CentralHubService) centralHub;
}
```

Still another addon`GoodCommunicatorPlus` realizes that the `central-hub` could use an upgraded protocol for `dispatch`, so it imports and extends the original class.

```ts
import CentralHubService from 'good-communicator/services/central-hub';

export default AdvancedCentralHubService extends CentralHubService {
   dispatch() {
      if (...) {
        // ....
      } else {
        return super.dispatch(...arguments);
      }
   }
}
```

And it (or another addon) uses this version of the service class:

```ts
import AdvancedCentralHubService from 'good-communicator-plus/services/central-hub';

export default MyComponent extends Component {
  @service(AdvancedCentralHubService) centralHub;
}
```

Now finally, a consuming application consumes `GoodCommunicatorPlus` as an addon and it decides to use the service, so it adds it to its services, adding in some config.

```ts
import CentralHubService from 'good-communicator-plus/services/central-hub';
import config from './hub-config';

export default ConfiguredCentralHubService extends CentralHubService {
   config = config;
}
``` 

The situation we are now in is that we have 3 distinct classes that all are serving as keys to what is intended to be the same service.

If the consuming application does nothing, then we would instantiate three separate services, two of which would be missing the desired configuration.

If however the consuming application wanted to ensure that we continued to only have a single service, it would need to remap the keys. Something like:

```ts
import CentralHubService from 'good-communicator/services/central-hub';
import AdvancedCentralHubService from 'good-communicator-plus/services/central-hub';
import ConfiguredHubService from '../services/central-hub';

export function initialize(application) {
  application.unregister(CentralHubService);
  application.unregister(AdvancedCentralHubService);
  application.register(CentralHubService, ConfiguredHubService);
  application.register(AdvancedCentralHubService, ConfiguredHubService);
};

export default {
  initialize
};
```

With this remapping complete, we would now be back to a single "central-hub" service.

But there remains a problem: the class we've used as a key in so many places is not the class that we've instantiated. Any "go-to-definition" features we take advantage of out of the box are going to send us to the wrong class for 2 of the 3 potential keys in this example. And remember, there's nothing preventing an application from using all 3 key variants (the two from addons and it's own).

**In parting, a final concern:** 

This sort of keying by class is also problematic if someone desires multiple instances of a given service. By using the class as the key, we lose the ability to re-use the same class multiple times if desired.

pzuraq:
@runspired I think your examples are exaggerating the issue a bit, as they don't use the API the way it is described by the RFC. If we were to move forward with this RFC, in order for _any_ addon to override a service, it would have to manually reregister that service, so the addon that provides the `AdvancedCentralHubService` would require users to do something like this:

```js
import CentralHubService from 'good-communicator/services/central-hub';
import AdvancedCentralHubService from 'good-communicator-plus/services/central-hub';

export function initialize(application) {
  application.register(CentralHubService, AdvancedCentralHubService);
};

export default {
  initialize
};
```

And the addon itself would inject the `CentralHubService`:

```
// good-communicator-plus/components/my-component.js
import CentralHubService from 'good-communicator/services/central-hub';

export default MyComponent extends Component {
  @service(CentralHubService) centralHub;
}
```

I'm not particularly worried about this pattern being difficult or smelly, because IMO it _should_ be. Attempting to override a service from another addon like this should require some manual steps, because it is generally an antipattern - the functionality should be contributed up stream, ideally, or contained in a separate service that the downstream addon has full control over.

Now, your second example is a bit more where my concerns come from as well, because it isn't uncommon to have a service that should be configured, or an interface for a service that should be provided by the consuming application, and Ember Data is moving this direction so it'll become even more common. In these cases, it's unavoidable to have a different implementation, and this alternate implementation would need to be _registered_, and then tooling would have to _know_ about that registration so it could Control+Click to the proper definition. While we could absolutely do this, we no longer get it for "free", like this RFC was hoping for.

I think if we were providing interfaces, this would be a more logical leap. Users would go-to the definition, see it was an interface, and then know they had to find the proper implementation. I agree that taking them to a class, even an "abstract class", probably isn't enough of a hint that there could be a _subclass_ being used, and could result in a lot of confusion.

sukima:
Disclaimer: I haven't read this yet, But I did have a related yet independent idea.

What is all service classes inherited a **static** `inject` method which could look up the singleton via a `Map`/`WeakMap` with the class itself being the key?

```js
import MyService from 'my-app/services/my-service';
…

  @MyService.inject myService;
}
```

wagenet:
Where does this stand?

NullVoxPopuli:
@chancancode has a user-land addon that explores this space. :thinking: 

and @wycats has some thoughts, too.

There is also: https://github.com/emberjs/ember.js/issues/20095
- where I said I'd try to prototype this out soon (though, in that comment, I said not :tm:, but it's def :tm: at this point) -- I got distracted

wagenet:
This is being moved to the Exploring state.

lifeart:
While reading https://netbasal.com/lazy-load-services-in-angular-bcf8eae406c8 

I think we could have something like this https://gist.github.com/lifeart/fbcc7bd8747562aa85d79b42ca991493

In short, all properties on lazy service may be `promisified` by default, with ability to call `toSync` to get "reference" service object.

thanks to @NullVoxPopuli navigating me here.

Some version of provided gist:
```ts
class Bar {
    doSomething() {
        console.log('do something');
    }
    name: string;
}

type PromisifyProps<T> = {
    [P in keyof T]: T[P] extends (...args: infer A) => infer R ? (...args: A) => Promise<R> : Promise<T[P]>;
};

// a function to accept service load using import
// and it should return same type as service but all it's methods and properties should be promises
// we can use this function to make lazy loading of services
// under the hood we use proxy to make all methods and properties to be promises
function lazyService<T extends object>(service: () => Promise<T>): PromisifyProps<T> & {
    toSync(): Promise<T>;
} {
  let loadedService: T;
  const proxy = new Proxy({}, {
    get(_, prop) {
      return new Promise(async (resolve, reject) => {
        if (!loadedService) {
          try {
            loadedService = await service();
          } catch(e) {
            reject(e);
          }
        }
        if (prop === 'toSync') {
            return resolve(loadedService);
        }
        const value = Reflect.get(loadedService, prop);
        if (typeof value === 'function') {
          return resolve((...args: any[]) => Promise.resolve(value.apply(loadedService, args)));
        } else resolve(value);
      });
    },
  });

  return proxy as PromisifyProps<T> & {
    toSync(): Promise<T>;
  }
}

class Foo {
    bar = lazyService<Bar>(() => import('./bar'));
    async onClick() {
        // auto-load and invoke service method
        await this.bar.doSomething();
        // get service property
        const name = await this.bar.name;

        // convert async service to sync (auto-load)
        const sync = await this.bar.toSync();
        const secondName  = sync.name;
        
    }
}
```

ef4:
This was discussed at the spec meeting this week, the main open point of discussion was wether we want to keep the decorator syntax despite typescript still not inferring it, or use a field assignment syntax instead that would infer correctly (with an explicit `this`):

```js
  cookie = service(this, CookieService);
```

chancancode:
Expanding on @ef4's comment above:

Today, service lookups are keyed on the tuple `(owner, "string-key")`. This means that:

1. we can't easily tell from the import graph which services are needed by what modules, thus we have to eagerly load/register all services upfront
2. in order to provide services that are scoped to engines (and that is the only kind of scoping we provide today), we have to create a "child owner" for each engine (the first part of that key tuple)

In Polaris, we plan to solve both of these problems by changing the lookup key to `(owner, SomeJavaScriptValue)`.

1. in order to inject the service, you would have to acquire it somehow, typically by importing the module where the service is defined
2. standard JavaScript features allows you to control the scoping of the service, for example:
    1. un-exported values can only be used locally with the module
    2. `exports` map in `package.json` allows you to keep the value private within the package
    3.  packages can ensure they share (or not) services with the appropriate `peerDependencies` relationship
    4. you can always register that as a global, pass it as argument, `{{yield}}` it to child components, etc

In terms of practical day-to-day usages, it will probably look something like this:

```js
// app/services/session.js
import { setOwner } from "@ember/owner";
import { tracked } from "@glimmer/tracking";

export default class SessionService {
  // Alternatively, we can still provide a superclass in the framework
  // to deal with this boilerplate, just have to be a different import
  // and not subclassing from `Ember.Object`
  constructor(owner) {
    setOwner(this, owner);
  }

  @tracked currentUser = null;
}
```

```js
// app/components/menu.js
import { service } from "@ember/service";
import Component from "@glimmer/component";
import Session from "my-app/services/session";

export default class Menu extends Component {
  @service(Session) accessor session;
}
```

...or...

```js
// app/components/menu.js
import { service } from "@ember/service";
import Component from "@glimmer/component";
import Session from "my-app/services/session";

export default class Menu extends Component {
  // About the same number of characters but possibly work better with TypeScript today
  session = service(Session, this);
}
```

For the primitives, we will need:

```ts
// The definition of "Owner" in this context is different/more relaxed than the traditional one
// We only need it to be a WeakMap-key-able, that should probably be rectified across the framework
export type Owner = object;

export type ServiceKey<T> = /* ...defined later... */;
export type ServiceInstanceType<K extends ServiceKey<any>> = /* ...defined later... */;

type InstantiatedServices = WeakMap<ServiceKey<any>, ServiceInstanceType<ServiceKey<any>>>;
const Services = new WeakMap<Owner, InstantiatedServices>;

export function lookupService<K extends ServiceKey<any>>(owner: Owner, key: K): ServiceInstanceType<K> {
  let services = servicesFor(owner);
  let service: ServiceInstanceType<K> | undefined = services.get(key);

  if (service === undefined) {
    service = instantiate(owner, key);
    services.set(key, service);
  }

  return service!;
}

function servicesFor(owner: Owner): InstantiatedServices {
  let map = Services.get(owner);
  
  if (map === undefined) {
    map = new WeakMap();
    Services.set(owner, map);
  }

  return map;
}

function instantiate<K extends ServiceKey<any>>(owner: Owner, key: K): ServiceInstanceType<K> {
  /* ...defined later... */
}
```

On top of which we can build the convenience API.

From the primitive's perspective, we can accept any WeakMap-key-able (`object`) as a "service key". However, we need to know how to instantiate that key into the service instance. Usually, we would have some kind of manager for this purpose. On the other hand, we don't really care what the instance type is – for example, `boolean` as a service (`service(isMobile, this)` service) works just fine.

ef4:
Linking to @chancancode's addon which explores this design and has some discussion issues around open questions: https://github.com/chancancode/ember-polaris-service

text/0502-explicit-dependency-injection.md
```diff
@@ -0,0 +1,688 @@
...
+    }
+
+    export default { initialize };
+    ```
+
+    A key can be a pure contract with no runtime footprint at all.
+
+#### the hierarchy check
+
+Logic will be added to the register method to ensure that the lookup type either is the same as the service instance's type or is an ancestor type. This will prevent the ability to register unrelated classes that would break the implied class hierarchy that is assumed with dependency injection.
```
les2, line 305: there may be cases where people do not want or care about static typing and would like to rely on "duck typing" -- any object that responds to the "quack()" message shall be a "duck".

should the validation be configurable globally or per registration? or is it per lookup?

e.g., in testing, i might inject a `sinon.mock()` instance that is programmed to respond to a subset of methods and i might not use the abstract base class at all.

NullVoxPopuli: duck typing specifically I think has to be a separate RFC, because achieving that in a language that doesn't support such a thing will take an RFC _as long_ as this one.

I've called this "looking up by shape" elsewhere in this existing RFC

>  programmed to respond to a subset of methods and i might not use the abstract base class at all.

that is fine as long as the key is the same, and/or the key-hierarchy is the same. the keys are independent of the implementation (outside of initial instantiation)

NullVoxPopuli:
Some implementation / example demos here: https://github.com/NullVoxPopuli/ember-rfc-502-playground/tree/main/app/demos

BoussonKarel:
When I click on the **Rendered** link of this issue, it's throwing a 404:
image

runspired:
> Important design issues uncovered by experimentation work:

something I'd note - no referential token intended to be used by multiple packages is safe. Even same version, if a dep is ever duplicated you have a different token. There are only 2 exceptions to this: `Symbol.for` (not Symbol) and strings. Also note that `Symbol.for` is not safe from a typescript perspective - only strings satisfy the unique reference + unique type constraint if you care about the potential for code duplication.

We hit this *a lot* in WarpDrive, and ended up writing special utils to handle defining and retrieving references for objects, strings and symbols. (For objects, we keep a global registry with a string key, first reference registered wins).

NullVoxPopuli:
for all situations where you would have duplicates, you  fall in to two categories
- the project's deps are messed up (either peer, externals, or package manager issue)
- you _do_ want separate copies of services, because perhaps you have multiple majors / versions of a dependency (and thus the service it provides), and the code for those multiple dependencies is only written for the specific version of the service that was published with its major. if we were to use strings, whichever copy of the dep is out of date would fail with runtime errors

so, I dare argue the opposite: referential keys gets you more stability, even if it means you don't always get the same copy of a service that every copy of a dependency has (the problem is the duplicate deps)

runspired:
> the project's deps are messed up (either peer, externals, or package manager issue)

this is aspirational but no package manager today is capable of creating accurate dependency trees (isolated or hoisted) without using duplicate copies of the same version of a package.

I'm not necessarily arguing for strings here - I'm just pointing out where things are likely to get a little rough. The ember ecosystem tends to do something I have not seen in any other ecosystem so far - have community packages that depend on other community packages as peers as part of one global ecosystem. That makes us far more susceptible to package reference issues imo than say tanstack or millionjs deps that tend to pull from their own ecosystem only.

NullVoxPopuli:
thankfully, what I'm proposing here does not get rid of or discourage the use of string keys -- as that's the existing behavior.

If folks want to keep using the old way, that's fine. 

This RFC is for net new, and since the string lookup implementation is like.. 2 lines of code, I see no reason it can't stick around


Folks today can already use private services, and services-by-reference if they wish (I do this with utilities that ship as optional imports in ember-primitives (but also the code to do it without a dep is very small))

I still maintain that string services should be the exception for cases where users can't get their deps correct, and I _am_ proposing the default guides materials be updated to use non-string keys

NullVoxPopuli:
TODO for me:
- PR to the guides for how to manage state for library authors, and app authors a like
  - these will not use services
  - touch an the different lifetimes
    - explain how to wire up an instance to the container, if needed
    - explain how to wire up destruction, if needed
    - wire up to the container -> singleton
    - wire up to the module -> singleton, but be careful (also might be what you want)
    - wire up to a component -> unit-testable state, not singleton
    - how to idempotent

gossi:
Hmm, I had some concerns about the names. It took me a while, since `@service` didn't feel right anymore and @NullVoxPopuli demo shows, this is now about any reference.

API wise I think, this is a better fit:

```ts
import { lookup, register, inject } from '@ember/di';
```

potentially speaking (I know this goes to far) but to play the surface:

```ts
import { getContainer, type Container } from '@ember/di';
import { doBusinessLogic } from '#business-logic';

class MyComponent {

  doSomeBusinessLogic = (someObject: object) => {
    const container = getContainer(this);

    doBusinessLogic(someObject, container);
  }
}
```

at which point one can realize the DI system is no longer a necessity for ember to function, people can choose something externally (I ignored older setups, I know).

I wanna project the idea here, to _recommend_ something in the wider ecosystem, than to maintain our own, which binds capacity. Or to seek agreement/common standard as they did for [StandardSchema](https://standardschema.dev/).

PS. I still know this from my php times with [PSR-11: Container interface](https://www.php-fig.org/psr/psr-11/)

NullVoxPopuli:
I took a stab at some docs: https://github.com/ember-learn/guides-source/pull/2237
(which, @gossi cover the scenario you described)

johanrd:
Some experience from a mid-size app (vite/gts/ ~15 app services) that spiked moving some app state out of services to module scope. The main goal was **direct access from templates and plain-function helpers**.

• Re `service(this, X)` and `@service(X)`: **template-only components and plain-function helpers have no `this`**. Reaching a service from templates or helpers today takes a class-based helper or a custom wrapper per helper (we used ember-serviceable-helper), which also forced `helper.call(this, …)` from ts. Moving to module state was very tempting for this reason alone, as very few of our services really used an owner. `ember-intl` in particular was needed heavily outside templates (exporters, chart options, dialogs) – but very few use cases of `ember-intl` itself really requires an owner? (none in our case). 
• On the one-off needs in templates: Maybe something like a template helper `{{#let (service X)` could bridge the need.
• Boot wiring became explicit calls (`registerIntl`, `attach(apolloClient)`) instead of resolution. Test isolation is now manual, each state module needs a `reset()` called from shared test setup.

PR #507, opened 2019-06-22
v2 Addon Format (Embroider compatibility)

ef4:
[Rendered](https://github.com/emberjs/rfcs/blob/embroider/text/0000-embroider-v2-package-format.md).

### DIFF
--- a/text/0000-embroider-v2-package-format.md
+++ b/text/0000-embroider-v2-package-format.md
@@ -138,6 +138,14 @@ Modules in **Own Javascript** are allowed to import template files. This is comm
 
 Modules in **Own Javascript** are allowed to import JSON files. This is supported in Node and it's enabled by default in popular Javascript bundlers like Webpack, so it's common to encounter code on NPM that expects it to work. We interpret JSON as an ECMA module whose `default` export is the JSON object.
 
+### Own Javascript: Co-located Component Templates
+
+Modules in **Own Javascript** get no _automatic_ component template co-location. They should use Ember's `setComponentTemplate` explicitly. They can access their compiled templates either by importing them or by inlining them via the `hbs` macro provided by ember-cli-htmlbars.
+
+We're assuming that template co-location will be handled by the build step _before_ an addon is published to NPM in V2 format.
+
+Similarly, template-only components must be explicitly present as Javascript modules that invoke Ember's `templateOnlyComponent()` and `setComponentTemplate`.
+
 ### Own Javascript: Transpilation of imported modules
 
 Any module you import, whether from an Ember package or a non-Ember package, gets processed by the app's babel configuration by default. This ensures that the app's `config/targets.js` will always be respected and you won't accidentally break your older supported browsers by importing a dependency that uses newer ECMA features.
--- a/text/0000-embroider-v2-package-format.md
+++ b/text/0000-embroider-v2-package-format.md
@@ -416,7 +416,7 @@ importSync("some-dependency").default;
 
 `importSync` exists to do a thing that standard Javascript does not do: synchronous dynamic import. Ember historically needs synchronous dynamic import (it's what our runtime AMD `require` does). Until some future date at which Ember has migrated away from synchronous module resolution we need `importSync`.
 
-`importSync` is defined as behaving exactly like the standards-compliant `import` except instead of returning `Promise<Module>` it returns `Module`.
+`importSync` is defined as behaving exactly like the standards-compliant `import()` except instead of returning `Promise<Module>` it returns `Module`.
 
 In this RFC we don't state explicitly what `importSync` compiles to. It compiles to whatever the Javascript bundler we're using supports in order to achieve synchronous dynamic import. For example, if we're internally using Webpack we can compile `importSync("something")` to `require("something")`, because Webpack supports CommonJS `require` anywhere.
 
@@ -444,7 +444,7 @@ const myConfig = { coolFeature: true };
 // assuming your ownConfig is `{ coolFeature: true }`.
 ```
 
-Since `getOwnConfig` accesses the output of your `configure` build-hook, you have a place to run arbitrary
+Since `getOwnConfig` accesses the output of your `configure` build-hook, you have a place to run arbitrary build-time code and inject the results into runtime Javascript.
 
 ### Javascript Macro: getConfig
 
@@ -463,7 +463,7 @@ Instead of actually doing the stripping, the ember-test-selector plugin would co
 
 ### JavaScript Macro: macroCondition
 
-`macroCondition` acts like a function that takes a boolean value and returns that same boolean value. But whenever `macroCondition` appears directly inside the predicate of an `if` statement or as the predicate of a ternary expression, it tells the macro system to do branch elimination based on the predicate. Here is an example that combines all three macros we've seen so far:
+`macroCondition` acts like a function that takes a boolean value and returns that same boolean value. But whenever `macroCondition` appears directly inside the predicate of an `if` statement or as the predicate of a ternary expression, it tells the macro system to do branch elimination based on the predicate. Here is an example that combines three of the macros we've seen so far:
 
 ```js
 // This example:
@@ -499,6 +499,7 @@ It is a build error if `macroCondition` cannot statically determine the truth st
 Here is an example of `macroCondition` in a ternary expression:
 
 ```js
+// this example:
 const flavor = macroCondition(getOwnConfig().prefersChocolate)
   ? "chocolate"
   : "vanilla";
@@ -520,20 +521,21 @@ if (macroConditional(getOwnConfig().x)) {
 
 Q: Why not allow `if (getOwnConfig().thing)` instead of `if (macroCondition(getOwnConfig().thing))`?
 
-A: Because we don't want to leave any confusion over whether branch elimination will be done. Boolean expressions that include a macro like `getOwnConfig` alongside other runtime-only values are perfectly legal. But those expressions would not allow branch elimination. The ambiguity means you might accidentally defeat branch elimination without noticing. `macroCondition` is intended to signal -- both to the reader and to the compile -- that this place absolutely _must_ do branch elimination. It's an error if we can't eliminate one branch or the other.
+A: Because we don't want to leave any confusion over whether branch elimination will be done. Boolean expressions that include a macro like `getOwnConfig` alongside other runtime-only values are perfectly legal. But those expressions would not allow branch elimination. The ambiguity means you might accidentally defeat branch elimination without noticing. `macroCondition` is intended to signal -- both to the reader and to the compiler -- that this place absolutely _must_ do branch elimination. It's an error if we can't eliminate one branch or the other.
 
-### JavaScript macro: macroEach
+### JavaScript macro: each
 
-`macroEach` allows you to unroll a loop based on an array value provided by another macro. It behaves like the identity function (returning its argument unchanged), but it provided the special guarantee that if you use it as the argument of a `for ... of` loop, the loop will be unrolled:
+`each` allows you to unroll a loop based on an array value provided by another macro. It behaves like the identity function (returning its argument unchanged), but it provides the special guarantee that if you use it as the argument of a `for ... of` loop, the loop will be unrolled:
 
 ```js
-import { getOwnConfig, macroEach, importSync } from "@ember/macros";
+// This example:
+import { getOwnConfig, each, importSync } from "@ember/macros";
 let plugins = [];
-for (let plugin of macroEach(getOwnConfig().registeredPlugins)) {
+for (let plugin of each(getOwnConfig().registeredPlugins)) {
   plugins.push(importSync(plugin).default);
 }
 
-// Could compile to this, given OwnConfig
+// could compile to this, given OwnConfig
 // containing { registeredPlugins: ['@bigco/bar-chart', '@bigco/line-chart']}
 
 let plugins = [];
@@ -541,7 +543,7 @@ plugins.push(importSync("@bigco/bar-chart").default);
 plugins.push(importSync("@bigco/line-chart").default);
 ```
 
-It is a static error if the argument to `macroEach` does not evaluate to a statically known array.
+It is a static error if the argument to `each` does not evaluate to a statically known array.
 
 ### JavaScript Macro: moduleExists
 
--- a/text/0000-embroider-v2-package-format.md
+++ b/text/0000-embroider-v2-package-format.md
@@ -113,7 +113,7 @@ First, here’s the list of things a v2 package can provide. More detail on each
 
 The public `main` (as defined in `package.json`) of a v2 package points to its **Own Javascript**. The code is formatted as ES modules that follow the **Ember Language Standard** (which is defined later in this RFC). Addon authors can still author in whatever dialect they want, but they must transpile to Ember Language Standard before publishing to NPM.
 
-Templates are in hbs format. No custom AST transforms are supported. Addon authors can still use whatever custom AST transforms they want, but those transforms must have already been applied before publication to NPM.
+Templates are in hbs format. No custom AST transforms are supported. Addon authors can still use custom AST transforms, but those transforms must have already been applied and re-emitted into nhbs source text before publication to NPM. This means that only AST transforms whose output are serializable into hbs source text are permitted.
 
 Unlike v1 addons, there is no `/app` or `/addon` directory that is magically removed from the runtime paths to the modules. All resolution follows the prevailing Node rules.
 
--- a/text/0000-embroider-v2-package-format.md
+++ b/text/0000-embroider-v2-package-format.md
@@ -364,7 +364,7 @@ The `configureDependencies` hook is the _only_ way to disable child packages. Th
 ### Build Hook: skipBabel
 
 ```ts
-skipBabel({ package: string, semverRange?: string }[]): void;
+skipBabel(): { package: string, semverRange?: string }[]
 ```
 
 By default, all imported dependencies (and their recursive imported dependencies) go through the app's babel config. This ensures browser compatibility safety. However, we provide `skipBabel` as an opt-out to work around transpilation problems in cases where the developer has verified that transpilation of a given package isn't needed.
@@ -463,13 +463,15 @@ This is a low-level power tool. It's mostly useful as a compile target for custo
 
 Instead of actually doing the stripping, the ember-test-selector plugin would compile the user's code into code that uses `macroCondition` and `getConfig('ember-test-selectors')`. In this way, we get the powerful custom behavior, but only the standard core macros are needed at the time when the app itself is building.
 
+Another example is that a package like `fastboot` could expose config that says whether you're in Fastboot vs the browser, so that apps and addons could say `getConfig('fastboot').isFastboot`. Notice that for an addon to do this, it should declare `fastboot` as an optional peerDependency (see **Optional Peer Dependencies** section), because `getConfig` can only see your **Allowed Dependencies**.
+
 ### JavaScript Macro: macroCondition
 
 `macroCondition` acts like a function that takes a boolean value and returns that same boolean value. But whenever `macroCondition` appears directly inside the predicate of an `if` statement or as the predicate of a ternary expression, it tells the macro system to do branch elimination based on the predicate. Here is an example that combines three of the macros we've seen so far:
 
 ```js
 // This example:
-import { macroCondition, getOwnConfig, importSync } from '@ember/macros';
+import { macroCondition, getOwnConfig, importSync } from "@ember/macros";
 
 let implementation;
 
@@ -489,12 +491,12 @@ export default implementation;
 
 // ===============
 // Or compile down to this if OwnConfig contains { useNewVersion: false }
-let implementation
+let implementation;
 implementation = importSync("./old-component");
 export default implementation;
 ```
 
-It is a build error if `macroCondition` cannot statically determine the truth status of its argument.
+It is a build error if `macroCondition` cannot statically determine the truth status of its argument. It's a build error if `macroCondition` appears anywhere other than as the predicate of an if statement or ternary expression.
 
 `macroCondition` supports boolean logic, like `macroCondition(getOwnConfig().a && getOwnConfig().b)`.
 
@@ -562,6 +564,8 @@ if (macroCondition(moduleExists("ember-data"))) {
 ```
 
 Remember that you're always only allowed to `import` from your own **allowed dependencies**. So if an addon wants to optionally use another package only if that package is present, that package must be listed as an **Optional Peer Dependency**.
+
+`moduleExists` differs in one respect from all the other macros: there is no reliable build-time polyfill for it in non-Embroider builds. So when you use it in a non-Embroider build, you get a runtime implementation instead (equivalent to today's `require.has()`. If you need build-time branch elimination, try to use `dependencySatisfies` instead, which works correctly in both Embroider and non-Embroider builds.
 
 ### JavaScript Macro: dependencySatisfies
 
--- a/text/0000-embroider-v2-package-format.md
+++ b/text/0000-embroider-v2-package-format.md
@@ -134,7 +134,7 @@ Modules in **Own Javascript** are allowed to use dynamic `import()`, and the spe
 
 This is designed to allow limited pattern matching of possible targets for your dynamic `import()`. The rules ensure that we can always tell which NPM package you are talking about (the `@` rule covers namespaced NPM packages, so you can't depend on `@ember/${whatever}` but you can depend on `@ember/translations/${lang}`). If your pattern matches zero files, we consider that a static build error. (For more background on the thought-process behind this feature, see [Embroider #198: Allow dynamic imports to accept more dynamic inputs](https://github.com/embroider-build/embroider/issues/198).
 
-Modules in **Own Javascript** are allowed to import template files. This is common in today’s addons (they import their own layout to set it on their Component class). But in v2 packages, import specifiers of templates are required to explicitly include the `.hbs` extension.
+Modules in **Own Javascript** are allowed to import template files. This is common in today’s addons (they import their own layout to set it on their Component class). The `.hbs` file extension resolves with lower priority than `.js` (meaning if you have both, the `.js` wins).
 
 ### Own Javascript: Transpilation of imported modules
 
--- a/text/0000-embroider-v2-package-format.md
+++ b/text/0000-embroider-v2-package-format.md
@@ -1,0 +1,772 @@
+- Start Date: 2019-06-21
+- Relevant Team(s): Ember CLI, Ember.js, Learning
+- RFC PR:
+- Tracking:
+
+# v2 Addon Format (Embroider Compatibility)
+
+## Summary
+
+This RFC defines a new package format that is designed to make Ember packages (meaning both apps and addons) statically analyzable and more compatible with the rest of the NPM & Javascript ecosystem. This RFC is the first step in stabilizing [Embroider](https://github.com/embroider-build/embroider) as our next-generation build system.
+
+## Motivation
+
+One of the good things about Ember is that apps and addons have a powerful set of build-time capabilities that allow lots of shared code with zero-to-no manual integration steps for the typical user. We have been doing “zero config” since before it was a cool buzzword (it was just called “convention over configuration”). And we’ve been broadly successful at maintaining very wide backward- and forward-compatibility for a large body of highly-rated community-maintained addons.
+
+But one of the challenging things about Ember is that our ecosystem’s build-time capabilities are more implementation-defined than spec-defined, and the implementation has accumulated capabilities organically while only rarely phasing out older patterns. I believe the lack of a clear, foundational, build-time public API specification is the fundamental underlying issue that efforts like the various packaging / packager RFCs have tried to work around.
+
+The benefits to users for this RFC (and Embroider in general) are:
+
+- faster builds and faster NPM installs
+- “zero-config import from NPM — both static and dynamic” as a first-class feature that works for both third-party libraries and Ember addons
+- support for arbitrary code splitting
+- tree-shaking of unused modules, components, helpers, etc from the app and all addons
+- a layered build system with clearly documented APIs between the layers, so it's easier to experiment and contribute
+- a build system that can take advantage of current and future investments by the wider Javascript ecosystem into code bundling & optimization.
+
+## Key Ideas
+
+### Fully Embrace ES Modules
+
+Ember was one of the earliest adopters of ECMAScript modules, and Ember core team members were directly involved in helping move that features through TC39. Ember’s early experiences with modules influenced the spec itself. _Yet we have lagged in truly embracing modules._
+
+For example, how do Ember apps express that they depend on a third-party library? The [app.import](https://ember-cli.com/user-guide/#javascript-assets) API. This should be ECMA standard `import`.
+
+Another way to state the problem is that apps and addons all _push_ whatever code they want into the final built app. Whereas ES modules can _pull_ each other into the build as needed.
+
+### Play nice with NPM Conventions
+
+The ECMA module spec by itself doesn’t try to define a module resolution algorithm. But the overwhelmingly most popular convention is the [node_modules resolution algorithm](https://nodejs.org/api/all.html#modules_all_together).
+
+Ember addons do respect node_module resolution for build-time code, but they do not respect it for runtime code. This is an unhelpful distinction.
+
+### Verbose, Static Javascript as a Compiler Target
+
+Ember’s strong conventions mean that many kinds of dependencies can be inferred (including _statically_ inferred) without requiring the developer to laboriously manage them. This is a good thing and I believe the current fad in the wider Javascript ecosystem for making developers hand-write verbose static imports for everything confuses the benefits of having static analysis (which is good) with the benefits of hand-managing those static imports (which is unnecessary cognitive load when you have clear conventions and a compiler).
+
+This design is about compiling today’s idiomatic Ember code into more “vanilla” patterns that leverage ES modules, node_modules resolution, and spec-compliant static and dynamic `import` to express the structure of an Ember application in a much more “vanilla Javascript” way.
+
+This compile step lets us separate the authoring format (which isn’t changing in any significant way in this RFC) from the packaging format (which can be more verbose and static than we would want in an authoring format).
+
+# Detailed design
+
+## Definitions
+
+**package**: every addon and app is a package. Usually synonymous with “NPM package”, but we also include in-repo packages. The most important fact about a package is that it’s often the boundary around code that comes from a particular author, team, or organization, so coordination across packages is a more sensitive design problem than coordination within apps.
+
+**app**: a package used at the root of a project.
+
+**addon**: a package not used at the root of a project. Will be an **allowed dependency** of either an **app** or an **addon**.
+
+**allowed dependency**: For **addons**, the **allowed dependencies** are the `dependencies` and `peerDependencies` in `package.json` plus any in-repo addons. For **apps**, the **allowed dependencies** are the `dependencies`, `peerDependencies`, and `devDependencies` in `package.json` plus any in-repo addons.
+
+**Ember package metadata**: the `ember-addon` section inside `package.json`. This already exists in v1, we’re going to extend it.
+
+**v2 package**: a package with `package.json` like:
+
+    "keywords": [ "ember-addon" ],
+    "ember-addon": {
+      "version": 2
+    }
+
+**v1 package**: a package with `package.json` like:
+
+    "keywords": [ "ember-addon" ]
+
+and no `version` key (or version key less than 2) in **Ember package metadata**.
+
+**non-Ember package**: a package without `keywords: ["ember-addon"]`
+
+## Scope of this RFC
+
+This RFC is intended as the base level spec for v2 Ember packages. **It does not attempt to cover everything a v1 package can do today**. For example, no provision is made in this RFC for:
+
+- providing dev middleware
+- providing commands and blueprints
+- preprocessing your parent package's code
+- modifying your parent package's babel config
+- injecting content into index.html (contentFor)
+
+It is understood that all of these are legitimate things for Ember addons to do. Defining these capabilities within v2 packages will be done in followup RFCs. It is simply too much scope to cover in one RFC.
+
+Because we're hyper-focused on backward- and forward-compatibility, there is no harm in progressively converting some addons to v2 (which provides immediate benefits in terms of build performance and reduced fragility under Embroider) while others need to stay as v1 until we offer the features they need.
+
+## Package Public API Overview
+
+The format we are about to describe _is a publication format_. Not necessarily an authoring format. By separating the two, we make it easier to evolve the authoring format without breaking ecosystem-wide compatibility. The publication format is deliberately more explicit and less dynamic that what we may want for an authoring format.
+
+First, here’s the list of things a v2 package can provide. More detail on each of these will follow:
+
+- **Own Javascript**: javascript and templates under the package’s own namespace (the v1 equivalent is `/addon/**/*.{js,hbs}/`)
+- **App Javascript**: javascript and templates that must be merged with the consuming app’s namespace (the v1 equivalent is `/app/**/*.{js,hbs}`). Other RFCs are working to move Ember away from needing this feature, but we are not gated on any of those and fully support **App Javascript**.
+- **CSS**: available for `@import` by other CSS files (both in the same package and across packages) and by ECMA `import` directives in Javascript modules (both in the same package and across packages).
+- **Implicit Dependencies**: scripts, modules, and stylesheets that should be implicitly included in the app or the app's tests whenever this addon is active. This is a backward-compatibility feature.
+- **Renaming Rules**: allow a package to declare that some of its modules should be available at different import paths than their real, resolvable path. This is a backward-compatibility feature that new addons should not use.
+- **Externals**: allows a package to declare that it imports some things that are not **allowed dependencies**. Instead of being resolved at build time, **externals** are deferred until runtime and get handled by the traditional `loader.js` `require()`.
+- **Assets**: any files that must be available in the final built application directory such that they have public URLs (typical examples are images and fonts).
+- **Build Hooks**: code that runs within Node at application build time. The v1 equivalent is an addon's `index.js` file.
+
+## Own Javascript
+
+The public `main` (as defined in `package.json`) of a v2 package points to its **Own Javascript**. The code is formatted as ES modules using ES latest features, meaning: stage 4 features only, unless otherwise explicitly mentioned elsewhere in this spec. (Addon authors can still use whatever custom syntax they want, but those babel plugins must run before publication to NPM.)
+
+Templates are in hbs format. No custom AST transforms are supported. (Addon authors can still use whatever custom AST transforms they want, but those transforms must have already been applied before publication to NPM.)
+
+Unlike v1 addons, there is no `/app` or `/addon` directory that is magically removed from the runtime paths to the modules. All resolution follows the prevailing Node rules.
+
+The benefit of this design is that it makes our packages understandable by a broader set of tooling. Editors and build tools can follow `import` statements across packages and end up in the right place.
+
+In v1 packages, `main` usually points to a build-time configuration file. That file is moving and will be described in the **Build Hooks** section below.
+
+### Own Javascript: Imports
+
+Modules in **Own Javascript** are allowed to use ECMA static `import` to resolve any **allowed dependency**, causing it to be included in the build whenever the importing module is included. This replaces `app.import`. Resolution follows prevailing Node rules. (This usually means the node_modules algorithm, but it could also mean Yarn PnP. The difference shouldn't matter if you are correctly declaring all your **allowed dependencies**.)
+
+Notice that a package’s **allowed dependencies** do not include the package itself. This is consistent with how Node resolution works. To import files from within your own package you must use relative paths. This is different from how run-time AMD module resolution has historically worked in Ember Apps. (`@embroider/compat` implements automatic adjustment for this case when compiling from v1 to v2).
+
+Modules in **Own Javascript** are allowed to use dynamic `import()`, and the specifiers have the same meanings as in static import. However, we impose a restriction on what syntax is supported inside `import()`. The only supported syntax inside `import()` is:
+
+- a string-literal
+- or a template string literal
+  - that begins with a static part
+  - where the static part is required to contain at least one `/`
+  - or at least two `/` when the path starts with `@`.
+
+This is designed to allow limited pattern matching of possible targets for your dynamic `import()`. The rules ensure that we can always tell which NPM package you are talking about (the `@` rule covers namespaced NPM packages, so you can't depend on `@ember/${whatever}` but you can depend on `@ember/translations/${lang}`). If your pattern matches zero files, we consider that a static build error. (For more background on the thought-process behind this feature, see [Embroider #198: Allow dynamic imports to accept more dynamic inputs](https://github.com/embroider-build/embroider/issues/198).
+
+Modules in **Own Javascript** are allowed to import template files. This is common in today’s addons (they import their own layout to set it on their Component class). But in v2 packages, import specifiers of templates are required to explicitly include the `.hbs` extension.
+
+### Own Javascript: Transpilation of imported modules
+
+Any module you import, whether from an Ember package or a non-Ember package, gets processed by the app's babel configuration by default. This ensures that the app's `config/targets.js` will always be respected and you won't accidentally break your older supported browsers by importing a dependency that uses newer ECMA features.
+
+There is an explicit per-package opt-out for cases where you're _sure_ that transpilation is not needed and not desirable. (See **Build Hooks** for details on the `skipBabel` option.)
+
+## Own Javascript: Supported module formats for non-Ember packages
+
+As already stated, V2 Ember packages must contain only ES modules. However, non-Ember packages in your **allowed dependencies** are allowed to contain ES modules _or_ CommonJS modules. This provides the best compatibility with general-purpose NPM utilities.
+
+### Own Javascript: Macros
+
+The V2 format deliberately eliminates many sources of app-build-time dynamism from addons. Instead, we provide an equivalently-powerful macro system and consider it an always-supported language extension (the macros are always available to every V2 package, ambiently, and we promise to give them their faithful build-time semantics).
+
+See **Macro System** for the full details.
+
+## App Javascript
+
+To provide **App Javascript**, a package includes the `app-js` key in **Ember package metadata**. For example, to duplicate the behavior of v1 packages, you could say:
+
+    "ember-addon": {
+      "version": 2,
+      "app-js": "./app"
+    }
+
+Like the **Own Javascript**, templates are in place in hbs format with any AST transforms already applied. Javascript is in ES modules, using only ES latest features. ECMA static and dynamic imports from any **allowed dependency** are supported. (Even though the app javascript will be addressable within the _app's_ module namespace, your own imports still resolve relative to your addon.)
+
+By making `app-js` an explicit key in **Ember package metadata**, our publication format is more durable (you can rearrange the conventional directory structure in the future without breaking the format) and more performant (less filesystem traversal is required to decide whether the package is using the **App Javascript** feature.
+
+## CSS
+
+To provide **CSS**, a package can include any number of CSS files. These files can `@import` each other via relative paths, which will result in build-time inclusion (as already works in v1 packages).
+
+If any of the **Own Javascript** or **App Javascript** modules depend on the presence of a CSS file in the same package, it should say so explicitly via an ECMA relative import, like:
+
+    import '../css/some-component.css';
+
+This is interpreted as a build-time directive that ensures that before the Javascript module is evaluated, the CSS file's contents will be present in the DOM. ECMA import of CSS files must always include the explicit `.css` extension.
+
+> Q: Does this interfere with the ability to do CSS-in-JS style for people who like that?
+
+> A: No, because that would be a preprocessing step before publication. It’s a choice of authoring format, just like TypeScript or SCSS. CSS-in-JS people would compile all their things to ES modules before we deal with it.
+
+It is also possible for other packages (including the consuming application) to depend on a CSS file in any of its **allowed dependencies**, from either Javascript or CSS. From Javascript it looks like:
+
+    // This will resolve the `your-addon` package and find
+    // './some-component.css' relative to the package root.
+    // The .css file extension is mandatory
+    import 'your-addon/some-component.css';
+
+And from CSS it looks like:
+
+    @import 'your-addon/some-component';
+
+What about SCSS _et al_? You’re still free to use them as your authoring format, and they should be transpiled to CSS in your publication format. If you want to offer the original SCSS to consuming packages, you’re free to include it in the publication format too. Since we’re making all packages resolvable via normal node rules, it’s now dramatically easier to implement a preprocessor that supports inter-package dependencies. (The same logic applies to TypeScript.)
+
+## Implicit Dependencies
+
+Within **Ember package metadata** we support several flavors of implicit dependencies:
+
+- implicit-modules
+- implicit-scripts
+- implicit-styles
+- implicit-test-modules
+- implicit-test-scripts
+- implicit-test-styles
+
+Each one is a list of package-relative paths to files within the package.
+
+Whenever your package is active, all of its implicit dependencies will be included in the build. The `-test-` variants will be included only in test suites, and the non-`-test-` variants are always included.
+
+`implicit-modules` and `implicit-test-modules` mean that the app should be built as if someone has explicitly typed ECMA import statements for each of the listed modules.
+
+`implicit-scripts` and `implicit-test-scripts` are for Javascript in a script context. Script context is different from module context (as defined by the ECMA spec). This is how an addon can push things into the equivalent of the traditioanl `vendor.js`, which is in script context.
+
+`implicit-styles` and `implicit-test-styles` are for stylesheets. This is how an addon can push things into the equivalent of the traditional vendor.css.
+
+While **Implicit Dependencies** are a fully-supported part of the v2 spec, v2 packages are encouraged to use direct ECMA `import` or CSS `@import` instead, whenever possible. A direct `import` provides finer-grained dependency information: we know exactly _which_ module inside your package actually depends on the thing, rather than needing to assume that your entire package depends on it.
+
+For example, if one of your components depends on a third-party library, you should `import` that library directly from your component. Then the library will only be included if somebody uses that particular component. Whereas if you use `implicit-scripts`, the library will always be included, even if nobody uses the component that needs the library.
+
+## Renaming Rules
+
+V1 Addons have multiple ways (at least five that I've found so far!) of emitting modules that escape the addon's own package namespace. Examples:
+
+- `ember-lodash` remaps all of its modules to the package name `lodash`
+- `@ember/test-helpers` provides some modules under its true name, but also some modules under `ember-test-helpers`.
+
+In order for Embroider compile these packages to valid V2 packages, we give V2 packages the ability to express renaming rules using the following properties in **Ember package metadata**:
+
+- `renamed-packages`: a map from the package name a user would type to the real package name that provides it. Example:
+
+  ```js
+  {
+    "renamed-packages": {
+      "lodash": "ember-lodash"
+    }
+  }
+  ```
+
+- `renamed-modules`: a map from modules that a user may try to import to the real paths where those modules live:
+
+  ```js
+   "renamed-modules": {
+    "ember-test-helpers/index.js": "@ember/test-helpers/ember-test-helpers/index.js"
+   }
+  ```
+
+  When Embroider compiles a V1 package like `@ember/test-helpers` it ensures that the modules that would have "escaped" the package end up _inside_ the package, so that this kind of renaming works.
+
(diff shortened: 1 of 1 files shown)


NullVoxPopuli:
is it going to be flexible enough to handle a structure change? like maybe future src top level directory for addons and apps?

text/0000-embroider-v2-package-format.md
```diff
@@ -0,0 +1,772 @@
...
+
+The impact on addon authors is more significant. This design is fully backward compatible, and the intention is that all existing addons continue to work (some with worse compatibility hacks than others in the v1-to-v2 compiler). But there will be a demand for addons published in v2 format, since it is expected to result in faster build times. My prediction is that people who are motivated to get their own app build times down will send a lot of PRs to the addons they’re using.
+
+In many cases, converting addons to v2 makes them simpler. For example, today many addons use custom broccoli code to wrap third-party libraries in a fastboot guard that prevents the libraries from trying to load in Node (where they presumably don’t work). In v2, they can drop all that custom build-time code in favor of a macro-guarded `importSync`.
+
+This design does _not_ advocate loudly deprecating any v1 addon features. Doing that all at once would be unnecessarily disruptive. I would rather rely on the carrot of faster builds and Embroider stability than the stick of deprecation warnings. We can choose to deprecate v1 features in stages at a later time.
```
jenweber: Some additional points to cover in this section:

- Where to document this system (CLI guides)
- How to divide up the learning: general overveiw, change tutorial for new addons, write a tutorial for addon authors to help them migrate their addons, a tutorial for configuration that app users can reference if they run into problems with their addons, and abilities that unlock by having webpack compat. We would also need to revise some existing materials for regular ember-cli builds so that we appropriately present both options. (side note, what would we actually recommend as the happy path).

text/0000-embroider-v2-package-format.md
```diff
@@ -0,0 +1,772 @@
...
+The benefits to users for this RFC (and Embroider in general) are:
+
+- faster builds and faster NPM installs
+- “zero-config import from NPM — both static and dynamic” as a first-class feature that works for both third-party libraries and Ember addons
+- support for arbitrary code splitting
+- tree-shaking of unused modules, components, helpers, etc from the app and all addons
+- a layered build system with clearly documented APIs between the layers, so it's easier to experiment and contribute
+- a build system that can take advantage of current and future investments by the wider Javascript ecosystem into code bundling & optimization.
+
+## Key Ideas
```
jenweber: This section would benefit from a link to an existing, Embroidered Ember app or two (Octane + Super Rentals) and an addon (???). The necessary detail in this RFC makes it sound very complicated to use Embroider, while the median experience is pretty straightforward.

ef4:
> is it going to be flexible enough to handle a structure change? like maybe future src top level directory for addons and apps?

There are two levels of indirection that make it flexible. 

One is that this is a publication format, not an authoring format. But I don't think that's a totally satisfying answer, because there will be a natural desire over time to make the two conform more closely.

The other indirection is that no directories in this format are hard-coded. For example, to use a feature like **App Javascript**, there's no special directory. Instead there's a property in package.json (`ember-addon.app-js`) that you can point at any directory in the package.

But there *is* a hard rule here: whatever paths you publish to NPM in your package, those are the actual paths your users can import from. That is an area where too much flexibility has been bad for us, and we need to follow a clear standard.

text/0000-embroider-v2-package-format.md
```diff
@@ -0,0 +1,772 @@
...
+
+## Definitions
+
+**package**: every addon and app is a package. Usually synonymous with “NPM package”, but we also include in-repo packages. The most important fact about a package is that it’s often the boundary around code that comes from a particular author, team, or organization, so coordination across packages is a more sensitive design problem than coordination within apps.
+
+**app**: a package used at the root of a project.
+
+**addon**: a package not used at the root of a project. Will be an **allowed dependency** of either an **app** or an **addon**.
+
+**allowed dependency**: For **addons**, the **allowed dependencies** are the `dependencies` and `peerDependencies` in `package.json` plus any in-repo addons. For **apps**, the **allowed dependencies** are the `dependencies`, `peerDependencies`, and `devDependencies` in `package.json` plus any in-repo addons.
```
buschtoens: I recognize that we want to keep `devDependencies` as an **allowed dependency** for **apps**, since this is how it's always been conventionally for Ember **apps** and we want to reduce churn as much as possible.

Out of curiosity I would still like to understand, why it ever was this way. Is it just a side-effect of the fact that **apps** are always the root of any given project and thus will always have access to `devDependencies`?

I would argue that, while still officially supporting it, we should discourage adding runtime dependencies for **apps** to `devDependencies` instead of `dependendencies`, for the following reasons:

- It is confusing for Ember newbies.
- It is inconsistent.
- When I turn an **app** into an engine (an **addon**) or extract certain parts of an **app** into an **addon** (or engine), `devDependencies` suddenly turn into `dependencies`, just because the code was moved.

Instead, I would recommend that we encourage the same mental model for **apps** and **addons**. Everything that needs to be available at runtime / should influence the production build, has to be a `dependency`. Everything else, like testing or development support, linters, etc. are `devDependencies`.

What is your opinion regarding this?

If you agree, we could have an opt-in flag to enforce the same semantics for **apps** and **addons**.

chriskrycho: I strongly agree with this but would go a step further: if at all possible, I would very much like to see us get to a world where devDependencies and dependencies mean what they sound like: the latter is genuinely runtime-only, the former genuinely development- and build-time only.

It’s possible there are reasons that cannot be the case, in which case I’d love clarification there, because my understanding of the three-phase build pipeline for Embroider suggests that precisely that distinction should be possible. But I very well may be missing something!

ef4: Yes, for addons the distinction gets quite strong under this RFC. And it will reduce the number of packages needed by apps (count how many copies of ember-cli-htmlbars and ember-cli-babel you have).

For apps, I agree with:

>  we should discourage adding runtime dependencies for apps to devDependencies instead of dependencies

But most apps don't have any runtime dependencies. If you do `ember b`, and then `rm -rf node_modules`, and then serve `dist` to a browser, your app still runs. There are no runtime dependencies. Your addons may be used at runtime, but the actual addon NPM packages are only needed during the build.

The exception is Fastboot. Fastboot does imply having some runtime dependencies. Using `dependencies` for this would simplify and improve fastboot, which today tries to generate a whole separate package.json.

text/0000-embroider-v2-package-format.md
```diff
@@ -0,0 +1,812 @@
...
+
+```hbs
+<SomeComponent @config={{macroGetOwnConfig}} />
+{{! ⬆️compiles to ⬇️ }}
+<SomeComponent @config={{hash items=(array (hash score=42))}} />
+```
+
+### Handlebars Macro: macroCondition
+
+Used as a helper within a block `{{#if}}` or inline `{{if}}`. Just like the JS `macroCondition`, it ensures that branch elimination will happen.
```
jelhan: This is still a little unclear to me. While working on a macro-based build system for Ember Bootstrap to handle Bootstrap version-specific builds I faced some open questions:

- Can another template helper be used as an argument for `macroCondition` as long as the value could be determined statically? E.g. `{{if (macroCondition (and (macroGetOwnConfig "isFoo") (macroGetOwnConfig "isBar")))}}`
- Are multiple positional arguments supported by `macroCondition`? E.g. `{{if (macroCondition (macroGetOwnConfig "isFoo") (macroGetOwnConfig "isBar"))}}`.
  - If it's not supported, a build time error should be thrown in my opinion.
  - If it's supported, are they combined by _AND_ or by _OR_?

Using `macroCondition` together with other template helpers would allow domain-specific macros without requiring an AST transform. For example in Ember Bootstrap we could use something like:

```hbs
{{#if (macroCondition (has-bootstrap-version 3))}}
  {{! Bootstrap 3 only }}
{{/if}}
```

With a `{{has-bootstrap-version}}` template helper like this:

```js
import { helper } from "@ember/component/helper";
import { getConfig } from "@ember/macros"; 

function hasBootstrapVersion([version]) {
  return getOwnConfig('version') === version;
}

export default helper(hasBootstrapVersion);
```

I don't see such a strong use case for support of multiple conditions. It would require all arguments to be statically determinable and in that case they could also be combined beforehand. So instead of `{{macroCondition (getOwnConfig "isFoo") (getOwnConfig "isBar)}}` a user could just add a `isFooAndBar` value to the config. If other template helpers are allowed as arguments even an `{{and}}` or `{{or}}` template helper could be used. But I nevertheless see the need to clarify that point.

ef4: An arbitrary helper like `(has-bootstrap-version 3)` isn't going to work because we're not going to be able to run it at build time. But the intention is that you would instead put whatever code you want to run at build time in your `configure` build hook:

```js
configure() {
  let hasBootstrap3 = figureThisOutHowever();
  return { hasBootstrap3 };
}
```

And use the macro like:

```hbs
{{#if (macroCondition (macroGetOwnConfig "hasBootstrap3"))}}
  <UseSomeBootstrap3 />
{{/if}}
```

ef4: As for boolean combinations, if Ember standardizes `and`, `or`, and `not` helpers I think it would be simple to support them here. But that is not anticipated in this RFC.

(The equivalent JS macro *does* support boolean operators.)

text/0000-embroider-v2-package-format.md
```diff
@@ -0,0 +1,817 @@
...
+```js
+import { dependencySatisfies, macroCondition, failBuild } from "@ember/macros";
+if (macroCondition(dependencySatisfies("ember-data", "^3.0.0"))) {
+  // include code here for ember-data 3.0 compat
+} else {
+  failBuild(`We don't support ember-data versions other than ^3.0.0`);
+}
+```
+
+### Handlebars Macro: macroGetOwnConfig
```
simonihmig: The Handlebars macro `macroGetConfig` is not mentioned in the RFC, I assume by mistake!?

text/0000-embroider-v2-package-format.md
```diff
@@ -0,0 +1,817 @@
...
+```
+
+It can be placed on both HTML elements and angle bracket component invocations.
+
+### Handlebars Macro: macroFailBuild
+
+Like the JS `failBuild` macro.
+
+```hbs
+{{#if (macroCondition (dependencySatisfies "some-peer-dep" "^3.0.0")) }}
```
simonihmig:
```suggestion
{{#if (macroCondition (macroDependencySatisfies "some-peer-dep" "^3.0.0")) }}
```

text/0000-embroider-v2-package-format.md
```diff
@@ -0,0 +1,817 @@
...
+There is one place where `{{#if}}` doesn't work: within "element space". If you want to _sometimes_ set an attribute, but sometimes not, this doesn't work:
+
+```hbs
+<div {{#if this.testing}} data-test-target={{@id}} {{/if}} />
+```
+
+`macroMaybeAttrs` exists to conditionally compile away attributes and arguments out of element space:
+
+```hbs
+<div {{macroMaybeAttrs (getConfig "ember-test-selectors" "enabled") data-test-target=@id }} />
```
simonihmig:
```suggestion
<div {{macroMaybeAttrs (macroGetConfig "ember-test-selectors" "enabled") data-test-target=@id }} />
```

rwjblue:
We discussed this in todays core team meeting, and are moving this into final comment period. On guard 🤺

text/0000-embroider-v2-package-format.md
```diff
@@ -0,0 +1,817 @@
...
+
+It is no longer the `main` entrypoint of the package (see **Own Javascript**). Instead, it’s located via the `build` key in **Ember package metadata**, which should point at a Javascript file. `build` is optional — if you don’t have anything to say, you don’t need the file.
+
+It is now an ECMA module, not a CJS file. The default export is a class that implements your build hooks (there is no required base class).
+
+Here is a list of build hooks, each of which will have its own section below.
+
+- configure
+- configureDependencies
+- skipBabel
```
simonihmig, line 319: With the above hooks only, addons that want to plugin into the build process, to process files owned by the *app*, so obviously not available at publish-time (of the addon), seem to not be possible in v2 format (yet). Things like `ember-cli-sass` for example. These use cases are currently supported by hooks like [`postprocessTree`](https://ember-cli.com/api/classes/Addon.html#method_postprocessTree) or [`setupPreprocessorRegistry`](https://ember-cli.com/api/classes/Addon.html#method_setupPreprocessorRegistry). 

Maybe the RFC should state more clearly the way forward here:
* are the three hooks mentioned above the only ones supported for v2 addons (for now), so with this RFC not all v1 addons are migratable to v2 (basically only addons that can implement their features at publish-time, like addons exposing new run-time parts (e.g. components) to the consumer
* which will be extended further by following RFCs? (this seems what the "Supporting References" chapter suggests, but this was easy to miss)
* or are the other v1 hooks still supported in a v2 addon for now, something which the original [SPEC draft](https://github.com/embroider-build/embroider/blob/master/SPEC.md) goes into more detail.

text/0000-embroider-v2-package-format.md
```diff
@@ -0,0 +1,817 @@
...
+- **Renaming Rules**: allow a package to declare that some of its modules should be available at different import paths than their real, resolvable path. This is a backward-compatibility feature that new addons should not use.
+- **Externals**: allows a package to declare that it imports some things that are not **allowed dependencies**. Instead of being resolved at build time, **externals** are deferred until runtime and get handled by the traditional `loader.js` `require()`.
+- **Assets**: any files that must be available in the final built application directory such that they have public URLs (typical examples are images and fonts).
+- **Build Hooks**: code that runs within Node at application build time. The v1 equivalent is an addon's `index.js` file.
+
+## Own Javascript
+
+The public `main` (as defined in `package.json`) of a v2 package points to its **Own Javascript**. The code is formatted as ES modules using ES latest features, meaning: stage 4 features only, unless otherwise explicitly mentioned elsewhere in this spec. (Addon authors can still use whatever custom syntax they want, but those babel plugins must run before publication to NPM.)
```
simonihmig, line 112: Given that decorators are on the one hand still not stable (i.e. stage 2), but so pivotal already in the Ember Octane (and before) programming model, shouldn't we treat them as an exception to the rule here, and guarantee v2 addons that they will be transpiled, so they don't have to bother setting up a prepublish build step just because of the use of decorators? (which I believe is the only non-standard language feature broadly in use)

text/0000-embroider-v2-package-format.md
```diff
@@ -0,0 +1,817 @@
...
+
+# How we Teach This
+
+This RFC should have no direct impact on what app authors need to learn. They keep using addons the same way they always have. Future RFCs that take Embroider mainstream _will_ have impact, but that can be discussed then.
+
+The impact on addon authors is more significant. This design is fully backward compatible, and the intention is that all existing addons continue to work (some with worse compatibility hacks than others in the v1-to-v2 compiler). But there will be a demand for addons published in v2 format, since it is expected to result in faster build times. My prediction is that people who are motivated to get their own app build times down will send a lot of PRs to the addons they’re using.
```
patocallaghan, line 797: You mention that the success of getting this format adopted throughout the community will be down to maintainers migrating their addons or other individuals putting together PRs, should there be more emphasis on how we plan to teach and empower people to do this? @jenweber brings up some good points in https://github.com/emberjs/rfcs/pull/507/files#r296484559. 

Will there be guides, migration tutorials, quest issues? Could there be a codemod which gets addons 80% of the way there (if not 100% of the way in some cases)?

text/0000-embroider-v2-package-format.md
```diff
@@ -0,0 +1,817 @@
...
+```js
+// this example:
+import { getOwnConfig } from "@ember/macros";
+const shouldEnableCoolFeature = getOwnConfig().coolFeature;
+// might compile to:
+const shouldEnableCoolFeature = true;
+// assuming your ownConfig is `{ coolFeature: true }`.
+```
+
+`getOwnConfig()` behaves like a function that returns your `OwnConfig`, as determined by the return value of your `configure` **Build Hook**. You're allowed to chain property accesses (including array indices) off of `getOwnConfig()`. Since the `OwnConfig` is required to be JSON-serializable, any subset of it can be accessed this way, and we inline that value directly into the code.
```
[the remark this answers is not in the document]
chancancode: Is there reliable constant folding in babel and is that enabled by default? For example, would something like this work?

```js
const isCoolFeatureEnabled = "coolFeature" in getOwnConfig() && getOwnConfig().coolFeature.enabled === true;
const isCoolFeatureSupported = dependencySatisfies("ember-source", "^3.0.0");

if (macroCondition(isCoolFeatureEnabled && isCoolFeatureSupported)) {
  // ...
} else if (macroCondition(isCoolFeatureEnabled && !isCoolFeatureSupported)) {
  failBuild(...);
}
```

text/0000-embroider-v2-package-format.md
```diff
@@ -0,0 +1,817 @@
...
+
+```js
+if (macroConditional(getOwnConfig().x)) {
+  await import("x");
+} else {
+  await import("y");
+}
+```
+
+Q: Why not allow `if (getOwnConfig().thing)` instead of `if (macroCondition(getOwnConfig().thing))`?
```
chancancode, line 539: Isn't this still "allowed"? `macroCondition` is just an optional "lint" isn't it?

text/0000-embroider-v2-package-format.md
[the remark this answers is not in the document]
chancancode: Per the `macroCondition` rationale, isn't it better to do something like `getOwnConfig("some.nested.path")`, where it is clear what is static (whatever the path micro syntax allows you to express) and what is not (using the return value in JavaScript), rather than accidentally turning this into runtime code? (e.g. `getOwnConfig()["foo"]` vs `getOwnConfig()[foo]`)

text/0000-embroider-v2-package-format.md
```diff
@@ -0,0 +1,817 @@
...
+export default implementation;
+
+// ===============
+// Or compile down to this if OwnConfig contains { useNewVersion: false }
+let implementation;
+implementation = importSync("./old-component");
+export default implementation;
+```
+
+It is a build error if `macroCondition` cannot statically determine the truth status of its argument. It's a build error if `macroCondition` appears anywhere other than as the predicate of an if statement or ternary expression.
```
chancancode, line 512: Why this restriction? It seems useful to have this lint for things like this:

```js
const isCoolFeatureEnabled = macroCondition(...);
const isCoolFeatureSupported = macroCondition(...);
```

Per my question above, I am not sure if the following is allowed:

```js
if (macroCondition(isCoolFeatureEnabled && isCoolFeatureSupported)) {
  // ...
}
```

I think we should in theory be able to support that with good constant folding. But even if it's not currently supported due to some limitation, I can still do:

```js
if (isCoolFeatureEnabled && isCoolFeatureSupported) {
  // ...
}
```

If I can be sure that the constants were assigned a static boolean value, then even if our macro system can't handle it, I can use some other babel plugin to do the constant folding + dead code elimination. The bottomline is, it seems useful to be able to use `macroCondition` outside of if statements or ternary expressions, since it's just a lint to make sure the expressions are completely static.

text/0000-embroider-v2-package-format.md
```diff
@@ -0,0 +1,817 @@
...
+
+Then:
+
+```hbs
+<SomeComponent @score={{macroGetOwnConfig "items" 0 "score" }} />
+{{! ⬆️compiles to ⬇️ }}
+<SomeComponent @score={{42}} />
+```
+
+If you don't pass any arguments, you can get the whole thing (although this makes your template bigger, so use arguments when you can):
```
chancancode, line 627: Why not compile it into a runtime helper? the hash/array literals in handlebars may also have subtle difference than `JSON.parse` (prototype, frozen bit, etc)

text/0000-embroider-v2-package-format.md
```diff
@@ -0,0 +1,817 @@
...
+
+```hbs
+<SomeComponent @canAnimate={{macroDependencySatisfies "liquid-fire" "*"}} />
+{{! ⬆️compiles to ⬇️ }}
+<SomeComponent @canAnimate={{true}} />
+```
+
+### Handlebars Macro: macroMaybeAttrs
+
+There is one place where `{{#if}}` doesn't work: within "element space". If you want to _sometimes_ set an attribute, but sometimes not, this doesn't work:
```
chancancode, line 662: Doesn't glimmer already treat `={{false}}` as "remove this attribute?

chancancode: What about modifiers?

text/0000-embroider-v2-package-format.md
```diff
@@ -0,0 +1,817 @@
...
+
+Notice that a package’s **allowed dependencies** do not include the package itself. This is consistent with how Node resolution works. To import files from within your own package you must use relative paths. This is different from how run-time AMD module resolution has historically worked in Ember Apps. (`@embroider/compat` implements automatic adjustment for this case when compiling from v1 to v2).
+
+Modules in **Own Javascript** are allowed to use dynamic `import()`, and the specifiers have the same meanings as in static import. However, we impose a restriction on what syntax is supported inside `import()`. The only supported syntax inside `import()` is:
+
+- a string-literal
+- or a template string literal
+  - that begins with a static part
+  - where the static part is required to contain at least one `/`
+  - or at least two `/` when the path starts with `@`.
```
jelhan, line 134: This should mention dynamic imports from external URLs explicitly. They are different from other imports as they aren't linked at build-time. It's currently unclear to me if they are supported and what supported means in this context.

There are three different types of dynamic imports from external URLs:

1. From a static URL
   ```js
   import('https://examples.com/foo.js');
   ```
2. From a URL template
   ```js
   import(`https://examples.com/${file}`);
   import(`https://${url}`);
   import(`//${url}`);
   ```
3. From a variable
   ```js
   let url = 'https://examples.com/foo.js');
   import(url);
   ```

Only for first and second example it's known at build-time that the target of the dynamic import is an external URL. The compiler does not know anything about the third example.

All three examples are valid JavaScript. They are [supported natively in most browsers](https://caniuse.com/#feat=es6-module-dynamic-import). Only IE11 and Edge versions that are not based on Chrome yet do not support dynamic imports.

They work in current Ember applications unless the app uses ember-auto-import. Ember-auto-import throws a build-time error for second and third example that the import target must be a string literal. If the app does not use ember-auto-import the dynamic import will not be transpiled and included in the build untouched.

The limitation of ember-auto-import was [raised in this issue](https://github.com/ef4/ember-auto-import/issues/276). Four possible solutions were discussed in the issue:

1. Support template literals if they start with `https://` or `//`.
   ```js
   import(`https://${url}`);
   import(`//${url}`);
   ```
2. Skip build-time transform with magic comments.
   ```js
   /* ember-auto-import-disable */
   import(url);
   /* ember-auto-import-enable */
   ```
3. Reserve `import()` for runtime imports and add a dedicated build-time-`import()` macro.
4. Add a lint rule forbidding dynamic imports. Silently skip dynamic imports at build-time and leave them untouched instead of throwing.

The first option would break some valid JavaScript. It may even break existing Ember applications. But it would still allow all use cases cause the [list of possible protocols is known upfront](https://github.com/ef4/ember-auto-import/issues/276#issuecomment-611597946).

The recommended solution for webpack [seems to be](https://github.com/webpack/webpack/issues/8341) magic comments: `import(/* webpackIgnore: true */ "/js/hello.js")`. But [@ef4 explained](https://github.com/ef4/ember-auto-import/issues/276#issuecomment-613087368) that it's the most complex option from implementation perspective.

The dedicated macro for build-time dynamic imports would be the simplest option from implementation perspective. But it would be different from the rest of the JS ecosystem and therefore increases the teaching effort dramatically as pointed out by @ef4 in the same comment.

The last option would be a change to the overall strategy. A dynamic import which target is not known at build-time would not trigger a build-time error but would be silently skipped. Instead the developer would be warned by a lint rule. It may cause strange and hard to debug issues if the developer ignores such a lint error. But valid JavaScript would just work. To be honest this is my preferred option.

A potential use case for dynamic imports from external URLs is expanding the components available to an application at run-time. E.g. there might be a registry of web components to which the user can publish new components at any time. The application does not know which web components exists at build-time. But it can resolve all of them using dynamic imports from an external URL at run-time.

text/0000-embroider-v2-package-format.md
pzuraq: I believe this is addressed in the [Scope of this RFC](https://github.com/emberjs/rfcs/blob/embroider/text/0000-embroider-v2-package-format.md#scope-of-this-rfc) section:

> No provision for:
> ...
> * preprocessing your parent package's code
> * modifying your parent package's babel config
> ...
>
> It is understood that all of these are legitimate things for Ember addons to do. Defining these capabilities within v2 packages will be done in followup RFCs. It is simply too much scope to cover in one RFC.

The three hooks are the only APIs currently in v2 addons, which means that some addons will not be able to migrate to the v2 format yet. The v1 format will continue to be supported however, it just means that those addons will not be able to benefit from things like tree-shaking, etc.

simonihmig:
Another thing that came to my mind: are there any implication on what a v2 addon can depend on, more specifically do all addons in `dependencies` also have to be in v2 format?

I am thinking of a concrete example actually: an addon uses a newer Ember feature, that requires a polyfill for older Ember versions. Say `in-element` (could be anything else that requires an AST transform, like ABI). So it has `ember-in-element-polyfill` as a dependency, which is still a v1 addon, so relies on a *build-time*  AST transform (basically doing `{{#in-element ...}}` -> `{{#-in-element ...}}` when the polyfill is needed). 

At *publish-time*  this transform will not get automatically applied (it even doesn't make any sense, as we don't know the Ember version in use at this point in time), so the original addon has everything pre-compiled, but still uses `{{in-element}}` directly in its templates.

So will the v1 polyfill still work at build-time, transforming the v2 addons templates? Which is probably too late as templates have been already precompiled AFAIU, and it seems to contradict the basic idea of pre-compilation and this: 
> Addon authors can still use whatever custom AST transforms they want, but those transforms must have already been applied before publication to NPM.

Or would the polyfill also have to be rewritten to be somehow v2 aware (it's a different case though as it does not provide any importable modules by itself), so basically
* have some means to plug into the prepublish build of the v2 addon (any defined hooks / conventions for this?)
* do a very different AST transform, something like from: `{{#in-element ...}}` to:
```hbs
{{#if (macroCondition (macroDependencySatisfies "ember-source" "^3.20.0"))}}
  {{#in-element ...}}
{{else}}
  {{#-in-element ...}}
{{/if}}
```

pzuraq:
@simonihmig I believe that the polyfill would have to expose its transforms in a way that could be used at build time, and then you would have to run it through HTMLbars ahead as part of your prepublish. This would mean a shift in our addon ecosystem, for sure, but the base code of the transforms wouldn't have to change, they would just have to align more with tools like Babel and TypeScript where they expect you to compile (or at least partially compile) beforehand.

text/0000-embroider-v2-package-format.md
simonihmig: Thanks for the clarification! So this basically confirms my suspicion here:

> which will be extended further by following RFCs? (this seems what the "Supporting References" chapter suggests, but this was easy to miss)

But given the already quite long text, as I said this was easy to miss. So for future readers maybe a little note here would make sense.

simonihmig:
> This would mean a shift in our addon ecosystem, for sure, but the base code of the transforms wouldn't have to change

I don't think this is true. As given in my example above, the current line of polyfills does very different things at *build-time* (of the app) based on the Ember version (see e.g. https://github.com/ember-polyfills/ember-angle-bracket-invocation-polyfill/blob/master/index.js#L14-L16), which is not known beforehand at *publish-time* . That's why the transform would have to be very different to work for a v2 addon, making use of `@embroider/macros` as in the example I gave above.

> I believe that the polyfill would have to expose its transforms in a way that could be used at pre-publish build time

Yeah, I agree *something* like that would have to happen. But this is obviously very vague, but we need to have some common API / conventions at some point in time, to not force each and every addon author to figure this out somehow (differently), and cause fragmentation in the ecosystem.

I can expect what will come now as a response, and I already agree: it's out of scope for *this* RFC, but I just want to make sure this is on the radar, and will somehow be looked into, probably a follow-up RFC?

Gaurav0:
It looks like Ember Twiddle can continue to interpret v2 addons with ember-cli as regular addons for now? Also @ef4 at some point I will need your guidance on how to giftwrap addons with embroider (hopefully before any embroider specific addons start to become available).

Here is the code currently being used to giftwrap addons for Ember Twiddle today: https://github.com/joostdevries/twiddle-backend/blob/36b63b1021d22b974d534f01f3c723e50c06a1c7/addon-build-configs/3.18.0/ember-cli-build.js

text/0000-embroider-v2-package-format.md
```diff
@@ -0,0 +1,817 @@
...
+
+`ember-source` provides many "virtual" packages like `@ember/component`. If they were real packages, they would be `peerDependencies`, but having non-real packages in package.json is likely to result in errors. Pedantically, they can be listed in **externals** instead. In practice, they are a well-known set that we can always handle correctly automatically, so it's not very important whether an addon includes them in **externals**.
+
+### Optional Peer Dependencies
+
+Some addons optionally use another addon if it happens to be available in the app. In order to resolve such a dependency, we really need **Optional Peer Dependencies**.
+
+NPM doesn't have a concept of optional peer dependencies. It has "optional dependencies", but they are something different and pretty useless.
+
+Yarn did an [RFC for optional peer dependency support](https://github.com/yarnpkg/rfcs/blob/master/accepted/0000-optional-peer-dependencies.md). It is basically compatible with NPM, with the only caveat being that if you use NPM you may see a spurious warning at install time. As non-actionable peerDependency warnings are rife throughout the NPM ecosystem this doesn't seem like a big deal.
```
kellyselden, line 727: That RFC doesn't mention how one would consume such an addon. What happens if you try to `import` from a missing dep?

text/0000-embroider-v2-package-format.md
```diff
@@ -0,0 +1,817 @@
...
+> Q: Should we just automatically namespace them instead?
+> A: That was considered, but it makes backward compatibility harder, and public URLs are not always free to choose/change.
+
+## Build Hooks
+
+In today’s v1 addon packages, the `index.js` file is the main entrypoint that allows an addon to integrate itself with the overall ember-cli build pipeline. The same idea carries forward to v2, with some changes.
+
+It is no longer the `main` entrypoint of the package (see **Own Javascript**). Instead, it’s located via the `build` key in **Ember package metadata**, which should point at a Javascript file. `build` is optional — if you don’t have anything to say, you don’t need the file.
+
+It is now an ECMA module, not a CJS file. The default export is a class that implements your build hooks (there is no required base class).
```
kellyselden, line 305: What is the reasoning of this decision? Is it using esm for broad node support?

text/0000-embroider-v2-package-format.md
```diff
@@ -0,0 +1,817 @@
...
+
+- it solves the longstanding problem that when you run your tests by visiting `localhost:4200/tests` the tests see the `development` environment, not the `test` environment. To get the test environment you can't use `ember serve`, you must use `ember test`. This has remained unfixed because it's expensive to do the whole build twice for the two environments.
+
+  We can solve this problem by producing a _single_ build containing _both_ environments, guarded by the macro system. The macros can evaluate at runtime, allowing each environment to get the right thing. In production builds, test-only or dev-only branches will still be eliminated.
+
+- it makes Fastboot builds simpler because we can guard the fastboot-only and browser-only code with the macro system. In development, we can run a single build that leaves both branches in and evaluates the macros at runtime.
+
+The macros package (`@ember/macros` as proposed, `@embroider/macros` as implemented) will work in both regular ember-cli and in Embroider. And it will work in both V1 and V2 packages.
+
+## Peer Dependencies
```
kellyselden, line 711: Are you debating failing the build when a peer dep isn't satisfied? I always imagine the following scenario:

package A has a peer dep on package B
package C has a direct dep on package B
App includes package A, then indirectly satisdfies the peer dep by including package C (on purpose, package C is a bundling package)

npm will warn like crazy, even though the app consumer may know what they are doing. I wouldn't want a strict implementation of peer deps to fail in this case.

text/0000-embroider-v2-package-format.md
ef4: Correct. I just pushed an update that tries to make this clearer by breaking out a separate section of the document for "Ember Standard Language", which explicitly allows decorators and some other features.

text/0000-embroider-v2-package-format.md
ef4: Yes, this is definitely the highest priority once the addon is merged. In response to your feedback I added a section to the "How we teach" section that says explicitly that we should release an official package with the tooling for addon authors and that package's documentation should serve as the starting point for teaching them.

text/0000-embroider-v2-package-format.md
```diff
@@ -0,0 +1,817 @@
...
+
+and no `version` key (or version key less than 2) in **Ember package metadata**.
+
+**non-Ember package**: a package without `keywords: ["ember-addon"]`
+
+## Scope of this RFC
+
+This RFC is intended as the base level spec for v2 Ember packages. **It does not attempt to cover everything a v1 package can do today**. For example, no provision is made in this RFC for:
+
+- providing dev middleware
```
kellyselden, line 84: Out of curiousity. If everyone seriously objected to one or more of these following RFCs, would it disrupt the whole system? And leave the implementation of this RFC in a difficult spot? Or can this RFC exist on its own without one of these pieces?

text/0000-embroider-v2-package-format.md
```diff
@@ -0,0 +1,817 @@
...
+**addon**: a package not used at the root of a project. Will be an **allowed dependency** of either an **app** or an **addon**.
+
+**allowed dependency**: For **addons**, the **allowed dependencies** are the `dependencies` and `peerDependencies` in `package.json` plus any in-repo addons. For **apps**, the **allowed dependencies** are the `dependencies`, `peerDependencies`, and `devDependencies` in `package.json` plus any in-repo addons.
+
+**Ember package metadata**: the `ember-addon` section inside `package.json`. This already exists in v1, we’re going to extend it.
+
+**v2 package**: a package with `package.json` like:
+
+    "keywords": [ "ember-addon" ],
+    "ember-addon": {
```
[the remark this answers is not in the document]
ef4: I agree we can adopt `ember` instead of `ember-addon`. 

Using a real market package would have distinct disadvantages. NPM would really download  it, and it would get traversed over and over in the graph because it would be a leaf beneath every addon. And we really *don't* want people to import any code out of it, because that's the kind of implementation-defined behavior this RFC is trying to move away from.

As for making the version more like a package string, I don't think that's a good fit. If an integer turns out to be insufficient, the next step up would be a capability set, not a semver string. This is much more like the capabilities on a component manager than the version of a package.

text/0000-embroider-v2-package-format.md
```diff
@@ -0,0 +1,817 @@
...
+- **Externals**: allows a package to declare that it imports some things that are not **allowed dependencies**. Instead of being resolved at build time, **externals** are deferred until runtime and get handled by the traditional `loader.js` `require()`.
+- **Assets**: any files that must be available in the final built application directory such that they have public URLs (typical examples are images and fonts).
+- **Build Hooks**: code that runs within Node at application build time. The v1 equivalent is an addon's `index.js` file.
+
+## Own Javascript
+
+The public `main` (as defined in `package.json`) of a v2 package points to its **Own Javascript**. The code is formatted as ES modules using ES latest features, meaning: stage 4 features only, unless otherwise explicitly mentioned elsewhere in this spec. (Addon authors can still use whatever custom syntax they want, but those babel plugins must run before publication to NPM.)
+
+Templates are in hbs format. No custom AST transforms are supported. (Addon authors can still use whatever custom AST transforms they want, but those transforms must have already been applied before publication to NPM.)
```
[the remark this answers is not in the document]
ef4: I incorporated this change, due to unrelated changes I can't actually use the "commit suggestion" button.

text/0000-embroider-v2-package-format.md
```diff
@@ -0,0 +1,817 @@
...
+
+Modules in **Own Javascript** are allowed to import template files. This is common in today’s addons (they import their own layout to set it on their Component class). The `.hbs` file extension resolves with lower priority than `.js` (meaning if you have both, the `.js` wins).
+
+Modules in **Own Javascript** are allowed to import JSON files. This is supported in Node and it's enabled by default in popular Javascript bundlers like Webpack, so it's common to encounter code on NPM that expects it to work. We interpret JSON as an ECMA module whose `default` export is the JSON object.
+
+### Own Javascript: Co-located Component Templates
+
+Modules in **Own Javascript** get no _automatic_ component template co-location. They should use Ember's `setComponentTemplate` explicitly. They can access their compiled templates either by importing them or by inlining them via the `hbs` macro provided by ember-cli-htmlbars.
+
+We're assuming that template co-location will be handled by the build step _before_ an addon is published to NPM in V2 format.
```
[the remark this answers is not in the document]
ef4: I think the right place to address this is in the shared tooling. I've added to the RFC that we should ship an official package that gives addon authors tooling for this.

(Deciding which hook to use isn't in scope for the RFC because you can build your package using whatever bizarre system you want as long as the output conforms to the standrad, but shipping one good shared solution is definitely important!)

text/0000-embroider-v2-package-format.md
```diff
@@ -0,0 +1,817 @@
...
+
+The API that packager provides is also incomplete compared with this design. For example, to take the packager output and build it using Webpack, Rollup, or Parcel still requires a significant amount of custom code. Whereas taking a collection of v2 formatted Ember packages and building them with any of those tools requires very little Ember-specific code.
+
+The prebuilt addons RFC addresses build performance by doing the same kind of work-moving as this design. Addons can do much of their building up front, thus saving time when apps are building. But it only achieves a speedup when apps happen to be using the same build options that addons authors happened to publish. This design takes a different approach that preserves complete freedom for app authors to postprocess all addon Javascript, including dead-code-elimination based on the addon features their app is using. The prebuilt addons RFC also doesn’t attempt to specify the contents of the prebuilt trees — it just accepts the current implementation-defined contents. This is problematic because shared builds artifacts are long-lived, so it’s worth trying to align them with very general, spec-compliant semantics.
+
```
kellyselden, line 891: Perhaps this needs a Drawbacks section? One thing that comes to mind is my comment on the prebuilt assets RFC https://github.com/ember-cli/rfcs/pull/118#issuecomment-386441383. Furthering on that, what is the workflow for symlinking your addon into your app as part of your development flow. Does making a change in your addon get picked up automatically by your app's running live reload server, or do you need a second live reload system to generate the addon's dist?

text/0000-embroider-v2-package-format.md
```diff
@@ -0,0 +1,817 @@
...
+### Own Javascript: Transpilation of imported modules
+
+Any module you import, whether from an Ember package or a non-Ember package, gets processed by the app's babel configuration by default. This ensures that the app's `config/targets.js` will always be respected and you won't accidentally break your older supported browsers by importing a dependency that uses newer ECMA features.
+
+There is an explicit per-package opt-out for cases where you're _sure_ that transpilation is not needed and not desirable. (See **Build Hooks** for details on the `skipBabel` option.)
+
+### Own Javascript: Supported module formats for non-Ember packages
+
+As already stated, V2 Ember packages must contain only ES modules. However, non-Ember packages in your **allowed dependencies** are allowed to contain ES modules _or_ CommonJS modules. This provides the best compatibility with general-purpose NPM utilities.
```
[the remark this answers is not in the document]
ef4: **allowed dependencies** is defined explicitly in the *Definitions* section of the document.

Are you asking if it's OK to have CJS modules in your package that are intended for consumption by non-ember projects? 

I would say yes, for the same reason you can have a file like `.eslintrc.js` that isn't actually an ES module -- as long as nobody tries to consume it from the ember app, we aren't going to go out of our way to error and shame you for shipping a non-ES-module.

text/0000-embroider-v2-package-format.md
```diff
@@ -0,0 +1,817 @@
...
+See **Macro System** for the full details.
+
+## App Javascript
+
+To provide **App Javascript**, a package includes the `app-js` key in **Ember package metadata**. For example, to duplicate the behavior of v1 packages, you could say:
+
+    "ember-addon": {
+      "version": 2,
+      "app-js": "./app"
+    }
```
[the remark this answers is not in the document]
ef4: I incorporated these suggestions.

text/0000-embroider-v2-package-format.md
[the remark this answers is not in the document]
chancancode: Right, I was asking about the `must contain` and `allowed to contain`, not the **allowed dependencies** terminology. For example, is it possible for the same package to be an ember addon and "normal" npm package (let's say the author of `momentjs` happen to be an ember user, and wants the package to work both as a regular package and also optionally adds the template helpers if inside ember)? Is there any conflicts re: the `main` field in `package.json` and such?

text/0000-embroider-v2-package-format.md
```diff
@@ -0,0 +1,817 @@
...
+The Handlebars macros are:
+
+- macroGetOwnConfig
+- macroGetConfig
+- macroCondition
+- macroDependencySatisfies
+- macroMaybeAttrs
+- macroFailBuild
+
+The difference in naming is because the JS macros get explicitly imported from `@ember/macros`, whereas the Handlebars macros do not, so they need an appropriate namespace prefix. (If we land template imports, I'm fine with adjusting this RFC to make the names align.)
```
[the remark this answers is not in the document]
ef4: True, though these macros also need to work in non-strict-mode templates, so the names in this RFC probably stay too.

Now that I think about it, I don't actually understand yet the implementation strategy for making the macros importable. Today they are an AST transform, but if we make them importable they would become ambiently-scoped values that we can't statically identify inside the ast transform, so the work would need to move up into babel where we can see the ES imports, and have that import the template compilation nested within it.

Right now, I expect the currently non-imported, namespaced names would work in both non-strict and strict templates, and that's not a bad place to start from.

text/0000-embroider-v2-package-format.md
```diff
@@ -0,0 +1,817 @@
...
+
+To provide test-support code, make a separate module within your package and tell people to `import` it from their tests. As long as it is only imported from tests, it will not be present in non-test bundles. (Things get simpler when you respect the module dependency graph.)
+
+## Macro System
+
+v1 packages can run arbitrary Node code that completely alters their runtime code. This makes them impossible to analyze. v2 packages are not allowed to do this. There are no "treeFor\*" hooks. Instead, they can influence their runtime code only through the macro system.
+
+It helps to think about the macro system as an extension to Javascript itself that we allow in v2 packages. Because the macros are allowed to appear in any published V2 package, and because the macros are _not_ a dependency that each package can control (you don't get to bring your own separate macro system version with you), it's important that we design a small core that we can support for the long-term. We probably can't make breaking changes to the macro system, we can only introduce new macros.
```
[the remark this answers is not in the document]
ef4: This was maybe more paranoid than necessary, just because a lot of the existing build tooling out there is bad at per-package configuration. For example, telling babel (a *single* babel, which is important!) to run different plugins for different packages requires a lot of bring-your-own infrastructure to make that work, where you end up doing relatively expensive searches to reverse engineer which package owns which file.

That said, it turns out we already needed that infrastructure and we do have it. So we could choose to use the `ember.version` key that we discussed elsewhere to influence the meaning of the macros if we really need to in the future.

I think it's fine to keep the RFC in this more conservative position, and it doesn't foreclose the ability to do those things if we need them in the future.

text/0000-embroider-v2-package-format.md
```diff
@@ -0,0 +1,817 @@
...
+
+The semver range is useful to disambiguate if there are multiple versions of the same package involved in the app, and in cases where a developer has manually verified that transpilation isn't needed, it's good practice to use the semver range so that `skipBabel` doesn't accidentally apply to a future version of the package that may indeed need transpilation.
+
+## What about Test Support?
+
+v1 packages can provide `treeForTestSupport`, `treeForAddonTestSupport`, and `app.import` with `type="test"`. All of these features are dropped.
+
+To provide test-support code, make a separate module within your package and tell people to `import` it from their tests. As long as it is only imported from tests, it will not be present in non-test bundles. (Things get simpler when you respect the module dependency graph.)
+
+## Macro System
```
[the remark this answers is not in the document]
ef4: That's correct.

It's enforced statically at build time if you try to use the macros.

text/0000-embroider-v2-package-format.md
ef4: Correct, it would still be allowed and it *might* even get your branch elimination if a tool like uglify sees the output of this.

It's a little more than a lint because it does change the point within the build process at which branch elimination happens. With macroCondition, branch elimination will happen *before* you traverse beyond the current module to discover more modules. So if you eliminate a branch that imports a non-parseable file, macroCondition prevents the parse error whereas relying on uglification would not.

text/0000-embroider-v2-package-format.md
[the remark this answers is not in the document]
ef4: If the key is not found it's an error. If you want to check for an optional key, we explicitly support optional chaining: `getConfig().thing?.enabled`.

We don't currently support general constant folding, though implementing increasingly more analysis seems fine, as long as we only increase the set of cases that is supported and don't decrease it. There is no risk of accidentally allowing dynamic cases through because the macros will error if they can't statically understand a scenario.

> Per the macroCondition rationale, isn't it better to do something like getOwnConfig("some.nested.path"), where it is clear what is static (whatever the path micro syntax allows you to express) and what is not (using the return value in JavaScript), rather than accidentally turning this into runtime code? (e.g. getOwnConfig()["foo"] vs getOwnConfig()[foo])

That's not a bad idea. The cost of it is inventing a microsyntax. The benefit is that you can immediately see which parts are build-time evaluated.

I'm not actually sure if it's important to make those cases look different, however. Certainly `getOwnConfig()["foo"]` and `getOwnConfig()[foo]` are both legitimate cases, and they both work correctly today. They don't seem to have any ambiguity, they do exactly what the javascript says they do. The mental model for users should be that `getOwnConfig()` returns an object, and you can use it in every way you use any other object. We just happen to optimize how much of the object actually gets inline into the code when we happen to notice that you're only using some of it. But that is purely optimization, and doesn't impact correctness.

text/0000-embroider-v2-package-format.md
ef4: That would also be a fine implementation. I regret implying one over-specific implementation here in the spec. Any implementation that works *as if* we had inlined your config satisfies the spirit of the design.

text/0000-embroider-v2-package-format.md
ef4: I think these are good ideas that could improve the ergonomics of the macro system but 
 - they are additive to this RFC
 - the ergonomics of the macro system aren't super high-stakes, because we're still talking about a publication format. People can iterate on more ergonomic tooling that *emits* the macros as output. 

My biggest goals for the macros in this RFC are that we 
 1. don't ship something that we regret needing to support in the future
 2. have reasonable completeness for all the common things addons are doing with custom broccoli today.

text/0000-embroider-v2-package-format.md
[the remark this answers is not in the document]
chancancode: The strict mode RFC split things into two categories. Importables and keywords, importables must be imported in strict mode and are essentially global variables. They can be shadowed by local variables and can be passed around as values. On the other hand, the keywords/macros (e.g. `{{yield}}`) does not have an import path and cannot be used as a value (may or may not be shadow-able, not 100% sure, but I think we have to work with what we have got today).

text/0000-embroider-v2-package-format.md
ef4: > Doesn't glimmer already treat ={{false}} as "remove this attribute?

Does it do that as a compile-time optimization? If so, then that's great and we can rely on that.

> What about modifiers?

That is a good question, perhaps we still have a small hole here. I came up with `maybeMacroAttrs` because it was a clear case that popular addons are doing today. I'm not aware of any that conditionally remove a modifier invocation, but that seems plausible.

text/0000-embroider-v2-package-format.md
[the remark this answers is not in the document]
chancancode: > But that is purely optimization, and doesn't impact correctness.

Sure, but if you accidentally inlined a huge JSON blob that cannot be minified out, that could be about as bad as not eliminating a branch, which was the rationale of `macroCondition`. The config for things like eslint, etc, can be very big. Or it maybe all the localization strings for your entire app.

text/0000-embroider-v2-package-format.md
ef4: I have now updated the document to reflect that we would allow URLs as long as we can tell from their static prefix that they're URLs.

text/0000-embroider-v2-package-format.md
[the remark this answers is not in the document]
chancancode: An alternative to the micro syntax is something more like the design for `macroCondition`:

```js
import { ownConfig } from '@ember/macros';
````

Plus:

```
macroValue(ownConfig.foo.bar)
```

Verses:

```js
macroValue(ownConfig.foo)[this.dynamicConfigKey]
```

`macroValue` can ensure whatever is in the parens are inlined as a static literal value, otherwise it errors. If any part of the build wants to use `ownConfig` outside of `macroValue`, it can be emitted into a JavaScript module and the import rewritten to point to that generated module. That way, at least it's still only once per app at most, and none at all if all usages is wrapped in `macroValue()`.

ef4:
@simonihmig: 

> do all addons in dependencies also have to be in v2 format?

No, that would indeed be bad. Each package can update to v2 independently. The embroider build's first stage compiles away those differences.

> I am thinking of a concrete example actually: an addon uses a newer Ember feature, that requires a polyfill for older Ember versions. Say in-element (could be anything else that requires an AST transform, like ABI). So it has ember-in-element-polyfill as a dependency, which is still a v1 addon, so relies on a build-time AST transform (basically doing {{#in-element ...}} -> {{#-in-element ...}} when the polyfill is needed).

That would indeed be a problem, but not because ember-in-element-polyfill is a v1 package. It's a problem because the transform is running on the wrong input (it's running based on the version of ember in your addon's dummy app, not the version of ember in the addon's consumers' apps).

v2 packages will need to use polyfills that are updated to work safely with embroider. For example, the in-element polyfill can compile`{{#in-element}}...` o 

```hbs
{{#if (macroCondition (dependency-satisfies "ember-source" ">=4.0.0")) }}
  {{#in-element}}...
{{else}}
  {{#-in-element}}...
{{/if}}
```

This would make the template bigger, but only on NPM. Once that template is compiled by the consuming app, it will only contain the correct branch for the app's ember version.

text/0000-embroider-v2-package-format.md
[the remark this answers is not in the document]
chancancode: Right, but I do want to make sure that there is _some_ hook that works with linking, installing from GitHub, doesn't have any undesirable side-effect (e.g. using postinstall and forcing a multi-minute wait for the build after running `yarn install`). At some point this was a mess and none of the option seem to work in all cases or have notable downsides. Not sure if it has changed since, but I certainly hope so. Maybe @rwjblue will know.

text/0000-embroider-v2-package-format.md
chancancode: > Does it do that as a compile-time optimization? If so, then that's great and we can rely on that.

I don't think it does, but assuming I was correct, and that this behavior is universal (i.e. not conditional on "oh let's see if we are doing an attribute vs property, etc" then I don't see why not). But even if it's not optimized yet, if that behaves correctly (i.e. not adding the attribute to the DOM), then I don't think `maybeMacroAttrs` is necessary.

ef4:
> It looks like Ember Twiddle can continue to interpret v2 addons with ember-cli as regular addons for now?

Yes, that would work. We really really do need to make sure everybody publishing v2 addons includes the tooling to make them work correctly within non-embroider builds too. We don't want to split the ecosystem.

This RFC is a step toward making the builds *way* simpler and that will help twiddle, but before we can really get to that point we need to get past the need for all the compatibility stuff. Once everything is natively v2, and everybody is using strict mode templates, etc, a lot of the build tooling can just go away.

ef4:
At this point I have responded to the new feedback since FCP and incorporated several good suggestions to clarify the document.

Framework core discussed on our call today and agree FCP should continue this coming week so everyone has a chance to respond to the latest round.

text/0000-embroider-v2-package-format.md
ef4: import from a missing dep would be a static build error.

To actually import things out of an optional dependency you'd need to guard it with the macro system:

```js
if (macroCondition(dependencySatisfies("the-optional-thing", "*"))) {
  let thing = importSync("the-optional-thing");
  doSomethingWith(thing);
}
```

text/0000-embroider-v2-package-format.md
ef4: It's authored as ES for the same reason that all other code in ember apps and addons is authored as ES. That's the only format that's part of ECMAScript standard.

How it actually runs is an implementation detail (which of course is already true of all the other ES modules in an Ember package).  We could certainly use esm or even babel-to-CJS.

text/0000-embroider-v2-package-format.md
ef4: This RFC doesn't fail the build for a missing peerDep.

But if you actually try to resolve a peerDep within your Ember code, Embroider implements that resolution strictly. You won't find an indirectly satisfied peerDep because that's just not safe to rely on. It happens to work often based on hoisting, but it also fails sometimes, and I have seen it fail many times in the wild.

Plus, you'll have a bad time adopting Yarn PnP or some other future more optimizable package system if you are relying on this kind of implicit dependencies.

So embroider is strict. A v2 package can't resolve things that aren't **allowed dependencies**, as defined in the definitions section.

The solution to your example is that package C should have a peerDep on B, so that B can be added once to the app and shared by both A and C (and the app). 

"bundling package" is a common pattern today in Ember addons, but for historical and no-longer-true-reasons. Now that importing libraries directly from NPM is easy, bundling packages either go away (if they were only bundling) or become separate libraries (if they were providing some ember-specific things like components and helpers that work nicely with the library).

text/0000-embroider-v2-package-format.md
ef4: It can exist by itself. In practice it would mean that v1 and v2 packages live side-by-side indefinitely because v2 never finishes implementing all the things v1 packages can do.

text/0000-embroider-v2-package-format.md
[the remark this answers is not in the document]
chriskrycho: The import design here is massively preferable over the string-key/micro syntax style from the perspective of TypeScript consumers: microsyntaxes are extremely hostile to providing types for correctly, and these kinds of boundaries/leaves are often some of the first places teams can and should introduce TS in their codebase because their impact tends to be so distributed. (While we don’t officially support TS at the framework level, keeping these considerations in view will make TS users’ and the Typed Ember team’s lives much easier.)

simonihmig:
> v2 packages will need to use polyfills that are updated to work safely with embroider. For example, the in-element polyfill can compile{{#in-element}}... 

Yes, understood. This is basically the same transform I posted [above](https://github.com/emberjs/rfcs/pull/507#issuecomment-624633700). In the same comment I asked this:

> Or would the polyfill also have to be rewritten to be somehow v2 aware (it's a different case though as it does not provide any importable modules by itself), so basically 
> * have some means to plug into the prepublish build of the v2 addon (any defined hooks / conventions for this?)
> * do a very different AST transform, something like from: {{#in-element ...}} to:

And that's still not entirely clear to me. So when the polyfill does the transformation we agree upon, it needs to do so at `prepublish` time, right? And if so, we currently have no infrastructure (APIs, convention) to allow the polyfill to somehow hook into the build process of the addon, before it gets published. So this needs to be addressed somehow, right? (Not necessarily by *this* RFC, but by *something*, like an addon build support library)

text/0000-embroider-v2-package-format.md
ef4: This RFC is careful not to saying anything about authoring format. So the answer is, "it depends".

It's possible to author addons directly in the format laid out here, in which case there's no build step at all. In practice, I do expect addon authors to want more creature comforts, implying a build step, but how that build tooling works now becomes something it's safe to experiment on without breaking any compatibility.

I intend this RFC to be similar to the level of the Component Manager RFC -- it establishes the baseline primitive that we can commit to supporting, which then frees people to experiment on top of the primitive to figure out the best developer experience.

ef4:
> Not necessarily by this RFC, but by something, like an addon build support library

Right, I've added text proposing this package is called `@embroider/addon` and that it gives addon authors the tooling they need.

I don't think polyfills will need much new infrastructure. A key thing to realize is that they can just always emit macros, they don't need to know whether they are running at publish time or app build time. The decision to evaluate the macros or not is delegated to the macro system so the polyfill doesn't need to think about it.

text/0000-embroider-v2-package-format.md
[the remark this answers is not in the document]
ef4: @chriskrycho that is a good point.


That said, I don't actually know how we could type `getOwnConfig` correctly anyway, because its meaning changes depending who's asking.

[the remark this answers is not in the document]
ef4: `importSync` is also not possible to type automatically today, although the TS team has expressed interest in a feature that would make it possible. Today you need to say:

```importSync("./some/thing") as import("./some/thing")```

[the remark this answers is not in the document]
Comment (chriskrycho) [reply text/0000-embroider-v2-package-format.md]: I'll reread on Monday and see (and I may ping you in Discord to follow up on that), but I _think_ that there are a couple ways we could wrangle that, including with interface declaration merging. People still might need to do some small hoop-jumping-through to make it work, but the key is *small* (vs. the large hoops that you have to jump through for microsyntaxes).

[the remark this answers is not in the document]
Comment (ef4) [reply text/0000-embroider-v2-package-format.md]: The not-yet-real feature would be something like:

```ts
importSync<T extends string>(specifier: T): import(T)
```

text/0000-embroider-v2-package-format.md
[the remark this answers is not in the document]
ef4: I remembered the biggest reason why you don't get to bring your own version of the macros: because their implementation is tied to a specific version of babel (or ember-cli-htmlbars). That has caused big pain in the ecosystem in the past.

For example, babel 8 is coming. As long as we port embroider/macros to babel 8, an app using all v2 addons will be free to upgrade to babel 8 unilaterally, with no changes in its addons. Whereas if we allow addons to bring their own macro implementations, you can't upgrade until the lowest common denominator addon has done the work to upgrade their potentially-idiosyncratic macro implementation.

We could lift this restriction in the future by inventing a more declarative API for defining macros.

igorT:
A question to explore in a future rfc is whether it is enough to expose global flags like env being `prod` or `dev` to a macro that would look something like `getConfig('app')` or we would also want the `configure` method to have access to it as well

ef4:
Agreed. It seems likely we will add a way for the macro system to talk about global things that are truly global. As I was working on fastboot compatibility last week the convenience of that became apparent.

rwjblue:
We discussed this in todays Ember core team meeting, and are still quite excited to move forward here!

Lets do it!!!

NullVoxPopuli:
in v2 addons, do we have a way of writing tests?
or do we need to establish a blueprint for all that boilerplate still?

ef4:
That question is beyond the scope of this RFC, because you're free to write tests however you want as long as you publish to NPM in this RFC's format, but I agree we also want to establish good patterns here.

> in both glimmer-apollo and ember-welcome-page, neither addon has its own tests -- is this intentional? 

Just because the tests live within a test harness application doesn't make them any less the addon's own tests. IMO it makes things more transparent and understandable. The classic pattern of putting `/tests` outside the `/tests/dummy` app was a source of unnecessary confusion and not a small amount of pain in `@embroider/compat`, because it doesn't align the code with the actual module namespacing (`./tests/components/foo.js` looks like it should be the `your-addon/tests/components/foo.js` module, but it's actually the `dummy/tests/components/foo.js` module).

> but do we need a test app for unit and rendering tests? I imagine rendering tests can be done without the whole application stack?

That's a whole lot of effort to get a strictly less-capable and less-representative test suite that would require a whole new set of learning materials. Whereas working within a fully-normal Ember app means that anyone who learns how to write tests for their app already knows how to write tests for an addon too.

I realize that heavily-mocked component tests (that often don't even run in a real browser) are popular in some parts of the JS ecosystem, but IMO that stems from pretty deep confusion about the point of tests.

NullVoxPopuli:
>  beyond the scope of this RFC

Makes sense -- I didn't know where else to ask though. Are these types of questions better for the forums? Or somewhere else? I feel like it's a bit of uncharted territory because we need to figure out / discuss things that are fundamental to how addons will be built, and since module unification (may it rest in peace), I don't recall the last time that happened 🤔 

> putting /tests outside the /tests/dummy app was a source of unnecessary confusion and not a small amount of pain

I believe it -- for the people I've worked with making their first addons, this comes up regularly

> I realize that heavily-mocked component tests (that often don't even run in a real browser)

oh I wasn't suggesting that, I still think those sorts of tests are terrible. I only imagine that it's doable (conceptually, maybe not technically (yet)), to have `setupRenderingTest` and `setupTest` either start their own empty shell of an app, or we provide start-app.js a la ye olde times. For example, @lifeart has [this set of scripts](https://gist.github.com/lifeart/b6df15f7197331314477b4a7a5349d6f) to demonstrate starting an app up without ember-cli.

ef4:
Yes, if you already have a monorepo making the test-app and the addon separate workspaces is very straightforward. You can organize them however makes sense to you (except that yarn doesn't allow nesting of workspaces, so one cannot be the child of the other).

For people who don't want a monorepo, a small amount of special handling is required to get the linking and dependency management correct. That is the approach that I took in ember-welcome-page, where it uses some utilities provided by `@embroider/addon-shim` (which I'm about to move into a dedicate `@embroider/addon-dev` package instead, since they're really only dev-time concerns) to handle linking and dependency synchronization.

PR #509, opened 2019-07-01
fix some typos

Draggha:
Hi @ef4,

I found some typos that you might want to correct. I hope this helps! :)

### DIFF
--- a/text/0000-embroider-v2-package-format.md
+++ b/text/0000-embroider-v2-package-format.md
@@ -404,7 +404,7 @@ The Handlebars macros are:
 - macroMaybeAttrs
 - macroFailBuild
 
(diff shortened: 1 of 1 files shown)


text/0000-embroider-v2-package-format.md
```diff
@@ -487,7 +487,7 @@ export default implementation;
 // ===============
 // Or compile down to this if OwnConfig contains { useNewVersion: false }
 let implementation
-implementation = importSync("./new-component");
+implementation = importSync("./old-component");
```
ef4: This one is not a bug, it's intentional to show that the output is different if the `useNewVersion` flag is false.

ef4: Oh wait, sorry, you're right, I had it backwards.

ef4:
Thanks!

PR #510, opened 2019-07-01
Add `load` hook to Route

michaelrkn:
This is almost entirely based off a blog post by @chadhietala. Thanks to @NullVoxPopuli for putting together #499 and reviewing the first draft of this.

[Rendered](https://github.com/michaelrkn/rfcs/blob/route-data/text/0000-route-data-loading.md)

### DIFF
--- a/text/0000-route-data-loading.md
+++ b/text/0000-route-data-loading.md
@@ -1,0 +1,72 @@
+- Start Date: 2019-07-01
+- Relevant Team(s): Ember.js
+- RFC PR: 
+- Tracking: 
+
+# Route Data Loading for Components
+
+## Summary
+
+#499 proposes an API for rendering a component from a route. This RFC builds on it by proposing an API to pass multiple arguments to a component from a route. It's based off the [design proposed by @chadhietala](https://gist.github.com/chadhietala/50b977a7d3476069892d351c65af418c) and much of the text is lifted from there.
+
+## Motivation
+
+The `model()` hook on routes is confusing, because it implies that only one piece of data will be fetched, and it uses the term "model" that is more typically associated with backend development. We should have a better name, and an API that allows data to be passed from a route into a component as multiple named arguments.
+
+## Detailed design
+
+In a world where we have routes invoking components, we can simply state that routes can fetch data and invoke components with named arguments as defined in the route. This might look like:
+
+
+```js
+// routes/profile.js
+export default class extends Route {
+  @service store;
+  async load({ profile_id }) {
+    return {
+      profile: await this.store.findRecord('profile', profile_id),
+    }
+  }
+}
+```
+
+```hbs
+{{! components/profile.hbs }}
+
+<h1>{{@profile.firstName}} {{@profile.lastName}}</h1>
+```
+
+So the keys of the object returned from `load()` become the named arguments to the component.
+
+The `load()` hook can also take responsibility for redirection:
+
+```js
+// routes/profile.js
+export default class extends Route {
+  @service store;
+  @service router;
+  @service user
+  async load({ profile_id }) {
+    if (this.user.loggedIn) {
+      return {
+        profile: await this.store.findRecord('profile', profile_id),
+      };
+    } else {
+      await this.router.transitionTo('login');
+    }
+  }
+}
+```
+
+## How we teach this
+
+Should #499 be accepted, this is a fairly straightforward and complementary change that would reduce confusion from the `model()` hook passing in a `@model` argument to a component. The Guides would need to be updated to use `load()` instead of `model()`, `beforeModel()`, and `afterModel()`.
+
+## Drawbacks
+
+In a time where a lot is changing in the Ember ecosystem, one might argue that this is too much change happening at once.
+
+## Alternatives
+
+* Stick with the `model()` and associated hooks.
+* Add `beforeLoad()` and `afterLoad()` hooks as are currently used in conjunction with `model()`.
--- a/text/0000-route-data-loading.md
+++ b/text/0000-route-data-loading.md
@@ -55,7 +55,7 @@ The `load()` hook can also take responsibility for redirection:
 
 ```js
 // routes/profile.js
-export default class extends Route {
+export default class ProfileRoute extends Route {
   @service store;
   @service router;
   @service user
--- a/text/0000-route-data-loading.md
+++ b/text/0000-route-data-loading.md
@@ -19,18 +19,20 @@ This RFC proposes adding a `load()` hook to `Route`.
 
 The `load()` hook will receive the same arguments as `model()`.
 
-The route is in the `loading` state while this hook is run. When this hook has finished running and any promises it returns have resolved, the route will change to the `resolved` state (and the `render` component will be rendered).
+The route is in the `loading` state while this hook is run. When this hook has finished running, the route will change to the `resolved` state (and the `render` component will be rendered).
 
 The keys of the object (what `Object.keys()` would return, i.e. only the own (`hasOwnProperty`) props, and nothing from prototypes) returned from `load()` become the named arguments to the component.
 
 ```js
 // routes/profile.js
+import { hash as resolveValues } from 'rsvp';
 export default class extends Route {
   @service store;
   async load({ profile_id }) {
-    return {
-      profile: await this.store.findRecord('profile', profile_id),
-    }
+    return await resolveValues({
+      profile: this.store.findRecord('profile', profile_id),
+      comments: this.store.query('comments', { profile_id })
+    });
   }
 }
 ```
--- a/text/0000-route-data-loading.md
+++ b/text/0000-route-data-loading.md
@@ -15,8 +15,7 @@ The `model()` hook on routes is confusing, because it implies that only one piec
 
 ## Detailed design
 
-In a world where we have routes invoking components, we can simply state that routes can fetch data and invoke components with named arguments as defined in the route. This might look like:
-
+This RFC proposes adding a `load()` hook to `Route`. The route is in the 'loading' state while this hook is run. The keys of the object returned from `load()` become the named arguments to the component.
 
 ```js
 // routes/profile.js
@@ -35,8 +34,6 @@ export default class extends Route {
 
 <h1>{{@profile.firstName}} {{@profile.lastName}}</h1>
 ```
-
-So the keys of the object returned from `load()` become the named arguments to the component.
 
 The `load()` hook can also take responsibility for redirection:
 
--- a/text/0000-route-data-loading.md
+++ b/text/0000-route-data-loading.md
@@ -7,7 +7,7 @@
 
 ## Summary
 
-[#499 An alternative to Controllers](https://github.com/emberjs/rfcs/pull/499) proposes an API for rendering a component from a route. This RFC builds on it by proposing an API to pass multiple arguments to a component from a route. It's based off the [design proposed by @chadhietala](https://gist.github.com/chadhietala/50b977a7d3476069892d351c65af418c) and much of the text is lifted from there.
+[#731 Add setRouteComponent API](https://github.com/emberjs/rfcs/pull/731) proposes an API for rendering a component from a route. This RFC builds on it by proposing an API to pass multiple arguments to a component from a route. It's based off the [design proposed by @chadhietala](https://gist.github.com/chadhietala/50b977a7d3476069892d351c65af418c) and much of the text is lifted from there.
 
 ## Motivation
 
@@ -19,7 +19,7 @@ This RFC proposes adding a `load()` hook to `Route`.
 
 The `load()` hook will receive the same arguments as `model()`.
 
-The route is in the `loading` state while this hook is run. When this hook has finished running, the route will change to the `resolved` state (and the `render` component will be rendered).
+The route is in the `loading` state while this hook is run. When this hook has finished running, the route will change to the `resolved` state (and the `setRouteComponent` component will be rendered).
 
 The keys of the object (what `Object.keys()` would return, i.e. only the own (`hasOwnProperty`) props, and nothing from prototypes) returned from `load()` become the named arguments to the component.
 
@@ -73,13 +73,13 @@ export default class ProfileRoute extends Route {
 
 ## How we teach this
 
-Should #499 be accepted, this is a fairly straightforward and complementary change that would reduce confusion from the `model()` hook passing in a `@model` argument to a component. The Guides would need to be updated to use `load()` instead of `model()`, `beforeModel()`, and `afterModel()`.
+Should #731 be accepted, this is a fairly straightforward and complementary change that would reduce confusion from the `model()` hook passing in a `@model` argument to a component. The Guides would need to be updated to use `load()` instead of `model()`, `beforeModel()`, and `afterModel()`.
 
 We could also include the `load()` hook in the route blueprint, returning an empty object, to make the hook more easily discoverable.
 
 ## Drawbacks
 
-In a time where a lot is changing in the Ember ecosystem, one might argue that this is too much change happening at once.
+Given that `setRouteComponent` is intended to unlock experimentation rather than being a final solution, adding this API at the present time could be premature.
 
 ## Alternatives
 
--- a/text/0000-route-data-loading.md
+++ b/text/0000-route-data-loading.md
@@ -15,7 +15,13 @@ The `model()` hook on routes is confusing, because it implies that only one piec
 
 ## Detailed design
 
-This RFC proposes adding a `load()` hook to `Route`. The route is in the `loading` state while this hook is run. The keys of the object returned from `load()` become the named arguments to the component.
+This RFC proposes adding a `load()` hook to `Route`. 
+
+The `load()` hook will receive the same arguments as `model()`.
+
+The route is in the `loading` state while this hook is run. When this hook has finished running and any promises it returns have resolved, the route will change to the `resolved` state (and the `render` component will be rendered).
+
+The keys of the object (what `Object.keys()` would return, i.e. only the own (`hasOwnProperty`) props, and nothing from prototypes) returned from `load()` become the named arguments to the component.
 
 ```js
 // routes/profile.js
@@ -34,6 +40,14 @@ export default class extends Route {
 
 <h1>{{@profile.firstName}} {{@profile.lastName}}</h1>
 ```
+
+If something other than an object is returned, an exception is thrown.
+
+If both `model()` and `load()` are present, an exception is thrown.
+
+As suggested in [Always run model hook #283](https://github.com/emberjs/rfcs/pull/283), the `load()` hook will always be run, even if a model is supplied to `{{link-to}}`.
+
+The `load()` hook will not support the "magic" behavior of the model hook, in which it will make loading work magically in simple scenarios without actually implementing the hook (http://api.emberjs.com/ember/release/classes/Route/methods/model?anchor=model).
 
 The `load()` hook can also take responsibility for redirection:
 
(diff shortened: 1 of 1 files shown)


text/0000-route-data-loading.md
```diff
@@ -0,0 +1,69 @@
...
+This RFC proposes adding a `load()` hook to `Route`. The route is in the `loading` state while this hook is run. The keys of the object returned from `load()` become the named arguments to the component.
+
+```js
+// routes/profile.js
+export default class extends Route {
+  @service store;
+  async load({ profile_id }) {
+    return {
+      profile: await this.store.findRecord('profile', profile_id),
+    }
```
buschtoens: Can you update this to use [`RSVP.hash`](http://api.emberjs.com/ember/3.10/functions/rsvp/hash) and multiple models?

```js
return await hash({
  profile: this.store.findRecord('profile', profile_id),
  comments: this.store.query('comments', { profile_id })
});
```

michaelrkn: What do you see as the advantage of using `RSVP.hash` over native JavaScript?

simonihmig: Jan's example above will make the network requests run in parallel, while...

```js
return {
  profile: await this.store.findRecord('profile', profile_id),
  comments: await this.store.query('comments', { profile_id })
}
```

... will make them run sequentially.

text/0000-route-data-loading.md
```diff
@@ -0,0 +1,69 @@
...
+
+[#499 An alternative to Controllers](https://github.com/emberjs/rfcs/pull/499) proposes an API for rendering a component from a route. This RFC builds on it by proposing an API to pass multiple arguments to a component from a route. It's based off the [design proposed by @chadhietala](https://gist.github.com/chadhietala/50b977a7d3476069892d351c65af418c) and much of the text is lifted from there.
+
+## Motivation
+
+The `model()` hook on routes is confusing, because it implies that only one piece of data will be fetched, and it uses the term "model" that is more typically associated with backend development. We should have a better name, and an API that allows data to be passed from a route into a component as multiple named arguments.
+
+## Detailed design
+
+This RFC proposes adding a `load()` hook to `Route`. The route is in the `loading` state while this hook is run. The keys of the object returned from `load()` become the named arguments to the component.
```
simonihmig: While I like the approach, I see a few things that are underspecified or deserve to be defined more explicitly:
* It will continue to wait for returned promises as `model()` does, affecting the `loading` state as well as the rendering of the component, as proposed by RFC499. "while this hook is run" is not very precise in this regard.
* I assume the arguments will be the same as for `model()`?
* Will `load()` always be called, even when the "model" is supplied in the `link-to`. I.e. how does it relate to https://github.com/emberjs/rfcs/pull/283
* The `model` hook has a default implementation that will make loading work magically in simple scenarios without actually implementing the hook, see http://api.emberjs.com/ember/release/classes/Route/methods/model?anchor=model. Do we want to copy over this (too) magic behaviour? I think the new `load()` API proposed here would be a good way to get rid of this, and force users to make everything explicit!
* The "keys" are what `Object.keys()` would return, i.e. only the own (`hasOwnProperty`) props, and nothing from prototypes, right?
* What happens when the returned (resolved) value is not an object? (a primitive value, or an array) How is the route component called in this case? Or is the error component rendered? Or does it simply throw an exception?

text/0000-route-data-loading.md
```diff
@@ -0,0 +1,69 @@
...
+      await this.router.transitionTo('login');
+    }
+  }
+}
+```
+
+## How we teach this
+
+Should #499 be accepted, this is a fairly straightforward and complementary change that would reduce confusion from the `model()` hook passing in a `@model` argument to a component. The Guides would need to be updated to use `load()` instead of `model()`, `beforeModel()`, and `afterModel()`.
+
```
simonihmig: The route blueprint should be updated as well.

tleperou:
> The Guides would need to be updated to use `load()` instead of `model()`, `beforeModel()`, and `afterModel()`.

Is there a replacement of `beforeModel` and `afterModel`? They are frequently used to grant access (or not) to routes.

tleperou:
Put into the perspective of the of this proposal, [controller-migration-path](https://github.com/NullVoxPopuli/rfcs/blob/controller-migration-path/text/0000-controller-migration-path.md) and the explicitness' trend*, may I wonder myself on the impact and the adoption of the `load` hook?

It would make sense to get _in fine_ under `/routes` basic components. Hence, `render modifiers` can be explicitly performed to handle the `model`, `beforeModel`, `afterModel`, `activate`, `deactivate` hooks. This would reproduce the same behavior of Controllers and Routes, right?

(*) [Add queryParams to the router service](https://github.com/NullVoxPopuli/rfcs/blob/router-computed-query-params/text/0000-router-service-computed-query-params.md), Khttps://github.com/nullvoxpopuli/rfcs/blob/explicit-dependency-injection/text/0000-explicit-dependency-injection.md, probably others

text/0000-route-data-loading.md
NullVoxPopuli: Hang on now, that's not totally fair. The native js way to do parallel is:

```ts
let [profile, comments] = await Promise.all([
  this.store.findRecord('profile', profile_id),
  this.store.query('comments', { profile_id })
)];

return { profile, comments };
```

simonihmig: Sure. I was not saying it's not possible to do natively, just not in the way the original example was. :)

michaelrkn: I think we should try to stick to native JS if possible, but the `Promise.all` approach feels verbose. What it we just let the user provide an object, and if any of the values are promises, then the `load()` hook will wait for them to resolve before rendering the component (much like the `model()` hook currently does)?

text/0000-route-data-loading.md
michaelrkn: > It will continue to wait for returned promises as model() does, affecting the loading state as well as the rendering of the component, as proposed by RFC499. "while this hook is run" is not very precise in this regard.

That's the intent. I will change this to "The route is in the loading state while this hook is run. When this hook has finished running and any promises it returns have resolved, the route will change to the resolved state."

> I assume the arguments will be the same as for model()?

Yes, good clarification. I will add that.

> Will load() always be called, even when the "model" is supplied in the link-to. I.e. how does it relate to #283

I wasn't previously familiar with #283, but I think the proposal is good. Maybe this is a good opportunity to achieve the goals of #283 without introducing a new base class and the potentially-confusingly-similar `findModel` hook. I will update this to specify that the `load` hook will always be run, and suggest that this subsume #283. As a follow-up to this RFC, we also might want to reconsider the word "model" for `link-to`...

> The model hook has a default implementation that will make loading work magically in simple scenarios without actually implementing the hook, see http://api.emberjs.com/ember/release/classes/Route/methods/model?anchor=model. Do we want to copy over this (too) magic behaviour? I think the new load() API proposed here would be a good way to get rid of this, and force users to make everything explicit!

Agreed - I will specify that `load()` will not support the "magic" behavior.

> The "keys" are what Object.keys() would return, i.e. only the own (hasOwnProperty) props, and nothing from prototypes, right?

Righto. I will be more explicit about this.

> What happens when the returned (resolved) value is not an object? (a primitive value, or an array) How is the route component called in this case? Or is the error component rendered? Or does it simply throw an exception?

I think we throw an exception. The `error` state is for errors with the data that is being fetched like being unauthorized, but in this case, the error is a mistake in the front-end code.

michaelrkn:
> Is there a replacement of `beforeModel` and `afterModel`? They are frequently used to grant access (or not) to routes.

@tleperou The intent is that such behavior could be taken care of in the `load` hook. I will be more explicit about this.

> It would make sense to get in fine under /routes basic components. Hence, render modifiers can be explicitly performed to handle the model, beforeModel, afterModel, activate, deactivate hooks. This would reproduce the same behavior of Controllers and Routes, right?

I'm not totally clear on your question here, sorry!

text/0000-route-data-loading.md
michaelrkn: If I'm not mistaken, the current route blueprint doesn't include the `model()` hook. Do you think it should include a `load()` hook, maybe that returns an empty object, or just keep things the way they are?

simonihmig: > If I'm not mistaken, the current route blueprint doesn't include the model() hook.

Oh, I was expecting that tbh, but you're right, it does not have it. Nevermind.

NullVoxPopuli: really, we should probably include the hooks in the route blueprint. It's easier to delete code than it is to try to find out what's possible

text/0000-route-data-loading.md
NullVoxPopuli: @michaelrkn, does the model hook do that if you declare it as async?

```ts
async model(params) {
  let requiredData = this.store.query('something',  { params });

  return {
     foo: requiredData.promiseReturningThing,
     bar: requiredData.someOtherPromise
  }
}
```

I think RSVP.hash has a special behavior, because it wraps the object in a _single_ promise.
I don't think model works with an object of promises.

Really, what we need is native Promise to implement `.hash`

text/0000-route-data-loading.md
michaelrkn: I've added the blueprint suggestion to the RFC.

text/0000-route-data-loading.md
michaelrkn: I'm thinking that we just wrap whatever is returned from `load()` in an RSVP hash (or a native equivalent), something like this:

```js
function load() {
  return {
    foo: this.store.findAll('foo'),
    bar: "bar data"
  };
}

async function whateverCallsLoad() {
  let data = load();
  await hash(data);
  // render component, etc.
}
```

Then the user doesn't have to get hung up on how to deal with promises - we just take care of them in the implementation (much like how the `model` hook currently does).

NullVoxPopuli: I guess you can try it out here: https://github.com/emberjs/ember.js/pull/18173

relevant change here: https://github.com/emberjs/ember.js/pull/18173/files#diff-7e252e0880a2b1de772cbcabaf3ddfb4R1195

michaelrkn: I started to update this PR, but I realized there's a use case this doesn't help with. For simply loading data, we're fine:

```js
load({ profile_id }) {
  return {
    profile: this.store.findRecord('profile', profile_id)
  }
}
```

And for redirecting before loading data we're fine:

```js
@service router;
@service user
load({ profile_id }) {
  if (this.user.loggedIn) {
    return {
      profile: this.store.findRecord('profile', profile_id)
    };
  } else {
    this.router.transitionTo('login');
  }
}
```

But for redirecting after loading data, you have to break the pattern:

```js
load({ profile_id }) {
  let data = {
    profile: this.store.findRecord('profile', profile_id)
  };
  if (data.profile.admin) { // this doesn't work, as the promise isn't resolved
    return data;
  } else {
    this.router.transitionTo('accessDenied');
  }
}
```

I'm a bit torn. We could consider this an edge case, and have the Guides tell people to use RSVP.hash in this situation (or native Promise.all). Or we could tell people to use RSVP.hash in all situations, which would be more explicit and make it straightforward for people to figure out how to handle this case, but which would require everybody learning about an Ember-specific tool. Thoughts?

NullVoxPopuli: sounds like your a step away from bringing back before/after model :p

michaelrkn: Ha, it crossed my mind. I don't think beforeModel solves anything. I think I'd prefer RSVP.hash or Promise.all to afterModel, but I'm open to other opinions.

michaelrkn: Okay, after thinking on this for a bit, I'm right back where we started - I think RSVP.hash is the way to go. It doesn't have any "magic" that the user needs to learn, and it's got a concise syntax that will make sense to most people. I'd like to use a more descriptive name than "hash", though, that gets away from it's Ruby roots and better describes what it's doing. I'm going to go with `resolveKeys` for now, but I'm open to other ideas.

NullVoxPopuli: Ruby probably got it from https://en.wikipedia.org/wiki/Hash_table 😉

text/0000-route-data-loading.md
```diff
@@ -0,0 +1,87 @@
...
+The `load()` hook will receive the same arguments as `model()`.
+
+The route is in the `loading` state while this hook is run. When this hook has finished running, the route will change to the `resolved` state (and the `render` component will be rendered).
+
+The keys of the object (what `Object.keys()` would return, i.e. only the own (`hasOwnProperty`) props, and nothing from prototypes) returned from `load()` become the named arguments to the component.
+
+```js
+// routes/profile.js
+import { hash as resolveValues } from 'rsvp';
+export default class extends Route {
```
efx:
```suggestion
export default class ProfileRoute extends Route {
```

text/0000-route-data-loading.md
```diff
@@ -0,0 +1,87 @@
...
+
+As suggested in [Always run model hook #283](https://github.com/emberjs/rfcs/pull/283), the `load()` hook will always be run, even if a model is supplied to `{{link-to}}`.
+
+The `load()` hook will not support the "magic" behavior of the model hook, in which it will make loading work magically in simple scenarios without actually implementing the hook (http://api.emberjs.com/ember/release/classes/Route/methods/model?anchor=model).
+
+The `load()` hook can also take responsibility for redirection:
+
+```js
+// routes/profile.js
+export default class extends Route {
```
efx:
```suggestion
export default class ProfileRoute extends Route {
```

tleperou:
@michaelrkn, I think that [render modifiers](https://github.com/emberjs/ember-render-modifiers) can replace the hooks, with those equivalents:
* activate, beforeModel, model: `will-render` (illustrated below)
* deactivate: `will-destroy`

Then, routes render a component, only (simplifying the APIs).

It may be out of the scope of this RFC, though :sweat_smile: 

```hbs
{{! templates/profile.hbs }}
<div {{will-render (perform this.load)}}>
  {{! ... }}
</div>
```

```js
// routes/profile.js
@service store;
@service queryParams;

@task load = function*() {
  const { profile_id } = this.queryParams 
  try {
    this.profile = yield this.store.findRecord('profile', profile_id);
    if(this.profile.admin) {
      this.router.transitionTo('accessDenied');
    }
  } catch(e) {
    // ...
  }
}
```

michaelrkn:
@tleperou Interesting! I don't think I'm qualified to weigh in on the advantages of loading data from the route versus from the component, but I'd be happy to close this if people more qualified than I am believe that's the best way forward.

webark:
For me, the main advantages of a route, and loading data from the route, is the clear communication it provides. If i’m encounter a bug, or i want to make additions to a feature, even if i have never worked in a project before, i can go to the router, see what route i’m on, then go to the route (and it’s heirchy), and find out everything that’s been loaded up, all the data i have available, and have a pretty good  understanding of how the data is handled in the app. (which is a reason why i am a fan of route actions too, but that’s another battle).

Whenever we have moved a chunk of loading to the components (generally due to ember concurrency) it has always been a source of confusion and incoherent patterns. (and often added to these insert elements, inits, will render hooks, observers 🤢)

Even though sometimes simpler and potentially even more elegant solutions can be devised with putting the loading directly on a component, you sacrifice that clear communication.

The thing i enjoy about moving from “model” to “load” is that it is more explicit about what is happening, it’s easier to explain to newcomers, and it avoids the awkward feeling when you load multiple items into one model hook, and then need to set aliases in the controller.

the requiring of an object i don’t love. There are many times where i just need a single item in a route. Something like a default of “data” or an equivalent could be nice. (i could see some awkward instances where someone returns an actual data object and it gets splattered all over the rendering component. (splattered sounded more dramatic then spread). 2¢

tleperou:
I think I am out of the scope of the RFC which steps on in the simplification of the route's API. @webark 

These,

> For me, the main advantages of a route, and loading data from the route, is the clear communication it provides

> The thing i enjoy about moving from “model” to “load” is that it is more explicit about what is happening, it’s easier to explain to newcomers, and it avoids the awkward feeling when you load multiple items into one model hook, and then need to set aliases in the controller.

sounds utterly true for me too.

For years now, Ember spotlights the `component`. Such as, [Create Router Service](https://github.com/emberjs/rfcs/pull/95), [Deprecate Route render APIs](https://github.com/emberjs/rfcs/blob/master/text/0418-deprecate-route-render-methods.md) and [An alternative to Controllers](https://github.com/emberjs/rfcs/pull/499) (plus others I missed) which get features off of Route/Controller from Ember.

From my point of view, the "clear communication [Ember] provides" through the different application's layers is essential. The route makes the app a **Web** application. Indeed, we use an URL.

Considering a route as a URL defined into the router.js which instantiates a Component, would provide the same features to developers with different development experience (no Controller, no Route):
* (pre/post) process local or remote data: `@action`, `@task` do the job
* Interact with the hooks of the route:  the Component's lifecycle and render modifiers fit with
* Provide a clear communication: keep the file structure
* Workaround the query parameters
* handle error and loading states: as we already do in components

Sounds evident there are some caveats but it fits the initial motivation that @michaelrkn, @chadhietala, and @NullVoxPopuli described. It hugely simplifies the APIs and the way to teach.

NullVoxPopuli:
For the record, I did react for 3 years where all you had was components. I'm against replacing routes' mode lifecycles with components. :/
Data loading is special enough to be it's own thing.

text/0000-route-data-loading.md
buschtoens: FWIW the broader (non-RSVP) JS community uses [`p-props`](https://github.com/sindresorhus/p-props).

Leooo:
My opinion is that naming the Ember route _loading_ hook "model" is a very conscious choice to suggest that the default object to be yielded by this hook should in general be a part of the domain model - typically an ember-data model.

A book route should yield a book model to the templates. Just MVC where the routes are the C and should yield (in general) ONE model to the view layer.

@samselikoff is saying it brilliantly in [this article](https://medium.com/@sam.selikoff/because-returning-a-single-domain-object-from-the-route-is-the-ember-pattern-the-very-name-of-the-94c4abf4ad58). Were this hook to be renamed `loading`, I would be very worried to see even more devs yielding huge `RSVP.hash({})` objects in their routes instead of leveraging the model layer, which is in my opinion strongly a bad pattern.

mehulkar:
My first thought at reading of a `load` hook was also to load specific components required by the route. (i.e. lazy loading, code splitting). I'm assuming this would already be *possible* with dynamic imports and ember-auto-import, but might be a good idea to mention it in this RFC as well.

michaelrkn:
> My first thought at reading of a `load` hook was also to load specific components required by the route. (i.e. lazy loading, code splitting). 

@mehulkar This RFC builds off of #499, which creates an API to load a single component from the route. If it doesn't seem clear in the context of #499, can you comment/PR to help me improve the clarity? Thanks!

webark:
@michaelrkn did you ever toss around just a “data” hook rather then “load”?

michaelrkn:
> did you ever toss around just a “data” hook rather then “load”?

@webark Yes, but I wanted to base this off [@chadhietala's proposal](https://gist.github.com/chadhietala/50b977a7d3476069892d351c65af418c) where he proposes that this hook also take responsibility for redirection, which I think is a good idea. `data` doesn't convey that redirection could occur here. `load` also matches with the `loading` state in #499.

webark:
oh, and @michaelrkn did you ever read over the discussion around https://github.com/emberjs/rfcs/pull/97 ?

It’s a little tangential, but there’s some good discussion around the semantics of loading data in a route.

michaelrkn:
@webark Thanks for pointing that out! Do you think the proposed `load` hook should have the parallel loading behavior suggested in #97? Are there any other takeaways?

michaelrkn:
Looks like things are going in a different direction and this isn't needed: https://wycats.github.io/polaris-sketchwork/routing.html

PR #512, opened 2019-07-08
Release a major version when any deprecations are removed

Kerrick:
[Rendered](https://github.com/Kerrick/rfcs/blob/intimate-api-changes-cause-major-version/text/0000-intimate-api-changes-cause-major-version.md)

---

This was put together after a discussion on the Ember Community Discord. Thanks to @mixonic, @jenweber, @NullVoxPopuli, @acorncom, @rwjblue, and @samselikoff for their feedback as I discussed the problem and drafted the RFC!

### DIFF
--- a/text/0000-intimate-api-changes-cause-major-version.md
+++ b/text/0000-intimate-api-changes-cause-major-version.md
@@ -31,15 +31,13 @@ This _will_ mean more major versions get released than today; but since Editions
 
 This proposal also does not suggest any changes to the community's current vocabulary. Intimate APIs are still Intimate APIs, and are considered different from Public APIs. Like today, not all Intimate API changes would warrant a deprecation warning (and therefore a major version change). Intimate APIs can still be taught as a thing to be avoided. The big change here is that in today's world, those who choose not to avoid Intimate APIs are punished by minor version upgrades. In the proposed world, they instead have to deal with breaking changes in a major version but everybody else gets that major version "for free."
 
-Finally, this proposal also does not suggest removing deprecated public APIs on a faster (calendar-based) timeline than we currently have. Many features are marked for deprecation "until 4.0.0" -- but that's a minimum guarantee, not a maximum. If an Intimate API removal causes a version 4.0.0 to be released before the community is ready for the deprecated public features to be removed, then don't remove them yet. Nothing about SemVer requires every major release to 
-
 ## How we teach this
 
 In a way, this is already largely what we teach publicly. Between the language on the [deprecations site](https://deprecations.emberjs.com/), and the content of the [Editions page](https://emberjs.com/editions/), this is how many people who are _not_ active participants in the Ember RFCs process might already expect things to work. And they are a large bulk of the Ember Community -- I work with people who have used Ember for years and have never heard of RFCs or Intimate APIs.
 
 The harder part might be teaching core team members and other heavily-enfranchised individuals. Upon reading the title and/or summary of this RFC, they may get gut reactions of visceral fear that this will slow down development (it shouldn't; just release more major versions) or cause lots of teaching problems (it shouldn't; teach "octane" paradigms not "v3" paradigms).
 
-The important thing when teaching this is to stress that _everything is the same as it has been, except that you bump a different integer when releasing any time a deprecated feature is removed_. **Whether to deprecate an Intimate API removal?** Same as before. **When to release LTS versions?** Same as before. **How many months to leave deprecated features in the framework?** Same as before. (etc.)
+The important thing when teaching this is to stress that _everything is the same as it has been, except that you bump a different integer when releasing any time a deprecated feature is removed_. **Whether to deprecate an Intimate API removal?** Same as before. **When to release LTS versions?** Same as before. (etc.)
 
 ## Drawbacks
 
--- a/text/0000-intimate-api-changes-cause-major-version.md
+++ b/text/0000-intimate-api-changes-cause-major-version.md
@@ -1,0 +1,58 @@
+- Start Date: 2019-07-08
+- Relevant Team(s): Steering, Ember.js, Ember Data, Ember CLI, Learning
+- RFC PR: (after opening the RFC PR, update this with a link to it and update the file name)
+- Tracking: (leave this empty)
+
+# Intimate API Changes That Warrant Deprecations Should Cause Major Version Releases Upon Their Removal
+
+## Summary
+
+When [Ember released its first LTS version](https://blog.emberjs.com/2016/02/25/announcing-embers-first-lts.html), the concept of [editions](https://emberjs.github.io/rfcs/0364-roadmap-2018.html) did not yet exist. Major versions were responsible both for [maintaining backwards compatibility](https://semver.org/) _and_ for [marking a new paradigm](https://github.com/emberjs/rfcs/blob/9c7fe3f4e947b5f79050214334a98673494c25d7/text/0000-editions.md). Today, the latter problem belongs to editions so major version releases no longer need much ceremony. Instead, they can be relegated to marking backwards-incompatible changes.
+
+[Intimate API](https://twitter.com/wycats/status/918644693759488005) changes already [undergo an analysis that sometimes results in deprecation messages](https://blog.emberjs.com/2016/02/25/announcing-embers-first-lts.html) before removal. I propose that in the future, any Intimate API removal that would get a deprecation warning should trigger a major version release upon its removal.
+
+## Motivation
+
+Ember.js has followed SemVer for many years. However, SemVer is incomplete in its definition of the term "public API." A while ago, the Ember community identified a gray-area between public and private APIs, which it dubbed "Intimate APIs." These are the parts of Ember.js that are being used by the community, but don't exist (at least completely) in its documentation.
+
+Ember's current release process allows backwards-incompatible changes to Intimate APIs to happen in a minor release. But since the very definition of an Intimate API means that people are using it, this runs the real risk of breaking a production application. Even with the current policy of a deprecation existing for at least one LTS cycle, the following tragic scenario is possible:
+
+> Chandra has an Ember.js application. Her company doesn't follow every best practice out there, so they don't use `package-lock.json` or `yarn.lock` to pin dependencies. Instead, their build system grabs exactly what is specified in `package.json`.
+>
+> Chandra reads [Ember's documentation](https://deprecations.emberjs.com/) and finds this: "Until a major revision such as 2.0 lands, code firing such deprecations is still supported by the Ember community. After the next major revision lands, the supporting code may be removed."
+>
+> Thus, Chandra feels safe setting `"ember-source": "^3.0.0"` in her package.json, comforted by the presence of SemVer to keep her application safe. She deploys her company's application on June 28, 2018. Her app runs on v3.2.2 and all is well. In her `beforeModel` hook, she reads `transition.targetName` (found in [the official Ember documentation](http://api.emberjs.com/ember/3.10/classes/Route/methods?anchor=resetController )) and `transition.queryParams` ([quite](http://shotgundebugging.blogspot.com/2019/02/impersonation-in-emberjs-elegantly.html) well [documented](https://codeandtechno.com/posts/user-impersonation-ember-simple-auth-doorkeeper/) by [community](https://discuss.emberjs.com/t/getting-query-params-and-segment-values-from-parent-route/6628/6) blog [posts]( https://www.tilcode.com/tag/ember-queryparams-tutorial/) and [answers](https://stackoverflow.com/a/26706095) all [over](https://stackoverflow.com/a/43310476)).
+> 
+> Chandra doesn't need to deploy again for a full year, as the application was developed perfectly from the start. But on July 4, 2019 the production server goes down and an automated build process checks out the repository, resolves `"ember-source": "^3.0.0"` to v3.11.0, and [her production application breaks](https://github.com/emberjs/ember.js/pull/17843)!
+>
+> Chandra never saw the [deprecation warnings](https://deprecations.emberjs.com/v3.x#toc_transition-state), because she never upgraded to v3.6. And now she has to leave her USA Independence Day celebration early, putting down the fireworks with family to fight metaphorical fires at work.
+
+## Detailed design
+
+In the current process, Intimate API changes already undergo a vetting process to see if they warrant deprecation warnings. This proposal does not suggest any changes to that process. Instead, it suggests that the outcome of the answer, "yes, this needs a deprecation," changes from, "so don't remove it until the next minor release after an LTS," to "so trigger a major version change when it releases."
+
+This _will_ mean more major versions get released than today; but since Editions now drive paradigm shifts and marketing efforts, it shouldn't greatly affect how people talk about "what's new in Ember."
+
+This proposal also does not suggest any changes to the community's current vocabulary. Intimate APIs are still Intimate APIs, and are considered different from Public APIs. Like today, not all Intimate API changes would warrant a deprecation warning (and therefore a major version change). Intimate APIs can still be taught as a thing to be avoided. The big change here is that in today's world, those who choose not to avoid Intimate APIs are punished by minor version upgrades. In the proposed world, they instead have to deal with breaking changes in a major version but everybody else gets that major version "for free."
+
+Finally, this proposal also does not suggest removing deprecated public APIs on a faster (calendar-based) timeline than we currently have. Many features are marked for deprecation "until 4.0.0" -- but that's a minimum guarantee, not a maximum. If an Intimate API removal causes a version 4.0.0 to be released before the community is ready for the deprecated public features to be removed, then don't remove them yet. Nothing about SemVer requires every major release to 
+
+## How we teach this
+
+In a way, this is already largely what we teach publicly. Between the language on the [deprecations site](https://deprecations.emberjs.com/), and the content of the [Editions page](https://emberjs.com/editions/), this is how many people who are _not_ active participants in the Ember RFCs process might already expect things to work. And they are a large bulk of the Ember Community -- I work with people who have used Ember for years and have never heard of RFCs.
+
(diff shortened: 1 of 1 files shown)


text/0000-intimate-api-changes-cause-major-version.md
```diff
@@ -0,0 +1,52 @@
...
+> 
+> Chandra doesn't need to deploy again for a full year, as the application was developed perfectly from the start. But on July 3, 2019 she hears from a colleague that the latest 3.X version of Ember has some performance improvements that'll really help during the holiday weekend rush. Comforted by SemVer, she upgrades to v3.11.0 and deploys on the way out the door. Little does she know, [her production application broke](https://github.com/emberjs/ember.js/pull/17843)!
```
rwjblue: > as the application was developed perfectly from the start

😜 INDEED!

text/0000-intimate-api-changes-cause-major-version.md
```diff
@@ -0,0 +1,52 @@
...
+>
+> Chandra never saw the [deprecation warnings](https://deprecations.emberjs.com/v3.x#toc_transition-state), because she never upgraded to v3.6. When the reports start flowing in the next morning, she has to leave her USA Independence Day celebration early, putting down the fireworks with family to fight metaphorical fires at work.
+
+## Detailed design
+
+In the current process, Intimate API changes already get deprecation warnings in some scenarios. This proposal does not suggest any changes to that process. Instead, it suggests that the outcome of the answer, "yes, this needs a deprecation," changes from, "so don't remove it until the next minor release after an LTS," to "so trigger a major version change when it releases."
+
+This _will_ mean more major versions get released than today; but since Editions now drive paradigm shifts and marketing efforts, it shouldn't greatly affect how people talk about "what's new in Ember."
```
rwjblue: > This _will_ mean more major versions get released than today

I don't think this is necessarily true (though IMHO it's fine if it is true). I think this may well be the forcing function for "svelte" which has long been discussed (and is internally supported in the codebase) but is not actually usable by apps today.

The _rough_ pitch for svelte is that deprecated codepaths are flagged (nearly identically to new features!), and the application would declare which version of Ember it is compatible with "deprecation free". This allows the build system to remove all deprecated code paths for features that the application has declared that it doesn't use. This dramatically reduces the "need" to remove deprecations since it goes from "we have to remove this to shave bytes" to "we generally would like to remove this so we can have cleaner code in the repo".

text/0000-intimate-api-changes-cause-major-version.md
```diff
@@ -0,0 +1,52 @@
...
+
+This proposal also does not suggest any changes to the community's current vocabulary. Intimate APIs are still Intimate APIs, and are considered different from Public APIs. Like today, not all Intimate API changes would warrant a deprecation warning (and therefore a major version change). Intimate APIs can still be taught as a thing to be avoided. The big change here is that in today's world, those who choose not to avoid Intimate APIs are punished by minor version upgrades. In the proposed world, they instead have to deal with breaking changes in a major version but everybody else gets that major version "for free."
```
rwjblue: This seems a tad "muddy" to me, there are three types of APIs:

* Public APIs - any API documented as public; these already can't be removed until a major version bump
* Private APIs - any API that is not explicitly marked as "public" should be considered private; these can be removed at will, and are not considered "breaking" in the least bit
* Intimate APIs - this one is hard to define 😸; these would be Private APIs that have enough community usage to force us to think carefully about removing arbitrarily

Specifically, sections like:

> Like today, not all Intimate API changes would warrant a deprecation warning (and therefore a major version change)

I think this is slightly wrong, if an API is "intimate" it probably _does_ need a deprecation warning. I'd rephrase this to be something like:

> Like today, not all private API changes would warrant a deprecation

rwjblue:
Deprecations are added in order to be _helpful_ to folks that might be using a given API. They are definitely not something that is "required" at all for private APIs. Adding a policy that intentionally penalizes folks for trying to be helpful seems like the complete opposite direction that you are trying to push for, and would almost certainly result in _fewer_ deprecations and therefore **more pain** for addon and app authors.

wycats:
Thanks for the well-considered RFC!

You're right that the "intimate API" policy is fairly unusual. It can be difficult for people to understand unless they're deeply involved with the Ember planning process.

In a way, this RFC proposes that we have a process to upgrade private APIs to public APIs instead of the compromise policy we have now.

This feels like a reasonable approach to me, and one that I would support.

This RFC (as written), suggests that we would never issue deprecations for private APIs, because the act of issuing a deprecation means that an API is public. This might be going too far.

I would support replacing our current Intimate API policy with a process for marking a private API as public. I would also support retaining the ability to deprecate private APIs.

rwjblue:
> The harder part might be teaching core team members and other heavily-enfranchised individuals.

There is a specific tone here that I find troubling. The onus of this RFC absolutely falls on the project maintainers and core team members, and this phrasing makes it seem that the fundamental goals of core team members are not aligned with whats good for the users of the project. If that is true then we have a **very big problem**.

runspired:
I have three concerns:

1) I suspect this encourages more private API usage
2) how this interplays with lockstep may cause release cadence issues for other teams
3) it suggests that "major versions for deprecation cleanup" as an alternative, but practically that is the only real solution. Were we instead to have any sort of "every X" version or worse "every Nth from when a new deprecation is introduced" we would find ourselves with major versions for most releases. This RFC should recommend a cadence for deprecation cleanup.

Kerrick:
I do not want to suggest anything that makes it _less_ likely for an intimate API to become deprecated. That's why this RFC **does not** suggest keeping deprecated APIs around longer.

If the RFC wasn't clear on this point, I'd love some advice on how to make it clear.

Essentially, I want to introduce no changes to process besides a change to whether the Minor or Major version number gets bumped. Things I don't want include:
- Changing how decisions are made on which private/intimate APIs to deprecate
- Changing whether deprecations are added for private/intimate APIs.

runspired:
I think maybe I should clarify the points I'm making a bit more:

> 1. I suspect this encourages more private API usage

Today, if your app or addon uses private API you understand that you are taking on a support risk. APIs become intimate only once enough app's / addon's have taken this risk, but even there it was an acknowledged risk. With this RFC it becomes somewhat easier for addons to abuse private not-previously-intimate APIs with the understanding that by doing so they are forcing support and a major version.  This may not be a huge risk, and I think most developers wouldn't be so cruel, but I do think it is a risk since the disincentives of abusing private APIs are lessened.

This may be a risk we happily take on to reduce the risks of "semver drift" still breaking apps that were depending on intimate API.  That said, we still have the case of addons like model-fragments, which was presented a public API to replace an intimate API and which refactored instead to use a new private API. E.g. this risk I see is not without precedent.

> 2. how this interplays with lockstep may cause release cadence issues for other teams
> 3. This RFC should recommend a cadence for deprecation cleanup.

Today deprecations are removed:

- After a major release (intimate or public)
- After an LTS release (intimate)

Keep in mind there are multiple core libraries all with their own teams to coordinate and intimate and public API deprecations to handle. 

Currently this RFC suggests issuing a major any time a deprecation is removed, but it doesn't specify how long a deprecation should live. Currently deprecations are set to live until the next LTS or next major, but if a major is determined by deprecation removal this policy would result in every release after an LTS release becoming a major release. That would seem to void the value of LTS releases, at least in their current every-4th-release form.

Alternatively a policy could be that deprecations must live for at least-N versions (where N is some number >= 4 in all likelihood).  But a policy like that would struggle from every single release becoming a major release, because new deprecations are introduced in most releases.

Formalizing a cadence for deprecation removal would avoid issues with teams forcing majors either unexpectedly or too often on other teams.  The cadence should be regular and consistent, and ideally not too long nor too short in between. Too short, we eliminate the value of LTS releases. Too long, the risk of a team missing being ready for a major with an important deprecation effort becomes too high.

maxbogue:
Reading through this RFC and PR, it seems to me like "Intimate APIs" are the crux of the problem. The perspective I see in the comments here seems to assume that people are perfectly aware of the difference between public and intimate APIs and are taking on that risk willingly.

I would wager the real danger that this RFC is trying to address is that most people using intimate APIs have absolutely no idea that they are doing so, and therefore are completely blindsided by a minor version bump breaking their application.

I'm not sure what the options are to correct that, but a separate change to clarify what is public and what is private somehow (a naming convention like a prefixed `_` in Python?) may be worth considering to address the issue of intimate APIs entirely.

rwjblue:
The line between public and private is fairly clear anything that is documented **must** be documented as either `@public` or `@private` (enforced by documentation linting), and anything undocumented is private.  I totally agree that it may not be clear when hacking around in the debugger, and preferring underscored method names is absolutely a good idea (which I _think_ we try to do?).

jelhan:
> Currently deprecations are set to live until the next LTS or next major, but if a major is determined by deprecation removal this policy would result in every release after an LTS release becoming a major release. That would seem to void the value of LTS releases, at least in their current every-4th-release form.

@runspired Could you elaborate a little bit more why this would void the value of LTS releases? I don't get that point. Of course such a policy would mean that deprecation of public APIs must life for more than one major release cause they shouldn't be removed that often as intimate APIs but I having a fixed end of life for deprecated public APIs would also make planing easier.

I could see some good arguments for a policy like:

- Every 4th release bumps the major version and removes deprecated APIs (intimate and public).
- Every release before a major version is bumped is declared an LTS release.
- An intimate public API must be deprecated at least one minor release before it's removed.
- A public API must be deprecated at least four two major releases before it's removed.

This would ensure that every intimate API is deprecated at least six weeks (1 release) and every public API at least ~ 1 ½ years (2 major releases + 1 release = 12 + 1 releases ≈ 78 weeks) before they are removed. Having a date as end of life than a vague `v4` would also be more helpful for planing.

maxbogue:
@runspired I'm also having trouble following your logic. You seem to be conflating deprecating something (which can be done as often as needed) and removing deprecated things (what this RFC is suggesting only happen at major version bumps). There's no reason it would affect the LTS schedule, it's just that LTS releases wouldn't remove deprecated things anymore.

runspired:
@jelhan @maxbogue This RFC voids the value of LTS releases because de-facto as-written every release post-current-LTS-cycle becomes a major release.

LTS exists so that there are clearly supported points within the minor release cycle to upgrade between: but if there are only 4 releases between majors then having a specific LTS policy becomes meaningless as there are no minor releases that are LTS releases that aren't simply "the last 1.0/2.0/3.0/4.0" release.  Basically, this RFC as-written has the side-effect of replacing the every 4th release being an LTS with every 4th release being a major: resulting in roughly two major versions per year.

This also means that churn goes way up, because with "GC" releases occurring far more frequently the time between deprecation and removal becomes very short.

The biggest downside is that teams would no longer be upgrading from LTS to LTS with the last LTS before a major being a special and more rare occurrence: instead they would be upgrading from major to major.

**To be very clear I am not opposed to the ideas this RFC is presenting**, but it needs to propose the exact cadence of major releases, and that cadence should not be 1:1 with LTS releases.

Kerrick:
> This also means that churn goes way up, because with "GC" releases occurring far more frequently the time between deprecation and removal becomes very short.

It doesn’t. If a deprecated public API was not intended to be GCed until at least June 2021, nothing says that v4.0 releasing in Feb 2020 due to deprecated private APIs being removed means that public feature has to be GCed. That is based on a flawed premise that ALL deprecations must be GCed on EACH major release.

runspired:
@Kerrick every deprecation receives an `until` major version, they aren't scheduled for a specific date.

Kerrick:
And as I mentioned in the RFC, the “until” date can be a minimum version, not a maximum. They can stick around past that version.

runspired:
Also note, if we started doing majors based on length of time a deprecation exists, we get back to the other problem I pointed out which is that very quickly every release becomes a major version.

Data deprecates something nearly every minor, and we aren't the only library adding deprecations. All libraries adhering to lockstep would be forcing majors.

runspired:
> And as I mentioned in the RFC, the “until” date can be a minimum version, not a maximum. They can stick around past that version.

This policy would introduce confusion, having an implicit agreement that some deprecations will live past their `until` versions would be difficult to manage and explain.

jelhan:
@runspired I'm sorry, but I don't get your point.

We could do majors on a fix schedule. I proposed every forth-release. So there wouldn't be any issue that each release could become a major. That would also be similar to the process currently used for  intermediate APIs. As today they would be removed every forth-release (one after LTS). The changes to the current process would be: a) the version removing intermediate APIs is a major one; b) there is a fixed date for removing public API in opposite to an unscheduled `v4`.

An API could be deprecated nth-major releases in advance. E.g. if we go with four-major releases in advance for public APIs, a deprecation of a public API introduced in `v7.2` would have an `until` of `v11`. If we notice in the process that the ecosystem isn't catching up quick enough, the deprecation's `until` might be changed in `v9.0` to `v13`. I don't think that would introduce any teaching issues.

We would still be able to garbage collect deprecated intermediate APIs quickly if we decide that they could be removed in next major. That would mean that an intermediate API introduced in `v7.3` will be removed in `v8`.

Having garbage collection even for public APIs on a regular schedule would fit perfectly to the _stability without stagnation_ concept.

runspired:
@jelhan it seems you do get my point. A regular cadence of majors is necessary for lockstep to work, for teams to be able to GC, and for users to have clear expectations of when things will be removed. Whether that is every 4th or 8th or something else.

What introduces teaching issues is if we use `until` to implicitly mean different things as was suggested by others. (e.g. a public API "until 4.0" really means until "5.0 or 6.0" where a private API really does mean "until 4.0", that sort of confusion).

My personal recommendation is that "if" we were to keep lockstep then we should do a major release every 8th release. There is a much  more important discussion that this RFC does not acknowledge but which needs to be had and consensus reached upon as to whether lockstep is still worth it and still viable, as it has led to a lot of semver issues (including some of the "breaking" LTS cleanups) and made it difficult for individual teams to plan and execute the removal of deprecations in a timely way.

e.g. a pattern like the following which would roughly equate to two LTS releases and one major release per year.

3.12 LTS
4.0 -> 4.4 LTS -> 4.8 LTS
5.0 -> 5.4 LTS -> 5.8 LTS
6.0

Kerrick:
I had never considered that lockstep releases could be the root cause of the issues that caused me to write this RFC. That’s an interesting perspective...

chriskrycho:
My own take is twofold:

1. "Intimate API" was a bad idea. I know why we did it, but sometimes it's possible to *over*-do trying to help people. When you choose to use private API, that's on you! And it's okay; sometimes it's the right call. But then it's debt to be paid, and you have to understand the tradeoffs you make when you're making that move. That's part of the job!

2. Kill lockstep. We need a new way of keeping track of how things stay in sync, though, to make that work – perhaps Ember CLI Update becomes a core primitive of the ecosystem, and/or packages have to advertise the ranges they support. This is more *honest* in some ways—per @runspired's comments about effectively-breaking changes within LTS releases, esp. in Ember Data—but it's also *harder*. There's a lot of design work to be done there. I think it's the right thing, and likely necessary, but it's *hard*.

Gaurav0:
People have already said most of what I'd like to say, so I'll just generally agree with @chriskrycho and @runspired

I'd like to add the following suggestions:

1. Release 4.0 soonish
2. When deprecating an intimate API, as part of the deprecation process, search for add-ons that use it and publish a list, perhaps in the PR.
3. Announce upcoming intimate api deprecations in places like Ember Times proactively.
4. Make wifely used apis public.
5. If there are any frowned upon public apis that might be deprecated soon, find a way to let people know.

MelSumner:
@chriskrycho I reconciled with intimate API as a necessary thing when you're trying to create something that people use IRL- kind of like the architect who waited to install sidewalks on a campus until he saw how people decided to get from place to place. I wouldn't want it advertised in official documentation, because then it might as well not exist; but I do think it should be  found when driven by genuine need to _do_ something, as it can point out the gaps we have for making useful things (since no one can be expected to know how to make all the things).

Kerrick:
What's the best way to move forward with the ideas presented in this RFC and the discussion on this PR?

acorncom:
@Kerrick Per the discussion of ["The Process"](https://github.com/emberjs/rfcs#the-process), it looks like we're probably at the "find a champion on the relevant core team" stage here ...

lupestro:
If a private API is used widely enough for deprecation to be deemed wise, that is an acknowledgment that it has historically met a broad need that isn't addressed by the public APIs. If so, we need to complete that acknowledgement by declaring it public in the same release in which it is deprecated. Then:
1. we take on the same design difficulties we have supporting a public API that needs to continue to be supported as the world changes around it.
2. we don't have to do anything new to our cadence except hang onto supporting the old behavior until a major release.

Making more frequent major releases will make our problems worse:  
1. We no longer have the confidence that if we see "deprecated" in our development runs, unless we schedule starting work on it now, we're going to be in major pain when the {n+1}.0.0 comes out. That's a tremendous incentive for us.
2. Dealing with even a single deprecation of a practice that we routinely were doing - especially if we weren't aware that we were doing it - can sometimes require peeling off one team member for several months of work on the day the deprecation first appears and we need all the lead time we can get to address it.
3. The whole point of LTS releases is that teams can move their code from LTS to LTS without dealing with the churn of intermediate releases. Hence, we will first see a deprecation triggered in our own code on the first LTS *after* that deprecation first appears.

wagenet:
There is active movement around #830 which may address the issues raised here. I'll leave it open for now but it's possible that it will be superseded.

chriskrycho:
We discussed this RFC at today’s Framework Core meeting and decided to move it into **Final Comment Period to *Close***. (FCP to close means that we will close this next week unless there are substantive *new points* brought up in discussion which are not already represented in the discussion above.) We believe RFC #830 addresses the underlying issues this RFC was pointing at, and does so by addressing them at a more fundamental level in our process, by providing a more regular cadence for dealing with not only private and intimate API but also *public* API.

---

Two really important notes we all agreed we wanted to call out here:

1. This RFC was really helpful in “priming the pump” for the direction we are now heading with #830, and as @jelhan pointed out over there, this discussion was one of the first places where the idea of a more regular cadence come up. Thanks very much to @Kerrick for taking the time to write up the idea and for kicking off the discussion here. Anyone who has written an RFC knows that it is a lot of work!

2. *In general*, we do not see closing an RFC like this as a sign that the RFC was bad, misguided, wrong, etc. Instead, it often means the exact opposite: that a given RFC started a conversation, or got things moving in the right direction, or provided useful background and discussion along the way.

PR #521, opened 2019-07-29
[DATA] findRecord/peekRecord via Identifier

runspired:
[Rendered](https://github.com/runspired/rfcs/blob/fetch-by-identifier/text/0521-find-by-identifier.md#data-findrecordpeekrecord-via-identifier)

### DIFF
--- a/text/0521-find-by-identifier.md
+++ b/text/0521-find-by-identifier.md
@@ -1,0 +1,121 @@
+- Start Date: (fill me in with today's date, 2019-08-29)
+- Relevant Team(s): EmberData
+- RFC PR: https://github.com/emberjs/rfcs/pull/521
+- Tracking: (leave this empty)
+
+# [DATA] findRecord/peekRecord via Identifier
+
+## Summary
+
+Users should be able to peek or find records based on `lid`.
+
+## Motivation
+
+Apps and Addons making use of [Identifiers](https://github.com/emberjs/rfcs/pull/403)
+may wish to peek or find the record for a known `identifier`. This
+RFC would allow them to do so.
+
+## Detailed design
+
+A new method signature would be added to `findRecord`
+`peekRecord` and `getReference`. The existing signatures would not be
+deprecated at this time.
+
+For the case where calling `findRecord` would result in a request
+being necessary but either no `type` or `id` information being known
+we would error.
+
+This is because there is no meaningful way for the store to create a
+network request for a record that does not exist yet (a
+`NewResourceIdentifierObject`), or for one for which the only thing
+we know is a locally assigned id (`Identifier`).
+
+```ts
+interface Identifier {
+  lid: string;
+}
+
+export interface ExistingResourceIdentifierObject {
+  id: string;
+  type: string;
+  lid?: string;
+}
+
+export interface NewResourceIdentifierObject {
+  id: string | null;
+  type: string;
+  lid: string;
+}
+
+export type Peekable =
+  | Identifier
+  | ExistingResourceIdentifierObject
+  | NewResourceIdentifierObject;
+
+class Store {
+  peekRecord(type: string, id: string | number): Record | null {}
+  peekRecord(identifier: Peekable): Record | null {}
+
+  findRecord(type: string, id: string | number, options?: any): PromiseRecord {}
+  findRecord(identifier: Peekable, options?: any): PromiseRecord {}
+
+  getReference(modelName: string, id: string | number): RecordReference {}
+  getReference(identifier: Peekable): RecordReference {}
+}
+```
+
+## How we teach this
+
+For existing usage of these methods no migration or changes are necessary. For folks wishing to use the new
+APIs an example route is below that both peeks a record and makes a findRecord request.
+
+**current style**
+
+```ts
+{
+  @service('store') store;
+
+  model({ user_id, post_id }) {
+    let user = this.store.peekRecord('user', user_id);
+
+    return hash({
+        user,
+        post: this.store.findRecord('post', post_id)
+    });
+  }
+}
+```
+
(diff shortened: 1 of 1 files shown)


text/0521-find-by-identifier.md
```diff
@@ -0,0 +1,95 @@
...
+Apps and Addons making use of [Identifiers](https://github.com/emberjs/rfcs/pull/403)
+may wish to peek or find the record for a known `identifier`. This
+RFC would allow them to do so.
+
+## Detailed design
+
+A new method signature would be added to both `findRecord` and
+`peekRecord`. The existing signatures would be deprecated.
+
+Additionally `findRecord` would return a `Promise` resolving to a
```
rwjblue: I'm definitely 👍 on moving away from `PromiseRecord`, but I'm a bit worried about having the same API emit different conceptual things (a stable object in the non-identifier signature, and a promise in the identifier based signature).

This isn't a super strong objection, and I'm totally willing to believe that this is just one of the ways we move away from the proxying but I wanted to notate the concern so we can discuss it specifically.

text/0521-find-by-identifier.md
```diff
@@ -0,0 +1,122 @@
+- Start Date: (fill me in with today's date, 2019-08-29)
+- Relevant Team(s): EmberData
+- RFC PR: https://github.com/emberjs/rfcs/pull/521
+- Tracking: (leave this empty)
+
+# [DATA] findRecord/peekRecord via Identifier
+
+## Summary
+
+Users should be able to peek or find records based on `lid`.
```
rwjblue: Can you include a succinct definition of "lid" (either here or below in detailed design)? It seems that saying "identifier" would have made more sense to me when I read through it.

text/0521-find-by-identifier.md
```diff
@@ -0,0 +1,122 @@
...
+
+## Motivation
+
+Apps and Addons making use of [Identifiers](https://github.com/emberjs/rfcs/pull/403)
+may wish to peek or find the record for a known `identifier`. This
+RFC would allow them to do so.
```
rwjblue: Minor quibble here, but since the existing APIs are deprecated in this RFC (AFAICT) this doesn't _allow_ folks to use identifiers it _requires_ it...

text/0521-find-by-identifier.md
runspired: we use `identifier` to mean an object with the `lid` property, whereas `lid` is the unique string itself.

text/0521-find-by-identifier.md
runspired: You are misreading the signature

```
ExistingResourceObject | NewResourceObject | Identifier
```

This allows for three distinct scenarios:

- `ExistingResourceObject` would allow for "just type + id" which is the current signature,
- `NewResourceObject` allows you to be able to find again a record you know you have newly created but perhaps not held on to a reference for. It does so by utilizing `lid`.
- `Identifier` allows you to lookup by "just lid" without type or id being required.

text/0521-find-by-identifier.md
runspired: Typescript allows us to specify which signature returns which return-type. Since users would be migrating to the new call signature in a location where they also handle the return, the would be able to account for this change all in one go. I think this is preferable to introducing a fully new API for querying for records at this time.

rwjblue:
We chatted about this in todays ember-data meeting, and had a few take aways:

* Remove mentions of deprecations
* Remove changes of the return value based on the input arguments

Is that a good summary @runspired / @igorT / @dgeb / @hjdivad / @heroiceric?

text/0521-find-by-identifier.md
```diff
@@ -0,0 +1,167 @@
...
+```
+
+API Documentation and Guides would need updated to use these new method
+signatures and to discuss how these APIs fit with Identifiers.
+
+## Drawbacks
+
+- It changes the function signatures of two of the most commonly used
+  EmberData APIs. However it also aligns those APIs better with more
+  recently designed APIs.
```
acorncom: Given that we're considering updating two quite common EmberData apis, do we envision releasing a code mod to help make this change mechanical?

text/0521-find-by-identifier.md
```diff
@@ -0,0 +1,167 @@
...
+```ts
+{
+  @service('store') store;
+
+  model({ user_id, post_id }) {
+    let user = this.store.peekRecord({ type: 'user', id: user_id });
+
+    return hash({
+        user,
+        post: this.store.findRecord({ type: 'post', id: post_id })
```
acorncom: In the scenario that this type of request is being done with an options hash, would that translate to this?

```
this.store.findRecord({ type: 'post', id: post_id}, { preload: { post: 1 } });
```

Or are we figuring that the main hash passed in might also optionally have an `options` key on it?

mikkopaderes: Curious with this one too as an adapter I made makes use of the `adapterOptions`.

text/0521-find-by-identifier.md
runspired: Removed for deprecation at a later time

text/0521-find-by-identifier.md
runspired: we decided to not deprecate the old signatures for now as well

text/0521-find-by-identifier.md
runspired: we could at such time as we deprecate the existing signature, but since we have decided not to this isn't something we need now

text/0521-find-by-identifier.md
runspired: updated to show the extra arg remains

runspired: updated to show the extra arg remains

PR #522, opened 2019-07-30
RFC - deprecate default serializers and adapters

pete-the-pete:
[Rendered](https://github.com/pete-the-pete/rfcs/blob/master/text/0000-default-serializers-and-adapters.md)

### DIFF
--- a/text/0000-default-serializers-and-adapters.md
+++ b/text/0000-default-serializers-and-adapters.md
@@ -1,0 +1,49 @@
+- Start Date: 2019-07-27
+- Relevant Team(s): Ember Data
+- RFC PR: https://github.com/emberjs/rfcs/pull/522
+- Tracking:
+
+# Deprecate default Adapter and Serializer fallbacks
+
+## Summary
+As part of Project Trim, https://github.com/emberjs/data/issues/6166, this deprecates the fallback and default adapter and serializer across ember-data including:
+1. deprecate -default serializer fallback in Store.serializerFor
+2. deprecate `adapter.serializer` and `adapter.defaultSerializer` instance property fallbacks (which currently default to -json-api).
+3. deprecate `store.defaultAdapter` instance property (which defaults to `-json-api`) and the `-json-api` adapter fallback behavior in `adapterFor`.
+4. deprecate `record.toJSON` instance method since this relies on the `-json` serializer.
+
+## Motivation
+
+The adapter and serializer packages provide reference implementations and base classes that are not required for applications that implement their own following the required interfaces for adapters and serializers as defined in their respective base classes.  Deprecating them allows us to simplify the lookup pattern and remove automatic injection and registration of potentially unused classes.
+
+In addition to removing use of initializer injection, this takes a significant step toward simplifying the mental model for how to determine what adapter/serializer is in use. Removing the defaults forces app developers to be more cognizant about the type of application level concerns vs model-specific concerns; they will now need to explicitly define and use specific adapters/serializers. After this deprecation RFC lands, apps will always use an adapter/serializer explicitly put into your application and the rule will be "the adapter matching modelName falling back to application".
+
+## Detailed design
+The injection of `-default` and `-json-api` serializers will be removed in the next major version (4.0). Since this changes some core assumptions we will deprecate the reliance on the existence of the defaults. All deprecation warnings will only be shown in Dev mode.
+
+##### deprecate -default serializer fallback in store.serializerFor
+A deprecation warning will be shown when [no model, application or adapter serializer](https://github.com/emberjs/data/blob/67affb0eca7048a1a0edc856af46d1305cd1fc1d/packages/store/addon/-private/system/store.ts#L2909) has specified and the default must be used. We will recommend implementing an application serializer.
+
+##### deprecate adapter.serializer and adapter.defaultSerializer fallbacks
+A deprecation warning will be shown when accessing the [adapter's defaultSerializer](https://github.com/emberjs/data/blob/67affb0eca7048a1a0edc856af46d1305cd1fc1d/packages/store/addon/-private/system/store.ts#L2896). This will be distinct from the warning about using the application level default. We will recommend implementing an application serializer.
+
+##### deprecate store.defaultAdapter (-json-api) and the -json-api adapter fallback behavior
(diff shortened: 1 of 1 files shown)


text/0000-default-serializers-and-adapters.md
```diff
@@ -0,0 +1,53 @@
...
+
+In addition to removing use of initializer injection, this takes a significant step toward simplifying the mental model for how to determine what adapter/serializer is in use. Removing the defaults forces app developers to be more cognizant about the type of application level concerns vs model-specific concerns; they will now need to explicitly define and use specific adapters/serializers. After this RFC lands, apps will always use an adapter/serializer you explicitly put into your application and the rule will be "the adapter matching modelName falling back to application".
+
+## Detailed design
+The injection of `-default` and `-json-api` serializers will be removed in the next major version (4.0). Since this changes some core assumptions we will deprecate the reliance on the existence of the defaults. All deprecation warnings will only be shown in Dev mode.
```
efx: could we link to this RFC / decision?

efx: is dev mode the development version of `ember build`? i.e. ember's default.

text/0000-default-serializers-and-adapters.md
```diff
@@ -0,0 +1,53 @@
...
+
+##### deprecate Store.defaultAdapter (-json-api) and the -json-api adapter fallback behavior
+A deprecation warning will be shown when the [defaultAdapter](https://github.com/emberjs/data/blob/b0cf3225662bfb806cd0c02b55b763e37a319b32/packages/store/addon/-private/system/store.ts#L309) is accessed.
+
+##### deprecate Model.toJSON
+A deprecation warning will be shown when toJSON is called since it uses a serializer to create a JSON representation of [the model](https://guides.emberjs.com/release/models/customizing-serializers/#toc_customizing-serializers). The message will suggests using a `debugData` method to accomplish the same goal.
+
+## How we teach this
+
+Today we have extensive documentation about creating custom serializers, but we will need to update the guides to specify the desired serializer in [app/serializers/application.js](https://guides.emberjs.com/release/models/customizing-serializers/#toc_customizing-serializers)
```
efx: My question applies to both serializers and adapters but am using the word adapters for brevity.
What practices should we encourage for application authors that do need various versions of adapters?
The guides recommend extending a base adapter and then implementing the hooks you need in each subclass. In this proposed update would we recommend writing independent functions within the application adapter?
```js
// app/adapters/application.js
import DS from 'ember-data';

export default DS.JSONAPIAdapter.extend({
  urlForQueryRecord({ slug }, modelName) {
    let baseUrl = this.buildURL();
    if (modelName === 'user') {
      return `${baseUrl}/other-place/${encodeURIComponent(slug)}`;
    }
    return `${baseUrl}/${encodeURIComponent(slug)}`;
  }
});
```

text/0000-default-serializers-and-adapters.md
pete-the-pete: > could we link to this RFC / decision?

I'm not sure what you're referring to? I linked to [Project Trim](https://github.com/emberjs/data/issues/6166) in the summary section. Is it my phrasing of 'this RFC lands'? In that case I'm talking about these deprecations.



> is dev mode the development version of `ember build`? i.e. ember's default.

Yes. I am just calling it out to make sure the deprecations don't show up in a production build.

efx: Ah, got it thanks. I had thought that was part of another RFC or something.

text/0000-default-serializers-and-adapters.md
pete-the-pete: The adapters/serializers would be have them same when in use. This RFC is only about deprecating the default injection of serializers. As it stands now, apps don't have to specify a adapter/serializer for their application because ember-data makes one available by default. This deprecates that default behavior, so in the future apps will have to explicitly specify an adapter/serializer per model and/or for their application.

runspired: Also note that in practice the number of applications that choose not to provide an application adapter/serializer seems to be fairly low

runspired:
Thank you for taking on this RFC!

text/0000-default-serializers-and-adapters.md
```diff
@@ -0,0 +1,53 @@
...
+- Tracking:
+
+# Deprecate default Adapter and Serializer fallbacks
+
+## Summary
+As part of Project Trim, https://github.com/emberjs/data/issues/6166, this deprecates the fallback and default adapter and serializer across ember-data including:
+1. deprecate -default serializer fallback in Store.serializerFor
+2. deprecate Adapter.serializer and Adapter.defaultSerializer fallbacks (which currently default to -json-api.
+3. deprecate Store.defaultAdapter (-json-api) and the -json-api adapter fallback behavior
+4. deprecate Model.toJSON and support the scenario for quick debug-mode inspection via debugData() method which in DEBUG would print a similar object to the current toJSON method.
```
hjdivad: I'm very much of the view we should restrict this RFC to only deprecating API and treat adding a new `debugData` function on `DS.Model` in a separate RFC.

But definitely 👍 on deprecating the fallbacks to default serializer behaviour

text/0000-default-serializers-and-adapters.md
```diff
@@ -0,0 +1,53 @@
...
+
+##### deprecate Model.toJSON
+A deprecation warning will be shown when toJSON is called since it uses a serializer to create a JSON representation of [the model](https://guides.emberjs.com/release/models/customizing-serializers/#toc_customizing-serializers). Users may call record.serialize() or implement their own toJSON instead.
+
+## How we teach this
+
+Today we have extensive documentation about creating custom serializers, but we will need to update the guides to specify the desired serializer in [app/serializers/application.js](https://guides.emberjs.com/release/models/customizing-serializers/#toc_customizing-serializers)
+
+The deprecation guide app will be updated with examples showing how to
+migrate away from relying on the defaults.
```
snewcomer: Will this include updates to any blueprints?  For example adding a default adapter with `ember new`?

pete-the-pete: That sounds like a great idea, but I'm not sure when it should be mentioned...the RFC is about deprecation, but I think updating the blueprints for the deprecation guide is a good idea. @runspired what are your thoughts?

runspired: I think that would be a different RFC if at all

runspired:
After discussion we have agreed to FCP but also wish to update the individual deprecation details to include guidance on each.

PR #528, opened 2019-08-15
Deprecate Events

ggayowsky:
[Rendered](https://github.com/ggayowsky/rfcs/blob/deprecate-evented-mixin/text/0000-deprecate-evented-mixin.md)

### DIFF
--- a/text/0000-deprecate-evented-mixin.md
+++ b/text/0000-deprecate-evented-mixin.md
@@ -1,6 +1,6 @@
 - Start Date: 2019-08-15
 - Relevant Team(s): Ember.js, Learning
-- RFC PR: (after opening the RFC PR, update this with a link to it and update the file name)
+- RFC PR: https://github.com/emberjs/rfcs/pull/528
 - Tracking: (leave this empty)
 
 # Deprecate Evented Mixin
--- a/text/0000-deprecate-evented-mixin.md
+++ b/text/0000-deprecate-evented-mixin.md
@@ -3,59 +3,151 @@
 - RFC PR: https://github.com/emberjs/rfcs/pull/528
 - Tracking: (leave this empty)
 
-# Deprecate Evented Mixin
+# Deprecate Eventing from Ember
 
 ## Summary
 
-Deprecation support for the `Evented` Mixin.
+Deprecation support for Events within Ember.
+
+This includes deprcating the `Evented` Mixin, and methods exposed in the `@ember/object/events` and `@ember/object/evented` packages
 
 ## Motivation
 
-Mixins in Ember can only be used when inheriting from an EmberObject, and with the introduction of native javascript classes the syntax to use them looks a little odd.
+Eventing is something that is no longer core to Ember, as eventing is no longer used internally after changes from all the work on Octane.
 
-```
-class MyClass extends EmberObject.extend(Evented) { }
-```
-
-Furthermore, the Evented Mixin is just a wrapper around a set of functions from the `@ember/object/events` package.
-
-Deprecating the Evented Mixin, should make it easier for applications reduce the dependency on EmberObject and make it easier to switch to native class syntax.
+One pillar of the Ember 2019-2020 roadmap is to "Continue Simplifying Ember". This deprecation aligns with that pillar as deprecating eventing will reduce the surface area of Ember and ensure that Ember is not maintaining a feature that is available and stable in several other libraries.
 
 ## Transition Path
 
-As mentioned in the motivation, the `@ember/objects/events` package mostly exposes the functionality of the Evented mixin (aside from the method `Evented.has`)
+If a developer still wishes to use the eventing pattern, the recommendation is to use a third-party library.
 
-There is a straightforward mapping from an Evented mixin method to a function provided by `@ember/object/events` as listed below:
+There are several eventing libraries available which are stable and well supported. Some examples include, but are not limited to:
+* events (https://www.npmjs.com/package/events)
+* eventemitter3 (https://www.npmjs.com/package/eventemitter3)
+* event-kit (https://www.npmjs.com/package/event-kit)
 
-* On:  
-`eventedObj.on(eventName, target, method)` -> `addListener(eventedObj, eventName, target, method)`
+Any objects that currently make use eventing pattern will need to create an instance of an emitter and use that emitter instance to subscribe to and emit events.
 
-* One:  
-`eventedObj.one(eventName, target, method)` -> `addListener(eventedObj, eventName, target, method, true)`
+For example this code which makes use of the Evented mixin:
 
-* Off:  
-`eventedObj.off(eventName, target, method)` -> `removeListener(eventedObj, eventName, target, method)`
+```
+import Evented, { on } from `@ember/object/evented';
 
-* Trigger:  
-`eventedObj.trigger(eventName, param1, param2, ..., paramN)` -> `sendEvent(eventedObj, eventName, [param1, param2, ..., paramN])`
+const EventedObject = EmberObject.extend(Evented, {
+    eventHandler: on('outputMessage', function() {
+        console.log('Outputting message from handler inside object');
+    })
+});
+const eventedObj = EventedObject.create({})
 
-* Has:  
-`eventedObj.has(eventName)` -> `hasListeners(eventedObj, eventName)`  
-Currently `hasListeners` is not a function that can be imported from `@ember/object/events`. `hasListeners` would need to be exported for this deprecation to be possible.
+const handler = (params1, param2) => {  console.log(`Let's ${param1} ${param2}!`); };
+eventedObj.on('outputMessage', handler);
+eventedObj.trigger('outputMessage', 'Deprecate', 'Events');
+/* console log
+ * Outputting message from handler inside object
+ * Let's Deprecate Events!
+ */
+eventedObj.off('outputMessage', handler)
+```
 
-To assist with this transition, it should be possible to provide a codemod that updates the application to make use of the events functions.
+and this functionally equivalent code which makes use of the methods from the `@ember/object/events` package:
+
+```
+import { addListener, removeListener, sendEvent } from `@ember/object/events;
+
+const EventedObject = EmberObject.extend({
+    init() {
+        addListener(this, 'outputMessage', this, 'eventHandler');
+    },
+
+    eventHandler() {
+        console.log('Outputting message from handler inside object');
+    }
+});
+const eventedObj = EventedObject.create({});
+
+const handler = () => {  console.log('Outputting message from handler outside object'); };
+addListener(eventedObj, 'outputMessage', handler);
+sendEvent(eventedObj, 'outputMessage', ['Deprecate', 'Events']);
+/* console log
+ * Outputting message from handler inside object
+ * Let's Deprecate Events!
+ */
+removeListener(eventedObj, 'outputMessage', handler);
+```
+
+can be rewritten using the third-party library `eventemitter3` as such
+
+```
+import EventEmitter from 'eventemitter3';
+
+const EventedObject = EmberObject.extend({
+    init() {
+        this.emitter = new EventEmitter();
+        this.emitter.on('outputMessage', this.eventHandler, this);
+    },
+
+    eventHandler() {
+        console.log('Outputting message from handler inside object');
+    }
+});
+
+const eventedObj = EventedObject.create({});
+const handler = (params1, param2) => {  console.log(`Let's ${param1} ${param2}!`); };
+eventedObj.emitter.on('outputMessage', handler);
+eventedObj.emitter.emit('outputMessage', 'Deprecate', 'Events');
+/* console log
+ * Outputting message from handler inside object
+ * Let's Deprecate Events!
+ */
+eventedObj.emitter.off('outputMessage', handler);
+```
+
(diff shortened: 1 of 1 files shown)


CvX:
Great RFC! 🙂

Could you update the "Rendered" link? (it points to master instead of `deprecate-evented-mixin`)

pzuraq:
I'm very much in favor of this RFC 😄 I actually think it could go further. I'm planning on submitting quite a few deprecation RFCs post-Octane, and two that I have planned are:

1. Mixins, in general
2. Event listeners (from `@ember/object/events`), in general

This would be a good first step toward either of these, but it may also be worth expanding the scope here since all of these features are going to eventually be deprecated anyways. If you're interested, we can dive in and discuss what that would look like!

ggayowsky:
I am amenable to expanding the scope of the PR and would enjoy discussing further in discord.

I agree with deprecating Mixins entirely. As far and I can tell, the Evented Mixin is the only Mixin that is publicly exposed through the ember packages. I figured that deprecating only Evented is simple first step, that would help reduce churn (compared to deprecating all Mixins) since the Evented Mixin seems to have a very straightforward transition path for deprecation (with potential for a codemod).

If the plan is to deprecate the `@ember/object/events` package (and not expose as an add-on), then I would want to update this RFC's transition path to include suggestions that will remove dependency on said package.

runspired:
I'm also in favor of this, and since EmberData has already shipped the deprecation of our own usage of the Evented mixin this shouldn't be terribly difficult: #329

ggayowsky:
After discussion with @pzuraq in the discord, this RFC will have its scope expanded to consider deprecation all eventing from ember (`Evented`, `@ember/object/evented` and `@ember/object/evented`)

gossi:
I want to throw in an alternative to this is to provide a decorator that takes over the job of the mixin. Similar to what orbitjs does: https://github.com/orbitjs/orbit/blob/master/packages/%40orbit/core/src/evented.ts

That can keep the API consistent only the `extends EmberObject.extend(Evented)` needs to change to a decorator here.

And I'm summoning @dgeb into this.

pzuraq:
@gossi: @ggayowsky and I discussed this possibility in Discord, and it definitely is something that could be done. However, eventing is just not something that is core to Ember anymore. The framework itself does not use eventing internally anymore, and it really makes sense to remove it so we aren't maintaining an orthogonal feature in the same codebase.

We _could_ create our own `@evented` decorator and maintain it as an official Ember library, but there are already several (non-decorator) eventing libraries that are well maintained on NPM:

* [EventEmitter3](https://www.npmjs.com/package/eventemitter3)
* [events](https://www.npmjs.com/package/events)
* [event-kit](https://www.npmjs.com/package/event-kit)

Since this is functionality that isn't used very often, I think the best path forward is to point users who need it to these libraries instead so we lessen the maintenance burden on our ecosystem.

mehulkar:
Should this RFC include the work required to move Ember itself off this mixin? E.g. the `Router` uses mixes in `Evented`, as I'm sure do other things.

locks:
@mehulkar that's what the RFC and the thread is about, deprecating the mixin and all the events.

lukemelia:
As a reasonably heavy user of the Evented mixin, I’m not opposed to this change but would suggest that we should let the absence of an Ember implementation of eventing stop us from providing a codemod that updates code to use one of the listed libraries. At best, it substantially eases migration risk and at worst it provides an example for developers who want to migrate to a different lib.

st-h:
I personally think that removing stuff that no longer is used internally from ember core is a reasonable decision. However, I also think that just removing features will have at least implications on developer experience, as long as we do not provide a clear way forward.

I have started to make use of the Evented Mixin after I noticed that popular addons like ember-simple-auth were using it and found it quite handy. Despite this feature not being promoted heavily on the emberjs website, it seems like this is a feature that is not just very rarely used.

Especially for addons a situation where there is no "official" replacement will likely have implications, as this will encourage addon developers to make use of an arbitrary 3rd party dependency or write their own code. This in terms will bring up a situation for all app devs where they could easily end up with multiple js dependencies that just do the same thing which will only necessarily bloat their apps.

I am especially concerned that ember will no longer provide any means to do things like for instance: Make use of a service to wrap a third party event (very simple example the `online` [event](https://developer.mozilla.org/en-US/docs/Web/API/NavigatorOnLine/Online_and_offline_events)). And then allow other parts of the app to subscribe to that event, so they could display the correct state. 

I think we should continue to provide an official add-on which brings back removed features (as this usually has been the case in the past) to mitigate all those issues, if we want to remove events from ember core.

text/0000-deprecate-evented-mixin.md
```diff
@@ -0,0 +1,194 @@
...
+
+## Drawbacks
+
+This RFC introduces churn in existing apps that make use of the Evented Mixin and Events.
+
+Due to transitioning away from an Ember package and using a third-party library of the developers choice, this seems like it will be difficult / impossible to codemod and be a very manual process. 
+
+## Alternatives
+
+* Do not deprecate Ember Events
```
locks:
```suggestion
```

text/0000-deprecate-evented-mixin.md
```diff
@@ -0,0 +1,194 @@
...
+* Deprecate only the `@ember/object/evented` package and Evented Mixin, and keep the `@ember/object/events` package
+
+## Unresolved questions
+
+Can this process be somwhat automated by providing a codemod?
```
locks:
```suggestion
Can this process be somewhat automated by providing a codemod?
```

text/0000-deprecate-evented-mixin.md
```diff
@@ -0,0 +1,194 @@
...
+- RFC PR: https://github.com/emberjs/rfcs/pull/528
+- Tracking: (leave this empty)
+
+# Deprecate Eventing from Ember
+
+## Summary
+
+Deprecation support for Events within Ember.
+
+This includes deprcating the `Evented` Mixin, and methods exposed in the `@ember/object/events` and `@ember/object/evented` packages
```
locks:
```suggestion
This includes deprecating the `Evented` Mixin, and methods exposed in the `@ember/object/events` and `@ember/object/evented` packages.
```

text/0000-deprecate-evented-mixin.md
```diff
@@ -0,0 +1,194 @@
...
+
+## Motivation
+
+Eventing is something that is no longer core to Ember, as eventing is no longer used internally after changes from all the work on Octane.
```
locks:
```suggestion
Eventing is something that is no longer core to Ember, and is no longer used internally after changes from all the work on Octane.
```

st-h:
@pzuraq Thanks a lot for putting this together. I think this addresses all my concerns and in my opinion this would make up a great `How we teach this` section for the RFC (and of course content for the guides). I am not very familiar with all the details about the octane edition yet - mostly because I was running into issues each time I tried to make use of glimmer components, but didn't have the time to dig any deeper.
As long as we are able to solve those use cases by using core ember features, I don't think we need  to consider maintaining a dedicated add-on. I was mainly concerned that we would remove a feature just for the sake of no longer needing it internally, but that does not seem to be the case at all as you pointed out.
And yes, writing code to implement a publisher subscriber pattern can usually be done without much hassle and a few lines of code. I was adding something similar to an app quite a while ago, however the reason was that I needed to bypass the runloop as I needed more frequent updates to keep drawing a canvas smooth.

lindyhopchris:
So having given this a lot of thought, I'm totally unsure about this RFC. I use events extensively, because it's such a common design pattern for application development (both backend and frontend) - i.e. for message based systems.

Although I like the example above that uses auto-tracking, and can see how I could use that to replace some of my use of events (and would definitely be better), I'm struggling to see how I can replace *every* use of events in my applications. The example assumes that you are using events to communicate state changes, but that's not always the case.

As an example, we receive events from the backend server over websockets, that we then re-fire in our Ember app... i.e. so that logically an event on the backend is effectively replicated in the frontend - which makes development simpler because backend and frontend developers work *with the same events* (even if they are not actually the same events!)

In this scenario our service that receives, processes and fires the events in the Ember app doesn't hold state... i.e. it has no logical conception of what is listening to it, or whether the event is actually going to be used - its job is just to ensure the event that occurred on the backend has an equivalent in the Ember app.

Now I totally accept that there are other event packages that I could choose to use if Ember decides to remove its event system as per this RFC. So in that sense I'm not worried about this RFC.

What does worry me though is event based messaging systems are a common application design pattern. So I see it as totally feasible that addons might not be consistent in choosing the same event based package, meaning we end up with multiple installed in our apps which is not a great position to be in.

Considering Ember is an application framework, I personally think a light-weight event capability should legitimately be provided by the framework. However, I can see the case for it being an optional install...

I.e. I'm proposing that while the framework might want to remove events from core, there should be an official Ember event addon. That could either keep the existing code that is extracted from core, or wrap one of the available npm event packages. Not bothered which, as long as it kept some unity in what you use if your app or addon wants to use an event-based design pattern.

pzuraq:
@lindyhopchris can you provide more concrete examples of the types of events that your backend sends, and that you re-fire on the frontend? This does sound like it may be a a good example of a use case for an eventing pattern, but it would help to make the discussion a bit more concrete.

I'm also curious as to what types of use cases you think would cause addons to be affected. This seems like an application concern, so I don't think most addons would be affected by this particular issue, and I find it difficult to see how this would become an ecosystem problem.

Regarding providing an eventing library, would the core team recommending a specific existing library (especially if one is better than the others), as opposed to creating/maintaining a wrapper, be a good solution to your concerns? Keep in mind, this affect isn't just about the Ember ecosystem - other JS libraries may be using eventing as well, and may actually be more likely too in the future, since Ember is moving away from eventing as a common pattern, but the rest of the ecosystem isn't necessarily.

lindyhopchris:
@pzuraq thanks for your response. here's the answers to your questions...

> can you provide more concrete examples of the types of events that your backend sends, and that you re-fire on the frontend?

We have multiple pieces of hardware connected to the system that our Ember app is the UI for. The user controls those pieces of hardware, but things can happen independently on the hardware. As one example, a device can be rebooted. So we fire device rebooting and device rebooted events on the backend, that then get broadcast to the front-end, where the events are displayed/notified to the user, plus prevent the user from doing things while those events are happening.

Potentially we could refactor these as each device does have an Ember Data model, so we could wrap up something that keeps track of the model, plus state properties that are changed when we receive the events, etc. But I'm not sure how this is an improvements, as the event-based system works well and accurately reflects what's going on.

> I'm also curious as to what types of use cases you think would cause addons to be affected. 

One of the addons we use is Ember websockets... that does everything via events on the object that represents the socket. Not sure what the alternative pattern would be for that, and whether an alternative pattern would work?
https://github.com/thoov/ember-websockets

> Regarding providing an eventing library, would the core team recommending a specific existing library (especially if one is better than the others), as opposed to creating/maintaining a wrapper, be a good solution to your concerns? 

Yes, I suppose that would cover it. I suppose I'm raising my concerns not in a *please don't do this!* way, but more a *if you're going to do this, I feel there still needs to be an Ember-way of doing events*. Hope that makes sense?

NullVoxPopuli:
Fwiw, I have a websockets powered app, https://emberclear.io, and I don't use events. 

I'm explicitly calling functions defined on other services.

lindyhopchris:
> I'm explicitly calling functions defined on other services.

Hmm, yeah I can see that works but I wouldn't normally architect an event-message system like that, because it gives one service too much explicit knowledge of the other services, rather than decoupling them. To be clear, not saying it's wrong as a lot of these approaches are down to developer preference - just saying it wouldn't be my choice. (I.e. please don't interpret this comment as a criticism!)

Anyway, I think I've probably said everything I wanted to to input into this discussion. Interested to see what the decision on this RFC will be!

pzuraq:
> One of the addons we use is Ember websockets... that does everything via events on the object that represents the socket. 

Ah, I think maybe I should clarify here, it's clear that certain data patterns would require eventing, for instance if you use websockets in this way. If you use Apollo, it has it's own Observable based eventing, so that would be another example where a data solution would bring eventing back into an app.

What's not clear is whether or not this would be a common pattern throughout the ecosystem. Would addons that provide components be affected by this? Would addons that provide authentication/authorization, such as `ember-simple-auth`, be affected? Are there addons that provide additional functionality on top of `ember-websockets` that would be affected?

Genuine questions here. Basically, if it's unlikely that an application will include multiple eventing libraries in the first place, because there isn't much overlap in the types of things that require eventing, then I think there is less of a concern than if they do, if that makes sense.

> if you're going to do this, I feel there still needs to be an Ember-way of doing events. Hope that makes sense?

Definitely makes sense, I think ideally we would provide some kind of guidance here, for sure 😄 thanks for clarifying btw, really appreciate it, this helps a lot!

NullVoxPopuli:
Would rx.js with a test waiters integration be a better path forward? I wonder if that'd be code moddable somehow...

NullVoxPopuli:
Reason why I bring up rx.js, is that it's a huuuuuge community-wide used library, and super lightweight and solid for event/observable patterns

wagenet:
I'm pretty sure this is something we need to do. I'll bring this up with the core team and see if we can't get this finalized.

wagenet:
I believe this falls under the umbrella of #832. I'm going to close this issue to push any discussion there. Despite being closed, we will certainly use this issue for reference and to inform the future RFC. Thank you all for the discussion!

PR #549, opened 2019-10-22
Ember dev for other platforms

bakerac4:
[rendered](https://github.com/bakerac4/rfcs/blob/master/text/0000-platforms.md)

### DIFF
--- a/text/0000-platforms.md
+++ b/text/0000-platforms.md
@@ -1,6 +1,6 @@
 - Start Date: 10/22/2019
 - Relevant Team(s): Steering, Learning
-- RFC PR: (after opening the RFC PR, update this with a link to it and update the file name)
+- RFC PR: https://github.com/emberjs/rfcs/pull/549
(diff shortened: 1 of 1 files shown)


bakerac4:
FWIW - Yes Im trying to create Glimmer Native but Im not even sure Im going to be able to get past the list view blockers I have. I'd like to see this happen whether or not I succeed. Its IMO just something we are dropping the ball on compared to other frameworks and is something that (again IMO) need to compete

wagenet:
@bakerac4 is this something you still want to move forwards?

wagenet:
After actually looking over this, I'm not at all sure that this RFC is going to work for us. We can't create a new core team out of thin air. If there are actually people already doing the work and who need more official support then that's great, but until such a team exists I'm going to go ahead and close this RFC. If you think I'm mistaken, please let me know. I'm happy to discuss this and move it forwards if it makes sense.

PR #554, opened 2019-11-08
Deprecate getWithDefault

chrisrng:
[Rendered](https://github.com/emberjs/rfcs/blob/e5e9d8249a26fcfe6c0199f7fe0e092876ff1e1f/text/0554-deprecate-getwithdefault.md)


NullVoxPopuli:
I'm a fan. This is codemodable. :+1:

text/0554-deprecate-getwithdefault.md
```diff
@@ -0,0 +1,71 @@
...
+## Transition Path
+
+Ember will start logging deprecation messages for `getWithDefault` usage. 
+
+We can codemod our current usage of `getWithDefault` with the equivalent behaviour using plain JavaScript. The migration guide will cover this example:
+
+Before:
+
+```js
+let result = getWithDefault(obj, ‘some.key', defaultValue);
```
xg-wang:
```suggestion
import { getWithDefault } from '@ember/object';

let result = getWithDefault(obj, ‘some.key', defaultValue);
```

text/0554-deprecate-getwithdefault.md
```diff
@@ -0,0 +1,71 @@
...
+```
+
+After:
+
+```js
+let result = get(obj, 'some.key');
```
xg-wang:
```suggestion
import { get } from '@ember/object';

let result = get(obj, 'some.key');
```

text/0554-deprecate-getwithdefault.md
```diff
@@ -0,0 +1,71 @@
...
+if (result === undefined) {
+  result = defaultValue;
+}
+```
```
snewcomer:
```js
let result = get(obj, 'some.key') || defaultValue;
```

Is this an alternative option?  Also, should the codemod require `get` or can we use plain property accessors?

chrisrng: I think it makes sense to use `get` to minimize behaviour differences

buschtoens: Codemodding paths and interpolated paths will be significantly harder as well:

```js
getWithDefault(obj, 'foo.bar');
getWithDefault(obj, `foo.${key}`);
getWithDefault(obj, path);
```

I'd also suggest to codemod towards `get` and then potentially run another codemod from `get` towards native accessors in the future.

rwjblue:
Thanks for working on this @chrisrng! Seems like a good idea to me...

text/0554-deprecate-getwithdefault.md
simonihmig: FWIW, there is already a codemod for `get()` to native: https://github.com/ember-codemods/es5-getter-ember-codemod. (at least for simple properties, no paths yet)

And someone even suggested covering `getWithDefault()` already: https://github.com/ember-codemods/es5-getter-ember-codemod/issues/1 😝

EWhite613:
I don't think to have a code mod to do is really useful.
```
let result = get(obj, 'some.key');
if (result === undefined) {
  result = defaultValue;
}
```
 I would suspect most people would modify the code mod to use their own util or use something like lodash (if they want that default behavior). Cause having one line turn into 4 while also increasing the complexity is not something I would think people would want in their code

chrisrng:
FYI @steventsao added this to eslint-plugin-ember https://github.com/ember-cli/eslint-plugin-ember/pull/594

steventsao:
> I don't think to have a code mod to do the following is really useful.
> 
> ```
> let result = get(obj, 'some.key');
> if (result === undefined) {
>   result = defaultValue;
> }
> ```
> 
> I would suspect most people would modify the code mod to use their own util or use something like lodash (if they want that default behavior). Cause having one line turn into 4 while also increasing the complexity is not something I would think people would want in their code

Yeah turning one assignment into a block would not be ideal. But this can be a one-liner, right? 

```js
// I imagine the value of `get()` is cached 
const result = get(obj, 'some.key') === undefined ? get(obj, 'some.key') || defaultValue;
```

[Optional chaining](https://github.com/tc39/proposal-optional-chaining) and [nullish coalescing](https://github.com/tc39/proposal-nullish-coalescing):

```js
const result = obj.some?.key ?? defaultValue;
```

chriskrycho:
@steventsao @EWhite613 Codemodding it in the safest way possible is the important bit here. You can definitely do it with a ternary, and that's fine with me as well. I suggested this particular output to @chrisrng because it seemed the clearest to me; a ternary would also be fine but does impose the slight extra cognitive overhead of having to *think* about the caching question when you're reading it. I'm not married to the `let` + `if` block for sure.

text/0554-deprecate-getwithdefault.md
```diff
@@ -0,0 +1,128 @@
...
+
+let result = get(obj, 'some.key');
+if (result === undefined) {
+  result = defaultValue;
+}
+```
+
+#### Using Nullish Coalescing Operator
+
+We cannot codemod directly into the nullish coalescing operator since the expected behaviour of `getWithDefault` is to only return the default value if it is strictly `undefined`. The nullish coalescing operator accepts either `null` or `undefined` as _falsey_ values to show the default value.
```
rwjblue:
```suggestion
We cannot codemod directly into the nullish coalescing operator since the expected behaviour of `getWithDefault` is to only return the default value if it is strictly `undefined`. The nullish coalescing operator accepts either `null` or `undefined` to show the default value.
```

text/0554-deprecate-getwithdefault.md
```diff
@@ -0,0 +1,128 @@
...
+
+Tooling Support:
+
+- [Babel](https://babeljs.io/) already supports the [nullish coalescing operator](https://babeljs.io/docs/en/next/babel-plugin-proposal-nullish-coalescing-operator.html) so we can use that for future use cases where we need to check if a property is `null` or `undefined` before applying a default value.
+
+- [TypeScript](https://github.com/microsoft/TypeScript), similarly, as of [version 3.7](https://devblogs.microsoft.com/typescript/announcing-typescript-3-7/#nullish-coalescing) also supports the operator so we will not be breaking that flow either.
+
+#### Using Object Destructuring With Defaults
+
+If we are only looking to check If both `null` or `undefined` should return the default value in a particular use case, using [object destructuring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment) with defaults can also be an option.
```
rwjblue:
```suggestion
If we would like to return the default value if the existing value is falsey, using [object destructuring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment) with defaults can also be an option.
```

text/0554-deprecate-getwithdefault.md
```diff
@@ -0,0 +1,128 @@
...
+
+Object destructuring with defaults **will return** the default value when the provided value is `undefined` or `null`:
```
rwjblue:
```suggestion
Object destructuring with defaults **will return** the default value when the provided value is _falsey_:
```

text/0554-deprecate-getwithdefault.md
```diff
@@ -0,0 +1,128 @@
...
+let obj = {
+  nullValue: null,
+};
+
+// Returns defaultValue 1, undefinedKey = 1
+let { undefinedKey = defaultValue } = obj;
+
+// Returns defaultValue 1, nullValue = 1
+let { nullValue = defaultValue } = obj;
+```
```
rwjblue: Can you also add an example for `falseValue: false`?

text/0554-deprecate-getwithdefault.md
```diff
@@ -0,0 +1,140 @@
...
+
+## Summary
+
+Deprecate support for `getWithDefault` in Ember's Object module (@ember/object) – both the [function](https://api.emberjs.com/ember/release/functions/@ember%2Fobject/getWithDefault) and the [class method](https://api.emberjs.com/ember/release/classes/EmberObject/methods/getWithDefault?anchor=getWithDefault) – because its expected behaviour is confusing to Ember developers.
+
+## Motivation
+
+The problem with `getWithDefault` is that its behaviour is confusing to Ember developers. The API will only return the default value when the value of the property retrieved is `undefined`. This behaviour is often overlooked when using the function where a developer might expect that `null` or other _falsy_ values will also return the default value.
+
+Given the JavaScript language will soon (currently in Stage 3) give us the appropriate tool for this use case using the [Nullish Coalescing Operator `??`](https://github.com/tc39/proposal-nullish-coalescing), we can deprecate usage of `getWithDefault` and use that instead.
```
mixonic: Some text should be added here: Ember can only suggest nullish coalescing syntax once Ember CLI supports that syntax out of the box. It should be supported in a stable release of Babel, but additionally we need to be sure that `ember new` includes that version of Babel.

And the deprecation instructions should explain how to get the proper version installed.

rwjblue:
We discussed this in todays core team meeting and have decided to move this into final comment period.

steventsao:
fyi adding codemod https://github.com/steventsao/ember-no-get-with-default-codemod

PR #558, opened 2019-12-05
Specifying and detecting an applications edition.

rwjblue:
[Rendered](https://github.com/emberjs/rfcs/blob/edition-detection/text/0558-edition-detection.md)

## Motivation

As Ember approaches it's first edition (see [emberjs/rfcs#364](https://github.com/emberjs/rfcs/pull/364)) various addons need to modify their behavior based on the edition that is being used. An initial implementation (done without RFC) used the `setEdition` method from `@ember/edition-utils` inside the application's (or addon's) `.ember-cli.js` file to specify which edition was to be used. That implementation worked well enough throughout the initial preview period, but a number of major issues were (rightfully!) surfaced by the community:

1. it seems unnecessary/redundant
2. it's not clear what this flag actually does (likely due to having no RFC!)
3. it's not statically analyzable (and therefore cannot be used by things like codemods)

This RFC will review these issues in light of the updated implementation, showing how each of the concerns have been met.

## Timeline

This RFC will have a significantly compressed timeline (hopefully in FCP tomorrow 🙀), as we attempt to move forward with "shipping octane" (🤞in a version **very** soon). I am sorry that it has taken me so long to put together, but hopefully we can come to a quick consensus and fix the mistakes of the past.

### DIFF
--- a/text/0558-edition-detection.md
+++ b/text/0558-edition-detection.md
@@ -97,12 +97,12 @@ their `.ember-cli.js`) a period of time in order to migrate.
 ## How we teach this
 
 The official guides at `https://guides.emberjs.com/release/configuring-ember/`
-will be updated to include documentation of the new `pacakge.json`
+will be updated to include documentation of the new `package.json`
 configuration and clearly explain what the edition flag is used for (warnings
 and blueprints).
 
-This will will not be a difficult concept to teach to folks (most users won't
-care, and will get upgrade as part of a future `ember-cli` blueprint update).
+This will not be a difficult concept to teach to folks (most users won't
+care, and will get upgraded as part of a future `ember-cli` blueprint update).
 
 ## Drawbacks
 
--- a/text/0558-edition-detection.md
+++ b/text/0558-edition-detection.md
@@ -64,6 +64,23 @@ property in the `package.json` from their `config/ember-try.js` scenarios.
 The edition flag should only be used by addons to determine what blueprint
 output to generate and to provide helpful warnings (or errors) at build time.
 
+Note that the above definition does **not** allow for an addon to detect the
+edition configuration and change its fundamental implementation. This is quite
+intentional!
+
+Instead, addons should rely on feature detection techniques like the following
+to alter implementations:
+
+* Check to see if given methods/properties exist (this is essentially like how
+  browser polyfills work)
+* Use a tool like
+[ember-compatibility-helpers](https://github.com/pzuraq/ember-compatibility-helpers)'s
+`gte` method (e.g. `if (gte('ember-source', '3.14.0')) { }`)
+* Use
+  [@ember/optional-features](https://github.com/emberjs/ember-optional-features)'s
+  [`isFeatureEnabled`
+  method](https://github.com/emberjs/ember-optional-features#at-build-time-from-an-addon)
+
 ### Detecting the edition
 
 The existing `@ember/edition-utils` package will still be used by addons to
--- a/text/0558-edition-detection.md
+++ b/text/0558-edition-detection.md
@@ -154,6 +154,10 @@ Instead of adding the `emberEdition` value to the `package.json` we could add
 it to the existing `.ember-cli` file. However, doing this would **not** satisfy
 the static analysis constraint mentioned in the motivation section (because
 `.ember-cli.js` is transparently supported by `ember-cli`'s build system).
+In addition, any values that are included in `.ember-cli` are automatically
+passed in to every command invocation which would be both unintended (we don't
+want commands to access the edition in this way) and possibly breaking (if the
+command already accepted an option with whatever value we chose).
 
 ## Unresolved questions
 
--- a/text/0558-edition-detection.md
+++ b/text/0558-edition-detection.md
@@ -52,6 +52,13 @@ The value will be expected to be one of the valid edition names (currently
 `classic` and `octane`). Using the `package.json` for this allows us to ensure
 that the value is statically analyzable and easy to discover.
 
+For applications specifying `ember: { edition: 'octane' }` in `package.json` is
+generally all that is needed. In an addon this new property would be specifying
+the edition that the addons own dummy app uses. However, many addons may want
+to test against _multiple_ editions that they support. In order to support
+this, ember-try will be updated to allow specifying (and merging) the `ember`
+property in the `package.json` from their `config/ember-try.js` scenarios.
+
 ### Valid use of the edition value
 
 The edition flag should only be used by addons to determine what blueprint
--- a/text/0558-edition-detection.md
+++ b/text/0558-edition-detection.md
@@ -1,6 +1,6 @@
 - Start Date: (2019-11-20)
 - Relevant Team(s): (Ember.js, Ember CLI, Ember Data)
-- RFC PR: (after opening the RFC PR, update this with a link to it and update the file name)
+- RFC PR: https://github.com/emberjs/rfcs/pull/558
 - Tracking: (leave this empty)
 
 # Edition detection
--- a/text/0000-edition-detection.md
+++ b/text/0000-edition-detection.md
@@ -1,0 +1,136 @@
+- Start Date: (2019-11-20)
+- Relevant Team(s): (Ember.js, Ember CLI, Ember Data)
+- RFC PR: (after opening the RFC PR, update this with a link to it and update the file name)
+- Tracking: (leave this empty)
+
+# Edition detection
+
+## Summary
+
+Introduces a mechanism that an application can use to specify which Edition of
+Ember it is intending to target. This RFC will define:
+
+* How to specify the edition that the application is using
+* How other packages (addons, codemods, etc) can detect the applications intended edition
+* What the edition should be used for
+
+## Motivation
+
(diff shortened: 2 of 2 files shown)


NullVoxPopuli:
:100: I think the ember root key in package.json gives us room to play around in the future, too.

text/0558-edition-detection.md
```diff
@@ -0,0 +1,136 @@
...
+
+The API documentation for `@ember/edition-utils` would be:
+
+```ts
+module '@ember/edition-utils' {
+  /**
+    Determine if the application that is running is running under a given Ember
+    Edition.  When the edition in use is _newer_ than the requested edition it
+    will return `true`, and if the edition in use by the application is _older_
+    than the requested edition it will return `false`.
```
buschtoens: What, when we get the first edition, that is a major bump (4.x)? Would `octane` still return `true`, since the next edition is strictly newer?

text/0558-edition-detection.md
```diff
@@ -0,0 +1,136 @@
...
+```
+
+The value will be expected to be one of the valid edition names (currently
+`classic` and `octane`). Using the `package.json` for this allows us to ensure
+that the value is statically analyzable and easy to discover.
+
+### Valid use of the edition value
+
+The edition flag should only be used by addons to determine what blueprint
+output to generate and to provide helpful warnings (or errors) at build time.
```
buschtoens: What about changing the build? For instance loading different class implementations. 

What about feature "detection"? Should we use https://github.com/pzuraq/ember-compatibility-helpers instead?

jrjohnson:
Couple of add-on author questions. Is this also where addons declare their support? The edition for the dummy app? How do addons test multiple editions?

text/0558-edition-detection.md
rwjblue: Yep. The idea is that calling `has('octane')` will be true if the app specifies any edition from octane and later. Without that functionality every addon that wants to do any work here would have to duplicate the logic to know which strings are newer and which are older (and we **definitely** don’t want to use version numbers)

text/0558-edition-detection.md
```diff
@@ -0,0 +1,136 @@
...
+For a period of time the `@ember/edition-utils` package will continue to
+support existing users of `setEdition` when an edition is not detected via the
+new mechanism. This allows users that have been testing out Ember Octane
+(either via the `@ember/octane-app-blueprint` or manually using `setEdition` in
+their `.ember-cli.js`) a period of time in order to migrate.
+
+## How we teach this
+
+The official guides at `https://guides.emberjs.com/release/configuring-ember/`
+will be updated to include documentation of the new `pacakge.json`
```
runspired:
```suggestion
will be updated to include documentation of the new `package.json`
```

text/0558-edition-detection.md
rwjblue: Yes. Feature detection (providing alternate implementations) is generally **not** about what edition the host app is using, but instead about feature sets of a given package (most often Ember, but also applies to ember-data, or any other package). 

Using ember-compatibility-helpers’s `gte` is the right tool for that job.

runspired: edition really is specifying a minimum version and set of features, you'll still want to use feature detection to go beyond that. You can mix: user-has-specified-octane and user-has-feature using a combo of compatibility helpers and debug-macros+this property.

text/0558-edition-detection.md
```diff
@@ -0,0 +1,136 @@
...
+configuration and clearly explain what the edition flag is used for (warnings
+and blueprints).
+
+This will will not be a difficult concept to teach to folks (most users won't
```
chriskrycho:
```suggestion
This will not be a difficult concept to teach to folks (most users won't
```

text/0558-edition-detection.md
```diff
@@ -0,0 +1,136 @@
...
+care, and will get upgrade as part of a future `ember-cli` blueprint update).
```
chriskrycho:
```suggestion
care, and will get upgraded as part of a future `ember-cli` blueprint update).
```

text/0558-edition-detection.md
buschtoens: My only concern was future outdated addons that rely on features introduced in `octane` or implicitly rely on other features, that may be removed with 4.x, would break then. But I think this is a pretty minor hazard, if any at all.

rwjblue:
> Couple of add-on author questions. Is this also where addons declare their support? The edition for the dummy app? How do addons test multiple editions?

@jrjohnson - Great questions! I have added a snippet to the RFC to specifically address this. Here is the relevant verbiage added:

> For applications specifying `ember: { edition: 'octane' }` in `package.json` is
generally all that is needed. In an addon this new property would be specifying
the edition that the addons own dummy app uses. However, many addons may want
to test against _multiple_ editions that they support. In order to support
this, ember-try will be updated to allow specifying (and merging) the `ember`
property in the `package.json` from their `config/ember-try.js` scenarios.

rwjblue:
> What about changing the build? For instance loading different class implementations.

@buschtoens - Great question, I've updated the prose to explain what the best path forward is here. Snippet below:

Note that the above definition does **not** allow for an addon to detect the
edition configuration and change its fundamental implementation. This is quite
intentional!

> Note that the above definition does **not** allow for an addon to detect the
edition configuration and change its fundamental implementation. This is quite
intentional!
>
> Instead, addons should rely on feature detection techniques like the following
> to alter implementations:
> 
> * Check to see if given methods/properties exist (this is essentially like how browser polyfills work)
> * Use a tool like [ember-compatibility-helpers](https://github.com/pzuraq/ember-compatibility-helpers)'s `gte` method (e.g. `if (gte('ember-source', '3.14.0')) { }`)
> * Use [@ember/optional-features](https://github.com/emberjs/ember-optional-features)'s [`isFeatureEnabled` method](https://github.com/emberjs/ember-optional-features#at-build-time-from-an-addon)
>

text/0558-edition-detection.md
rwjblue: Updated prose to state the official position explicitly, thanks for surfacing the question!

text/0558-edition-detection.md
rwjblue: > My only concern was future outdated addons that rely on features introduced in octane or implicitly rely on other features, that may be removed with 4.x, would break then.

Ya, but that hazard is fairly unrelated to the edition. APIs removed in a major version bump would always have this pitfall (and is one of the reasons we have so few major bumps!).

kellyselden:
Can you clarify why .ember-cli won't work? Is it because .ember-cli only looks like JSON, but is interpreted, so who knows what could be in there?

Or is it because we also support .ember-cli.js, it would force them to have both files to get the static edition?

rwjblue:
> Can you clarify why .ember-cli won't work?

@kellyselden - Ya, I had this question specifically listed in the "alternatives" section, but I've updated the prose there to hopefully make it even clearer. Please let me know if that helps!

Snippet:

> Instead of adding the `emberEdition` value to the `package.json` we could add
it to the existing `.ember-cli` file. However, doing this would **not** satisfy
the static analysis constraint mentioned in the motivation section (because
`.ember-cli.js` is transparently supported by `ember-cli`'s build system).
In addition, any values that are included in `.ember-cli` are automatically
passed in to every command invocation which would be both unintended (we don't
want commands to access the edition in this way) and possibly breaking (if the
command already accepted an option with whatever value we chose).

kellyselden:
> and possibly breaking (if the command already accepted an option with whatever value we chose).

Isn't that also true of any apps that use an "ember" property in package.json for whatever reason?

The snippet doesn't really explain the relationship between .ember-cli.js and .ember-cli. I know it's not a requirement of the RFC to explain it, but it might be nice for readers to explain why supporting a file called .ember-cli.js means we can't use static JSON in .ember-cli for this new property. Thanks.

rwjblue:
Thank you all for your help in moving this forward! We discussed this in the Ember core team meeting today and are moving this into final comment period.

In addition, due to the shortened cycle for this RFC (apologies again :sad:), we will be working toward landing implementations in ember-cli's current beta cycle (targeting 3.15's release in a few days). Regardless of landing the implementation in the beta branch, any *actual* release will wait for this RFC cycle to be completed (assuming no new information / issues are brought up in the thread here, this means a release could happen as early as next Friday). I realize this is a somewhat unusual (we normally like to have a full cycle for new features), but hope that y'all understand.

rwjblue:
> The snippet doesn't really explain the relationship between .ember-cli.js and .ember-cli. I know it's not a requirement of the RFC to explain it, but it might be nice for readers to explain why supporting a file called .ember-cli.js means we can't use static JSON in .ember-cli for this new property.

@kellyselden - Ya, I'll add a bit more detail there. I agree the current "oh, btw, there is also .ember-cli.js" doesn't _actually_ explain why that is a deal breaker.

rwjblue:
The core team met to discuss this RFC today, we are still in favor of moving forward here, there were no new concerns brought up during the FCP period, the implementation has gone smoothly, we are ready to land this.

Thank you everyone for providing feedback, we absolutely ended up on a much better solution!

PR #561, opened 2019-12-08
Adding Numeric Comparison Operators to Templates

cibernox:
Extracted from RFC #388 

[Rendered](https://github.com/cibernox/rfcs/blob/add-numeric-comparison-operators-to-templates/text/0561-add-numeric-comparison-operators.md)

### DIFF
--- a/text/0000-add-numeric-comparison-operators.md
+++ b/text/0000-add-numeric-comparison-operators.md
@@ -1,0 +1,66 @@
+- Start Date: 2019-12-08
+- Relevant Team(s): (fill this in with the [team(s)](README.md#relevant-teams) to which this RFC applies)
+- RFC PR: https://github.com/emberjs/rfcs/pull/561
+- Tracking: (leave this empty)
+
+# Adding Numeric Comparison Operators to Templates
+
+## Summary
+
+Add new built-in template `{{lt}}` and `{{gt}}` helpers to perform basic numeric comparison operations in templates, similar to those included in `ember-truth-helpers`.
+
+This RFC is a subset of the changes proposed in #388.
+
+## Motivation
+
+It is a very common need in any sufficiently complex Ember app to perform some numeric comparison operations and often the most convenient place to do it is right in the templates.
+Because of that, [ember-truth-helpers](https://github.com/jmurphyau/ember-truth-helpers) is one of the most installed addons out there, either directly by apps or indirectly by
+other addons that those apps consume.
+
+The fact that `ember-truth-helpers` is so popular is a good signal that this it is filling a perceived gap in Ember's functionality.
+
+A second reason is that it might help make Ember more approachable to newcomers that have some experience in other frameworks.
+Most if not all web frameworks have some way of comparing numbers in the templates and it's surprising that Ember requires an third party package to perform
+even the most basic operations.
+
+
+## Detailed design
+
+Add `{{lt}}` and `{{gt}}` helpers.
+
+#### `{{lt}}`
+Binary operation. Throws an error if not called with exactly two arguments.
+Equivalent of <arg1> < <arg2>
+This is identical to the `{{lt}}` helper in `ember-truth-helpers`
+
+#### `{{gt}}`
+Binary operation. Throws an error if not called with exactly two arguments.
+Equivalent of <arg1> > <arg2>
+This is identical to the `{{gt}}` helper in `ember-truth-helpers`, except for the name.
+
+This RFC intentionally leaves the implementation details unspecified, those could be implemented in Glimmer VM or
+in a higher level in Ember itself.
+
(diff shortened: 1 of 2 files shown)


text/0000-add-numeric-comparison-operators.md
```diff
@@ -0,0 +1,66 @@
...
+## Alternatives
+
+One alternative path is don't do anything and let users continue to define their own helpers (or install `ember-truth-helpers`).
+
+## Unresolved questions
+
+- The proposed version of those helpers mimic the behavior of the `<` and `>` operators in Javascript. While this is the
+  least surprising thing to do, one could argue that we could add extra logic to protect users from the usual pitfalls
+  and edge cases of those operators, **like throwing exceptions if any of the values is `NaN` or a value other than a number**.
+- We must decide if it's worth adding `{{lte}}` and `{{gte}}` helpers, equivalent to `<=` and `>=` respectively.
```
Turbo87: I think I would prefer having them supported too

rwjblue:
:wave: Just a quick update here. We discussed this in todays core team meeting, and are generally in favor. We also strongly that `{{lte}}` and `{{gte}}` should be included here. There are a few details that we want to chat through around just what that means, specifically around the equality side of things. We _just_ landed the `{{eq}}` / `{{neq}}` RFC which very intentionally uses `===` style equality, and it seems pretty odd conceptually for `{{lte}}` to be referring to a **different** kind of equality than `{{eq}}` does.

Basically, the question is, should `{{lte}}` (for example) be essentially `{{or (lt first second) (eq first second)}}`? It _seems_ like the answer ought to be "yes", but this means a slightly different thing than `>=` in JavaScript (since `>=` refers to `==` style equality). It seems somewhat unlikely to be a giant issue, but we do want to make sure we have rationalized the answer properly.

We plan to discuss this with the GlimmerVM team and again at next weeks core team meeting, hopefully I'll have another update for folks next week.

text/0561-add-numeric-comparison-operators.md
```diff
@@ -0,0 +1,83 @@
...
+
+## Detailed design
+
+The new comparison operators will be made keywords, so they are easily accessible in current and future templates without needing to be imported.
+
+#### `{{lt}}`
+
+Binary operation. Throws an error if not called with exactly two arguments.
+Equivalent of <arg1> < <arg2>
+This is identical to the `{{lt}}` helper in `ember-truth-helpers`
```
mixonic, line 41:
```suggestion
Equivalent of `<arg1> < <arg2>`.
This is identical to the `{{lt}}` helper in `ember-truth-helpers`.
```

text/0561-add-numeric-comparison-operators.md
```diff
@@ -0,0 +1,83 @@
...
+
+#### `{{lte}}`
+
+Binary operation. Throws an error if not called with exactly two arguments.
+Equivalent of <arg1> <= <arg2>
+This is identical to the `{{lte}}` helper in `ember-truth-helpers`
```
mixonic, line 47:
```suggestion
Equivalent of `<arg1> <= <arg2>`.
This is identical to the `{{lte}}` helper in `ember-truth-helpers`.
```

text/0561-add-numeric-comparison-operators.md
```diff
@@ -0,0 +1,83 @@
...
+
+#### `{{gt}}`
+
+Binary operation. Throws an error if not called with exactly two arguments.
+Equivalent of <arg1> > <arg2>
+This is identical to the `{{gt}}` helper in `ember-truth-helpers`, except for the name.
```
mixonic, line 53:
```suggestion
Equivalent of `<arg1> > <arg2>`
This is identical to the `{{gt}}` helper in `ember-truth-helpers`.
```

text/0561-add-numeric-comparison-operators.md
```diff
@@ -0,0 +1,83 @@
...
+
+#### `{{gte}}`
+
+Binary operation. Throws an error if not called with exactly two arguments.
+Equivalent of <arg1> >= <arg2>
+This is identical to the `{{gte}}` helper in `ember-truth-helpers`, except for the name.
```
mixonic, line 59:
```suggestion
Equivalent of `<arg1> >= <arg2>`.
This is identical to the `{{gte}}` helper in `ember-truth-helpers`.
```

rwjblue:
We discussed this at todays core team meeting, and with @pzuraq's recent updates (adding `gte`/`lte`) we are happy to move this into final comment period. 🎉 

@locks - Is going to update some materials for "How we teach this" (that will have to be done prior to merging), but we don't think it will be an issue for the final comment period itself.

NullVoxPopuli:
has this landed?

PR #562, opened 2019-12-08
Adding Logical Operators to Templates

cibernox:
Extracted from RFC #388 

[Rendered](https://github.com/cibernox/rfcs/blob/add-logical-operators-to-templates/text/0000-add-logical-operators.md)

### DIFF
--- a/text/0000-add-logical-operators.md
+++ b/text/0000-add-logical-operators.md
@@ -1,0 +1,95 @@
+- Start Date: 2019-12-08
+- Relevant Team(s): (fill this in with the [team(s)](README.md#relevant-teams) to which this RFC applies)
+- RFC PR: https://github.com/emberjs/rfcs/pull/562
+- Tracking: (leave this empty)
+
+# Adding Logical Operators to Templates
+
+## Summary
+
+Add new built-in template `{{and}}`, `{{or}}` and `{{not}}` helpers to perform basic logical operations in templates, similar to those included in `ember-truth-helpers`.
+
+This RFC is a subset of the changes proposed in #388.
+
+## Motivation
+
+It is a very common need in any sufficiently complex Ember app to perform some logical comparison operations and often the most convenient place to do it is right in the templates.
+Because of that, [ember-truth-helpers](https://github.com/jmurphyau/ember-truth-helpers) is one of the most installed addons out there, either directly by apps or indirectly by
+other addons that those apps consume.
+
+The fact that `ember-truth-helpers` is so popular is a good signal that this it is filling a perceived gap in Ember's functionality.
+
+A second reason is that it might help make Ember more approachable to newcomers that have some experience in other frameworks.
+Most if not all web frameworks have some way of comparing numbers in the templates and it's surprising that Ember requires an third party package to perform
+even the most basic operations.
+
+A third reason, this time technical, is that by implementing those helpers at a lower-level, we can make them more performant by making them short-circuit.
+Right now helpers implemented using public APIs like those in `ember-truth-helpers` eagerly consume their arguments. In the case of logical operations like `{{and a b}}`,
+once the first argument (`a`) is evalued to _falsey_, the second argument (`b`) is irrelevant. Sometimes arguments can be expensive properties to calculate,
+and by short-circuiting we can avoid computing them at all sometimes.
+
+
+## Detailed design
+
+Add `{{and}}`, `{{or}}` and `{{not}}` helpers.
+
+#### `{{and}}`
+Takes at least two positional arguments, bound or unbound. Raises an error if invoked with less than two arguments.
+It evaluates arguments left to right, returning the first one that is not _truthy_ (**by handlebar's definition of truthiness**)
+or the right-most arguments if **all** evaluate to _truthy_.
+This is *NOT* equivalent to the `{{and}}` helper from `ember-truth-helpers` because unlike this proposed helper, the one in `ember-truth-helpers`
+uses Javascript's definition of truthiness.
+
+#### `{{or}}`
+Takes at least two positional arguments, bound or unbound. Raises an error if invoked with less than two arguments.
+It evaluates arguments left to right, returning the first one that is _truthy_ (**by handlebar's definition of truthiness**) or the
+right-most argument if all evaluate to _falsy_.
+This is *NOT* equivalent to the `{{or}}` helper from `ember-truth-helpers` because unlike this proposed helper, the one in `ember-truth-helpers`
+uses Javascript's definition of truthiness.
+
+#### `{{not}}`
+Unary operator. Raises an error if invoked with more than one positional argument. If the given value evaluates to a _truthy_ value (**by handlebar's definition of truthiness**),
+the `false` is returned. If the given value evaluates to a _falsy_ value (**by handlebar's definition of truthiness**) then it returns `true`.
+This is *NOT* equivalent to the `{{not}}` helper from `ember-truth-helpers` because unlike this proposed helper, the one in `ember-truth-helpers`
+uses Javascript's definition of truthiness.
+
+#### Handlebar's definition of truthiness
+This is the most important detail of this proposal because it's where it deviates from `ember-truth-helpers`.
+Handlebars has its own definition of _truthyness_, which is similar to Javascripts except that empty arrays are
+considered **falsy**, while in JS are considered **truthy**.
+
+
+This RFC intentionally leaves the implementation details unspecified, but one can think of those helpers as macros that
+expand to combinations of `if`s.
+
+##### `{{and}}`
+- `{{and a b}}` is equivalent to `{{if a b a}}`
+- `{{and a b c}}` is equivalent to  `{{if a (if b c b) a}}`
+- and so on
+
+##### `{{or}}`
+- `{{or a b}}` is equivalent to `{{if a a b}}`
+- `{{and a b c}}` is equivalent to  `{{if a a (if b b c)}}`
+- and so on
+
+##### `{{not}}`
+- `{{not a}}` is equivalent to `{{if a false true}}`
+
+## How we teach this
+
+The introduction of these helpers does not impact the current mental model for Ember applications.
+
+In addition to API and Guides documentation with illustrative examples of usage of the various helpers, stressing out
+where the definition of truthiness of handlebars is different from the one in Javascript.
+
+## Drawbacks
+
+Adding new helpers increases the surface area of the framework and the code the core team commits to support long term.
+
+## Alternatives
+
+One alternative path is to not take any action and let users continue to define their own helpers (or install `ember-truth-helpers`).
+
+## Unresolved questions
+
+- Consider following Javascript's definition of truthiness. This would also help with the transition from `ember-truth-helpers`.
--- a/text/0000-add-logical-operators.md
+++ b/text/0000-add-logical-operators.md
@@ -69,7 +69,7 @@ expand to combinations of `if`s.
 
 ##### `{{or}}`
 - `{{or a b}}` is equivalent to `{{if a a b}}`
-- `{{and a b c}}` is equivalent to  `{{if a a (if b b c)}}`
+- `{{or a b c}}` is equivalent to  `{{if a a (if b b c)}}`
 - and so on
 
 ##### `{{not}}`
--- a/text/0000-add-logical-operators.md
+++ b/text/0000-add-logical-operators.md
@@ -77,10 +77,18 @@ expand to combinations of `if`s.
 
 ## How we teach this
 
-The introduction of these helpers does not impact the current mental model for Ember applications.
+While the introduction of these helpers doesn't introduce new concepts, as helpers like these could be
+written and in fact were written for a long time, it might affect slightly how we frame some concepts in the guides.
 
-In addition to API and Guides documentation with illustrative examples of usage of the various helpers, stressing out
-where the definition of truthiness of handlebars is different from the one in Javascript.
+Previously users were encouraged to put computed properties in the javascript file of the components, even for
+the most simple tasks like negating a boolean condition using `computed.not` or adding them with `computed.and`.
+
+With the addition of these helpers users don't have to resort to computed properties for simple operations, which sometimes
+forced users to create javascript files for what could have been template-only components.
+
+In addition to documenting the new helpers in the API docs, the Guides should be updated to favour the usage of helpers
+over computed properties where it makes more sense, adding illustrative examples and stressing out where
+the definition of truthiness of handlebars differs from the one of Javascript.
 
 ## Drawbacks
 
--- a/text/0000-add-logical-operators.md
+++ b/text/0000-add-logical-operators.md
@@ -20,7 +20,7 @@ other addons that those apps consume.
 The fact that `ember-truth-helpers` is so popular is a good signal that this it is filling a perceived gap in Ember's functionality.
 
 A second reason is that it might help make Ember more approachable to newcomers that have some experience in other frameworks.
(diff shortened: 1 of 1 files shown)


text/0000-add-logical-operators.md
```diff
@@ -0,0 +1,95 @@
...
+## Motivation
+
+It is a very common need in any sufficiently complex Ember app to perform some logical comparison operations and often the most convenient place to do it is right in the templates.
+Because of that, [ember-truth-helpers](https://github.com/jmurphyau/ember-truth-helpers) is one of the most installed addons out there, either directly by apps or indirectly by
+other addons that those apps consume.
+
+The fact that `ember-truth-helpers` is so popular is a good signal that this it is filling a perceived gap in Ember's functionality.
+
+A second reason is that it might help make Ember more approachable to newcomers that have some experience in other frameworks.
+Most if not all web frameworks have some way of comparing numbers in the templates and it's surprising that Ember requires an third party package to perform
```
chriskrycho:
```suggestion
Most if not all web frameworks have some way of performing logical operations in the templates and it's surprising that Ember requires an third party package to perform
```

text/0000-add-logical-operators.md
```diff
@@ -0,0 +1,95 @@
...
+once the first argument (`a`) is evalued to _falsey_, the second argument (`b`) is irrelevant. Sometimes arguments can be expensive properties to calculate,
+and by short-circuiting we can avoid computing them at all sometimes.
+
+
+## Detailed design
+
+Add `{{and}}`, `{{or}}` and `{{not}}` helpers.
+
+#### `{{and}}`
+Takes at least two positional arguments, bound or unbound. Raises an error if invoked with less than two arguments.
```
chriskrycho: For those of us following along who are less intimately acquainted with some of these details 😅 can you explain (not necessarily in this paragraph!) what is meant by “bound or unbound”? Thanks!

text/0000-add-logical-operators.md
```diff
@@ -0,0 +1,95 @@
...
+
+Adding new helpers increases the surface area of the framework and the code the core team commits to support long term.
+
+## Alternatives
+
+One alternative path is to not take any action and let users continue to define their own helpers (or install `ember-truth-helpers`).
+
+## Unresolved questions
+
+- Consider following Javascript's definition of truthiness. This would also help with the transition from `ember-truth-helpers`.
```
chriskrycho: Probably belongs under “Alternatives” instead.

text/0000-add-logical-operators.md
cibernox: I meant that you can pass bound arguments like local properties/arguments (e.g `{{or this.someProp @someArg}}`) or literal values like strings/numbers/booleans (e.g. `{{or @someArg "defaultValue"}}`)

I don't know if there is a better term for that.

chriskrycho: Ah, got it. I’d probably drop it entirely as unnecessary, then, unless there’s a specific reason to call it out. 🤔

text/0000-add-logical-operators.md
cibernox: I'd argue that it belongs here. The way I understand RFCs, alternatives is where you list alternative options to implementing the RFC, not so much about doubts regarding what the best implementation is.

If anyone can confirm that I got it wrong, I'm happy to change it.

text/0000-add-logical-operators.md
cibernox: You're maybe right, all helpers allow both afaik. I'll drop it.

text/0000-add-logical-operators.md
Gaurav0: I think an exact implementation of ember-truth-helpers logical operators (except maybe not) is not what anyone wants, given the desire for short circuit logic. But alternatives sections are not always fully specced out, so I think with just a little more detail, you could write an alternative.

webark: Being someone who believes that the lack of logical operators is the main benefit that handlebars provides, and one who has never had a need to install this addon in the many ember apps i have built, simply assuming that this is something that is required to be installed does not sit well with me. I would like somewhere in this process to highlight the benefits that not having these be a part of the framework provides.

Gaurav0:
> Being someone who believes that the lack of logical operators is the main benefit that handlebars provides, and one who has never had a need to install this addon in the many ember apps i have built, simply assuming that this is something that is required to be installed does not sit well with me. I would like somewhere in this process to highlight the benefits that not having these be a part of the framework provides.

I think this needs more discussion in the other helpers, but for these particular helpers, we can't build short circuit versions without direct core support. Would it be possible to build some primitives to make short circuiting possible while keeping the helpers (and actual semantics) in an addon?

text/0000-add-logical-operators.md
cibernox: @webark You can certainly live without those helpers, but I'll add some examples of things that are much nicer with helpers like those:

## Default Values in template-only components
Imagine a template only component that wants to have a default value for say, `@name`

```hbs
<img class="avatar" src={{or @user.avatar "/imgs/default-avatar.png">
```

## Perform an operation on each iteration of a loop without extracting another component
```hbs
{{#each @users as |user|}}
  Name: {{user.name}} 
  {{#if (and session.hasAdminPrivileges (not user.inactive))}}
    <button {{on "click" this.activate}}>Delete</button>
  {{/if}}
{{/each}}
```

In both situation you can find a way to do the same things without using any of those helpers, but it involve either create a JS file for a component only to put a computed property on it, or to extract a new component to call it on each iteration of a loop, just to perform a very simple logical operation.

It's a matter of convenience and as any tool it can be abused, but the popularity of ember-truth-helpers proves that devs really find it convenient.

I think that Ember providing just the most basic and used ones (`and`, `or`, `not`, `eq` are the main 4 for me, but this RFC focuses on the first 3) and optimized at the VM level we are alleviating a pain.

cibernox:
>I think this needs more discussion in the other helpers, but for these particular helpers, we can't build short circuit versions without direct core support. Would it be possible to build some primitives to make short circuiting possible while keeping the helpers (and actual semantics) in an addon?

@Gaurav0 I'll add it as a note, but even if it was possible and short-circuiting was possible in user-land, I still think those helpers are useful regardless of any optimization we may add, so I don't want to focus the RFC on the possible performance improvements. 

Probably on the big picture the perf improvements we may squeeze from short-circuiting will be rather small in anything but pretty contrived examples.

text/0000-add-logical-operators.md
webark: call me a curmudgeon (and i’ll postpone my “at this point we might as well be using jsx” rant for another day) I just feel it should be mentioned why this was never in core to begin with. Your examples, to me, are the base examples of why. Just a brief “handlebars was originally meant to be “logic-less” but it has grown, and the community has pushed it, past that for sometime now” blurb would be helpful, at least for a historical sense.

text/0000-add-logical-operators.md
```diff
@@ -0,0 +1,97 @@
...
+
+##### `{{not}}`
+- `{{not a}}` is equivalent to `{{if a false true}}`
+
+## How we teach this
+
+The introduction of these helpers does not impact the current mental model for Ember applications.
+
+In addition to API and Guides documentation with illustrative examples of usage of the various helpers, stressing out
+where the definition of truthiness of handlebars is different from the one in Javascript.
```
rwjblue: I think we need to beef this section a bit more. Specifically, I _do_ think this changes some of the mental models. For example, commonly we punt folks into JS-land instead of using ember-truth-helpers in the guides/tutorial (e.g. we might suggest using `computed.not`). This would definitely change if we introduce these helpers.

text/0000-add-logical-operators.md
```diff
@@ -0,0 +1,97 @@
...
+expand to combinations of `if`s.
+
+##### `{{and}}`
+- `{{and a b}}` is equivalent to `{{if a b a}}`
+- `{{and a b c}}` is equivalent to  `{{if a (if b c b) a}}`
+- and so on
+
+##### `{{or}}`
+- `{{or a b}}` is equivalent to `{{if a a b}}`
+- `{{and a b c}}` is equivalent to  `{{if a a (if b b c)}}`
```
chancancode:
```suggestion
- `{{or a b c}}` is equivalent to  `{{if a a (if b b c)}}`
```

cibernox:
@rwjblue I updated the How we teach section, check it out.

simonihmig:
Any updates on this? Given that the need for these helpers is quite ubiquitous I believe and trivial to implement (except for the short-circuiting maybe), advancing this RFC seems like a very low-hanging fruit to me!

I would like to add one thing though: has another helper like `{{default}}` been considered? It might not be truly a "truth helper", but `{{or}}` is also not *really* used as a boolean logic operator in most cases, but rather as a way to provide a default value, as in @cibernox's previous example:

```hbs
<img class="avatar" src={{or @user.avatar "/imgs/default-avatar.png"}}>
```

Tbh, I don't like this pattern that much, because
* it does not clearly express the *intent*, as this is not about some boolean operation that *returns* a boolean. Rather it relies on JavaScript's loosely typed interpretation of boolean logic, which in this case would return the second argument if the first evaluates to falsy 
* given that loosely typed behavior, this happens to work for string values as in the example above, but fails miserably for other values that should be considered valid values (i.e. which should not get overridden by the default value) but evaluate to falsy (`false`, `0`, `""`). Take this example: 

```hbs
<Modal @open={{or @modalOpen true}}/>
```

(The modal would always be open here, no matter what you pass as `@modalOpen`)

That's why I am so happy about the nullish coalescing (`??`) operator that we have now in JS-land, as it handles both these concerns (intent and behavior) much better than `||`. 

*Side node: you see I am not a fan of JavaScript's loose interpretation of boolean logic in general. Maybe because of the stronger mathematical education I enjoyed in school and university. I just find that boolean logic is soo easy to reason about when constrained to functions that take one or more booleans and return a boolean (truth tables!). As it was supposed to be! But JS semantics are obviously not going to change. So here we go, and sorry for my slightly off-topic rant here!* 😆 

So basically I would propose to add something like `{{default}}` (we can argue about naming of course), which is basically [implemented using `??`](https://github.com/kaliber5/ember-bootstrap/blob/master/addon/helpers/bs-default.js), just as `{{or}}` would *basically* use `||` (except for the slightly different handlebars semantics of truthiness).

You could argue that this can be added in its own RFC, *however*  I would argue that once you add `{{or}}` to Ember's core, it would perpetuate that use (IMHO misuse), and it would be hard to roll it back later.

rwjblue:
We chatted about this in todays core team meeting, and are in favor of moving forward here.

One thing to call out explicitly, this RFC does not propose import paths (which helpers would require [see the strict mode RFC](https://github.com/emberjs/rfcs/blob/master/text/0496-handlebars-strict-mode.md) for details). At the time of authoring, that distinction was not present (which is why it isn't discussed here), but after discussion with the core team we think that making these keywords (which do not need imports) is the correct path _anyways_ (therefore no changes are required to the RFC).

Additionally, landing this in Ember will require a small deprecation when using `ember-truth-helpers`'s `and`/`or`/`not` (since the semantics are subtlely different). In order to preserve the correct behavior in your application, Ember will use a resolved helper for these names if one is found (along with a deprecation).

---

All of that being said, we are moving this into final comment period (finally 😉).

cibernox:
Great. I think that making them keywords (so they don't have to be imported all the time) makes sense for such basic operators. Let's have it merged!

jelhan:
The RFC proposes that `{{and}}` and `{{or}}` keywords do not return a boolean but the value of one of its arguments. I don't think that's a good design choice.

The main motivation that I have seen for returning the value rather than a boolean is using `{{or}}` to define default values. I think @simonihmig raised valid concerns with that pattern. Additionally there is another RFC, _Argument Default Primitives_ (#695), since two weeks, which addresses the same problem. There wouldn't be any need to use `{{or}}` for default values in templates if _Argument Default Primitives_ or something similar lands.

I fear _not_ returning a boolean from keywords called `{{and}}` and `{{or}}` would cause confusion and will cause bugs. Especially if the keywords are not used with `{{if}}` or `{{unless}}` but to pass values to components, helpers or modifiers.

Let's assume a `<Foo>` component with a `@show` argument, which takes a boolean. Let's further assume that a consumer invokes it like this: `<Foo @show={{and (array)}} />`. Let's see what could go wrong depending on implementation details of `<Foo>`:

1. The component could validate the argument and throw because `@show` is not a boolean but an array. :boom: If the _Argument Validation Primitives_ RFC (#694) lands these kind of argument validation may get way more common that it is today.
2. The component consumes the value without validating it. Depending if consumes it in the template or in its JavaScript class it would be either considerd _falsy_ or _truthy_ . :scream: 

A consumer would need to cast the value returned by `{{and}}` and `{{or}}` to boolean to avoid these problems. So the correct invocation would look like this: `<Foo @show={{not (not (and (array)))}} />`. :exploding_head:

chriskrycho:
@jelhan these are all wholly valid points… but they’re also fundamental to how *JavaScript* works, and one of the things I see quite consistently as a challenge for folks working with Glimmer/Handlebars templates vs. working in React, Svelte, etc. is the points where the semantics of *these* kinds of things differ. They're surprising, and needlessly so, at least in context.

```js
let value = 1 || false; // 1
let flipped = false || 1; // 1
let anded = 1 && 2; // 2
let flipThat = 2 && 1; // 2
```

etc.

If we give `{{and}}` and `{{or}}` different semantics on this front in templates than in JavaScript, it increases the teaching burden *and* removes symmetry between these helpers and any such helpers you would implement yourself. In particular, given a world of template imports (🔜!), you can imagine someone being confused why the built-in version and the version they might implement themselves like this (with a hand-wave syntax) have different semantics:

```js
function or([a, b]) {
  return a || b;
});

export default<template>
  <div>{{or @something this.someOtherValue}}</div>
</template>
```

More/worse, there are times when you might *need* to switch from the built-ins to something slightly more complicated, and then you have to keep the differences in your head: refactoring from `{{or}}` to something which uses JS `||` is no longer straightforward but can have (sometimes wildly!) different semantics. The combination of increased burden on teaching (esp. for folks coming from other modern frameworks) *and* of ongoing things to keep in your head both point pretty conclusively in the direction of matching JS’s semantics here, even though JS’s own semantics are a mess.

chancancode:
> Let's assume a <Foo> component with a @show argument, which takes a boolean. Let's further assume that a consumer invokes it like this: <Foo @show={{and (array)}} />.

This is indeed not how it should be used and too complicated anyway. If you want strict Boolean values, you can just do `<Foo @show={{if array true false}} />`.

knownasilya:
The missing hole is `{{eq}}` that's the other one I use most.

jelhan:
@chriskrycho Thanks a lot for explaining the motivation. Wasn't clear to me that it should work as `||` and `&&` in JavaScript. But actually that makes sense.

I guess definition of truthiness is only exception for consistency within the template? To avoid a situation in which `{{and}}` consider an empty array as truthy which is considered falsy by `{{if}}`?

chrism:
> The missing hole is {{eq}} that's the other one I use most.

Honestly, I think its a great idea to include all these by default but this is the same issue for me.

Usually I add `ember-truth-helpers` in large part because `{{eq}}` is so useful.

So that means I would still probably need to add it anyway and then this difference in behaviour...

> This is NOT equivalent to the {{not}} helper from ember-truth-helpers because unlike this proposed helper, the one in ember-truth-helpers uses Javascript's definition of truthiness.

I guess would change subtlety and so require some care not to introduce bugs after the install?

rwjblue:
> The missing hole is `{{eq}}` that's the other one I use most.

Ya, that is being proposed in #560. Hopefully we can finalize our thoughts there (I made a comment explaining the current mental state of the core team recently) and release these two features _roughly_ together.

PR #569, opened 2020-01-03
TrackedList

pzuraq:
[Rendered](https://github.com/emberjs/rfcs/blob/tracked-list/text/0569-tracked-list.md)

### DIFF
--- a/text/0569-tracked-list.md
+++ b/text/0569-tracked-list.md
@@ -1,6 +1,6 @@
 - Start Date: 2019-12-23
 - Relevant Team(s): Ember.js
-- RFC PR: (after opening the RFC PR, update this with a link to it and update the file name)
+- RFC PR: https://github.com/emberjs/rfcs/pull/569
 - Tracking: (leave this empty)
 
 # TrackedList
--- a/text/0000-tracked-list.md
+++ b/text/0000-tracked-list.md
@@ -1,0 +1,306 @@
+- Start Date: 2019-12-23
+- Relevant Team(s): Ember.js
+- RFC PR: (after opening the RFC PR, update this with a link to it and update the file name)
+- Tracking: (leave this empty)
+
+# TrackedList
+
+## Summary
+
+Introduces a new array-like class whose usage is autotracked by default, and
+is designed with performance in mind.
+
+```js
+import { TrackedList } from '@glimmer/tracking';
+
+class Person {
+  friends = TrackedList.from(['Yehuda', 'Melanie', 'Ricardo']);
+
+  addFriend(newFriend) {
+    this.friends.push(newFriend);
+  }
+}
+```
+
+## Motivation
+
+`@tracked` in Ember Octane was primarily designed around the use case of
+tracking changes to _properties_ on classes. However, there are other forms of
+state that need to be changed in an Ember app, and the most common one is
+tracking changes to _arrays_.
+
+When `@tracked` was introduced, one of the suggested possibilities for tracking
+changes to arrays was the _immutable pattern_:
+
+```js
+class Person {
+  @friends = ['Yehuda', 'Melanie', 'Ricardo'];
+
+  addFriend(newFriend) {
+    this.friends = [...this.friends, newFriend];
+  }
+}
+```
+
+This pattern is solid and works very well for many Octane users, but it is
+fairly opinionated toward a functional-immutable style of programming, one that
+is different from many Ember apps historically, and that conflicts with
+established programming styles in the ecosystem. This is not necessarily a bad
+thing - one of autotracking's strengths is that it is unopinionated about
+programming style, and can be used with many different paradigms - but it means
+that users who prefer an Object Oriented Programming (OOP) style don't have a
+modern, ergonomic equivalent.
+
+We received feedback to this effect from a good number of Octane early adopters,
+which is why we decided to switch back to using `EmberArray` within the guides
+and documentation for our basic examples.
+
+```js
+import { A } from '@ember/array';
+
+class Person {
+  @friends = A(['Yehuda', 'Melanie', 'Ricardo']);
+
+  addFriend(newFriend) {
+    this.friends.pushObject(newFriend);
+  }
+}
+```
+
+`EmberArray` is reactive, so changes to it will update autotracked values and it
+has the behaviors most OOP users want. However, it has a lot of its own baggage:
+
+- KVO methods like `pushObject` and `popObject` have to be learned in order to
+  use the array, which means learning a whole new array API to be effective in
+  Ember.
+- Custom extra methods like `mapBy` and `sortBy` add to the confusion, since
+  they extend the core array API, and sometimes have conflicting names (e.g.
+  `any` vs `some`)
+- Prototype extensions can still be used, and it can be confusing to new users
+  since their DX is fairly nice, even though they have many downsides and are
+  generally recommended against.
+- Performance is not great, since they either rely on patching the array
+  prototype, which is very bad, or on re-implementing most array methods under
+  the hood.
+
+### Introducing `TrackedList`
+
+The goal of `TrackedList` is to provide a better `EmberArray`, in the short
+term, for Ember users who prefer this style of programming; One that lines up
+more directly with standard array behaviors, and is also reactive.
+
+```js
+import { TrackedList } from '@glimmer/tracking';
+
+class Person {
+  friends = TrackedList.from(['Yehuda', 'Melanie', 'Ricardo']);
+
+  addFriend(newFriend) {
+    this.friends.push(newFriend);
+  }
+
+  get firstFriend() {
+    let [first, ...rest] = this.friends;
+
+    return first;
+  }
+}
+```
+
+`TrackedList` will be as close to 1-1 with the native array API as possible.
+However, it will not attempt to be a drop-in replacement for native arrays,
+since this is not possible without native `Proxy` (which is not supported in
+older browsers) and would have major performance caveats even if it were. This
+is also the motivation behind the name: `TrackedArray` would imply a tracked
+version of native arrays, where `TrackedList` implies that it is its own data
+structure, with potentially divergent behaviors.
+
+## Detailed design
+
+`TrackedList` will re-implement the array API 1-1 for the most part, including
+native iterables in environments where they are available (enabling native
+destructuring syntax). Any existing prototype and static methods on `Array` will
+also exist on `TrackedList`, and future methods that are added will be added as
+soon as possible, without the need for an RFC (unless it would require
+significant changes to the way `TrackedList` works).
+
+As such, the rest of the design portion of this RFC will cover the tracking
+dynamics of `TrackedList`, and the ways in which it diverges from native arrays.
+
+### Tracking Dynamics
+
+In general, any changes to a `TrackedList` will be propagated as changes to the
+_whole_ list, rather than to individual indexes within the list. This is for two
+main reasons:
+
+1. Most changes will result in downstream consumers having to reprocess the
+   entire list, in general.
+2. It is much more performant to track changes to the list as a whole rather
+   than to each individual item within the list.
+
+In the following example, for instance, the `firstFriend` getter would be
+invalidated every time a new friend is added, even though it only accesses the
+first item in the list:
+
+```js
+import { TrackedList } from '@glimmer/tracking';
+
+class Person {
+  friends = TrackedList.from(['Yehuda', 'Melanie', 'Ricardo']);
+
+  addFriend(newFriend) {
+    this.friends.push(newFriend);
+  }
+
+  get firstFriend() {
+    let [first] = this.friends;
+
+    return first;
+  }
+}
+```
+
+From a change tracking perspective, we can see that this is actually very
+similar to the immutable pattern described at the beginning of this RFC - we
+are essentially telling the tracking system that some changes have occured, and
+now `friends` is _effectively_ a new array that needs to be reprocessed.
+
+In general, methods that _access_ the list and don't mutate it will entangle
+the list in the autotracking context. Methods that _mutate_ the list will
+invalidate, but not entangle the list, preventing backflow issues when
+rendering.
+
+### `getAt` and `setAt`
+
+It will not be possible to access the list items directly using `[]` syntax,
+like on normal arrays, since this would require native proxies (and would incur
+performance penalties). `TrackedList` will have two extra methods instead for
+direct index access: `getAt` and `setAt`. These methods will operate similar to
+`objectAt` and `replace` on `EmberArray` today, allowing the user to pass a
+numeric index as the first value, and in the case of `setAt` to pass the value
+to set the index to as the second:
+
+```ts
+declare class TrackedList<T> {
+  getAt(index: number): T;
+  setAt(index: number, value: T): T;
+}
+```
+
+### `sort` and `reverse`
+
+`sort` and `reverse` are unusual array methods, because they both _read_ the
+value, _and_ mutate the array, in place. Most of the time, users want to do one
+or the other, and from a change tracking perspective this is not ideal behavior.
+
+As such, `TrackedList` will instead always return a new list when either of
+these methods are called, allowing users to derive state without accidentally
+mutating upstream values.
+
+### `arr` in Array Callbacks
+
+Many array callbacks, such as `map` and `reduce`, receive the array as the third
+argument:
+
+```js
+[1, 2, 3].map((value, index, arr) => value + 1);
+```
+
+For performance reasons, `TrackedList` will not provide the list itself as the
+last argument. Instead, it will provide the underlying array, which is used as a
+storage mechanism to back the list. In `DEBUG` mode, this array will be wrapped
+with a Proxy which will:
+
+1. Alert users if they attempt to use non-standard methods such as `getAt` and
+   `setAt`.
+2. Throw an error if users try to mutate the array in any way. It should be
+   considered read-only for the purposes of usage within callbacks.
+
+This will allow common array-mapping libraries to be used interoperably with
+lists, without causing issues by accidentally mutating the internal storage of
+the list.
+
+### `Array.isArray`, `isEmberArray`, and Instance Checks
+
+`Array.isArray` and `isEmberArray` will return `false` for instances of
+`TrackedList`. Users can check to see if it is an instance using `instanceOf`:
+
+```js
+let friends = new TrackedList();
+
+friends instanceOf TrackedList; // true
+```
+
+### `length`
+
+`length` will be an immutable/readonly property on `TrackedList`. This is to
+prevent complications around sparse lists and sparse list iteration, and keep
+lists simpler overall. This restriction could be lifted in future RFCs.
+
+### `Ember.A`
+
+`Ember.A` will do nothing if it receives a `TrackedList`, and return the
+original value instead.
+
+## How we teach this
+
+`TrackedList` will be the default reactive array class recommended moving
+forward, and is meant to replace `EmberArray` entirely. The main guides should
+be updated to reflect this. In the advanced guides, there should be a section
+covering the functional-immutable style of updating arrays and objects.
+
+### API Docs
+
+TODO
+
+## Drawbacks
+
+- Introduces a new array-like class, which may cause confusion as the community
+  transitions.
+
+- Introduces an array class that _may_ have a limited shelf life. IE11 support
+  will eventually be removed, and allow us to use native `Proxy` to
+  transparently wrap real native arrays instead. However, this potential class
+  would have a few of its own caveats:
+  - Performance. Proxies have gotten much better, but they are still much less
+    performant compared to native arrays.
+  - General interop concerns. Arrays have a lot of odd behaviors that have been
+    built up over the years, such as the behaviors of sparse arrays, the
+    differences between how numeric and string based keys are handled, etc.
+  Given these caveats, and the fact that IE11 is still supported and will be for
+  the indeterminate future, a minimal array-like class that improves on the
+  existing DX seems like a good middle-ground solution for the time being.
+
+## Alternatives
+
+- `getAt` and `setAt` could be changed to match the `objectAt` and `replace`
+  methods that exist on `EmberArray`. This would be better for interoperability,
+  but comes with caveats:
+  - `objectAt` is not accurate for values that contain primitive values
+  - `replace` is a not an ideal method for updating a value at a predetermined
+    index. It's API is more similar to `splice`, which is confusing and
+    difficult to work with.
+  Some potential alternatives include:
+  - `objectAt` and `replaceAt`
+  - `valueAt` and `replaceAt`
+  - `valueAt` and `setValueAt`
+  `TrackedList` could also implement `objectAt` and `replace` in order to remain
+  mostly compatible with legacy code, but recommend that `getAt`/`setAt` be used
+  by convention.
+
+- Passing the underlying array as the third argument to methods like `map` and
+  `forEach` may be confusing to users. However, passing the list itself instead
+  would remove many of the browsers optimizations for these methods. They are
+  implemented in most modern runtimes as precompiled, pre-JITed code, and
+  matching anything near the level of native performance is likely impossible.
+  The other alternative would be to not pass a third argument at all (in
+  development mode), which would prevent users from relying on it in the first
+  place. This would limit the interoperability of `TrackedList` somewhat, but
+  would likely not have a massive impact.
+
+## Unresolved questions
+
+- How does this impact the potential functional array-methods RFC (functional
+  `objectAt` and `replace`, etc)?
+- Should `EmberArray` implement `getAt` and `setAt` to match the new API for
+  getting/setting indexes?


text/0569-tracked-list.md
```diff
@@ -0,0 +1,306 @@
...
+tracking changes to _properties_ on classes. However, there are other forms of
+state that need to be changed in an Ember app, and the most common one is
+tracking changes to _arrays_.
+
+When `@tracked` was introduced, one of the suggested possibilities for tracking
+changes to arrays was the _immutable pattern_:
+
+```js
+class Person {
+  @friends = ['Yehuda', 'Melanie', 'Ricardo'];
```
josemarluedke:
```suggestion
  @tracked friends = ['Yehuda', 'Melanie', 'Ricardo'];
```

text/0569-tracked-list.md
```diff
@@ -0,0 +1,306 @@
...
+
+We received feedback to this effect from a good number of Octane early adopters,
+which is why we decided to switch back to using `EmberArray` within the guides
+and documentation for our basic examples.
+
+```js
+import { A } from '@ember/array';
+
+class Person {
+  @friends = A(['Yehuda', 'Melanie', 'Ricardo']);
```
josemarluedke:
```suggestion
  friends = A(['Yehuda', 'Melanie', 'Ricardo']);
```

text/0569-tracked-list.md
```diff
@@ -0,0 +1,306 @@
...
+class Person {
+  friends = TrackedList.from(['Yehuda', 'Melanie', 'Ricardo']);
+
+  addFriend(newFriend) {
+    this.friends.push(newFriend);
+  }
+}
+```
+
+## Motivation
```
chriskrycho: I'd add to this that there's another (and IMO even more important than paradigm-preference) reason that this is valuable: performance. Immutable functional updates have many advantages, but without something like [persistent data structures](https://en.wikipedia.org/wiki/Persistent_data_structure), they can carry  a high performance penalty.

`TrackedList` allows users to make that tradeoff as they like, whatever their programming paradigm preference—and there's no reason that you can't provide a functional abstraction around it if that's your jam.

text/0569-tracked-list.md
```diff
@@ -0,0 +1,306 @@
...
+  they extend the core array API, and sometimes have conflicting names (e.g.
+  `any` vs `some`)
+- Prototype extensions can still be used, and it can be confusing to new users
+  since their DX is fairly nice, even though they have many downsides and are
+  generally recommended against.
+- Performance is not great, since they either rely on patching the array
+  prototype, which is very bad, or on re-implementing most array methods under
+  the hood.
+
+### Introducing `TrackedList`
```
chriskrycho: One concern I have with this name is that `List` suggests, well, *lists*, which are a standard abstraction with specific characteristics. At a minimum we will need to document how this `List` type is implemented, so that people will have a sense of the performance characteristics – and specifically, if this is implemented on top of JS arrays rather than as a linked list, we should be explicit about that (presumably adding a note to that effect in the **How We Teach This** section).

text/0569-tracked-list.md
```diff
@@ -0,0 +1,306 @@
...
+to set the index to as the second:
+
+```ts
+declare class TrackedList<T> {
+  getAt(index: number): T;
+  setAt(index: number, value: T): T;
+}
+```
+
+### `sort` and `reverse`
```
chriskrycho: Borrowing an idea from Swift, perhaps we should name these `sorted` and `reversed`, the difference of names implying the difference in outcome?

text/0569-tracked-list.md
```diff
@@ -0,0 +1,306 @@
...
+
+`TrackedList` will be as close to 1-1 with the native array API as possible.
+However, it will not attempt to be a drop-in replacement for native arrays,
+since this is not possible without native `Proxy` (which is not supported in
+older browsers) and would have major performance caveats even if it were. This
+is also the motivation behind the name: `TrackedArray` would imply a tracked
+version of native arrays, where `TrackedList` implies that it is its own data
+structure, with potentially divergent behaviors.
+
+## Detailed design
```
chriskrycho: It'd be great if we included the full API spec in this section!

text/0569-tracked-list.md
```diff
@@ -0,0 +1,306 @@
...
+  implemented in most modern runtimes as precompiled, pre-JITed code, and
+  matching anything near the level of native performance is likely impossible.
+  The other alternative would be to not pass a third argument at all (in
+  development mode), which would prevent users from relying on it in the first
+  place. This would limit the interoperability of `TrackedList` somewhat, but
+  would likely not have a massive impact.
+
+## Unresolved questions
+
+- How does this impact the potential functional array-methods RFC (functional
```
chriskrycho: Whoops: incomplete sentence!

text/0569-tracked-list.md
```diff
@@ -0,0 +1,306 @@
...
+  `forEach` may be confusing to users. However, passing the list itself instead
+  would remove many of the browsers optimizations for these methods. They are
+  implemented in most modern runtimes as precompiled, pre-JITed code, and
+  matching anything near the level of native performance is likely impossible.
+  The other alternative would be to not pass a third argument at all (in
+  development mode), which would prevent users from relying on it in the first
+  place. This would limit the interoperability of `TrackedList` somewhat, but
+  would likely not have a massive impact.
+
+## Unresolved questions
```
chriskrycho: Is `TrackedArray` to be subclassed? If so, what invariants do subclasses need to uphold?

text/0569-tracked-list.md
```diff
@@ -0,0 +1,306 @@
...
+
+```js
+let friends = new TrackedList();
+
+friends instanceOf TrackedList; // true
+```
+
+### `length`
+
+`length` will be an immutable/readonly property on `TrackedList`. This is to
```
chriskrycho:
```suggestion
`length` will be a readonly property on `TrackedList`. This is to
```

I *think* if I'm understanding the intent here correctly, we should describe this as `readonly`, since it *will* change if users `push` or `pop` from the list.

Also, is `length` auto-tracked? (I'd assume yes, but I'd love if it were explicit here.)

text/0569-tracked-list.md
```diff
@@ -0,0 +1,306 @@
...
+}
+```
+
+`TrackedList` will be as close to 1-1 with the native array API as possible.
+However, it will not attempt to be a drop-in replacement for native arrays,
+since this is not possible without native `Proxy` (which is not supported in
+older browsers) and would have major performance caveats even if it were. This
+is also the motivation behind the name: `TrackedArray` would imply a tracked
+version of native arrays, where `TrackedList` implies that it is its own data
+structure, with potentially divergent behaviors.
```
chriskrycho: A couple questions as I'm considering this:

- What would those "potentially divergent behaviors be"?
- Is it possible to ship a `TrackedArray` *now*, with future performance improvements unlocked when IE11 is removed (whenever Ember 4 ships), and if so, how do the tradeoffs of doing that compare to the tradeoffs of introducing this class now and then potentially *another* class later?

pzuraq: 1. Specifically, the two divergences in this RFC are:
  * The difference in behaviors of `sort` and `reverse`
  * The readonly nature of `length`
2. It is not possible to ship `TrackedArray` without native Proxy, so we would have to wait for IE11 support. The native Proxy version is still much much slower than a native array based solution, by at least an order of magnitude, currently.

chriskrycho: Thanks for the clarifications!

text/0569-tracked-list.md
hjdivad: I had exactly this question.

Also, what is the intended API for a subclass to manually invalidate itself?

pzuraq: `TrackedList` is not intended to be subclassed, though there isn't a way to prevent it in JS, and there is no API for manual invalidation. If users want to implement a custom iterable, the recommendation would instead be to use the standard `Symbol.iterable` API to do so, as it should be fully supported by Ember at this point. If there are still some gaps, they should probably be addressed in a separate RFC.

bendemboski:
Here's something to maybe argue about here, or maybe just to add to the drawbacks section (or as a sub-point of the first drawback). I'm generally not a fan of object APIs that look really really like standard/native object APIs, but differ slightly. All too often I find one of the significant challenges in reading and understanding code in dynamically typed languages is figuring out what a particular thing is. Good documentation or, better yet, typescript should largely ameliorate this, but it's still a significant factor.

So I'm not a huge fan of implementing something that looks really really like a native array, but is missing one or two pieces (`[]` getter/setter). With `EmberArrays`, a very high percentage of API calls are typically using the non-native(-looking) methods, so it's generally easy to figure out when something is an `EmberArray` and I need to use its methods. But I worry that it will be fairly common to encounter

```javascript
function doSomeStuff(items) {
  let things = items.map((i) => i.getThing());
  for (let thing of things) {
    thing.doStuff();
  }
}
```

and maybe change it to

```javascript
function doSomeStuff(items) {
  items[0].getThing().doStuff();
  let things = items.slice(1).map((i) => i.getThing());
  for (let thing of things) {
    thing.doDifferentStuff();
  }
}
```

except whoops, `items` is (or sometimes is) a `TrackedList` and my code blows up, or sometimes blows up because I can't `[0]`.

I'm kinda ambivalent because I'm not sure I'd really prefer a situation where all the methods do the same thing as `Array` methods, but have different names (like `EmberArray`). But whether I want to or not, I often rely on mental duck typing when reading code, and having secret agent impostor ducks makes thing harder.

lifeart:
Why we can't use `[]` for values access/write?

```js
Object.defineProperty([], '0', {
  get: function() { return this._impl[0]; },
  set: function(newValue) { this._impl[0] = newValue; },
  enumerable: false,
  configurable: true
});
```

Writing to "uninitialized" index of array - it's really "corner" case and we can skipt it at all, mention it as anti-pattern in documentation.

pzuraq:
@lifeart that would be very inefficient, and a leaky abstraction since there is no way to know when a value has been initialized. At that point it would be better to wait for and use native Proxy.

@bendemboski I understand the concerns you've raised, and they're totally valid. Unfortunately, we cannot make a fully transparent array class at the moment without native Proxy, and even when it becomes possible it may not be performant enough for some use cases (my own naive micro-benchmarks show that native arrays are still 100-times faster than a naive Proxy wrapper).

Personally, I feel like `EmberArray` is the worst of both worlds that you describe. Some methods like `pushObject` and `popObject` do signal that the object is in fact an `EmberArray`, but plenty methods do not - `map` for instance is probably one of the most common methods I use in an array, and I could use an `EmberArray` in many places only using `map` or `forEach`, and thus no longer have a reliable hint. On top of that, you are forced to learn a new, confusing API if you want to use `EmberArray`. I think I'd prefer an API where _all_ methods are hinted, or _none_ are, so it is at least consistent.

We also _can_ use native proxy in development, so for instance if you tried to do:

```js
let first = trackedList[0];
```

We can throw an error to the user and let them know they need to use `getAt(0)`. Maybe this would help with the uncanny valley effect you're describing.

lifeart:
@pzuraq why `no way`? there is limited set methods to `add` value to array

pzuraq:
```js
let arr = [];

arr[0] = 123;
```

Is valid, and not possible to instrument without Proxy. It would mean that you cannot transparently pass the array around, especially in non Ember code, because some other method would have to be used to setup the (extremely expensive) native getter and setter.

lifeart:
do we really have such performance problem? (in real life apps)?
there is cases we have > 10k arrays, but I don't think it's > 20% of all app arrays (in user space)

lifeart:
> ```js
> let arr = [];
> 
> arr[0] = 123;
> ```
> 
> Is valid, and not possible to instrument without Proxy. It would mean that you cannot transparently pass the array around, especially in non Ember code, because some other method would have to be used to setup the (extremely expensive) native getter and setter.

we can use Babel and replace `[]` to `new MyArray([])` and we will be able instrument 
`a[0] = 1` without problems

// svelte transforms

pzuraq:
I’m personally more worried about the leaky abstraction than the performance problem. IMO, any solution that attempts to use native get/set syntax must be able to do so perfectly, so we must wait for native Proxy.

Users who don’t need to support IE11 today can of course use native proxies to create an array wrapper as is. For an official Ember solution, I think the options are to continue using EmberArray until we drop support for IE11, or to pursue something like TrackedList.

pzuraq:
> we can use Babel and replace [] to new MyArray([]) and we will be able instrument
a[0] = 1 without problems

Let me rephrase:


```js

let arr = new MyArray([]);

arr[0] = 123;
```

Is also impossible to instrument. It is impossible to do this for any and all JS objects without native Proxy.

lifeart:
@pzuraq possible https://github.com/emberjs/rfcs/pull/569#issuecomment-572920736

In software development there is no perfect solution, it's always balance between simplicity, performance, maintainability, extensibility.

pzuraq:
@lifeart that abstraction is still leaky, which is my overall point. There may not be perfect solutions, but there are _correct_ solutions, and native Proxy is the only one that is correct in the way you want it to be. I don't believe we can ship anything less correct.

`TrackedList` may indeed end up being a temporary solution. I think the performance angle is still worth considering, though I tend to agree that unless performance is absolutely critical, the benefits of a drop in wrapper are probably worth it. But the fact is, we cannot have `TrackedArray` today.

This has been a continual problem in our design process, that we let the perfect be the enemy of the good. The point of this proposal is to make things incrementally better. `TrackedList` would simplify the current design and give us a path off native prototype extensions, custom array methods, etc. It can be implemented quickly, and since its API lines up with native arrays very closely, it would make a transition path toward a hypothetical `TrackedArray` incrementally easier than directly from `EmberArray`.

Personally, I want a slightly better Ember today much more than a much better Ember tomorrow. We may just have to agree to disagree on that.

lifeart:
Is it possible to ignore `mutable` arrays at all?
-> Offer immutable arrays?
(don't fight with a[b]=c issues, and mention it on documentation level)

pzuraq:
The immutable pattern is fully supported, and we definitely should mention it in the guides and have an In Depth guide on its usage.

text/0569-tracked-list.md
```diff
@@ -0,0 +1,306 @@
...
+  since their DX is fairly nice, even though they have many downsides and are
+  generally recommended against.
+- Performance is not great, since they either rely on patching the array
+  prototype, which is very bad, or on re-implementing most array methods under
+  the hood.
+
+### Introducing `TrackedList`
+
+The goal of `TrackedList` is to provide a better `EmberArray`, in the short
+term, for Ember users who prefer this style of programming; One that lines up
```
hjdivad: The RFC should probably mention, if only as an aside, what the long term preference is, if not `TrackedList`.

hjdivad: Presumably it's `Proxy`, but it's not totally obvious since later on when `Proxy` is discussed a couple of important caveats are emphasized.

Gaurav0: IMHO, if this is not the long term solution, it should not be added to core and should remain in an addon.

marcoow:
I understand that this can be helpful for people who prefer the pre-Octane way of dealing with mutable/tracked (Ember) arrays but isn't the fact that `@tracked` removes the need for Ember-specific data structures and the cognitive overhead that comes with them actually an advantage? This would add something that's very similar to `Ember.Array` effectively with similar cognitive overhead again.

So while I'm not completely opposed to this, the above is just the first thought I had when seeing this. Would it be an option to release this in an addon if we want it at all and not make it part of standard Ember? The same comment applies to #577 of course.

pzuraq:
@marcoow yeah, I definitely understand that and agree that in general we want to make the programming model really feel like it's "Just JavaScript" still, and any extra layering on top of vanilla JS should be carefully considered and thought through, since it'll add cognitive overhead.

For context, the guides and official recommendations in Octane still all use `EmberArray`, and I don't believe there is consensus to change that in the near future. The issue is that the "immutable model" that was originally put forward in the Tracking RFCs, where you would reassign the array via `this.arr = [...this.arr, newValue]`, had some pitfalls:

* There are performance implications for duplicating arrays and objects every time with `...spread`, and it many cases it makes more sense to mutate the array.
* Removing an item from an array using `...spread` syntax is really difficult, as is adding an item in the middle of an array.
* Notification style syntax (e.g. `arr.push(foo); this.arr = this.arr;`) also doesn't feel like vanilla JS, and neither does manual notification (e.g. (`invalidate(this, 'arr')`)). Both add cognitive overload, and in can be very error prone if someone forgets to notify.
* There isn't really a good, ergonomic way to notify in all the locations an array is used, if it happens to be passed around and assigned/used in multiple locations. Both of the above styles can only provide notifications locally, and to children, but there are many cases where an array might be used in multiple locations (e.g. Ember Data collections).
* Collections like `Map` and `Set` don't really have an ergonomic solution here, spread syntax does not work as well. `WeakMap` and `WeakSet` _cannot_ be updated with spread syntax, and many of the common use cases for them (e.g. associating values with a module-scope `WeakMap`) also cannot use the notification style, since they aren't associated with a specific property.

This is not to say that it would be completely off the table to change the guides and recommendations to use this style, it's just that these patterns need to be figured out. Additionally, potential future JS features like [Records and Tuples](https://github.com/tc39/proposal-record-tuple) would make this style much nicer and possibly solve many of the concerns listed above. Libraries like Immer exist today, but I feel like we would have to be very confident in them to recommend them by default.

I suppose from a personal perspective, I really started to feel like this was the right solution as I began to work more with autotracking and think through its principles, from a theory perspective. The fundamental assumption that autotracking makes is that all root state that is input to the system is tracked. Currently, we only expose ways to tracked specific properties by default, but mutable collections like arrays generally exist, and we want to track changes to them as well. The current solution feels like it's breaking that principle by asking you to notify on something unrelated when you mutate an array, map, or set.

FWIW, I'm really excited about Records and Tuples though and could really see them be used this way. Since they're immutable, they don't introduce any of these problems, which is perfect! I could also definitely see a guide for using Immer in Ember apps.

mehulkar:
As exciting and cool as this is, I think the focus of 2020 should be to make the surface area of Ember smaller, not bigger. In real world scenarios, the tradeoff at PR review time is to either nudge the author of code towards understanding the reactivity model and its caveats or using `TrackedList`. I would much prefer the first, as it's much easier to grasp why Ember can't react to internal changes of an object, than to understand why we can't use `Array` as a rule (or worse, having to decide when to use an Array or TrackedList).

I would like to see this shipped as a library completely agnostic of Ember though (not an Ember addon) and let users opt in.

marcoow:
@pzuraq thanks for the in-depth reply. I understand there are pitfalls with the immutable model and agree there needs to be a way to get around those. I think I'm just worried (similar to what @mehulkar is saying) that as we're now in a situation where `@tracked` allows us to get rid of many of Ember's specialties (like `Ember.Array`) we should not immediately go back to a similar situation as before by adding new versions of these specialties when encountering problems with the new approach (of course there are substantial differences between `Ember.Array` and `TrackedList` but essentially the latter is very similar to the former in the sense that it is Ember's special data type for arrays/lists that one has to learn in addition to learning Ember, one has to be aware when to use which etc.). I do think `TrackedList` makes sense in general and it would be good if it was available in some form for people that run into the above pitfalls but it seems like this could perfectly well be in an addon for now which would still not mean it could never be added to the framework itself later on or the addon could be added to the default blueprint etc.

pzuraq:
@mehulkar @marcoow I definitely understand these concerns. I think the current state of things is though that we _haven't_ gotten rid of those extra wrappers. Specifically, we currently cannot deprecate or remove `EmberArray`, since it is the default and recommended approach for dealing with lists in the framework.

We can come up with alternative solutions in the community, but until we can get consensus on an alternative default for arrays/lists, I believe `EmberArray` and its quirks will have to stay. Maybe it would be worth waiting for `Proxy` at least then, since it would mean we could potentially design a drop-in array replacement, but I personally would like to be in a position where we can deprecate `EmberArray` sooner rather than later.

256dpi:
I personally like that auto-tracking is just limited to classes and the root level properties. In the past I had situations where a `EmberArray` was leaked to some other place and caused unintended re-computations as it was changed. Since the methods are different to native arrays the problem was easy to spot and fix. However, I don't want to think about what happens if a proxied native array gets passed into some external library and cause re-renders all over the place. Even if it would be possible, I wouldn't want arrays, POJOs etc. to magically cause re-renders if they are mutated. For that reason, I would not use and recommend tracking of such basic primitive types.

But, I agree that the auto-tracking system is not very ergonomically when it comes to arrays, POJOs and other non-trackable types. The current example in the guides shows that first hand:

```js
class SimpleCache {
  @tracked _cache = {};

  set(key, value) {
    this._cache[key] = value;

    // trigger an update
    this._cache = this._cache;
  }

  get(key) {
    return this._cache[key];
  }
}
```

Yet, I belief there are other solutions that could be explored here. A quick idea of the back of my head would be to provide a magic `tracked` property that provides access to all other properties but automatically schedules a re-computation in the next cycle when accessed:

```js
class SimpleCache {
  @tracked _cache = {};

  set(key, value) {
    this.tracked._cache[key] = value;
  }

  get(key) {
    return this._cache[key];
  }
}
```

Maybe with something like that there is no need for `TrackedX`. And if there is still one, it can be quickly implemented by hand and adjusted for the specific use case.

sandstrom:
I think something proxy-based would be much better and [browser support is very good](https://caniuse.com/#search=proxy). I think it would be much better to design something for the 93% of browsers that already support proxies. Especially since IE 11 is in rapid decline.

When will Ember drop IE 11 support?

Many large software companies has already stopped supporting it, for example:

- https://blog.developer.atlassian.com/retiring-ie11-support-for-atlassian-cloud-server-and-data-center-products/

- https://support.zendesk.com/hc/en-us/articles/360002084367-Removal-of-support-for-Internet-Explorer-11

- https://support.qlik.com/articles/000079718

Regarding the performance drawback, I don't think it would be a big problem under most real-world scenarios.

luxferresum:
> There isn't really a good, ergonomic way to notify in all the locations an array is used, if it happens to be passed around and assigned/used in multiple locations. Both of the above styles can only provide notifications locally, and to children, but there are many cases where an array might be used in multiple locations (e.g. Ember Data collections).

Could we not `Array` and add a `notify` method? Then I do `arr[2] = foo; arr.notify()`.

I dont like the current way no notify a change `this.foo = this.foo` but I think we wont find a way around manual notification. And actually this is not bad. But whenever we get data from an external non ember-aware library we need to do manual notification. Not matter if its a nested object or an array. So learning this is important, and then better use the same pattern then to introduce another custom Array-Like structure.

pzuraq:
@luxferresum there are two main issues I see with that potential solution:

1. Modifying the built-in prototypes is generally considered a very bad idea. Not only can it cause future compatibility issues (e.g. if `Array` adds its own `notify` method in the future), it also deoptimizes the browser's built in version of the array, which is pre-JIT'd and ready to run, forcing it to recompile and re-optimize all of the array's methods. If we didn't add the method directly to `Array.prototype`, then we'd be in the same boat as before with introducing a new class, and it would probably make sense to do more than adding `notify`.

2. Entanglement is not just about notification - we must also _consume_ the array, which can only be done when the array is used. Consider this example, which models the game [Lights Out](https://en.wikipedia.org/wiki/Lights_Out_(game)):

   ```js
   class LightsOut {
     lights = [
       [0, 0, 1, 0, 0],
       [0, 1, 0, 1, 0],
       [1, 0, 0, 0, 1],
       [0, 1, 0, 1, 0],
       [0, 0, 1, 0, 0],
     ];

     isLightOn(x, y) {
       return this.lights[y][x] === 1;
     }
   }
   ```

   In the autotracking model, whenever an autoracking frame wraps a calculation, all state that can affect the _outcome_ of that calculation should entangle itself. In this case, the calculation is whenever we call `isLightOn` with a particular `x` and `y` coordinate. With `notify()` we would have no way to tell the consuming API that the arrays have been _used_ when we call `isLightOn`. Normally, accessing the property that is `@tracked` does this, but in this case it won't work since we're working with a 2d array. The best solution here would be to notify all of the lights when any one of them has changed, by notifying the `@tracked` property itself.

> But whenever we get data from an external non ember-aware library we need to do manual notification.

This is actually not how we expect most users to have to work with external libraries. Autotracking is a very flexible reactivity system, and is designed to allow libraries to be wrapped transparently. In general, manually entangling and instrumenting notifications from external libraries is something that should be done once in a wrapper library, and then within Ember apps it should be consumable without having to worry about the details of how the library autotracks. The assumption should be that if the library mutates state, it also correctly entangles and notifies that state.

For examples of what this looks like in action, we can look at Ember Data and [tracked-built-ins](https://github.com/pzuraq/tracked-built-ins). In the case of Ember Data, it provides the `@attr`, `@belongsTo`, and `@hasMany` decorators, which all autotrack seamlessly. These are currently computed properties, but in the future they will likely be stripped down to the bare minimum and become thin wrappers around lower level autotracking entanglement and notification APIs. In the case of tracked-built-ins, the library instruments a very basic external (at the moment) data source, and notifies on any changes to it.

This same strategy could be applied to, say, Apollo, Redux, or any number of other external libraries. We are working on stabilizing autotracking APIs so that instrumenting libraries is easier and the community can begin experimenting more easily, but I definitely think something we want to avoid is manually notifying for every single external library, every time it is used in an app. How to do it should be well documented and understood though, it's not that these wrapper layers would provide "magic", they would just would place the entanglement where it logically belongs - with the source of the data.

wagenet:
How does this compare to https://github.com/tracked-tools/tracked-built-ins?

chriskrycho:
I believe this is superseded by the combination of #669 and #812. #669 gave us the tools needed to implementing things like this in user-land, and #812 adds the most common built-in versions that users will need, implemented using those primitives.

sandstrom:
@wagenet Awesome with all the RFC cleanup lately! 🎉

PR #570, opened 2020-01-04
URL Manager

NullVoxPopuli:
[rendered](https://github.com/nullvoxpopuli/rfcs/blob/url-primitives/text/0000-url-primitives.md)

### DIFF
--- a/text/0000-url-primitives.md
+++ b/text/0000-url-primitives.md
@@ -30,7 +30,6 @@ when interacting with query params.
 
 If no URL Manager is configured, Ember's current URL behavior will be preserved.
 
-
 ## Examples
 
 The following examples of the proposed API are written in TypeScript in _userland / application-space_ to better 
--- a/text/0000-url-primitives.md
+++ b/text/0000-url-primitives.md
@@ -1,6 +1,6 @@
-- Start Date: (fill me in with today's date, YYYY-MM-DD)
-- Relevant Team(s): (fill this in with the [team(s)](README.md#relevant-teams) to which this RFC applies)
-- RFC PR: (after opening the RFC PR, update this with a link to it and update the file name)
+- Start Date: 2020-01-04
+- Relevant Team(s): Ember.js
+- RFC PR: https://github.com/emberjs/rfcs/pull/570
 - Tracking: (leave this empty)
 
 # URL Manager
--- a/text/0000-url-primitives.md
+++ b/text/0000-url-primitives.md
@@ -149,7 +149,7 @@ class CustomURLManager extends URLManager {
     };
   }
 
-  toURL(routeInfo: RouteInfo) {
+  toURL(routeInfo: RouteInfo): string {
     let { 
       mapInfo: {
         segments // [blogs, :blogId, posts, :postId]
@@ -238,6 +238,25 @@ class CustomURLManager extends URLManager {
 
 1. URLManager is a container-controlled object, so that it may utilize the dependency injection system.
   Primary need for this is to access the router service to utilize existing APIs for transforming URLs and `RouteInfo`s
+  
+`URLManager` has the following API:
+
+```ts
+class URLManager {
+  static create(injections: any) {
+    return new this(injections);
+  }
+
+  constructor(injections: any) {
+    Object.assign(this, injections);
+  }
+  
+  public toURL(routeInfo: RouteInfo, dynamicSegments: object): string;
+  public fromURL(url: string): RouteInfo;
+}
+```
+
+and `toURL` and `fromURL` implement today's URL (de)serialization.
 
 ### Notes
 
--- a/text/0000-url-primitives.md
+++ b/text/0000-url-primitives.md
@@ -67,7 +67,7 @@ Router.map(function() {
 
 ### Naïve Query Params without Controllers
 
-This example proposes the possibility of an alternate strategy for managing query params without the need for controllers. 
+This example proposes the possibility of an alternative strategy for managing query params without the need for controllers. 
 Note that this would not change or alter the existing query param behavior in any way.
 
 ```ts
--- a/text/0000-url-primitives.md
+++ b/text/0000-url-primitives.md
@@ -341,7 +341,6 @@ module('Unit | Route | blogs/blog/posts/post', function(hooks) {
 Prior Art: 
 - the manager patterns from elsewhere in Ember.
 
-
 ## Unresolved questions
 
 - Should the URL Manager exist as a static config or an instantiable object per-route? The above proposal is a static config / singleton, but allowing an instance per route would allow for more varied state buckets, but could also make debugging harder as there would be an URL Manager for each route segment.
--- a/text/0000-url-primitives.md
+++ b/text/0000-url-primitives.md
@@ -237,6 +237,7 @@ class CustomURLManager extends URLManager {
   Primary need for this is to access the router service to utilize existing APIs for transforming URLs and `RouteInfo`s
 
 ### Notes
+
 1. `MapInfo` is static or "frozen" -- it is only constructed at the time of router setup.
 2. `MapInfo` represents the optional `object` argument of `Router.map`'s `this.route`. Empty object if not configured.
 3. If needed, the URL Manager may be used from a component: 
--- a/text/0000-url-primitives.md
+++ b/text/0000-url-primitives.md
@@ -32,7 +32,7 @@ If no URL Manager is configured, Ember's current URL behavior will be preserved.
 
 ## Examples
 
-The following examples of the proposed API are written in TypeScript in _userland / application-space_ to better 
+The following examples of the proposed API are written in TypeScript in _userland_/_application-space_ to better 
 demonstrate intended API and available data.
 
 The same naming semantics and conventions in today's router system will still apply. 
--- a/text/0000-url-primitives.md
+++ b/text/0000-url-primitives.md
@@ -50,7 +50,7 @@ class CustomURLManager extends URLManager {
 export default class Router extends EmberRouter {
   location = config.locationType;
   rootURL = config.rootURL;
-  urlManager = CustomURLManager; // Defined in Examples
+  urlManager = CustomURLManager;
 });
 
 Router.map(function() {
--- a/text/0000-url-primitives.md
+++ b/text/0000-url-primitives.md
@@ -11,7 +11,7 @@ Supersedes [Add queryParams to the router service](https://github.com/emberjs/rf
 
 A URL Manager will provide the ability to manipulate the URL, 
 both when writing to `window.location` (serializing), 
-and reading from the `window.location` (deserializing). 
+and reading from `window.location` (deserializing). 
 The ultimate goal of the URL Manager will be to enable 
 app and addon authors to customize the use of the URL -- 
 including i18n'd routes, dynamic segment slugs / aliases, 
--- a/text/0000-url-primitives.md
+++ b/text/0000-url-primitives.md
@@ -268,7 +268,7 @@ import config from 'app-name/config/environment';
 export default class Router extends EmberRouter {
   location = config.locationType;
   rootURL = config.rootURL;
-});
+}
 
 Router.map(...);
 ```
--- a/text/0000-url-primitives.md
+++ b/text/0000-url-primitives.md
@@ -113,6 +113,7 @@ class CustomURLManager extends URLManager {
 ```
 
 ### i18n routes
+
 In this scenario, we may want to internationalized route segments.
 
 For example, in English, we may want a route to be `/blogs/1/posts/2`, 
--- a/text/0000-url-primitives.md
+++ b/text/0000-url-primitives.md
@@ -71,7 +71,6 @@ Router.map(function() {
 This example proposes the possibility of an alternate strategy for managing query params without the need for controllers. 
 Note that this would not change or alter the existing query param behavior in any way.
 
-
 ```ts
 // app/router.js/ts
 import { inject as service } from '@ember/service';
--- a/text/0000-url-primitives.md
+++ b/text/0000-url-primitives.md
@@ -171,6 +171,7 @@ class CustomURLManager extends URLManager {
 ```
 
 ### Custom URL Management per-route
+
 There are situations in which links and their routes don't conform to the rest of the application. For these situations, here is an example showing how manage the URL separately for those routes.
 
 ```ts
--- a/text/0000-url-primitives.md
+++ b/text/0000-url-primitives.md
@@ -10,7 +10,7 @@ Supersedes [Add queryParams to the router service](https://github.com/emberjs/rf
 ## Summary
 
 A URL Manager will provide the ability to manipulate the URL, 
-both when writing to the `window.location` (serializing), 
+both when writing to `window.location` (serializing), 
 and reading from the `window.location` (deserializing). 
 The ultimate goal of the URL Manager will be to enable 
 app and addon authors to customize the use of the URL -- 
--- a/text/0000-url-primitives.md
+++ b/text/0000-url-primitives.md
@@ -216,7 +216,7 @@ class CustomURLManager extends URLManager {
 
 ## Detailed design
 
-### Additions to existing Public APIs
+### Additions to existing public APIs
 1. add full list of dynamicSegments to RouteInfo so that the task of building out the map of dynamic segments to their values isn't in user-space
 2. Restriction of Query Params on controllers / `<LinkTo />` needs to have a way to opt-out. 
    Today, if a query param is added to a `<LinkTo />` and that query param is not present on the target `route`'s controller, the query param is removed from the link. 
--- a/text/0000-url-primitives.md
+++ b/text/0000-url-primitives.md
@@ -5,8 +5,6 @@
 
 # URL Manager
 
-TODO: check if routeInfo or route has a way to build a URL from routeInfo (need full dynamic segments)
-
 Supersedes [Add queryParams to the router service](https://github.com/emberjs/rfcs/pull/380)
 
 ## Summary
@@ -223,7 +221,8 @@ class CustomURLManager extends URLManager {
 2. Restriction of Query Params on controllers / `<LinkTo />` needs to have a way to opt-out. 
    Today, if a query param is added to a `<LinkTo />` and that query param is not present on the target `route`'s controller, the query param is removed from the link. 
    Query Param allow/deny lists could be re-implemented using the Router `MapInfo` / options.
-3. The router's urlFor should be able to take a RouteInfo / RouteInfo should be able to be converted to an URL
+3. The router's urlFor helper function should be able to take a RouteInfo / RouteInfo should be able to be converted to an URL
+   - delegates to `urlManager.toURL` for `RouteInfo`
 4. Static `MapInfo` object reference added to each `RouteInfo`.
 5. routerService.mapInfoFrom should take: path / url / routeInfo -- uses existing recognize method
 
@@ -250,6 +249,7 @@ class CustomURLManager extends URLManager {
       }
     }
     ```
+4. `transitionTo` and `replaceWith` APIs are unaffected.
 
 ### The Default URL Manager
 
--- a/text/0000-url-primitives.md
+++ b/text/0000-url-primitives.md
@@ -14,7 +14,7 @@ both when writing to `window.location` (serializing),
 and reading from `window.location` (deserializing). 
 The ultimate goal of the URL Manager will be to enable 
 app and addon authors to customize the use of the URL -- 
-including i18n'd routes, dynamic segment slugs / aliases, 
+including internationalized routes, dynamic segment slugs / aliases, 
 and alternate query params behavior -- though, 
 implementation of those specific things is outside the scope of *this* RFC. 
 
--- a/text/0000-url-primitives.md
+++ b/text/0000-url-primitives.md
@@ -15,7 +15,7 @@ and reading from `window.location` (deserializing).
 The ultimate goal of the URL Manager will be to enable 
 app and addon authors to customize the use of the URL -- 
 including internationalized routes, dynamic segment slugs / aliases, 
-and alternate query params behavior -- though, 
+and alternative query params behavior—though, 
 implementation of those specific things is outside the scope of *this* RFC. 
 
 ## Motivation
--- a/text/0000-url-primitives.md
+++ b/text/0000-url-primitives.md
@@ -9,7 +9,7 @@ Supersedes [Add queryParams to the router service](https://github.com/emberjs/rf
 
 ## Summary
 
-An URL Manager will provide the ability to manipulate the URL -- 
+A URL Manager will provide the ability to manipulate the URL, 
 both when writing to the `window.location` (serializing), 
 and reading from the `window.location` (deserializing). 
 The ultimate goal of the URL Manager will be to enable 
--- a/text/0000-url-primitives.md
+++ b/text/0000-url-primitives.md
@@ -232,6 +232,7 @@ class CustomURLManager extends URLManager {
    This allows more control over the allow/deny list behavior of query params filtering.
 
 ### Additional APIs / Behavior
+
 1. URLManager is a container-controlled object, so that it may utilize the dependency injection system.
   Primary need for this is to access the router service to utilize existing APIs for transforming URLs and `RouteInfo`s
 
--- a/text/0000-url-primitives.md
+++ b/text/0000-url-primitives.md
@@ -5,6 +5,8 @@
 
 # URL Manager
 
+TODO: check if routeInfo or route has a way to build a URL from routeInfo (need full dynamic segments)
+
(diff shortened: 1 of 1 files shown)


text/0000-url-primitives.md
```diff
@@ -0,0 +1,349 @@
...
+- RFC PR: https://github.com/emberjs/rfcs/pull/570
+- Tracking: (leave this empty)
+
+# URL Manager
+
+Supersedes [Add queryParams to the router service](https://github.com/emberjs/rfcs/pull/380)
+
+## Summary
+
+An URL Manager will provide the ability to manipulate the URL --
```
locks:
```suggestion
A URL Manager will provide the ability to manipulate the URL, 
```

text/0000-url-primitives.md
```diff
@@ -0,0 +1,349 @@
...
+- Tracking: (leave this empty)
+
+# URL Manager
+
+Supersedes [Add queryParams to the router service](https://github.com/emberjs/rfcs/pull/380)
+
+## Summary
+
+An URL Manager will provide the ability to manipulate the URL -- 
+both when writing to the `window.location` (serializing),
```
locks:
```suggestion
both when writing to `window.location` (serializing), 
```

text/0000-url-primitives.md
```diff
@@ -0,0 +1,349 @@
...
+
+# URL Manager
+
+Supersedes [Add queryParams to the router service](https://github.com/emberjs/rfcs/pull/380)
+
+## Summary
+
+An URL Manager will provide the ability to manipulate the URL -- 
+both when writing to the `window.location` (serializing), 
+and reading from the `window.location` (deserializing).
```
locks:
```suggestion
and reading from `window.location` (deserializing). 
```

text/0000-url-primitives.md
```diff
@@ -0,0 +1,349 @@
...
+Supersedes [Add queryParams to the router service](https://github.com/emberjs/rfcs/pull/380)
+
+## Summary
+
+An URL Manager will provide the ability to manipulate the URL -- 
+both when writing to the `window.location` (serializing), 
+and reading from the `window.location` (deserializing). 
+The ultimate goal of the URL Manager will be to enable 
+app and addon authors to customize the use of the URL -- 
+including i18n'd routes, dynamic segment slugs / aliases,
```
locks:
```suggestion
including internationalized routes, dynamic segment slugs / aliases, 
```

text/0000-url-primitives.md
```diff
@@ -0,0 +1,349 @@
...
+
+## Summary
+
+An URL Manager will provide the ability to manipulate the URL -- 
+both when writing to the `window.location` (serializing), 
+and reading from the `window.location` (deserializing). 
+The ultimate goal of the URL Manager will be to enable 
+app and addon authors to customize the use of the URL -- 
+including i18n'd routes, dynamic segment slugs / aliases, 
+and alternate query params behavior -- though,
```
locks:
```suggestion
and alternative query params behavior—though, 
```

text/0000-url-primitives.md
```diff
@@ -0,0 +1,349 @@
...
+
+Supersedes [Add queryParams to the router service](https://github.com/emberjs/rfcs/pull/380)
+
+## Summary
+
+An URL Manager will provide the ability to manipulate the URL -- 
+both when writing to the `window.location` (serializing), 
+and reading from the `window.location` (deserializing). 
+The ultimate goal of the URL Manager will be to enable 
+app and addon authors to customize the use of the URL --
```
locks:
```suggestion
app and addon authors to customize the use of the URL—
```

text/0000-url-primitives.md
```diff
@@ -0,0 +1,349 @@
...
+This means that we are unable to provide internationalized URLs, 
+we cannot customize the (de)serialization of query params 
+(which is important because there is no standard format for query params).
+
+These APIs will also provide an opportunity to iterate on a better default query params implementation that will enable an objectively better development experience for app devs 
+when interacting with query params.
+
+If no URL Manager is configured, Ember's current URL behavior will be preserved.
+
+
```
locks:
```suggestion
```

text/0000-url-primitives.md
```diff
@@ -0,0 +1,349 @@
...
+## Examples
+
+The following examples of the proposed API are written in TypeScript in _userland / application-space_ to better
```
locks:
```suggestion
The following examples of the proposed API are written in TypeScript in _userland_/_application-space_ to better 
```

text/0000-url-primitives.md
```diff
@@ -0,0 +1,349 @@
...
+        // index route would be the "list" of posts
+        this.route('post', { path: ':id' });
+      })
+    });
+  })
+});
+```
+
+
+### Naiive Query Params without Controllers
```
locks:
```suggestion
### Naïve Query Params without Controllers
```

text/0000-url-primitives.md
```diff
@@ -0,0 +1,349 @@
...
+
+This example proposes the possibility of an alternate strategy for managing query params without the need for controllers.
```
locks:
```suggestion
This example proposes the possibility of an alternative strategy for managing query params without the need for controllers. 
```

text/0000-url-primitives.md
```diff
@@ -0,0 +1,349 @@
...
+
+    let query = qs.stringify(queryParams);
+
+    return `/${url}?${query}`; // => /blogs/1/posts/2?foo=bar
+  }
+}
+```
+
+### i18n routes
+In this scenario, we may want to internationalized route segments.
```
locks:
```suggestion

In this scenario, we may want to internationalized route segments.
```

text/0000-url-primitives.md
```diff
@@ -0,0 +1,349 @@
...
+    let query = qs.stringify(queryParams);
+
+    return `/${url}?${query}`; // => /블로그/1/게시물/2?foo=bar
+  }
+}
+
+
+```
+
+### Custom URL Management per-route
```
locks:
```suggestion
### Custom URL Management per-route

```

text/0000-url-primitives.md
```diff
@@ -0,0 +1,349 @@
...
+There are situations in which links and their routes don't conform to the rest of the application. For these situations, here is an example showing how manage the URL separately for those routes.
```
locks:
```suggestion
There are situations in which links and their routes don't conform to the rest of the application. For these situations, here is an example showing how to manage the URL separately for those routes.
```

text/0000-url-primitives.md
```diff
@@ -0,0 +1,349 @@
...
+
+    // ...
+  }
+}
+```
+
+## Detailed design
+
+### Additions to existing Public APIs
+1. add full list of dynamicSegments to RouteInfo so that the task of building out the map of dynamic segments to their values isn't in user-space
```
locks:
```suggestion

1. Add full list of dynamicSegments to `RouteInfo` so that the task of building out the map of dynamic segments to their values isn't in userspace;
```

text/0000-url-primitives.md
```diff
@@ -0,0 +1,349 @@
...
+    if (toURL) return toURL(routeInfo, dynamicSegments);
+
+    // ...
+  }
+}
+```
+
+## Detailed design
+
+### Additions to existing Public APIs
```
locks:
```suggestion
### Additions to existing public APIs
```

text/0000-url-primitives.md
```diff
@@ -0,0 +1,349 @@
...
+  - can often match on incorrect portions of the Route if not thoroughly tested
+
+- Only adding QueryParams modifications (per RFC #380)
+  - not flexible enough
+  - forces a single query params implementation
+
+Prior Art: 
+- the manager patterns from elsewhere in Ember.
+
+
```
locks:
```suggestion
```

text/0000-url-primitives.md
```diff
@@ -0,0 +1,349 @@
...
+});
+```
+
+
+### Naiive Query Params without Controllers
+
+This example proposes the possibility of an alternate strategy for managing query params without the need for controllers. 
+Note that this would not change or alter the existing query param behavior in any way.
+
+
```
locks:
```suggestion
```

text/0000-url-primitives.md
```diff
@@ -0,0 +1,349 @@
...
+   Today, if a query param is added to a `<LinkTo />` and that query param is not present on the target `route`'s controller, the query param is removed from the link. 
+   Query Param allow/deny lists could be re-implemented using the Router `MapInfo` / options.
+3. The router's urlFor helper function should be able to take a RouteInfo / RouteInfo should be able to be converted to an URL
+   - delegates to `urlManager.toURL` for `RouteInfo`
+4. Static `MapInfo` object reference added to each `RouteInfo`.
+5. routerService.mapInfoFrom should take: path / url / routeInfo -- uses existing recognize method
+
+
+### Changes to Internals
+1. `<LinkTo />` and other related router helpers use `toURL` and `fromURL` from the URL Manager
```
locks:
```suggestion

1. `<LinkTo />` and other related router helpers use `toURL` and `fromURL` from the URL Manager.
```

text/0000-url-primitives.md
```diff
@@ -0,0 +1,349 @@
...
+   This allows more control over the allow/deny list behavior of query params filtering.
+
+### Additional APIs / Behavior
+1. URLManager is a container-controlled object, so that it may utilize the dependency injection system.
```
locks:
```suggestion

1. URLManager is a container-controlled object, so that it may utilize the dependency injection system.
```

text/0000-url-primitives.md
```diff
@@ -0,0 +1,349 @@
...
+  Primary need for this is to access the router service to utilize existing APIs for transforming URLs and `RouteInfo`s
+
+### Notes
+1. `MapInfo` is static or "frozen" -- it is only constructed at the time of router setup.
```
locks:
```suggestion

1. `MapInfo` is static or "frozen" -- it is only constructed at the time of router setup.
```

text/0000-url-primitives.md
```diff
@@ -0,0 +1,352 @@
...
+
+  fromURL(url: string): RouteInfo { // /블로그/1/게시물/2?foo=bar
+    let [path, query] = url.split('?');
+    let segments = path.split('/');
+
+    // this dosen't have to be english, but it does have to match the names as
+    // defined in Router.map(...);
+    let english = 
+      path.split('/')
+          .segments.map(segment => {
```
buschtoens: I think you meant:

```ts
let english = segments.map(segment => {
```

text/0000-url-primitives.md
```diff
@@ -0,0 +1,352 @@
...
+            return this.i18n.lookup(`routes.from.${segment}`, 'en-us') || segment;
```
buschtoens: Wouldn't this imply that the `en-us` locale contains the mappings from all other languages' segments to the names used in the code? Shouldn't this mapping actually be part of every individual locale?

text/0000-url-primitives.md
```diff
@@ -0,0 +1,352 @@
...
+## Summary
+
+A URL Manager will provide the ability to manipulate the URL, 
+both when writing to `window.location` (serializing), 
+and reading from `window.location` (deserializing). 
+The ultimate goal of the URL Manager will be to enable 
+app and addon authors to customize the use of the URL—
+including internationalized routes, dynamic segment slugs / aliases, 
+and alternative query params behavior—though, 
+implementation of those specific things is outside the scope of *this* RFC.
```
Turbo87: isn't that something that the `Location` API can already handle? 🤔

text/0000-url-primitives.md
```diff
@@ -0,0 +1,352 @@
...
+
+The same naming semantics and conventions in today's router system will still apply. 
+
+All examples will have the following in common:
+```ts
+// router.js
+import EmberRouter, { URLManager } from '@ember/routing/router';
+import config from 'app-name/config/environment';
+
+// ... Example CustomURLManager here ...
```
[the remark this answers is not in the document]
NullVoxPopuli: This is much clearer! thanks!

text/0000-url-primitives.md
```diff
@@ -0,0 +1,352 @@
...
+```ts
+// app/router.js/ts
+import { inject as service } from '@ember/service';
+import qs from 'qs';
+
+class CustomURLManager extends URLManager {
+  @service router;
+
+  fromURL(url: string): RouteInfo { // /blogs/1/posts/2?foo=bar
+    let [path, query] = url.split('?');
```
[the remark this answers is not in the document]
NullVoxPopuli: are multiple `?` valid? wouldn't you need to escape additional `?`? 

fwiw, I'm not proposing that these examples be actual implementation. They need tests, and there are def cases not covered

text/0000-url-primitives.md
```diff
@@ -0,0 +1,352 @@
...
+  - forces a single query params implementation
+
+Prior Art: 
+- the manager patterns from elsewhere in Ember.
+
+## Unresolved questions
+
+- Should the URL Manager exist as a static config or an instantiable object per-route? The above proposal is a static config / singleton, but allowing an instance per route would allow for more varied state buckets, but could also make debugging harder as there would be an URL Manager for each route segment.
+
+- `toURL` / `fromURL` or `serialize` / `deserialize`?
```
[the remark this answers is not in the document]
NullVoxPopuli: is engine support totally in addon space? I know nothing of engines? how would URL manipulation affect engines?

or is it more just that there could be multiple routers?

text/0000-url-primitives.md
NullVoxPopuli: Location api can't handle nested query params, which is a common ask

text/0000-url-primitives.md
NullVoxPopuli: I did!

text/0000-url-primitives.md
NullVoxPopuli: > Wouldn't this imply that the en-us locale contains the mappings from all other languages' segments to the names used in the code?

yes.

> Shouldn't this mapping actually be part of every individual locale?

maybe. I don't know how molt people would want to implement this. Ember's route's are typically named in english, hence the approach here.

text/0000-url-primitives.md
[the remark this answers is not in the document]
buschtoens: Engine support is in core and addon space. [Routable engines](http://ember-engines.com/guide/core-concepts#routablevsrouteless) partake in the routing process. They define their [routing map in `routes.js`](http://ember-engines.com/guide/creating-an-engine#addingroutesforroutableengines) and are [mounted in the parent routing map](http://ember-engines.com/guide/mounting-engines#routableengines) via [`mount(name: string, options?: MountOptions)`](https://github.com/emberjs/ember.js/blob/f5d9878fdcd0d73970aaf5445cfe46cdba91fe58/packages/%40ember/-internals/routing/lib/system/dsl.ts#L33-L34).

The engine routing maps are basically merged with the host routing map — it's a bit more complicated — but engines don't have their own `Router` instances.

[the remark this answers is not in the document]
NullVoxPopuli: ah ok, then I think everything should be compat with this.

App defines _the_ CustomURLManager, and if needed, engines could override that per route

text/0000-url-primitives.md
buschtoens: I just wanted to note, that if used like this, it'd probably be wiser to remove the `en-us` in favor of the current locale.

text/0000-url-primitives.md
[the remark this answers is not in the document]
buschtoens: Yes, they are valid: https://stackoverflow.com/a/2924187/420747

Not really common though.

text/0000-url-primitives.md
```diff
@@ -0,0 +1,353 @@
...
```
gossi: > `toURL` / `fromURL` or `serialize` / `deserialize`?

definitely `toURL` / `fromURL` - with `(de)serialize` you are constantly question which direction is serialize? 😇

text/0000-url-primitives.md
```diff
@@ -0,0 +1,353 @@
...
+
+All examples will have the following in common:
+```ts
+// router.js
+import EmberRouter, { URLManager } from '@ember/routing/router';
+import config from 'app-name/config/environment';
+
+class CustomURLManager extends URLManager {
+  // Concrete example implementations are shown in the following sections.
+}
```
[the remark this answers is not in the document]
NullVoxPopuli: that's what I had initially -- this was added as a clarification by another reviewer. The examples don't duplicate the router setup. imo, for learning intent purposes, it's good to keep. Would def not recommend an empty class if this were implemented.

text/0000-url-primitives.md
```diff
@@ -0,0 +1,353 @@
...
+      mapInfo,
+      queryParams: {}
+      dynamicSegments: { blog: '1', post: '2' },
+    }
+    let resultRouteInfo = fromURL(toURL(sampleRouteInfo));
+
+    assert.deepEqual(resultRouteInfo, sampleRouteInfo);
+  });
+});
+```
```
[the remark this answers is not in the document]
NullVoxPopuli: where would a test for the url-manager live? `tests/unit/url-manager.js`?

text/0000-url-primitives.md
```diff
@@ -0,0 +1,353 @@
...
+    let { mapInfo: { toURL } } = routeInfo;
+
+    if (toURL) return toURL(routeInfo, dynamicSegments);
+
+    // ...
+  }
+}
+```
+
+## Detailed design
```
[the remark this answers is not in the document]
NullVoxPopuli: Sure thing. never sure how much duplicate to add in these things

NullVoxPopuli:
> but my bigger concern is: adding further complexity to the internals of the router seems extremely high risk to me. 

I'm all ears for other solutions that would allow people to customize query param (de)serialization as well as control how the URL is (de)serialized in general. 

This is RFC is entirely in response to comments from #380. ;)

> the worst of the bugs we fixed were extremely subtle issues related to changes deep in the router.

should this RFC be implemented, all existing tests should be unchanged. :-\
I think that it's possible to implement all the existing behavior using what's proposed here, and if we have gaps in tests... we'll just need to add more tests.

> I'd very much like to see Ember's router become much simpler

Do you have ideas for how this could be simpler?

> As such, I'm concerned by the amount of new API surface around the existing router implementation.

🤷‍♂ it's 1 extra router config option, and 2 configs per route. :-\
Doesn't _seem_ like that big of a deal regarding API Surface. The router config already has minimal API surface.

text/0000-url-primitives.md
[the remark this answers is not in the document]
NullVoxPopuli: added

text/0000-url-primitives.md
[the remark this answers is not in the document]
NullVoxPopuli: legit. so, whenever someone implements this, they should def not do a naive split on '?', but split on the first occurrence of `?`

chriskrycho:
Don't have time or mental bandwidth to do design work, alas, but—

> Doesn't seem like that big of a deal regarding API Surface. The router config already has minimal API surface. But maybe I'm missing something? idk.

It's not just that it's more API surface (though that's *a* concern) but that every one of these kinds of things introduces more *internal* implementation, and much of the router's current behavior—not least around URL resolutions, including *especially* for QPs—is most accurately described as "emergent and incredibly complex". Every new code path has surprising opportunities to introduce bugs there, and "just add more tests" isn't really a good answer there: there's already a *ton* of test coverage, but because the existing implementation has all that emergent complexity, there are many interactions which are difficult to cover exhaustively, and there are many places where the issues come down to subtle timing details. And of course: if we *knew* what needed more coverage, someone would write it; the problems that come up in this space are from *unknown unknowns*.

To be clear, I'm not trying to come crap on this design. I'm just saying that all of those concerns should be considered very high priority constraints on this and other designs in this space, given the relatively high likelihood of accidental breakage even when extreme care is taken in the design.

NullVoxPopuli:
> introduces more internal implementation

I was planning on replacing as much of the internal implementation with this stuff as makes sense.
Also, isn't this true of _any_ feature? I'm confused.

> is most accurately described as "emergent and incredibly complex"

that's exactly why I'm proposing some primitives for us to un-complexify URL/QP behavior.

> To be clear, I'm not trying to come crap on this design.

of course ;) 'just also kind feels like FUD. :-\

chriskrycho:
> replacing as much of the internal implementation with this stuff as makes sense.

Frankly, then: this *will* cause exactly the kind of issues I’m describing. This is not FUD; that’s frankly rude. I’m happy to help further clarify why and how this is risky, but I’m not making crap up; I’m highlighting a risk area that it’s *perfectly fair and reasonable* for you to have been unaware of.  I was unaware until a few months ago! :smile:

chriskrycho:
Let me try to put this a different way: I think this idea or something like it is good and necessary. We can’t avoid making changes to high risk areas forever because otherwise it just gets worse and worse.

However, we may need a *different strategy* for how we run a transition in this area because of the high risk (especially for large apps like the one I work on!) than we need for most features. The point I’m trying to make is not “this is bad; let’s not do it” but “this is much more complicated than it appears; let’s think further about what it will take to accomplish this, possibly including very different paths to getting to the desired end state proposed”.

mehulkar:
Love the idea, bur echoing @cheiskrycho, I would also like to see the existing router reduce complexity and expose existing abstractions for experimentation before expose introducing a new concept. I think a URLManager is a good end goal but agree that it’s high risk to try to get here before going through some intermediate steps first.

NullVoxPopuli:
@mehulkar what existing stuff would you like to see exposed?

NullVoxPopuli:
the apis proposed here are intended to be the low level entrypoints into anything you'd want from the router -- as is, 
This is how the URL is interpreted: https://github.com/emberjs/ember.js/blob/release/packages/%40ember/-internals/routing/lib/system/router.ts#L663
This is how query params are serialized: https://github.com/emberjs/ember.js/blob/release/packages/%40ember/-internals/routing/lib/system/router.ts#L712 -- this RFC would be moving this code, along with deserialize: https://github.com/emberjs/ember.js/blob/release/packages/%40ember/-internals/routing/lib/system/router.ts#L757

atm, I don't know where the router URL gets pushed back in to the address bar.

I don't think it'd be a good idea to make these methods public, as it would make testing the URL transforming behavior more difficult if it were customized, as you'd need to setup the entire router -- and that would take you out of the realm of unit testing. :-\

text/0000-url-primitives.md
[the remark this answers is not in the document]
villander: I agree with @chriskrycho, we should write integration and acceptance tests for them

mehulkar:
I wrote about some ideas here: https://mehulkar.com/blog/2019/12/post-octane-ember-routing/. I'm not particularly saying the a URLManager for serialization/deserialization is bad, but more that there's no good place to put it yet, because the routing surface area is so big. I'd like to see `the `Router` class go away entirely, for example. We don't have to agree on that particular part (and it's not part of this discussion), but my point is that I would "clean the house before adding more furniture".

NullVoxPopuli:
@mehulkar 
> I'd like to see the `Router` class go away entirely, for example

what do you imagine would take it's place? ideal world, no consequences

mehulkar:
@NullVoxPopuli https://mehulkar.com/blog/2019/12/post-octane-ember-routing/. tldr; 

- app/router.js can be replaced with a function that defines the route map
- `Application` can receive the other configs (e.g. locationType directly instead of through Router class
- `router:main` is exposed in the container, but should be replaced with `service:router`.

Gaurav0:
> It's not just that it's more API surface (though that's a concern) but that every one of these kinds of things introduces more internal implementation, and much of the router's current behavior—not least around URL resolutions, including especially for QPs—is most accurately described as "emergent and incredibly complex". Every new code path has surprising opportunities to introduce bugs there, and "just add more tests" isn't really a good answer there: there's already a ton of test coverage, but because the existing implementation has all that emergent complexity, there are many interactions which are difficult to cover exhaustively, and there are many places where the issues come down to subtle timing details. And of course: if we knew what needed more coverage, someone would write it; the problems that come up in this space are from unknown unknowns.

To be honest I'm not an expert on the router (@machty is the only expert), but *any* backward compatible way of dealing with query params is going to add more surface area to the api until we deprecate controllers. We do need to find a way forward.

NullVoxPopuli:
Closing this for now due to lack of time / community interest :) 
Can resurrect if there is interest

sandstrom:
FWIW I think modularity around URL serialization/deserialization can be useful and liked the idea of this RFC. Sorry to hear that there wasn't enough interest.

For example, we've implemented our custom location type (https://api.emberjs.com/ember/release/classes/Location) to do locale-prefixing of routes. Basically, all routes have a `/locale-code` prefix, e.g. `/en/about` and `/de/about` and really just one single `/about` route in the router.

But by putting the locale in the route we get two versions of the same website, with different languages, plus make it easier for search engines to parse.

Obviously possible already today with custom Locations, but this seems like an even more flexible solution.

PR #572, opened 2020-01-10
Promise based initializers

mehulkar:
[rendered](https://github.com/mehulkar/rfcs/blob/promise-initializers/text/0000-promise-based-initializers.md)

### DIFF
--- a/text/0000-promise-based-initializers.md
+++ b/text/0000-promise-based-initializers.md
@@ -12,15 +12,24 @@ and pause application boot until that promise resolves.
 
 ## Motivation
 
-Currently, the earliest place that developers can write asynchronous code is in the ApplicationRoute.
-For some use cases, this is not early enough, and it is useful to be able to run asynchronous code
-before the application begins routing.
+**Earlier Async Hooks**
 
-For example, when booting an Ember application in an embedded environment, it may be necessary to
-listen for events from the environment to determine the initial state to determine the initial URL.
+Currently, the earliest place developers can write blocking asynchronous code is
+in the ApplicationRoute's `beforeModel()` hook. There are two problems with this:
 
-Another example may be to interact with 3rd party libraries that need to be initialized early but
-have an asynchronous API.
+1. Addons cannot easily add behavior to this hook. To do so, they must reopen the
+`Ember.Route` class, and then require that all Routes in the application remember
+to call `super`. This approach does not scale well, but will also become harder
+as Ember.Route moves away from extending from `CoreObject`, and thus losing the
+`.reopen()` API.
+1. Semantically, code can make more _sense_ to belong in the application lifecycle,
+rather than the routing stack. For example, if your Ember application depends on an
+embedded environment (an Electron shell or even an iframe), it may make more sense to
+implement blocking code _before_ hitting the routing stack. This can also translate
+to technical wins, as initializers run before routing starts, so, for example,
+an initializer could determine the starting URL of the application (much like
+window.location does today). Yet another example of better codifying developer patterns
+is the ability to import or initialize 3rd part libraries with async APIs.
 
 ## Detailed Design
 
@@ -41,12 +50,12 @@ For example:
 ```js
 // app/initializers/foo.js
 export function initialize() {
-    return Promise.resolve();
+  return Promise.resolve();
 }
 
 export default {
-    initialize,
-    async: true
+  initialize,
+  async: true,
 };
 ```
 
@@ -63,19 +72,21 @@ If implemented, future RFCs could:
 1. The guides would need to document the `async` key of the exported object from initializers and instance-initializers.
 2. Some advanced logging to see when initializer functions are called may be useful.
 3. Some information about how this relates to `deferReadiness` and `advanceReadiness` would be useful.
-It's possible that deprecating these APIs as part of this RFC is appropriate, so there's a clear
-signal that one is preferred over the other.
+   It's possible that deprecating these APIs as part of this RFC is appropriate, so there's a clear
+   signal that one is preferred over the other.
 
 ## Drawbacks
 
 - There could be some performance concerns over introducing promises at boot time, but because this
-is an opt-in feature, they can be avoided.
+  is an opt-in feature, they can be avoided.
 
 ## **Alternatives**
 
 - The existing `{defer|advance}Readiness` APIs could be introduced for instance-initializers as well.
-Although promise-based async isn't exactly the same as this API, and there could be subtle differences
-between both, it is unlikely that end-users would actually end up caring.
+  Although promise-based async isn't exactly the same as this API, and there could be subtle differences
+  between both, it is unlikely that end-users would actually end up caring.
+- Deprecate initializers entirely and create new primitives for hooking into the
+application boot lifecycle using the `Application` class.
 
 ## **Unresolved questions**
 
@@ -83,4 +94,4 @@ between both, it is unlikely that end-users would actually end up caring.
 - What happens if a promise is created in an async initializer but not returned?
 - Does this affect the algorithm that orders initializers based on `before` and `after` APIs?
 - Does this affect the `boot` vs `_bootSync` of Application and ApplicationInstance at all and
-should that be cleaned up *before* attempting this?
+  should that be cleaned up _before_ attempting this?
--- a/text/0000-promise-based-initializers.md
+++ b/text/0000-promise-based-initializers.md
@@ -1,6 +1,6 @@
 - Start Date: 2020-01-09
 - Relevant Team(s): Ember.js
-- RFC PR: (after opening the RFC PR, update this with a link to it and update the file name)
+- RFC PR: https://github.com/emberjs/rfcs/pull/572
 - Tracking: (leave this empty)
 
 # Promise-based Initializers
--- a/text/0000-promise-based-initializers.md
+++ b/text/0000-promise-based-initializers.md
@@ -1,0 +1,86 @@
+- Start Date: 2020-01-09
+- Relevant Team(s): Ember.js
+- RFC PR: (after opening the RFC PR, update this with a link to it and update the file name)
+- Tracking: (leave this empty)
+
+# Promise-based Initializers
+
+## Summary
+
+This RFC proposes the ability to return a promise from an initializer or instance-initializer
+and pause application boot until that promise resolves.
+
+## Motivation
+
+Currently, the earliest place that developers can write asynchronous code is in the ApplicationRoute.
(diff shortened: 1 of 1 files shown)


text/0000-promise-based-initializers.md
```diff
@@ -0,0 +1,86 @@
...
+and pause application boot until that promise resolves.
+
+## Motivation
+
+Currently, the earliest place that developers can write asynchronous code is in the ApplicationRoute.
+For some use cases, this is not early enough, and it is useful to be able to run asynchronous code
+before the application begins routing.
+
+For example, when booting an Ember application in an embedded environment, it may be necessary to
+listen for events from the environment to determine the initial state to determine the initial URL.
```
NullVoxPopuli: would using the beforeModel in the application route work?

mehulkar: For this particular use cases, the application route is too late because there are subtle differences between what it means to *start* routing at a URL and what it means to redirect to a URL. When an Ember app runs in a browser, it does the former.

That said, I think promise based initializers will have benefits beyond just determining the initial state of the application.

text/0000-promise-based-initializers.md
```diff
@@ -0,0 +1,86 @@
...
+## Summary
+
+This RFC proposes the ability to return a promise from an initializer or instance-initializer
+and pause application boot until that promise resolves.
+
+## Motivation
+
+Currently, the earliest place that developers can write asynchronous code is in the ApplicationRoute.
+For some use cases, this is not early enough, and it is useful to be able to run asynchronous code
+before the application begins routing.
```
kpfefferle: Do we have known use cases where the Application route "is not early enough"? I'm not doubting the truth of this statement, but having an example use case or cases would make it easier for me to wrap my head around how this would be useful.

mehulkar: Sure! If you have a webview embedded in a native layer that sends you events (say when the native app is launched via deeplink). Capturing that event and operating on it is an asynchronous operation. If you wait till the `ApplicaitonRoute`, the Ember app has already started routing, so the only option to get to the desired route is `redirect()`. That's starkly different from what happens when the Ember app is initialized at a subpath in the browser, however and it can cause redirect hell if you receive more than one of these events.

Gaurav0:
Really like this!

Could we add an afterAsync key or something like that so that one initializer could wait for another one to resolve?

text/0000-promise-based-initializers.md
```diff
@@ -0,0 +1,86 @@
...
+have an asynchronous API.
+
+## Detailed Design
+
+The `Application` and `ApplicationInstance` (extending from `Engine` and `EngineInstance`
+classes respectively) are responsible for running initializers and instance initializers. Each
+implements a method that loops through the loaded initializers and executes the exported function.
+
+Initializer and Instance Initializer files each export an object with an `initialize` property
+that references a function. For backwards compatibility, an `async` property should be added, and
```
xg-wang: If the `initialize` return type can be detected do we still need `async`? I think handling async initializers based on this option or the actual return type affects unresolved questions below.

text/0000-promise-based-initializers.md
NullVoxPopuli: You can't redirect from the application route though. It,s the parent of everything.

NullVoxPopuli:
I'd be more of a fan of deprecating intializers as a whole. I don't think they're needed.

Ime, everything initializers can do is better handled in the application route. Either in a before model, or the init/constructor.

text/0000-promise-based-initializers.md
mehulkar: Good call out. I debated over the need for this key too. The main reason it's useful is that you cannot detect the return type without running the initializer. The purpose of knowing *earlier* is to opt the user into the promise-based flow (the whole thing would need to be wrapped in a promise). Without any `async` declarations, existing apps would be able to retain the exact same behavior as they have today.

mehulkar:
> I'd be more of a fan of deprecating intializers as a whole. I don't think they're needed.

Few thoughts:

- The `app/initializers` and `app/instance-initilalizers` directories can definitely go away IMO. #573 is my proposal for that.
- Whether or not "code run early in the application lifecycle" is unnecessary is debatable. I've heard the argument for "just put everything in the `ApplicationRoute#beforeModel`" argument, and I'm not sure how I feel about it. To some extent, my opinion is "who cares what you call it, as long as it runs at the right time", but I also feel like it *is* useful to indicate that some code needs to run at app initialization because it has nothing to do any of the routing. I find that stuffing code into `ApplicationRoute#beforeModel` is similar to having a top-level wrapper component in React that does a bunch of stuff, only because there's no other place to do it.

In any case, I think removing initializers completely is a bit out of scope here and would require a lot more discussion and feedback from the community. This seems like a pretty easy win that would put us on the path to a fully async app boot, and then we can fiddle around with where code lives.

mehulkar:
> Could we add an afterAsync key or something like that so that one initializer could wait for another one to resolve?

I think the existing `before`/`after` keys would still be expected to work. Once you add *one* async initializer, you're forced to wrap *all* initializers in a promise. And since the ordering is done before the initializers run anyway, these existing keys would work. Let me know if I should add something to the text of the RFC to make that more clear. 

As a side note, I think the better solution to ordering is the approach in #573, rather than continuing down the before/after path.

text/0000-promise-based-initializers.md
mehulkar: redirecting is just a fancy word for saying " cancel the current transition and start a new one". You *can* intercept a transition in the `ApplicationRoute` that would have ended at `/foo` and and redirect to `/bar`, for example.

text/0000-promise-based-initializers.md
```diff
@@ -0,0 +1,86 @@
...
+For some use cases, this is not early enough, and it is useful to be able to run asynchronous code
+before the application begins routing.
+
+For example, when booting an Ember application in an embedded environment, it may be necessary to
+listen for events from the environment to determine the initial state to determine the initial URL.
+
+Another example may be to interact with 3rd party libraries that need to be initialized early but
+have an asynchronous API.
+
+## Detailed Design
```
rwjblue: Some important details (to me at least 😸) are not mentioned here:

* What happens if you say you are `after` an intializer that is async, are you called after its promise resolves? The RFC should explicitly state this behavior.
* What is the benefit of adding `async: true` (vs doing something like promise-map-series)?

text/0000-promise-based-initializers.md
rwjblue: Even accepting this case as gospel, I don't see why the changes proposed in this RFC would be needed. The specific case seems to be that you want to avoid autobooting your application (until these events have come back), that is already a supported operation (without changes AFAICT).

This feels to me to be introducing a new feature to Ember for what seems like a very very specific/targeted scenario for a single application (that already has argubly better solutions).

text/0000-promise-based-initializers.md
```diff
@@ -0,0 +1,86 @@
...
+The `Application` and `ApplicationInstance` (extending from `Engine` and `EngineInstance`
+classes respectively) are responsible for running initializers and instance initializers. Each
+implements a method that loops through the loaded initializers and executes the exported function.
+
+Initializer and Instance Initializer files each export an object with an `initialize` property
+that references a function. For backwards compatibility, an `async` property should be added, and
+should default to `false` if omitted. If true (or truthy), the Application/ApplicationInstance would
+expect a promise to be returned from the function and wait for the promise to resolve
+before running the next initializer in the queue. The value resolved from the promise would be
+ignored. Promises that reject would throw an exception in development and be caught in production
```
rwjblue: > and be caught in production

What does this mean? If initialization of an application fails, I cannot reason about how to make that app functional again...

text/0000-promise-based-initializers.md
```diff
@@ -0,0 +1,86 @@
...
+};
+```
+
+For maximum backwards compatibility, the boot sequence that calls these initializers could implement
+an early check to see if any async initializers are defined and forgo all async handling.
+
+If implemented, future RFCs could:
+
+- change the default value of the `async` property to `true`.
+- deprecate `deferReadiness` and `advanceReadiness` entirely.
```
rwjblue: 👍👍 - Seems like a great thing to do when we deprecate initializers 😃

text/0000-promise-based-initializers.md
mehulkar: skipping autoboot is very much NOT the happy path, regardless of what the intention of the `visit` API was. The reason I introduced this API is because initializers are one of the few non-async code paths in Ember APIs today and it seems like a good thing to have them regardless of the use cases I've listed here.

We do have other use cases also:

1. Making a network call that sets some cookies before we do anything else
1. Initializing 3rd libraries that have async APIs (or make their own async calls)

This stuff *can* happen in the route, but I'd argue that not all of them *should* happen in a route, if we have the concept of an initializer. If initializers weren't a thing at all, I'd focus more on the manual boot scenario and improving that.

text/0000-promise-based-initializers.md
mehulkar: - yes
- `async: true` was dreamed up for opt-in beahvior and to make it easier to sell :D. I don't know what promise-map-series is, but it sounds like wrapping each initializer in a promise and executing them in order? i'm also fine with that, but I was worried that could have performance implications or could make it possible for some other framework code microtasks to run in between initializers. If neither of those are concerns, I have no problems with it.

text/0000-promise-based-initializers.md
mehulkar: it's up to do the developer to determine that. For example, we make a network request at the beginning to set some cookies. Without those cookies the app is perfectly functional, but could have unintended consequences later. Catching errors in production would just be a failsafe for apps who opts into this feature, but doesn't think through that. I'd be fine with removing this part.

text/0000-promise-based-initializers.md
mehulkar: 👍. It's up to core team what order to do all this in. This RFC seemed like a good path to "just use JS" that didn't require changing the initializer paradigm!

mehulkar:
@chancancode do you have any thoughts here (considering you did the seminal work for async boot in https://github.com/emberjs/ember.js/commit/1a423865c464847539a7f9535a49a2983d626fb4)?

mehulkar:
After talking with @rwjblue and @pzuraq, I think it makes more sense to pursue #573 instead of this. Moving away from initializers will be better for debugability, more explicit code, and some bundle size improvements. This RFC could probably move to `FCP to close`!

mehulkar:
Closing this down!

PR #573, opened 2020-01-10
App Boot Hooks

mehulkar:
[rendered](https://github.com/mehulkar/rfcs/blob/app-boot-hooks/text/0000-app-boot-hooks.md)

### DIFF
--- a/text/0573-app-boot-hooks.md
+++ b/text/0573-app-boot-hooks.md
@@ -89,6 +89,11 @@ Good reasons for separate files per initializer are:
 but it would require re-implementing a large part of the interaction between Application and
 ApplicationInstance, it is easily possible for an application to dig itself into a maintenance
 nightmare.
+- Make initializers more robust and allow them to return promises. This was detailed in [RFC #572],
+but the cost of initializers (both in bundle size and maintenance) is already high and it would be
+better to drop them entirely. One benefit of initializers that would we would lose is the ability
+for addons to inject initialization behavior into their host apps implicitly. While this is useful,
+it is also harder to debug this pattern and ultimately makes it harder to maintain Ember apps.
 
 ## Unresolved questions
 
--- a/text/0573-app-boot-hooks.md
+++ b/text/0573-app-boot-hooks.md
@@ -33,28 +33,27 @@ be progressively enhanced.
 
 ## Motivation
 
-- The existing convention of initializer code comes from Ruby on Rails (which has similar
-`config/initializers/*.rb` files). This technique works and can be nice for organizing code, but it
-is a poor onboarding experience and misses out on a teaching opportunity about how Applications work.
-By making it possible to write code as simple as `console.log()` in `app.js`, Ember developers can
-be exposed to the entry point of the application much sooner and avoid the complication of additional
-directories.
-- Currently the order of initializers is not guaranteed without some extra annotations
-(`before`/ `after`) that are both error-prone and not obvious. It can be debated whether
-two separate initializers *should* care about whether the other one has run or not, it is reasonable
-to expect that two unrelated pieces of code will run in the same order every time. This may already
-be technically true due to the implemented sorting algorithm, it is not guaranteed by Ember, and
-can be messy to grapple with. By using the hooks proposed in the Summary section, Ember developers
-can simply write Javascript and expect that it runs the way they intend it to.
+- **Moving away from initializers**. The existing convention of initializer code comes from
+Ruby on Rails (which has similar `config/initializers/*.rb` files). This technique works
+and can be nice for organizing code, but it also has several downsides:
+
+    - bookkeeping code to determine the order of execution
+    - confusing `before`/`after` API to change order of execution
+    - implicit injections from addons by way of merging directories
+    - lack of support for blocking asynchronous behavior for instance-initializers (`{defer|advance}Readiness` exists for initializers)
+    - a poor onboarding experience and misses out on a teaching opportunity about how Ember Applications work.
+
+    By providing a way to hook into the existing, async boot sequence more directly,
+    application developers can write declarative, explicit code about how they want to
+    customize their application.
 - **Blocking async code before routing**. Currently, the earliest place an application can write
 asynchronous blocking code is in the `beforeModel` hook of the `ApplicationRoute`.
 In some cases, this is too late. For example, if an application needs to wait for an event
 from its host environment e.g. an iframe or an electron shell, and use it to determine
 the starting URL, it cannot do so.
 - **First Transition**: Similarly, if one wants to add an event listener to the
-`routeWillChange` event,
-it is too late to do that in the beforeModel hook, because the event has already
-been dispatched for the first route transition.
+`routeWillChange` event, it is too late to do that in the beforeModel hook, because the
+event has already been dispatched for the first route transition.
 
 ## Detailed design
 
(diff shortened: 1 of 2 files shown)


gossi:
Sounds like something I wrote about earlier, read here: https://gos.si/blog/thoughts-on-improving-developer-ergonomics-for-emberjs/

Since stages of app boot are unclear to me, I was wondering whose those are and at which your methods would be called. One regarding the initializing process (and having control over the sequence here), second is also what currently goes into `ApplicationRoute.beforeModel()` hook. Stuff like loading `ember-intl` and setting it's language. Would love to see more details on this.

Second, I'm strongly against `on*` methods. It is `application.setup()` rather than `application.onBoot()`. The first not only tells a better story, the method is an _active_ part not a reactive one.

mehulkar:
> Second, I'm strongly against on* methods. It is application.setup() rather than application.onBoot(). The first not only tells a better story, the method is an active part not a reactive one.

I'm ok with a different name, but the part that we're trying to hook into is [called `boot()` already](https://github.com/emberjs/ember.js/blob/v3.15.0/packages/@ember/engine/instance.js#L74-L117). I was going for a name that said "this is related to `boot`, but we're not guaranteeing *exactly* where in `boot` it runs"

> second is also what currently goes into ApplicationRoute.beforeModel() hook. 

💯. This is one of the ergonomic problems I'd like to solve, by way of this RFC and #572. Currently, the deciding factor usually ends up being "is this code sync or async", rather than "does this belong in a Route or in app initialization". If #572 is accepted and we can align the feature set between the two places, I think we'll find ourselves in a better position to really answer this question. ApplicationRoute `beforeModel` is currently a shoehorn for every promise resolution that needs to happen early in the app, and I think we can improve that.

gossi:
> I'm ok with a different name, but the part that we're trying to hook into is [called `boot()` already](https://github.com/emberjs/ember.js/blob/v3.15.0/packages/@ember/engine/instance.js#L74-L117) [](/emberjs/ember.js/blob/HEAD@{2020-01-10T19:26:33Z}/packages/@ember/engine/instance.js#L74-L117). I was going for a name that said "this is related to `boot`, but we're not guaranteeing _exactly_ where in `boot` it runs"

Here are a few: `preBoot()`, `postBoot()` rsp `beforeBoot()`, `afterBoot()` =)

mehulkar:
> Here are a few: preBoot(), postBoot() rsp beforeBoot(), afterBoot()

I'm open to this, but we'd have to first define "boot". If we simply called `postBoot` at the end of the existing `boot` method, we might introduce a problem because initializers would run after the `_booted` flag is set.

Also, the major reason for these hooks (for me anyway), is to have a place to put initializers. Since the "post/after" version of the hook is the thing that matters, I'm not sure if there is enough justification to have a "pre" version (there might be some use cases for non autobooted apps though) 🤔.

mehulkar:
> but I do not want to land a change to how initializers work only to subsequently deprecate them

1. This wouldn't change how they work! It would just mean exposing the point at which they run so that uses can replace them if they want. If anything, I think it'll make it easier to deprecate.
1. Deprecating initializers would mean that they are replaced by something right? Is manual boot the correct replacement then? Or do you not anticipate any need to yield to the app after creating an Application/ApplicationInstance, but before starting routing?

mehulkar:
@rwjblue any additional thoughts about this one?

text/0573-app-boot-hooks.md
```diff
@@ -0,0 +1,114 @@
...
+- Relevant Team(s): Ember.js
+- RFC PR: https://github.com/emberjs/rfcs/pull/573
+- Tracking:
+
+# App Boot Hooks
+
+## Summary
+
+This RFC proposes adding two hooks in the `Application` class in `app.js` that, by default, load
+and run initializers and instance-initializers. Here's an eaxmple of what this could look like:
```
SolPier, line 11: typo :+1

mehulkar: @solpier what's the typo?

pangratz, line 11:
```suggestion
and run initializers and instance-initializers. Here's an example of what this could look like:
```

SolPier: `eaxmple`

xg-wang:
Can you add more usecases / examples to explain how one would use the new APIs?

And if I don't follow the default:

```js
import { loadInitializers, loadInstanceInitializers } from 'ember-load-initializers';
import Application from '@ember/application';
import config from './config/environment';
export default class App extends Application {
    onBoot() {
        loadInitializers(this, config.modulePrefix);
        this.runInitializers();
    }
    onInstanceBoot(appInstance) {
        loadInstanceInitializers(this, config.modulePrefix)
        this.runInstanceInitializers();
    }
}
```

Will any of the addon provided initializer not run? In that case how would an addon author communicate with the app consumer?

text/0573-app-boot-hooks.md
```diff
@@ -0,0 +1,114 @@
...
+
+    By providing a way to hook into the existing, async boot sequence more directly,
+    application developers can write declarative, explicit code about how they want to
+    customize their application.
+- **Blocking async code before routing**. Currently, the earliest place an application can write
+asynchronous blocking code is in the `beforeModel` hook of the `ApplicationRoute`.
+In some cases, this is too late. For example, if an application needs to wait for an event
+from its host environment e.g. an iframe or an electron shell, and use it to determine
+the starting URL, it cannot do so.
+- **First Transition**: Similarly, if one wants to add an event listener to the
```
xg-wang, line 54: There's discussion first transition should be handled by an instance-initializer
https://github.com/emberjs/rfcs/pull/680#issuecomment-747739416

Will there be any recommendation change after this RFC?

text/0573-app-boot-hooks.md
```diff
@@ -0,0 +1,114 @@
...
+    - bookkeeping code to determine the order of execution
+    - confusing `before`/`after` API to change order of execution
+    - implicit injections from addons by way of merging directories
+    - lack of support for blocking asynchronous behavior for instance-initializers (`{defer|advance}Readiness` exists for initializers)
+    - a poor onboarding experience and misses out on a teaching opportunity about how Ember Applications work.
+
+    By providing a way to hook into the existing, async boot sequence more directly,
+    application developers can write declarative, explicit code about how they want to
+    customize their application.
+- **Blocking async code before routing**. Currently, the earliest place an application can write
```
xg-wang, line 49: Is this currently achieved by `deferReadiness`? Can you describe what would the new alternative look like?

mehulkar: deferReadiness is not available for instance initializers and it's an awkward API to begin with when we could just use promises.

text/0573-app-boot-hooks.md
mehulkar: It can be handled by a function that runs in `boot()` instead. Note that this RFC doesn't fundamentally change the fact that you can run code during application boot. It just changes the developer ergonomics to enable new features.

mehulkar:
> Will any of the addon provided initializer not run? In that case how would an addon author communicate with the app consumer?

`import { loadInitializers, loadInstanceInitializers } from 'ember-load-initializers';`

This would do the same thing that currently happens, which involves gathering up all the files in the app and addons initializers/instance-initializers directories.

text/0573-app-boot-hooks.md
mehulkar: ah thanks, fixed!

wagenet:
What's the status of this? Are we just waiting on more review?

mehulkar:
I still believe this is a good idea and the right abstraction for an ember application. I am no longer working primarily in Ember though, so I’m not sure how I’ll be able to keep contributing. If any core team feedback appears, I would be happy to address it though.

PR #576, opened 2020-01-10
Move HTTP Mock to addon

mehulkar:
[rendered](https://github.com/mehulkar/rfcs/blob/http-mock/text/0000-http-mock.md)


wagenet:
This seems reasonable to me. I'll see what I can do to get it moving.

bertdeblock:
I've added this RFC to next week's agenda.

runspired:
Moving to an addon seems great if doable; however, it would be better if we exposed a serverMiddleware hook directly to ember apps for this purpose. The dance otherwise is very complicated and not forward-thinking as we move away from older addon patterns. I'd like to see http-mocks be the default story, mirage is .... not an ideal solution for anyone.

wagenet:
@runspired could you expand on this a bit. I'm guessing you'd suggest an ember-cli RFC for this change?

runspired:
@wagenet currently to implement anything as a server-middleware it must be created as an addon. EmberCLI has no ability currently for sever middleware to be added except that mechanism. This makes creating solutions for apps that hook into the serve process very complicated: which includes any proposed http-mock replacement. Instead of mocks in-app you end up with mocks in an addon in the app.

runspired:
Basically my suggestion is for embercli to add registerMiddleware to the EmberApp in ember-cli-build

kategengler:
We think the medium-term plan should be to deprecate HTTP mock, rather than extract it. To that end, we are putting this in FCP to close.

PR #577, opened 2020-01-10
Tracked Maps and Sets

pzuraq:
[Rendered](https://github.com/emberjs/rfcs/blob/tracked-maps-and-sets/text/0577-tracked-maps-and-sets.md)

### DIFF
--- a/text/0000-tracked-maps-and-sets.md
+++ b/text/0000-tracked-maps-and-sets.md
@@ -1,0 +1,265 @@
+- Start Date: 2020-01-02
+- Relevant Team(s): Ember.js
+- RFC PR: (after opening the RFC PR, update this with a link to it and update the file name)
+- Tracking: (leave this empty)
+
+# Tracked Maps and Sets
+
+## Summary
+
+Adds tracked versions of JavaScript's built-in `Map`, `WeakMap`, `Set`, and
+`WeakSet` classes, and updates the `{{get}}` helper to work with `Map` and
+`WeakMap`.
+
+```js
+class Store {
+  inventory = new TrackedMap([
+    ['socks', 123],
+    ['shoes', 456],
+  ]);
+
+  employees = new TrackedSet([
+    'Ed',
+    'Katie',
+    'Leah',
+  ]);
+}
+```
+
+## Motivation
+
+Autotracking has overall been a huge success in Ember Octane, and has received
+a lot of good feedback in general. However, there are a few use cases that are
+not supported very well currently. One of those use cases is tracking changes to
+_dynamic collections of values_, where its not possible to decorate properties
+with the `@tracked` decorator because the properties themselves are constantly
+changing.
+
+There are four types of primitive/built-in collections in JavaScript:
+
+1. Plain-Old-JavaScript-Objects (POJOs), which can operate as simple key-value
+   stores, but have many other unrelated use cases
+2. Maps, which are explicit key-value stores (and somewhat more flexible than
+   POJOs)
+3. Arrays, for storing lists of items
+4. Sets, which are effectively unique arrays, and are usually used specifically
+   to define groups and check whether a value exists within said group
+
+These collections are used frequently for a variety of purposes in standard JS,
+and each was introduced to solve use cases that the others could not. In
+particular, `WeakMap` and `WeakSet` are very valuable for storing private meta
+information about an object while still allowing it to be garbage collected, and
+this ability is used very commonly in library and framework code.
+
+Based on early feedback, it has become apparent that providing tracked versions
+of each of these collections is important. In particular, a tracked version of a
+key-value store is something that many developers have been wanting. Meanwhile,
+the [`tracked-built-ins`](https://github.com/pzuraq/tracked-built-ins) library
+has shown that tracked versions of `WeakMap` and `WeakSet` can be built, and are
+very useful for developing various tracking utilities (such as the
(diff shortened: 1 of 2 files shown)


Gaurav0:
I'd prefer this to be in a separate package rather than part of `@glimmer/tracking` so that we can tree shake it.

NullVoxPopuli:
> I'd prefer this to be in a separate package rather than part of @glimmer/tracking so that we can tree shake it.

tree shaking can work on multiple imports of the same package

Gaurav0:
> > I'd prefer this to be in a separate package rather than part of @glimmer/tracking so that we can tree shake it.
> 
> tree shaking can work on multiple imports of the same package

I am unaware of how this works. Right now we can only do "manual tree shaking" in ember-cli; i.e. not load a package. While Embroider may change this someday, we must work with what we have now. Once Ember core is property separated into *real* separate packages, with today's ember-cli we will be able to only load or not load the entire package.

pzuraq:
This would not be implemented as a separate package either way, so manual treeshaking would not be possible. We should build for automatic treeshaking.

Gaurav0:
Sorry, why not? It is in a separate addon now.

pzuraq:
`@glimmer/tracking` is not a separate addon. You can install the addon for types, but it is still provided by `ember-source`.

Gaurav0:
Yes. I would like this to be separate from `@glimmer/tracking`

NullVoxPopuli:
@Gaurav0 I think we have too many imports as is.

Gaurav0:
@NullVoxPopuli Why? The imports went through the RFC process and were agreed to by the community.

NullVoxPopuli:
@Gaurav0 I just mean that I feel like we have too many imports as is -- not that there is anything we can do about it -- all the imports make sense. 
Adding more tracked utilities should be on `@glimmer/tracking`, as the RFC states, because they are very related. Splitting that up doesn't make sense to me.  
I understand the desire to have tree shaking, but we need to design for the future, and not hold ourselves back with existing tech... idk if that makes sense. I need to eat. lol

pzuraq:
@Gaurav0 for pragmatic purposes, this would not be implemented as a separate addon for the time being, which is my overall point.

Gaurav0:
> Adding more tracked utilities should be on @glimmer/tracking, as the RFC states, because they are very related. Splitting that up doesn't make sense to me.

IMHO, I disagree. The `tracked` function and decorator is likely to be much more widely used than any additional utilities. This is similar to how `computed` is on `@ember/object` and macros are on `@ember/object/computed`.

The only additional thing I think should be added to this import

sandstrom:
Good suggestion, I really like this idea.

**However, I think designing something proxy-based would be much better and [browser support is excellent](https://caniuse.com/#search=proxy). Let's come up with something modern, with longer shelf-life.**

IE 11 is in rapid decline and only have a few percentage points of usage these days.

Many large software companies has already stopped supporting it, for example:

- [Atlassian has dropped support for IE 11](https://blog.developer.atlassian.com/retiring-ie11-support-for-atlassian-cloud-server-and-data-center-products/)

- [Zendesk has dropped IE 11 support](https://support.zendesk.com/hc/en-us/articles/360002084367-Removal-of-support-for-Internet-Explorer-11)

- [Qlik View let go of IE 11 too](https://support.qlik.com/articles/000079718)

- We're not a huge company, but we dropped support for IE11 and none of our ~1000 customers complained. When we checked usage logs, for the few companies that still had IE11 usage 100% of them had other browsers installed on their machines already, and somewhere around 50-90% of their employees already used another browser.

The days where IE11 was the only browser installed on a computer are long gone.

Ember should drop IE 11 support and focus time on using the modern JS toolset to provide a nice framework. Any Ember user that need to extend support a while longer, e.g. end of 2020, could simply linger on an Ember LTS a while longer. It doesn't need to hold back new framework development.

*Regarding the performance drawback with proxies, I don't think it would be a big problem under most real-world scenarios.*

pzuraq:
@sandstrom these classes do not require a proxy to wrap transparently, since all access is done through methods, unlike arrays. Ultimately, the exact implementation should likely be whatever is most performant, and whether or not that is a subclass or a proxy should be an implementation detail.

runspired:
Would these tracked classes be iterable by {{#each ...}} in a template or would that require wrapping in a Proxy (and dropping IE11 support)?

If yes, can we expose the underlying mechanism so that libraries like EmberData can provide custom iterables without the Proxy wrapper requirement?

pzuraq:
> Would these tracked classes be iterable by {{#each ...}} in a template or would that require wrapping in a Proxy (and dropping IE11 support)?

A proxy is not necessary for this either way, and they should be iterable as is.

> If yes, can we expose the underlying mechanism so that libraries like EmberData can provide custom iterables without the Proxy wrapper requirement?

Yes, but I believe that should be a separate RFC specifically dedicated to outlining our support of native iterables. I also think that, because native iterables rely on Symbol, we will still need to wait for IE11 support to be fully dropped before we can deprecate alternatives to native iterables, but we can introduce support immediately in the meantime.

NullVoxPopuli:
IE11 is now dropped, so what's left for this RFC? 
It'd be great to have more of the tracked-toolbox family of stuff built-in :D

wagenet:
I believe this should be closed with the same rationale as https://github.com/emberjs/rfcs/pull/569#issuecomment-1193163334. If I'm mistaken, please let me know and we can work to get this moving again.

PR #580, opened 2020-01-11
Destroyables

pzuraq:
[Rendered](https://github.com/emberjs/rfcs/blob/destroyables/text/0580-destroyables.md)


text/0580-destroyables.md
```diff
@@ -0,0 +1,163 @@
...
+* Modifiers
+
+Any future classes that are added and have a container managed lifecycle should
+also be marked as destroyables.
+
+## How we teach this
+
+This is a generally low level API, meant for usage in the framework and in
+addons. It should be taught primarily through API documentation, and potentially
+through and in-depth guide.
```
jrjohnson: typo: should be `an in-depth`.

text/0580-destroyables.md
```diff
@@ -0,0 +1,163 @@
...
+}
+```
+
+## Motivation
+
+Ember manages the lifecycles and lifetimes of many built in constructs, such as
+components, and does so in a hierarchical way - when a parent component is
+destroyed, all of its children are destroyed as well. This is a well established
+software pattern that is useful for many applications, and there are a variety
+of libraries, such as `ember-lifeline` and `ember-concurrency`, that would
```
rwjblue:
```suggestion
of libraries, such as [ember-lifeline](https://github.com/ember-lifeline/ember-lifeline) and [ember-concurrency](https://github.com/machty/ember-concurrency), that would
```

text/0580-destroyables.md
```diff
@@ -0,0 +1,163 @@
...
+
+The API consists of 3 functions:
+
+```ts
+declare function registerLifetime(parent: object, child: object): void;
+declare function registerDestructor(destroyable: object, destructor: () => void): void;
+declare function destroy(destroyable: object): void;
+```
+
+### `registerLifetime`
```
rwjblue: What do you think about changing this slightly to infer that we are registering a lifetime _limit_ (not a "lifetime" specifically)?

text/0580-destroyables.md
```diff
@@ -0,0 +1,163 @@
...
+
+### `registerDestructor`
+
+Receives a destroyable and a destructor function, and associates the function
+with it. When the destroyable is destroyed, or when its parent is destroyed, the
+destructor function will be called. Multiple destructors can be associated with
+a given destroyable, and they can be associated over time, allowing libraries
+like `ember-lifeline` to dynamically add destructors as needed.
+
+### `destroy`
```
rwjblue: Can you add prose to explain that calling `destroy(obj); destroy(obj);` is "safe" (e.g. calling destroy a second time is essentially inert

text/0580-destroyables.md
```diff
@@ -0,0 +1,163 @@
...
+
+`destroy` manually initiates the destruction of a destroyable. This allows users
+to manually manage the lifecycles of destroyables that are shorter lived than
+their parents (for instance, destroyables that are defined an a service, which
+lives as long as an application does), cleaning them up as needed.
+
+### Built In Destroyables
+
+The root destroyable of an Ember application will be the instance of the
+application itself. All framework managed classes are destroyables, including:
```
rwjblue: > will be the instance of the application itself

technically it is the instance of Ember.ApplicationInstance (not an "instance of the application itself")

Maybe you could rephrase to "owner"?

text/0580-destroyables.md
rwjblue: We need to either assert if calling `registerLifetime(thing1, child); registerLifetime(thing2, child);` or define the behavior for that scenario.

text/0580-destroyables.md
```diff
@@ -0,0 +1,163 @@
...
+   there is no way for `ember-concurrency` to know how to destroy tasks on
+   non-framework classes that users may have added themselves.
+
+This RFC proposes a streamlined API that disconnects the exact implementation
+from any interface, allows for multiple destructors per-destroyable, and
+maximizes interoperability in general.
+
+## Detailed design
+
+The API consists of 3 functions:
```
rwjblue: F2F feedback (from @ef4): Can you also add an `unregisterDestructor` method?

text/0580-destroyables.md
```diff
@@ -0,0 +1,163 @@
...
+  hooks. Since this is a low-level API, it shouldn't be too problematic - most
+  users will be guided toward using the standard lifecycle hooks, and this API
+  will exist for libraries like `ember-concurrency` and `ember-lifeline`.
+
+## Alternatives
+
+- Continue using existing lifecycle hooks for public API, and don't provide an
+  independent API.
+
+## Unresolved questions
```
rwjblue: Can you update the prose to add an import location, or add it to unresolved questions?

rwjblue:
@pzuraq - Another feature that ember-lifeline has, that is pretty nice (and we would either have to drop there or not be able to migrate to this builtin API) is that we can assert that all destroyables that were ever registered **have** been ran. This is used in testing infrastructure (see [the documentation here](https://ember-lifeline.github.io/ember-lifeline/docs/testing)) to ensure that folks don't _accidentally_ forget to properly `destroy(obj)` something that they should have (the assumption here is that at the end of the test, all destroyables that were created during the test are ran).

text/0580-destroyables.md
```diff
@@ -0,0 +1,243 @@
...
+maximizes interoperability in general.
+
+## Detailed design
+
+The API consists of 6 main functions, imported from `@ember/application`:
+
+```ts
+declare function associateDestroyableChild(parent: object, child: object): void;
+declare function registerDestructor(destroyable: object, destructor: () => void): void;
+declare function unregisterDestructer(destroyable: object, destructor: () => void): void;
```
btecu: `unregisterDestructor`

text/0580-destroyables.md
```diff
@@ -0,0 +1,243 @@
...
+   there is no way for `ember-concurrency` to know how to destroy tasks on
+   non-framework classes that users may have added themselves.
+
+This RFC proposes a streamlined API that disconnects the exact implementation
+from any interface, allows for multiple destructors per-destroyable, and
+maximizes interoperability in general.
+
+## Detailed design
+
+The API consists of 6 main functions, imported from `@ember/application`:
```
rwjblue: FWIW, I am not sure that `@ember/application` is quite the correct layer here. A few points:

* I think it would be pretty rare to actually entangle something with the destruction of the app (the more common thing would be to associate with a service, component, route, controller, etc)
* We need to be able to use this concept in Glimmer.js applications, and having it come from `@ember/application` means that we will have a different module import in Glimmer.js than in Ember (increasing the difference in source <-> source compat between Ember and Glimmer.js). Also, it seems likely that the actual implementation of these destroyable APIs internally will come from a glimmer-vm abstraction.

text/0580-destroyables.md
```diff
@@ -0,0 +1,243 @@
...
+
+#### `associateDestroyableChild`
+
+This function is used to associate a destroyable object with a parent. When the
+parent is destroyed, all registered children will also be destroyed.
+
+- Attempting to associate a parent or child that has already been destroyed will
+  throw an error.
+- Attempting to associate a child with a parent that is not a destroyable should
+  throw an error.
```
rwjblue: I don't understand what "is not a destroyable" is here. Can you be more specific?

text/0580-destroyables.md
```diff
@@ -0,0 +1,243 @@
...
+
+Receives a destroyable and a destructor function, and associates the function
+with it. When the destroyable is destroyed, or when its parent is destroyed, the
+destructor function will be called. Multiple destructors can be associated with
+a given destroyable, and they can be associated over time, allowing libraries
+like `ember-lifeline` to dynamically add destructors as needed.
+
+- Registering a destructor on a destroyed object should throw an error.
+- Attempting to register the same destructor multiple times should throw an
+  error.
```
rwjblue: This seems slightly odd. Since we aren't passing any values in to the destructor function, every destructor that is registered will have to leverage closure scope along with an anonymous  function (AFAICT).

text/0580-destroyables.md
```diff
@@ -0,0 +1,243 @@
...
+- Attempting to register a destructor on a non-destroyable should throw an
```
rwjblue: I'd like a more concrete definition of "destroyable" (similar to my other comment above)

text/0580-destroyables.md
```diff
@@ -0,0 +1,243 @@
...
+
+This function asserts that all active destroyables have been destroyed at the
+time it is called. It is meant to be a low level hook that testing frameworks
+like `ember-qunit` and `ember-mocha` can use to hook into and validate that all
+destroyables have in fact been destroyed.
+
+### Timing Semantics
+
+Destroyables first call their own destructors, then destroy their associated
+children.
```
rwjblue: I'd like a bit more flesh here, and I think we need some tweaks. Specifically, I think that the order of operation should be:

1. mark the destroyable such that `isDestroying(destroyable)` returns `true`
2. call the destructors
3. mark _all_ children associated such taht `isDestroying(child)` returns `true`
4. call the destructors for the child
5. recurse.....

text/0580-destroyables.md
```diff
@@ -0,0 +1,243 @@
...
+
+### Built In Destroyables
```
rwjblue: I'd like to rationalize `Ember.Object`s current `obj.destroy()`, `obj.isDestroying`, `obj.isDestroyed`, `obj.willDestroy` in terms of this RFC. Limiting destroyables to just the roots listed makes it quite difficult to "build higher" with your own classes that aren't derived from these Ember built-ins.

text/0580-destroyables.md
pzuraq: I don't think it is if you take into account `unregisterDestructor`. Users may end up storing off closures or actions, and could accidentally register one twice.

text/0580-destroyables.md
```diff
@@ -164,45 +166,42 @@ destructor from the destroyable.
...
-lives as long as an application does), cleaning them up as needed.
+`destroy` initiates the destruction of an destroyable object. It runs all
+associated destructors, and then destroys all children recursively.
 
-Calling `destroy` multiple times on the same object is safe. It will not throw
-an error, and will not take any further action.
+Destruction via `destroy()` follows these steps:
 
-- Calling `destroy` on a non-destroyable should throw an error.
+1. Mark the destroyable such that `isDestroying(destroyable)` returns `true`
```
rwjblue: Does it also mark each of the children?

text/0580-destroyables.md
pzuraq: I don't think we're ready to move it to `@glimmer` just yet, especially because the ideal place would likely be `@glimmer/core`, which is still an evolving package. I think we should wait until we have an RFC for source-to-source compatibility to resolve this. Updated the RFC to `@ember/destroyable` for now, and we can move the import once we have a more thorough plan for Glimmer imports.

text/0580-destroyables.md
```diff
@@ -164,45 +166,42 @@ destructor from the destroyable.
...
-This function asserts that all active destroyables have been destroyed at the
-time it is called. It is meant to be a low level hook that testing frameworks
-like `ember-qunit` and `ember-mocha` can use to hook into and validate that all
-destroyables have in fact been destroyed.
-
-### Timing Semantics
-
-Destroyables first call their own destructors, then destroy their associated
-children.
+This function asserts that all objects which has associated destructors or
```
rwjblue:
```suggestion
This function asserts that all objects which have associated destructors or
```

text/0580-destroyables.md
pzuraq: It does while calling `destroy()` on the children, since it recurses. It does not mark them as destroying before running its own destructors.

text/0580-destroyables.md
```diff
@@ -175,9 +175,14 @@ associated destructors, and then destroys all children recursively.
 Destruction via `destroy()` follows these steps:
 
 1. Mark the destroyable such that `isDestroying(destroyable)` returns `true`
-2. Call the destroyable's destructors
-3. Call `destroy()` on each of the destroyable's associated children
-4. Mark the destroyable such that `isDestroyed(destroyable)` returns `true`
+2. Schedule calling the destroyable's destructors
+4. Call `destroy()` on each of the destroyable's associated children
+3. Schedule setting destroyable such that `isDestroyed(destroyable)` returns `true`
```
rwjblue:
```suggestion
3. Call `destroy()` on each of the destroyable's associated children
4. Schedule setting destroyable such that `isDestroyed(destroyable)` returns `true`
```

text/0580-destroyables.md
```diff
@@ -175,9 +175,14 @@ associated destructors, and then destroys all children recursively.
...
+
+This algorithm results in the entire tree of destroyables being first marked as
+destroyed, then having all of their destructors called, and finally all being
```
rwjblue:
```suggestion
destroying, then having all of their destructors called, and finally all being
```

text/0580-destroyables.md
```diff
@@ -0,0 +1,247 @@
...
+  will exist for libraries like `ember-concurrency` and `ember-lifeline`.
+
+## Alternatives
+
+- Continue using existing lifecycle hooks for public API, and don't provide an
+  independent API.
+
+## Unresolved questions
+
+- Should `willDestroy` and `didDestroy` (e.g. mark-and-sweep style semantics) be
```
MichalBryxi: Just throwing a thought: Will this be a footgun? Seems like first thing I would try if `willDestroy` get deprecated and codemods are not working for me (for whatever reason):
```js
registerDestructor(this, this.willDestroy);
```

pzuraq: This specifically was referencing _synchronous_ `willDestroy` semantics, which are a footgun for performance, but are necessary for things like running code before a component is removed from the DOM (e.g. `willDestroyElement`). Generally, you should do the minimal amount of work synchronously, and schedule the rest, but often this doesn't happen. I'll reword this to make that more clear.

pzuraq: FWIW, the current semantics _would_ be exactly the same as `willDestroy` is currently.

text/0580-destroyables.md
```diff
@@ -0,0 +1,247 @@
...
+* Services
+* Routes
+* Controllers
+* Helpers
+* Modifiers
+
+Any future classes that are added and have a container managed lifecycle should
+also be marked as destroyables.
+
+## How we teach this
```
MelSumner: We've said previously that we don't want this section to say how we should theoretically teach this, but rather be an MVP of documentation for the guides. I would ask that this section be updated accordingly.

rwjblue:
After discussing this in the meeting today, we are very excited to move this RFC into final comment period!

ming-codes:
Given how @pzuraq loves decorators, I’m surprised to see there are no mention of decorators in this RFC 😏

BryanCrotaz:
Suggestion. Create a codemod to convert willDestroy to using destroyables instead

text/0580-destroyables.md
```diff
@@ -0,0 +1,385 @@
...
+
+  @action pinModal() {
+    unregisterDestructor(this, this.modalDestructor);
+  }
+}
+```
+
+- Calling `unregisterDestructor` on a destroyed object should throw an error.
+- Calling `unregisterDestructor` with a destructor that is not associated with
+  the object should throw an error.
```
buschtoens, line 246: If calling `unregisterDestructor` with a non-registered destructor throws and error and there is no function to check whether a destructor is associated with a destroyable, it means that users need to implicitly or explicitly bookkeep what destructors were registered with what destroyables.

I like that it throws an error, but I would recommend to expose another function like `isDestructorRegistered(obj, destructor)`.

text/0580-destroyables.md
```diff
@@ -0,0 +1,385 @@
...
+This algorithm results in the entire tree of destroyables being first marked as
+destroying, then having all of their destructors called, and finally all being
+marked as `isDestroyed`. There won't be any in between states where some items
+are marked as `isDestroying` while destroying, while others are not.
+
+Calling `destroy` multiple times on the same destroyable is safe. It will not
+throw an error, and will not take any further action.
+
+Calling `destroy` with a destroyable that has no destructors or associated children
+will not throw an error, and will do nothing.
```
buschtoens, line 281: Will `isDestroying` / `isDestroyed` still return `true` though?

text/0580-destroyables.md
```diff
@@ -0,0 +1,385 @@
...
+
+1. Mark the destroyable such that `isDestroying(destroyable)` returns `true`
+2. Schedule calling the destroyable's destructors
+3. Call `destroy()` on each of the destroyable's associated children
+4. Schedule setting destroyable such that `isDestroyed(destroyable)` returns `true`
+
+This algorithm results in the entire tree of destroyables being first marked as
+destroying, then having all of their destructors called, and finally all being
+marked as `isDestroyed`. There won't be any in between states where some items
+are marked as `isDestroying` while destroying, while others are not.
```
buschtoens, line 275: I overlooked the keyword `Schedule`, meaning that the action happens asynchronously / next tick / etc. and thought that this algorithm actually results in marking the parent, calling the parent destructors already and only _then_ marking the children and calling the children's destructors.

Just adding this here, in case I am not the only one.

Do we wanna make any guarantees with regards to the timing of scheduling? Is it scheduled onto a specific runloop queue, next tick, or no guarantees on that at all?

text/0580-destroyables.md
```diff
@@ -0,0 +1,385 @@
...
+    // obj will also be destroyed, and have all of its destructors triggered.
+    this.obj = associateDestroyableChild(this, {});
+  }
+}
+```
+
+Returns the associated child for convenience.
+
+- Attempting to associate a parent or child that has already been destroyed
+  should throw an error.
```
buschtoens, line 148:
```suggestion
- Attempting to associate a parent or child that has already been destroyed
  or is being destroyed should throw an error.
```

Is it legal to associate a parent or child while either is being destroyed (`isDestroying`)?

text/0580-destroyables.md
```diff
@@ -0,0 +1,385 @@
...
+
+  constructor() {
+    this.resize.register(this, this.layout);
+
+    registerDestructor(this, unregisterResize);
+  }
+}
+```
+
+- Registering a destructor on a destroyed object should throw an error.
```
buschtoens, line 217:
```suggestion
- Registering a destructor on a destroyed object or object that is being destroyed should throw an error.
```

Is it legal to register a destructor on an object while it is being destroyed (`isDestroying`)?

text/0580-destroyables.md
```diff
@@ -0,0 +1,385 @@
...
+    this.modalDestructor = registerDestructor(this, () => this.modals.remove(this));
+  }
+
+  @action pinModal() {
+    unregisterDestructor(this, this.modalDestructor);
+  }
+}
+```
+
+- Calling `unregisterDestructor` on a destroyed object should throw an error.
```
buschtoens, line 244:
```suggestion
- Calling `unregisterDestructor` on a destroyed object or object that is being destroyed should throw an error.
```

Is it legal to unregister a destructor from an object while it is being destroyed (`isDestroying`)?

text/0580-destroyables.md
```diff
@@ -0,0 +1,385 @@
...
+Receives a destroyable, and returns `true` if the destroyable has begun
+destroying. Otherwise returns false.
+
+```js
+let obj = {};
+
+isDestroying(obj); // false
+destroy(obj);
+isDestroying(obj); // true
+```
```
buschtoens, line 294:
```suggestion
```js
let obj = {};

isDestroying(obj); // false
destroy(obj);
isDestroying(obj); // true

// ...sometime later, after scheduled destruction
isDestroyed(obj); // true
isDestroying(obj); // false
```
```

Will `isDestroying` return `true` or `false` _after_ the destruction has finished (`isDestroyed`)?

File: text/0580-destroyables.md [hunk 28]
```diff
@@ -0,0 +1,385 @@
...
+
+Receives a destroyable, and returns `true` if the destroyable has begun
+destroying. Otherwise returns false.
+
+```js
+let obj = {};
+
+isDestroying(obj); // false
+destroy(obj);
+isDestroying(obj); // true
```
buschtoens, line 293:
```suggestion
let obj = {};

isDestroying(obj); // false
destroy(obj);
isDestroying(obj); // true

// ...sometime later, after scheduled destruction
isDestroyed(obj); // true
isDestroying(obj); // false
```

Will `isDestroying` return `true` or `false` _after_ the destroying has finished (`isDestroyed`)?

text/0580-destroyables.md
```diff
@@ -0,0 +1,385 @@
...
+isDestroyed(obj); // true
+```
+
+#### `assertDestroyablesDestroyed`
+
+This function asserts that all objects which have associated destructors or
+associated children have been destroyed at the time it is called. It is meant to
+be a low level hook that testing frameworks like `ember-qunit` and `ember-mocha`
+can use to hook into and validate that all destroyables have in fact been
+destroyed.
```
buschtoens, line 318: It would be helpful to know which objects did not finish destroying in this case. The objets could be exposed as a property on the error.

text/0580-destroyables.md
buschtoens: For `CoreObject` `isDestroying` continues to return `true` _after_ `isDestroyed` became `true`, so I would expect it to be the same here:

```suggestion
let obj = {};
isDestroying(obj); // false
destroy(obj);
isDestroying(obj); // true
// ...sometime later, after scheduled destruction
isDestroyed(obj); // true
isDestroying(obj); // true
```

buschtoens:
I made a polyfill for this RFC: [`ember-destroyable-polyfill`](https://github.com/buschtoens/ember-destroyable-polyfill)

It's not perfect yet: The test suite could be expanded and the integration with `CoreObject` / the hooks-based system could be improved, but AFAICT it works.

text/0580-destroyables.md
buschtoens: In my naive polyfill implementation, this function leaks state / failures between test runs, as there is no way to reset the hidden module state: https://github.com/buschtoens/ember-destroyable-polyfill/issues/2

```ts
test('should fail', function () {
  const obj = {};
  registerDestructor(obj, () => {});
  // run(() => destroy(obj));
  assertDestroyablesDestroyed();
});

test('should pass, but fails', function () {
  const obj = {};
  registerDestructor(obj, () => {});
  run(() => destroy(obj));
  assertDestroyablesDestroyed();
});
```

How are `ember-qunit` / `ember-mocha` supposed to reset the module state in `@ember/destroyable` after each test run? `require.unsee('@ember/destroyable')`? Or is this actually a design flaw?

pzuraq: `assertDestroyablesDestroyed` could probably do the cleanup during test builds. We would have to track some extra state in DEBUG builds, use a normal Map instead of a WeakMap, but that's something we've done before.

text/0580-destroyables.md
pzuraq: Yes, it will.

text/0580-destroyables.md
pzuraq: We do not want to make guarantees about timing, as timing could change per-environment, or could change in the future. The only guarantee is relative timing, as outlined in this RFC.

text/0580-destroyables.md
pzuraq: Yes, `isDestroying` always returns true after being set. It does not get unset at any point in time.

text/0580-destroyables.md
rwjblue: I've commented on the issue linked above (in the polyfill) to explain how this should work. I don't think this is a gap in the public APIs proposed (but the assertion would leverage internal private APIs).

rwjblue:
Thanks for working through this with us!

sandstrom:
@buschtoens Awesome addon! Just installed it, works great! 💯

PR #581, opened 2020-01-14
New Test Waiters

scalvert:
[Rendered](https://github.com/scalvert/rfcs/blob/new-test-waiters/text/0000-new-test-waiters.md)


text/0000-new-test-waiters.md
```diff
@@ -0,0 +1,111 @@
+- Start Date: 2020-01-14
+- RFC PR: (leave this empty)
+- Ember Issue: (leave this empty)
+
+# Summary
+
+Test waiters have been around in Ember in one form or another since version 1.2.0, and provide a way for developers to signal to the testing framework system that async operations are currently active, when to keep waiting, and when those async operations have completed. This allows the active test to wait during the test in a deterministic fashion, and only proceed once the active async is completed,
```
rwjblue:
```suggestion
Test waiters have been around in Ember in one form or another since version 1.2.0, and provide a way for developers to signal to the testing framework system that async operations are currently active, when to keep waiting, and when those async operations have completed. This allows the active test to wait during the test in a deterministic fashion, and only proceed once the active async is completed.
```

text/0000-new-test-waiters.md
```diff
@@ -0,0 +1,111 @@
...
+
+The new system will provide a few benefits:
+
+1. A new API that's _explicit_ and _straightforward_
+1. A more robust way to gather debugging information for the test waiter
+1. Default test waiters with the ability to author your own, more complex test waiters
+
+# Detailed design
+
+Ember’s test framework has an internal concept of settledness, that is used by all of its internal helpers. Settledness can be defined as **_all known active async operations have completed, and there’s no outstanding work to be done_**. This is codified in the [settled](https://github.com/emberjs/ember-test-helpers/blob/master/addon-test-support/@ember/test-helpers/settled.ts#L158) helper.
```
rwjblue: > This is codified in the [settled](https://github.com/emberjs/ember-test-helpers/blob/master/addon-test-support/@ember/test-helpers/settled.ts#L158) helper

Can you tweak this to point to https://github.com/emberjs/ember-test-helpers/blob/master/API.md#settled?

text/0000-new-test-waiters.md
```diff
@@ -0,0 +1,111 @@
...
+
+The settled helper, as noted, wires itself up to known asynchronous behaviors. Those include whether there
+
+- is an active runloop (more on the runloop)
+- are any pending timers within the runloop (run.later, run.debounce, run.throttle)
+- are any pending test waiters (more on waiters later!)
+- are any pending AJAX requests
```
rwjblue:
```suggestion
- are any pending `jQuery.ajax` requests
```

text/0000-new-test-waiters.md
```diff
@@ -0,0 +1,111 @@
...
+
+The current test waiters implementation has a simple but confusing API, and the test waiters themselves lack some key features. This RFC proposes replacing them with a new test waiters system: [ember-test-waiters](https://github.com/rwjblue/ember-test-waiters).
+
+# Motivation
+
+Recently, an updated replacement for the original test waiters API was created. This new library, ember-test-waiters, seeks to provide an easy-to-use API that can be used to interleave unmanaged async behaviors with Ember’s test framework. Using a test waiter can help mark begin and end points for your async operations, allowing the test to deterministically pause during execution.
+
+The new system will provide a few benefits:
+
+1. A new API that's _explicit_ and _straightforward_
```
rwjblue: FWIW, "straightforward" is a bit subjective 😄, I'd suggest a smallish tweak to:

```suggestion
1. A new API that removes the existing footguns (e.g. "Do I return `false` or `true` if I want to continue waiting?!?")
```

text/0000-new-test-waiters.md
```diff
@@ -0,0 +1,111 @@
...
+
+During the development of the test isolation validation feature, we discovered that most asynchronous operations used in the settled check provided good debug information that could be provided to the end user, with the exception of the existing test waiters. Those waiters only provided rudimentary information that could be exposed, specifically whether there were any active test waiters pending, but nothing more.
+
+To address this, a new addon was written to experiment on a new test waiter system that would provide a number of things (as noted above):
+
+1. A new API that's _explicit_ and _straightforward_
+1. A more robust way to gather debugging information for the test waiter
+1. Default test waiters with the ability to author your own, more complex test waiters
+
+This would allow developers to utilize this new test waiters system to annotate their asynchronous operations not tracked by the settled check, and for those annotations to provide useful debugging information in the event their async extended past the expected duration of the test.
```
rwjblue:
```suggestion
This allows developers to utilize `ember-test-waiters` to annotate their asynchronous operations that are not tracked by an `await settled()` check, and for those annotations to provide useful debugging information in the event their async extended past the expected duration of the test.
```

text/0000-new-test-waiters.md
```diff
@@ -0,0 +1,111 @@
...
+import { registerWaiter } from '@ember/test';
+
+registerWaiter(function() {
+  return myPendingTransactions() === 0;
+});
+```
+
+While reading the above is straightforward, when writing a test waiter using the old system it's easy to forget what the expected return value is: `true` or `false`. Additionally, it's a bit more cognitive overhead to derive what the intended result of the particular boolean return value is: does returning `true` result in the test waiter waiting or not?
+
+As mentioned before, there's no additional information provided via `registerWaiter`, and no way to capture stack traces at the call site.
```
rwjblue: Maybe flesh this out a tad bit more. Specifically, I'd like to see a bit of prose that explains that when some async that is being waited on your test is "hung" (and may time out) but it is not possible for you to identify which of the (possibly hundreds!) waiters in your application was "stuck".

text/0000-new-test-waiters.md
```diff
@@ -0,0 +1,111 @@
...
+
+The old test waiters system will be deprecated in its own deprecation RFC.
+
+# How We Teach This
+
+This new test waiters system should be included in the Ember guide's testing section. Information and examples should be provided to allow users to correctly author asynchronous code that can be correctly managed by the testing system.
+
+# Drawbacks
+
+- added work to integrate the new test waiters system with `ember-mocha`
```
rwjblue: IMHO, this isn't a drawback, it's should be part of the detailed design.

text/0000-new-test-waiters.md
```diff
@@ -0,0 +1,111 @@
...
+
+This new test waiters system has been through multiple iterations of refinement, and is in use and integrated with the test isolation validation system. Furthermore, it is backwards compatible with the old test waiters system, allowing applications and addons to gradually migrate to using the new system.
+
+## Deprecating the old test waiters
+
+The old test waiters system will be deprecated in its own deprecation RFC.
+
+# How We Teach This
+
+This new test waiters system should be included in the Ember guide's testing section. Information and examples should be provided to allow users to correctly author asynchronous code that can be correctly managed by the testing system.
```
rwjblue: Can you link to a few of the existing guide pages that will be rewritten?

text/0000-new-test-waiters.md
```diff
@@ -0,0 +1,111 @@
...
+
+In the above example, a new test waiter is built that is identified via a `name` string passed into the `buildWaiter` function. This allows the waiter to be identifiable, and that name is ultimately used with test isolation validation to help developers narrow down problems in their tests.
+
+The API used to signal whether an asynchronous operation has begun and ultimately ended is through the paired calls of `beginAsync` and `endAsync`: begin to denote the start of the asynchronous operation, and end to denote the end. Unique instances of async operations are identified using a `token` returned from `beginAsync`, which is subsequently provided to the `endAsync` call.
+
+This new test waiters system has been through multiple iterations of refinement, and is in use and integrated with the test isolation validation system. Furthermore, it is backwards compatible with the old test waiters system, allowing applications and addons to gradually migrate to using the new system.
+
+## Deprecating the old test waiters
+
+The old test waiters system will be deprecated in its own deprecation RFC.
```
rwjblue: Let's tweak this to be an overall "compatibility" section with:

* registerWaiter continues to work, and is not deprecated
* addons using ember-test-waiters work even if consumed in applications that do *not* use a new enough ember-test-helpers version (they don't get the nice new fancy helpful output, but it still works)

text/0000-new-test-waiters.md
```diff
@@ -0,0 +1,111 @@
...
+This new test waiters system should be included in the Ember guide's testing section. Information and examples should be provided to allow users to correctly author asynchronous code that can be correctly managed by the testing system.
+
+# Drawbacks
+
+- added work to integrate the new test waiters system with `ember-mocha`
+- the new test waiters system ships code to production bundles, though this is quite small
+
+# Alternatives
+
+- the existing test waiters system could be left in place
```
rwjblue: 😡

text/0000-new-test-waiters.md
scalvert: Ya I agree, and wasn't sure where to put this, so decided to be a bit snarky and put it here :)

I'll update this and call it out in the detailed design section.

text/0000-new-test-waiters.md
scalvert: I don't think any particular page should be rewritten. I think a new section should be added to https://guides.emberjs.com/release/testing/, which clearly defines what a test waiter is, and how and when to use them.

text/0000-new-test-waiters.md
scalvert: 😃

text/0000-new-test-waiters.md
scalvert: In fact, now that I think about it this should already work with `ember-mocha`, since we're merely depending on the integration with the `settled` check in `@ember/test-helpers`.

text/0000-new-test-waiters.md
```diff
@@ -0,0 +1,204 @@
...
+        // this unique async operation has ended.
+        waiter.endAsync(token);
+      });
+  }
+}
+```
+
+### `waitForPromise`
+
+The `waitForPromise` utility provides a convenience wrapper around the `TestWaiter` class for use with promises. It ensures the `endAsync` call is invoked in the `finally` of the configured promise.
```
Gaurav0: Can you add an example of how to use `waitForPromise`?

text/0000-new-test-waiters.md
```diff
@@ -0,0 +1,204 @@
...
+import { registerWaiter } from '@ember/test';
+
+registerWaiter(function() {
+  return myPendingTransactions() === 0;
+});
+```
+
+While reading the above is straightforward, when writing a test waiter using the old system it's easy to forget what the expected return value is: `true` or `false`. Additionally, it's a bit more cognitive overhead to derive what the intended result of the particular boolean return value is: does returning `true` result in the test waiter waiting or not?
+
+As mentioned before, there's no additional information provided via `registerWaiter`, and no way to capture stack traces at the call site. Unmanaged async that 'hangs' can cause your tests to stall and ultimately timeout. This is particularly problematic when trying to identify which of many test waiters has caused this timeout.
```
Gaurav0: How do you capture stack traces in the new test waiters design?

rwjblue: @Gaurav0 - When `.beginAsync` is called on a test waiter we capture the stack trace (in debug builds) and save it off.

text/0000-new-test-waiters.md
```diff
@@ -85,21 +85,114 @@ export default class Friendz extends Component {
...
+
+```ts
+export type WaiterName = string;
+export type Token = unknown;
+
+export interface IWaiter {
+  name: WaiterName;
+  waitUntil(): boolean;
+  debugInfo(): ITestWaiterDebugInfo[];
+}
```
rwjblue: Can you expand this slightly to include some API docs for each of the methods? Specifically, I'd like to explain the semantics of `waitUntil` (as long as this function returns `true` the waiter will be considered "waiting") and `debugInfo` (this information will be available in the test harnesses `getSettledState()` result, and is printed to the console when a test or a `waitFor`/`waitUntil` times out)

text/0000-new-test-waiters.md
```diff
@@ -85,21 +85,114 @@ export default class Friendz extends Component {
...
+
+```ts
+export interface ITestWaiter<T = Token> extends IWaiter {
+  beginAsync(token?: T, label?: string): T;
+  endAsync(token: T): void;
+  reset(): void;
+}
+```
+
+This interface is used for the two internal concrete types: `TestWaiter` and `NoopTestWaiter`. These types for the basis for the addon, and will likely satisfy the majority of use cases.
```
rwjblue: > These types for the basis

This confuses me a bit, maybe you meant:

```
These types form the basis
```

text/0000-new-test-waiters.md
```diff
@@ -85,21 +85,114 @@ export default class Friendz extends Component {
...
+
+The most common practice is to import and invoke the `buildWaiter` function to create a new test waiter. The recommendation is to do so at the module level, which allows a single waiter to be created per type. A single waiter is then usable across multiple instances.
```
rwjblue: We should propose an additional rule for eslint-plugin-ember to enforce the module scope-ness here

text/0000-new-test-waiters.md
scalvert: Most definitely.

text/0000-new-test-waiters.md
```diff
@@ -85,21 +85,114 @@ export default class Friendz extends Component {
...
+
+```ts
+function buildWaiter(name: string): ITestWaiter
+```
+
+In anything but a production build, this function will return a `TestWaiter` instance. When in production mode, a `NoopTestWaiter` will be returned, which as the name suggests is effectively a noop class. Since test waiters are intended to be called from application or addon code, but are only required to be _active_ when in tests, we use the `NoopTestWaiter` to noop the class' methods when run in production. This has a negligible impact on performance.
```
rwjblue: I think we can avoid mentioning NoopTestWaiter specifically, but instead mention that in production (not testing) builds that the `TestWaiter` class is "inert" and optimized for "speed".

text/0000-new-test-waiters.md
rwjblue: Ya, good idea

text/0000-new-test-waiters.md
```diff
@@ -85,21 +85,114 @@ export default class Friendz extends Component {
...
+- `registerWaiter` continues to work, and is not deprecated
+- addons using `ember-test-waiters` work _even if consumed in applications that do not use a new enough `ember-test-helpers` version_ (they won't get additional output via test isolation validation such as test waiter names or stack traces)
+
+The old test waiters system ultimately should be deprecated in its own deprecation RFC.
 
 # How We Teach This
 
 This new test waiters system should be included in the Ember guide's testing section. Information and examples should be provided to allow users to correctly author asynchronous code that can be correctly managed by the testing system.
 
+Specifically, calling this out in a separate section will allow readers to
```
rwjblue: Seems like this is cut off?

text/0000-new-test-waiters.md
scalvert: I added a brief description of how we do it.

text/0000-new-test-waiters.md
scalvert: Ya that's a typo.

text/0000-new-test-waiters.md
scalvert: I see no harm in mentioning it, just so people understand how it works slightly more than hand waving. Any issues with my leaving it in?

text/0000-new-test-waiters.md
scalvert: Correct. I left the rest in my brain. Fixed now.

text/0000-new-test-waiters.md
scalvert: Done!

text/0000-new-test-waiters.md
scalvert: Expanded!

text/0000-new-test-waiters.md
rwjblue: The issue is that the _how_ we make it inert is not really important (and definitely isn't public api). I think making it clear what our commitment is (that it is effectively zero cost in production), and that "for example, you could imagine that we use an alternative class implementation in production that makes all operations a no op".

scalvert: Sure. I can go either way, so will remove it.

text/0000-new-test-waiters.md
```diff
@@ -58,7 +58,7 @@ registerWaiter(function() {
...
 
-As mentioned before, there's no additional information provided via `registerWaiter`, and no way to capture stack traces at the call site. Unmanaged async that 'hangs' can cause your tests to stall and ultimately timeout. This is particularly problematic when trying to identify which of many test waiters has caused this timeout.
+As mentioned before, there's no additional information provided via `registerWaiter`, and capturing stack traces at the call site is currently not implemented. Unmanaged async that 'hangs' can cause your tests to stall and ultimately timeout. Not having stack traces is particularly problematic when trying to identify which of many test waiters has caused this timeout, as it's like looking for a needle in a haystack. The new system captures the stacks lazily, when the waiter's `beginAsync` method is called (more on `beginAsync` later). This allows for identifying the offending code more easily.
```
rwjblue: > The new system captures the stacks lazily, when the waiter's `beginAsync` method is called (more on `beginAsync` later).

I think this phrasing is _slightly_ confusing. We eagerly create the error object, but only force the stack to be reified when it's asked for (e.g. via `getSettledState`). I'm not sure how to say this better than what you have here though. 🤔

text/0000-new-test-waiters.md
```diff
@@ -104,6 +104,11 @@ export interface IWaiter {
 }
 ```
 
+- `name`: The name of the test waiter, which is used to help identify it in test isolation validation output.
+- `waitUntil`: Used to determine if the waiter system should still wait for async
+operations to complete. The `waitUntil` method will return `true` to signal completion.
+- `debugInfo`: Returns the `debugInfo` for each item tracking async operations in a waiter. The `debugInfo` for each waiter item is ultimately used in `@ember/test-helpers`' `getSettledState` function, which is ultimately used for test isolation validation output.
```
rwjblue: s/ultimately//

text/0000-new-test-waiters.md
scalvert: I also felt the same. I have an idea. Will update.

scalvert: Updated. Take a look and let me know what you think.

text/0000-new-test-waiters.md
scalvert: Ultimately I'll remove it.

text/0000-new-test-waiters.md
```diff
@@ -0,0 +1,238 @@
...
+
+Specifically:
+
+- `registerWaiter` continues to work, and is not deprecated
+- addons using `ember-test-waiters` work _even if consumed in applications that do not use a new enough `ember-test-helpers` version_ (they won't get additional output via test isolation validation such as test waiter names or stack traces)
+
+The old test waiters system ultimately should be deprecated in its own deprecation RFC.
+
+# How We Teach This
+
```
jenweber: I think we should add some more details about where the API docs themselves will live. Will they be in the addon README or would they be included in the api.emberjs.com docs?

text/0000-new-test-waiters.md
```diff
@@ -0,0 +1,238 @@
...
+The `ember-test-waiters` addon is backwards compatible with the old test waiters system, allowing applications and addons to gradually migrate to using the new system.
+
+Specifically:
+
+- `registerWaiter` continues to work, and is not deprecated
+- addons using `ember-test-waiters` work _even if consumed in applications that do not use a new enough `ember-test-helpers` version_ (they won't get additional output via test isolation validation such as test waiter names or stack traces)
+
+The old test waiters system ultimately should be deprecated in its own deprecation RFC.
+
+# How We Teach This
```
mixonic: I think the "How we teach this" should include guidance for how we communicate to addon developers vs. app developers.

For example we need to communicate to addon developers that they should expose waited async and avoid passing the burden of using the wrappers to the end user.

text/0000-new-test-waiters.md
jenweber: I would also like to see some bullet points on what information should go in cli.emberjs.com as advice for addon authors. cc @mixonic

scalvert: For sure. Added a callout that API docs will be available there.

text/0000-new-test-waiters.md
scalvert: This is a great callout. I've added a specific section that indicates that addon authors should have further direction on the importance of adding test waiters to their addon's unmanaged async.

scalvert:
Thanks for the feedback, @jenweber and @mixonic. Updated the RFC accordingly.

jherdman:
I recently had to use the current test waiter system with a third party evented API, e.g.

```javascript
SomeSystem.on('someEvent', function() {
  // ...
});

SomeSystem.on('error', function() {
  // ...
});

SomeSystem.asyncThing(arg1, arg2);
```

In this case you could call the async API function many times and end up with either the error or the handler hook. Maybe my imagination is failing me a little at this moment, but I'm not really sure how you connect one async call to a waiter token. An example might be helpful for this sort of problem.

scalvert:
@jherdman this is a really good question, and a tricky problem to solve. 

The test waiter system is predicated on the supposition that you are expected to execute _balanced_ calls. That means that there needs to be a reasonable expectation of the `endAsync` call happening after `beginAsync`. Depending on the library, that's not always the case. 

I think for your example above, the primitives in the `ember-test-waiters` package could be utilized to come up with test waiter implementation that was more fault tolerant. You'd want to make sure that whatever the implementation, your waiter reasonably covers your async behaviors without unintentionally timing your tests out (`endAsync` not being called for a specific `beginAsync`).

If there was some identifying feature of the unique calls, such as unique parameters, I could see creating a mechanism that stores tokens identified by call. This could be a simple map that's keyed on the call, and retrieves the token to call `endAsync` on.

If there isn't an identifying feature of the unique calls, then this poses a trickier problem.

OmShiv:
> This addon is great given a complex state of unmanaged component behavior. I however feel that this add-on will encourage devs to go more upwards in the testing pyramid, widening the middle (Integration) and top (Acceptance) layers.
> For eg., a developer may be encouraged to only test the parent component from an app-user's consumption point of view, where the test is waiting for some blackbox functionality (using this waiter) to complete before asserting behavior. For this we already have application tests. Thoughts?!

This was a summarization. Left comments on specific lines in the `.md` file, but can't see them. LMK if you can see, else, can paste here.

scalvert:
@OmShiv I don't believe this to be the case.

The test waiter system _already exists_; we're further refining it to make it more usable and to provide a better path to debugging async behavior.

While I believe that what you say could happen, we can intentionally call this out in the documentation. That said, testing and the philosophy people use to test their apps, can be a contentious discussion. My preference would be to provide great tools and APIs to help developers succeed, without being overly prescriptive about how they get there.

OmShiv:
@scalvert 
Totally. I guess better would have been to ask this: 
_With this pull-request, are you not expecting review on the proposition itself and taking for granted that it has inherent value, while only looking for review on the implementation/detail/documentation?_

In other words, are we taking for granted that "New Test Waiters" add value and this fact itself is **exempt** from this RFC? If YES, ignore from here...

**If NO, read on - here's my concern.** 

It's debatable that this add-on will add value to developers concerned with async problems in tests. In addition it's a 3-cost jump:

1. This add-on **doesn't fix** async tests (as we saw in our previous mutual efforts), but will helps debugging. The existing waiter support from Ember is sufficient to pin-point the async and apply the fix
2. It adds test-related code that you have to ship in Production
3. And most important, this add-on will encourage writing Integration tests more top-to-bottom (acceptance like) - misbalancing the test pyramid (widening the middle and top layers, and shrinking the Unit layer at the bottom). It prevents devs from refactoring and decoupling the async - that in many cases (like we saw in ours) is the root cause of tests timing out. 

Consider this:
```js
// Parent component
<ChildImageComponent props=imageProps/>

// Child Image component
init() {
  workOnReceivedProps(); // async operation
  drawImage(); // async operation
}

// Parent test
await render(hbs`<ParentComponent .../>`)
"do some new test waiter magic" // since we can't control child's async
assert.dom(image was drawn...) // likely to timeout 
```
Such tests should be broken down into 
- parent component only tests it's own stuff
- parent tests that child was rendered (but not the drawing of the image)
- child, in its own separate test, verifies that when options are given, image is drawn
- if possible, child should (in some cases, must) **mock** the async when testing. 

This add-on however will promote testing more of child behavior directly from parent, misbalancing test pyramid greatly. For a stable pyramid, you should be able to move tests from Acceptance into Integration, and further break them into more and more unit tests - which is best both for faster build time and rugged tests. (Imagine a pyramid with wider base and thinner apex, extremely difficult to fall and is a rock stable structure, compared to a thin base pyramid). 

With any tool/methodology, a little prescription (not a lot) is **needed**, and ember test strategy shouldn't be an exception. Otherwise it may be the case that async timeout issues are a result of greatly deviating from the prescription in the first place and we may be overlooking the opportunity to fix them by promoting wait helpers.

scalvert:
@OmShiv 

Precisely. This isn't really a debate about the merits of test waiters themselves. They're an established tool that currently exist in the test infrastructure, and are already used by a number of libraries to manage async.

As to the latter part, this library doesn't profess to fix any async in tests magically, but merely to provide a tool, as previously stated, that helps you address patterns in your code that aren't easily handled another way. The waiters don't merely help with debugging; they also help with signaling to the test system when to wait in a deterministic fashion.

I don't disagree with your assertion about the correct orientation of the test system, and I also think that discussion and what follows it is best reserved for a separate forum.

rwjblue:
> Are we taking for granted that "New Test Waiters" do add value - and this fact itself is exempt from this RFC?

I don't think things are quite this simple. Test waiters are already a concept in Ember (and have been for quite a long time), and that is _basically_ not going to change (though this RFC does make significant enhancements).

> It's debatable that this add-on adds value to developers concerned with async problems in tests. Also, it's a 3-cost jump:
>
> * This add-on doesn't fix async tests (as we saw in our previous efforts), but only helps debugging. The existing waiter support from Ember is sufficient to pin-point the async and apply the fix

Nothing we can do will magically fix all async issues in tests. We shouldn't even really strive for that. Hiding async (e.g. via mocking) is fundamentally flawed, and will become a never ending game of whack-a-mole. The goal of the current waiter system in Ember is to allow the specific unit that introduces asynchrony to properly instruct the test environment when it is completed. If everything in the system that introduces it's own async leverages the waiter system (either the current one, or the one proposed in this RFC) the tests **will** not exhibit issues stemming from uncontrolled async.

> * It adds test-related code that you have to ship in Production

Indeed, this is true. The size increase this infers is mitigated by ensuring that we ship as little code as possible in production. One thing to keep in mind here is that, while this _does_ propose actually shipping some test related bits in your app (or addon) code that "cost" has a very specific and distinct benefit: we will be able to make it possible to properly run your tests in production builds (which we cannot do with today's system).

> * Most importantly, it will encourage writing Integration tests more top-to-bottom (acceptance like) - misbalancing the test pyramid by widening the middle and top layers, and shrinking the Unit layer at the bottom. It prevents devs from refactoring and decoupling the async - that in many cases (like we saw in ours) can be the root cause of tests timing out.

I think this is actually massively misleading, and possibly the largest point of disagreement here. Let's be super clear: **NOTHING** in this RFC suggests that tests should not be properly composed. Extracting logic to simpler pieces (stand alone functions, simple classes, etc) that are directly unit tested is absolutely a great practice, and this RFC certainly does not suggest otherwise. That sort of extraction leads to very nice composition in actual use and _enables_ the very test pyramid that you are describing.

> Consider this:
> 
> ```js
> // Parent component
> <ChildImageComponent props=imageProps/>
> 
> // Child Image component
> init() {
>   workOnReceivedProps(); // async operation
>   drawImage(); // async operation
> }
> 
> // Parent test
> await render(hbs`<ParentComponent .../>`)
> "do some new test waiter magic" // since we can't control child's async
> assert.dom(image was drawn...) // likely to timeout 
> ```

I don't really understand the inference you are making here. If the child component encapsulates the async that it is introducing, no changes are needed in the parents tests (e.g. that `await render(hbs('<ParentComponent ... />'))` will properly wait until the child component's async is completed). Specifically, the two inline comments RE: "do some new test waiter magic" and "likely to timeout" seem invalid.

@scalvert - We should consider how to update the RFC to make sure readers do not fall into this faulty assumption...

> Such tests should be broken down into

I do *not* believe this is the correct way to break down tests for this, and in fact if you did do this your tests would be extremely brittle (where a refactoring that resulted in the same DOM but different underlying implementation, would fail many of these tests).

Let's break it down in detail though.

> * parent component only tests it's own stuff
> * parent tests that child was rendered (but not the drawing of the image)

A significant part of the contract provided by the parent component is that it renders the child component. Testing its own stuff *means* rendering the child component by definition. If you refactored these tests to provide a replacement for `ChildComponent` (which is pretty easy to do, and has nothing to do with this RFC at all), you would very easily end up in situations where you would do an internal refactor to `ParentComponent`s template that results in passing tests (because you've essentially provided a stub for `ChildComponent`) but a completely broken reality, or alternatively that the tests all fail but things actually _do_ work properly. Both cases are a sign of brittle tests.

> * child, in its own separate test, verifies that when options are given, image is drawn

Totally agree! Components should test what they do. 😃 

> * if possible, child should (in some cases, must) **mock** the async when testing.

🙀 This _sounds_ good, but is actually a bit concerning and takes care to do properly. I do agree with some form of statement around the child "mocking async", but I think we disagree with what that means. What I mean is that you mock your API requests with something like Pretender, what I suspect you mean is that you monkey patch the methods on your component to avoid doing the async in tests. These two approaches are _very_ different.

---

@OmShiv - Thank you very much for chatting through these things here. I think it is extremely good to have these types of conversations, so that hopefully others can learn from our mistakes (and from what we do correctly). In the end, this RFC is **not** about the theology of testing or about what kinds of tests that you write (which is where it seems that our disagreement lies), but is instead about providing good primitives that allow each piece of the system to properly indicate to the rest of the test infrastructure that there is currently async underway. Nothing in this RFC prevents you from authoring tests in the style that you seem to prefer (though as you might imagine, I strongly suggest you rethink that style).

text/0000-new-test-waiters.md
```diff
@@ -0,0 +1,248 @@
...
+
+This interface is used for the concrete `TestWaiter` type. This type forms the basis for the addon, and will likely satisfy the majority of use cases.
+
+The most common practice is to import and invoke the `buildWaiter` function to create a new test waiter. The recommendation is to do so at the module level, which allows a single waiter to be created per type (this should likely be enforced via a lint rule added to `eslint-plugin-ember`). A single waiter is then usable across multiple instances.
+
+```ts
+function buildWaiter(name: string): ITestWaiter
+```
+
+In anything but a production build, this function will return a `TestWaiter` instance. When in production mode, we make this instance _inert_ and essentially no cost to invoke. Since test waiters are intended to be called from application or addon code, but are only required to be _active_ when in tests, this process of making the instance _inert_ is important. Even though code is still invoked, this has a negligible impact on performance.
```
mixonic: How do we feel about the plausibility of a system which would make this configurable. Basically users who want to optimize for test-ability of production builds can leave the inert code, and users who want to optimize for every byte can strip the inert code out?

Or is this already designed in such a way that it would be trivial to author an addon which stripped the inert code from production builds, allowing byte-watchers to opt-in to that size optimization? (I think it would be easy, but I'd like to hear from the authors)

rwjblue: I think there are two parts to this:

1. ensuring that the addon's code itself is not shipped outside of testing
2. authoring a babel plugin that allows the statements in app and addon code to be removed

Both are doable and should be investigated during implementation (though I don't think it's particularly important to happen _before_ we land the RFC).

rwjblue:
After review and discussion with the core team at today's meeting, we are moving this RFC into final comment period.

text/0000-new-test-waiters.md
scalvert: These are good callouts, @mixonic. Early on @rwjblue and I discussed the possibility of stripping code from prod builds that weren't destined to be tested. I think as Rob describes above, we could absolutely do this. 

We also need to add a separate flag other than `DEBUG`, such as `FOR_TESTING` (don't worry about the name; naming is hard yada yada), but that will allow us to do prod builds that will allow for testing. Obviously this is a broader change that impacts a number of things, but is something we've also discussed.

scalvert:
@jherdman I wanted to circle back on your comment, as my reply to your comment really only touched on what a possible solution could be WRT addressing using waiters with unique async call patterns. 

I've also [opened a PR](https://github.com/rwjblue/ember-test-waiters/pull/112) on `ember-test-waiters` to ensure that `endAsync` won't error if called repeatedly after a previously used token is used again. Hopefully this will help address some of your concerns when using this library with evented systems.

text/0000-new-test-waiters.md
```diff
@@ -0,0 +1,248 @@
...
+
+- the new test waiters system ships code to production bundles, though this is quite small
+
+# Alternatives
+
+- the existing test waiters system could be left in place
+
+# Unresolved questions
+
+- what is the correct order of operations to roll a change like this out?
```
mixonic: Before merge please remove this line. Under detailed design you can add a section "rollout" which makes clear the proposal is:

* rename the addon
* add it to the default ember-cli blueprint
* there is no schedule for deprecating the old waiters, requires a new RFC

If that is correct this comment is simply an organizational, doesn't require FCP restart.

dgeb:
Sorry for the last minute comment, but the core team discussed the names of the interfaces used in this RFC during today's meeting. It occurred to me that we are almost exclusively using interfaces in both Glimmer and Ember that avoid the `I` prefix (It seems that an `IRegistry` interface has snuck in to the codebase, but it's not exposed as a public interface).

There is clearly a healthy debate in the TS community (and many others) about this topic, and there are even [linting rules](https://palantir.github.io/tslint/rules/interface-name/) that favor each side. Hopefully we can all agree that consistency is the most important part of this decision. And in the Ember ecosystem we've favored `never-prefix`.

@rwjblue @scalvert Can we explore refactoring `IWaiter` -> `Waiter` and `ITestWaiter` -> `TestWaiter`? The implication then is that the `TestWaiter` class would need to be renamed. A standard but perhaps unergonomic choice would be to append `Impl` to class names that otherwise conflict with interfaces. Perhaps `BaseTestWaiter` would differentiate this class better? The point is that there's a reason that all implementations of the `TestWaiter` interface are not also instances of this class, and that's the distinction that the name should get at. I'm trying to not be too prescriptive or opinionated here - just trying to enforce consistency. Hopefully we can arrive at a nice alternative for the class names.

rwjblue:
The classes themselves are not exposed as public api, only the `buildWaiter` which returns (currently) the `ITestWaiter` interface. I don’t think removing the `I` prefix is a substantial change.

text/0000-new-test-waiters.md
scalvert: Will do.

scalvert:
@dgeb I'm totally happy to go with whatever convention the rest of the codebase dictates. Interestingly, @stefanpenner and I just had a discussion regarding `I` vs. no `I` for interface prefixes. I largely do it because I have a decade of C# standards drilled into me, but I can absolutely see the argument both ways.

In this case, we'll rename it.

dgeb:
@rwjblue @scalvert thank you both for your quick responses and your willingness to be flexible here! I'm glad that neither of you think this will be a substantial change. Hopefully we'll be able to restart the FCP very soon.

scalvert:
@dgeb https://github.com/rwjblue/ember-test-waiters/pull/117

OmShiv:
@rwjblue, agree with all of your thoughts here, with **ONLY** disagreement around brittleness. Firstly, I appreciate you giving detailed explanation, and appropriate consideration to my concerns. My disagreement probably is best suited for a different forum. But pardon my intent to consider this as an opportunity to describe here.
___
In short, I differ here per my understanding:
>Testing its own stuff means rendering the child component by definition. 

If you compared how child is tested for rendering v/s how parent consumes it - the markup is same (by design. In fact, former is more robust/thorough)
```hbs
<div> parent's dom ...
{{!How parent uses child 👇}}
<ChildComponent prop1={{this.parentProp}}/>
```
```js
// How child's test looks like: child-test.js
await render(hbs`<ChildComponent prop1={{testProp}}/>)
```
The way child is **EVER** going to be used, specifically, `<ChildComponent prop1={{testProp}}/>` - is thoroughly tested at child (including behavior). This includes parent's usage (what I assume you called "reality") as a **subset of all possible usages**. Otherwise, it would be un-scalable, i.e., more than 1 parent consumers testing child end-to-end, including covering for child's async via a waiter-mechanism. That is both slow and non DRY. Also, one can't keep that test in a particular component of choice either, and call that all contracts are verified because  Parent-1's contract is verified. 
___
With this approach, however, I agree we are missing one thing - the specific contract with **given** parent. Verification of that contract should **move a layer up** into Acceptance, coz otherwise the contract doesn't make sense. In my opinion, any waiter mechanism is best suited there, including this new Test waiters - where I definitely see value!

More so, all such contracts, since they are only important as long as the user story is, should be **compacted** into even slimmer acceptance criteria, constituting the basis for Test Pyramid where top layer is thinner. Acceptance should be broken into Integration, and further, Integration into more units since a wider pyramid at base is more stable and infallible. 
>"Acceptance tests should be written using the standard agile framework of a **user story**" 
>-BDD, Wikipedia
___
Following this, these statements stand incorrect acc. to me. **Implementation wise:**
> a refactoring that resulted in the same DOM but different **underlying implementation**

_Reason_: As long as my user story (acceptance crit.) is verified. I don't care if underlying implementation was different. Then next, **"Reality" wise:**
> you would do an internal refactor ... that results in passing tests ... but a completely **broken reality**

_Reason:_ The acceptance test should fail if reality was changed. If it doesn't, I shouldn't care.
>...alternatively that the tests all fail but things actually do work properly

Won't happen either. Because if all tests fail, I must have broken down a single user story into integration, and further into unit tests. They are failing is a hint of things are not working properly. 
___
Brittleness, and mocking. Pardon me for further exemplifying this with a known dummy example.
- testing an `add` function that takes in two numbers, downloads an image asynchronously, and then returns the sum
  - **my understanding:** test waiter provides mechanism that helps the `add` function to notify its consumers correctly when the image download is complete
  - **my argument**: this is helpful, but not needed for thorough testing
  - I would mock the image download service here, and test it separately
  - my test for `add` will only await the mocked download promise (which is resolved) and immediately verify the result
    - _probably we are in agreement here_
  - this is **sufficient**
    - _probably we differ here_
- testing another function `multiplyAfterAdd` which consumes `add`
  - I would also mock `add` entirely and take for granted that it works
    - _probably we differ here, and this is what you are pointing as brittleness - correct me if wrong_ 

One may argue that the signature of `add` can change while tests for only `add` are updated. So all tests collectively pass, but things have changed, which is worrisome. This is **incorrect for an application with dependencies** like this, because if `add` function changed signature, it no longer adds. It is **highly unlikely** in an application that `multiplyAfterAdd` is not updated to use a different dependency. Because it no longer multiplies after adding, and it should use a different dependency.

In other words, to circumvent brittleness, tests for `multiplyAfterAdd` can not be the ideal place. I am understanding test-waiter acting as an aid to devs to write application-like tests for `multiplyAfterAdd` by correctly tracking `add`'s async. I strongly believe it's not **needed** and a more scalable approach is to move up the pyramid to verify all `add` contracts at a single top-level place, **once** (may be as part of user story) where a test-waiter including existing ones can prove useful.

text/0000-new-test-waiters.md
scalvert: Done.

rwjblue:
@OmShiv - Thanks for adding more detail there. It sounds like we agree on quite a few things, which is great!

The primary thing that I think we will continue to disagree about is leveraging mocks to replace internal collaborators of a given unit. I absolutely agree that it is important to be _able_ to do this (and would consider blockers to this ability to be a bug that must be fixed), but I likely could not disagree much more with the idea that this is something that should be done in common practice.

The good news here is that I don't need to convince you that I'm right (and you don't need to convince me that you are right), because we **both** see the requirement for the system proposed in this RFC. We only differ in the specific circumstances that we'd individually reach for it. I'm glad that we were able to iron out that agreement here, thank you again for your time in explaining your position.

rwjblue:
We discussed this at todays core team meeting, and are very much in favor of moving forward!

PR #585, opened 2020-01-27
Registry APIs: deprecate micro-syntax; introduce replacement API

chriskrycho:
[Rendered RFC](https://github.com/chriskrycho/ember-rfcs/blob/registry-apis/text/0585-improved-ember-registry-apis.md)

Thanks to @rwjblue, @cafreeman, @jamescdavis, @pzuraq, @dfreeman, and @suchitadoshi1987 for reviewing earlier drafts of this and providing helpful feedback and contributions to the design laid out here!

---

Introduce a new, object-based API for all registry APIs; deprecate the current string-based microsyntax registry APIs; and introduce a `capabilities` property to the resolver to safely support existing resolvers. Extend the new API for service and controller injections, and deprecate their microsyntax for namespaced injections.

Today the registry APIs are all of shapes roughly like this:

```js
getOwner(this).lookup('service:session');
```

This RFC proposes that they would instead be written like this:

```js
getOwner(this).lookup({ type: 'service', name: 'session' })
```

Service and controller injections which include a namespace currently look like this:

```js
class Example {
  @service('global@session')
  session;
}
```

This RFC proposes that these would instead be written like this:

```js
class Example {
  @service({ namespace: 'global', name: 'session' })
  session;
}
```

### DIFF
--- a/text/0585-improved-ember-registry-apis.md
+++ b/text/0585-improved-ember-registry-apis.md
@@ -302,10 +302,10 @@ The `'<namespace>@<name>'` form is deprecated and will be removed at 4.0. The no
     import { inject as service } from '@ember/service';
 
     export default class Example {
-      @inject foo;
-      @inject('bar') barRenamed;
-      @inject({ name: 'quux' }) quuxViaIdentifier;
-      @inject({ namespace: 'baz', name: 'neato' }): neatoNamespaced;
+      @service foo;
+      @service('bar') barRenamed;
+      @service({ name: 'quux' }) quuxViaIdentifier;
+      @service({ namespace: 'baz', name: 'neato' }): neatoNamespaced;
     }
     ```
 
@@ -315,7 +315,7 @@ The `'<namespace>@<name>'` form is deprecated and will be removed at 4.0. The no
     import { inject as service } from '@ember/service';
 
     export default class Example {
-      @inject('baz@neato') neatoNamespaced;
+      @service('baz@neato') neatoNamespaced;
     }
     ```
 
--- a/text/0585-improved-ember-registry-apis.md
+++ b/text/0585-improved-ember-registry-apis.md
@@ -223,8 +223,6 @@ Some APIs refer to *factories* and *factory managers*. The API for these classes
  }
 ```
 
-**TODO: nail down what here is public API during RFC discussion.**
-
 #### `Owner` API diff
 
 ```diff
--- a/text/0585-improved-ember-registry-apis.md
+++ b/text/0585-improved-ember-registry-apis.md
@@ -360,6 +360,7 @@ The majority of uses are likely to be codemoddable, but not *all* will.
 >     copyToClipboard(source) {/* no-op */}
 >   }
 > )
+> this.owner.inject('controller', 'session', 'service:session');
 > ```
 >
 > Should be changed to:
--- a/text/0585-improved-ember-registry-apis.md
+++ b/text/0585-improved-ember-registry-apis.md
@@ -323,7 +323,7 @@ All existing *static* microsyntax invocations can be straightforwardly migrated 
 
 For example, this (in a test) is migrate-able:
 
-```js
+```diff
 - this.owner.lookup('service:session');
 + this.owner.lookup({ type: 'service', name: 'session' });
 ```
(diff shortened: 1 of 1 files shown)


text/0585-improved-ember-registry-apis.md
```diff
@@ -0,0 +1,793 @@
...
+
+The `'<namespace>@<name>'` form is deprecated and will be removed at 4.0. The no-argument form, and the name-only string form are still allowed. Users *may* invoke using `{ name: '<name>' }` but are not required to.
+
+- **Good** (using services as an example):
+
+    ```js
+    import { inject as service } from '@ember/service';
+
+    export default class Example {
+      @inject foo;
```
NullVoxPopuli: import renamed, but used without the rename

text/0585-improved-ember-registry-apis.md
```diff
@@ -0,0 +1,793 @@
...
+    }
+    ```
+
+- **Deprecated:**
+
+    ```js
+    import { inject as service } from '@ember/service';
+
+    export default class Example {
+      @inject('baz@neato') neatoNamespaced;
```
NullVoxPopuli: same here

text/0585-improved-ember-registry-apis.md
chriskrycho: Ah, nice catch on these; thank you! (No matter how many times you proof-read, as *soon* as it goes up all the mistakes are obvious. 😂)

rwjblue:
Thanks for working on this @chriskrycho, I'm in favor of moving in this direction...

suchitadoshi1987:
@chriskrycho Great Work on introducing this new API, this fits perfectly for any futuristic changes coming to the framework!

jamescdavis:
I'm very much in favor of this!

chriskrycho:
As I've been mulling on names, I'm leaning slightly toward preferring `FactoryIdentifier` and `FactoryTypeIdentifier` for the primary identifier names, which make a nice complement to `InjectionIdentifier` for the service and controller injection identifiers. Thoughts?

text/0585-improved-ember-registry-apis.md
```diff
@@ -0,0 +1,798 @@
...
+
+### `Resolver`
+
+Ember’s `Resolver` function is public API, designed to be customized and overridden. To support backwards compatibility with existing custom resolvers, and following the lead of the [Custom Components RFC (#0213)](https://github.com/emberjs/rfcs/blob/master/text/0213-custom-components.md#capabilities) and the [Element Modifiers RFC (#0373)](https://github.com/emberjs/rfcs/blob/master/text/0373-Element-Modifier-Managers.md#capabilities), we introduce a `capabilities` property to the `Resolver` API, which may be `undefined` or the return value of a new `capabilities` function, which defines a `Capabilities` type.
+
+The `capabilities` function is a public export of the `@ember/application` module:
+
+```ts
+interface OptionalCapabilities {
+  modulesBased?: boolean;
```
buschtoens: Can you please explain what `modulesBased` is? Is this just an example capability or are there concrete concepts associated with it already?

chriskrycho: Excellent call-out. This is rationalizing the existing flag (referenced in the prior art section) using the new capabilities system; I'll add a note to that effect!

jmdejno:
Great direction for supporting Typescript in Ember more easily and deprecating and archaic and niche Ember syntax! Awesome @chriskrycho, and totally agree on the API signatures / interfaces defined in your RFC vs. the other options considered.

text/0585-improved-ember-registry-apis.md
chriskrycho: Resolved in af36f6d!

Gaurav0:
I just wanted to say this is one of the most well written RFCs I've seen. Great job.

rwjblue:
After review and discussion with the core team at today's meeting, we are moving this RFC into final comment period.

rwjblue:
We discussed this at todays meeting, and are very excited to land this!

chriskrycho:
Excellent—I look forward to helping get it landed! Thanks @rwjblue for championing it!

PR #599, opened 2020-03-02
Test Co-Location

NullVoxPopuli:
[rendered](https://github.com/NullVoxPopuli/rfcs/blob/colocated-tests/text/0000-colocated-tests.md)

### DIFF
--- a/text/0000-colocated-tests.md
+++ b/text/0000-colocated-tests.md
@@ -1,0 +1,151 @@
+- Start Date: 2020-03-02
+- Relevant Team(s): Ember CLI
+- RFC PR: (after opening the RFC PR, update this with a link to it and update the file name)
+- Tracking: (leave this empty)
+
+# Co-located Tests
+
+## Summary
+
+As apps grow, and need to be refactored, the ability to have tests be co-located with their concerned tested behavior enables faster refactors and better overall upkeep as related files are grouped together.
+
+## Motivation
+
+Similar to why people choose the Pods layout, or Module Unification (back when it was available in the canary build), and more recently, component class/template co-location, the ability to group tightly coupled things together allows for much quicker implementation, iteration, and maintenance of any unit of work.
+
+## Detailed design
+
+Nothing about existing tests needs to change. All tests currently living in the `tests/` directory may remain there and coexist with co-located tests.
+
+Primary changes needed:
+
+When building an ember app, based on if tests are included in the build or not, as they would be for development and test builds, ember-cli must configure broccoli to skip over test files within the `app` directory.
+
+Additionally, for the test bundle, the `app` directory must be added to search for files that are tests.
+
+Test files could remain using the current hyphenated suffix that they currently use -- example: `my-component-test.js`.
+But, because the only name part that could be searched for is `-test` at the end of the filename, there is the possibility that valid components, controllers, routes, etc could legitimately end with `-test`, causing files to incorrectly get removed or added to the app or test bundles.
+To counter this conflict, co-located tests must end with `.test`, e.g.: `my-component.test.js`. Because having extra periods in the filename isn't allowed in framework objects, using `.test` should be fairly safe to search for when building the app and test bundles.
+
+More concisely:
+
+For the app bundle,
+ - previously
+   - include: `app/**/*.{js,hbs}`
+ - now
+   - include: `app/**/*.{js,hbs}`
+   - exclude: `app/**/*.test.{js,ts}`
+
+For the test bundle,
+ - previously
+   - include: `tests/**/*.{js}`
+ - now
+   - include: `tests/**/*.{js}`
+   - include: `app/**/*.test.{js}`
+
+
+Like with component co-location, there will be different preferences with how to group files together. It may be undesirable to have a flat list of 20 routes, and then an additional flat list of 20 tests for those routes all in the same folder.
+The `.ember-cli` file provides the option to further nest components' class and template files as `app/components/{component-name}/index.{js,hbs}` via the `componentStructure` property (set to `nested`).
+This same flexibility should be available for routes, controllers, etc -- but is out of scope for this RFC as other resolver implementations may be in-flight.
+Avoiding implementing `componentStructure` for the rest of the `app` directory does not have any impact on test co-location as the resolver is not involved with finding test file locations. For example, it would be valid to have `app/routes/{route-name}.js` and tests located at `app/routes/__tests__/{route-name}.test.js`.
+
+
+### Reference: Where do the existing tests get migrated to?
+While tests in `app/` could live anywhere, these are to be the default locations for the generators to use.
+
+#### Route
+
+previously:
+ - `app/routes/{route-name}.js`
+ - `tests/unit/{route-name}-test.js`
+now:
+ - `app/routes/{route-name}.js`
+ - `app/routes/{route-name}.test.js`
+
+#### Service
+
+previously:
+ - `app/services/{service-name}.js`
+ - `tests/unit/{service-name}-test.js`
+now:
+ - `app/service/{service-name}.js`
+ - `app/service/{service-name}.test.js`
+
+#### Controller
+
+previously:
+ - `app/controllers/{controller-name}.js`
+ - `tests/unit/{controller-name}-test.js`
+now:
+ - `app/controllers/{controller-name}.js`
+ - `app/controllers/{controller-name}.test.js`
+
+#### Component
+
+previously:
+ - `app/components/{component-name}.js`
+ - `app/components/{component-name}.hbs`
+ - `tests/integration/components/{component-name}-test.js`
+now:
+ - `app/components/{component-name}.js`
+ - `app/components/{component-name}.hbs`
+ - `app/components/{component-name}.test.js`
+
+or using `"componentStructure": "nested"`:
+ - `app/components/{component-name}/index.js`
+ - `app/components/{component-name}/index.hbs`
+ - `app/components/{component-name}/.test.js`
+
+#### Helpers
+
+previously:
+ - `app/helpers/{helper-name}.js`
+ - `tests/integration/helpers/{helper-name}-test.js`
+now:
+ - `app/helpers/{helper-name}.js`
+ - `app/helpers/{helper-name}.test.js`
+
+#### Utils
+
+previously:
+ - `app/utils/{util-name}.js`
+ - `tests/unit/{util-name}-test.js`
+now:
+ - `app/utils/{util-name}.js`
+ - `app/utils/{util-name}.test.js`
+
+#### Acceptance Tests
+
+Acceptance tests do not have a core affiliation with any other file, so co-locating them _may_ not make sense, depending on what a particular test file is doing.
+
+While tests may live anywhere in the `app` directory, the default location for acceptance tests will still be `tests/acceptance/`.
+
+
+## How we teach this
+
+> What names and terminology work best for these concepts and why? How is this
+idea best presented? As a continuation of existing Ember patterns, or as a
+wholly new one?
+
+> Would the acceptance of this proposal mean the Ember guides must be
+re-organized or altered? Does it change how Ember is taught to new users
+at any level?
+
+> How should this feature be introduced and taught to existing Ember
+users?
+
+## Drawbacks
+
+While the community migrates, there will be a number of apps that either haven't migrated, or will be in the middle of migrating for a while (depending on app size).
+It would be benificial to provide a CLI migration tool to automatically move files from today's standard locations to the new locations.
+
+## Alternatives
+
+Prior Art:
+ - Jest projects -- where co-location happens everywhere
+ - Module Unification -- every was co-locatable, even tests -- though there were bugs wit the tests showing up in the app bundle.
+
+
+## Unresolved questions
+
+None (yet).
--- a/text/0000-colocated-tests.md
+++ b/text/0000-colocated-tests.md
@@ -81,6 +81,9 @@ previously:
 now:
  - `{app,addon}/routes/{route-name}.js`
  - `{app,addon}/routes/{route-name}.test.js`
+pods:
+ - `{pod-namespace}/{route-name}/route.js`
+ - `{pod-namespace}/{route-name}/.test.js`
 
 #### Service
 
@@ -99,6 +102,9 @@ previously:
 now:
  - `{app,addon}/controllers/{controller-name}.js`
  - `{app,addon}/controllers/{controller-name}.test.js`
+pods:
+ - `{pod-namespace}/{controller-name}/controller.js`
+ - `{pod-namespace}/{controller-name}/.test.js`
 
 #### Component
 
@@ -115,6 +121,7 @@ or using `"componentStructure": "nested"`:
  - `{app,addon}/components/{component-name}/index.js`
  - `{app,addon}/components/{component-name}/index.hbs`
  - `{app,addon}/components/{component-name}/index.test.js`
+
 
 #### Helpers
 
--- a/text/0000-colocated-tests.md
+++ b/text/0000-colocated-tests.md
@@ -162,6 +162,18 @@ users?
 
 This RFC does not deprecate the old test location -- it only suggests adding test locations -- so people can take their time migrating or completely avoid migrating altogether and wait for a tool to automatically move (nearly) everything for them.
 
+It should be noted that this RFC is not proposing anything happen with non-test files in the test folder.
+ - `tests/helpers/**`
+ - `tests/index.html`
+ - `tests/test-setup.js`
+
+During migration, the `tests/helpers/**` files should switch to absolute imports, if they are not already.
+
+For apps: `<app-name>/tests/helpers/**`
+
+For addons, these helpers may be moved to `addon-test-support`, but also `<addon-name>/tests/helpers/**` should be feasible as well.
+
+
 ## Drawbacks
 
 While the community migrates, there will be a number of apps that either haven't migrated, or will be in the middle of migrating for a while (depending on app size).
--- a/text/0000-colocated-tests.md
+++ b/text/0000-colocated-tests.md
@@ -94,7 +94,7 @@ now:
 or using `"componentStructure": "nested"`:
  - `app/components/{component-name}/index.js`
  - `app/components/{component-name}/index.hbs`
- - `app/components/{component-name}/.test.js`
+ - `app/components/{component-name}/index.test.js`
 
 #### Helpers
 
--- a/text/0000-colocated-tests.md
+++ b/text/0000-colocated-tests.md
@@ -61,6 +61,14 @@ The `.ember-cli` file provides the option to further nest components' class and 
 This same flexibility should be available for routes, controllers, etc -- but is out of scope for this RFC as other resolver implementations may be in-flight.
 Avoiding implementing `componentStructure` for the rest of the `app` directory does not have any impact on test co-location as the resolver is not involved with finding test file locations. For example, it would be valid to have `app/routes/{route-name}.js` and tests located at `app/routes/__tests__/{route-name}.test.js`.
 
+### Addons, in-repo Addons and Engines
+
+For normal addons, living as their own entire project, the lookup rules for apps will apply.
+For in-repo addons and engines, there is the possibility that the implementation of this feature would enable things not possible today.
+For example, if the "test finder" followed the glob: `**/*(\.test.js|-test.js)`, (excluding node\_modules, etc), that would include tests that are in in-repo addons and engines,
+which today, do not support their own tests -- the in-repo addon or engine still wouldn't support their own tests, exactly, as they would run as part of the parent
+app's test suite. This might be desired behavior as it would allow the in-repo addon and engine to finally co-locate their tests with the behavior that those units implement.
+
 
 ### Reference: Where do the existing tests get migrated to?
 While tests in `app/` could live anywhere, these are to be the default locations for the generators to use.
@@ -172,4 +180,5 @@ Prior Art:
 
 ## Unresolved questions
 
-None (yet).
+- Given that in-repo addons can't have their own tests, can engines?
+- Would bundling all tests in an app into a single test suite have undesirable consequences?
--- a/text/0000-colocated-tests.md
+++ b/text/0000-colocated-tests.md
@@ -147,6 +147,8 @@ This RFC does not deprecate the old test location -- it only suggests adding tes
 While the community migrates, there will be a number of apps that either haven't migrated, or will be in the middle of migrating for a while (depending on app size).
 It would be benificial to provide a CLI migration tool to automatically move files from today's standard locations to the new locations.
 
+When moving test files, it's possible that the import paths will be wrong. For example, if someone had `../../app/utils/my-utils.js` in their test file, the path would be wrong if the file moved to a new location.
+
 ## Alternatives
 
 Prior Art:
--- a/text/0000-colocated-tests.md
+++ b/text/0000-colocated-tests.md
@@ -180,5 +180,12 @@ Prior Art:
 
 ## Unresolved questions
 
+- Should the root of test resolution be the project root, or app/addon folder?
+  - if project root,
+    - this gives ultimate flexibility, and people could literally do whatever they want.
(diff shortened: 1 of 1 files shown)


text/0000-colocated-tests.md
```diff
@@ -0,0 +1,159 @@
...
+ - `tests/integration/components/{component-name}-test.js`
+now:
+ - `app/components/{component-name}.js`
+ - `app/components/{component-name}.hbs`
+ - `app/components/{component-name}.test.js`
+
+or using `"componentStructure": "nested"`:
+ - `app/components/{component-name}/index.js`
+ - `app/components/{component-name}/index.hbs`
+ - `app/components/{component-name}/.test.js`
```
simi: :thinking: What about `index.test.js` to prevent this file behave as a hidden on some systems?

NullVoxPopuli: oh right! I forgot windows does that. good call. will update

NullVoxPopuli: actually, I think Mac does, too.

and now that I think of it.. everyone does _by default_.

whoops

text/0000-colocated-tests.md
```diff
@@ -0,0 +1,161 @@
...
+previously:
+ - `app/utils/{util-name}.js`
+ - `tests/unit/{util-name}-test.js`
+now:
+ - `app/utils/{util-name}.js`
+ - `app/utils/{util-name}.test.js`
+
+#### Acceptance Tests
+
+Acceptance tests do not have a core affiliation with any other file, so co-locating them _may_ not make sense, depending on what a particular test file is doing.
```
void-mAlex: in the interest of not leaving a class of tests alone in the tests folder which would really act to minimize their importance, I would propose a default location for acceptance tests under the `app/routes/` folder

reasoning behind that is that in those tests you start off by visiting a route then testing the behavior of it as a whole

NullVoxPopuli: is this how most acceptance tests are? I thought to leave them in `tests/acceptance/` because I haven't seen a clear pattern (not that I've seen a ton of different people's / companies' acceptance tests)

void-mAlex: that is partly why I left this comment
also judging by the guide https://guides.emberjs.com/release/tutorial/part-1/automated-testing/
that's what an acceptance test looks like:
`await visit('/')` click around assert stuff

Gaurav0:
Currently, all tests must end with `-test.js` even if they are in `tests` directory. Why wouldn't it make sense to continue to use `-test.js` instead of changing it to `.test.js`?

NullVoxPopuli:
>  Why wouldn't it make sense to continue to use -test.js instead of changing it to .test.js?

the RFC says this: https://github.com/emberjs/rfcs/pull/599/files#diff-9e4ea543242ddd0a6ef8209ac8a058e8R27

> because the only name part that could be searched for is `-test` at the end of the filename, there is the possibility that valid components, controllers, routes, etc could legitimately end with `-test`, causing files to incorrectly get removed or added to the app or test bundle.

text/0000-colocated-tests.md
```diff
@@ -0,0 +1,161 @@
+- Start Date: 2020-03-02
+- Relevant Team(s): Ember CLI
+- RFC PR: [#599](https://github.com/emberjs/rfcs/pull/599)
+- Tracking: (leave this empty)
+
+# Co-located Tests
+
+## Summary
+
+As apps grow, and need to be refactored, the ability to have tests be co-located with their concerned tested behavior enables faster refactors and better overall upkeep as related files are grouped together.
```
jrjohnson: Disagree that co-location makes development an easier so this statement will need some evidence to back it up. I very much prefer tests to be in their own space and not cluttering up the already extremely crowded components directory. The ember language server in VSCode now inserts a link to the test at the top of the file anyway making it very easy to navigate to.

text/0000-colocated-tests.md
jrjohnson: The fact that acceptance tests won't be co-located makes this proposal significantly weaker. Now there are two places to look for tests.

text/0000-colocated-tests.md
NullVoxPopuli: good point -- I should bring in some of the philosophy around module unification in here. 
where, drag-and-drop refactoring is allowed, because all folder sub-tree are totally valid, and should remain valid as you move them around your app.

NullVoxPopuli: I, and many others, don't use VS Code, so.. while tooling can solve part of the problem, it's not all of the problem! <3

text/0000-colocated-tests.md
NullVoxPopuli: I started a poll on twitter: https://twitter.com/tommyjr/status/1234563047873708032

my thoughts are -- what would they be co-located with?
you can co-locate them yourself, as all .test.js files in the app folder would be detected, but I have a hard time justifying changing the default for acceptance tests.

text/0000-colocated-tests.md
NullVoxPopuli: I've updated the Summary and Motivation sections as well as added tooling as an alternative (in the alternatives section)

text/0000-colocated-tests.md
mehulkar: collocating acceptance tests with routes would not work. We have several application level tests where the entry URL doesn't matter, we're just using it as a shell to test some complex interaction / behavior.

mehulkar:
Seems good for component and unit tests, but the complication with acceptance tests does seem to make it harder to justify. That said, it may be possible to put acceptance tests at `app/tests`, so it's "co-located" with... the app. 

Couple other concerns: 

- how does it affect the `tests/dummy` in addons
- how does this affect what we call the different test types? e.g. unit vs integration tests for components & helpers. cc #575

NullVoxPopuli:
> put acceptance tests at app/tests, so it's "co-located" with... the app.

that's an interesting idea! would we just want them in `tests`, or make the directory stand out somehow? `__tests__`, `-tests`, etc?

> how does it affect the tests/dummy in addons

ha, ya, I totally forgot about addons.
I think for simplicity, I don't want to mess the dummy app, it's already awkward, and I think we are one ore two RFCs from it being less awkward? would be nice to have actual apps instead of a dummy app, imo.

for addon, non-acceptance tests, I could see those being co-located following all the same rules outlined for the app. (just with `addon` instead of `app` as the root. During implementation, I would also need to make sure that tests in the addon aren't bundled with a consuming app 

> how does this affect what we call the different test types? e.g. unit vs integration tests for components & helpers. cc #575

integration vs unit doesn't matter today, aside from the arbitrary folders that those tests get thrown in to. 
good question though -- since those concepts wouldn't matter here, I suppose the test type doesn't matter either.
maybe the generated tests are just... tests.... that happen to use `setupTest` or `setupRenderingTest`
like,
```ts
module('Component | my-component', ...);
module('Service | my-service', ...);
module('Helper | my-helper', ...);
// etc
```

mehulkar:
> integration vs unit doesn't matter today

The blueprints are different

NullVoxPopuli:
> The blueprints are different

I meant location-wise

camerondubas:
How would this work for in-repo engines and add-ons? Currently their tests are co-located in the `tests` directory of the host app.

I work on a large app with multiple in-repo engines, and being able to have test files co-located with their relevant files would be a huge improvement to the development experience.

Regardless, I really like this RFC. Again, working in a big app makes finding test files quite cumbersome as is.

nightire:
I like the idea of test co-location in general, but the acceptance test is the biggest roadblock in front of this. Most people write acceptance tests in the feature/scenario oriented way(according to the latest survey results), it is not necessary to be bounded with routings. Thus, the acceptance tests deserve a dedicated place and it should be easy to navigate.

I suggest we don't have to be picky too much, just add another folder called `app/(application|acceptance|feature)-tests/` to serve this purpose.

As for the end of tests filename, I agree to keep `-test.js` form in order to help developers migrate easily. But it should be easy to support `.test.js` at the same time.

In-repo add-ons/engines could be another big issue here, but I don't know too much about it, I prefer to using monorepo structure these days. Though I think we don't have to put acceptance tests in the in-repo add-ons because "in-repo" implicates serving and only serving the host application, so the application should take care of the acceptance tests.

What about unit tests and rendering tests in the in-repo add-ons, does co-locate support it? I'm not sure, but it should be addressed in this RFC.

Another thing is once test co-location is settled, the `app/` folder is actually not just *app* anymore, it contains all types of source code, so it should be renamed as `src/` or other else...wow, welcome back, unifications! Anyway, this could be another RFC on another day.

Just my personal opinion, but I would like to see test co-location becomes a success.

jelhan:
> integration vs unit doesn't matter today, aside from the arbitrary folders that those tests get thrown in to.
> good question though -- since those concepts wouldn't matter here, I suppose the test type doesn't matter either.
> maybe the generated tests are just... tests.... 

Agree that we may not want to make such a big deal anymore of the difference but the RFC must include how to deal with it:

1. How to migrate tests for a component if both integration and unit tests exist? Can't move both to the same location.
2. What should happen if user tries to generate an unit test with blueprints for a component that already has an integration test?

The same question apply to template helpers and their tests. Possible strategies would be:

- Use a `-integration-test.js` and `-unit-test.js` suffix in that cases or
- skip the tests on automated migration or/and
- treat it as an attempt to replace existing test if user tries to generate an unit test for a component that already has an integration test. Ember CLI would ask if it should be overwritten as today.

NullVoxPopuli:
> How would this work for in-repo engines and add-ons? Currently their tests are co-located in the tests directory of the host app.

> What about unit tests and rendering tests in the in-repo add-ons, does co-locate support it? I'm not sure, but it should be addressed in this RFC.

Just updated the RFC to answer the questions. In short, tests *could* be _anywhere_ within the entire repo. Maybe not just scoped to `app`?.

--------------------

> but the acceptance test is the biggest roadblock in front of this. 
> I suggest we don't have to be picky too much, just add another folder called app/(application|acceptance|feature)-tests/ to serve this purpose.

The intent here (for me) is to provide the least dramatic changes while still enabling a lot of flexibility. 
If there is enough deviciveness about acceptance tests, It'd totally be cool to add a default application test location to the `.ember-cli` file.

------------------

> As for the end of tests filename, I agree to keep -test.js form in order to help developers migrate easily. But it should be easy to support .test.js at the same time.

I think this is a technical limitation, and nothing we can do anything about. supporting `-test.js` in the app directory could be a breaking change as `-test` files would be omitted from the app's bundle -- which is breaking if someone has a component or route ending in `-test`. Like, maybe `drivers-test` for a route or something.

--------------------

> How to migrate tests for a component if both integration and unit tests exist? Can't move both to the same location.

Component unit tests, imo, should be deprecated, as they rely heavily on implementation and slow maintenance, buuuut, the generator will already warn and prompt you when there is a file conflict and the migrator will need to de the same.

that said, my advice, is if you want both component integration and unit tests to be co-located, you could certainly rename things to be be `-unit-test.js` and `-integration-test.js`. maybe if the migrator can be a little more opinionated about this and create those files for you if it detects a conflict, but I don't think it's a requirement for this RFC.

> What should happen if user tries to generate an unit test with blueprints for a component that already has an integration test?
> The same question apply to template helpers and their tests.

the generator will warn in case of a conflict, and the user can decide what to do from there. the automated migrator should probably do the same.


hope this helps! I've added some unresolved questions to the bottom of the RFC, and would like people's feedback for in-repo-addons and engines

srowhani:
I like the proposal - would suit integration and unit tests quite well. Thanks for preparing this

bendemboski:
I think this proposal should address:
1. What about addons?
2. Where does `tests/test-helper.js` go?
3. Where does `tests/index.html` go?
4. Where does `tests/helpers` go?

I think the answer to (1) is pretty clear (`s/app/addon/g`), just should be called out in the RFC.

The answers to (2) and (3) could definitely be "it stays where it is," but should be addressed in the RFC.

The answer to (4) could be "it stays where it is" although then we need to talk about importing test helpers since I believe most people do a relative import that never traverses outside the `tests/` directory. A relative import would still work in this RFC, but I think it would piss off most tooling because I believe the runtime relative path from `app/foo.js` to `tests/foo.js` is actually `./tests/foo.js`, not `../tests/foo.js` as the directory structure would suggest. Maybe we advise people switch to absolute imports for test helpers since it's probably less confusing.

I would propose another option for (4), which is also what I proposed in #575, which is to move test helpers to `/test-support` (which mirrors addons), and then recommend people use absolute imports. I suppose we could argue for putting `test-helper.js` and `index.html` there as well...I'm less sure of that...

NullVoxPopuli:
@bendemboski re: #1, did you see this section: https://github.com/NullVoxPopuli/rfcs/blob/colocated-tests/text/0000-colocated-tests.md#addons-in-repo-addons-and-engines ?

re: the things that aren't tests -- I don't know if this RFC needs to address them, because this RFC is explicitly about `{-,.}test.*` files? idk. Just trying to keep scope low, ya know?

bendemboski:
@NullVoxPopuli oh yeah, I read it but I guess then forgot about it. That seems to imply that the addon tests would live in `app/` instead of `addon/`? I assuming that's not what you meant...so maybe it just needs a little clarification?

I guess not addressing (2) and (3) is valid, although I think it would be helpful to call them out as explicitly not addressed by this RFC because I think the RFC strongly suggests the question of what happens to them.

For (4), I suppose you could do the same thing, but unlike `test-helper.js` and `index.html`, the location of the test helpers has an impact on what needs to be inside the test files, so I think if we're providing an option to move test files somewhere else, we should have a story/advice/something for how they should import their helpers. Could just be suggesting absolute imports, and then we could discuss moving the helpers directory in a separate RFC, but IMO the RFC is incomplete without saying something about how to import test helpers.

NullVoxPopuli:
> That seems to imply that the addon tests would live in app/ instead of addon/

I linked to a paragraph explicitly talking about addon :P

anywho, I've updated everything to use `{app,addon}` where appropriate, so people don't think I'm talking about just one thing all the time. It's a placeholder.

NullVoxPopuli:
>  Could just be suggesting absolute imports,

I'm a fan of this. 

Now that I've had some time to think about your suggestions, I think the'd all be useful under the "how we teach this" section.

bendemboski:
> I linked to a paragraph explicitly talking about addon :P

Right, that's what I was referring to when I said `I read it but I guess then forgot about it`. This is the sentence that was confusing me:

> For normal addons, living as their own entire project, the lookup rules for apps will apply.

Now that the rules reference `app,addon` it's perfectly clear though 😄

text/0000-colocated-tests.md
```diff
@@ -0,0 +1,191 @@
...
+ - `{app,addon}/service/{service-name}.test.js`
+
+#### Controller
+
+previously:
+ - `{app,addon}/controllers/{controller-name}.js`
+ - `tests/unit/{controller-name}-test.js`
+now:
+ - `{app,addon}/controllers/{controller-name}.js`
+ - `{app,addon}/controllers/{controller-name}.test.js`
```
gossi: please provide examples for pod layout

text/0000-colocated-tests.md
```diff
@@ -0,0 +1,191 @@
...
+ - `{app,addon}/components/{component-name}/index.test.js`
+
+#### Helpers
+
+previously:
+ - `{app,addon}/helpers/{helper-name}.js`
+ - `tests/integration/helpers/{helper-name}-test.js`
+now:
+ - `{app,addon}/helpers/{helper-name}.js`
+ - `{app,addon}/helpers/{helper-name}.test.js`
```
gossi: what if I want unit _and_ rendering test?

- `{app,addon}/helpers/{helper-name}.unit.test.js`
- `{app,addon}/helpers/{helper-name}.rendering.test.js`

like this? If so, please add as examples

text/0000-colocated-tests.md
```diff
@@ -0,0 +1,191 @@
...
+ - `tests/unit/{util-name}-test.js`
+now:
+ - `{app,addon}/utils/{util-name}.js`
+ - `{app,addon}/utils/{util-name}.test.js`
+
+#### Acceptance Tests
+
+Acceptance tests do not have a core affiliation with any other file, so co-locating them _may_ not make sense, depending on what a particular test file is doing.
+
+While tests may live anywhere in the `{app,addon}` directory, the default location for acceptance tests will still be `tests/acceptance/`.
```
gossi: Engines do have a weird setup with the dummy app, so that might be the co-location in that case.
Also I'm not a big fan of the dummy app being used for docs (e.g. ec-addon-docs) and is loaded in (part of) tests... can sb free it up from the identity crisis? :D

text/0000-colocated-tests.md
NullVoxPopuli: do unit tests for helpers exist?

text/0000-colocated-tests.md
```diff
@@ -0,0 +1,191 @@
...
+While tests in `{app,addon}/` could live anywhere, these are to be the default locations for the generators to use.
+
+#### Route
+
+previously:
+ - `{app,addon}/routes/{route-name}.js`
+ - `tests/unit/{route-name}-test.js`
+now:
+ - `{app,addon}/routes/{route-name}.js`
+ - `{app,addon}/routes/{route-name}.test.js`
```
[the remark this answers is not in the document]
NullVoxPopuli: added

text/0000-colocated-tests.md
NullVoxPopuli: added

text/0000-colocated-tests.md
bendemboski: Why use different files? Why not

```javascript
module('...', function() {
  module('...', function(hooks) {
    setupTest(hooks);
    // ...
  });
  module('...', function(hooks) {
    setupRenderingTest(hooks);
    // ...
  });
});
```

bendemboski: @NullVoxPopuli I believe before RFC232, helpers were tested by importing the helper function as testing it as a Javascript function. I don't think that makes sense now that we have rendering tests, but there are probably still legacy helper unit tests out there.

NullVoxPopuli: I mean, those legacy tests will still work. because (imo), it may not be the happy path, people could either leave their files in `tests/` or manually rename?

bendemboski: @NullVoxPopuli I agree with you, was just trying to answer your direct question. In fact, if mixing unit and rendering tests, I'd either put them in different modules in the same file (see above), or just put them all in a rendering test module -- you can do "unit tests" in a rendering test environment -- just don't call `render()`!

I think the one case where tests don't mix in the same module is application and rendering tests. As rwjblue [likes to point out](https://github.com/emberjs/rfcs/pull/575#pullrequestreview-355235832), `setupTest()` and `setupApplicationTest()` do almost the same thing -- both fully boot an application. `setupRenderingTest()` is the odd one since it sets up a sandboxed rendering environment. And I have on occasion written both application and rendering tests for the same subject, but I'd argue those are pretty rare, and nested modules (or just custom test file naming) gives a clean answer there.

So IMO maaaaaaaybe there's a "how we teach this" point in here, but doesn't seem necessary since anything like this is pretty off the beaten path and also has a fine answer.

NullVoxPopuli:
@gossi 

> Generally in favor of this. FWIW, this folder structure:

Personally, if there would be multiple tests, I'd make a tests folder

NullVoxPopuli:
The need for this capability has come up a few times in discord (and at work) over the past week or two ;)

lowski:
@NullVoxPopuli what's the status of this?

NullVoxPopuli:
I think at this point, it needs to be reviewed by a core team member?

text/0000-colocated-tests.md
MichalBryxi: I'm personally on the side of co-locating acceptance test with wherever the route lives. Most of the acceptance tests I've written do have some entry URL. I think it's the "sane default" here. And IMO this does not break other people's workflows. If it does not work for your case, you can still use the old locations.

Will the generator be able to use old locations? With some flag?

Also note that since the crawling code will use `**/*.test.js` we can use any location we like.

text/0000-colocated-tests.md
```diff
@@ -0,0 +1,210 @@
...
+pods:
+ - `{pod-namespace}/{route-name}/route.js`
+ - `{pod-namespace}/{route-name}/.test.js`
```
MichalBryxi, line 86: This raises a red flag. Won't this collide with controller tests?

```
❯ ember g route foo --pods
  create app/foo/route.js
  // therefore: app/foo/.test.js
❯ ember g controller foo --pods
  create app/foo/controller.js
  // therefore: app/foo/.test.js
```

This does not seem right?

jelhan, line 86: A file starting with a dot is very likely to cause confusion as it's hidden on some operation systems. I think this should be avoided.

I get the argument that collision with legitim names for non-test code must be avoided. But I think we need to find a better solution for pod file system layout. Maybe `route.test.js` and `controller.test.js`?

text/0000-colocated-tests.md
bendemboski: The current acceptance test blueprint is pretty tied to a single route. I suppose you could then navigate to other routes, but then the auto-generated module and filename would not be very clear, etc. Whenever I create an acceptance test that isn't just/primarily testing a single route, I don't use the blueprint.

So I could definitely see adding another blueprint, e.g. `ember g route-acceptance-test my-route` vs. `ember g acceptance-test test-my-whole-app`, but there is clearly a need for a single-route-focused blueprint, and I do think it makes sense to co-locate that one.

So I think my opinion is:
1. Co-locate acceptance tests as the blueprint currently understands them
2. Maybe add another blueprint for non-route-focused blueprints that probably get generated in `tests/<somewhere>`.

text/0000-colocated-tests.md
NullVoxPopuli: I've been under the impression that controller tests are rarely used and there wouldn't be a conflict.

_but_, the controller blueprint could totally check if there is a name collision and then output `app/foo/.controller.test.js

NullVoxPopuli: > A file starting with a dot is very likely to cause confusion as it's hidden on some operation systems.

are all the top-level project files hidden? .eslintrc, .ember-cli, etc?

NullVoxPopuli:
So, here are the remaining todos of this RFC

- [ ] Find acceptable naming convention for tests while using pods
- [ ] Lay out details for how blueprints work
   - [ ] Controller unit tests 
   - [ ] Acceptance tests now co-located with routes
   - [ ] new multi-route acceptance test blueprint

Needs more discussion:
 -  [Do we also want to change how we name tests? or leave that for a different RFC?](https://github.com/emberjs/rfcs/pull/599#pullrequestreview-372318117)
     - rendering vs unit vs application as the new names (maybe "feature"? idk)

Please comment if I missed something ;)

text/0000-colocated-tests.md
MichalBryxi: - But this would not be a top-level project files?
- I think mostly those that are hidden are ... configuration files ... for libraries?

I agree that having test file as "hidden" (dot in front of the name) seems wrong.

NullVoxPopuli: aight, how about a hyphen (which is more inline with our current naming)

jelhan: > > A file starting with a dot is very likely to cause confusion as it's hidden on some operation systems.
> 
> are all the top-level project files hidden? .eslintrc, .ember-cli, etc?

Yes. At least for linux systems. They aren't shown in most file explorers nor in output of `ls` terminal command.

gnclmorais: I agree with the dot files not looking that great… I associate them with configuration and settings, not with test or any actual code written for an application. The hyphen sounds like a good compromise.

wagenet:
@NullVoxPopuli what's the status of this. I'd love to get it moving again.

NullVoxPopuli:
It's been a while, and I need to test this, but I think with some clever packager loader rules (webpack for now) using embroider strictest mode, we might be able to achieve this today.  I don't think there is anything for ember core teams to do, as this is now a packager responsibility. 

If/when I do try this again, I'll probably uncover something special about the tests tree, and I'll open a more specific issue

PR #615, opened 2020-04-18
Autotracking Memoization

pzuraq:
[Rendered](https://github.com/emberjs/rfcs/blob/autotracking-memoization/text/0615-autotracking-memoization.md)

### DIFF
--- a/text/0615-autotracking-memoization.md
+++ b/text/0615-autotracking-memoization.md
@@ -1,6 +1,6 @@
 - Start Date: 2020-04-17
 - Relevant Team(s): Ember.js
-- RFC PR: (after opening the RFC PR, update this with a link to it and update the file name)
+- RFC PR: https://github.com/emberjs/rfcs/pull/615
 - Tracking: (leave this empty)
 
 # Autotracking Memoization
--- a/text/0615-autotracking-memoization.md
+++ b/text/0615-autotracking-memoization.md
@@ -103,7 +103,7 @@ import {
 
 ### Usage
 
-`cache` receives a function, and returns an cache instance for that function.
+`cache` receives a function, and returns a cache instance for that function.
 Users can call `getValue()` on the cache instance to run the function and get
 the value of its output. The cache will then return the same value whenever
 `getValue` is called again, until one of the tracked values that was _consumed_
--- a/text/0615-autotracking-memoization.md
+++ b/text/0615-autotracking-memoization.md
@@ -9,7 +9,7 @@
 
 Provides a low-level primitive for memoizing the result of a function based on
 autotracking, allowing users to create their own reactive systems that can
-respond to changes in autotracked state
+respond to changes in autotracked state.
 
 ```js
 import { tracked } from '@glimmer/tracking';
@@ -56,7 +56,7 @@ Ember's templating layer.
 An example of where users might want to do this is the [`@cached`](https://github.com/emberjs/rfcs/pull/566)
 decorator for getters. This decorator only reruns its code when the tracked
 state it accessed during its previous computation changes. Currently, it would
-not be possible to build this decorator with public APIs.
+be quite difficult and error prone to build this decorator with public APIs.
 
 For a more involved example, we can take a look at [ember-concurrency](http://ember-concurrency.com/),
 which allows users to define async tasks. Ember Concurrency has an `observes()`
@@ -103,7 +103,7 @@ import {
 
 ### Usage
 
-`cache` receives a function, and returns a cache instance for that function.
+`createCache` receives a function, and returns a cache instance for that function.
 Users can call `getValue()` on the cache instance to run the function and get
 the value of its output. The cache will then return the same value whenever
 `getValue` is called again, until one of the tracked values that was _consumed_
@@ -182,7 +182,7 @@ getValue(counter); // 1
 
 When this happens, it often means that optimizations can be made in the code
 surrounding the computation. For instance, in the Glimmer VM, we don't emit
-updating bytecodes if we detect that a memoized function is constant, because
+updating bytecodes if we detect that a memoized function can never change, because
 it means that this piece of DOM will never update.
 
 In order to check if a memoized function is constant or not, users can use the
@@ -258,7 +258,7 @@ Some possibilities include:
 
 ### API Docs
 
-#### `cache`
+#### `createCache`
 
 Receives a function, and returns a wrapped version of it that memoizes based on
 _autotracking_. The function will only rerun whenever any tracked values used
--- a/text/0615-autotracking-memoization.md
+++ b/text/0615-autotracking-memoization.md
@@ -159,13 +159,9 @@ arguments is passed to the function, or if a different set is passed.
 
 #### Function Name
 
-The `cache` name was chose for two main reasons:
-
-1. It's a fairly verbose name, which discourages common usage. It is meant to be
-   a low-level API, and shouldn't make its way into common usage in app code.
-2. The function name begins with `memo` instead of `track`, which means that
-   auto-import completion won't show it when users type `@tracked`. This will
-   make it less likely for users to stumble upon it without any context.
+The `cache` name was chosen because this is effectively what the function
+produces, a cache function. It is unlikely to be mistaken for a decorator, since
+the tense is different, and the import path distinguishes it as a primitive.
 
 ### Constant Functions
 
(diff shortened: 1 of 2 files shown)


BryanCrotaz:
Looks good. Solves my usages of observes, for example playing a video element on a data change.

BryanCrotaz:
This should be a decorator?

pzuraq:
@BryanCrotaz no, this is meant to be a low-level API only. Decorators like `@cached` will be based on it (I need to update the text of that RFC, have updated the name already), and it would also allow us to build things like the `@use` and Resources RFC in userland for the most part.

richard-viney:
Couple minor typos in the first line of the usage section: `cache` => `createCache()`, and "an cache" => "a cache".

text/0615-autotracking-memoization.md
```diff
@@ -0,0 +1,385 @@
...
+
+## Summary
+
+Provides a low-level primitive for memoizing the result of a function based on
+autotracking, allowing users to create their own reactive systems that can
+respond to changes in autotracked state
+
+```js
+import { tracked } from '@glimmer/tracking';
+import { createCache, getValue } from '@glimmer/tracking/cache-primitive';
```
scottmessinger, line 16:
```suggestion
import { createCache, getValue } from '@glimmer/tracking/primitives/cache';
```
This makes the import consistent with the rest of the document.

rwjblue:
We discussed this in todays core team meeting, and are moving it into final comment period. Time to read through it again! 😸 👓 🤓

text/0615-autotracking-memoization.md
```diff
@@ -0,0 +1,385 @@
...
+  createCache,
+  getValue,
+  isCache,
+  isConst,
+} from '@glimmer/tracking/primitives/cache';
+```
+
+### Usage
+
+`cache` receives a function, and returns a cache instance for that function.
```
snewcomer, line 106: `createCache` receives a function ?

text/0615-autotracking-memoization.md
```diff
@@ -0,0 +1,385 @@
...
+getValue(counter); // 1
+getValue(counter); // 1
+getValue(counter); // 1
+
+// ...
+```
+
+When this happens, it often means that optimizations can be made in the code
+surrounding the computation. For instance, in the Glimmer VM, we don't emit
+updating bytecodes if we detect that a memoized function is constant, because
```
snewcomer, line 185: It may make sense to further define `is constant`.  I think it shares the same properties in meaning as the JS construct "constant", but could be also gain further meaning.  Does it also mean "steady" or "invariable"?

text/0615-autotracking-memoization.md
```diff
@@ -0,0 +1,385 @@
...
+class State {
+  @tracked value;
+}
+
+let state = new State();
+let count = 0;
+
+let counter = createCache(() => {
+  // consume the state
+  state.value;
```
snewcomer, line 122: What about consuming multiple states, both of which are tracked?  This seems unnecessary in the context you have described in this RFC, but just curious!

```js
createCache(() => {
  state.value;
  state.otherValue;
});
```

text/0615-autotracking-memoization.md
```diff
@@ -0,0 +1,385 @@
...
+
+This topic is one that is meant for advanced users and library authors. It
+should be covered in detail in the Autotracking In-Depth guide in the Ember
+guides.
+
+This guide should cover how memoization works, and various techniques for using
+memoization. It should cover a variety of ways to use memoization to accomplish
+common tasks of other reactivity systems. Pull-based reactivity is unfamiliar to
+many programmers, so we should try to familiarize them with as many common
+examples as possible.
```
snewcomer, line 248: Agreed.  Even as we migrate applications over, I'm sometimes not applying `@tracked` in a condoned way

text/0615-autotracking-memoization.md
```diff
@@ -0,0 +1,385 @@
...
+- Building the `@cached` decorator from scratch.
+- Building a data layer that syncs changes to models to localStorage or a
+  backend in real time, as the changes occur (note: requires polling of some
+  kind, or a component to do this).
+- Building a `RemoteData` implementation, a helpful wrapper that sends a fetch
+  request to a remote url and loads data whenever the url input changes.
+
+### API Docs
+
+#### `cache`
```
snewcomer, line 261: `createCache`?

text/0615-autotracking-memoization.md
snewcomer: Perhaps an additional adjective could be used in the docs section if you felt there was an improvement!

text/0615-autotracking-memoization.md
pzuraq: Right, if both of those values are tracked, and either of them were updated, then the cache would be invalidated.

text/0615-autotracking-memoization.md
```diff
@@ -0,0 +1,385 @@
...
+- RFC PR: https://github.com/emberjs/rfcs/pull/615
+- Tracking: (leave this empty)
+
+# Autotracking Memoization
+
+## Summary
+
+Provides a low-level primitive for memoizing the result of a function based on
+autotracking, allowing users to create their own reactive systems that can
+respond to changes in autotracked state
```
rwjblue, line 12:
```suggestion
respond to changes in autotracked state.
```

text/0615-autotracking-memoization.md
```diff
@@ -0,0 +1,385 @@
...
+import { createCache, getValue } from '@glimmer/tracking/primitives/cache';
+
+let count = 0;
+
+class Person {
+  @tracked firstName = 'Jen';
+  @tracked lastName = 'Weber';
+
+  #fullName = createCache(() => {
+    ++count;
```
rwjblue, line 25: Might be a bit more clear to call this `++computeCount`, what do you think?

text/0615-autotracking-memoization.md
```diff
@@ -0,0 +1,385 @@
...
+been highly successful so far in its usage and adoption. However, users today
+can only integrate with autotracking via the `@tracked` decorator, which allows
+them to _create_ tracked root state. There is no way to write code that responds
+to changes in that root state directly - the only way to do so is indirectly via
+Ember's templating layer.
+
+An example of where users might want to do this is the [`@cached`](https://github.com/emberjs/rfcs/pull/566)
+decorator for getters. This decorator only reruns its code when the tracked
+state it accessed during its previous computation changes. Currently, it would
+not be possible to build this decorator with public APIs.
```
rwjblue, line 59: > Currently, it would not be possible to build this decorator with public APIs.

Is this actually true? Couldn't you use a combination of `computed` + `dependentKeyCompat`? I agree that it doesn't _really_ matter, but it would be good to avoid the absolutist terminology if it is actually not correct.

Maybe:

```suggestion
state it accessed during its previous computation changes. Currently, it would
be quite difficult and error prone to build this decorator with public APIs.
```

text/0615-autotracking-memoization.md
rwjblue: Yep, should be updated to 
 
```suggestion
`createCache` receives a function, and returns a cache instance for that function.
```

text/0615-autotracking-memoization.md
```diff
@@ -0,0 +1,385 @@
...
+  getValue,
+  isCache,
+  isConst,
+} from '@glimmer/tracking/primitives/cache';
+```
+
+### Usage
+
+`cache` receives a function, and returns a cache instance for that function.
+Users can call `getValue()` on the cache instance to run the function and get
```
rwjblue, line 107: This seems a bit misleading, saying "can call getValue on the cache instance" infers that they can do:

```js
let cache = createCache(() => {})
cache.getValue();
```

But that is actually not what we are intending.

text/0615-autotracking-memoization.md
rwjblue:
```suggestion
updating bytecodes if we detect that a memoized function can never change, because
```

text/0615-autotracking-memoization.md
rwjblue: Yep, needs to be updated

```suggestion
#### `createCache`
```

text/0615-autotracking-memoization.md
```diff
@@ -0,0 +1,385 @@
...
+This RFC proposes four functions to be added to Ember's public API:
+
+```ts
+interface Cache<T = unknown> {}
+
+function createCache<T>(fn: () => T): Cache<T>;
+
+function getValue<T>(cache: Cache<T>): T
+
+function isCache(maybeCache: object): boolean;
```
rwjblue, line 87:
```suggestion
```

text/0615-autotracking-memoization.md
```diff
@@ -0,0 +1,385 @@
...
+The `isCache` function can be used to determine if a value is a cache, in cases
+where it's not possible to know if the value is a cache or not.
+
+```js
+let cache = createCache(() => {});
+let notACache = () => {};
+
+isCache(cache); // true
+isCache(notACache); // false
+```
```
rwjblue, line 161:
```suggestion
```

text/0615-autotracking-memoization.md
```diff
@@ -0,0 +1,385 @@
...
+import { createCache, isCache } from '@glimmer/tracking/primitives/cache';
+
+let cache = createCache(() => {});
+let notACache = () => {};
+
+
+isCache(cache); // true
+isCache(notACache); // false
+```
+
```
rwjblue, line 330:
```suggestion
```

wycats:
For historical reference:

I was momentarily worried that this RFC doesn't provide an easy way to check whether a computation is const without creating unnecessary garbage.

This problem isn't semantic: it's not a semantic problem to create a `Cache` and immediately throw it away. I was worried, at first, that this API would make it difficult to implement Glimmer's strategy of running a computation, checking if it was const, and only bothering to cache the value at all if the value wasn't const.

@pzuraq persuaded me that the current spec allows this implementation: the implementation could check to see if the computation is `const` and not bother to store the value at all if it is.

This whole thing only matters if the cost of storing the cache momentarily is a notable cost. If it turns out to be a serious issue, and @pzuraq's thinking turns out to be wrong, we could always revisit this issue.

rwjblue:
FYI - @pzuraq created a polyfill for this that works back to ember-source@3.13 (going further back is very very difficult since autotracking in general is missing).

PR #619, opened 2020-04-23
Better environment handling

mydea:
[Rendered](https://github.com/mydea/rfcs/blob/better-env-handling/text/0000-better-env-handling.md)

### DIFF
--- a/text/0000-better-env-handling.md
+++ b/text/0000-better-env-handling.md
@@ -1,6 +1,6 @@
 - Start Date: 2020-04-23
 - Relevant Team(s): Ember.js, Learning
-- RFC PR: (after opening the RFC PR, update this with a link to it and update the file name)
+- RFC PR: [#619](https://github.com/emberjs/rfcs/pull/619)
 - Tracking: (leave this empty)
 
 # Better environment handling
@@ -58,6 +58,8 @@ We could also add `@glimmer/env` to the Ember API docs, like `@glimmer/component
 We could add something like `TESTING` to `@glimmer/env`. The same concerns as mentioned above apply here too.
 
 We could also do nothing and propose users to continue using `Ember.testing` or a home-grown solution instead. However, I think that these problems are common enough to warrant a nice, built-in solution for them.
+
+We could also use a different module name, or put it into an existing module.
 
 ## Unresolved questions
 
--- a/text/0000-better-env-handling.md
+++ b/text/0000-better-env-handling.md
@@ -63,12 +63,58 @@ It is false in all other cases, like these:
 
 In addition, the `@glimmer/env` module should be added to the [Ember API docs](https://api.emberjs.com/), like `@glimmer/tracking`.
 
-Finally, we should deprecate these older functions/modules:
+### Added deprecations
+
+We deprecate this older functionality:
 
 - `Ember.testing`
 - `runInDebug` from `@ember/debug`
 
 We will deprecate them immediately, as the implementation of TESTING will work retroactively - anyone that can use ember-cli-babel@7 would be able to use it, regardless of ember-source version.
+
+#### Transition path
+
+`Ember.testing` can be replaced by `TESTING`. For example:
+
+```js
+// Before
+import Ember from "ember";
+
+function doSomething() {
+  let time = Ember.testing ? 1 : 5000;
+}
+
+// After
+import { TESTING } from "@glimmer/env";
+
+function doSomething() {
+  let time = TESTING ? 1 : 5000;
+}
+```
+
+`runInDebug()` can be replaced by `DEBUG` like this:
+
+```js
+// Before
+import { runInDebug } from "@ember/debug";
+import { myFunction } from "my-function";
+
+function doSomething() {
+  runInDebug(myFunction);
+}
+
+// After
+import { DEBUG } from "@glimmer/env";
+import { myFunction } from "my-function";
+
+function doSomething() {
+  if (DEBUG) {
+    myFunction();
+  }
+}
+```
+
+This has the added benefit that it will also actually strip the code from production builds (which doesn't happen with`runInDebug()`).
 
 ## How we teach this
 
--- a/text/0000-better-env-handling.md
+++ b/text/0000-better-env-handling.md
@@ -26,6 +26,8 @@ There are three main ways to solve this, as of now:
 
 This is not really recommended anymore. There is no module export for this (you need to use the `ember` import for it).
 
+It cannot be used in module scope at all (see [no-ember-testing-in-module-scope eslint-plugin-ember rule](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/no-ember-testing-in-module-scope.md)).
+
 2. `import config from 'my-app/config/environment` && `config.environment === 'test'`
 
 This only works inside of an app. Addons cannot use this approach, as they don't know the application name.
@@ -66,6 +68,8 @@ Finally, we should deprecate these older functions/modules:
 - `Ember.testing`
 - `runInDebug` from `@ember/debug`
 
+We will deprecate them immediately, as the implementation of TESTING will work retroactively - anyone that can use ember-cli-babel@7 would be able to use it, regardless of ember-source version.
+
 ## How we teach this
 
 In addition to adding this to the Ember API docs, a sub page should be added to the Ember.js Guides. This could either live under "Application concerns" or under "Configuration > Configuring your app".
@@ -84,4 +88,4 @@ We can also encourage users to use `import config from 'my-app/config/environmen
 
 ## Unresolved questions
 
-Should we deprecate the older functionality immediately or leave a grace period?
+-
--- a/text/0000-better-env-handling.md
+++ b/text/0000-better-env-handling.md
@@ -7,62 +7,68 @@
 
 ## Summary
 
-This RFC proposes to add a tiny module similar to `@glimmer/env`, to allow to more easily work with common environment questions.
+This RFC proposes to add a `TESTING` export to `@glimmer/env`, to allow to more easily work with common environment questions.
+Additionally, it proposes to add the `@glimmer/env` module to the [API docs](https://api.emberjs.com/), like `@glimmer/tracking`.
 
 ## Motivation
 
-This RFC tries to solve two common problems:
+Currently, there is no _one_ way to identify if the application is running in test mode.
 
-1 - **Handling test-specific code**.
-
-The old way of doing this, `Ember.testing`, is not really ideal and recommended anymore.
-
-Example use cases where you might want to use this:
+Some reasons why you would want to check for that in your code include:
 
 - Disabling polling code in tests
-- Preventing test-breaking code like `window.reload()`
+- Preventing test-breaking code like window.reload()
 - Preventing log output in tests
 
-2 - **Handling debug-specific (=non-production) code**.
+There are three main ways to solve this, as of now:
 
-You can use `import { DEBUG } from "@glimmer/env"` for this, or `import { runInDebug } from "@ember/debug"`, although the later [seems to be buggy as of now](https://github.com/emberjs/ember.js/issues/18912).
+1. `Ember.testing`
 
-An exmaple use case where you might want to use this:
+This is not really recommended anymore. There is no module export for this (you need to use the `ember` import for it).
 
-- Auto-signing in a mocked user in development environment
+2. `import config from 'my-app/config/environment` && `config.environment === 'test'`
 
-This RFC proposes to provide a more more streamlined, approved and documented way to handle these cases, by introducing an `@ember/env` namespace which can be used like this:
+This only works inside of an app. Addons cannot use this approach, as they don't know the application name.
 
-```js
-import { DEBUG, TESTING } from "@ember/env";
-```
+3. `getOwner(this).resolveRegistration('config:environment').environment === 'test'`
+
+This requires access to the container, and is also a rather complicated syntax for such a thing.
+It might not be immediately clear what exactly is happening here,
+and the `resolveRegistration()` syntax is probably not commonly used by many developers.
+
+This RFC wants to solve this by providing a single, unified & community-approved solution to this problem.
+
+It also wants to provide an easy solution to test production builds of an application.
 
 ## Detailed design
 
 ```js
-import { DEBUG } from "@ember/env";
+import { TESTING } from "@glimmer/env";
 ```
 
-This works exactly as `import { DEBUG } from "@glimmer/env"`.
-The reasoning for moving this into the `@ember` namespace is that it is not clear why this should be imported from glimmer, which is the rendering engine.
+This is a constant (does not change / live update without a refresh).
 
-```js
-import { TESTING } from "@ember/env";
-```
+It is true in the following cases:
 
-This works like `Ember.testing`.
+- `ember test`
+- `ember test --environment=production`
+- `ember serve` & `http://localhost:4200/tests`
 
-Together with this, we should deprecate (or plan to deprecate) the older functions:
+It is false in all other cases, like these:
+
+- `ember serve` & any other path
+- `ember build --environment=production`
+
+In addition, the `@glimmer/env` module should be added to the [Ember API docs](https://api.emberjs.com/), like `@glimmer/tracking`.
+
+Finally, we should deprecate these older functions/modules:
 
 - `Ember.testing`
 - `runInDebug` from `@ember/debug`
-- `DEBUG` from `@glimmer/env`
 
 ## How we teach this
 
-This probably doesn't need to be in the guides, just in the API docs for developers looking for this.
-
-As the list of alternatives below shows, there are quite a few ways to _kind of_ achieve this outcome right now, but none of them are really documented right now. In the spirit of Ember, it would be helpful to converge on a single, officially documented & recommended way, which will also make teaching this much easier (as there will be a single point to direct developers to, when they run into these issues).
+In addition to adding this to the Ember API docs, a sub page should be added to the Ember.js Guides. This could either live under "Application concerns" or under "Configuration > Configuring your app".
 
 ## Drawbacks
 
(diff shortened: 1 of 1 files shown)


mehulkar:
I would like to have this also, but I also feel that environment-specific code tends to hint at other problems. E.g. a lot of `Ember.testing` branch code tends to be a last resort when animations or whatever cause tests to fail. The real problem is a test suite that doesn't accurately reflect the user experience, and I'd want to fix that. Dev specific code, in my experience is usually to configuration that can be done in config/environment.js or logging that can be handled by the existing `debug` function in `@ember/debug`. Maybe more pain points would be useful to see if there are other alternative solutions?

Gaurav0:
@mydea I believe the currently recommended solution for determining if we are in testing is to use:

```
import config from 'my-application/config/environment';

// config contains the contents of ENV from your config/environment.js

const { environment } = config;
const isTesting = environment === 'test';
```

mydea:
> @mydea I believe the currently recommended solution for determining if we are in testing is to use:
> 
> ```
> import config from 'my-application/config/environment';
> 
> // config contains the contents of ENV from your config/environment.js
> 
> const { environment } = config;
> const isTesting = environment === 'test';
> ```

Yeah, there are definitely ways, but it's not ideal, not necessarily easy to find for new developers, and harder to use e.g. from addons - there are also ways there, like `getOwner(this).resolveRegistration('config:environment')`, but it still feels like there should be an easier way to do this. I will add this to the text as alternatives as well.

mydea:
> I would like to have this also, but I also feel that environment-specific code tends to hint at other problems. E.g. a lot of `Ember.testing` branch code tends to be a last resort when animations or whatever cause tests to fail. The real problem is a test suite that doesn't accurately reflect the user experience, and I'd want to fix that. Dev specific code, in my experience is usually to configuration that can be done in config/environment.js or logging that can be handled by the existing `debug` function in `@ember/debug`. Maybe more pain points would be useful to see if there are other alternative solutions?

I will add some example use cases I personally came across to the motivation. Some I could think of are:

`TESTING`

* Disabling polling code in tests
* Preventing test-breaking code like `window.reload()`
* Preventing log output in tests

`DEBUG`

* Auto-signing in a mocked user in development environment

Overall, I am definitely positive that for all of these use cases, it is possible to find a decent workaround. However, I feel that a more agreed upon and documented solution for these _is_ missing from the public API right now. I also think this is a need that most developers will stumble over at some point. Since I just searched for this in the last days I can attest to the fact that it is, as of now, not easily possible to find the "best"/recommended solution to these problems easily right now. I spent quite some time going through the API docs to search if, by chance, anything like that has been added (in the process, stumbling over the not-really-working anymore `runInDebug()`). There are multiple ways to _somehow_ achieve these outcomes, but not really _one_ well defined way.

sandstrom:
I also think what @Gaurav0 says makes sense.

If something would be added to Ember itself, perhaps there are other things we'd like to add too? For example the Ember version, or other things that are "locked in" at build-time. And have some shared name for that, e.g. `import { version, environment, …etc } from '@ember/static'` or whatever we'd like to call that module.

Also, why would you want to export both `DEBUG` and `TESTING`, wouldn't it be better to export the environment and then check `env === 'development'`?

mehulkar:
@mydea the polling and `window.reload` examples are great. I really think that there should be better fixes for the root causes here. E.g. disabling polling should really happen by being able to register/inject dummy code (this is mostly possible today, but with caveats https://github.com/ember-learn/guides-source/issues/1262). The reload example would be fixed by a test runner that runs in an iframe rather than in a different root div. I've had both of these problems and would love a *real* fix that moves ember application tests closer to selenium-like.

The dev example for logging in as a mocked user is also good, and I think that environment specific configuration for addons is the right approach in the long term.

Anyway, I totally agree that sooomething would be good, and this is a low enough level of effort to be justifiable. I'd just be worried about encouraging the antipattern of env specific code

simonihmig:
> Anyway, I totally agree that sooomething would be good, and this is a low enough level of effort to be justifiable. I'd just be worried about encouraging the antipattern of env specific code

I agree with that, both with the concern, but also that we need something better than `Ember.testing` as an escape hatch. It has some [considerable use today](https://emberobserver.com/code-search?codeQuery=Ember.testing), but no proper import path suggested by [RFC176](https://github.com/emberjs/rfcs/blob/master/text/0176-javascript-module-api.md), so should be addressed *somehow*.

> Preventing test-breaking code like window.reload()

Shameless plug: https://github.com/kaliber5/ember-window-mock 😏

mydea:
I basically agree with everything said here, for many things there are (or there should be) ways to handle them more gracefully, however there still needs to be an escape hatch (and be it just for addons to be able to provide these better abstractions).

And even for occurrences where using such flags might not be the ideal solution, it is probably still better to use an "approved" and clear solution than everybody using their own custom way to kind of get the same outcome.

text/0000-better-env-handling.md
```diff
@@ -0,0 +1,89 @@
...
+```
+
+## Detailed design
+
+```js
+import { DEBUG } from "@ember/env";
+```
+
+This works exactly as `import { DEBUG } from "@glimmer/env"`.
+The reasoning for moving this into the `@ember` namespace is that it is not clear why this should be imported from glimmer, which is the rendering engine.
```
rwjblue, line 47: `@glimmer/*` is not just about the rendering engine though we haven't _really_ done a great job of explaining how it fits into things. For example `@glimmer/tracking` is not really fundamentally tied to rendering.

text/0000-better-env-handling.md
```diff
@@ -0,0 +1,89 @@
...
+
+```js
+import { TESTING } from "@ember/env";
+```
+
+This works like `Ember.testing`.
```
rwjblue, line 53: I actually think this is a bit off from a design perspective:

* `Ember.testing` **cannot** be checked in module scope, it is only set when starting an individual test and unset when the test is completed.
* We want this value to mean something more akin to "this is a testing enabled build"

In other words, we do **not** want this to be the same as `Ember.testing` we want it to be:

* constant (does not change / live update without a refresh)
* `true` for all of the following:
  * `ember test`
  * `ember test --environment=production`
  * `ember serve` + `/tests/`
* `false` for all of the following:
  * `ember serve` + any non `/tests/` path
  * `ember build --environment=production` (non-testing related production build)

text/0000-better-env-handling.md
```diff
@@ -0,0 +1,89 @@
...
+
+Together with this, we should deprecate (or plan to deprecate) the older functions:
+
+- `Ember.testing`
+- `runInDebug` from `@ember/debug`
+- `DEBUG` from `@glimmer/env`
+
+## How we teach this
+
+This probably doesn't need to be in the guides, just in the API docs for developers looking for this.
```
rwjblue, line 63: I think there is a place for this in the application concerns section of the main guides, and in the addon authoring guides in cli.emberjs.com.

text/0000-better-env-handling.md
```diff
@@ -0,0 +1,89 @@
...
+- RFC PR: [#619](https://github.com/emberjs/rfcs/pull/619)
+- Tracking: (leave this empty)
+
+# Better environment handling
+
+## Summary
+
+This RFC proposes to add a tiny module similar to `@glimmer/env`, to allow to more easily work with common environment questions.
+
+## Motivation
```
rwjblue, line 12: IMHO, this misses a very important issue / motivation:

Using `import { DEBUG } from '@glimmer/env'` is **inappropriate** for testing related guards because it means you _can not_ run your tests against production builds. I am 100% to blame for this issue (I added the `@glimmer/env` functionality, and kicked off a bunch of PRs to use `DEBUG` to guard testing related code) :sob:.

text/0000-better-env-handling.md
```diff
@@ -0,0 +1,89 @@
...
+
+```js
+import { DEBUG, TESTING } from "@ember/env";
+```
+
+## Detailed design
+
+```js
+import { DEBUG } from "@ember/env";
+```
```
rwjblue, line 44: I'd much prefer to keep with `@glimmer/env` and teach it better, than introduce another new thing. It is my opinion (personal, not an official core team position just yet) that we _really do_ want to move forward with the mental model that Ember itself is layered on top of `@glimmer/*` (very much like the "npm install your way to Ember" terminology from waaaaaay back when Glimmer.js was originally introduced), and we are actually kinda close to that these days.

If we do that, this change becomes quite a bit smaller as well (adding a `TESTING` named export to `@glimmer/env`).

text/0000-better-env-handling.md
Gaurav0: What is the difference between this and `environment === 'test'` ?

rwjblue: Not much from an applications perspective, but addons cannot use `import config from '../config/environment'` (because they have no clue what the app name they are consumed is is, and really shouldn't _have_ to care).

text/0000-better-env-handling.md
mydea: I will add this then as a suggestion!

text/0000-better-env-handling.md
mydea: I am also fine with that! I will update the RFC accordingly, which will then suggest adding a `TESTING` export from `@glimmer/env`, plus this module to the API docs.

text/0000-better-env-handling.md
mydea: Yeah, basically it is:

1. `import config from 'my-app/config/environment` only works in host apps
2. `getOwner(this).resolveRegistration('config:environment')` only works with access to the container
3. Both of these are IMHO less explicit, also I always need to lookup the syntax for 2. as that is the only place I ever need it

text/0000-better-env-handling.md
mydea: Will add this as well!

text/0000-better-env-handling.md
mydea: :+1: will update this accordingly!

mydea:
I have updated the RFC content based on @rwjblue's feedback. It now proposes to just add an `TESTING` export to `@glimmer/env`, as well as adding the module to the API docs.

text/0000-better-env-handling.md
Gaurav0: We already have the `ember-get-config` addon.

mydea: Yes. However, 

a) it feels weird to require an addon to solve this.
b) there is current `Ember.testing` which is much more convenient. However, it is not encouraged to use it anymore, without having a clear replacement.
c) Having a clear import for this is also less error prone & clearer in intent (e.g. consider `config.environment === 'test'` vs. `config.environment === 'testing'` - oops). It is easier to find and can be auto-completed by an editor.

text/0000-better-env-handling.md
```diff
@@ -0,0 +1,87 @@
...
+
+- Disabling polling code in tests
+- Preventing test-breaking code like window.reload()
+- Preventing log output in tests
+
+There are three main ways to solve this, as of now:
+
+1. `Ember.testing`
+
+This is not really recommended anymore. There is no module export for this (you need to use the `ember` import for it).
```
rwjblue, line 27: We should also mention that it cannot be used in module scope at all ([see `no-ember-testing-in-module-scope` eslint-plugin-ember](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/no-ember-testing-in-module-scope.md) rule).

text/0000-better-env-handling.md
```diff
@@ -0,0 +1,87 @@
...
+
+We could also do nothing and propose users to continue using `Ember.testing` or a home-grown solution instead. However, I think that these problems are common enough to warrant a nice, built-in solution for them.
+
+We can also encourage users to use e.g. `getOwner(this).resolveRegistration('config:environment').environment === 'test'`, however this will not work in plain classes without Owner access.
+
+We can also encourage users to use `import config from 'my-app/config/environment'`, however that does not work in addons.
+
+## Unresolved questions
+
+Should we deprecate the older functionality immediately or leave a grace period?
```
rwjblue, line 87: I think deprecation immediately is probably fine, since the implementation of `TESTING` will work retroactively (basically anyone that can use ember-cli-babel@7 would be able to use it, regardless of ember-source version).

text/0000-better-env-handling.md
rwjblue: > We already have the ember-get-config addon.

Sure, and that might be fine when you actually _need_ the configuration, but determining when you should or should not include some test support code should be a built in concept. 

Additionally, it is pretty important for these code paths to be stripped when `TESTING` is **not** true (e.g. `ember build --environment="production"`), that is not possible with `ember-get-config`.

text/0000-better-env-handling.md
mydea: I have updated this accordingly.

text/0000-better-env-handling.md
mydea: I have added this.

text/0000-better-env-handling.md
```diff
@@ -0,0 +1,91 @@
...
+
+- `ember serve` & any other path
+- `ember build --environment=production`
+
+In addition, the `@glimmer/env` module should be added to the [Ember API docs](https://api.emberjs.com/), like `@glimmer/tracking`.
+
+Finally, we should deprecate these older functions/modules:
+
+- `Ember.testing`
+- `runInDebug` from `@ember/debug`
```
Gaurav0, line 69: Please specifically mention these deprecations in the summary.

Please add a `Transition Path` section for these deprecations. It should include enough information to write a deprecation guide pull request to https://github.com/ember-learn/deprecation-app.

It should specifically include the code each of these is refactored to (I know this is obvious, but still should be included for the deprecation guide.)

rwjblue: > Please specifically mention these deprecations in the summary.

Deprecations aren't really a part of the summary for RFCs that are not explicitly deprecations RFCs. In this case, I believe they belong here in the detailed design.

> Please add a Transition Path section for these deprecations. It should include enough information to write a deprecation guide pull request to ember-learn/deprecation-app.
> 
> It should specifically include the code each of these is refactored to (I know this is obvious, but still should be included for the deprecation guide.)
> 

Ya, good call out.

mydea: I added a transition path for the deprecated functionality!

CvX: `Ember.testing` is currently a wrapper for `isTesting()` and `setTesting()` from `@ember/debug`. Should those functions/exports also be affected?

mydea: These are not public? They don't even show up as private here: https://api.emberjs.com/ember/release/classes/@ember%2Fdebug?show=inherited%2Cprotected%2Cprivate

rwjblue: @CvX - I don't think we have module exports for those, that is just how Ember internally is organized. The deprecation here would be upon touching `Ember.testing` which means it would go into the `getTesting` method.

ef4:
We now have [a merged RFC](https://github.com/emberjs/rfcs/blob/master/text/0507-embroider-v2-package-format.md) that contains a macro system that overlaps with the main idea in this proposal.

Environment checking really wants to compose with the macro system so you can strip out code per environment.

Plus the macro system is already implemented. Adding a new macro named like `environment` that evaluates to `"development"`, `"test"`, or `"production"` is a trivial extension to what we already have, and then it gives you everything in this RFC, plus build-time code-stripping.

So my suggestion is to reword this RFC so adds a new `environment` macro. Usage would look like:

```js
import { environment } from "@ember/macros";
if (environment() === "test") {
  // ...
}
```

The type signature of environment would be:

```ts
function environment(): "development" | "test" | "production";
```

(Please note these three *really are* the **only** environments Ember has or has ever had. People sometimes think they can make their own extra environments, but that isn't really true, what they really want are separate ember-cli-deploy targets, but every target falls into one of these three environments.)

You could use `environment()` anywhere and get the actual value, or you can use it with the other macros to get guaranteed-build-time branch elimination:

```js
import { macroCondition, environment } from '@ember/macros';
if (macroCondition(environment() === 'test')) {
  // ...
}
```

ef4:
After getting more detail from @rwjblue, my previous comment didn't account for the fact that this is trying to reduce the number of environments from three to two. That's great too, we can detangle "test vs not-test" from "dev vs prod" as separate questions, but I still advocate asking those questions via `@ember/macros`.

rwjblue:
Ya, agree in general @ef4. I do also want to keep in mind (since I have to do a bit of work to support this) that we will want these same features in Glimmer.js / GlimmerX style apps, but that is a bit orthogonal.

ef4:
I would be fine with supporting the ember macros in glimmer apps too. They really don’t need much to integrate with a new build system (that is part of what they’re for).

mydea:
> (Please note these three _really are_ the **only** environments Ember has or has ever had. People sometimes think they can make their own extra environments, but that isn't really true, what they really want are separate ember-cli-deploy targets, but every target falls into one of these three environments.)

I mean, I get the point that this is how it _should_ be, but I've worked on apps that actually _had_ more environments, and it all worked just fine (with some adjustments in the ember-cli-build.js). So I am not sure if I would agree with that statement - it is perfectly possible to set any other environment. It might make sense to _actually_ deprecate that at some point (allowing to use a different environment), but I guess that's a different issue.

Personally, I don't really care if we import this from `@ember/macros`or from `@glimmer/debug` - I will defer to you as you have way more insight into this than I have. However, I personally would definitely still prefer to have boolean exports/functions for this, than to have to do string comparisons. But I guess we could also have e.g.:

```js
import { isDebug, isTesting } from '@ember/macros`
```

mydea:
Sidenote, I just ran into this case when trying to use `@embroider/macros` to do some checks.

Basically, setting a macro value for tests will not work like this:

```js
module.exports = function (defaults) {
  let env = process.env.EMBER_ENV;

  let app = new EmberApp(defaults, {
    '@embroider/macros': {
      setOwnConfig: {
        mockApi: env === 'test'
      },
    },
}
```

As this will be false when running `ember s` and visiting `localhost:4200/tests`. 
So in order for us to be able to strip/include stuff for the test environment with the macro system, the changes proposed here are necessary.

ef4:
@mydea your example is nearly enough to make that scenario work. What's missing is that, when you run `ember s` the macros are left un-evaluated until runtime, so there's actually another opportunity to set the testing flag to true right as the test suite is about to start.

I haven't opened this up as a public API yet because I was letting it get stabilized first, but we already do this for `fastboot.isRunning` here:

https://github.com/embroider-build/embroider/blob/2a9802c8a99041218ac54b4d43fd82de8b61625c/packages/core/src/app.ts#L716-L720

We will do the same for an `isTesting` flag.

mydea:
So, if this is being implemented in @embroider/macros, do we need an/this RFC? Or how does that work then?

mydea:
I just saw that this is now available in the latest release of `@embroider/macros`. Usage:

```js
import { isTesting, isDevelopingApp, macroCondition } from '@embroider/macros';

function myFunction() {
  if (macroCondition(isTesting())) {
    // do something test specific
  }

  if (macroCondition(isDevelopingApp())) {
    // do something development specific
  }
}
```

This works nicely, as far as I can see, and is properly tree-shaked. So I guess all that remains is potentially some documentation, but that affects all of `@embroider/macros` I guess..!

Should I close this RFC, or leave it open until the question of documentation is resolved?

mydea:
I think we can close this, as this can be pretty nicely accomplished now using `@embroider/macros`:

```js
import { isTesting, isDevelopingApp } from '@embroider/macros';
```

Works like a charm in our app, and can properly tree-shake code when used with `macroCondition` :+1:

PR #620, opened 2020-04-29
Revise RFC #389 with details about its argument

chancancode:
This PR revises RFC #389 with some additional details about its argument handling based on implementation experience in https://github.com/tildeio/ember-element-helper

### DIFF
--- a/text/0389-dynamic-tag-names.md
+++ b/text/0389-dynamic-tag-names.md
@@ -45,8 +45,8 @@ With the `element` helper proposed in this RFC, this can be accomplished with so
 
 We propose to add a new `element` helper that takes a single positional argument.
 
-* When passed a non-empty string it will generates a contextual component that, when invoked, renders an element with the same tag name as the passed string, along with the passed attributes (if any), modifiers (if any) and yields to given block (if any).
-* When passed an empty string, it will generates a contextual compoment that, when invoked, yields to the given block without wrapping it an element and ignores any passed modifiers and attributes.
+* When passed a non-empty string it generates a contextual component that, when invoked, renders an element with the same tag name as the passed string, along with the passed attributes (if any), modifiers (if any) and yields to given block (if any).
+* When passed an empty string, it generates a contextual compoment that, when invoked, yields to the given block without wrapping it an element and ignores any passed modifiers and attributes.
 * When passed `null` or `undefined`, it will return `null`.
 * When passed any other values (e.g. a boolean or a number), it will result in a development mode assertion.
 
--- a/text/0389-dynamic-tag-names.md
+++ b/text/0389-dynamic-tag-names.md
@@ -43,14 +43,29 @@ With the `element` helper proposed in this RFC, this can be accomplished with so
 
 ## Detailed design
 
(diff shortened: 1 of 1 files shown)


chancancode:
cc @cibernox

text/0389-dynamic-tag-names.md
```diff
@@ -43,14 +43,29 @@ With the `element` helper proposed in this RFC, this can be accomplished with so
...
+* When passed an empty string, it will generates a contextual compoment that, when invoked, yields to the given block without wrapping it an element and ignores any passed modifiers and attributes.
+* When passed `null` or `undefined`, it will return `null`.
+* When passed any other values (e.g. a boolean or a number), it will result in a development mode assertion.
 
 Example:
 
 ```hbs
 {{#let (element @htmlTag) as |Tag|}}
-  <Tag>...</Tag>
+  <Tag class="my-element" {{on "click" this.tagClicked}}>Hello</Tag>
```
chancancode, line 57: I'm aware this is probably not the most sensible (and accessible) example, but it's really just the simplest example to show that "modifiers get passed through" without too much bloat. Happy to change it to something else though.

text/0389-dynamic-tag-names.md
```diff
@@ -43,14 +43,29 @@ With the `element` helper proposed in this RFC, this can be accomplished with so
 
 ## Detailed design
 
-We propose to add a new `element` helper that takes a tag name and generates a contextual component that, when invoked, renders the selected element.
+We propose to add a new `element` helper that takes a single positional argument.
+
+* When passed a non-empty string it generates a contextual component that, when invoked, renders an element with the same tag name as the passed string, along with the passed attributes (if any), modifiers (if any) and yields to given block (if any).
+* When passed an empty string, it generates a contextual compoment that, when invoked, yields to the given block without wrapping it an element and ignores any passed modifiers and attributes.
```
chancancode, line 49: Despite unspecified, this is how the proof-of-concept addon already behaves today. It's largely just a happy accident due to implementation uses an `Ember.Component` under-the-hood, but there are tests for this behavior and seems like a useful feature in general.

text/0389-dynamic-tag-names.md
```diff
@@ -43,14 +43,29 @@ With the `element` helper proposed in this RFC, this can be accomplished with so
...
+* When passed `null` or `undefined`, it will return `null`.
```
chancancode, line 50: The proof-of-concept addon currently throws here. The motivation for this is to mostly match the `(component)` helper's behavior with this, and acknowledge that there are useful patterns arising from this. Note that when `null` is "invoked" it's a no-op, which is why this works for components.

chancancode:
These changes are now available in ember-element-helper v0.3.1!

nightire:
What if I want to yield a dynamic element with pre-bound attributes?

```hbs
{{yield (element (or @htmlTag "div") class="foo bar" (on "click" this.onClick))}}
```

Is the above snippet works?

Or maybe like this?

```hbs
{{#let (element (or @htmlTag "div")) as |Tag|}}
  {{yield (component Tag class="foo bar" (on "click" this.onClick))}}
{{/let}}
```

PR #625, opened 2020-05-06
Helper Managers

pzuraq:
[Rendered](https://github.com/emberjs/rfcs/blob/helper-managers/text/0625-helper-managers.md)


text/0625-helper-managers.md
```diff
@@ -0,0 +1,504 @@
...
+instance, a timing to call the effect using `requestIdleCallback`, when the
+browser has finished rendering and handling higher priority work, could be
+added. However, this is out of scope for this RFC.
+
+## How we teach this
+
+Helper managers are a low-level construct that is generally only meant to be
+used by experts and addon authors. As such, it will only be taught through API
+documentation. In addition, for precision and clarity, the API docs will include
+snippets of TypeScript interfaces where appropriate.
```
chriskrycho, line 289: This is good, but I think probably somewhat incomplete. These should certainly be covered in <abbr title="application programming interface">API</abbr> docs, but <abbr title="application programming interface">API</abbr> docs are insufficient for actually building the appropriate mental model for consumers. (Both the CLI and Data docs here are prime examples of this problem historically for Ember! Even when they had full coverage of the API surface, it has been difficult to *learn* or *master*.)

[Effective documentation](https://documentation.divio.com/introduction/) includes not only <i>reference materials</i> but also <i>tutorials</i> (introducing the , <i>how-to-guides</i> (like “cookbooks”), and <i>explanations</i>. This RFC itself would suffice for the *explanation*, and <abbr title="application programming interface">API</abbr> docs will handle <i>reference materials</i>, but I think we should also identify what appropriate kinds of <i>tutorial</i> and <i>how-to</i> material are appropriate here—presumably in an <b>Advanced Guides</b> section for addon authors specifically.

pzuraq: The issue here is this is not even a "standard" advanced Ember topic. I would say the current advanced section (In-Depth Topics) is currently stuff that most intermediate and senior Ember devs will eventually need to know or learn. Ideally, _most_ Ember apps will never include a custom manager directly, and most senior Ember devs will not _have_ to write one. 

I think a guide would be useful, and I'd be happy to write one for managers in general, but I'm not sure where that would live. Maybe "Advanced Topics" would be good to add to the guides, we've shied away from the term because we didn't want to discourage beginners who are curious (thus the name "in-depth") but this would be a time when that would make sense.

locks: Just wanted to pitch in that this is something the learning team has considered previously and recently discussed at one of the meeting. If there's interest enough interest to kick-off an RFC…

chriskrycho: I’d be happy to co-author, @locks! I actually mentioned as much to @pzuraq in a DM very recently!

text/0625-helper-managers.md
```diff
@@ -0,0 +1,504 @@
...
+users to design high-level helper APIs, without Ember needing to worry about the
+details. High-level APIs can be experimented with and iterated on while the
+core of Ember helpers remains stable, and new APIs can be introduced gradually
+over time to existing code bases.
+
+`setHelperManager` receives two arguments:
+
+1. A factory function, which receives the `owner` and returns an instance of a
+   helper manager.
+2. The object to associate the factory function with.
```
rwjblue, line 314: Would you mind making it clear that `function` is also allowed? I know the TS interface makes it clear but it is easy to miss, and doesn't seem to hurt to say "function or object" here.

text/0625-helper-managers.md
```diff
@@ -0,0 +1,504 @@
...
+
+The first time the object is looked up, the factory function will be called to
+create the helper manager. It will be cached, and in subsequent lookups it the
```
rwjblue, line 317:
```suggestion
create the helper manager. It will be cached, and in subsequent lookups the
```

text/0625-helper-managers.md
```diff
@@ -0,0 +1,504 @@
...
+cached helper manager will be used instead.
+
+Only one helper manager exists per helper factory, so many helpers will end up
```
rwjblue, line 320: > per helper factory

I don't think the term "helper factory" is actually defined here (this is the only reference in the whole document). Would helper definition work here?

text/0625-helper-managers.md
rwjblue: I think you should also tweak this to say "helper definition" to make it a easier association with the TS interfaces you drew up above.

```suggestion
2. A helper definition, which is the object or function to associate the factory function with.
```

text/0625-helper-managers.md
```diff
@@ -0,0 +1,504 @@
...
+
+The capabilities property should be provided using the `capabilities()` function
+imported from the same module as `setHelperManager`:
+
+```js
+import { capabilities } from '@ember/helper';
+
+class MyHelperManager {
+  capabilities = capabilities();
+}
```
rwjblue, line 350: This is actually not valid:

* doesn't have either `getValue` or `runEffect` methods
* does not pass the required version argument
* does pass options to `capabilities` (which even though typed as optional, it is not actually optional because you must specify either `hasValue` or `isScheduledEffect`)

```suggestion
class MyHelperManager {
  capabilities = capabilities('3.21.0', { hasValue: true });
  
  // ...snip...
}
```

text/0625-helper-managers.md
```diff
@@ -0,0 +1,504 @@
...
+The first argument to capabilities is a version string, which is the version of
+Ember that the capabilities were defined in. Ember can add new versions at any
+time, and these may have entirely different behaviors, but it will not remove
+old versions until the next major version.
+
+```js
+capabilities('3.x');
+```
+
+The second argument is an optional object of capabilities and boolean values
```
rwjblue, line 374: I do not actually think this is optional. As defined above, you must either `hasValue` or `isScheduledEffect` to actually do anything.

text/0625-helper-managers.md
```diff
@@ -0,0 +1,504 @@
...
+indicating whether they are enabled or disabled.
+
+```js
+capabilities('3.x', {
+  hasValue: true,
+  hasDestructor: true,
+});
+```
+
+If no value is specified, then the default value will be used.
```
rwjblue, line 384: What is the default value? That should be clearly defined...

text/0625-helper-managers.md
```diff
@@ -0,0 +1,504 @@
...
+- The `isScheduledEffect` capability could be broken out into a separate RFC.
+  It is mostly separable, except for the impact it has on `getValue`. Value-less
+  and effect-less helpers don't really make sense, so in isolation `getValue`
+  would probably not be an optional hook, and the `hasValue` capability wouldn't
+  exist.
+
+  Capabilities can change from version to version, so this is still not a major
+  issue, but it seems like it would be easier to add from the get go.
+
+## Appendix
```
rwjblue, line 415: I mentioned this above also, would you mind adding a proposed implementation of the `{{title}}` helper from `ember-page-title` as an example of an effecting helper?

text/0625-helper-managers.md
```diff
@@ -0,0 +1,504 @@
...
+This RFC proposes adding the `setHelperManager` and `capabilities` APIs,
+imported from `@ember/helper`. Like `setComponentManager` and
+`setModifierManager`, `setHelperManager` receives a callback that is passed
+the owner, and returns an instance of the helper manager. When a helper
+definition is resolved by Ember, it will look up the manager recursively on the
+definition's prototype chain until it finds a helper manager. If it does not
+find one, it will throw an error.
+
+```ts
+type HelperDefinition = object;
```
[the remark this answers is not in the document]
chriskrycho: You can thank JS for this one. It’s wild, but because you can do this—

```js
function foo() {
  return 'JavaScript is wild';
}
foo.bar = true;
```

—what else is TS going to do? :sob:

text/0625-helper-managers.md
```diff
@@ -0,0 +1,504 @@
...
+
+  getDestroyable?(bucket: HelperStateBucket): object;
+}
+```
+
+Let's dig into these hooks one by one:
+
+### Hooks
+
+#### `createHelper`
```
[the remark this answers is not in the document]
pzuraq: The issue with autotracking the creation hook is that it would break precedent, and we don't really have a clear meaning for it. We could call `getValue` or `runEffect` the next time that dirties, but it is already effectively "dirty" because we _will_ call those hooks anyways, and we don't want to teardown and recreate the helper because then we're taking control away from the user (they can no longer control their own tracking semantics).

Instead, users can use the `cache` API to wrap any code that they want to autotracking in this hook, and they can do anything they want with that information. They could return a cached value in the first `getValue`, they could destroy and recreate the instance of the helper, etc.

Really, `createHelper` just means "you exist in the template now", IMO. We shouldn't define any semantics for it beyond that.

text/0625-helper-managers.md
```diff
@@ -0,0 +1,504 @@
...
+interface HelperCapabilitiesOptions {
+  hasValue?: boolean;
+  hasDestroyable?: boolean;
+  isScheduledEffect?: boolean;
+}
+
+type HelperCapabilities = Opaque;
+
+export declare function capabilities(
+  version: string,
```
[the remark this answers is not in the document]
chriskrycho: That's exactly what we did in #585, FWIW.

text/0625-helper-managers.md
```diff
@@ -0,0 +1,504 @@
...
+`capabilities` function for helper managers receives a version string as the
+first parameter and a options object as the second with optional flags as
+booleans. It produces an opaque `HelperCapabilities` object, which is assigned
+to the helper manager.
+
+```ts
+interface HelperCapabilitiesOptions {
+  hasValue?: boolean;
+  hasDestroyable?: boolean;
+  isScheduledEffect?: boolean;
```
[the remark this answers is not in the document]
pzuraq: I was thinking they would default to false, and we would assert on invalid input, forcing the user to make a conscious choice. That way we wouldn't need complex defaulting logic (e.g. if we default `hasValue = true`, and they pass `isScheduledEffect = true`, then we need to determine if they gave us invalid input or if it's the default value and change it. Better for the user to decide here.)

I'll add language to make that explicit.

text/0625-helper-managers.md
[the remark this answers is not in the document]
rwjblue: FWIW, that doesn't seem very bizarre to me 😸, and in fact `function() {} instanceof Object` is true (which also supports what TS is doing). But `typeof function() {}` is **not** `object` so `typeof` and `instanceof` disagree a bit.

Either way, IMHO, it doesn't _hurt_ to say `object | Function` as far as I can tell.

text/0625-helper-managers.md
```diff
@@ -0,0 +1,504 @@
...
+### Capabilities
+
+There are three proposed capabilities for helper managers:
+
+* `hasDestroyable`
+* `isScheduledEffect`
+* `hasValue`
+
+Out of these capabilities, one of `isScheduledEffect` or `hasValue` _must_ be
+enabled. The other must _not_ be enabled, meaning they are mutually exclusive.
```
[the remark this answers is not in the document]
chriskrycho: Related, perhaps we can explicitly specify the type (and presumably make sure we ship types for it one way or another—via Typed Ember support or via the package itself, at Framework Core’s discretion) as a union type?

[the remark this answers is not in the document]
pzuraq: FWIW, for effects, using untracked state is actually something that may not be unusual. The point of an effect is to push state out of the system, into an untracked world. If there is a coalescing intermediate state that is untracked, that's perfectly fine, as long as the rest of the app does not need to _react_ to those changes. And if it _does_, then it begs the question of why you are using an effect in the first place.

text/0625-helper-managers.md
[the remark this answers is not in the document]
rwjblue: FWIW, I wasn't trying to convince you to change it (though I think we should 😉), but the fact that it is missing from being explicitly stated is my fundamental problem here.

text/0625-helper-managers.md
pzuraq: hmm, I think it was supposed to be "one helper manager per helper _manager_ factory". The semantics are that we weakmap the function that produces the manager to the instance of the manager, so you could use the same factory function with many different definitions, and they would all use the same instance. That's the thing we need to capture here.

text/0625-helper-managers.md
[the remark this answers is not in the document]
rwjblue: Sure, but then the options are not in fact optional (because you **must** chose either `hasValue` or `isScheduledEffect`)

text/0625-helper-managers.md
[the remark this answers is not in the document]
rwjblue: > FWIW, for effects, using untracked state is actually something that may not be unusual.

Yes, I totally understand/agree, but if you go this route it is fundamentally impossible to _also_ allow title tokens to be pushed into the service directly (which is a thing you can do in `ember-page-title` today).

text/0625-helper-managers.md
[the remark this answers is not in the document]
chriskrycho: Yeah, the interplay between `Object`, `object`, `typeof`, and `Function` in TS is admittedly a bit funky. `object` ≠ `typeof something === 'object'`, and there’s no equivalent to `'function'` from typeof (or most of the others). 😭

text/0625-helper-managers.md
rwjblue: Ya, my concern wasn't about that coupling, but it was with the terminology used to talk about which thing the factory function is associated with. There is no conceptual thing introduced in this RFC as a "helper factory". There is a `HelperDefinition` (which is what this argument is actually called throughout the rest of this document), but no "helper factory". 

Also, conceptually calling this a "helper factory" is misleading. There is no reason for us to infer that this value is a "factory" at all (factory usually inferring that you can do `new Foo()` or `Foo.create()` in Ember parlance). For example, it is perfectly valid for this value to be `{}`.

pzuraq: right, but again, I'm not talking about the definition, I'm talking about the function that creates the helper manager itself. That is a different thing entirely, which we need a new term for, it is _not_ the `HelperDefinition`

text/0625-helper-managers.md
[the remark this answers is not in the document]
pzuraq: I think that in the future, with a different set of capabilities, they _could_ be, but currently they are not. So yeah, will change the types to reflect that.

text/0625-helper-managers.md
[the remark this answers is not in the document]
pzuraq: > also allow title tokens to be pushed into the service directly (which is a thing you can do in ember-page-title today).

I'm not sure this is true, based on looking at the implementation. `{{title}}` schedules an update from within the helper, and _only_ from within the helper, for changes. That update happens async in the next actions queue, and is the way that `page-title` currently avoids the backtracking rerender assertion (essentially opting into a second render pass).

It'd be possible to do this from a service as well, but I don't think it's possible to avoid a second render pass in general using a declarative approach. By the time you hit a `{{title}}` helper, the actual header has already rendered, so it will necessarily cause a backtracking assertion.

IMO It is _completely fine_ to require a second render pass, and effects as spec'd would definitely allow this. In fact, I don't think there is a way to _disallow_ this. This is the strategy I'll take in the example implementation. The main thing we want is people to do this consciously and carefully, and to document that it is an advanced technique which should be used sparingly (basically, don't just start scheduling willy-nilly if you hit the assertion)

text/0625-helper-managers.md
```diff
@@ -0,0 +1,504 @@
...
+
+**Never**
+- called if the `isScheduledEffect` capability is disabled
+
+#### `getDestroyable`
+
+`getDestroyable` is an optional hook that users can use to register a
+destroyable object for the helper. This destroyable will be registered to the
+containing block or template parent, and will be destroyed when it is destroyed.
+See the [Destroyables RFC](https://github.com/emberjs/rfcs/blob/master/text/0580-destroyables.md)
```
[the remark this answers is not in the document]
pzuraq: It does exist in that RFC actually, though it could be easy to miss:

> For the remainder of this RFC, the terms "destroyable" and "destroyable object" will be used to mean any object which is a valid WeakMap key (e.g. typeof obj === 'object' || typeof obj === 'function'). Any JS object that fulfills this property can be used with this system.

text/0625-helper-managers.md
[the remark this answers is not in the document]
rwjblue: I think this is why the demo of ember-page-title will be helpful. In this case, I could easily see the solution being to use two helpers. But what your really want is:

* `{{title` helper just does the pushing of tokens onto the services tracked "title tokens" array
* another helper that actually does the effect that you have to included yourself in `application.hbs` (e.g. `{{set-document-title}}`), this one leverages the tracked title tokens, so it always re-runs the effect as needed)

But you don't _really_ want the first thing to produce a value (the value is meaningless, and the helper ATM returns `''` because it can't signify that it returns nothing). IMHO, it may point to the fact that you want to do "something" during the render cycle, but still not return a value.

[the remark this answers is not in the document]
pzuraq: Sorry, I was basing my last comment on the structure of `{{page-title}}` specifically, which uses `ember-cli-head`, which renders a full template snippet in the header. So mutating a value during render would not be possible, it really does consume the value first, and `{{title}}` really does need to use untracked state and schedule an update itself.

More generally though, I could see the pattern of mutating in a value-helper and then using an effect-helper to finalize the effect. You can compose a value helper and effect helper together to get this property if needed, so I think mutual exclusivity still makes sense.

text/0625-helper-managers.md
rwjblue: Hmm, why would we do a different thing here than for component managers / modifier managers? We do not cache the factory instance on the manager factory function, we call the manager factory function for each component base class we find.

See this demo repo if you'd like to fact check me 😜:

https://github.com/rwjblue/manager-invocation-demo

text/0625-helper-managers.md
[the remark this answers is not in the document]
rwjblue: Aha! Thank you! 🤦

text/0625-helper-managers.md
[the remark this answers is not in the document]
rwjblue: Ya, but in the case that I'm referring to I just don't think producing a value actually makes. One helper is for collecting the tokens, the other is for queuing the side effect, and the one that collects tokens really has no valid "value".

text/0625-helper-managers.md
pzuraq: Ah, this is my mistake then. This is how managers work in Glimmer.js currently, I must've forgotten where we put that optimization. It looks like it first came from `ember-functional-modifiers`, which makes sense because that addon needed to assign the same modifier manager to many functions and we wanted to avoid creating an instance per function: https://github.com/ember-modifier/ember-modifier/blob/master/addon/-private/functional/modifier.js

Personally I think this optimization is common enough that we should consider making it the default, but we can make it something users have to implement if they need it. I'm not sure if it would be a breaking change to update the existing managers - in theory, they shouldn't depend on instance state per-definition (they may depend on state overall, or per owner), that would somewhat defeat the purpose.

text/0625-helper-managers.md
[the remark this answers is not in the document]
pzuraq: Can you make that case more concrete by providing an example helper that I can also convert? Currently I'm focusing on `ember-page-title`, which is not using that strategy, and I think a solid example use case would really help. 

Adding a hook explicitly for side-effects that _does_ allow internal state mutation would essentially be bringing back observers. When we previously discussed this, everyone agreed that would probably not be ideal, and that was why the limitation about mutating state was placed on effect-helpers. We can always remove that restriction in the future, if we find motivating use cases. Users using the `getValue()` hook to frequently circumvent the restriction would, for instance, definitely show that it would have value, as would users frequently scheduling changes from effects directly.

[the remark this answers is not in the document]
pzuraq: Another thing to think about, is that once we add `invokeHelper` into the mix, `getValue()` will _not_ trigger unless used directly in JS. This really implies that side-effects _should not_ happen during render, since there's no way to ensure they happen predictably.

I think we would need to add a _second_ effect queue if we were to move this direction.

1. The first effect queue _would_ allow internal app state changes, and would basically be the same as async observers today.
2. The second queue _would not_ allow internal app state changes, and it would be about reflecting state elsewheres.

text/0625-helper-managers.md
rwjblue: > Personally I think this optimization is common enough that we should consider making it the default

Ya, I'm sure. I think that might be fine, but adding caching for something that is super light weight (`new Something()` where that constructor does _basically_ nothing) is actually going to make this slower a lot of times.

Even if we do this, the prose here is wrong. It is not one per manager factory function, it is one per manager factory function per owner. I think that is conceptually fine though, just needs tweaking to be clearer I guess.

rwjblue: If we do that, we should also consider doing the same for the other manager types. I'd hate to introduce another small difference while our overall goal is to _reduce_ the differences. I think we can just do the implementation change for the others though, I don't particularly think it is a breaking change or anything.

text/0625-helper-managers.md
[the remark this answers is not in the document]
rwjblue: To be clear, I don't think my concern here is blocking. Just something we need to have a good answer / explanation for. Starting out more restrictive then relaxing that restriction is absolutely better than starting out too permissive and being stuck.

text/0625-helper-managers.md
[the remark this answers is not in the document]
pzuraq: I added a section noting that the `createHelper` function is not tracked, and strategies for how users can track instance creation if they want to. We can keep discussing whether or not the hook should be tracked, I think the biggest question there is if it is, what would it do when it invalidates?

text/0625-helper-managers.md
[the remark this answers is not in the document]
rwjblue: Also, either way having the `ember-page-title` case study will help and be a nice justification for the new features.

text/0625-helper-managers.md
```diff
@@ -0,0 +1,845 @@
...
+}
+
+setHelperManager((owner) => new EffectHelperManager(owner), Effect.prototype);
+```
+
+```js
+// ember-page-title/addon/services/page-title-list
+
+import Service, { inject as service } from '@ember/service';
+import { getOwner } from '@ember/application
```
chriskrycho, line 588:
```suggestion
import { getOwner } from '@ember/application';
```

text/0625-helper-managers.md
```diff
@@ -0,0 +1,845 @@
...
+      .filter(token => Boolean(token.title))
+      .map((token, index) => {
+        if (index + 1 < tokens.length) {
+          return token.title + token.separator;
+        }
+
+        return token.title;
+      })
+      .join('');
+  },
```
chriskrycho, line 789:
```suggestion
  }
```

text/0625-helper-managers.md
```diff
@@ -0,0 +1,845 @@
...
+    if (this._hasFastboot()) {
+      return;
+    }
+
+    let titles = document.getElementsByTagName('title');
+    for (let i = 0; i < titles.length; i++) {
+      let title = titles[i];
+      title.parentNode.removeChild(title);
+    }
+  },
```
chriskrycho, line 805:
```suggestion
  }
```

text/0625-helper-managers.md
```diff
@@ -0,0 +1,845 @@
...
+import { inject as service } from '@ember/service';
+
+export default class Title extends Effect {
+  @service pageTitleList;
+
+  constructor(owner, args) {
+    super(owner, args);
+
+    this.pageTitleList.push({ id: guidFor(this) });
+  },
```
chriskrycho, line 826:
```suggestion
  }
```

text/0625-helper-managers.md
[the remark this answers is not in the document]
rwjblue: > I think the biggest question there is if it is, what would it do when it invalidates?

Ya, this is **exactly** the question. 

@krisselden - I think you had thoughts here, what do you think?

text/0625-helper-managers.md
```diff
@@ -502,3 +520,326 @@ export function helper(fn) {
...
+  }
+
+  getDestroyable({ instance }) {
+    registerDestructor(instance, instance.teardown.bind(instance));
+
+    return instance;
+  }
+}
+
+setHelperManager((owner) => new EffectHelperManager(owner), Effect.prototype);
```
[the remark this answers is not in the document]
pzuraq: Ah, I got instance and definition prototype chains mixed up 🙃

text/0625-helper-managers.md
```diff
@@ -0,0 +1,845 @@
...
+
+  getValue?(bucket: HelperStateBucket, args: TemplateArgs): unknown;
+
+  runEffect?(bucket: HelperStateBucket, args: TemplateArgs): void;
+
+  getDestroyable?(bucket: HelperStateBucket): object;
+}
+```
+
+The capabilities property should be provided using the `capabilities()` function
```
jelhan, line 358: It's inconsistent if this is a requirement (*must*) or a recommendation (*should*). It uses *must* just a little bit later.

text/0625-helper-managers.md
```diff
@@ -0,0 +1,861 @@
...
+This hook has the following timing semantics:
+
+**Always**
+- called the first time the helper value is requested
+- called after autotracked state has changed
+
+**Never**
+- called if the `hasValue` capability is disabled
+
+#### `runEffect`
```
rwjblue, line 216: Can you add some prose to clarify what the behavior is in SSR / FastBoot? As you know we do not run modifiers in SSR, and in general SSR does not have the ability to run update opcodes. I could easily imagine that we say `runEffect` should **not** run in SSR (essentially saying that you shouldn't have external side effects), but we should be clear and we should consider it carefully.

rwjblue: I think we _probably_ should say they don't run in SSR for now. I can absolutely imagine wanting them eventually in SSR, but I think the safe default to start with (especially while introducing this new concept) is to disallow it. In the future, we can add a capability to explicitly declare that you do or do not want to run in SSR which means this isn't really blocking future functionality.

rwjblue:
We discussed this in todays framework meeting, and are in favor of moving forward here! 

Look out! It's final comment period time!! 💬 🎭

rwjblue:
ZOMG ZOMG ZOMG ZOMG ZOMG ZOMG ZOMG ZOMG ZOMG ZOMG ZOMG ZOMG ZOMG ZOMG ZOMG ZOMG ZOMG ZOMG ZOMG ZOMG ZOMG ZOMG ZOMG ZOMG ZOMG ZOMG ZOMG ZOMG ZOMG ZOMG ZOMG ZOMG 

LETS DO IT!!!!!

BryanCrotaz:
Rendered link gives a 404??

rwjblue:
Updated (it was the link to the branch, which was deleted upon merging).

PR #626, opened 2020-05-08
`invokeHelper`

pzuraq:
[Rendered](https://github.com/emberjs/rfcs/blob/invoke-helper/text/0626-invoke-helper.md)

### DIFF
--- a/text/0626-invoke-helper.md
+++ b/text/0626-invoke-helper.md
@@ -389,10 +389,6 @@ export default class Search extends Component {
 }
 ```
 
-## Dependencies
-
-This RFC depends on the [Helper Manager RFC](https://github.com/emberjs/rfcs/pull/625).
-
 ## Detailed design
 
 This RFC proposes adding the `invokeHelper` function, imported from
--- a/text/0626-invoke-helper.md
+++ b/text/0626-invoke-helper.md
@@ -1,6 +1,6 @@
 - Start Date: 2020-04-30
 - Relevant Team(s): Ember.js
-- RFC PR: (after opening the RFC PR, update this with a link to it and update the file name)
+- RFC PR: https://github.com/emberjs/rfcs/pull/626
 - Tracking: (leave this empty)
 
 # JavaScript Helper Invocation API
--- a/text/0626-invoke-helper.md
+++ b/text/0626-invoke-helper.md
@@ -16,48 +16,22 @@ import Component from '@glimmer/component';
 import Helper from '@ember/component/helper';
 import { invokeHelper } from '@ember/helper';
 
-class FetchTask {
-  @tracked isLoading = true;
-  @tracked isError = false;
-  @tracked result = null;
-
-  constructor(url) {
-    this.run(url);
-  }
-
-  async run(url) {
-    try {
-      let response = await fetch(url);
-      this.result = await response.json();
-    } catch {
-      this.isError = true;
-    } finally {
-      this.isLoading = false;
-    }
+class PlusOne extends Helper {
+  compute([num]) {
+    return number + 1;
   }
 }
 
-class RemoteData extends Helper {
-  compute([url]) {
-    return new FetchTask(url);
-  }
-}
-
-export default class DataLoader extends Component {
-  data = invokeHelper(this, RemoteData, () => {
-    positional: [this.args.url]
+export default class PlusOne extends Component {
+  plusOne = invokeHelper(this, RemoteData, () => {
+    return {
+      positional: [this.args.number],
+    };
   });
 }
 ```
 ```hbs
-<!-- app/components/data-loader.hbs -->
-{{#if this.data.value.isLoading}}
-  Loading...
-{{else if this.data.value.isError}}
-  Something went wrong!
-{{else}}
-  {{this.data.value.result}}
-{{/if}}
+{{this.plusOne.value}}
 ```
 
 ## Motivation
@@ -89,11 +63,13 @@ as such, they have a _lifecycle_. Unlike a function, a component can _update_
 over time, and will be _destroyed_ at some unknown point in the future.
 
 Components previously exposed this lifecycle directly via a number of
-lifeycle hooks, but there were many issues with these hooks. These stemmed from
-the fact that components were the smallest _atom_ for reactive composition. In
-the world of functions, a piece of code can always be broken out into a new
-function, giving the user the ability to extract repeated functionality,
-abstracting common patterns and concepts and reducing brittleness.
+lifeycle hooks, making components the smallest _atom_ for reactive composition.
+This presented an issue for composability in general. In the world of functions,
+a piece of code can always be broken out into a new function, giving the user
+the ability to extract repeated functionality, abstracting common patterns and
+concepts and reducing brittleness. For instance, in the following example we
+extract several portions of the `myProgram` function to make it clearer what
+each section is doing, and isolate its behavior.
 
 ```js
 // before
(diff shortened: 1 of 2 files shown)


jelhan:
Would it be possible to implement `invokeHelper` in an addon using the proposed Helper Manager?

pzuraq:
It would not. There is no way to lookup a helper manager instance using public APIs, and the hooks on helper managers are meant to be wrapped by the VM directly. For instance, `runEffect` needs to be scheduled by the VM at the correct point in the lifecycle of the helper.

This is the minimum API we can expose in order to invoke a helper in the same way as it would be if done by the VM, which is the goal.

text/0626-invoke-helper.md
```diff
@@ -0,0 +1,475 @@
...
+meant to be functional, but not particularly readable or ergonomic. This API can
+be wrapped with higher level, more ergonomic APIs in the ecosystem, until we're
+sure what the final API should be.
+
+## How we teach this
+
+This API is meant to be a low-level primitive which will eventually be replaced
+with higher level wrappers, possibly decorators, that will be much easier to use
+and recommend to average app developers. As such, it will only be taught through
+API documentation.
```
jelhan, line 391: I fear it's a little bit more complex than this. Helpers are currently reached as *template* helpers. This proposal decouples them from the template. They could be invoked everywhere. It might not be important to teach the low-level API to invoke them. But it would be important to reflect in the guides that they could also be used outside of template context.

pzuraq: Yes, I do agree there. I think we should update the guides to reflect this once we have a solid community example, similar to what we did with `ember-modifier`. I'll add that to the RFC.

text/0626-invoke-helper.md
```diff
@@ -0,0 +1,475 @@
...
+
+class RemoteData extends Helper {
+  compute([url]) {
+    return new FetchTask(url);
+  }
+}
+
+export default class DataLoader extends Component {
+  data = invokeHelper(this, RemoteData, () => {
+    positional: [this.args.url]
```
allthesignals, line 48: This is awesome, I've needed this many times. When I first looked at this, I expected it to be a little closer to the way the helper looks in a template (`{{remote-data 'https://google.com'}}`). I think I understand that that's not entirely possible given the nature of tracked properties (and also passing these params at instantiation).

I wonder if something like this would work?

```javascript
get data() {
  return invokeHelper(this, RemoteData, [this.args.url], /* { } */);
}
```

That might get around the tracked properties issue, but then the helper would be invoked each time.

Is "invoke" the right word for this? "bindHelper"?

text/0626-invoke-helper.md
```diff
@@ -0,0 +1,475 @@
...
+    } catch {
+      this.isError = true;
+    } finally {
+      this.isLoading = false;
+    }
+  }
+
+  refresh() {
+    this.run();
+  }
```
rwjblue, line 228: Can we just name `run` as `refresh`?

text/0626-invoke-helper.md
```diff
@@ -0,0 +1,475 @@
...
+    super(...arguments);
+
+    use(this, poll(() => [() => this.data.refresh(), this.pollPeriod]));
+  }
+}
+```
+
+## Dependencies
+
+This RFC depends on the [Helper Manager RFC](https://github.com/emberjs/rfcs/pull/625).
```
rwjblue, line 287: Probably can remove this (since that RFC already landed)

text/0626-invoke-helper.md
```diff
@@ -0,0 +1,475 @@
...
+encourages users to ensure they've properly handled the lifecycle of their
+helper.
+
+#### `definition`
+
+This is the helper definition. It can be any object, with the only requirement
+being that a helper manager has been associated with it via the
+`setHelperManager` API.
+
+#### `argsGetter`
```
rwjblue, line 333: Can we call this `computeFunction` or something? I think of "getter" generally in the ES5 `get someFunction() {}` sense...

text/0626-invoke-helper.md
```diff
@@ -0,0 +1,475 @@
...
+function invokeHelper(
+  context: object,
+  definition: HelperDefinition,
+  argsGetter?: (context: object) => TemplateArgs
+): Helper;
+```
+
+Let's step through the arguments to the function one by one:
+
+#### `context`
```
rwjblue, line 316: Can we call this something like "parent destroyable"?

text/0626-invoke-helper.md
```diff
@@ -0,0 +1,475 @@
...
+
+This is the parent context for the helper definition. The helper will be
+associated as a destroyable to this parent context, using the destroyables API,
```
rwjblue, line 319: Maybe link to the destroyables RFC here?

text/0626-invoke-helper.md
```diff
@@ -0,0 +1,475 @@
...
+
+This allows helper's lifecycles to be entangled correctly with the parent, and
+encourages users to ensure they've properly handled the lifecycle of their
+helper.
+
+#### `definition`
+
+This is the helper definition. It can be any object, with the only requirement
+being that a helper manager has been associated with it via the
+`setHelperManager` API.
```
rwjblue, line 331: Can you link to the detailed design section in the helper manager RFC?

text/0626-invoke-helper.md
```diff
@@ -0,0 +1,475 @@
...
+with a `positional` property that is an array and/or a `named` property that is
+an object.
+
+This getter function will be _autotracked_ when it is run, so the process of
+retrieving the arguments is autotracked. If any of the values used to create the
+arguments object change, the helper will be updated, just like in templates.
+
+### Return Value
+
+The function returns an instance of the helper. The public API of the instance
```
rwjblue, line 346: > The function returns an instance of the helper

I find this **very confusing**, when I read "instance of the helper" I assume you mean the result of creating the thing _I_ put in `app/helpers/foo.js`, but this is actually talking about some wrapper object.

Specifically, as written, I would expect the `this` in this `constructor` to **be** the "helper instance":

```js
// app/helpers/foo.js

export default class extends Helper {
  constructor() {
    this; // <- **THIS THING IS THE INSTANCE OF THE HELPER**
  }
}
```

text/0626-invoke-helper.md
```diff
@@ -142,124 +118,252 @@ function myProgram(data = []) {
...
+    <DocTitle @title='Search Result for "{{@query}}"'/>
+
+    <FetchData @url="www.example.com/search?query={{@query}}" as |data|>
+      ...
+    </FetchData>
+    ```
+
+   This structure is also not ideal because the components aren't being used for
+   templating, they're just being used for logic effectively. So, there's no
+   reason to add the overhead of a component here.
```
rwjblue:
```suggestion
   templating, they're just being used for logic effectively.
```

text/0626-invoke-helper.md
```diff
@@ -142,124 +118,252 @@ function myProgram(data = []) {
...
+
+3. We could use other template constructs, such as helpers or modifiers. Both
+   helpers and modifiers have lifecycles, like components, and can be used to
+   contain functionality. Modifiers aren't really a good choice here though,
+   because it would require us to add an element that we don't need. So, helpers
+   are the better option.
+
+   Helpers work, but like components they require us to move some of our logic
+   into the template, even if that isn't really necessary:
```
rwjblue, line 286: I'm not sure I follow this part. What is the unnecessary part in the template?

text/0626-invoke-helper.md
```diff
@@ -142,124 +118,252 @@ function myProgram(data = []) {
...
+    <!-- app/components/search.hbs -->
+    {{doc-title 'Search Result for "{{@query}}"'}}
+
+    {{#let (fetch-data "www.example.com/search?query={{@query}}") as |data|}}
+      ...
+    {{/let}}
+    ```
+
+Out of these options, helpers are the closest to what we want - they are low
+overhead, they produce computed values directly without a template, and
```
rwjblue, line 335: FWIW, I strongly dislike the "overhead" argument. I just don't see any rationale for how helpers are inherently lower overhead than a component in modern Ember (e.g. the Octane mental model). Both helpers and components (ala `@glimmer/component`) are _effectively_ the cost of doing `new SomeThing()` where `SomeThing` has a very shallow inheritance structure and very little code in `constructor`.

text/0626-invoke-helper.md
```diff
@@ -142,124 +118,252 @@ function myProgram(data = []) {
...
+with the recent addition of effect helpers they can be used side-effect to
+accomplish tasks like setting the document title. The only downside is that they
+can only be invoked in templates, so they require you to design your components
+around using them in templates only. This can be difficult to do in many cases,
+where the data wants to be accessed to create derived state for instance.
+
+This RFC proposes adding a way to create helpers within JavaScript directly,
+extending the reactive model in a way that allows users to extract common
+reactive code and patterns, and reuse them transparently. This will make helpers
+the new reactive atom of the system - the reactive equivalent of a "function" in
```
rwjblue, line 345: To be clear, I don't think this actually changes the atom, it just makes helpers an equal peer to components. Right?

text/0626-invoke-helper.md
```diff
@@ -142,124 +118,252 @@ function myProgram(data = []) {
...
+our incremental model. Like components, they have a lifecycle, and can update
+over time. Unlike components, they can exist nested in JavaScript classes _and_
+in templates, and they can produce any type of value, making them much more
+flexible.
```
rwjblue, line 349: I'm not sure I'm following this part, is the point you are making that components can't be nested in JavaScript?

text/0626-invoke-helper.md
```diff
@@ -305,20 +409,21 @@ interface Helper {
...
 
-#### `context`
+#### `parentDestroyable`
 
-This is the parent context for the helper definition. The helper will be
-associated as a destroyable to this parent context, using the destroyables API,
-so that its lifecycle is tied to the parent. The only requirement of the parent
-is that is an object of some kind that can be destroyed.
+This is the parent for the helper definition. The helper will be associated as a
+destroyable to this parent context, using the destroyables API, so that its
```
rwjblue, line 426: Can you add a link to the destroyables RFC here?

text/0626-invoke-helper.md
```diff
@@ -305,20 +409,21 @@ interface Helper {
...
+lifecycle is tied to the parent. The only requirement of the parent is that is
```
rwjblue, line 427:
```suggestion
lifecycle is tied to the parent. The only requirement of the parent is that it is
```

text/0626-invoke-helper.md
pzuraq: Right, exactly. You can't refer to the value of the helper in JS anymore, so any computed properties you have that may rely on the value of the helper now need to be in the template, any other business logic now needs to be moved into the template, etc.

text/0626-invoke-helper.md
pzuraq: Helpers are already a peer. The issues is neither helpers nor components can be used in JS-only. This API makes helpers better than components for many use cases, and makes them usable in many more places, so I would say they are the reactive atom now. Especially considering Glimmer components have no lifecycle hooks, so they can't even do the same things anymore.

text/0626-invoke-helper.md
pzuraq: Yes, and that makes them less flexible.

pzuraq: Yes, and that makes them less flexible.

text/0626-invoke-helper.md
pzuraq: > That might get around the tracked properties issue, but then the helper would be invoked each time.

Right, exactly. You would need to do something like:

```js
get data() {
  return invokeHelper(this, 'data', RemoteData, [this.args.url] /*, { } */);
}
```

Where the second argument is a unique key specifically for this instance. Overall this API would be much less flexible, which is why I proposed the one in this RFC. We can expose higher level APIs built on top of the low level one that work more like the way you described.

> Is "invoke" the right word for this? "bindHelper"?

Binding typically doesn't create an instance of something, it curries it and returns a new definition. This actually is creating an instance, so invoking is more correct I believe.

text/0626-invoke-helper.md
rwjblue: Thank you for clarifying! From my perspective _some_ of this concern is mitigated with upcoming changes to allow components to be "just values", but still seems good to state this here (I was just confused).

text/0626-invoke-helper.md
rwjblue: Sure, but the prose here infers that we should now always decompose components into helpers because "they are the smallest atom", which I don't think is correct either.

text/0626-invoke-helper.md
rwjblue: Hmm, I don't see that as unnecessary 🤔 

Either way (either with this RFC or with templates) you have to invoke the helper. The only question is where that invocation actually is.

text/0626-invoke-helper.md
```diff
@@ -0,0 +1,564 @@
...
+
+```ts
+interface TemplateArgs {
+  positional?: unknown[],
+  named?: Record<string, unknown>
+}
+
+type HelperDefinition = object;
+
+interface Helper {
```
rwjblue, line 406: I still think this is a large issue. This is **not** a "helper", this is like a "helper value wrapper thingy". 🤔

text/0626-invoke-helper.md
```diff
@@ -0,0 +1,564 @@
...
+with a `positional` property that is an array and/or a `named` property that is
+an object.
+
+This getter function will be _autotracked_ when it is run, so the process of
+retrieving the arguments is autotracked. If any of the values used to create the
+arguments object change, the helper will be updated, just like in templates.
+
+### Return Value
+
+The function returns a reference to the helper. The public API of the reference
```
rwjblue, line 450: I don't think this is correct. You just don't get a reference to the helper, you get some wrapping value that has a `.value` property but **not** "the helper".

rwjblue:
We discussed this in today’s core team meeting, and think it is ready to move into final comment period. 

🥖🥳🕺

betocantu93:
Great feature, I definitely would use this. Just one quick question, not sure if I'm missing something but in your example

```ts
import { invokeHelper } from '@ember/helper';

class PlusOne extends Helper {
  compute([num]) {
    return number + 1;
  }
}

export default class PlusOne extends Component {
  plusOne = invokeHelper(this, RemoteData, () => {
    return {
      positional: [this.args.number],
    };
  });
}
```
You meant `RemoteData` to be PlusOne (the helper)? something like

```ts
import { invokeHelper } from '@ember/helper';

class PlusOneHelper extends Helper {
  compute([num]) {
    return number + 1;
  }
}

export default class PlusOne extends Component {
  plusOne = invokeHelper(this, PlusOneHelper, () => {
    return {
      positional: [this.args.number],
    };
  });
}
```

scottmessinger:
I'm excited for this to land -- definitely will be useful for us.

The part I'm confused by is the `computeArgs` return value. If the `positional` value is always an array and `named` is always an object, why not just return either an array or an object from the function? Being familiar with helpers, the idea of calling it `positional` makes sense, but if I didn't know helpers can have positional arguments, I think I'd be at a loss to explain this API.

Not knowing the limitations that might have led to this design choice, I would prefer if the types look more like this:

```diff
-interface TemplateArgs {
-  positional?: unknown[],
-  named?: Record<string, unknown>
-}
+type TemplateArgs = unknown[] | Record<string, unknown>

type HelperDefinition<T = unknown> = object;

function invokeHelper<T = unknown>(
  parentDestroyable: object,
  definition: HelperDefinition<T>,
  computeArgs?: (context: object) => TemplateArgs
): Cache<T>;
```

jelhan:
> If the positional value is always an array and named is always an object, why not just return either an array or an object from the function?

A helper could use both positional and named arguments at the same time. E.g. `{{format-date date timeZone="UTC"}}` from ember-intl.

pzuraq:
@scottmessinger @jelhan's explanation is correct, it is not an either-or API, both can be passed in at the same time. This API is also meant to be a primitive API - the idea is that higher level APIs will be written on top of this to provide easier to use, more ergonomic DX. Those APIs could add the logic you're describing for instance.

rwjblue:
We discussed this in Friday's core team meeting, and think it is ready to land. Thank you to @pzuraq and all of y'all that helped hone this proposal!

🚁 🐎 🚗

chriskrycho:
Note: OP needs an update to the final version of the RFC!

lougreenwood:
I wonder if the RFC should mention/clarify that this doesn't mean that helpers now replace utils (or perhaps it does) - in many cases a helper can also be useful as a util for re-use in some JS. At the moment it makes sense to write the helper as a util and then call the util from the helper:

`utils/my-reusable-util.js`
```js
export default function myReusableUtil(someService, thingy) {
  return this.someService.modifyThingy(thingy)
}
```

`helpers/my-reusable-util.js`
```js
import myReusableUtil from '../utils/helper';

export default class MyReusableUtilHelper extends Helper {
  @service someService;

  compute([thingy]) {
    return myReusableUtil(this.someService, thingy);
  }
}
```

However this PR could encourage the perspective that most/all utils should just always be helpers, since helpers can be easily invoked in JS or templates - so why bother with utils at all after this?

Issue #582, opened 2020-01-21
Ember 4.0


Gaurav0:
Let us set up a timeline for Ember 4.0 in order to be able to drop support for IE11.

Gaurav0:
For reference: https://blog.emberjs.com/2017/10/03/the-road-to-ember-3-0.html

NullVoxPopuli:
Things I would like to see landed before Ember 4:
 - embroider 
  + tree shaking
 - option to have no controllers
 - typed templates
 - template imports
   + ability to have components anywhere
   + file/project linter to enforce component locations.
- binary templates
- information on GlimmerVM as WASM (was this a good idea?)

All of this might be too far out there, but also, It could warrant another edition. 

If Ember 3 is carrying too much bloat with it, I'd be cool with a bump  to 4. However, because everyone is really slow at upgrading, I think we'd want to wait at least a couple of LTSs.. Maybe once 3.24 is LTS, it'd make sense to do Ember 4.0, and remove all the cruft from the pre-octane era

Gaurav0:
Why should we continue to support IE11 for another year or two while all that is being done?

NullVoxPopuli:
tbf, IE11 is holding everyone back that needs to support it.

lifeart:
IE 11 supporting may be opt-in feature with some codestyle limitations.

makepanic:
As an alternative to dropping IE11 one could also investigate what other frameworks are doing with [Differential Loading](https://angular.io/guide/deployment#differential-loading) that splits the build output into a evergreen and legacy bundle.

This allows newer browser to get a small payload while older browsers can still work by loading more javascript.

There's also the possibility of letting the developer handle old browsers themselves by making polyfills easaier to load, see e.g. [polyfills.ts in NG2](https://angular.io/guide/browser-support#mandatory-polyfills).

This won't work if ember uses API that isn't possible to polyfill though.

Panman82:
I might be wrong here, but I think there are newer things that cannot be polyfilled and therefore (once used) will forever break backwards compatibility with IE11.

pzuraq:
Certain features, such as native Proxy, are impossible to polyfill, and so the framework cannot use them for many production features. I think the ability to use Proxy would be incredibly helpful, and definitely should factor into the timeline.

[...]

PR #627, opened 2020-05-15
Timeline for Transitioning Away from IE11

Gaurav0:
Now that LinkedIn is dropping support for IE11 on July 1, 2020, I am able to open this RFC!

For #582

[Rendered](https://github.com/emberjs/rfcs/blob/20857f5fd742295704d27c98ad6a2f6f4d5cd2a5/text/0627-ember-version-4.md)

### DIFF
--- a/text/0000-ember-version-4.md
+++ b/text/0000-ember-version-4.md
@@ -1,0 +1,70 @@
+- Start Date: 2020-05-15
+- Relevant Team(s): All teams
+- RFC PR: (after opening the RFC PR, update this with a link to it and update the file name)
+- Tracking: (leave this empty)
+
+# Ember 4.0
+
+## Summary
+
+> Ember.js, Ember-Data, and Ember-CLI will release a new major version, 4.0,
+approximately six months from today, somewhere around mid November. Version
+3.23 will be the last version of the 3.x cycle, and will become an LTS release.
+In order to prevent a rush of deprecations submitted to Ember prior to the major
+version bump, any new proposed deprecations must target Ember 5.0.
+
+## Motivation
+
+> We will release a new major version of Ember in order to drop support for IE11.
+There are several major reasons for doing this.
+
+1. Supporting IE11 prevents us from using the latest web technologies in Ember,
+such as native proxies.
+
+1. Supporting IE11 requires us to ship code to the client that in order to polyfill
+features that IE11 does not support.
+
+1. Supporting IE11 adds a significant maintenance overhead for the Ember project.
+
+> We are now able to drop support for IE11 since major enterprises such as LinkedIn
+are doing so. However, we are waiting six months in order for companies to notify
+clients and make appropriate plans. Developers that still need to support IE11 may
+stay on Ember 3.x indefinitely, and it will be a supported LTS until Ember 4.4 becomes LTS,
+approximately mid May 2021.
+
+## Detailed design
+
+> No new features will be introduced in Ember 4.0. Instead, 
+all deprecated functionality targeting Ember 4.0 may be removed following its release.
+
+> There will be special handling for imports from `@ember/polyfills`. Specifically,
+these will not emit deprecation warnings by default in the Ember 3.x cycle so that both
+apps and addons can continue to work with IE11. A way to turn on deprecation warnings
+in some versions of Ember 3.x may be provided in a future RFC.
+In Ember 4.0, these functions will be replaced by stubs which emit a deprecation warning
+before delegating to native methods.
+These deprecated stubs will continue to exist until Ember 4.4 to give time for apps and addons
+to remove them replace them with native methods.
+They may be removed beginning in Ember 4.5.
+Please note that these imports shall not actually work in IE11 in Ember 4.x; they exist
+primarily to ease code migration.
+
+## How we teach this
+
+A blog post will be released announcing this as soon as this RFC is accepted, detailing
+all relevent information, similar to the one announcing
+[Ember 3.0](https://blog.emberjs.com/2017/10/03/the-road-to-ember-3-0.html).
+
+## Drawbacks
+
+> We may push some companies which must continue to support IE11 into looking into
+alternative frameworks.
+
+## Alternatives
+
+> We can continue to support IE11 for a longer period of time.
+
+## Unresolved questions
+
+> How can we allow some applications and/or addons to opt into receiving deprecation
+warnings for `@ember/polyfills` in the remainder of the 3.x cycle?
--- a/text/0627-ember-version-4.md
+++ b/text/0627-ember-version-4.md
@@ -8,7 +8,7 @@
 ## Summary
 
 > Ember.js, Ember-Data, and Ember-CLI will release a new major version, 4.0,
-at least six months from today, possibly as soon as December, but this date
+at least six months from today, possibly as soon as December 2021, but this date
 may slip for a variety of reasons.
 The last version in the 3.x cycle of each of these projects will be an LTS.
 In order to prevent a rush of deprecations submitted to Ember prior to the major
@@ -20,12 +20,20 @@ version bump, any new proposed deprecations must target Ember 5.0.
 There are several major reasons for doing this.
 
 1. Supporting IE11 prevents us from using the latest web technologies in Ember,
-such as native proxies.
+such as native proxies, that cannot be polyfilled.
 
 1. Supporting IE11 requires us to ship code to the client that in order to polyfill
 features that IE11 does not support.
 
 1. Supporting IE11 adds a significant maintenance overhead for the Ember project.
+
+1. If separate code is not shipped, IE11 polyfills can cause significant performance issues.
+They do not optimize well in V8 and other JavaScript engines.
+
+1. Many addons do not support IE11 but do not specifically document this. This can lead
+to confusion. Without adequate testing, many applications fail to support IE11 in practice.
+
+1. Many community tools, such as Ember Inspector and Ember Twiddle, do not work in IE11.
 
 > We are now able to drop support for IE11 since major enterprises such as LinkedIn
 are doing so. However, we are waiting at least six months in order for companies to notify
@@ -38,7 +46,7 @@ approximately mid May 2021.
 > In a separate RFC, a new browser support policy will be introduced for the 4.x cycle
 and possibly beyond.
 
-> No new features will be introduced in Ember 4.0. Instead, 
+> No new features will be introduced in Ember 4.0. Instead,
 all deprecated functionality targeting Ember 4.0 may be removed following its release.
 
 > There will be special handling for imports from `@ember/polyfills`. Specifically,
--- a/text/0627-ember-version-4.md
+++ b/text/0627-ember-version-4.md
@@ -1,6 +1,6 @@
 - Start Date: 2020-05-15
 - Relevant Team(s): All teams
-- RFC PR: (after opening the RFC PR, update this with a link to it and update the file name)
+- RFC PR: https://github.com/emberjs/rfcs/pull/627
 - Tracking: (leave this empty)
 
 # Ember 4.0
@@ -8,8 +8,9 @@
 ## Summary
 
 > Ember.js, Ember-Data, and Ember-CLI will release a new major version, 4.0,
-approximately six months from today, somewhere around mid November. Version
-3.23 will be the last version of the 3.x cycle, and will become an LTS release.
+at least six months from today, possibly as soon as December, but this date
+may slip for a variety of reasons.
+The last version in the 3.x cycle of each of these projects will be an LTS.
 In order to prevent a rush of deprecations submitted to Ember prior to the major
 version bump, any new proposed deprecations must target Ember 5.0.
 
@@ -27,12 +28,15 @@ features that IE11 does not support.
 1. Supporting IE11 adds a significant maintenance overhead for the Ember project.
 
 > We are now able to drop support for IE11 since major enterprises such as LinkedIn
-are doing so. However, we are waiting six months in order for companies to notify
+are doing so. However, we are waiting at least six months in order for companies to notify
 clients and make appropriate plans. Developers that still need to support IE11 may
 stay on Ember 3.x indefinitely, and it will be a supported LTS until Ember 4.4 becomes LTS,
 approximately mid May 2021.
 
 ## Detailed design
+
+> In a separate RFC, a new browser support policy will be introduced for the 4.x cycle
+and possibly beyond.
 
 > No new features will be introduced in Ember 4.0. Instead, 
 all deprecated functionality targeting Ember 4.0 may be removed following its release.


sandstrom:
Awesome!

Set a minimum Edge version of 79 for Ember 4.0, so that we don't have to deal with old edge stuff. Will save everyone a ton of time & effort.

Edge is evergreen so jumping straight to Edge 79 minimum shouldn't be a problem!

jrjohnson:
I'm not familiar with why some things needed to be pollyfilled in IE11 and others can be provided by babel and app specific `targets.js` file. I'm guessing there are other browsers who's support has been guaranteed by IE11 compatibility, does this mean ember gets a minimum ES version standard in 4.0? Or does babel cover anything that isn't one of these IE11 issues?

I'm thinking about [Firefox ESR)[https://www.mozilla.org/en-US/firefox/enterprise/] which is evergreen, but trends behind FF and is required for our app, but I'm sure there are others.

pzuraq:
I think rather than focusing on 4.0, we should likely focus on the new browser support policy and deprecating support for legacy browsers. Then, we can cut a 4.0 release at any time, and it will lead to us removing the support for deprecated browsers.

@sandstrom I don't think we should specify minimums here, the support policy should probably be based on either:

* Current - n versions for some set of browsers (this could include Firefox ESR, Opera, etc)
* Or usage statistics

This way we don't need to continually drop support in new releases, we can be evergreen along with the browsers.

@jrjohnson It really depends on the feature. Some features, like new global classes (e.g. `Map` and `Symbol`) bring a large amount of code and need to be polyfilled directly. Historically, Babel did not include this polyfill by default, because you had to include the _entire_ polyfill, not just the pieces that you used. This has changed recently, Babel tooling is getting much better, and with Embroider we are getting close to being able to include the polyfill based on usage rather than in its entirety.

Babel has also increasingly _assumed_ that the polyfill will be present. There was some pretty major breakage in the upgrade to Babel 7 in some cases because of this. So, I think part of the policy change should be that we also assume the presence of the polyfill, and enable it by default. This way Ember can use polyfilled features, like Symbol. Given we won't be supporting legacy browsers anymore, the size of the polyfill will likely be less of an issue, and when Embroider lands it'll become completely pay-go, so I think this would probably be fine.

wycats:
I agree with @pzuraq that we should focus on the next iteration of our browser support policy first, which will naturally have implications for Ember 4.0.

From my perspective, LinkedIn's IE11 needs are relevant to the Ember community because they provide a strong signal about where conservative companies are at with IE11 support. In the context of a browser support RFC, I'd want to hear from more companies across the spectrum about how IE11 support fits into their plans, both for good and ill.

Gaurav0:
I agree that we need a new browser support policy. I am happy to add it here or to write a separate RFC.

However, I feel the reasons stated in "Motivation" section still stand, that we really should drop IE11 support sooner rather than later. I feel that "deprecating" IE11 support brings us the worst of both worlds, requiring us to maintain support for an increasingly rarely used browser and having it being used less often and therefore being less tested, especially in community addons.

pzuraq:
@Gaurav0 I think we all agree on that point, and the tentative timeline you laid out is definitely in the range we would like. After the recent issues with the Octane release however, where we declared that 3.14 would be the release version for Octane, and then it slipped, I think concrete timelines and commitments are something we want to stay away from. 

Personally, I'd basically want to say "we have a new browser support policy, and we're going to major version ASAP. Looks like it'll be November or so, maybe a little later depending on logistics." Historically, we didn't have an RFC for v3.0, and v2.0 only had a sort of roadmap RFC which did not specify timelines or dates, so I'm not sure it's necessary to have an RFC for 4.0 specifically.

Gaurav0:
@pzuraq Yes. I'm happy to update the RFC to state the timeline along the lines you suggest.

The reason I felt the need for an RFC and a timeline is that people need to communicate with customers and plan. I also wanted a place to propose the changes with regards to `@ember/polyfills` and to discuss what the exact timeline should be. Also I generally feel we should be doing this sort of thing in public with input from community members like myself.

Let's discuss the new browser support policy.

As a quick strawman proposal, I propose we copy LinkedIn's: https://www.linkedin.com/help/linkedin/answer/4135

This means we will support the last two revisions of Chrome, Firefox, Safari, and Edge, and on mobile, Safari, Chrome, and Android.

Specifically this would mean we would not have to support Edge prior to Chromium.

jrjohnson:
@Gaurav0 that's a good strawman, but let me take a shot at it!  

Ember's support policy should be 18 months more conservative than what an app would potentially want to support. So if firefox does [5 releases a year](https://wiki.mozilla.org/Release_Management/Calendar) Ember should aim to support the last 7-10 major versions. Given that we can't guarantee the cadence of every browser vendor I think tying this to time is better that tying it to a version number. 

Translates to:

Ember supports every version of FF, Safari, and Chrome released in the past 18 months unless they are EOLed by the vendor and will only drop support for a specific browser version if it has fallen out of this time frame for at least one LTS release.

rwjblue:
> I think rather than focusing on 4.0, we should likely focus on the new browser support policy and deprecating support for legacy browsers. 

Yes, I strongly agree with this. The primary goal should be defining what our browser support policy should be come (including deprecations for those browsers that we currently support that don't fit into that policy) **not** when we should or shouldn't do a breaking change release.

sandstrom:
Agree, a browser support policy would be good!

For our company, we'd rather have Ember sacrifice some backwards compatibility in favour of progress. I'd also add my observation from talking to customers (companies with 500-10,000 employees) over the past 5-10 years is that the discussion has changed quite a bit over the past few years:

1. They used to update browser in their fleet much less often, but for security reasons they push updates much more frequently now.
2. Companies that need an older browser for internal systems will usually have another one installed, that is recent.
3. Even when not needed because of (2) above, most end-users (employees) have several browsers installed, usually Chrome/Edge on Windows and Chrome/Safari on Mac.
4. Asking employees to switch browser isn't perceived as much of a hassle these days, but it used to be.
5. Support for evergreen browsers, most recent release + 1-2 previous versions seems to be fairly standard policy in companies.
6. SaaS-companies are a bit more aggressive in dropping support for older browsers these days, probably reflecting the change in sentiment mentioned in the previous points.

I'd also add that if we could time things such that the next major release of Ember won't support Edge-before-chromium I think that would be neat and safe everyone a lot of work.

***

**Here are the schedules I could find:**

Chrome:
https://chromiumdash.appspot.com/schedule

Firefox:
https://hacks.mozilla.org/2019/09/moving-firefox-to-a-faster-4-week-release-cycle/

Safari:
https://en.wikipedia.org/wiki/Safari_version_history
https://developer.apple.com/documentation/safari_release_notes

Edge:
https://www.microsoftedgeinsider.com/en-us/download
https://techcommunity.microsoft.com/t5/discussions/edge-release-schedule/m-p/1123280

Opera:
https://en.wikipedia.org/wiki/History_of_the_Opera_web_browser#Timeline_of_releases

Gaurav0:
> Ember supports every version of FF, Safari, and Chrome released in the past 18 months unless they are EOLed by the vendor and will only drop support for a specific browser version if it has fallen out of this time frame for at least one LTS release.

FF, Safari, Chrome, and Edge currently EOL a revision as soon as the next one is released, never supporting more than one revision at a time.

Gaurav0:
> Yes, I strongly agree with this. The primary goal should be defining what our browser support policy should be come (including deprecations for those browsers that we currently support that don't fit into that policy) not when we should or shouldn't do a breaking change release.

While I am ok with changing the policy for Ember 4.0 going forward, we made a commitment to support IE11 during the entire 3.x cycle and therefore we should increment our major version when we drop this commitment.

See https://github.com/emberjs/rfcs/blob/49e72790bdb1408371e76503ea46800bafefa651/text/0252-browser-support-changes.md

luxferresum:
> Set a minimum Edge version of 79 for Ember 4.0,

please dont!
This thing is basically just out and far from widely spread. Especially in enterprise situations edge is *not* fully evergreen! Dropping the old Edge this year is just too early.

NullVoxPopuli:
Should we have a separate browser support policy RFC?

rwjblue:
> Should we have a separate browser support policy RFC?

IMHO, yes.

Gaurav0:
@pzuraq I've updated the RFC to a more flexible "at least six months from now" timeline.

jrock2004:
Are you implying by dropping support that an Ember app would no longer run on IE11? As long as IE11 is still an OS default install, I think deprecating IE11 should not be done at this time.

Gaurav0:
@jrock2004 IE11 is not a default install on any OS. Windows 10 comes with Edge by default.

jrock2004:
@Gaurav0 on my windows 10 machine I can click on start menu and type internet explorer and it comes up

Gaurav0:
@jrock2004 What I mean is, on a fresh install of windows 10, Edge is the default browser.

Gaurav0:
> Are you implying by dropping support that an Ember app would no longer run on IE11?

I am implying that new applications written in Ember 4.0 or applications rebuilt after upgrading to Ember 4.0 **may** not run correctly on IE11.

jrock2004:
> > Are you implying by dropping support that an Ember app would no longer run on IE11?
> 
> I am implying that new applications written in Ember 4.0 or applications rebuilt after upgrading to Ember 4.0 **may** not run correctly on IE11.

So then you would have to stop users of IE11 to use your site? 

Are there any other frameworks that have done this?

Gaurav0:
> So then you would have to stop users of IE11 to use your site?

No. If you would like to continue supporting IE11, you may continue to use Ember 3.x for as long as you'd like. Or you may try to "bring your own polyfills" as with other frameworks. There may be features in Ember 4.x that cannot be polyfilled, and you might have to avoid them.

> Are there any other frameworks that have done this?

Most other frameworks already require you to "bring your own polyfills". For example, see https://create-react-app.dev/docs/supported-browsers-features/

wycats:
@Gaurav0 What do you think about renaming this "Timeline for Ending IE11 Support" or something like that?

Gaurav0:
@wycats What is the motivation behind changing the name?

I suppose my feelings are that I intended for this RFC to advocate for dropping support for IE11 in a manner consistent with semver and our previous guarantees, plus provide a path for those that wish to continue to support IE11 without deprecations while also handling the breaking changes involved with removing the `@ember/polyfills` code.

I have, in previous comments, expressed my concern over the alternatives of "deprecating support" for IE11 or "dropping support" without signaling it with a 4.0 release. I could expand on these further in the RFC itself if you think that would be helpful.

wycats:
@Gaurav0 The reason I think it's sensible to change the name is that dropping IE11 support is only one of a number of changes that will likely be made in the 4.0 release. I think it's fine for this RFC to clearly state that it would be a semver breaking change and would need to happen along with a 4.0 release. We'll ultimately *need* a 4.0 RFC, and it will of course incorporate this RFC into it, if this RFC is approved.

(personal note: I would be very surprised if this RFC, in some form, was not approved)

Gaurav0:
@wycats ok, will update accordingly.

wycats:
@Gaurav0 awesome! Thank you so much.

armansa:
https://techcommunity.microsoft.com/t5/microsoft-365-blog/microsoft-365-apps-say-farewell-to-internet-explorer-11-and/ba-p/1591666

sandstrom:
Now that Microsoft 365 aren't even supporting IE11 anymore, maybe it's time for Ember to pull the plug?

image

wycats:
@sandstrom This change will be one of the major changes in Ember 4.0. We have to get all our ducks in a row on finalizing deprecations (especially ones that were already approved), but I can say with confidence that IE11 is going, going, GOOONE.

jrock2004:
Are we going to have any guides to help users that still might have to support IE11? I assume us dropping support will require work to still support it because technically a person could still visit the site on that browser

pzuraq:
@jrock2004 in dropping support, we'll likely using browser features in Ember that are unpolyfillable in IE11, such as native [Proxy](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy). If someone needs their Ember application to run in IE11, they will need to remain on Ember v3 because of this. 

We could definitely add a guide or create an addon for detecting if you are visiting using IE11, and showing a "please visit in a supported browser!" type page in that case, though, if that's what you mean.

jrock2004:
@pzuraq Well then I think that needs to be stated that if you still need IE11 support you cannot upgrade to 4.x

jrock2004:
So if you upgrade and someone from IE11 tries to hit your site, I guess the app blows up. Maybe we need to add some script tag to index html to check if IE 11, and if so, do not load the app but just load an H1 or something telling you to use a newer browser

locks:
> @pzuraq Well then I think that needs to be stated that if you still need IE11 support you cannot upgrade to 4.x

@jrock2004 yup! this is what dropping support means and what was done for Ember 2.0 and IE8 support for example.

rwjblue:
> So if you upgrade and someone from IE11 tries to hit your site, I guess the app blows up. Maybe we need to add some script tag to index html to check if IE 11, and if so, do not load the app but just load an H1 or something telling you to use a newer browser

Ya, I think there are a few different ways to handle this:

* webserver sniffs user agent, and...
  * sends request to a known good version of the site
  * sends request to a static "please use a modern browser" page
* use "tricks" to make your normal scripts not seen by IE11 (e.g. [`module`/`nomodule` trick](https://philipwalton.com/articles/deploying-es2015-code-in-production-today/))

Definitely seems good to document that this can be done (and possibly linking to the article above).

simonihmig:
After reading the RFC text for the first time, and skimming through the comments here, I am still a bit confused tbh, as it's still not quite clear what exactly the single concern of this RFC is. Is it the transition to Ember 4.0 (do we need an RFC for this?), or is it deprecating IE11 support? The RFC's summary only refers to the former, while the motivation only refers to the latter.

If it's about IE11 - as others have said before - I think it can be "just a regular deprecation RFC", i.e. it does not really have to provide details for how the transition to Ember 4.0 needs to be done, because the same mechanics apply for the many other (already merged) deprecation RFCs, right?

Also - very much nitpicking here - could you reformat the text a bit to not use block-quotes everywhere? This got be confused a bit at the beginning, as it suggested the text parts are not things the RFC says, but rather something it refers to...

> No. If you would like to continue supporting IE11, you may continue to use Ember 3.x for as long as you'd like. 

Let's be honest here: this sounds like a terrible idea. Like you will be stuck on a version that receives no bug fixes anymore, and - maybe even more importantly - no bug fixes for addons, that I expect will more or less quickly start adopting Ember 4.0 as their minimum supported version. And virtually nobody will back port bug fixes to older releases just so that some users get them for IE11 support. I think the more reasonable answer IMHO is that if you need IE11 support, Ember is not for you anymore.

And btw, I am very much in favor of deprecating IE11! 😬

Gaurav0:
> After reading the RFC text for the first time, and skimming through the comments here, I am still a bit confused tbh, as it's still not quite clear what exactly the single concern of this RFC is. Is it the transition to Ember 4.0 (do we need an RFC for this?), or is it deprecating IE11 support? The RFC's summary only refers to the former, while the motivation only refers to the latter.

This was originally written as the transition to 4.0, with dropping IE11 support as the primary motivation.

> If it's about IE11 - as others have said before - I think it can be "just a regular deprecation RFC", i.e. it does not really have to provide details for how the transition to Ember 4.0 needs to be done, because the same mechanics apply for the many other (already merged) deprecation RFCs, right?

Actually, no. We cannot just deprecate the IE11 polyfills, because they will be needed for anyone sticking with IE11 support.

> Also - very much nitpicking here - could you reformat the text a bit to not use block-quotes everywhere? This got be confused a bit at the beginning, as it suggested the text parts are not things the RFC says, but rather something it refers to...

This RFC is kinda stuck. I've been asked to stop contributing to Ember. If someone appropriate would like to take this over, they can. @pzuraq @wycats ?

> > No. If you would like to continue supporting IE11, you may continue to use Ember 3.x for as long as you'd like.
> 
> Let's be honest here: this sounds like a terrible idea. Like you will be stuck on a version that receives no bug fixes anymore, and - maybe even more importantly - no bug fixes for addons, that I expect will more or less quickly start adopting Ember 4.0 as their minimum supported version. And virtually nobody will back port bug fixes to older releases just so that some users get them for IE11 support. I think the more reasonable answer IMHO is that if you need IE11 support, Ember is not for you anymore.

This isn't necessarily true. Ember 1.13.14 and 1.13.15 were released following Ember 2.0 to address fixes. The last version of 3.0 will be an LTS, so it should be supported for a period of six months. It can be decided to support it longer if the community chooses.

> And btw, I am very much in favor of deprecating IE11! 😬

pzuraq:
I have been working on the next iteration of the new browser support policy and will be posting some updates soon! Sorry for the delays!

pzuraq:
I'm going to close this for now as we're working on a followup RFC. Will comment here as soon as its up!

Kerrick:
That followup RFC seems to have been this, which has been merged: https://github.com/emberjs/rfcs/pull/685

PR #628, opened 2020-05-18
💅 Add Prettier 💇

rwjblue:
[rendered](https://github.com/emberjs/rfcs/blob/pretty-please!/text/0628-prettier.md)

### DIFF
--- a/text/0628-prettier.md
+++ b/text/0628-prettier.md
@@ -133,7 +133,12 @@ and it will replace the previously created commits.
 
 We do not currently discuss linting or stylistic formatting in either guides.emberjs.com or cli.emberjs.com.
 
-This RFC proposes adding a new subsection that discusses linting under [`Basic Use` section of the CLI guides](https://cli.emberjs.com/release/basic-use/).
+This RFC proposes adding a new subsection that discusses linting under [`Basic
+Use` section of the CLI guides](https://cli.emberjs.com/release/basic-use/). In
+addition, when this RFC is implemented we should do a throughough audit of the
+`cli.emberjs.com` and `guides.emberjs.com` content to ensure that all code
+snippets are formatted correctly (e.g. using the Prettier specific formatting
+introduced here).
 
 ## Drawbacks
 
--- a/text/0628-prettier.md
+++ b/text/0628-prettier.md
@@ -80,21 +80,26 @@ be updated to ensure that they satisfy these new linting rules requirements.
 
 #### `package.json` scripts
 
-The `app` and `addon` blueprints will be updated to add the following additional entries to `scripts`:
+The `app` and `addon` blueprints will be updated to add the following
+additional entries to `scripts`:
 
 ```json
 {
   "scripts": {
+    "lint": "npm-run-all --aggregate-output --continue-on-error --parallel 'lint:!(fix)'",
     "lint:fix": "npm-run-all --aggregate-output --continue-on-error --parallel lint:*:fix",
     "lint:js:fix": "eslint . --fix",
     "lint:hbs:fix": "ember-template-lint . --fix"
   }
 }
 ```
+The `lint:fix`, `lint:js:fix`, `lint:hbs:fix` scripts are new (introduced with
+this RFC), and the `lint` script will be updated to ensure that it avoids
+running the new `lint:fix`.
 
-This configuration is specifically intending to allow users to add additional linters
-(e.g. `stylelint` or `markdownlint`) by adding scripts for them, and they would
-automatically be rolled up into `lint:fix`.
+This configuration is specifically intending to allow users to add additional
+linters (e.g. `stylelint` or `markdownlint`) by adding scripts for them, and
+they would automatically be rolled up into `lint:fix`.
 
 ### Codemod
 
--- a/text/0628-prettier.md
+++ b/text/0628-prettier.md
@@ -1,0 +1,135 @@
+- Start Date: 2020-05-18
+- Relevant Team(s): Ember CLI
+- RFC PR: https://github.com/emberjs/rfcs/pull/628
+- Tracking: (leave this empty)
+
+# Add Prettier
+
+## Summary
+
+This RFC proposes adding [Prettier](https://prettier.io/) to the blueprints
+that back `ember new` and `ember addon`.
+
+## Motivation
+
+Using Prettier removes the ever present and ongoing stylistic debates that tend
+to happen on teams. Prettier is incredibly freeing for both developers **and**
+reviewers of a codebase. The developer can author in whatever format they find
+easiest to type out, avoiding needless wasted time trying to get that
+indentation _just_ right, and either setup an automated "format on save"
+operation in their editor or use a quick command line invocation to `--fix` all
+files at once.
+
+This dove tails very nicely with Embers deeply ingrained goal of using shared
+solutions to solve common problems, and aligns very well with the general trend
+to play better with the general JavaScript ecosystem.
+
+## Detailed design
+
+Due to the pre-existing design (and improvements added to the blueprints in
+[ember-cli/rfcs#121](https://github.com/emberjs/rfcs/blob/master/text/0121-remove-ember-cli-eslint.md))
+implementation is very straight forward. The general idea is that we will
+update the `app` and `addon` blueprints to add a few Prettier related packages
+to the `package.json`, update the linting configuration to enforce Prettier
+styles, and update the other blueprint code to ensure that its output is
+Prettier compatible.
+
+### Packages
+
+The following packages will be added to the `package.json` of both `app` and `addon` blueprints:
+
+* [prettier](https://www.npmjs.com/package/prettier) - The main package (used by the other packages).
+* [eslint-config-prettier](https://www.npmjs.com/package/eslint-config-prettier) - Disables stylistic ESLint rules that would conflict with the Prettier output.
+* [eslint-plugin-prettier](https://www.npmjs.com/package/eslint-plugin-prettier) - Enables `eslint-config-prettier`s `recommended` set, and adds the `prettier/prettier` rule to enforce styles.
+
+### Configuration Changes
+
+#### ESLint
+
+The `.eslintrc.js` that is generated will be updated to add:
+
+```js
+{
+  extends: ['plugin:prettier/recommended'],
+}
+```
+
+#### Prettier
+
+A new `prettier` section will be added to the `package.json` file with the following contents:
+
+```json
+{
+  "prettier": {
+    "singleQuotes": true
+  }
+}
+```
+
+There is a [small number of configuration
+options](https://prettier.io/docs/en/options.html) that can be used in this
+file, but we are intentionally avoiding modifying the default values for things
+that are not _nearly_ universally agreed on in the Ember ecosystem.
+
+### Blueprint Changes
+
+In general, it is recommended that all blueprints provided by addons should
+satisfy the default linting configuration of a new Ember application. As such
+the blueprints provided by `ember-cli`, `ember-source`, and `ember-data` will
+be updated to ensure that they satisfy these new linting rules requirements.
+
+#### `package.json` scripts
+
+The `app` and `addon` blueprints will be updated to add the following additional entries to `scripts`:
+
+```json
+{
+  "scripts": {
+    "lint:fix": "npm-run-all --aggregate-output --continue-on-error --parallel lint:*:fix",
+    "lint:js:fix": "eslint . --fix",
+    "lint:hbs:fix": "ember-template-lint . --fix"
+  }
+}
+```
+
+This configuration is specifically intending to allow users to add additional linters
+(e.g. `stylelint` or `markdownlint`) by adding scripts for them, and they would
+automatically be rolled up into `lint:fix`.
+
+### Codemod
+
+In order to make the migration as simple as possible, a codemod will be created: `ember-cli-prettier-codemod`.
+
+The codemod will do the following:
+
+* Install the new/required dependencies
+* Migrate an application / addon to the new linting configuration
+* Commit _just_ the configuration changes.
+* Run the new `lint:fix` script to update all files to the new format.
+* Commit just the auto-fixed output.
+
+While this seems quite simple (and in fact it is pretty easy), leveraging a
+codemod will make it much easier to deal with rebasing large Prettier migration
+pull requests. If one of those pull requests gets out of date (e.g. due to a
+conflict in one of the files) you can simply re-run the codemod on the branch
+and it will replace the previously created commits.
+
+## How we teach this
+
+We do not currently discuss linting or stylistic formatting in either guides.emberjs.com or cli.emberjs.com.
+
+This RFC proposes adding a new subsection that discusses linting under [`Basic Use` section of the CLI guides](https://cli.emberjs.com/release/basic-use/).
+
+## Drawbacks
+
+The largest drawback is generally the cost of the _initial_ migration to Prettier. That migration tends to be quick, but it can cause pain for folks maintaining long lived branches.
+
+## Alternatives
+
+Do nothing.
+
+## Unresolved questions
+
+> In general, users of `yarn` could be perfectly happy with `yarn lint:js --fix` / `yarn lint:hbs --fix` (which works today). Should we still add the `lint:js:fix`/`lint:hbs:fix` scripts?
+
+I went with "yes" here in the initial RFC prose, as I'd general prefer to _reduce_ the differences between users of `npm` and `yarn` over time.
--- a/text/0628-prettier.md
+++ b/text/0628-prettier.md
@@ -173,7 +173,7 @@ introduced here).
 
 ## Drawbacks
 
-The largest drawback is generally the cost of the _initial_ migration to Prettier. That migration tends to be quick, but it can cause pain for folks maintaining long lived branches.
+> The largest drawback is generally the cost of the _initial_ migration to Prettier. That migration tends to be quick, but it can cause pain for folks maintaining long lived branches.
 
 ## Alternatives
 
--- a/text/0628-prettier.md
+++ b/text/0628-prettier.md
@@ -61,7 +61,7 @@ A new `prettier` section will be added to the `package.json` file with the follo
 ```json
 {
   "prettier": {
-    "singleQuotes": true
+    "singleQuote": true
   }
 }
 ```
--- a/text/0628-prettier.md
+++ b/text/0628-prettier.md
@@ -71,6 +71,12 @@ options](https://prettier.io/docs/en/options.html) that can be used in this
 file, but we are intentionally avoiding modifying the default values for things
 that are not _nearly_ universally agreed on in the Ember ecosystem.
 
+#### Git
+
+The `.gitignore` will be updated to add `.eslintcache` so that when using
+`eslint --cache` the cache file itself won't be added to the repositories
+history.
+
 ### Blueprint Changes
 
 In general, it is recommended that all blueprints provided by addons should
@@ -88,14 +94,18 @@ additional entries to `scripts`:
   "scripts": {
     "lint": "npm-run-all --aggregate-output --continue-on-error --parallel 'lint:!(fix)'",
     "lint:fix": "npm-run-all --aggregate-output --continue-on-error --parallel lint:*:fix",
+    "lint:js": "eslint . --cache",
     "lint:js:fix": "eslint . --fix",
     "lint:hbs:fix": "ember-template-lint . --fix"
   }
 }
 ```
-The `lint:fix`, `lint:js:fix`, `lint:hbs:fix` scripts are new (introduced with
-this RFC), and the `lint` script will be updated to ensure that it avoids
-running the new `lint:fix`.
+
+A few callouts here:
+
+* The `lint:fix`, `lint:js:fix`, `lint:hbs:fix` scripts are new (introduced with this RFC)
+* The `lint` script will be updated to ensure that it avoids running the new `lint:fix` script
+* The `lint:js` script will be updated to add caching
 
 This configuration is specifically intending to allow users to add additional
 linters (e.g. `stylelint` or `markdownlint`) by adding scripts for them, and
--- a/text/0628-prettier.md
+++ b/text/0628-prettier.md
@@ -56,20 +56,43 @@ The `.eslintrc.js` that is generated will be updated to add:
 
 #### Prettier
 
-A new `prettier` section will be added to the `package.json` file with the following contents:
+A `.prettierrc.js` will be added with the following contents:
 
-```json
-{
-  "prettier": {
-    "singleQuote": true
-  }
-}
+```js
+module.exports = {
+  singleQuote: true,
+};
 ```
 
 There is a [small number of configuration
 options](https://prettier.io/docs/en/options.html) that can be used in this
 file, but we are intentionally avoiding modifying the default values for things
 that are not _nearly_ universally agreed on in the Ember ecosystem.
+
+A `.prettierignore` will be added to match the `.eslintignore` contents:
+
+```
+# unconventional js
+/blueprints/*/files/
+/vendor/
+
+# compiled output
+/dist/
+/tmp/
+
+# dependencies
+/bower_components/
+/node_modules/
+
+# misc
+/coverage/
+!.*
+
+# ember-try
+/.node_modules.ember-try/
+/bower.json.ember-try
+/package.json.ember-try
+```
 
 #### Git
 
--- a/text/0628-prettier.md
+++ b/text/0628-prettier.md
@@ -141,7 +141,23 @@ The largest drawback is generally the cost of the _initial_ migration to Prettie
 
 ## Alternatives
 
-Do nothing.
+> Why not adopt [standard](https://standardjs.com/) instead of Prettier?
+
+Prettier is **much** more popular with [9,249,847 weekly
+downloads](https://www.npmjs.com/package/prettier) vs [265,658 weekly
+downloads](https://www.npmjs.com/package/standard) (we all know these download
+numbers don't mean a ton, but the order of magnitude can tell us something),
+and (aside from the defaulting of `singleQuotes`) is much more aligned with the
+Ember ecosystems inherent preferences.
+
+Additionally, a very large number of the Ember ecosystem
+projects are _already_ using Prettier internally. These include `ember-source`,
+`ember-cli`, `ember-data`, `@ember/test-helpers`, `eslint-plugin-ember`, the
+various packages composing the rendering engine, `@glimmer/component`,
+`ember-template-lint`, and many more. Additionally a large number of community
+maintained addons are also using it already ([here is a
+listing](https://emberobserver.com/code-search?codeQuery=prettier&fileFilter=package.json&sort=score&sortAscending=false)
+using EmberObservers awesome code search feature).
 
 ## Unresolved questions
 
--- a/text/0628-prettier.md
+++ b/text/0628-prettier.md
@@ -107,6 +107,14 @@ satisfy the default linting configuration of a new Ember application. As such
 the blueprints provided by `ember-cli`, `ember-source`, and `ember-data` will
 be updated to ensure that they satisfy these new linting rules requirements.
 
+In order to ensure blueprint output follows each individual projects custom
+stylistic linting settings as much as possible. The following will be updated
+to run `lint:fix` (when available) after their existing functionality:
+
+* `ember generate`
+* `ember init`
+* `ember-cli-update`
+
 #### `package.json` scripts
 
 The `app` and `addon` blueprints will be updated to add the following


kpfefferle:
The current `package.json` blueprint [includes a generic `lint` script command](https://github.com/ember-cli/ember-new-output/blob/3c66af7eeee9ecb8f5b766c01b65fb385f393908/package.json#L15) which uses `npm-run-all` to run all `lint:*` script commands.

How would this work with this RFC's proposed `lint:fix` script command, which similarly uses `npm-run-all` to run all `lint:*:fix` script commands? Would running the existing `lint` run both `lint:fix` *and* each individual `lint:*:fix` script, or does something prevent each command from running more than once?

I would assume we'd want to avoid running duplicate lint tasks, but I do like that the wildcards in these script commands conveniently pick up on similarly-named script commands that application authors may add (such as `lint:ts`).

text/0628-prettier.md
```diff
@@ -0,0 +1,135 @@
...
+  "prettier": {
+    "singleQuotes": true
+  }
+}
+```
+
+There is a [small number of configuration
+options](https://prettier.io/docs/en/options.html) that can be used in this
+file, but we are intentionally avoiding modifying the default values for things
+that are not _nearly_ universally agreed on in the Ember ecosystem.
```
ro0gr, line 72: I believe some of agreements exist in [editorconfig](https://github.com/ember-cli/ember-new-output/blob/master/.editorconfig) today. Are we going to respect editorconfig settings with prettier, or should we consider removing editorconfig at some point?

text/0628-prettier.md
```diff
@@ -0,0 +1,135 @@
...
+```
+
+#### Prettier
+
+A new `prettier` section will be added to the `package.json` file with the following contents:
+
+```json
+{
+  "prettier": {
+    "singleQuotes": true
```
kategengler, line 64: The option in the [documentation for prettier](https://prettier.io/docs/en/configuration.html) is `singleQuote` (no plural).

lifeart:
Should we add prettier for handlebars templates too?

rwjblue:
> How would this work with this RFC's proposed lint:fix script command, which similarly uses npm-run-all to run all lint:*:fix script commands? Would running the existing lint run both lint:fix and each individual lint:*:fix script, or does something prevent each lint:*:fix command from running more than once?

Great question @kpfefferle! I pushed an update in a32ad2d to clarify and show how to fix (tldr; you change from `lint:*` to `lint:!(fix)`).

text/0628-prettier.md
rwjblue: > Are we going to respect editorconfig settings with prettier?

Yep, Prettier already parses and uses the `.editorconfig` values for some of its defaults. It is a bit buried, but see [these docs](https://prettier.io/docs/en/api.html#prettierresolveconfigfilepath--options):

> If options.editorconfig is true and an .editorconfig file is in your project, Prettier will parse it and convert its properties to the corresponding prettier configuration. This configuration will be overridden by .prettierrc, etc. Currently, the following EditorConfig properties are supported:
> 
> * end_of_line
> * indent_style
> * indent_size/tab_width
> * max_line_length

text/0628-prettier.md
rwjblue: D'oh! 🤦  thank you @kategengler!

rwjblue: Updated in fa51038.

rwjblue:
@lifeart:

> Should we add prettier for handlebars templates too?

I don't think we should do that at this time, but as the formatting behaviors continue to improve in `.hbs` files we should reconsider ([follow along with that progress here](https://github.com/jgwhite/prettier/issues/1)). IMHO, doing this should follow _essentially_ the same pattern that this RFC currently lays out for ESLint (adding a `ember-template-lint-plugin-prettier` which will implement a `--fix`).

Alonski:
@lifeart Prettier for HBS files isn't completely ready for the limelight yet in my opinion. There are still a few kinks to remove before we can do this. :)

lifeart:
Prettier quest: https://github.com/prettier/prettier/issues/4908

Gaurav0:
The RFC adds lint:hbs:fix. This probably should be added regardless of whether prettier configuration is added.

In my experience with prettier, it generates lots of stylistic errors that often obscure "real" problems as well as issues I might temporarily not want autofixed (like console.logs). Is there a way to temporarily disable prettier errors on the command line? If so, can we add that to the config?

rwjblue:
@Gaurav0 

> The RFC adds lint:hbs:fix. This probably should be added regardless of whether prettier configuration is added.

Sure, but we have to add it somewhere. I guess I could remove `lint:hbs:fix` from this RFC because it is not prettier related, but that doesn't really seem like a great option to me.

> In my experience with prettier, it generates lots of stylistic errors that often obscure "real" problems

Perhaps, but FWIW, that is also the part that is super freeing for developers. As I mentioned in the RFC, you just author however you want to and let Prettier take care of the final formatting. Editor settings such as format on save or suggested code actions make it very trivial to automatically apply, and if you don't want to fiddle with your editor settings the `yarn lint:fix` is only a short terminal hop away.

> as well as issues I might temporarily not want autofixed (like debugger statements and console.logs).

Prettier will not remove `debugger` statements (and neither will ESLint as of https://github.com/eslint/eslint/pull/10509). The experience you had was almost certainly the issue reported in https://github.com/eslint/eslint/issues/10242.

> Is there some way we can add a git precommit hook to the boilerplate as well to automatically run lint:js:fix? I often break prettier linting with only minor touchups and forget to rerun it.

I personally dislike precommit hooks as a rule, and violently dislike those that rewrite file contents. In _theory_ we could add one that validates only ([see this example](https://github.com/malleatus/nyx/blob/87cdd0042df874a58f1a12b4d4d6f191f1d04f4d/package.json#L74-L81)), but I'd prefer not to.

jherdman:
This was/is a holy war for my team. Including Prettier was massively disrupting at first, though we have experimented with it on greenfield projects. That said, why Prettier? Why not Standard? Etc.

rwjblue:
@jherdman 

> Including Prettier was massively disrupting at first

Ya, I totally agree that initial adoption is definitely the most difficult part of the process. One thing to keep in mind is that this RFC is specifically talking about adding this setup to new apps/addons (e.g. greenfield projects). While it is true that `ember-cli-update` would pick these changes up and attempt to apply them to existing apps/addons, it is easy enough for the person doing the upgrade to clear those specific changes and hold off on them (once you do that for the first version that includes the changes, `ember-cli-update` is smart enough to avoid suggesting them again).

> Why Prettier? Why not Standard? Etc.

First up, I'll be perfectly frank: I chose Prettier for this RFC because I prefer it and have the most experience with it. Aside from my personal preferences Prettier is **much** more popular with [9,249,847 weekly downloads](https://www.npmjs.com/package/prettier) vs [265,658 weekly downloads](https://www.npmjs.com/package/standard) (we all know these download numbers don't mean a ton, but the order of magnitude can tell us something), and (aside from the defaulting of `singleQuotes`) is much more aligned with the Ember ecosystems inherent preferences.

Another data point here is that a very large number of the Ember ecosystem projects are _already_ using Prettier internally. These include `ember-source`, `ember-cli`, `ember-data`, `@ember/test-helpers`, `eslint-plugin-ember`, the various packages composing the rendering engine, `@glimmer/component`, `ember-template-lint`, and many more. Additionally a large number of community maintained addons are also using it already ([here is a listing](https://emberobserver.com/code-search?codeQuery=prettier&fileFilter=package.json&sort=score&sortAscending=false) using EmberObservers awesome code search feature).

ballPointPenguin:
FWIW I have established a pattern in my apps for using `eslint-config-prettier`. It's nice when prettier rules and linting (and fixing) are managed by eslint, in my experience, so that they are never in conflict and existing tooling around eslint (e.g. Text Editor plugins and pre-commit hooks) just work.

for reference
https://github.com/ballPointPenguin/impulse-sieve/blob/master/package.json
https://github.com/ballPointPenguin/impulse-sieve/blob/master/.eslintrc.js

I also include `standard` config, but that would be more disruptive to hoist onto the general public (no semicolons!)

rwjblue:
@ballPointPenguin

> FWIW I have established a pattern in my apps for using eslint-config-prettier. It's nice when prettier rules and linting (and fixing) are managed by eslint, in my experience, so that they are never in conflict and existing tooling around eslint (e.g. Text Editor plugins and pre-commit hooks) just work.

Yep, totally agree! Thats exactly the setup proposed here. :smiley:

rwjblue:
We discussed this in todays meeting, and are moving into final comment period.

ming-codes:
Doesn’t this overlaps with the current editorconfig?

pichfl:
Big fan. Prettier caused and removed quite a bit of discussion in many of my projects but it was worth it for the less noisy PRs alone.

1. Why `package.json` over `.prettierrc.js`? We have `.eslintrc.js` and so on as well, I think it's a bit odd that this config is moved into `package.json`:

```js
module.exports = {
  singleQuote: true,
};
```

2. While it is still experimental, Prettier does support hbs to some extend. Would it be good to show an example config? 

```js
module.exports = {
  singleQuote: true,
  overrides: [
    {
      files: '**/*.hbs',
      options: {
        singleQuote: false,
      },
    },
  ],
}
```

3. Using single quotes in yaml files will lead to issues in project which use `ember-intl` and other plugins that rely on yaml files with ICU translation syntax, which uses single quotes for escaping (no clue why). 

```js
module.exports = {
  singleQuote: true,
  overrides: [
    {
      files: '**/*{yml,yaml}',
      options: {
        singleQuote: false,
      },
    },
  ],
}
```

All three blocks merged make up the Prettier config that served me well in all recent ember projects.

Gaurav0:
This was introduced 4 days ago and has had considerable feedback. Why is it being rushed into FCP?

rwjblue:
@ming-codes 

> Doesn’t this overlaps with the current editorconfig?

[@ro0gr asked a similar question inline](https://github.com/emberjs/rfcs/pull/628#discussion_r426888426), but the thread is resolved so it doesn't display. Here was my reply:

> > Are we going to respect editorconfig settings with prettier?
> 
> Yep, Prettier already parses and uses the `.editorconfig` values for some of its defaults. It is a bit buried, but see [these docs](https://prettier.io/docs/en/api.html#prettierresolveconfigfilepath--options):
> 
> > If options.editorconfig is true and an .editorconfig file is in your project, Prettier will parse it and convert its properties to the corresponding prettier configuration. This configuration will be overridden by .prettierrc, etc. Currently, the following EditorConfig properties are supported:
> > 
> > * end_of_line
> > * indent_style
> > * indent_size/tab_width
> > * max_line_length

ming-codes:
> This configuration will be overridden by .prettierrc, etc

But in the context of Ember blueprints, are we going to deprecate editorconfig in favor of Prettier? Since prettier config can override certain properties, this can cause conflicts in editor plugins.

rwjblue:
@pichfl 

> Why `package.json` over `.prettierrc.js`? We have `.eslintrc.js` and so on as well, I think it's a bit odd that this config is moved into `package.json`

I don't feel super strongly here, I did originally have it as a `.prettierrc` file with basically that one setting set (note using `.prettierrc` is just JSON, vs your snippet is a CJS module). I ultimately went with moving it to `package.json` to avoid adding a very infrequently modified file to the repo root. Unlike `.eslintrc.js` we really do not expect folks to be modifying the prettier config _basically_ at all after their initial setup, and as such it doesn't really seem like it warrants its own top level config file.

> While it is still experimental, Prettier does support hbs to some extend. Would it be good to show an example config?

Not sure what you mean here, do you mean that we should add configuration for it by default? If so, I disagree: I don't think its quite ready for prime time (see [this comment](https://github.com/emberjs/rfcs/pull/628#issuecomment-630450832) and [my reply](https://github.com/emberjs/rfcs/pull/628#issuecomment-630456516)). I absolutely think a future RFC is likely to add that configuration, but it seems more likely (to me) that it would be done via `ember-template-lint` (via [ember-template-lint-plugin-prettier](https://github.com/ember-template-lint/ember-template-lint-plugin-prettier)) which would be directly in line with how we are doing `*.js`/`*.ts`.

> Using single quotes in yaml files will lead to issues in project which use ember-intl and other plugins that rely on yaml files with ICU translation syntax, which uses single quotes for escaping (no clue why)

Thank you for that! Though I am curious why this caused escaping issues in `.yml` files. Ultimately, this proposal is only setting up Prettier to run on `*.js` files (really whatever you have configured `eslint` to run against), and as such shouldn't require this override.

rwjblue:
@ming-codes 

> But in the context of Ember blueprints, are we going to deprecate editorconfig in favor of Prettier? Since prettier config can override certain properties, this can cause conflicts in editor plugins.

No, `.editorconfig` and Prettier configuration are complimentary. The `.prettierrc` _can_ override settings in `.editorconfig`, but we are not doing that.

rwjblue:
@Gaurav0 

> This was introduced 4 days ago and has had considerable feedback.

Yep, it is quite awesome that we all seem to agree here. 😸 

> Why is it being rushed into FCP?

I don't particularly think anything is being rushed. The point of final comment period is to signal that the core teams that are responsible for a given feature set modified by an RFC think that the RFC is "ready" and that it is a good time for community members that might prefer to avoid all the noise of evolving RFCs to take a look now that the proposal has settled down. That is exactly what is happening here.

tldr; We are🚶 with 🙇 not 🏃 with ✂️.

text/0628-prettier.md
```diff
@@ -0,0 +1,140 @@
...
+
+This RFC proposes adding a new subsection that discusses linting under [`Basic Use` section of the CLI guides](https://cli.emberjs.com/release/basic-use/).
+
+## Drawbacks
+
+The largest drawback is generally the cost of the _initial_ migration to Prettier. That migration tends to be quick, but it can cause pain for folks maintaining long lived branches.
+
+## Alternatives
+
+Do nothing.
```
locks, line 134: standardjs could be added as an alternative and rejected for the reasons mentioned in the thread, so we have the rationale documented.

text/0628-prettier.md
```diff
@@ -0,0 +1,140 @@
...
+
+While this seems quite simple (and in fact it is pretty easy), leveraging a
+codemod will make it much easier to deal with rebasing large Prettier migration
+pull requests. If one of those pull requests gets out of date (e.g. due to a
+conflict in one of the files) you can simply re-run the codemod on the branch
+and it will replace the previously created commits.
+
+## How we teach this
+
+We do not currently discuss linting or stylistic formatting in either guides.emberjs.com or cli.emberjs.com.
```
locks, line 124: While the teaching and documentation resources don't explicitly discuss linting/formatting, I'd like to see some words on making sure the code samples adhere to the prettier formatting. Unfortunately we do not have an automated code sample linting process, so we try to manually guarantee that code samples are free of linting warnings.

text/0628-prettier.md
rwjblue: Ya, great point! Will update.

text/0628-prettier.md
rwjblue: 👍👍 perfect, will do

Gaurav0:
> I don't particularly think anything is being rushed. The point of final comment period is to signal that the core teams that are responsible for a given feature set modified by an RFC think that the RFC is "ready" and that it is a good time for community members that might prefer to avoid all the noise of evolving RFCs to take a look now that the proposal has settled down. That is exactly what is happening here.

4 days from introduction to FCP feels rushed to me. I'm pretty sure it's a record. The advancement of RFC proposals is a nontransparent process to those of us not on the core team, and in this case I feel I did not have adequate time to put together my concerns nor respond to your feedback about the concerns I already posted.

> Is there a way to temporarily disable prettier errors on the command line? If so, can we add that to the config?

This question was not addressed. I would like to be able to look at and concentrate on "real" linting errors while developing.

> I personally dislike precommit hooks as a rule, and violently dislike those that rewrite file contents. In theory we could add one that validates only (see this example), but I'd prefer not to.

Is there another way to handle forgetting to run lint:fix? I believe ember-data already has a validating precommit hook, in order to avoid overuse of CI resources.

Another question:

Can we cache eslint by default? This would improve performance with large codebases. When prettier was first added to ember-data, eslint run time increased significantly.

rwjblue:
@Gaurav0 

> The advancement of RFC proposals is a nontransparent process to those of us not on the core team, and in this case I feel I did not have adequate time to put together my concerns nor respond to your feedback about the concerns I already posted.

I see, maybe we just have a misalignment here. You talk as if FCP means things are a done deal, but that is quite far from the truth. There have been many RFCs that have had very important issues brought up _after_ the final comment period (and those have been modified, withdrawn, etc to accommodate those concerns).

Please take your time, let me know what you think about the proposal and what we can make better here!

> > Is there a way to temporarily disable prettier errors on the command line? If so, can we add that to the config?
>
> This question was not addressed. I would like to be able to look at and concentrate on "real" linting errors while developing.

I don't really understand the question, can you please elaborate on the developer workflow you are referring to? 

You can definitely disable a specific linting rule from the command line via the `--rule` command line option (see [documentation here](https://eslint.org/docs/user-guide/command-line-interface#rule)). 

Something like this should do it:

```
yarn eslint --rule 'prettier/prettier: off' .
```

However, I don't think this should really be part of the developer workflow (and I would be opposed to adding it in this RFC), and instead if you are swapping to the command line to do linting you should just use `yarn lint:fix` to begin with. This indeed does **exactly** what you want ("show me the errors that I should care about") while also making your file beautiful 💅 💇 👯.

> > I personally dislike precommit hooks as a rule, and violently dislike those that rewrite file contents. In theory we could add one that validates only (see this example), but I'd prefer not to.
> 
> Is there another way to handle forgetting to run lint:fix?

Yep, make sure your `npm test` / `yarn test` properly fails when you have incorrectly formatted files.

> I believe ember-data already has a validating precommit hook, in order to avoid overuse of CI resources.

Yep, lots of projects will have tweaks to the default setup. That is absolutely great! I don't think sacrificing developer productivity to save CI time is terribly valuable (your time and my time is worth more than the machines 😺), and our default CI setup for Ember apps and addons would already avoid running many concurrent jobs if linting fails ([see the default Travis](https://github.com/ember-cli/ember-addon-output/blob/c65cc81505ee9826dea9a1ef602edf0f789c440c/.travis.yml#L38-L39)).

> Can we cache eslint by default? This would improve performance with large codebases.

Ya, seems good to me. I'll update with that.

rwjblue:
> > Can we cache eslint by default? This would improve performance with large codebases.
> 
> Ya, seems good to me. I'll update with that.

Added this in ecf524d.

pichfl:
@rwjblue thanks for answering. Prettier [does support config files with file endings](https://prettier.io/docs/en/configuration.html) which I do prefer over guess the right syntax. I used CJS because many other configs in an ember project do so as well. Given how it might be necessary to change up the config a little in the future, I'd rather prefer a separate file than trying to put everything in the `package.json`.

In regards to Glimmer/yaml: I was just wondering if it might make sense to have those two overrides in the blueprint, but commented out to provide a very easy path to enable either. I agree with Glimmer formatting not being ready though, so maybe just the yaml stuff.

Personally, I run prettier for js,yaml,md and anything it supports. 

---

On a different note, we should also add a `.prettierignore` file so `node_modules`, `dist` and `vendor` are not touched.

rwjblue:
@pichfl 

> Prettier does support config files with file endings which I do prefer over guess the right syntax. I used CJS because many other configs in an ember project do so as well.

Oh! I wasn't saying that I disliked using the CJS setup (I have a number of repos that use that setup), sorry about that aside in my reply. I still generally like the proposal here (using `package.json`) a bit better, but that preference hinges on the things I mentioned above:

1. We won't be modifying this very much / if at all
2. There are not very many settings that will be configured, and therefore a separate config file is not warranted
3. That the things we need to configure do not need explanation (since `.json` can't have comments). 

If these assumptions are not valid (e.g. your example of `.yml` files needing a different config), we should absolutely consider moving to the stand alone file setup.

I'm curious what other folks' perspective is here, do folks tend to use overrides / changes per file type like that? Either way, I'll add that to the alternatives section in the meantime.

> In regards to Glimmer/yaml: I was just wondering if it might make sense to have those two overrides in the blueprint, but commented out to provide a very easy path to enable either.

Generally speaking, I flag any commented out code while reviewing PRs 🙈. 

While early adopters are working through issues in the printer I don't think we should suggest folks use Prettier for their `.hbs` files, and leaving it there but commented out is just a bit too tempting. I definitely think that we could add a comment pointing to somewhere in the guides (either CLI or main guides) if we decide to go with an extra file for configuration, but other than that I think we should only include configuration that we actually want to run.

> On a different note, we should also add a .prettierignore file so node_modules, dist and vendor are not touched.

A few thoughts:

* Based on [their documentation](https://prettier.io/docs/en/cli.html) it seems that Prettier ignores `node_modules` by default.

* The suggested setup here only runs by way of running `eslint` and therefore follows the `.eslintignore` (which already contains these ignores).

We certainly could add a top level `.prettierignore`, I'm not terribly opposed but in practice I don't use one in most of the repos that I work in. I'm definitely open to being convinced, but I don't really think needing to customize it is common enough to make part of the default stack.

neojp:
I'm also in favor of using a `.prettierc` file over `package.json`.

While I agree on adding the `.prettierignore` file (it also should have the `tmp` directory), if the `.eslinignore` works for prettier, that's good enough for me.

Alonski:
Heads up. I just tried running this locally on Windows:
`npm-run-all --aggregate-output --continue-on-error --parallel 'lint:!(fix)'`
I got an error:
`' was unexpected at this time.`
Fix is to escape `\"`
`npm-run-all --aggregate-output --continue-on-error --parallel \"lint:!(fix)\"`

rwjblue:
Awesome, thanks @Alonski I’ll update to accommodate.

jrjohnson:
My preference is for `.prettierrc.js` in the same way we have `.eslintrc.js`. 

Over time I've found it much easier to work with blueprint changes to the `.js` file and it has been especially helpful for standardizing customization across projects and importing them. 

While I was originally on board with the idea that this config wouldn't change much, from the discussion here it sounds like there are several immediate possible changes or project specific customizations looming (`yaml`, `hbs`) and in a project as long lived as ember and the apps it supports I'd assume there will eventually be more.

NullVoxPopuli:
Because prettier config is so light, I'm totally fine with it in the package.json.

I usually tweak eslint settings as new rules come out, but I never touch prettier, so I'm happy with one less file in the root of the directory

pichfl:
@rwjblue 

I'm still not convinced not to have a separate config file. It shows straight away that the project uses prettier, without having to check an unrelated file and allows for commentary. While the latter isn't necessary, it can be helpful and could provide pointers on how prettier is applied in Ember projects. Especially for newcomers, it's helpful if a file tells you what it is for instead of you having to go on a wild chase to find that out.

> Generally speaking, I flag any commented out code while reviewing PRs 🙈.

`ember-cli-build.js` and `config/environment.js` both have examples of commented code, even tempting ones for Canary features. 😬

Re ignore file: If a prettier configuration is present in the project, people might see that as an option to use Prettier plugins in der editor. Which in turn will do unintended changes. While I know that is not our job to ensure people configure their editor correctly, Ember is still the framework with the best out-of-the-box experience. I think having the ignore file will ensure that.

btecu:
If there are two or more projects in a repository using Prettier, would it be possible to have `.prettierrc.js` in the parent directory and be used automatically, similarly how it works with `.eslintrc.js`?

This would not be possible if the configuration is stuffed inside `package.json`.

rwjblue:
@btecu - Hmm, I would think it would work the same actually.

rwjblue:
@pichfl - Ya, definitely good points. Thank you!

I’ll work on an update to move from package.json to stand alone files (likely will be a day or so to get that update pushed up, doing a bit of construction over the long weekend).

kellyselden:
I find it interesting there was no reference of the previous discussion around adding and reverting this https://github.com/ember-cli/ember-cli/pull/7458, unless I'm missing something. Back then my opposition was thoeretical based on looking at use in ember-cli. Now we've been using at my workplace, and the dislike is still there. The amount of times it destroys git blame history by collapsing or expanding lines is frustrating. I end up having to think about formatting and how prettier is going to uglify the code while I'm writing to try and get it to format it in a certain way, which breaks my concentration and flow. I think my team members (and possibly other ember community members) don't use git blame as much, which makes them like prettier a lot more than me. It seems like I'm very much in the minority and this is all but decided, but I still want to warn everyone.

rwjblue:
Thanks for chiming in here @kellyselden! 

I don’t find that Prettier makes using `git blame` much harder _other than_ the initial migration to Prettier. I do a ton of git archeology both in editor and in GitHub (knowing the “why” for a given change / logical condition / etc is super important before you change it). In both locations, having to traverse to history _prior_ to the most recent change is extremely common (unrelated to Prettier) and IMHO very often fundamental to understanding the code properly before modifying it. Both GitHub’s UI and in editor tools (Magit for Emacs, fugitive for vim, GitLens for VS Code are ones I’ve used extensively) make it pretty easy to navigate the git history. In the end, it is very important for this kind of deep dive / investigation to **not** be surface deep (e.g. only looking at the most recent commit/blame entry). 

That initial migration absolutely takes over the `git blame` output changing many lines of code that may not have previously been touched, but part of the goal of this RFC is to ensure that initial migration never has to happen (because we add it for you to begin with). After that initial migration is competed Prettier rarely modifies lines of code on its own that were not otherwise being modified by a human (though some early Prettier versions flip flopped a bit on syntax, so upgrades had a bit more noise while they figured out their process / policies).

rwjblue:
@kellyselden - Finally back to my computer, and wanted to reply to a few more points you made:

> I find it interesting there was no reference of the previous discussion around adding and reverting this ember-cli/ember-cli#7458, unless I'm missing something.

FWIW, initial drafts had a bit of a "history lesson" vibe, but it just wasn't very interesting or useful to the overall proposal. Your point in that PR was 100% correct: a change of this magnitude and impact to the community deserves an RFC. That is what this RFC is. 😸 

There was absolutely no malicious intent or burying of facts in leaving it out. 

> I end up having to think about formatting and how prettier is going to uglify the code while I'm writing to try and get it to format it in a certain way, which breaks my concentration and flow. 

I think this is likely the main point of disagreement here. The **entire point** of this proposal (and AFAICT Prettier itself) is _exactly_ the opposite: folks should **_not_** think about formatting at all when authoring or reviewing code. Your brain power is much better used on solving _actual_ problems in your domain, the computer can do the rote formatting for you.

In my experience, having bespoke formatting works when only one person is responsible to maintain it, but once you have a team of folks all with their own "style" you have:

* A large hodgepodge of styles in the same repo, making it quite difficult to know for some _new_ addition what its style should be
* Pull requests / merge requests get stuck in a quagmire of argument about formatting, when they _should_ be focusing on the code / feature / etc.
* Lots of wasted cycles discussing, changing, removing, custom bespoke styles

tldr; I believe that in the end, even if you disagree with Prettier's formatting of specific constructs it still lets you build higher, faster, with less team confusion and conflict. **THAT** is why it is especially important for the Ember ecosystem to adopt.

rwjblue:
> I’ll work on an update to move from package.json to stand alone files (likely will be a day or so to get that update pushed up, doing a bit of construction over the long weekend).

FYI - I've done that in 1276199.

kellyselden:
> Prettier rarely modifies lines of code on its own that were not otherwise being modified by a human

That's exactly the problem I'm refering to. Prettier _often_, in my experience, takes a simple change and makes it hard to trace in history. Given the example:

```
const aListOfStuff = [
  'item1',
  'item2',
  'item3',
  'item4',
  'item5',
  'item6',
];
```

Now removing item6 gives you:

```
const aListOfStuff = ['item1', 'item2', 'item3', 'item4', 'item5'];
```

Now trying to figure out when item2, for example, change takes two blame lookups. This is a simple example, you can imagine it gets pretty hair in real-world. This is what I meant when I said prettier makes me think of formatting while I'm coding, and the historical context change if I make certain changes to code.

All your responses to me seem like you think I'm arguing against the value of linting and autofixing. That's not true. I'm informing people that if you adopt prettier, you're probably going to lose reliable git blames, which was the case for me. (You also will get some pretty awkward one-liners, think array lists, destructuring, etc)

NullVoxPopuli:
The array example, and maybe small objects, are the only ones I know of where I personally prefer the non-prettier version.

Though, I still prefer prettier overall.

If prettier could instead have some way of determining when multi-line arrays can be formatted to look nicer... Then I'd have 0 problems with prettier.

But for my case, I had an array with a grid of 5x5 entries in the array, and prettier made it 25 lines.

:Shrug:

Idk. There seems to be no way to programmatically figure out how to format an array. So many options.

rwjblue:
@kellyselden 

> you're probably going to lose reliable git blames

I just can't agree here. _Perhaps_ you will loose reliable **first level** `git blame`, but as I mentioned that really isn't enough _anyways_.

rwjblue:
@kellyselden 

> All your responses to me seem like you think I'm arguing against the value of linting and autofixing.

Nah, I know we agree much more than disagree here.🥇 

Most likely I just value the overall reduction in both mental overhead when authoring code (no need to get format "just right") and bikeshedding that you get on teams when using :wave: some other tool :wave: to do stylistic formatting and it is therefore out of your control a little bit more.

kellyselden:
> first level git blame

What do you mean by first level? If an array/destructure has lots of activity, prettier will continually collapse and expand the line (seen in real life), adding a new git blame indirection each time.

rwjblue:
@kellyselden 

> What do you mean by first level?

I mean doing literally a single `git blame` without attempting to look at history prior to the most recent change on the line (or without looking at what the change to the line actually _was_).

> If an array/destructure has lots of activity, prettier will continually collapse and expand the line (seen in real life), adding a new git blame indirection each time.

Yep, definitely happens (I've seen it too!), though this seems like an exception and not the normal experience. It will only happen when a given statement is right on the cusp of being too long for the one liner.

Ultimately, even if you conclude that this line wrapping (or unwrapping) is not ideal (which I'm not really sure I agree with TBH) the benefits to the overall codebase (and team productivity) **_far_** outweigh that cost.

abhilashlr:
🎉 Nice to have this adopted as part of the ember's ecosystem and while creating new ember apps, this becomes the perfect way to start. However, I had a question on the default blueprint files that are generated using the `ember generate` command:

The default ember-source blueprints that generate test cases would need to be altered. For example, the default generated unit test case follows:
```
module('<name>, function(hooks) {
```
Now, Prettier would add an error on `function(hooks)` as it expects `function (hooks)`. I tried to search for a config that can alter this but couldn't find one or maybe my search was bad 😄 . 
As an example, we introduced this RFC approach in `upgrade-guide` ember app and I had to update the test cases with `function (hooks)` instead of `function(hooks)`. 

I believe the codemod you are suggesting (ember-cli-prettier-codemod) should handle this as well as the initial migration?

Gaurav0:
@abhilashlr There is no configuration option for this. For why, read https://prettier.io/docs/en/option-philosophy.html

abhilashlr:
> @abhilashlr There is no configuration option for this. For why, read https://prettier.io/docs/en/option-philosophy.html

Yep, I ended up with the same understanding, as the one you mentioned. However, this might tend to be huge changes considering large applications that have numerous files.
 
> Yes, I think it will be necessary to audit our blueprints for prettier errors.

Yep, this could also cascade across addons that generate tests, I believe? Although I don't have a concrete number of addons that currently generate test cases, if they do, they also need to update their blueprints. 

Few quick addons that I came across: 
1. https://github.com/typed-ember/ember-cli-typescript-blueprints/blob/master/blueprints/component-test/qunit-rfc-232-files/__root__/__testType__/__path__/__test__.ts
2. https://github.com/minutebase/ember-can/blob/master/blueprints/ability-test/index.js

ro0gr:
> I think it will be necessary to audit our blueprints for prettier errors.

I'm not sure if we should do it, until the code generated by blueprints is syntactically valid. 

Technically, apps may have some extra code style rules defined in their own eslint config or something. In this case blueprint output would also conflict with the certain project's codestyle. So in my opinion this seems to be a pre-existing issue.

I think this kind of issue may/should be solved via running(preferably automatically) a `lint --fix` after the blueprint output.

abhilashlr:
> > I think it will be necessary to audit our blueprints for prettier errors.
> 
> I'm not sure if we should do it, while the code generated by blueprints is syntactically valid.
> 
> Technically, apps may have some extra code style rules defined in their own eslint config or something. In this case blueprint output would also conflict with the certain project's codestyle. So in my opinion, this seems to be a pre-existing issue.
> 
> I think this kind of issue may/should be solved via running(preferably automatically) a `lint --fix` after the blueprint output.

@ro0gr Let's say, we create a new ember app using `ember new <sample-app>`, and then use `ember g component <new-component>` commands. That would be showing errors without even a single line of ESLint config in place, isn't it?

ro0gr:
> That would be showing errors

@abhilashlr  if ember-cli runs `lint --fix` afterwards, then no, it should not. But I suspect, this functionality might need to be considered separately, though it still feels better than changing the blueprints manually.

jgwhite:
We’ve been using Prettier for JS-files only on a green field codebase at work for about a year now and our experience has been overwhelmingly positive. I’m really excited about this RFC and support it whole-heartedly.

Regarding Prettier and HBS templates — this is definitely still ongoing but great progress has been made. When we are collectively happy with Prettier’s handling of HBS templates then we can push for the Prettier team to mark HBS templates as officially supported.

rwjblue:
@abhilashlr 

> However, I had a question on the default blueprint files that are generated using the ember generate command. The default ember-source blueprints that generate test cases would need to be altered.

Ya, definitely a good call out here. The prose in the RFC under the header "Blueprint Changes" is intending to make it clear that all addon provided blueprints (including ember-source, ember-data, ember-cli owned ones) should be updated to satisfy the (new) default linting requirements. 

Here is a copy of that prose (for easier reference):

> 
> ### Blueprint Changes
> 
> In general, it is recommended that all blueprints provided by addons should
> satisfy the default linting configuration of a new Ember application. As such
> the blueprints provided by `ember-cli`, `ember-source`, and `ember-data` will
> be updated to ensure that they satisfy these new linting rules requirements.
>

rwjblue:
@ro0gr 

> > I think it will be necessary to audit our blueprints for prettier errors.
> 
> I'm not sure if we should do it, while the code generated by blueprints is syntactically valid.
> 
> Technically, apps may have some extra code style rules defined in their own eslint config or something. In this case blueprint output would also conflict with the certain project's codestyle. So in my opinion, this seems to be a pre-existing issue.
> 
> I think this kind of issue may/should be solved via running(preferably automatically) a lint --fix after the blueprint output.

Ya, this is a good idea, thank you!

While I still think blueprints should be updated to conform to the default app/addon blueprint linting requirements (most addons that own blueprint would consider it a bug to have their blueprints fail linting immediately upon generation), running `yarn lint:fix` (when it is defined in the `package.json`) after running a blueprint (for `ember g`, `ember init`, and `ember-cli-update`) is **still** a good idea. I've pushed that tweak in 6ceceba.

abhilashlr:
> @abhilashlr
> 
> > However, I had a question on the default blueprint files that are generated using the ember generate command. The default ember-source blueprints that generate test cases would need to be altered.
> 
> Ya, definitely a good call out here. The prose in the RFC under the header "Blueprint Changes" is intending to make it clear that all addon provided blueprints (including ember-source, ember-data, ember-cli owned ones) should be updated to satisfy the (new) default linting requirements.

Apologies, I seemed to have missed it.

patocallaghan:
I'm a big fan of this change. I'll generally add Prettier to a new app or addon anyway :smile: I mightn't agree with all of Prettier's formatting decisions but the fact that I don't have to think about it is a big win for me.

We introduced Prettier on a large repo about 2 years ago and while the initial rollout was a bit bumpy as not everyone had their editors set up correctly, the developer feedback was overwhelmingly positive. At time we found that there was a ~60% reduction in lint errors on CI (~2K errors less to the previous month) as all stylistic lints were now being handled by Prettier.

luketheobscure:
Big fan of this change. I've seen firsthand how much PR chatter and code churn is eliminated when everyone is on prettier. In addition, getting eslint + prettier working together can be a bit confusing at first, so getting it all set up as the default is helpful.

I also agree with the sentiment here that prettier for hbs isn't quite ready to be the default. It's darn close, but there's at least two issues I know of that would be blockers for some people.

text/0628-prettier.md
```diff
@@ -0,0 +1,202 @@
...
+/.node_modules.ember-try/
+/bower.json.ember-try
+/package.json.ember-try
+```
+
+#### Git
+
+The `.gitignore` will be updated to add `.eslintcache` so that when using
+`eslint --cache` the cache file itself won't be added to the repositories
+history.
```
mehulkar, line 101: Why wouldn't you want to share the cache with your teammates?

jrjohnson: The thought of checking in a cache file gives me the heebie jeebies. I'm fine with just adding it to `.gitignore` myself though if I'm alone in that.

rwjblue: @mehulkar - This file would necessarily be changed by **every** commit that touches a `.js` file. That would make it **very** prone to merge conflicts, and AFAICT (did a quick local test) `eslint` doesn't automatically resolve these merge conflicts for you. Checking the file in would result in large amounts of pain, whereas not checking it in "just works" (and after that first linting run locally you get a nice speed boost).

wycats:
We discussed this RFC at today's framework meeting and believe it's ready to merge. Thanks to everyone for the great conversation. I'm personally really looking forward to this change!

PR #630, opened 2020-05-22
New browser support policy

Gaurav0:
In response to comments on #627 

[Rendered](https://github.com/emberjs/rfcs/blob/5c0eb34b26c5384cc259c797c9b51b5f13a69837/text/0630-new-browser-support-policy.md)

### DIFF
--- a/text/0000-new-browser-support-policy.md
+++ b/text/0000-new-browser-support-policy.md
@@ -8,7 +8,7 @@
 ## Summary
 
 > In dropping support for IE11 at some point in the future, we will need a new
-browswer support policy. This is the proposed policy.
+browser support policy. This is the proposed policy.
 
 ## Motivation
 
--- a/text/0000-new-browser-support-policy.md
+++ b/text/0000-new-browser-support-policy.md
@@ -1,0 +1,67 @@
+- Start Date: 2020-05-21
+- Relevant Team(s): All
+- RFC PR: (after opening the RFC PR, update this with a link to it and update the file name)
+- Tracking: (leave this empty)
+
+# New Browser Support Policy
+
+## Summary
+
+> In dropping support for IE11 at some point in the future, we will need a new
+browswer support policy. This is the proposed policy.
+
+## Motivation
+
+> It should be clear which browsers are supported by Ember.js
+
+## Detailed policy
+
+- Desktop
+
+    1. Google Chrome
+
+        - The current and previous stable releases are supported
+
+    2. Mozilla Firefox
+
+        - The current and previous rapid releases and the latest [ESR](https://support.mozilla.org/en-US/kb/choosing-firefox-update-channel) are supported
+
+    3. MS Edge
+
+        - The current and previous stable releases are supported.
+
+    4. Mac Safari
+
+        - The versions coming with the current and previous releases of MacOS are supported, as well as versions of Mac Safari available with updates to the current release.
+
+- Mobile
+
+    1. Google Chrome
+
+        - The versions coming with all versions of Android released in the past 18 months are supported, as well as the current and most stable releases available via update.
+
+    2. iOS Safari
+
+        - The versions coming with all versions of iOS released in the past 36 months are supported.
+
+    3. Android
+
+        - The versions released with all versions of Android released in the past 18 months are supported.
+
+All browsers not listed here may work but are not explicitly supported.
+
+## How we teach this
+
+> A blog post explaining the policy with a link to this RFC will be posted.
+
+## Drawbacks
+
+- See discussions on this RFC
+
+## Alternatives
+
+> See discussions on this RFC
+
+## Unresolved questions
+
+> Should a version of non Chromium MS Edge be supported, and if so for how long?
--- a/text/0630-new-browser-support-policy.md
+++ b/text/0630-new-browser-support-policy.md
@@ -8,13 +8,42 @@
 ## Summary
 
 > In dropping support for IE11 at some point in the future, we will need a new
-browser support policy. This is the proposed policy.
+browser support policy. This is the proposed policy for that point in the future.
+By properly documenting our browser support policy,
+we increase our flexibility and provide explicit clarification of intended support.
 
 ## Motivation
 
-> It should be clear which browsers are supported by Ember.js
+> It should be clear which browsers are supported by Ember.js,
+to eliminate any confusion and decrease the number of decisions
+we're required to make about any feature (existing or new).
 
-## Detailed policy
+## Existing Policy
+
+The existing policy is listed here for documentation purposes.
+
+- Desktop
+
+    1. Google Chrome, Mozilla Firefox, MS Edge, Mac Safari
+
+        - The current and previous stable releases are supported.
+
+    2. Internet Explorer
+
+        - Only Internet Explorer 11 is supported.
+
+- Mobile
+
+    1. All
+
+        - All mobile browsers may work but are not explicitly supported.
+
+Any browsers not listed here may work but are not explicitly supported.
+
+## Proposed policy
+
+This policy drops support for a major browser, and therefore can only be
+implemented with a major version bump.
 
 - Desktop
 
@@ -52,15 +81,26 @@ All browsers not listed here may work but are not explicitly supported.
 
 ## How we teach this
 
-> A blog post explaining the policy with a link to this RFC will be posted.
+- A blog post explaining the policy with a link to this RFC will be posted.
+- Browser support policy will be linked in repo
+- Browser support policy will be added to the website
+- Browser support policy will have an announcement in the Ember Times
 
 ## Drawbacks
 
-- See discussions on this RFC
+- The proposed policy adds significant testing work to make up for
+the dropping of IE11 support. This is more than made up for by clarity.
+In practice our users need to support many of these browsers
+anyway, and without Internet Explorer 11 compatibility requirement
+it is likely that older mobile browsers will become the new limits
+that prevent us from taking advantage of new features.
 
 ## Alternatives
 
-> See discussions on this RFC
+- Keep things as they are.
+- Document the existing policy but do not change it.
+- Support fewer browsers, for a shorter amount of time.
+- Support more browsers, for a longer amount of time.
 
 ## Unresolved questions
 
--- a/text/0630-new-browser-support-policy.md
+++ b/text/0630-new-browser-support-policy.md
@@ -104,4 +104,5 @@ that prevent us from taking advantage of new features.
 
 ## Unresolved questions
 
-> Should a version of non Chromium MS Edge be supported, and if so for how long?
+- Should a version of non Chromium MS Edge be supported, and if so for how long?
+- How often should this policy be reviewed?
--- a/text/0630-new-browser-support-policy.md
+++ b/text/0630-new-browser-support-policy.md
@@ -1,6 +1,6 @@
 - Start Date: 2020-05-21
 - Relevant Team(s): All
-- RFC PR: (after opening the RFC PR, update this with a link to it and update the file name)
+- RFC PR: https://github.com/emberjs/rfcs/pull/630
 - Tracking: (leave this empty)
 
 # New Browser Support Policy
@@ -24,7 +24,7 @@ browser support policy. This is the proposed policy.
 
     2. Mozilla Firefox
 
-        - The current and previous rapid releases and the latest [ESR](https://support.mozilla.org/en-US/kb/choosing-firefox-update-channel) are supported
+        - The current and previous rapid releases and any [ESR](https://support.mozilla.org/en-US/kb/choosing-firefox-update-channel) versions currently supported by Mozilla are supported
 
     3. MS Edge
 
@@ -38,7 +38,7 @@ browser support policy. This is the proposed policy.
 
     1. Google Chrome
 
-        - The versions coming with all versions of Android released in the past 18 months are supported, as well as the current and most stable releases available via update.
+        - The versions coming with all versions of Android released in the past 36 months are supported, as well as the current and most stable releases available via update.
 
     2. iOS Safari
 
@@ -46,7 +46,7 @@ browser support policy. This is the proposed policy.
 
     3. Android
 
-        - The versions released with all versions of Android released in the past 18 months are supported.
+        - The versions released with all versions of Android released in the past 36 months are supported.
 
 All browsers not listed here may work but are not explicitly supported.
 


Gaurav0:
@wycats @pzuraq @NullVoxPopuli @locks @rwjblue Comments appreciated, I have not received any input from core team in past week.

text/0000-new-browser-support-policy.md
```diff
@@ -0,0 +1,67 @@
...
+
+> It should be clear which browsers are supported by Ember.js
+
+## Detailed policy
+
+- Desktop
+
+    1. Google Chrome
+
+        - The current and previous stable releases are supported
```
NullVoxPopuli, line 23: Idk if this is enough. I know of companies that have automatic updates turned off and are on chrome 65 still

text/0000-new-browser-support-policy.md
```diff
@@ -0,0 +1,67 @@
...
+
+        - The current and previous rapid releases and the latest [ESR](https://support.mozilla.org/en-US/kb/choosing-firefox-update-channel) are supported
+
+    3. MS Edge
+
+        - The current and previous stable releases are supported.
+
+    4. Mac Safari
+
+        - The versions coming with the current and previous releases of MacOS are supported, as well as versions of Mac Safari available with updates to the current release.
```
NullVoxPopuli, line 35: Because safari moves muuuuuuch slower than chrome and Firefox. It,s possible that having this in the browserlist will accommodate concerns over chrome/Firefox support being too evergreen

text/0000-new-browser-support-policy.md
```diff
@@ -0,0 +1,67 @@
...
+
+- Mobile
+
+    1. Google Chrome
+
+        - The versions coming with all versions of Android released in the past 18 months are supported, as well as the current and most stable releases available via update.
+
+    2. iOS Safari
+
+        - The versions coming with all versions of iOS released in the past 36 months are supported.
```
NullVoxPopuli, line 45: Where are these months coming from? (For all browsers)
What methodology was used to determine that this is sufficient?

text/0000-new-browser-support-policy.md
```diff
@@ -0,0 +1,67 @@
...
+
+    3. Android
+
+        - The versions released with all versions of Android released in the past 18 months are supported.
+
+All browsers not listed here may work but are not explicitly supported.
```
NullVoxPopuli, line 51: Should we talk about browserlist? And non major browsers?
For example: 
https://browserl.ist/?q=%3E1%25%2C+not+dead

text/0000-new-browser-support-policy.md
Gaurav0: Agreed. I wish Chrome had something similar to Firefox's ESR. What do you suggest?

text/0000-new-browser-support-policy.md
Gaurav0: The months are fairly arbitrary, I'm trying not to overburden us. In fact, even 36 months is not "sufficient". My research suggests iPhone 7, for example, is still in wide use.

text/0000-new-browser-support-policy.md
Gaurav0: Perhaps. I'm concerned about the burden on testing from adding too many browsers.

text/0000-new-browser-support-policy.md
```diff
@@ -0,0 +1,67 @@
...
+## Summary
+
+> In dropping support for IE11 at some point in the future, we will need a new
+browser support policy. This is the proposed policy.
+
+## Motivation
+
+> It should be clear which browsers are supported by Ember.js
+
+## Detailed policy
```
jrjohnson, line 17: I think it needs to be clear *when* the support matrix is recalculated. I'd like to see this happen only after an LTS release, but I'm not sure quite how to phrase that. 
Maybe:
> Any specific browser version which falls out of support between LTS versions will only be dropped after an LTS supporting it has been released
?

This ensures it's not a constantly recalculated moving target that might shift an app into a non-working state during a minor update cycle.

text/0000-new-browser-support-policy.md
```diff
@@ -0,0 +1,67 @@
...
+
+- Desktop
+
+    1. Google Chrome
+
+        - The current and previous stable releases are supported
+
+    2. Mozilla Firefox
+
+        - The current and previous rapid releases and the latest [ESR](https://support.mozilla.org/en-US/kb/choosing-firefox-update-channel) are supported
```
jrjohnson, line 27: ESR has a bit of [overlap built in](https://support.mozilla.org/en-US/kb/firefox-esr-release-cycle)

>  For example, users of version 52 will be updated only when version 60.2.0 has been published

So this should read currently supported ESR versions.

text/0000-new-browser-support-policy.md
```diff
@@ -0,0 +1,67 @@
...
+
+    4. Mac Safari
+
+        - The versions coming with the current and previous releases of MacOS are supported, as well as versions of Mac Safari available with updates to the current release.
+
+- Mobile
+
+    1. Google Chrome
+
+        - The versions coming with all versions of Android released in the past 18 months are supported, as well as the current and most stable releases available via update.
```
jrjohnson, line 41: iOS and Android should have the same release length. I'm not sure why it would be different. I do think it is OK for us to rely on Android users to keep their devices updated so possibly:

> The current and previous version available on any Android device released in the past 36 months.

text/0000-new-browser-support-policy.md
jrjohnson: I originally didn't think this was long enough and that it would force application owners to consider dropping support for users who's phones are only a few years old. However given that you can update the iOS version (and therefor the browser) for several years I think 36 months is OK.

text/0000-new-browser-support-policy.md
Gaurav0: "any Android device" is likely too broad.

text/0630-new-browser-support-policy.md
```diff
@@ -0,0 +1,67 @@
...
+
+    3. Android
+
+        - The versions released with all versions of Android released in the past 36 months are supported.
+
+All browsers not listed here may work but are not explicitly supported.
+
+## How we teach this
+
+> A blog post explaining the policy with a link to this RFC will be posted.
```
mehulkar, line 55: Should also explain in the Guides what "browser support" means. There are two things I can think of:

- ember-source's test suite runs against these browsers
- default `config/targets.js` is configured to this list

Both of these should have explanations for new developers who may not be familiar with the challenges of browser support.

text/0630-new-browser-support-policy.md
```diff
@@ -0,0 +1,67 @@
+- Start Date: 2020-05-21
+- Relevant Team(s): All
+- RFC PR: https://github.com/emberjs/rfcs/pull/630
+- Tracking: (leave this empty)
+
+# New Browser Support Policy
+
+## Summary
+
+> In dropping support for IE11 at some point in the future, we will need a new
```
mansona, line 10: so you're saying that we are looking to drop IE11 `at some point in the future`, and this is the "right now" browser support policy. So doesn't that mean that this proposed policy should **include** IE11 and when we drop support for it we remove IE11 from the "official policy"? 🤔

Gaurav0: I said "This is the proposed policy". How can I make this more clear?

text/0630-new-browser-support-policy.md
```diff
@@ -0,0 +1,67 @@
+- Start Date: 2020-05-21
+- Relevant Team(s): All
+- RFC PR: https://github.com/emberjs/rfcs/pull/630
+- Tracking: (leave this empty)
+
+# New Browser Support Policy
```
MelSumner, line 6:
```suggestion
# Document Browser Support Policy
```
Recommend changing this to document the reality of the current browser support (so it can be policy that can be reviewed/changed at an official/project level). This sets the project up for more flexibility and future growth.

text/0630-new-browser-support-policy.md
MelSumner, line 10:
```suggestion
By properly documenting our browser support policy, we increase our flexibility and provide explicit clarification of intended support. 
```

text/0630-new-browser-support-policy.md
```diff
@@ -0,0 +1,67 @@
...
+
+## Summary
+
+> In dropping support for IE11 at some point in the future, we will need a new
+browser support policy. This is the proposed policy.
+
+## Motivation
+
+> It should be clear which browsers are supported by Ember.js
```
MelSumner, line 15:
```suggestion
It should be clear which browsers are supported by Ember.js, to eliminate any confusion and decrease the number of decisions we're required to make about any feature (existing or new).
```

text/0630-new-browser-support-policy.md
```diff
@@ -0,0 +1,67 @@
...
+
+## Detailed policy
```
MelSumner, line 17: For this entire section, I would simply say "Browser support" and then list all of the browsers. If you want to further indicate mobile devices, put that in parenthesis after the list item.

text/0630-new-browser-support-policy.md
MelSumner, line 55: This is not sufficient. Please add: 

- browser support policy will be linked in repo
- browser support policy will be added to the website
- browser support policy will have an announcement in the Ember Times
- consider: adding browser support to CLI output

text/0630-new-browser-support-policy.md
```diff
@@ -0,0 +1,67 @@
...
+
+All browsers not listed here may work but are not explicitly supported.
+
+## How we teach this
+
+> A blog post explaining the policy with a link to this RFC will be posted.
+
+## Drawbacks
+
+- See discussions on this RFC
```
MelSumner, line 59:
```suggestion
An explicit browser support policy will require regular review. Cadence TBD. 

The discussion/comments may identify additional drawbacks that have not been identified in this RFC yet, and will be added to this RFC if that is the case.
```

text/0630-new-browser-support-policy.md
```diff
@@ -0,0 +1,67 @@
...
+
+## Alternatives
+
+> See discussions on this RFC
```
MelSumner, line 63:
```suggestion
- Things stay as they are now
- We include assistive technology (largely unchanged in the past 25+ years) in our support policy

The discussion/comments may identify additional alternatives that have not been identified yet; they will be added to this RFC if that is the case. 
```

text/0630-new-browser-support-policy.md
```diff
@@ -0,0 +1,67 @@
...
+
+## Unresolved questions
+
+> Should a version of non Chromium MS Edge be supported, and if so for how long?
```
MelSumner, line 67:
```suggestion
- Should a version of non-Chromium MS Edge be supported, and if so for how long?
- What are the support schedules for the respective browser vendors? 
```

text/0630-new-browser-support-policy.md
Gaurav0: The current browser support policy is poorly documented, I agree. However, what I am proposing is a stark departure from current policy. I will document current policy in this RFC for comparison.

text/0630-new-browser-support-policy.md
Gaurav0: Given that mobile browsers are being given significantly different treatment from desktop browsers, this is impractical.

text/0630-new-browser-support-policy.md
Gaurav0: The support schedules are known and documented, however practically we must support more browsers for a longer period of time than browsers are supported by vendors.

text/0630-new-browser-support-policy.md
rwjblue: Sure, but regardless this RFC is attempting to explicitly document the _new_ policy (while deciding what it is).

text/0630-new-browser-support-policy.md
rwjblue: > however practically we must support more browsers for a longer period of time than browsers are supported by vendors

I'm not quite sure this is true. I don't think we should support browsers that themselves are not supported. 🤔

Gaurav0: For example, IE11 has not been officially supported by Microsoft since 2015 except for security updates.

rwjblue: > officially supported by Microsoft since 2015 except for security updates

These two parts are contradictory 😸

Security updates === supported :wink:

Gaurav0: In that case we may end up supporting IE11 indefinitely, as long as Windows exists.

text/0630-new-browser-support-policy.md
Gaurav0: I think assistive technology support should be covered in a separate RFC.

text/0630-new-browser-support-policy.md
Gaurav0: Not recommended for use === supported?

Also, as of January 2020, corporations and governments using IE11 are supposed to pay for those security updates.

text/0630-new-browser-support-policy.md
Gaurav0: @MelSumner @rwjblue Documented current policy. Please rereview.

text/0630-new-browser-support-policy.md
MelSumner: Can you explain more? I disagree that it's impractical, but maybe you have more information I'm missing.

text/0630-new-browser-support-policy.md
MelSumner: I disagree, I think saying "accessibility is different" is the root cause of accessibility issues on the web for the past 30+ years the web has existed. We can do better.

Gaurav0: I agree that "accessibility is not different". I disagree that I should be the one to propose the standards, due to my ignorance. :) If you'd like to write something up, I will try to add it, but I recommend that you write the RFC for an accessibility policy yourself, and I can link to it.

Gaurav0: I do not think adding "we include assistive technology" should be an *alternative*. It should be first class.

luxferresum:
IMHO only supporting the current and the previous release of chrome is *not* a good idea. Thats basically just 12 weeks.

In enterprise environments its not uncommon to run behind quite a bit. This also totally leaves out offline or/and embedded devices. We should at least support 12 months back IMHO while still allowing to jump from LTS to LTS.

Gaurav0:
@luxferresum See https://github.com/emberjs/rfcs/pull/630#discussion_r429170193

text/0630-new-browser-support-policy.md
Gaurav0: Browser support for Firefox on Desktop but not for mobile. Browser support for Android browser is mobile only, no desktop support.

For all other browsers, browser support is also much longer for mobile than desktop because browsers on mobile are typically only updated as long as phone vendors choose to support the phones which means that we have to support Chrome, Safari, and Android for a full 3 years.

If we list browsers first and then make exceptions for mobile, we make exceptions for every single browser.

Gaurav0:
@wycats @pzuraq @NullVoxPopuli @locks @rwjblue @MelSumner Still looking for comments from core team members.

text/0630-new-browser-support-policy.md
Gaurav0: > > however practically we must support more browsers for a longer period of time than browsers are supported by vendors

> I'm not quite sure this is true. I don't think we should support browsers that themselves are not supported. 🤔

I'll give another example. For Chrome, only the stable and previous release are supported by Google. However, many enterprise desktops and mobile users are running much older versions of Chrome. Enterprise users are kept on older browsers to avoid frequent testing of internal apps. Mobile users often stop upgrading their browsers once phone vendors stop shipping updates, leaving a large segment with old versions.

text/0000-new-browser-support-policy.md
jrjohnson: How about the same as iOS. All versions of chrome released in the past 36 months? (could be the same for FF, Safari too). So no longer tied to version but to release date.

text/0630-new-browser-support-policy.md
```diff
@@ -0,0 +1,108 @@
...
+
+        - The versions coming with all versions of Android released in the past 36 months are supported, as well as the current and most stable releases available via update.
+
+    2. iOS Safari
+
+        - The versions coming with all versions of iOS released in the past 36 months are supported.
+
+    3. Android
+
+        - The versions released with all versions of Android released in the past 36 months are supported.
```
sandstrom, line 78: What is "Android"? Any browser released for Android?

text/0630-new-browser-support-policy.md
```diff
@@ -0,0 +1,108 @@
...
+
+- Mobile
+
+    1. Google Chrome
+
+        - The versions coming with all versions of Android released in the past 36 months are supported, as well as the current and most stable releases available via update.
+
+    2. iOS Safari
+
+        - The versions coming with all versions of iOS released in the past 36 months are supported.
```
sandstrom, line 74: 36 months is a *very* long time. Do we really need to support browsers that far back?

Individual developers can still delay upgrading Ember if they need to retain support for some old browser.

sandstrom (commented):
A few thoughts:

- I think there is a balance between browser support in the framework, and letting the developer provide additional backwards compatibility by *not* upgrading their Ember version. As a Javascript framework, it's helpful for Ember to not be bogged down by support for old browsers too much (more resources for actual development + possible to take advantage of new features). We've recently updated our app to 3.18, but we used to run on 2.14 until 2 months ago. Worked just fine, we've been building out the app, even though we've been running a quite old version of Ember for some time. That wasn't for browser support reasons, but it could have been.
    
    In short, a framework can support an old browser via an *old* version of the framework, e.g. a policy could state that "A non-expired Ember LTS should support mobile browsers going back Y months, but the most recent version of Ember will only support mobile browsers going Y/2 months back."

- 36 months for mobile browsers is a *very* long time. Do we really need to support browsers that far back? I'd vote for 24 months on iOS/Android (or possible even shorter). The iOS ecosystem will usually see ~95% of devices running the most recent version of iOS. Those needing more backwards support can always wait updating Ember for a while. Most Android browsers are "evergreen".

- What's the definition of "support"? For example, if a new ember features would introduce a performance regression in old versions of some browser (but not new ones), is that okay? (I think it totally should be!)

- What about webviews vs. browsers on iOS and Android? Generally they're the same thing, but could be worth mentioning (since apps build via e.g. Cordova/Capacitor would run in a webview) if they are in scope or not.

- Chrome on iOS is essentially Safari (WKWebView). For mobile, would be better to first split by platform, and then list browsers per platform. For clarity and to avoid confusion (though most developers should know this!).

pzuraq:
I'm going to close this for now as we're working on a followup RFC. Will comment here as soon as its up!

sandstrom:
For anyone who's searching for it, the followup RFC is here: https://github.com/emberjs/rfcs/pull/685

Issue #629, opened 2020-05-20
Deprecate string based action handling


mehulkar:
We should deprecate all forms of string based actions. This includes:

- `{{action "foo"}}`
- `(action 'foo')`
- `this.send('foo')` (as discussed in #609)

rwjblue:
Thanks @mehulkar! I added a few labels to clarify where this is at. I personally think it should be possible to start charting out what that RFC would look like...

chriskrycho:
I was taking with team members around this same topic today, especially `this.send`. The only significant blocker for its removal I could think of is that it’s the normal and idiomatic way to trigger an action on a route from its controller, which is currently the only way to fully reload a route’s model (using [`Route.refresh`](https://api.emberjs.com/ember/release/classes/Route)). The only alternative I’m aware of is to look up the corresponding route using the container, e.g. `getOwner(this).lookup('route:same-as-controller').refresh()`.

An obvious intermediate step there might be to support refreshing a route on the router service. That would need to be designed independently of this, but it’s the only thing I know of which actually *requires* `this.send` in idiomatic Ember today. That would also be a nice step along the path of removing places we still *need* controllers today, and in a direction well-aligned with current directions in Ember API design. (I think there’s actually an open issue for this somewhere but I can’t find it and am typing on mobile at the moment.)

cc. @gabrielcsapo @jackson-dean

mehulkar:
FWIW, it's not required, per se. There are a few ways to do this without `.send`. None of them are good, but 🤷

- Set Route.refresh as a property on the controller in setupController
- Use lookup(`route:${RouterService.currentRouteName}`).refresh()
- Use [`ember-route-action-helper` addon](https://github.com/dockyard/ember-route-action-helper)

rwjblue:
It isn't immediately obvious that we need per-route `.refresh` functionality to me (though maybe folks could share good use cases), and if that is the case then I think a top level router service `.refresh` method might be all we need.

mehulkar:
> That would also be a nice step along the path of removing places we still need controllers today,

Also, for the record, my understanding from   https://github.com/emberjs/rfcs/pull/574#issuecomment-583636761 is that the core team was not in favor of incremental progress towards removing controllers.

chriskrycho:
Certainly every use care I’ve hit for `refresh` would be solved by a single method on the router service. I’d also be curious if others can supply scenarios where that’s insufficient!

[...]

Issue #592, opened 2020-02-08
Calling ".refresh()" from the router service


NullVoxPopuli:
Right now, refresh only exists on the route.
It'd be fantastic if we had a central place to call refresh.

### Update 1

```ts
@service router;
// ...
// refreshes current route
this.router.refresh() 
// refreshes the application and all child routes
this.router.refresh('application') 

// maybe typed this way
refresh(routeName?: string): Promise<void>;
```
-------------------
### Original Text

Maybe this is something that exists on each RouteInfo that makes of the list of route infos representing the current route?

Given:
```ts
this.route('posts', function() {
  this.route('show', { path: ':id' });
})
```
And I want to refresh just the show route, maybe I'd do:
```ts
@service router;

// ...
this.router.currentRoute.refresh();
```

And give that the route tree is represented as a linked list:
```
<RouteInfo#application>
   ⇩
<RouteInfo#posts>
   ⇩
<RouteInfo#show>  <- currentRoute
```
maybe I want to refresh just the posts route?
```ts
this.router.currentRoute.parent.refresh();
```

Thoughts?

nightire:
`this.router.refresh()` -> refresh current route.

`this.route.refresh(ROUTE_NAME)~ -> refresh a specific route (along with its children of course), and if it is not in the current route tree, throw an error in dev or do nothing (maybe?) in production.

jaydgruber:
I have questions about how this would work with queryParams. With sticky queryParams? How is this different from calling this.transitionTo(currentRouteName)?

jakebixbyavalara:
> this.router.refresh() -> refresh the current route.
>
> this.router.refresh(ROUTE_NAME) -> refresh a specific route (along with its children of course), and if it is not in the current route tree, throw an error in dev and do nothing (maybe?) in production.
>
> The reason that I'm not a fan of this.router.currentRoute.parent:
>
> Let's say I have a component that can refresh a specific route, 'posts' for example, I can render this component in any level of child route of 'posts', so the length of the parent chain can not be determined easily. So why not just use the route name as we used to in the transitionTo method?

I completely agree with this :point_up:. Because I see the RouteInfo class as more of a "snapshot", I think that treating the refresh method the same as `this.router.transitionTo` makes a bit more sense. 

If we were to put the `refresh` method on the router service, I think that the most common use case will be `this.router.refresh()` with no arguments because 90% of the time you'll want to refresh the route you're currently on. 

[...]

PR #631, opened 2020-05-23
add refresh method to router service

Gaurav0:
Part of issue #629 

[Rendered](https://github.com/emberjs/rfcs/blob/f2c75d1cd53d0a99c91a24dc22ac4ff3c3836a8e/text/0631-refresh-method-for-router-service.md)

### DIFF
--- a/text/0000-refresh-method-for-router-service.md
+++ b/text/0000-refresh-method-for-router-service.md
@@ -7,25 +7,34 @@
 
 ## Summary
 
-> Add a refresh method to the router service that calls refresh on the current route.
+> Add a refresh method to the router service that calls refresh on the route specified by `currentRouteName`.
 
 ## Motivation
 
-> We want to be able to refresh the current route without relying on the send api.
+> We want to be able to call refresh the route specified by `currentRouteName` without relying on the send api, which we would like to deprecate. This enables us to get the latest data from the model hook.
 
 ## Detailed design
 
-```
+```js
 class RouterService {
     refresh() {
-        this.owner.lookup(`route:${this.currentRouteName}`).refresh();
+        this._router._routerMicroLib.refresh();
     }
 }
 ```
 
 ## How we teach this
 
-> Documentation will be added to the method.
+> The following documentation will be added to the method:
+
+```js
+/**
+ * Refreshes the current route, updating it with new data from the model hook.
+ * 
+ * @method refresh
+ * @public
+ */
+```
 
 ## Drawbacks
 
@@ -33,4 +42,6 @@ class RouterService {
 
 ## Alternatives
 
-> We could provide a direct link to the current route via the router service.
+> We could provide a direct link to the current route via the router service. However,
+this would encourage people to use routes to store information and provide methods
+that should be idiomatically placed in a service.
--- a/text/0000-refresh-method-for-router-service.md
+++ b/text/0000-refresh-method-for-router-service.md
@@ -1,0 +1,36 @@
+- Start Date: 2020-05-23
+- Relevant Team(s): Ember.js
+- RFC PR: (after opening the RFC PR, update this with a link to it and update the file name)
+- Tracking: (leave this empty)
+
+# RouterService#refresh
+
+## Summary
+
+> Add a refresh method to the router service that calls refresh on the current route.
+
+## Motivation
+
+> We want to be able to refresh the current route without relying on the send api.
+
+## Detailed design
+
+```
+class RouterService {
+    refresh() {
+        this.owner.lookup(`route:${this.currentRouteName}`).refresh();
+    }
+}
+```
+
+## How we teach this
+
+> Documentation will be added to the method.
+
+## Drawbacks
+
+> This is a slight increase in API surface area.
+
+## Alternatives
+
+> We could provide a direct link to the current route via the router service.
--- a/text/0631-refresh-method-for-router-service.md
+++ b/text/0631-refresh-method-for-router-service.md
@@ -1,17 +1,19 @@
 - Start Date: 2020-05-23
 - Relevant Team(s): Ember.js
-- RFC PR: (after opening the RFC PR, update this with a link to it and update the file name)
+- RFC PR: https://github.com/emberjs/rfcs/pull/631
 - Tracking: (leave this empty)
 
 # RouterService#refresh
 
 ## Summary
 
-> Add a refresh method to the router service that calls refresh on the route specified by `currentRouteName`.
+> Add a refresh method to the router service that calls refresh on all currently active routes.
 
 ## Motivation
 
-> We want to be able to call refresh the route specified by `currentRouteName` without relying on the send api, which we would like to deprecate. This enables us to get the latest data from the model hook.
+> We want to be able to call refresh on all currently active routes from a centralized service
+or from a component. As a side benefit, we will be able to do this without relying on the send api,
+which we would like to deprecate. This enables us to get the latest data from the model hook.
 
 ## Detailed design
 
@@ -29,7 +31,9 @@ class RouterService {
 
 ```js
 /**
- * Refreshes the current route, updating it with new data from the model hook.
+ * Refreshes all currently active routes, doing a full transition.
+ * All resetController, beforeModel, model, afterModel, redirect, and setupController
+ * hooks will be called again. You will get new data from the model hook.
  * 
  * @method refresh
  * @public
--- a/text/0631-refresh-method-for-router-service.md
+++ b/text/0631-refresh-method-for-router-service.md
@@ -21,13 +21,13 @@ This enables us to get the latest data from the model hook.
 
 ## Detailed design
 
-The following pseudocode represents what the overall technical design of this method.
+The following pseudocode represents the overall technical design of this method.
 This code implementation is not normative.
 
 ```js
 class RouterService {
     refresh(pivotRouteName?: string): Transition {
-        let pivotRoute = pivotRouteName ?? lookupRoute(pivotRouteName);
+        let pivotRoute = pivotRouteName && lookupRoute(pivotRouteName);
         assert("If an argument provided must be the name of an active route", !pivotRouteName || isActiveRoute(pivotRoute));
         return this._router._routerMicroLib.refresh(pivotRoute);
     }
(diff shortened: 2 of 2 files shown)


text/0000-refresh-method-for-router-service.md
```diff
@@ -0,0 +1,36 @@
...
+class RouterService {
+    refresh() {
+        this.owner.lookup(`route:${this.currentRouteName}`).refresh();
+    }
+}
+```
+
+## How we teach this
+
+> Documentation will be added to the method.
```
rwjblue, line 28: The API documentation for the method should be included in this section. It should likely also include some details around reviewing the current guides to see if any can / should be updated to leverage the new feature.

text/0000-refresh-method-for-router-service.md
```diff
@@ -0,0 +1,36 @@
...
+
+## Drawbacks
+
+> This is a slight increase in API surface area.
+
+## Alternatives
+
+> We could provide a direct link to the current route via the router service.
```
rwjblue, line 36: Should include explanation of why we wouldn't want to do this.

rwjblue:
@mehulkar `Route.prototype.refresh` is a tad different than this new method being proposed. It refreshes whatever route you call it on and all routes “below” it in the route hierarchy. The method proposed here always refreshes all routes.

That being said, I think it would be good to mention the service method **from** the route documentation, to make it super easy for folks to see there is a service mechanism to do this also.

acorncom:
> It refreshes whatever route you call it on and all routes “below” it in the route hierarchy. The method proposed here always refreshes all routes.

Would we want to be opting for different behavior and refreshing all routes or is the implicit goal here to refresh only the current route (as defined from the perspective of the caller)? The “defined from the perspective of the caller” part seems quite tricky to implement well (and consistently) ...

Gaurav0:
> Should this also deprecate Route#refresh()? And/Or change its implementation to delegate to RouterService?

In my opinion, no. We should keep it around until we are sure we have filled in all the gaps. There may be a use case for refreshing an ancestor of the current route, possibly.

text/0000-refresh-method-for-router-service.md
```diff
@@ -0,0 +1,47 @@
+- Start Date: 2020-05-23
+- Relevant Team(s): Ember.js
+- RFC PR: (after opening the RFC PR, update this with a link to it and update the file name)
+- Tracking: (leave this empty)
+
+# RouterService#refresh
+
+## Summary
+
+> Add a refresh method to the router service that calls refresh on the route specified by `currentRouteName`.
```
rwjblue, line 10: These `currentRouteName` references still need to be removed.

text/0000-refresh-method-for-router-service.md
```diff
@@ -0,0 +1,47 @@
...
+
+## Motivation
+
+> We want to be able to call refresh the route specified by `currentRouteName` without relying on the send api, which we would like to deprecate. This enables us to get the latest data from the model hook.
```
rwjblue, line 14: I think there are a number of better motivations. Deprecating `.send` isn't _really_ a motivation for adding this feature IMO (it's a nice side benefit though 😄). A better motivation (that I've seen in the wild a few places) is wanting to be able to trigger a route refresh from a centralized service, or to trigger a refresh from within a component (e.g. a "refresh button").

rwjblue: Should also not mention `currentRouteName`, as the router service would be refreshing the models for **all** currently active routes not just the one referenced by `currentRouteName`.

text/0000-refresh-method-for-router-service.md
```diff
@@ -0,0 +1,47 @@
...
+}
+```
+
+## How we teach this
+
+> The following documentation will be added to the method:
+
+```js
+/**
+ * Refreshes the current route, updating it with new data from the model hook.
```
rwjblue, line 32: Needs to incorporate the details RE: the full route heirarchy. Should probably also explain what refreshing actually means and what specifically is entailed (e.g. does it re-run `beforeModel`,`model`,`afterModel`? or only `model`?).

text/0000-refresh-method-for-router-service.md
mehulkar: > all currently active routes

Just to make sure... "all active" means `currentRoute` and all its ancestors, right? Does this have a name? Maybe `currentRouteTree` or something could represent this in a data structure that makes sense?

Gaurav0: Yes, but I've decided to use "all currently active routes" because "active routes" does have a technical definition in the api.

rwjblue: > Just to make sure... "all active" means currentRoute and all its ancestors, right? 

Ya, exactly. I don’t know that it has a name other than “currently active routes” or some variant thereof.

Gaurav0:
@rwjblue revised.

text/0631-refresh-method-for-router-service.md
```diff
@@ -0,0 +1,51 @@
...
+
+## Summary
+
+> Add a refresh method to the router service that calls refresh on all currently active routes.
+
+## Motivation
+
+> We want to be able to call refresh on all currently active routes from a centralized service
+or from a component. As a side benefit, we will be able to do this without relying on the send api,
+which we would like to deprecate. This enables us to get the latest data from the model hook.
```
rwjblue, line 16: Maybe tweak this to not say "which we would like to deprecate" but say something like "which is being discussed as a possible deprecation in rfc#9999".

mike183:
@rwjblue @acorncom I think it would be useful to have a method on the router service for refreshing the current route (including child routes). 

I had a recent use case for providing a “Try Again” button when a route failed to load due to a timeout.

I did look for some method of refreshing the route but ended up resorting to just redirecting to a parent route which so happened to redirect back to the route that I wanted to refresh.

Gaurav0:
@mike183 How about adding an optional routeName argument as proposed in #592 ?

mike183:
> @mike183 How about adding an optional routeName argument as proposed in #592 ?

That seems like it would work 👍

text/0631-refresh-method-for-router-service.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Summary
+
+> Add a refresh method to the router service that calls refresh on all currently active routes.
+
+## Motivation
+
+> We want to be able to call refresh on all currently active routes from a centralized service
+or from a component. As a side benefit, we will be able to do this without relying on the send api,
+which is being discussed as a possible deprecation in RFC 632.
```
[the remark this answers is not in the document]
Gaurav0: Done.

Gaurav0:
@mike183 @rwjblue Updated. Please rereview.

text/0631-refresh-method-for-router-service.md
```diff
@@ -7,46 +7,63 @@
...
 ## Detailed design
 
+The following pseudocode represents what the overall technical design of this method.
+This code implementation is not normative.
+
 ```js
 class RouterService {
-    refresh() {
-        this._router._routerMicroLib.refresh();
+    refresh(pivotRouteName?: string): Promise<Transition> {
```
rwjblue, line 29: As it stands today we should be returning a `Transition` which _is_ a "thennable" (not a promise that resolves to a transition).

```suggestion
    refresh(pivotRouteName?: string): Transition {
```

text/0631-refresh-method-for-router-service.md
```diff
@@ -7,46 +7,63 @@
...
 /**
  * Refreshes all currently active routes, doing a full transition.
+ * If a pivotRouteName is provided and refers to a currently active route,
+ * it will refresh only that route and its descendents.
+ * Returns a promise that will be resolved once the refresh is complete.
  * All resetController, beforeModel, model, afterModel, redirect, and setupController
  * hooks will be called again. You will get new data from the model hook.
  * 
  * @method refresh
+ * @param pivotRouteName?: string
```
rwjblue, line 56:
```suggestion
 * @param {String} [pivotRouteName] the route to refresh (along with all child routes)
```

text/0631-refresh-method-for-router-service.md
```diff
@@ -7,46 +7,63 @@
...
+ * @return Promise<Transition>
```
rwjblue, line 57:
```suggestion
 * @return Transition
```

Gaurav0:
@ewjblue Thanks again for the review.

mike183:
Apologies for the delayed response.

The updates all look good to me, nice work 👍

mehulkar:
Re-reading this after a while, does my previous question/suggestion apply more, now that `pivotRouteName` is introduced? In other words, could `Route.prototype.refresh` be implemented as:

```js
{
  refresh() {
    routerService.refresh(this.fullName)
  }
}
```

rwjblue:
> Re-reading this after a while, does my previous question/suggestion apply more, now that `pivotRouteName` is introduced? In other words, could `Route.prototype.refresh` be implemented as:
> 
> 
> 
> ```js
> 
> {
> 
>   refresh() {
> 
>     routerService.refresh(this.fullName)
> 
>   }
> 
> }
> 
> ```

Ya, I think that makes sense, though I'm not sure it really matters from an RFC perspective (since it doesn't really change any existing API surface, it just makes the implementation make more sense).

rwjblue:
👋 - We discussed this at today's Ember core meeting and are moving it into final comment period.

k-dauda:
I would be nice to have an affordance like
`router.currentRoute.refresh()`

Gaurav0:
@k-dauda  This point is addressed in alternatives: https://github.com/emberjs/rfcs/blob/f2c75d1cd53d0a99c91a24dc22ac4ff3c3836a8e/text/0631-refresh-method-for-router-service.md#alternatives

rwjblue:
Sorry for the delays here, we discussed this in last weeks Ember core team meeting last week and are happy to land this!

MariannaAtPlay:
This is a great idea! Wondering what's the timeline for implementing this in Ember core?

snewcomer:
@MariannaAtPlay This was implemented in `4.1.0-beta.1.`!  

https://github.com/emberjs/ember.js/blob/master/CHANGELOG.md#v410-beta1-november-19-2021

PR #632, opened 2020-05-23
Deprecate string based actions

Gaurav0:
For issue #629 

[Rendered](https://github.com/emberjs/rfcs/blob/b50d9c60b6fb0b54f2db23ded31bf9782e1e6963/text/0632-deprecate-string-based-actions.md)

### DIFF
--- a/text/0000-deprecate-string-based-actions.md
+++ b/text/0000-deprecate-string-based-actions.md
@@ -7,14 +7,14 @@
 
 ## Motivation
 
-> In Ember 1.13 we introduced [closure actions](https://github.com/emberjs/rfcs/blob/00ac2685c86f27d41547012903f485a4ef338d27/active/0000-improved-actions.md) have been recommended over string based actions. This officially deprecates string based actions in favor of closure actions,
+> In Ember 1.13 we introduced [closure actions](https://github.com/emberjs/rfcs/blob/00ac2685c86f27d41547012903f485a4ef338d27/active/0000-improved-actions.md), which have been recommended over string based actions. This officially deprecates string based actions in favor of closure actions,
 which provide many benefits such as
 
 1. simpler passing through components
 2. less need for an actions hash
 3. ability to return a value
 4. typing in TypeScript
-5. compatible with Glimmer Components
+5. more in line with the Octane programming model
 
 ## Detailed Design
 
@@ -23,6 +23,16 @@ We will deprecate the following:
 1. The `send` api.
 2. The use of a string parameter to the `action` helper.
 3. The use of a string parameter to the `action` modifier.
+
+The `send` api may currently be used in components, controllers, and routes.
+It is used to target the `actions` hash. In calling the `actions` hash of the
+same object, `send` can be replaced by a function call once the method is
+moved outside the `actions` hash.
+
+For cases where `send` is used to go from a controller to a route or from
+a route to a parent route, no direct replacement for this "action bubbling"
+behavior will be provided.
+Instead, users are recommended to inject the router service.
 
 ## Transition Path
 
--- a/text/0000-deprecate-string-based-actions.md
+++ b/text/0000-deprecate-string-based-actions.md
@@ -1,0 +1,89 @@
+- Start Date: 2020-05-23
+- Relevant Team(s): Ember.js
+- RFC PR: (after opening the RFC PR, update this with a link to it and update the file name)
+- Tracking: (leave this empty)
+
+# Deprecate String Based Actions
+
+## Motivation
+
+> In Ember 1.13 we introduced [closure actions](https://github.com/emberjs/rfcs/blob/00ac2685c86f27d41547012903f485a4ef338d27/active/0000-improved-actions.md) have been recommended over string based actions. This officially deprecates string based actions in favor of closure actions,
+which provide many benefits such as
+
+1. simpler passing through components
+2. less need for an actions hash
+3. ability to return a value
+4. typing in TypeScript
+5. compatible with Glimmer Components
+
+## Detailed Design
+
+We will deprecate the following:
+
+1. The `send` api.
+2. The use of a string parameter to the `action` helper.
+
+## Transition Path
+
+Users will transition to closure actions.
+
+From:
+
+```js
+export default Component.extend({
+    someMethod() {
+        this.send('actionName');
+    }
+
+    actions: {
+        actionName() {
+            // do something
+        }
+    }
+});
+```
+
+To:
+
+```js
+export default Component.extend({
+    someMethod() {
+        this.actionName();
+    }
+
+    actionName() {
+        // do something
+    }
+});
+```
+
+From:
+
+```hbs
+<button {{action 'actionName'}}>Action Label</button>
+```
+
+To:
+
+```hbs
+<button {{action this.actionName}}>Action Label</button>
+```
+
+## How We Teach This
+
+> We already teach closure actions exclusively in the guides. Other than a new
+deprecation guide, no additional material should be necessary.
+
+## Drawbacks
+
+> Some older applications have thousands of string based actions throughout their codebases.
+We should consider writing a codemod to ease the transition.
+
+## Alternatives
+
+> Closure actions are now the standard in Ember, so other alternatives have not been
+seriously considered.
+
+## Unresolved questions
+
+> Can we also deprecate the use of the `actions` hash?
--- a/text/0000-deprecate-string-based-actions.md
+++ b/text/0000-deprecate-string-based-actions.md
@@ -22,6 +22,7 @@ We will deprecate the following:
 
 1. The `send` api.
 2. The use of a string parameter to the `action` helper.
+3. The use of a string parameter to the `action` modifier.
 
 ## Transition Path
 
--- a/text/0000-deprecate-string-based-actions.md
+++ b/text/0000-deprecate-string-based-actions.md
@@ -33,7 +33,7 @@ From:
 export default Component.extend({
     someMethod() {
         this.send('actionName');
-    }
+    },
 
     actions: {
         actionName() {
@@ -49,7 +49,7 @@ To:
 export default Component.extend({
     someMethod() {
         this.actionName();
-    }
+    },
 
     actionName() {
         // do something
@@ -86,4 +86,4 @@ seriously considered.
 
 ## Unresolved questions
 
-> Can we also deprecate the use of the `actions` hash?
+> Can we also deprecate the use of the `actions` hash? Can we also deprecate using `this.actions`?
--- a/text/0632-deprecate-string-based-actions.md
+++ b/text/0632-deprecate-string-based-actions.md
@@ -101,7 +101,7 @@ export default Component.extend({
 so that it can eventually become, once refactored to native classes:
 
 ```js
-export class extends Component {
+export default class extends Component {
     someMethod() {
         this.actionName();
     }
@@ -164,7 +164,7 @@ In native class syntax, this looks like:
 
 ```js
 // app/controllers/some-route.js
-export class SomeRouteController extends Controller {
+export default class SomeRouteController extends Controller {
     @service router;
 
     someMethod() {
@@ -176,6 +176,44 @@ export class SomeRouteController extends Controller {
         this.router.transitionTo('some.other.route');
     }
 }
+```
+
+If you are using `ember-route-action-helper`, you can instead
+move your action from the route to the controller.
+
+From:
+
+```js
+// app/routes/some-route.js
+export default Route.extend({
+    actions: {
+        someRouteAction() {
+            // do something
+        }
+    }
+}
+```
+
+```hbs
+<!-- app/templates/some-route.hbs -->
+<SomeComponent myAction={{route-action 'someRouteAction'}} />
+```
+
+To:
+
+```js
+// app/routes/some-controller.js
+export default class SomeController extends Controller {
+    @action
+    someAction() {
+        // so something
+    }
+}
+```
+
+```hbs
+<!-- app/templates/some-route.hbs -->
+<SomeComponent myAction={{someAction}} />
 ```
 
 ## How We Teach This
--- a/text/0632-deprecate-string-based-actions.md
+++ b/text/0632-deprecate-string-based-actions.md
@@ -7,7 +7,7 @@
 
 ## Motivation
 
-> In Ember 1.13 we introduced [closure actions](https://github.com/emberjs/rfcs/blob/00ac2685c86f27d41547012903f485a4ef338d27/active/0000-improved-actions.md), which have been recommended over string based actions. This officially deprecates string based actions in favor of closure actions,
+In Ember 1.13 we introduced [closure actions](https://github.com/emberjs/rfcs/blob/00ac2685c86f27d41547012903f485a4ef338d27/active/0000-improved-actions.md), which have been recommended over string based actions. This officially deprecates string based actions in favor of closure actions,
 which provide many benefits such as
 
 1. simpler passing through components
@@ -118,20 +118,20 @@ export class extends Component {
 
 ## How We Teach This
 
-> We already teach closure actions exclusively in the guides. Other than a new
+We already teach closure actions exclusively in the guides. Other than a new
 deprecation guide, no additional material should be necessary.
 
 ## Drawbacks
 
-> Some older applications have thousands of string based actions throughout their codebases.
+Some older applications have thousands of string based actions throughout their codebases.
 We should consider writing a codemod to ease the transition.
 
 ## Alternatives
 
-> Closure actions are now the standard in Ember, so other alternatives have not been
+Closure actions are now the standard in Ember, so other alternatives have not been
 seriously considered.
 
 ## Unresolved questions
 
-> Can we also deprecate the use of the `actions` hash? Can we also deprecate using `this.actions`?
+Can we also deprecate the use of the `actions` hash? Can we also deprecate using `this.actions`?
 Can the `target` property of `@ember/controller` and `@ember/component` be deprecated?
--- a/text/0632-deprecate-string-based-actions.md
+++ b/text/0632-deprecate-string-based-actions.md
@@ -1,6 +1,6 @@
 - Start Date: 2020-05-23
 - Relevant Team(s): Ember.js
-- RFC PR: (after opening the RFC PR, update this with a link to it and update the file name)
+- RFC PR: https://github.com/emberjs/rfcs/pull/632
 - Tracking: (leave this empty)
 
 # Deprecate String Based Actions
@@ -38,7 +38,9 @@ Instead, users are recommended to inject the router service.
 
 Users will transition to closure actions.
 
-From:
+While the action helper and modifiers are still supported with non-string arguments, the refactoring recommendation is to switch to the on modifier and @action decorator directly. It is possible that the action helpers and modifiers will be deprecated in future RFCs.
+
+For example, take this button that sends the foo action with the default click event:
 
 ```js
 export default Component.extend({
@@ -54,7 +56,12 @@ export default Component.extend({
 });
 ```
 
-To:
+```hbs
+<button {{action 'actionName'}}>Action Label</button>
+```
+
+Although this could be refactored to:
+
 
 ```js
 export default Component.extend({
@@ -68,16 +75,45 @@ export default Component.extend({
 });
 ```
 
-From:
-
-```hbs
-<button {{action 'actionName'}}>Action Label</button>
-```
-
-To:
-
 ```hbs
 <button {{action this.actionName}}>Action Label</button>
+```
+
+it is recommended that the code is refactored instead to:
+
+```js
+export default Component.extend({
+    someMethod() {
+        this.actionName();
+    },
+
+    action(function actionName() {
+        // do something
+    })
+});
+```
+
+```hbs
+<button {{on "click" this.actionName}}>Action Label</button>
+```
+
+so that it can eventually become:
+
+```js
+export class extends Component {
+    someMethod() {
+        this.actionName();
+    }
+
+    @action
+    function actionName() {
+        // do something
+    }
+}
+```
+
+```hbs
+<button {{on "click" this.actionName}}>Action Label</button>
 ```
 
 ## How We Teach This
@@ -98,3 +134,4 @@ seriously considered.
 ## Unresolved questions
 
 > Can we also deprecate the use of the `actions` hash? Can we also deprecate using `this.actions`?
+Can the `target` property of `@ember/controller` and `@ember/component` be deprecated?
--- a/text/0632-deprecate-string-based-actions.md
+++ b/text/0632-deprecate-string-based-actions.md
@@ -79,6 +79,7 @@ export default Component.extend({
 <button {{action this.actionName}}>Action Label</button>
 ```
 
+Because the action modifier might be deprecated in the future,
 it is recommended that the code is refactored instead to:
 
 ```js
@@ -97,7 +98,7 @@ export default Component.extend({
 <button {{on "click" this.actionName}}>Action Label</button>
 ```
 
-so that it can eventually become:
+so that it can eventually become, once refactored to native classes:
 
 ```js
 export class extends Component {
@@ -114,6 +115,67 @@ export class extends Component {
 
 ```hbs
 <button {{on "click" this.actionName}}>Action Label</button>
+```
+
+For the send api, it is now generally possible to avoid string based actions in the actions
+hash of any route, because the router service has all the methods that were once only
+available in routes. If there are any route actions, it is recommended to move them to a
+component, service, or a controller.
+
+For example, if you currently have a route action accessed by the `send` api like this:
+
+```js
+// app/routes/some-route.js
+
+actions: {
+    myAction() {
+        this.transitionTo('some.other.route');
+    }
+}
+```
+
+and are calling it from using `send` in your controller like this:
+
+```js
+// app/controllers/some-route.js
+someMethod() {
+    this.send('myAction');
+}
+```
+
+You can move the action from the route to the controller:
+
+```js
+// app/controllers/some-route.js
+export default Controller.extend({
+    router: service(),
+
+    someMethod() {
+        this.myAction();
+    }
+
+    myAction: action(function() {
+        this.router.transitionTo('some.other.route');
+    })
+});
+```
+
+In native class syntax, this looks like:
+
+```js
+// app/controllers/some-route.js
+export class SomeRouteController extends Controller {
+    @service router;
+
+    someMethod() {
+        this.myAction();
+    }
+
+    @action
+    myAction() {
+        this.router.transitionTo('some.other.route');
+    }
+}
 ```
 
 ## How We Teach This
--- a/text/0632-deprecate-string-based-actions.md
+++ b/text/0632-deprecate-string-based-actions.md
@@ -87,7 +87,7 @@ export default Component.extend({
         this.actionName();
     },
 
-    action(function actionName() {
+    actionName: action(function() {
         // do something
     })
 });
@@ -106,7 +106,7 @@ export class extends Component {
     }
 
     @action
-    function actionName() {
+    actionName() {
         // do something
     }
 }


text/0000-deprecate-string-based-actions.md
```diff
@@ -0,0 +1,89 @@
+- Start Date: 2020-05-23
+- Relevant Team(s): Ember.js
+- RFC PR: (after opening the RFC PR, update this with a link to it and update the file name)
+- Tracking: (leave this empty)
+
+# Deprecate String Based Actions
+
+## Motivation
+
+> In Ember 1.13 we introduced [closure actions](https://github.com/emberjs/rfcs/blob/00ac2685c86f27d41547012903f485a4ef338d27/active/0000-improved-actions.md) have been recommended over string based actions. This officially deprecates string based actions in favor of closure actions,
```
rwjblue, line 10: FWIW, I think the general phrasing here is a bit off. I don't think we are deprecating string based actions in favor of closure actions so much as we are deprecating them in favor of passing functions. Yes, this is what `(action` returns, but we _really_ don't want folks using `(action` at all. They should use the `@action` decorator along with `fn` (when you need to curry an argument).

text/0000-deprecate-string-based-actions.md
```diff
@@ -0,0 +1,89 @@
...
+which provide many benefits such as
+
+1. simpler passing through components
+2. less need for an actions hash
```
rwjblue, line 14: No need, right?

text/0000-deprecate-string-based-actions.md
```diff
@@ -0,0 +1,89 @@
...
+3. ability to return a value
+4. typing in TypeScript
+5. compatible with Glimmer Components
```
rwjblue, line 17: I'm not sure this is actually true, have you tested? I'm **pretty sure** you can use `@action whatever() {}` in a `@glimmer/component` and then use something like `<button {{action 'whatever'}}></button>` just fine.

text/0000-deprecate-string-based-actions.md
```diff
@@ -0,0 +1,89 @@
...
+
+## Detailed Design
```
rwjblue, line 19: This is missing quite a bit of info I think. 

* Which framework classes are affected?
* What are the replacement values for each of the various use cases for `.send`?
* Shouldn't it also deprecate directly accessing `this.actions` (this is a thing some folks have done, and the `@action` decorator still populates this actions hash)?
* Does it deprecate specifying an `actions` hash?
* What is the exact prose that will be included in the deprecation guide entry?

text/0000-deprecate-string-based-actions.md
```diff
@@ -0,0 +1,89 @@
...
+
+To:
+
+```js
+export default Component.extend({
+    someMethod() {
+        this.actionName();
+    }
+
+    actionName() {
```
rwjblue, line 54: This should use `@action`

text/0000-deprecate-string-based-actions.md
```diff
@@ -0,0 +1,89 @@
...
+});
+```
+
+To:
+
+```js
+export default Component.extend({
+    someMethod() {
+        this.actionName();
+    }
```
rwjblue, line 52: this is invalid syntax, when using `.extend` you'd need a comma here

text/0000-deprecate-string-based-actions.md
```diff
@@ -0,0 +1,89 @@
...
+From:
+
+```hbs
+<button {{action 'actionName'}}>Action Label</button>
+```
+
+To:
+
+```hbs
+<button {{action this.actionName}}>Action Label</button>
```
rwjblue, line 69: Honestly, we should not suggest using `action` here at all. While this RFC doesn't explicitly deprecate `{{action` / `(action` we shouldn't suggest folks migrate from `{{action 'foo'}}` to `{{action this.foo}}` when we _really think_ they ought to be using `{{on 'click' this.foo}}`. Doing that migration is just forcing double work / churn on folks.

text/0000-deprecate-string-based-actions.md
```diff
@@ -0,0 +1,89 @@
...
+
+## How We Teach This
+
+> We already teach closure actions exclusively in the guides. Other than a new
+deprecation guide, no additional material should be necessary.
+
+## Drawbacks
+
+> Some older applications have thousands of string based actions throughout their codebases.
+We should consider writing a codemod to ease the transition.
```
rwjblue, line 80: This RFC's detailed design or transition path should include the details that the codemod should use. What would the codemod's heuristics be? How would it select what to update? How would it determine what to update _to_?

text/0000-deprecate-string-based-actions.md
Gaurav0: We are not currently deprecating use of the `action` helper with a function.

text/0000-deprecate-string-based-actions.md
Gaurav0: I wasn't aware of this. I thought a quoted action required the actions hash. TIL.

text/0000-deprecate-string-based-actions.md
Gaurav0: Not entirely. I believe it is still need for use in routes for the `error` method.

text/0000-deprecate-string-based-actions.md
Gaurav0: `@action` will not work with `.extend(`; i.e. it would require a native class.

text/0000-deprecate-string-based-actions.md
Gaurav0: correct. Thanks.

text/0000-deprecate-string-based-actions.md
Gaurav0: Do you have an example RFC that does this?

text/0000-deprecate-string-based-actions.md
Gaurav0: Are we planning to deprecate the `action` helper entirely in the near future?

text/0000-deprecate-string-based-actions.md
rwjblue: Ya, exactly, that is why that isn't what I said 😛. 

I'm saying that we are deprecating using any form of `action` (either helper or modifier form) when passing a string, but the overall rationale is that it is much easier to "just pass a function". In the Octane programming model, you don't use `{{action` or `(action`.

text/0000-deprecate-string-based-actions.md
rwjblue: Well, it _does_ require an actions hash. The `@action` decorator (either in decorator form or computed style syntax) **creates** an actions hash for compatibility purposes.  


https://github.com/emberjs/rfcs/blob/master/text/0408-decorators.md#actions has more details on how that works, a small snippet:

> This would be implemented by creating the class's actions object and assigning the method to it, setting up inheritance and such in the process. The decorator leaves the method definition on the class, where it can be called like a normal method would, including super functionality. This maintains compatibility with the classic object model, while making the actions namespace an implementation detail rather than something users need to know about.

text/0000-deprecate-string-based-actions.md
Gaurav0: I was thinking it might be easier not to encourage people to go all the way to Octane recommendations, not to mention easier to write a codemod that just converted string based actions to closure actions?

text/0000-deprecate-string-based-actions.md
rwjblue: I'm using `@action` here to show what I mean, you can easily use that decorator in a classic Ember.Object.extend style class (just like computed):

```js
  actionName: action(function() {

  }),
})
```

text/0000-deprecate-string-based-actions.md
rwjblue: Neither helper nor modifier form of `{{action` are part of the octane programming model. Therefore, it seems very likely that both forms will be fully deprecated at some point in the future. 

IMHO, the only question here is what "soon" actually means. It seems unlikely that `{{action` or `(action` would be something that we would attempt to _remove_ in a 4.0 (it will take time for applications to migrate), but I would expect them to have been deprecated before then.

text/0000-deprecate-string-based-actions.md
Gaurav0: > Shouldn't it also deprecate directly accessing this.actions (this is a thing some folks have done, and the @action decorator still populates this actions hash)?

Added to unresolved questions for the moment.

> Does it deprecate specifying an actions hash?

Added to unresolved questions for the moment.

text/0000-deprecate-string-based-actions.md
rwjblue: The issue is, rewriting from one deprecated thing (deprecated in this RFC) to another thing that is likely to be deprecated soon (when there is a perfectly good, and easier to understand alternative that will not be deprecated) is just introducing churn for no good reason.

text/0000-deprecate-string-based-actions.md
locks: This means that there is no need for an actions hash in the previous discussion, right?

text/0000-deprecate-string-based-actions.md
Gaurav0: > IMHO, the only question here is what "soon" actually means.

Only if we intend to deprecate the action helper and modifier prior to deprecating Ember Components, Controllers, and Ember.Object, would the user really experience extra churn. However, if the action helper and modifier will exist until these are deprecated, then no extra churn will be created and there is no need to immediately stop using `action` modifier and helper altogether.

text/0000-deprecate-string-based-actions.md
Gaurav0: I'm doubtful we should be recommending this.

text/0000-deprecate-string-based-actions.md
Gaurav0: While it is possible to use other modifiers with classic Ember components, the need to use a kind of syntax not shown in any guides makes this less ergonomic than just continuing to use the `action` modifier IMO. It will also be significantly more difficult to codemod.

text/0000-deprecate-string-based-actions.md
mehulkar: It was recommended in the 3.11 release blog post https://blog.emberjs.com/2019/07/15/ember-3-11-released.html, we used this syntax pretty extensively while migrating!

text/0000-deprecate-string-based-actions.md
```diff
@@ -0,0 +1,100 @@
...
+
+## Alternatives
+
+> Closure actions are now the standard in Ember, so other alternatives have not been
+seriously considered.
+
+## Unresolved questions
```
mehulkar, line 98: A couple other questions I have:

1. Can the `target` property of core classes be deprecated also? I didn't dig deep enough, so I'm not sure what else it's used for
1. Can the `ActionHandler` mixin be deprecated entirely? It only handles the `send` method and merges action hashes, so in theory it could be deprecated. (Although, since it's private API, I'm not sure it's worth signaling that)

text/0000-deprecate-string-based-actions.md
mehulkar: Does an an alternative need to be detailed at all here?

text/0000-deprecate-string-based-actions.md
rwjblue: I’m not really following the logic here. What is the syntax that you are referring to (that isn’t in the rest of the guides)? Both decorators (if that is what you meant) and modifiers (if that is what you meant) are used heavily in the guides...

text/0000-deprecate-string-based-actions.md
rwjblue: I read that (in the list above right?) as referring to you _the user_ specifying an actions hash.

text/0000-deprecate-string-based-actions.md
rwjblue: If there is a gap in the guides (though I’m not sure there is) then we should fix it. That is something this RFC should work to address. 

That fact does **not** change the direction we should advise folks to migrate to. Migrating as proposed in this RFC at the moment would be a non-trivial waste of time, and ultimately (IMO) not worth moving forward RFC wise.

text/0000-deprecate-string-based-actions.md
```diff
@@ -0,0 +1,100 @@
...
+
+> Can we also deprecate the use of the `actions` hash? Can we also deprecate using `this.actions`?
```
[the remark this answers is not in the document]
Gaurav0: I think I agree. Would you be willing to PR this PR and help out?

text/0000-deprecate-string-based-actions.md
Gaurav0: > Can the target property of core classes be deprecated also? I didn't dig deep enough, so I'm not sure what else it's used for.

Not sure either. I can add it to the list of unresolved questions.

> Can the ActionHandler mixin be deprecated entirely? It only handles the send method and merges action hashes, so in theory it could be deprecated. (Although, since it's private API, I'm not sure it's worth signaling that)

I don't think it needs a separate deprecation, and throwing two deprecations when using `send` is noisy. It should be removable once `send` is removed from Ember.

text/0000-deprecate-string-based-actions.md
Gaurav0: @mehulkar I think an alternative should be detailed here, that is why I did so. It **definitely** needs to be detailed in the deprecation guide.

text/0000-deprecate-string-based-actions.md
Gaurav0: I'm not sure if it is possible for a deprecation warning to distinguish between the user providing an actions hash and Ember itself providing it?

text/0000-deprecate-string-based-actions.md
Gaurav0: The `action(function() {...})` syntax

text/0000-deprecate-string-based-actions.md
Gaurav0: > Which framework classes are affected?

Added.

> What are the replacement values for each of the various use cases for .send?

Partially added.

> What is the exact prose that will be included in the deprecation guide entry?

Help wanted here.

text/0000-deprecate-string-based-actions.md
mehulkar: Maybe this RFC / deprecation guide can say something like:

> While the action helper and modifiers are still supported with non-string arguments, the refactoring recommendation is to switch to the `on` modifier and `@action` decorator directly. It is likely that the `action` helpers and modifiers will be deprecated in future RFCs.

and give an example like this:

> For example, take this button that sends the `foo` action with the default click event:

> ```hbs
> <button {{action "foo"}}></button>
> ```

> Although this _could_ be refactored to:

> ```hbs
> <button {{action this.foo}}></button>
> ```

> where `foo()` is implemented in the backing context in the actions hash:

> ```js
> Component.extend({
>   actions: {
>     foo() {}
>   }
> })
> ```

> it is recommended that the code is refactored instead to:

> ```hbs
> <button {{on "click" this.foo}}></button>
> ```

> ```js
> Component.extend({
>   foo: action(function() {})
> })
> ```

This could also be written as the with the `diff` syntax to make it more clear what changed. It doesn't quite capture the case of refactoring away from an action that bubbles, but you could link to a page in the guides that covers it.

text/0000-deprecate-string-based-actions.md
mehulkar: Maybe this RFC / deprecation guide can say something like:

> While the action helper and modifiers are still supported with non-string arguments, the refactoring recommendation is to switch to the `on` modifier and `@action` decorator directly. It is likely that the `action` helpers and modifiers will be deprecated in future RFCs.

and give an example like this:

> For example, take this button that sends the `foo` action with the default click event:

> ```hbs
> <button {{action "foo"}}></button>
> ```

> Although this _could_ be refactored to:

> ```hbs
> <button {{action this.foo}}></button>
> ```

> where `foo()` is implemented in the backing context in the actions hash:

> ```js
> Component.extend({
>   actions: {
>     foo() {}
>   }
> })
> ```

> it is recommended that the code is refactored instead to:

> ```hbs
> <button {{on "click" this.foo}}></button>
> ```

> ```js
> Component.extend({
>   foo: action(function() {})
> })
> ```

This could also be written as the with the `diff` syntax to make it more clear what changed. It doesn't quite capture the case of refactoring away from an action that bubbles, but you could link to a page in the guides that covers it.

Gaurav0: This seems good. :+1:

text/0000-deprecate-string-based-actions.md
Gaurav0: Ok. I'm putting this in. Where in the guides can we discuss this?

text/0000-deprecate-string-based-actions.md
rwjblue: Since the mixin is on the global, it might still be good to explicitly issue a deprecation if you access it. The internal usages in Ember can be updated to avoid triggering that secondary / knock on deprecation.

text/0000-deprecate-string-based-actions.md
rwjblue: Sure, but by that logic we should avoid `.extend()` here too since the guides don’t cover it either. 😜🤷‍♂️

text/0632-deprecate-string-based-actions.md
```diff
@@ -0,0 +1,137 @@
...
+        // do something
+    }
+});
+```
+
+```hbs
+<button {{action this.actionName}}>Action Label</button>
+```
+
+it is recommended that the code is refactored instead to:
```
knownasilya, line 82:
```suggestion
it is recommended that the code is refactored instead to, because the above example looses its `this` context:
```

text/0632-deprecate-string-based-actions.md
```diff
@@ -0,0 +1,137 @@
...
+
+```js
+export default Component.extend({
+    someMethod() {
+        this.actionName();
+    },
+
+    action(function actionName() {
```
knownasilya, line 90:
```suggestion
    actionName: action(function actionName() {
```

text/0632-deprecate-string-based-actions.md
```diff
@@ -0,0 +1,137 @@
...
+so that it can eventually become:
+
+```js
+export class extends Component {
+    someMethod() {
+        this.actionName();
+    }
+
+    @action
+    function actionName() {
```
knownasilya, line 109:
```suggestion
   actionName() {
```

text/0632-deprecate-string-based-actions.md
Gaurav0: The action modifier does bind the this context correctly.

knownasilya: That was about the bare function above.

Gaurav0: Yes, if there is an action modifier in the template, the function can be bare.

text/0000-deprecate-string-based-actions.md
rwjblue: Ya, should be possible. Happy to spit ball specific implementation details to make that work, but internally when Ember merges mixins (the POJO passed to `.extend()` is technically treated as a mixin) we can handle explicit `actions` property there differently than the `action` helper imperatively creating it.

text/0000-deprecate-string-based-actions.md
rwjblue: Not 100% sure, maybe a "classic interop" thing in the advanced topics? 

@emberjs/learning-team-managers - Thoughts?

text/0632-deprecate-string-based-actions.md
knownasilya: If I remember correctly, once you do something like `someaction={{this.actionName}}` it breaks down. Which seems like a valid thing to want to do.

Gaurav0: Correct. However, you can fix it as `@someaction={{action this.actionName}}`.

text/0632-deprecate-string-based-actions.md
```diff
@@ -0,0 +1,137 @@
...
+}
+```
+
+```hbs
+<button {{on "click" this.actionName}}>Action Label</button>
+```
+
+## How We Teach This
+
+> We already teach closure actions exclusively in the guides. Other than a new
```
mehulkar, line 121:
```suggestion
We already teach closure actions exclusively in the guides. Other than a new
```

text/0632-deprecate-string-based-actions.md
```diff
@@ -0,0 +1,137 @@
...
+deprecation guide, no additional material should be necessary.
+
+## Drawbacks
+
+> Some older applications have thousands of string based actions throughout their codebases.
```
mehulkar, line 126:
```suggestion
Some older applications have thousands of string based actions throughout their codebases.
```

text/0632-deprecate-string-based-actions.md
```diff
@@ -0,0 +1,137 @@
...
+We should consider writing a codemod to ease the transition.
+
+## Alternatives
+
+> Closure actions are now the standard in Ember, so other alternatives have not been
```
mehulkar, line 131:
```suggestion
Closure actions are now the standard in Ember, so other alternatives have not been
```

text/0632-deprecate-string-based-actions.md
```diff
@@ -0,0 +1,137 @@
...
+seriously considered.
+
+## Unresolved questions
+
+> Can we also deprecate the use of the `actions` hash? Can we also deprecate using `this.actions`?
```
mehulkar, line 136:
```suggestion
Can we also deprecate the use of the `actions` hash? Can we also deprecate using `this.actions`?
```

text/0632-deprecate-string-based-actions.md
```diff
@@ -0,0 +1,137 @@
+- Start Date: 2020-05-23
+- Relevant Team(s): Ember.js
+- RFC PR: https://github.com/emberjs/rfcs/pull/632
+- Tracking: (leave this empty)
+
+# Deprecate String Based Actions
+
+## Motivation
+
+> In Ember 1.13 we introduced [closure actions](https://github.com/emberjs/rfcs/blob/00ac2685c86f27d41547012903f485a4ef338d27/active/0000-improved-actions.md), which have been recommended over string based actions. This officially deprecates string based actions in favor of closure actions,
```
mehulkar, line 10:
```suggestion
In Ember 1.13 we introduced [closure actions](https://github.com/emberjs/rfcs/blob/00ac2685c86f27d41547012903f485a4ef338d27/active/0000-improved-actions.md), which have been recommended over string based actions. This officially deprecates string based actions in favor of closure actions,
```

text/0632-deprecate-string-based-actions.md
```diff
@@ -0,0 +1,137 @@
...
+
+The `send` api may currently be used in components, controllers, and routes.
+It is used to target the `actions` hash. In calling the `actions` hash of the
+same object, `send` can be replaced by a function call once the method is
+moved outside the `actions` hash.
+
+For cases where `send` is used to go from a controller to a route or from
+a route to a parent route, no direct replacement for this "action bubbling"
+behavior will be provided.
+Instead, users are recommended to inject the router service.
```
patocallaghan, line 35: Could you explain this part? How does the router service remove the need to use "action bubbling"? As someone who has lots of instances of using `send` in controllers/routes it would be good to have a concrete example in the **Transition Path** section too.

Gaurav0: Basically, there is no longer a need to use actions in a route because the router service exposes all the methods that used to be available on routes.

It may be that you have a good use case for keeping your actions in a route. If you do, let me know.

This is the idea: if you have an action in your route that would transition to a different route, move it to your component or any service and just call the routerService's transitionTo method.

patocallaghan: Thanks, that clarifies things a bit. I wonder if you could include some of this text in the rfc because that sentence is a bit vague. I'd also still recommend having an example in the Transition Path section.

luxferresum:
I really like this, *but* there is one very important use case that we need to cover before we can do this. And this is the `Route#refresh` method.

While you proposed a [new `refresh` method](https://github.com/emberjs/rfcs/pull/631) on the RouterService, which is good, it does *not* cover this use-case! There must be a way to refresh the current Route from the template/controller.

Gaurav0:
> I really like this, _but_ there is one very important use case that we need to cover before we can do this. And this is the `Route#refresh` method.
> 
> While you proposed a [new `refresh` method](https://github.com/emberjs/rfcs/pull/631) on the RouterService, which is good, it does _not_ cover this use-case! There must be a way to refresh the current Route from the template/controller.

Why can't you inject the router sevice into the controller or whatever component backs the template?

luxferresum:
The problem is that I only want to refresh the current route, not all parent routes.

text/0632-deprecate-string-based-actions.md
```diff
@@ -0,0 +1,237 @@
...
+3. The use of a string parameter to the `action` modifier.
+
+The `send` api may currently be used in components, controllers, and routes.
+It is used to target the `actions` hash. In calling the `actions` hash of the
+same object, `send` can be replaced by a function call once the method is
+moved outside the `actions` hash.
+
+For cases where `send` is used to go from a controller to a route or from
+a route to a parent route, no direct replacement for this "action bubbling"
+behavior will be provided.
```
sandstrom, line 34: Maybe we could elaborate on this a bit?

Our app has probably ~30% of all actions in routes. Right now this is treated more like a side-effect, but for us the biggest change here would probably be what to do with all our route actions? Or perhaps I've misunderstood something?

Since things like routable components (or something to that effect) isn't available, there isn't really a clear migration path? (that I'm aware of)

Also, if all the docs are already teaching the closure actions, maybe this could be put on hold until a replacement for "actions in routes" is available?

Overall still in favor of the change, but the effect this would have on action bubbling could be expanded a bit.

text/0632-deprecate-string-based-actions.md
```diff
@@ -0,0 +1,237 @@
+- Start Date: 2020-05-23
+- Relevant Team(s): Ember.js
+- RFC PR: https://github.com/emberjs/rfcs/pull/632
+- Tracking: (leave this empty)
+
+# Deprecate String Based Actions
+
+## Motivation
+
+In Ember 1.13 we introduced [closure actions](https://github.com/emberjs/rfcs/blob/00ac2685c86f27d41547012903f485a4ef338d27/active/0000-improved-actions.md), which have been recommended over string based actions. This officially deprecates string based actions in favor of closure actions,
```
joukevandermaas, line 10: IMO this is not a motivation for deprecating string based actions, rather it is a motivation for adopting closure actions in new code. I agree with those points, but deprecating functionality like this that doesn't hurt anyone is not "stability without stagnation", it is just creating churn for people with old apps.

My recent experience coming back to an Ember 3.x app (started in 1.12) after a few years and trying to get to 4.x is that this framework has *incredible* amounts of churn because of deprecation like this. It just creates work for anyone with an older app without any stated benefit. I have spent weeks now pointlessly refactoring from one pattern to another simply because it was deprecated, while looking at RFCs there is no actual reason given for the deprecation (outside of, "we have invented another way to do it").

What I would really love to see in this motivation section is some reason why this feature is specifically harmful to keep around. For example, maybe it is complicating the code in a way that makes adding other features harder. That would at least be a compelling motivation to remove it.

PS. I hope I don't come across as unnecessarily harsh, I really do not mean it that way. I hope you understand where I am coming from.

wagenet:
Is there still interest in moving this forwards?

Gaurav0:
@ef4 Someone would need to take over this from me.

wagenet:
@chriskrycho do you want to include this in #832?

chriskrycho:
That would make sense to me. I keep forgetting these still exist—it got at least partially cleaned up by removing `sendAction` for v4.0, just needs the rest done.

wagenet:
Thanks for your work here! This will be addressed as part of #832.

PR #638, opened 2020-06-12
Interactive way to create new Ember apps

MelSumner:
This RFC introduces an interactive way to create a new Ember app.

[Rendered](https://github.com/emberjs/rfcs/blob/441c1f33c98fe71ae82266def661082c1b004474/text/0000-interactive-app-creation.md)


text/0000-interactive-app-creation.md
```diff
@@ -0,0 +1,142 @@
...
+
+Lastly, this RFC is equally motivated by providing 'global' application improvements to the `ember new` workflow. Briefly, these improvements can be summarized as:
+- improvements to applications caused by achieving WCAG SC-3.1.1 that aren't strictly accessibility-related (as outlined in [RFC 635](https://github.com/emberjs/rfcs/pull/635))
+- flattening the learning curve for new users by providing them with an interactive environment and helpful feedback during troubleshooting, error handling, etc.
+
+## Detailed design
+
+### Prior Art
```
NullVoxPopuli, line 31: maybe also Angular's Schematics.

text/0000-interactive-app-creation.md
```diff
@@ -0,0 +1,142 @@
...
+1. `ember new`
+1. `ember new --interactive`
+1. `ember new -i`
+1. `ember new my-app -i`
+
+### App/Addon 
+
+Starting a new Ember app or addon as we do now (at the time of this writing) will not change. If a user types `ember new my-app` they will still get a new Ember app generated in the usual ways. However, if the user types `ember new` they will enter into an interactive workflow that will ask them a few key questions about their application. 
+
+If a user were to type in existing flags that are also questions in the interactive workflow (i.e., `ember new my-app --lang en-US --interactive`), it will skip the question entirely.
```
NullVoxPopuli, line 66: I don't think this'd be a technical limitation. I can think of backing the interactive wizard with a statechart that would determine which questions to ask based on the presence or lack of flags passed at invocation time

text/0000-interactive-app-creation.md
```diff
@@ -0,0 +1,142 @@
...
+
+As this is a wholly new idea, it should be documented and added to the guides along with screenshots of the new workflow. Until that prototype exists, this section will largely remain empty. However, we intend to explain it similarly to Vue's documentation for a similar feature (https://cli.vuejs.org/guide/creating-a-project.html#vue-create).  
+
+## Drawbacks
+
+It could be considered a drawback to make this not the default for new Ember apps, even if an app name is defined. 
+
+The general drawback with wizards is that it can lead to a large number of combinations, which increases the risk for fragmentation. We believe that we have limited this risk by limiting the number of questions asked, but it is still a risk and should be identified as such.
+
+## Alternatives
```
NullVoxPopuli, line 131: I think Angular schematics cover a large chunk of what we'd want out of blueprints, optional interactivity, and interactivity provided by addons + additional blueprints.

as part of this RFC, I'd like to see that path explored a bit, as it already exists, and would provide a lot of the future functionality

text/0000-interactive-app-creation.md
Gaurav0: Also vue-cli?

text/0000-interactive-app-creation.md
```diff
@@ -0,0 +1,142 @@
...
+
+If a user were to type in existing flags that are also questions in the interactive workflow (i.e., `ember new my-app --lang en-US --interactive`), it will skip the question entirely. 
+
+When a user types `ember new` into their command line, they will be asked these questions (defaults indicated by a `>`):
+
+- `Is this an app or an addon? (use arrow keys)`
+  - `> app`
+  - `addon`
+- `Please provide the name of your app/addon:`
+- `Please provide the spoken/content language of your app/addon: (use arrow keys)`
```
Wolfr, line 74: Suggestion:

What is the default language of your app/addon? Please type your ISO639-1 language code. For example, the language code for English is "en" and the language code for Chinese is "chi". A full list can be found at https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes .

Motivation:

* Ensure equality for all languages
* Avoids discussion about which languages are "common"
* The wording for "default" gives people who are planning to do a multilingual app a path forward, whereas, the text above doesn't.

text/0000-interactive-app-creation.md
```diff
@@ -0,0 +1,142 @@
...
+
+It could be considered a drawback to make this not the default for new Ember apps, even if an app name is defined. 
+
+The general drawback with wizards is that it can lead to a large number of combinations, which increases the risk for fragmentation. We believe that we have limited this risk by limiting the number of questions asked, but it is still a risk and should be identified as such.
+
+## Alternatives
+
+1. We could decide on a different list of questions. See the notes from the [strike team discussion on May 20th](https://github.com/ember-a11y/core-notes/blob/ember-a11y/ember-a11y/2020-05/may-20.md) to review the other possibilities considered.
+1. We could make this the default for new Ember apps and add it in the appropriate version as a change.
+1. We could have this workflow be opt-in only, via the `--interactive` flag (e.g., `ember new --interactive`). Using the `ember new` command would error and trigger the help workflow (which would be updated to include the `--interactive` flag).
```
MichalBryxi, line 135: I'm always up for "sane defaults", which from my experience makes the flow more straightforward, expected (guessable) and more pleasant to use. In this case I see the sane default option the originally proposed one `ember new` => wizard.

text/0000-interactive-app-creation.md
MichalBryxi: For no real reason I've always used `en-US` for the language codes, never simple `en`. Would this make any difference?

Wolfr: I can imagine that this would for example determine the accent used by a screenreader if no default is set and different accents (e.g. British, US, Australian) are installed. But I would have to verify to be sure.

text/0000-interactive-app-creation.md
MelSumner: @NullVoxPopuli  Can you explain more? I think what you're describing is beyond the scope of this RFC, but I might be missing something.

text/0000-interactive-app-creation.md
MelSumner: > For no real reason I've always used `en-US` for the language codes, never simple `en`. Would this make any difference?

@MichalBryxi it can change things in small ways, like spellcheck, how things are pronounced, etc.

text/0000-interactive-app-creation.md
```diff
@@ -0,0 +1,142 @@
...
+1. Should the questions be different if it is an app vs addon?
+
+We will also endeavor to provide strong defaults for these questions, based on the results from the community surveys we have been conducting over the past few years and the defaults as stated by Ember itself (either explicitly or implicitly). **If these defaults change over time as the result of community RFCs, the default values or other responses for the interactive questions should be updated as part of those RFCs.**
+
+So, how then do we decide which questions to include and which not to include? It became necessary to then develop a rubric by which the proposed list of questions could be measured. 
+
+### Question Rubric
+We intend for this section to be normative; that is, we intend for it to define how questions are evaluated for inclusion both now and in the future. 
+
+Evaluating criteria:
```
kellyselden, line 47: These are great guidelines!

text/0000-interactive-app-creation.md
kellyselden, line 74: Have you considered the importance of the order of these questions? It seemed strange to me that the lang would be asked before package manager or CI. Maybe I have a preconceived notion that the options that are older/have been around longer should be first, or maybe how much code the question is responsible for changing, where package manager or CI would have more impact.

text/0000-interactive-app-creation.md
```diff
@@ -0,0 +1,142 @@
...
+
+#### App/Addon Name
+There will be no default for this question. The user must enter a response. 
+
+#### App/Addon Language
+
+- the default would be `en-US`
+- if your system has a different language set (which can be confirmed by using the `echo $LANG` command in the terminal), then _that_ language would be shown as the second option
+- if your system was already set to the default language, the second option would not be shown. 
+- for "manually define a language", we will validate the response against the allowed language codes
```
kellyselden, line 94: Should we be responsible for this? I relate this to something like timezones, it's really hard to get right. Are we sure we wouldn't be excluding anyone, like maybe an obsure or brand code is introduced that we don't support? I'm don't know how static the list is.

text/0000-interactive-app-creation.md
MelSumner: @kellyselden having a way to determine the order of the questions asked is a good idea.

text/0000-interactive-app-creation.md
MelSumner: @kellyselden we're already doing this with the `--lang` flag. You're right that it could be a tricky issue but we have a fairly high level of confidence in the library that we're using for this. [(is-language-code)](https://www.npmjs.com/package/is-language-code).

> Are we sure we wouldn't be excluding anyone, like maybe an obscure or brand code is introduced that we don't support?

While new regions can be created (due to things like changes in political landscape), it's unlikely that there will be a new language code. All languages have codes, and even languages that no longer have living native speakers still have language codes. For example, `htc` is the code for the [Hattic](https://en.wikipedia.org/wiki/Hattic_language) language, a historical language from 4,000 years ago.

text/0000-interactive-app-creation.md
```diff
@@ -0,0 +1,142 @@
...
+1. `ember new -i`
+1. `ember new my-app -i`
+
+### App/Addon 
+
+Starting a new Ember app or addon as we do now (at the time of this writing) will not change. If a user types `ember new my-app` they will still get a new Ember app generated in the usual ways. However, if the user types `ember new` they will enter into an interactive workflow that will ask them a few key questions about their application. 
+
+If a user were to type in existing flags that are also questions in the interactive workflow (i.e., `ember new my-app --lang en-US --interactive`), it will skip the question entirely. 
+
+When a user types `ember new` into their command line, they will be asked these questions (defaults indicated by a `>`):
```
gossi, line 68: This section confuses me and I guess needs more explanation:

Above is stated the normative for picking question and this section? Are these examples? Is this the draft for the first iteration?

Some background how this document _can be read/understood_: the top section is hinting this RFC is finding a strategy for such question which is now made inconsistent by providing thm.

text/0000-interactive-app-creation.md
```diff
@@ -0,0 +1,142 @@
...
+
+- `Is this an app or an addon? (use arrow keys)`
+  - `> app`
+  - `addon`
```
gossi, line 72: engines?

text/0000-interactive-app-creation.md
```diff
@@ -0,0 +1,142 @@
...
+- `Please provide the name of your app/addon:`
```
gossi, line 73: Next question:

- 'Where to create your project?:'

Default being the name of the project (from earlier), hitting enter will confirm that

text/0000-interactive-app-creation.md
```diff
@@ -0,0 +1,142 @@
...
+- `Please provide the spoken/content language of your app/addon: (use arrow keys)`
+  - `> en-US`
+  - `my computer's default language`
+  - `manually define a different language`
+- `Pick the package manager to use when installing dependencies:  (use arrow keys)`
+  - `> NPM`
+  - `Yarn`
+  - `ignore/skip`
```
gossi, line 81: 'ignore/skip' shouldn't be here and rather be auto-detected. If so, hide this question completely.

text/0000-interactive-app-creation.md
```diff
@@ -0,0 +1,142 @@
...
+
+- the default would be `en-US`
+- if your system has a different language set (which can be confirmed by using the `echo $LANG` command in the terminal), then _that_ language would be shown as the second option
+- if your system was already set to the default language, the second option would not be shown. 
+- for "manually define a language", we will validate the response against the allowed language codes
+- we have not added a "skip" option here, because the purpose is to focus on improved accessibility in Ember apps.
+
+#### Package Manager
+
+- ignore/skip has been added to cover the use case where the developer is in a workspace
```
gossi, line 99: see above, can be skipped completely when auto-detected.

text/0000-interactive-app-creation.md
```diff
@@ -0,0 +1,142 @@
...
+- for "manually define a language", we will validate the response against the allowed language codes
+- we have not added a "skip" option here, because the purpose is to focus on improved accessibility in Ember apps.
+
+#### Package Manager
+
+- ignore/skip has been added to cover the use case where the developer is in a workspace 
+
+#### CI Option
+
+Separate RFCs should further define more options for CI.
```
gossi, line 103: Also please take into consideration of auto-detection and a possible skip for such a question. Only if there is confidence in detection, anyway present highest probabilty as default.

text/0000-interactive-app-creation.md
```diff
@@ -0,0 +1,142 @@
...
+
+This interactive workflow will also ignore any ~/.ember-cli settings that already exist. A separate RFC should be written that makes it possible for these to things to co-exist and respect the settings of the other.
+
+### Future possibilities
+An example of a future RFC that could be specific to addons, would be to add a `--supported-version` flag for addons and add that to the wizard: 
+
+- `What is the earliest version of Ember that you intend to support? (use arrow keys)`
+  - `> recommended (last two LTS)`
+  - `last LTS`
+  - `manually define a version number`
```
gossi, line 116: From above:

> What is the maintenance cost? *For example, a nice question to add could be something like, Do you want to use a CSS preprocessor? But, because Ember does not deal with SASS, LESS... by default, the interactive workflow will have to install an addon (like `ember-cli-less`) which would imply that the CLI team will have to ensure it never fails.* 

I can say, this is absolutely a nice idea but hardly manageable. A future possibility is to provide hooks for enhancing this by installing more addons (globally or per project). Vue cli did this with v2 already, was happy to see it when I was reading their code.

text/0000-interactive-app-creation.md
Gaurav0: I'm not so sure about this. There is a page on Wikipedia about this, but it may not be up to date: https://en.wikipedia.org/wiki/Category:Languages_without_ISO_639-3_code

However, using a continually updated library is probably good enough since you can always edit index.html afterward.

text/0000-interactive-app-creation.md
```diff
@@ -0,0 +1,142 @@
...
+#### App/Addon Name
+There will be no default for this question. The user must enter a response. 
+
+#### App/Addon Language
+
+- the default would be `en-US`
+- if your system has a different language set (which can be confirmed by using the `echo $LANG` command in the terminal), then _that_ language would be shown as the second option
+- if your system was already set to the default language, the second option would not be shown. 
+- for "manually define a language", we will validate the response against the allowed language codes
+- we have not added a "skip" option here, because the purpose is to focus on improved accessibility in Ember apps.
```
luxferresum, line 95: what if I want to use `ember-intl` and then dont want a `lang` in the `<html>` tag of my `index.html`? Could we allow this as an option and install `ember-intl`? Or should `ember-intl` remove this during install? Or do we agree that there should always be a "default lang" in the `index.html`?

MelSumner: `ember-intl` will update the `lang` attribute (adding it if it's not there). It's still a recommendation to set a default content/spoken app language, for a few reasons, one of which is that we're also adding linting support (via ember-template-lint) to check for it.

text/0000-interactive-app-creation.md
MelSumner: What's the part that you think is not manageable? can you clarify what makes it not manageable? Thanks!

text/0000-interactive-app-creation.md
MelSumner: How do you think this can be auto-detected? Can you tell me more about that?

text/0000-interactive-app-creation.md
gossi: Hard to manage a selection of possible options for addons to install. It's always subjective to the author(s) (if lectored) but I don't think will provide individuals/comapnies a good option. That's what I think, working on an API so you can _customize_ it for yourself or your companies is hitting the mark - although I see this as a future addition, not even tomorrow but farther future

text/0000-interactive-app-creation.md
gossi: It can be detected within workspaces for sure or when using lerna.

Dunno if I remember correctly but I think ember (once) offered me the option to set yarn as default, right? If so, this is also a resource to that.

text/0000-interactive-app-creation.md
gossi: This is addressing the case, when you are _not_ using `ember-intl` and give the document a language.

text/0000-interactive-app-creation.md
MichalBryxi: It's impossible to accommodate to everyone's needs, but I'd say it's good idea to provide with an easily accessible set of commonly agreed useful options. I know that I want addons XY, because I'm comfortable with using them. In my company we know that we want AB, because that's what we've chosen.

But person that is new to the framework or even JS ecosystem in general will have no idea where to start. At that point being given hints that "you might want sass if you want flexible styles", "you might want eslint if you want linting", "you might want prettier if you want unified code formatter", etc. is a great help to get started.

In the future it might be a good idea to abstract this wizard functionality and let the users use external sets. But I would not over-engineer this at this point.

text/0000-interactive-app-creation.md
```diff
@@ -0,0 +1,142 @@
...
+- RFC PR: (after opening the RFC PR, update this with a link to it and update the file name)
+- Tracking: (leave this empty)
+
+# Interactive New Ember App Creation
+
+## Summary
+
+As part of the effort to make new Ember apps more conformant for digital accessibility requirements at a global scale, this RFC proposes an interactive workflow for new Ember apps. This will also have the benefit of assisting new users who prefer an interactive model of new app creation. 
+
+## Motivation
```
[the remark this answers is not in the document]
MelSumner: > The whole motivation section reads as if this is all justified through accessibility

That's correct. It was the motivation for the work represented in this RFC, even if it has other benefits.

text/0000-interactive-app-creation.md
MelSumner: I can update the language here- the prior art was meant to acknowledge the things we reviewed when working on this RFC, and does not intend to suggest that this is the only prior art that exists.

text/0000-interactive-app-creation.md
```diff
@@ -0,0 +1,142 @@
+- Start Date: 2020-06-12
+- Relevant Team(s): ember-cli, framework, learning
+- RFC PR: (after opening the RFC PR, update this with a link to it and update the file name)
+- Tracking: (leave this empty)
+
+# Interactive New Ember App Creation
+
+## Summary
+
+As part of the effort to make new Ember apps more conformant for digital accessibility requirements at a global scale, this RFC proposes an interactive workflow for new Ember apps. This will also have the benefit of assisting new users who prefer an interactive model of new app creation.
```
sandstrom, line 10: > As part of the effort to make new Ember apps more conformant for digital accessibility requirements at a global scale, this RFC proposes an interactive workflow for new Ember apps.

What's the reasoning behind this sentence?

How does an interactive app creation flow help "accessibility requirements at a global scale", what does it mean concretely?

What's the tax/cost for maintaining two flows for creating apps (interactive and non-interactive)? I don't know the answer to this, but someone on the ember-cli team should know. Do we want to take on this tax?

How does an interactive flow make things easier compared to a non-interactive, for a blind/deaf/etc person?

How does other terminal software, editors, etc. work? To effectively develop applications there are usually a ton of other tooling needed. Maybe these problems are better solved at a higher level of abstraction, i.e. at the [editor](https://github.com/microsoft/vscode/issues/88987), [terminal](https://github.com/tspivey/tdsr) and [os](https://www.apple.com/accessibility/mac/) level?

For other parts of Ember, we often contemplate whether a problem is best solved within the Ember project, or outside. For example, the upcoming changes to the build system (Embroider) are a lot about piggy-backing on existing tooling in the JS ecosystem, instead of building everything ourselves.

text/0000-interactive-app-creation.md
```diff
@@ -0,0 +1,142 @@
...
+## Alternatives
+
+1. We could decide on a different list of questions. See the notes from the [strike team discussion on May 20th](https://github.com/ember-a11y/core-notes/blob/ember-a11y/ember-a11y/2020-05/may-20.md) to review the other possibilities considered.
+1. We could make this the default for new Ember apps and add it in the appropriate version as a change.
+1. We could have this workflow be opt-in only, via the `--interactive` flag (e.g., `ember new --interactive`). Using the `ember new` command would error and trigger the help workflow (which would be updated to include the `--interactive` flag).
+1. We could set the default language to US English (the default language of the Ember project) and remove the language question from the interactive workflow.
+1. We could set the default language to US English (the default language of the Ember project) and not have an interactive option at all.
+
+## Unresolved questions
+1. Do we have a good way to make this a good experience for developers who use assistive technology?
```
sandstrom, line 140: How many Ember developers are using assistive technology? What problems to they have with Ember today?

If the underlying purpose here is to "reach" more developers? Are there other groups that are left out at the moment, for example developers that don't speak English very well? Perhaps translating the docs (or parts of them) to more languages would reach more developers?

text/0000-interactive-app-creation.md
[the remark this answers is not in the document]
rwjblue: >  It was the motivation for the work represented in this RFC, even if it has other benefits.

Hmm, I think this is slightly the wrong framing. I think it is perfect that the _reason_ y'all looked into it was accessibility to begin with, but framing the "reason for the feature" as an accessibility one seems flat out wrong. I agree with @gossi here that the motiviation should be refactored a bit to make it clear that DX/UX is the primary factor (and that that is _why_ it meets the accessibility goals as well!).

text/0000-interactive-app-creation.md
rwjblue: > How does an interactive app creation flow help "accessibility requirements at a global scale", what does it mean concretely?

Agree with this critique of the wording. The interactive flow just makes it easier to use the new `--lang` option without remembering it exists, AFAICT that is the only way this relates to A11Y.

> What's the tax/cost for maintaining two flows for creating apps (interactive and non-interactive)? I don't know the answer to this, but someone on the ember-cli team should know. Do we want to take on this tax?

The nice thing here is that **everything** here desugars into command line options that already exist. This means there isn't a duplicated cost of each option/etc, and the only new cost is the wizard / prompting itself. All of the underlying features are already implemented (or in the process of being implemented WRT to `--lang`) and tested.

From an ember-cli team perspective, I think this RFC is a very very large win for simplicity (both ember-cli's codebase, and ember-cli's users).

text/0000-interactive-app-creation.md
```diff
@@ -0,0 +1,142 @@
...
+
+Lastly, this RFC is equally motivated by providing 'global' application improvements to the `ember new` workflow. Briefly, these improvements can be summarized as:
+- improvements to applications caused by achieving WCAG SC-3.1.1 that aren't strictly accessibility-related (as outlined in [RFC 635](https://github.com/emberjs/rfcs/pull/635))
+- flattening the learning curve for new users by providing them with an interactive environment and helpful feedback during troubleshooting, error handling, etc.
+
+## Detailed design
+
+### Prior Art
+It would be remiss to introduce this RFC without mentioning the prior art that we have for wizard-like app creation. In the Ember community, [@gossi](https://gist.github.com/gossi) first introduced this idea with [ember-cli-create](https://github.com/gossi/ember-cli-create). In the wider web community, the Vue community offers this through `vue create` ([see docs](https://cli.vuejs.org/guide/creating-a-project.html#vue-create)).
```
MelSumner, line 32:
```suggestion
It would be remiss to introduce this RFC without mentioning the prior art that we analyzed for wizard-like app creation. In the Ember community, [@gossi](https://gist.github.com/gossi) first introduced this idea with [ember-cli-create](https://github.com/gossi/ember-cli-create). In the wider web community, the Vue community offers this through `vue create` ([see docs](https://cli.vuejs.org/guide/creating-a-project.html#vue-create)).
```

text/0000-interactive-app-creation.md
MelSumner: I've updated the language here to indicate that the prior art listed here is specifically ones used for completing the analysis necessary to write this RFC.

text/0000-interactive-app-creation.md
```diff
@@ -0,0 +1,142 @@
...
+
+### Question Rubric
+We intend for this section to be normative; that is, we intend for it to define how questions are evaluated for inclusion both now and in the future. 
+
+Evaluating criteria: 
+1. Is it hard (or time-consuming) to change it later? *For example, it's easy to remove the `ember-welcome-page` addon, so we did not include it. But it's time consuming to change from `yarn` to `npm` (or vice-versa) so we made sure to include that.* 
+1. Is it difficult to discover the option?
+1. What is the maintenance cost? *For example, a nice question to add could be something like, Do you want to use a CSS preprocessor? But, because Ember does not deal with SASS, LESS... by default, the interactive workflow will have to install an addon (like `ember-cli-less`) which would imply that the CLI team will have to ensure it never fails.* 
+1. Does the question affect another one? *For example, the LTS question can affect the CI file generated by the CI question (setup tests on CI with `ember-try`). Questions that modify the result of other questions too much should be avoided to prevent high maintenance.*
```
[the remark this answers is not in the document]
MelSumner: The question order is based on a few things, but specifically, I wanted to ask the language question while the developers cognitive energy is still fresh. We've already decided that it won't matter if you're building an app or addon for this question, so it's position is fine for both.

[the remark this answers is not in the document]
MelSumner: > Third - How to take out questions?

This is a good question, and I will think about it and add something in here. Do you have anything specific that you'd want considered, @gossi ?

text/0000-interactive-app-creation.md
MelSumner: The normative strategy for determining questions had two purposes: 
1. determine which of our questions on the draft list met the criteria for inclusion
2. provide the framework for adding future questions

text/0000-interactive-app-creation.md
MelSumner: Are engines officially supported?

text/0000-interactive-app-creation.md
MelSumner: I think we want to assume that they are already in the right directory.

text/0000-interactive-app-creation.md
MelSumner: @gossi Ahh yes, you were agreeing with the stated position then. We decided that it would be not manageable, which is why we didn't include it. 👍

text/0000-interactive-app-creation.md
MelSumner: I don't know the total number but I am aware that there are _some_; after digging into this unresolved question a bit more, I am fairly confident that we can remove it. The answer is, "the terminal has to do most of the heavy lifting here" at least where development is concerned. 

For the Ember team, providing clear error messages and ensuring that our guides and API docs are accessible will be the best support we can provide.

text/0000-interactive-app-creation.md
```diff
@@ -0,0 +1,142 @@
...
+## Alternatives
+
+1. We could decide on a different list of questions. See the notes from the [strike team discussion on May 20th](https://github.com/ember-a11y/core-notes/blob/ember-a11y/ember-a11y/2020-05/may-20.md) to review the other possibilities considered.
+1. We could make this the default for new Ember apps and add it in the appropriate version as a change.
+1. We could have this workflow be opt-in only, via the `--interactive` flag (e.g., `ember new --interactive`). Using the `ember new` command would error and trigger the help workflow (which would be updated to include the `--interactive` flag).
+1. We could set the default language to US English (the default language of the Ember project) and remove the language question from the interactive workflow.
+1. We could set the default language to US English (the default language of the Ember project) and not have an interactive option at all.
+
+## Unresolved questions
+1. Do we have a good way to make this a good experience for developers who use assistive technology?
```
MelSumner, line 140:
```suggestion
```

text/0000-interactive-app-creation.md
gossi: Yeah... Sometimes I speak in riddles. I'm pretty sure this question will arise sooner than later. I think ember shouldn't provide any answers here at all (as this would be prejudgement). Instead what I'm saying is - to reach there - it will be nice if ember would provide an API to hook into. That way people can install there own flavors (eg. `yarn install -g ember-cli-new-config-css-modules`) and then it will be picked up in the wizard, to give an idea. Truly the (over)next step.

text/0000-interactive-app-creation.md
gossi: Historically, there weren't npm scopes which changed the game. When I create the package `@foo/bar` I don't want to necessarily create it in that subbed folder - that was also one of the points to create `ec-create`.

text/0000-interactive-app-creation.md
sandstrom: Sounds reasonable, to solve most accessibility concerns on other levels than Ember. There are many interesting projects and great progress at the [editor](https://github.com/microsoft/vscode/issues/88987), [terminal](https://github.com/tspivey/tdsr) and [os](https://www.apple.com/accessibility/mac/) level. But I think these things are mostly better to address outside Ember.

If this RFC isn't mainly about accessibility though (sounds like it's more for first experience), maybe the motivation text in the RFC could be tweaked accordingly.

> The underlying motivation for this RFC is the desire to improve the accessibility of new Ember apps as previously documented in Issue 595/Section 4. Specifically, it aims to prevent new Ember apps from failing legal requirements for accessibility conformance according to WCAG 2.1, the standard used by most countries around the world to inform digital accessibility requirements.

text/0000-interactive-app-creation.md
sandstrom: That it desugars into existing (or future) command line options sounds good. Sort of like the `.ember-cli` file which afaik share much code with the CLI flags.

I don't know enough about ember-cli to determine whether the implementation cost is worth it. My point above was mainly to ask a few (uncomfortable) questions regarding prioritisation, and it's up to you on the core team to decide of course.

rwjblue:
We discussed this in depth at todays core team meeting. There were a few tweaks / suggestions from the team that need to be taken care of, but the team is overwhelmingly in favor of moving this into final comment period once these tweaks are done:

* Tweak the motivation section to have a bit more prose about the massively better developer ergonomics (in addition to the accessibility improvements)
* Add some details to the detailed design that makes it clear that the interactive `ember new` should **not** be considered "stable" (e.g. the exact wording and ordering of questions _could_ change), and it should not be used in a scripting context (for scripting you should absolutely use the desugared flags instead of the interactive workflow).

Thanks for working on this @MelSumner!

rwjblue:
As mentioned in my last comment, now that we have those last few tweaks here we are moving this into final comment period.

text/0000-interactive-app-creation.md
```diff
@@ -0,0 +1,142 @@
...
+### Future possibilities
+An example of a future RFC that could be specific to addons, would be to add a `--supported-version` flag for addons and add that to the wizard: 
+
+- `What is the earliest version of Ember that you intend to support? (use arrow keys)`
+  - `> recommended (last two LTS)`
+  - `last LTS`
+  - `manually define a version number`
+
+### Implementation
+This section is not normative, but provided for extra context for folks who might be new to this idea in general. We could use something like [inquirer.js](https://github.com/SBoudrias/Inquirer.js) to implement this wizard, if the mechanism doesn't already exist within the ember-cli codebase. However, it should be explicitly noted that inquirer is only being used as an example of a library that could be used, and this RFC isn't explicitly defining that inquirer.js should be or will be used.
```
rajasegar, line 115: I did some exploration for another project for having a CLI wizard, looks like there are other lightweight alternatives like [enquirer](https://github.com/enquirer/enquirer) and [prompts](https://github.com/terkelg/prompts). Let the ember-cli team come up with their own recommended library for the same task.

rwjblue: 👍 - Absolutely! The idea that is conveyed here is that we will do prompting. Exactly _how_ is an implementation detail in ember-cli (thats why the prose above says "we could use something like Inquirer.js").

rwjblue:
Awesome work here y'all! Thanks for everyones time and effort to push this forward, let's do it!!!!!

bertdeblock:
Took a stab: https://github.com/ember-cli/ember-cli/pull/9824.

PR #639, opened 2020-06-15
Replace terms blacklist & whitelist

mydea:
This RFC proposes to replace the terms `blacklist` and `whitelist` in Ember CLI.

[Rendered](https://github.com/mydea/rfcs/blob/replace-whitelist-blacklist/text/0000-replace-blacklist-whitelist.md)

### DIFF
--- a/text/0000-replace-blacklist-whitelist.md
+++ b/text/0000-replace-blacklist-whitelist.md
@@ -55,7 +55,7 @@ let app = new EmberApp(defaults, {
 });
 ```
 
-The old keys (`blacklist` and `whitelist`) should be aliased to the new ones, and show a deprecation warning. Functionally, they behave the same.
+The old keys (`blacklist` and `whitelist`) should be aliased to the new ones. Functionally, they behave the same.
 
 If both keys are used at the same time (so either `blacklist` AND `exclude` or `whitelist` AND `include`), an error is thrown.
 
--- a/text/0000-replace-blacklist-whitelist.md
+++ b/text/0000-replace-blacklist-whitelist.md
@@ -59,6 +59,8 @@ The old keys (`blacklist` and `whitelist`) should be aliased to the new ones. Fu
 
 If both keys are used at the same time (so either `blacklist` AND `exclude` or `whitelist` AND `include`), an error is thrown.
 
+The old keys should be removed at some point - there should be a dedicated deprecation RFC for this (at a later point), ideally in time for them to be removed in the next major release.
+
 ## How we teach this
 
 We should update the Ember CLI API docs with the new terms.
--- a/text/0000-replace-blacklist-whitelist.md
+++ b/text/0000-replace-blacklist-whitelist.md
@@ -1,0 +1,70 @@
+- Start Date: 2020-06-15
+- Relevant Team(s): Ember CLI, Learning
+- RFC PR: (after opening the RFC PR, update this with a link to it and update the file name)
+- Tracking: (leave this empty)
+
+# Replace terms blacklist & whitelist in Ember CLI
+
+## Summary
+
+Ember.js prides itself (rightly) on being an inclusive framework. To further improve on this, we should remove the terms "blacklist" and "whitelist" with more neutral replacements.
+
+## Motivation
+
+The terms "blacklist" and "whitelist" are considered racially insenstive. These terms are used in Ember CLI for rather advanced functionality, which should make providing a better alternative naming and deprecating the current one rather easy.
+
+## Detailed design
+
+You can use `blacklist` and `whitelist` to include/exclude certain addons from your build:
+
+https://ember-cli.com/user-guide/#whitelisting-and-blacklisting-assets
+
+We should replace these with `exclude` or `include`:
+
+```js
+// Before
+let app = new EmberApp(defaults, {
+  addons: {
+    blacklist: ["fastboot-app-server"],
+  },
+});
+
+// After
+let app = new EmberApp(defaults, {
+  addons: {
+    exclude: ["fastboot-app-server"],
+  },
+});
+
+// Before
+let app = new EmberApp(defaults, {
+  addons: {
+    whitelist: ["ember-cli-sass"],
+  },
+});
+
+// After
+let app = new EmberApp(defaults, {
+  addons: {
+    include: ["ember-cli-sass"],
+  },
+});
+```
+
+The old keys (`blacklist` and `whitelist`) should be aliased to the new ones, and show a deprecation warning. Functionally, nothing changes.
+
+## How we teach this
+
+We should update the Ember CLI API docs with the new terms.
+
+## Drawbacks
+
+This might be considered "churn" by some. However, it is a small enough change to not really impact anybody meaningfully, while ensuring that Ember.js feels inclusive to everybody.
+
+## Alternatives
+
+Leave it as it is.
+
+## Unresolved questions
+
+- Should we use different replacements than `exclude`/`include`?
--- a/text/0000-replace-blacklist-whitelist.md
+++ b/text/0000-replace-blacklist-whitelist.md
@@ -1,6 +1,6 @@
 - Start Date: 2020-06-15
 - Relevant Team(s): Ember CLI, Learning
-- RFC PR: (after opening the RFC PR, update this with a link to it and update the file name)
+- RFC PR: [#639](https://github.com/emberjs/rfcs/pull/639)
 - Tracking: (leave this empty)
 
 # Replace terms blacklist & whitelist in Ember CLI
@@ -11,7 +11,11 @@ Ember.js prides itself (rightly) on being an inclusive framework. To further imp
 
 ## Motivation
 
-The terms "blacklist" and "whitelist" are considered racially insenstive. These terms are used in Ember CLI for rather advanced functionality, which should make providing a better alternative naming and deprecating the current one rather easy.
+The terms "blacklist" and "whitelist" can be considered racially insenstive. While the origin of these terms in this context is not in itself racially motivated, the fact that in todays context it _can be considered racially insensitive_ has been discussed frequently - see for example [here](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC6148600/), [here](https://bugs.chromium.org/p/chromium/issues/detail?id=981129#c16) or [here](https://www.zdnet.com/article/uk-ncsc-to-stop-using-whitelist-and-blacklist-due-to-racial-stereotyping/).
+
+Other projects have already taken similar steps, for example [Go](https://go-review.googlesource.com/c/go/+/236857/), [Android](<https://android-review.googlesource.com/q/topic:%22soong_inclusive_language%22+(status:open%20OR%20status:merged)>), [curl](https://github.com/curl/curl/pull/5546), [PHPUnit](https://github.com/sebastianbergmann/phpunit/issues/4275), and many more.
+
+These terms are used in Ember CLI for rather advanced functionality, which should make providing a better alternative naming and deprecating the current one rather easy.
 
 ## Detailed design
 
@@ -63,7 +67,13 @@ This might be considered "churn" by some. However, it is a small enough change t
 
 ## Alternatives
 
-Leave it as it is.
+There are also other possible terms that could be used to replace `blacklist` and `whitelist`. For example:
+
+- `allowlist` / `denylist`
+- `allowlist` / `disallowlist`
+- `allow` / `deny`
+- `allowedlist` / `disallowedlist`
+- `safelist` / `blockedlist`
 
 ## Unresolved questions
 
--- a/text/0000-replace-blacklist-whitelist.md
+++ b/text/0000-replace-blacklist-whitelist.md
@@ -13,7 +13,7 @@ Ember.js prides itself (rightly) on being an inclusive framework. To further imp
 
 The terms "blacklist" and "whitelist" can be considered racially insenstive. While the origin of these terms in this context is not in itself racially motivated, the fact that in todays context it _can be considered racially insensitive_ has been discussed frequently - see for example [here](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC6148600/), [here](https://bugs.chromium.org/p/chromium/issues/detail?id=981129#c16) or [here](https://www.zdnet.com/article/uk-ncsc-to-stop-using-whitelist-and-blacklist-due-to-racial-stereotyping/).
 
-Other projects have already taken similar steps, for example [Go](https://go-review.googlesource.com/c/go/+/236857/), [Android](<https://android-review.googlesource.com/q/topic:%22soong_inclusive_language%22+(status:open%20OR%20status:merged)>), [curl](https://github.com/curl/curl/pull/5546), [PHPUnit](https://github.com/sebastianbergmann/phpunit/issues/4275), and many more.
+Other projects have already taken similar steps, for example [Go](https://go-review.googlesource.com/c/go/+/236857/), [Android](<https://android-review.googlesource.com/q/topic:%22soong_inclusive_language%22+(status:open%20OR%20status:merged)>), [curl](https://github.com/curl/curl/pull/5546) or [PHPUnit](https://github.com/sebastianbergmann/phpunit/issues/4275).
 
 These terms are used in Ember CLI for rather advanced functionality, which should make providing a better alternative naming and deprecating the current one rather easy.
 
@@ -55,7 +55,9 @@ let app = new EmberApp(defaults, {
 });
 ```
 
-The old keys (`blacklist` and `whitelist`) should be aliased to the new ones, and show a deprecation warning. Functionally, nothing changes.
+The old keys (`blacklist` and `whitelist`) should be aliased to the new ones, and show a deprecation warning. Functionally, they behave the same.
+
+If both keys are used at the same time (so either `blacklist` AND `exclude` or `whitelist` AND `include`), an error is thrown.
 
 ## How we teach this
 
@@ -72,8 +74,23 @@ There are also other possible terms that could be used to replace `blacklist` an
 - `allowlist` / `denylist`
 - `allowlist` / `disallowlist`
 - `allow` / `deny`
+- `allowed` / `disallowed`
 - `allowedlist` / `disallowedlist`
 - `safelist` / `blockedlist`
+
+For reference, this is how often these terms are used accross the Ember ecosystem according to Ember Observer - all searches done inside of index.js files for easier comparison:
+
+- blacklist: 16 addons
+- whitelist: 20 addons
+- allowlist, allowedlist, denylist, deniedlist, disallowlist, disallowedlist, blockedlist: 0 addons
+- blocklist: 2 addons
+- safelist: 1 addon
+- allow: 4 addons
+- deny: 1 addon
+- allowed: 18 addons
+- disallowed: 1 addon
+- include: 207 addons
+- exclude: 101 addons
 
 ## Unresolved questions
 


nightire:
I vote for `include/exclude`.

NullVoxPopuli:
I think it depends on the context which makes more sense.

Like, for example, in a firewall, allow/deny makes more sense

mydea:
I will collect some alternatives and put them in the RFC - although I haven't found anything I prefer over `include` and `exclude`. IMHO, include/exclude describe the functionality here pretty well and are easy to understand.

text/0000-replace-blacklist-whitelist.md
```diff
@@ -0,0 +1,70 @@
...
+
+# Replace terms blacklist & whitelist in Ember CLI
+
+## Summary
+
+Ember.js prides itself (rightly) on being an inclusive framework. To further improve on this, we should remove the terms "blacklist" and "whitelist" with more neutral replacements.
+
+## Motivation
+
+The terms "blacklist" and "whitelist" are considered racially insenstive. These terms are used in Ember CLI for rather advanced functionality, which should make providing a better alternative naming and deprecating the current one rather easy.
```
rwjblue, line 14: This seems to be making a specific claim (that the terms are racially motivated), can you cross link reference material to make it clear this isn't just an assumption being made.

text/0000-replace-blacklist-whitelist.md
```diff
@@ -0,0 +1,70 @@
...
+
+## How we teach this
+
+We should update the Ember CLI API docs with the new terms.
+
+## Drawbacks
+
+This might be considered "churn" by some. However, it is a small enough change to not really impact anybody meaningfully, while ensuring that Ember.js feels inclusive to everybody.
+
+## Alternatives
```
rwjblue, line 64: I think there are other alternatives that are worth mentioning (e.g. "allowlist" / "disallowlist"). Though I do personally think `include`/`exclude` is the better choice.

miguelcobain:
For completeness sake, I think it's important to point out that the origins of the blacklist/whitelist terms have nothing to do with race. More than that, it has never been historically used in any racial context.

Despite that, some people think that it's still valuable to get rid of them because some people might take offense, even when no offense was ever intended. I'm not debating that position here.

However, in that case, I think it's important to make sure that whatever the terms we end up replacing these with do not offend anyone else as a result. Just making sure that this RFC solves what it is trying to solve.

text/0000-replace-blacklist-whitelist.md
```diff
@@ -0,0 +1,70 @@
...
+
```
MelSumner, line 65:
```suggestion
Other suggestions for what we can use: 
- allowedlist/disallowedlist
- include/exclude
- safelist/blockedlist

```

mydea: I have added alternate terms!

text/0000-replace-blacklist-whitelist.md
mydea: :+1: added more alternatives

text/0000-replace-blacklist-whitelist.md
mydea: I have added some examples and references, hope that works better!

MelSumner: There is a lot of material that is available about the subliminal messaging that is pervasive in American culture whose underlying message is "white is good" and that has made its way into a lot of things as "the norm". 

While the original developer who first wrote "whitelist" and "blacklist" was perhaps not a racist (but rather using the words that had been made social norms by a white majority), we have enough public awareness to choose an elevated, more inclusive way of doing things for Ember projects.

text/0000-replace-blacklist-whitelist.md
```diff
@@ -0,0 +1,70 @@
...
+// After
+let app = new EmberApp(defaults, {
+  addons: {
+    include: ["ember-cli-sass"],
+  },
+});
+```
+
+The old keys (`blacklist` and `whitelist`) should be aliased to the new ones, and show a deprecation warning. Functionally, nothing changes.
```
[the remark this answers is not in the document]
MelSumner: There is a lot of material that is available about the subliminal messaging that is pervasive in American culture whose underlying message is "white is good" and that has made its way into a lot of things as "the norm".

While the original developer who first wrote "whitelist" and "blacklist" was perhaps not a racist (but rather using the words that had been made social norms by a white majority), we have enough public awareness to choose an elevated, more inclusive way of doing things for Ember projects.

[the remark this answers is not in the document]
mydea: I am not sure, as far as I see this only affects apps, not addons. And it will only ever appear exactly once at one place, and can be immediately changed without much thought.

I am also OK with not deprecating this now, but then, when would we? Will it be less churn if it it deprecated later? :shrug: But I also have no strong feelings otherwise. What timeline would you think about for deprecation?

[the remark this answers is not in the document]
Gaurav0: As long as this only affects apps (and addons not getting upgraded won't produce deprecations in consuming apps) I feel the noise is minimal and the cost of deprecating is the same as or less than the cost of maintaining both keys.

text/0000-replace-blacklist-whitelist.md
```diff
@@ -0,0 +1,80 @@
...
+
+We should update the Ember CLI API docs with the new terms.
+
+## Drawbacks
+
+This might be considered "churn" by some. However, it is a small enough change to not really impact anybody meaningfully, while ensuring that Ember.js feels inclusive to everybody.
+
+## Alternatives
+
+There are also other possible terms that could be used to replace `blacklist` and `whitelist`. For example:
```
simi, line 70: ℹ️ you can inspire at https://github.com/rails/rails/pull/33681/files as well. There's no need to use only one-worded terms.

MelSumner:
> For completeness sake, I think it's important to point out that the origins of the blacklist/whitelist terms have nothing to do with race. More than that, it has never been historically used in any racial context.

@miguelcobain There’s some reading and learning that I think is useful and necessary, that cover topics such as color connotations and racial attitudes (especially, but not limited to, America), covert racism, the ways that white supremacy is reproduced through casual discourse, internalized racism, unconscious bias, etc. It’s _really_ important to self-educate about these things. For better or for worse, words have meaning, and I don't think we should ignore them in our code repositories because _we_ didn’t mean for them to be racist. 

In our society at large, and especially in tech, it’s been easy for us to codify the biases we didn’t even know we had been conditioned to have. But _now that we have been very clearly and persistently been made aware_, the onus is on us to self-educate and take action.

There’s a clear path forward for us as a community, and IMO it’s to be actively anti-racist. With both kindness and firmness, the “well actually it’s not racist” type of reasoning isn’t going to be helpful in this conversation, but rather serve as a magnet for less productive discussions.

miguelcobain:
@MelSumner I disagree with many of your points, but since they don't have any impact on the RFC, I don't think we should debate them here.

In my comment I was just pointing out that if we're going to remove words because they might offend people, then there are a lot of words that we want to avoid.
So we have to make sure that we don't pick words at random, taking the risk of offending someone else. If this RFC is accepted, it sets a precedent. We need to be prepared other terminology in the future for coherence sake.
I think we can all agree on this.

text/0000-replace-blacklist-whitelist.md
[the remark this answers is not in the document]
rwjblue: @MelSumner 

> There is a lot of material that is available about the subliminal messaging that is pervasive in American culture whose underlying message is "white is good" and that has made its way into a lot of things as "the norm".
> 
> While the original developer who first wrote "whitelist" and "blacklist" was perhaps not a racist (but rather using the words that had been made social norms by a white majority), we have enough public awareness to choose an elevated, more inclusive way of doing things for Ember projects.

How does this have to do with what I said at all? We all really need to make sure to stay focused on the actual proposal here...

@mydea

> I am not sure, as far as I see this only affects apps, not addons. And it will only ever appear exactly once at one place, and can be immediately changed without much thought.

As I mentioned above, we should not introduce churn on existing application developers for this. If they want to migrate off of the previous terms: seems completely fine and we will provide a nice alternative (this RFC!). 

@Gaurav0 

> I feel the noise is minimal and the cost of deprecating is the same as or less than the cost of maintaining both keys.

The cost of deprecation is definitely more overall. It requires:

* adding support for both fields
* adding the deprecation notice itself
* requires all applications that use these APIs to modify their application code

Whereas supporting both fields has only the first. Now, in general, ember-cli developers (and addon authors) are commonly willing to absorb churn like the one proposed in this RFC in order to make the lives of our users better (hence the willingness to support a version of this RFC), but instead of reducing churn for our consumers by doing the right thing internally adding a deprecation _intentionally_ adds **exactly** the churn we intend to avoid.

text/0000-replace-blacklist-whitelist.md
rwjblue: > There is a lot of material that is available about the subliminal messaging that is pervasive in American culture whose underlying message is "white is good" and that has made its way into a lot of things as "the norm".

Awesome, that was my assumption when I made my comment above. It should be super easy to satisfy my request. 😃

rwjblue: > While the original developer who first wrote "whitelist" and "blacklist" was perhaps not a racist (but rather using the words that had been made social norms by a white majority), we have enough public awareness to choose an elevated, more inclusive way of doing things for Ember projects.

Not sure what this was replying to at all.

text/0000-replace-blacklist-whitelist.md
```diff
@@ -0,0 +1,80 @@
...
+
+## Summary
+
+Ember.js prides itself (rightly) on being an inclusive framework. To further improve on this, we should remove the terms "blacklist" and "whitelist" with more neutral replacements.
+
+## Motivation
+
+The terms "blacklist" and "whitelist" can be considered racially insenstive. While the origin of these terms in this context is not in itself racially motivated, the fact that in todays context it _can be considered racially insensitive_ has been discussed frequently - see for example [here](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC6148600/), [here](https://bugs.chromium.org/p/chromium/issues/detail?id=981129#c16) or [here](https://www.zdnet.com/article/uk-ncsc-to-stop-using-whitelist-and-blacklist-due-to-racial-stereotyping/).
+
+Other projects have already taken similar steps, for example [Go](https://go-review.googlesource.com/c/go/+/236857/), [Android](<https://android-review.googlesource.com/q/topic:%22soong_inclusive_language%22+(status:open%20OR%20status:merged)>), [curl](https://github.com/curl/curl/pull/5546), [PHPUnit](https://github.com/sebastianbergmann/phpunit/issues/4275), and many more.
```
rwjblue, line 16: Should probably remove this bit (or find some meta link to cross reference it with):

```suggestion
Other projects have already taken similar steps, for example [Go](https://go-review.googlesource.com/c/go/+/236857/), [Android](<https://android-review.googlesource.com/q/topic:%22soong_inclusive_language%22+(status:open%20OR%20status:merged)>), [curl](https://github.com/curl/curl/pull/5546), [PHPUnit](https://github.com/sebastianbergmann/phpunit/issues/4275).
```

text/0000-replace-blacklist-whitelist.md
```diff
@@ -0,0 +1,80 @@
...
+
+These terms are used in Ember CLI for rather advanced functionality, which should make providing a better alternative naming and deprecating the current one rather easy.
+
+## Detailed design
+
+You can use `blacklist` and `whitelist` to include/exclude certain addons from your build:
+
+https://ember-cli.com/user-guide/#whitelisting-and-blacklisting-assets
+
+We should replace these with `exclude` or `include`:
```
stefanpenner, line 26: include is, unfortunately, ambiguous, for example: "please include an Oreo MilkShake with my order." It does not limit my order to an Oreo MilkShake; rather, it implies my order will have at-least one Oreo MilkShake.

The existing whitelist implementation does not guarantee the existence of any of these mentioned addons; rather, the whitelist limits addons to the ones mentioned in the provided configuration. It is also worth noting; the existing implementation does also not require any of the "whitelisted" addons to be present.

Given that, I would recommend a more concise name, and the best I could come up with is `allowed,` which pairs nicely with `disallowed.`

text/0000-replace-blacklist-whitelist.md
```diff
@@ -0,0 +1,80 @@
...
+
+// After
+let app = new EmberApp(defaults, {
+  addons: {
+    include: ["ember-cli-sass"],
+  },
+});
+```
+
+The old keys (`blacklist` and `whitelist`) should be aliased to the new ones, and show a deprecation warning. Functionally, nothing changes.
```
stefanpenner, line 58: I believe this needs a little more fleshing out as nothing prevents the user, or a bad merge conflict, from resulting in both naming schemes being used. 

Given that, how will such conflicts be handled / limited / controlled for?

text/0000-replace-blacklist-whitelist.md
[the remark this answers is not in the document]
stefanpenner: Ember has an existing policy for breaking changes, following that process seems like a tried tested and true approach.

text/0000-replace-blacklist-whitelist.md
stefanpenner, line 58: Does this change also include updating blueprints?

text/0000-replace-blacklist-whitelist.md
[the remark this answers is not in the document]
mehulkar: I know y'all have good intentions, so just want to say... IMO it's not a requirement to make this transition easy or comfortable or following the appropriate schedule (whatever that is), if it's happening for the right reasons. I honestly can't see a good reason _not_ to immediately deprecate.

hergaiety:
@miguelcobain I agree that if this RFC is accepted it will set a precedent. It'll be good to have this example to reference if and when we find more instances to replace in the future. I'm okay with relatively small wins that push us in the right direction and believe the small change signifies something more for us as a community. It isn't that unlike implementing a project's first A11y improvement that will begin to say "A11y is important to us and we should continue to address it".

text/0000-replace-blacklist-whitelist.md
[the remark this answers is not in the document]
mydea: If this RFC is accepted I think it is out of question that it will _eventually_ be deprecated (as we would not want to indefinitely keep two option keys for the same thing). So as far as I see, the actual question is _when_ should it be deprecated.

As this is not a gradual transition, but a one-time change always done in the same place, at most once per application, I do not see the benefit in postponing this. Thinking about churn, there are these possible scenarios:

**Case A: We ship the change but deprecate it later**
**Option 1**: I read about the change in the release docs, so I go into my app to check if I've been using this, and if so, change it.
**Option 2**: I either don't realize this has been changed, or I do not want to change it right now. So I do nothing initially. X months later it is deprecated, so I do Option 1 then.

**Case B: We ship the change and deprecate it immediately**
I get notified if I've been using this and change it in my app.

I don't see anybody using a deprecation-free transition time of X months to slowly change this, as this cannot be slowly changed - either you update it in your app, or you don't.

Overall, I am not necessarily against deprecating it later, I simply don't see how it would help anybody. But I'm also happy to include a different deprecation strategy if you would prefer. What time horizon would make you feel more comfortable, deprecation wise?

text/0000-replace-blacklist-whitelist.md
mydea: As far as I know, this is not part of any blueprint. As far as I see it is really only a more "advanced" feature that can be found rather buried in the Ember CLI docs.

text/0000-replace-blacklist-whitelist.md
mydea: I don't really agree on this. I went to Ember Observer to check how often certain terms are used in the community. For easier comparison, I restricted searches to index.js - these were the results:

- blacklist: 16 addons
- whitelist: 20 addons
- allowlist, allowedlist, denylist, deniedlist, disallowlist, disallowedlist, blockedlist: 0 addons
- blocklist: 2 addons
- safelist: 1 addon
- allow: 4 addons
- deny: 1 addon
- allowed: 18 addons
- disallowed: 1 addon
- include: 207 addons
- exclude: 101 addons

Also, looking at some popular usages like broccoli-funnel, the general mode of function is pretty much the same. E.g. I can specify:

```js
funnel('src', {
  include: ['my-file', 'my-non-existing-file']
});
```

or

```js
funnel('src', {
  exclude: ['my-file']
});
```

Yes, I know that takes a glob pattern, but the mental model is the same for me.

So for me, it would feel quite natural to use the same wording for configuration of the addons.
I don't know of any occurrence in an Ember app where I used `include` in a way you described (meaning it adds this to a selection/list at least once), but I've used it frequently in the way broccoli-funnel uses it.

text/0000-replace-blacklist-whitelist.md
mydea: I added a sentence about this - I would suggest throwing an error in this case.

text/0000-replace-blacklist-whitelist.md
mydea: done!

text/0000-replace-blacklist-whitelist.md
stefanpenner: As a maintainer of broccoli-funnel I apologies for misleading you and can tell you with confidence, if it where to be written again it would not use `include`. Rather it would mimic `walkSync` `globs` + `ignore` verbiage.

text/0000-replace-blacklist-whitelist.md
stefanpenner: @mydea for users who run `ember init` or similar upgrade tools, a code-mod could be written to make this migration pretty seamless/automagical.

text/0000-replace-blacklist-whitelist.md
```diff
@@ -55,7 +55,9 @@ let app = new EmberApp(defaults, {
 });
 ```
 
-The old keys (`blacklist` and `whitelist`) should be aliased to the new ones, and show a deprecation warning. Functionally, nothing changes.
+The old keys (`blacklist` and `whitelist`) should be aliased to the new ones, and show a deprecation warning. Functionally, they behave the same.
+
+If both keys are used at the same time (so either `blacklist` AND `exclude` or `whitelist` AND `include`), an error is thrown.
```
stefanpenner, line 60: 👍

text/0000-replace-blacklist-whitelist.md
```diff
@@ -72,9 +74,24 @@ There are also other possible terms that could be used to replace `blacklist` an
 - `allowlist` / `denylist`
 - `allowlist` / `disallowlist`
 - `allow` / `deny`
+- `allowed` / `disallowed`
 - `allowedlist` / `disallowedlist`
 - `safelist` / `blockedlist`
 
+For reference, this is how often these terms are used accross the Ember ecosystem according to Ember Observer - all searches done inside of index.js files for easier comparison:
```
stefanpenner, line 81: The total usage isn't super useful, without knowing the context it's hard to tell if the usage and constraints are the same or if the author regrets the choice. Simply cargo culting verbiage would be be unfortunate.

Given the constraints variants of `allow.*` `disallow.*` appear to be the most clear.

text/0000-replace-blacklist-whitelist.md
wycats: `disallowed` seems like a good solution. I agree that `include` / `exclude` can be ambiguous, and have seen this problem crop up repeatedly in various OSS projects I've worked on.

There's another ambiguity than the one you referenced: what happens if you use both of the features together. I tend to not think that it's important to have a pair that sounds symmetrical, because it intrinsically opens up this question.

Spitballing, I prefer a mutually exclusive pair like `only` and `except`. The word "only" strongly implies that the other files aren't included. The word "except" strongly implies that the other files are included. The words "include" and "exclude" are much more neutral, which gives rise to the "what happens if both are used" confusion.

wycats: (btw, @mydea the fact that an idiom is popular doesn't mean it's not confusing or misleading -- in fact, that's the whole premise of this PR!)

text/0000-replace-blacklist-whitelist.md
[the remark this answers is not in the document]
wycats: As a general principle, deprecation RFCs don't try to guess the exact timing of when the deprecations will be released. Instead, they describe a phasing strategy for the deprecations, sometimes along with criteria that should be met for each stage.

In other words, deprecation RFCs are about the details of issuing deprecations, not about the exact versions when the deprecations will appear in the codebase.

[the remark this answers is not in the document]
rwjblue: > In other words, deprecation RFCs are about the details of issuing deprecations, not about the exact versions when the deprecations will appear in the codebase.

Yes, exactly. My point was that I am opposed to immediate deprecation. It should be a phased transition plan (introduce new API, deprecate in a future version, make an error upon deprecation removal), like we normally do.

wycats:
> For completeness sake, I think it's important to point out that the origins of the blacklist/whitelist terms have nothing to do with race. More than that, it has never been historically used in any racial context.

There are lots of words that are used in a clearly hurtful way that don't have hurtful etymological roots. For example, every time I hear someone say "The Final Solution to the X Problem", my brain immediately goes back to the Holocaust, even though the etymology of none of those words is anti-Semitic.

Etymology can occasionally inform us about the original motivation of certain terms, but in my opinion, the more appropriate question is whether people feel hurt by the term today. Since it's not always easy to tell when you're hurting people in an online environment, we should err on the side of caution, especially when it comes to terminology that other similar communities have come to believe is hurtful.

> Despite that, some people think that it's still valuable to get rid of them because some people might take offense, even when no offense was ever intended. I'm not debating that position here.

It's possible to hurt someone even without intending to hurt them. When we learn that something we're doing may be hurting somebody, our default reaction should be to stop engaging in the behavior.

The accessibility community gives us many examples of this: there are a huge number of terms that people use without intending to be hurtful that create tremendous amounts of pain in the affected communities. When people say that something we're doing is hurtful, my view is that we should stop doing it first and ask questions later (if ever).

I don't think that good intent absolves hurtful behavior from the hurt that it causes.

> However, in that case, I think it's important to make sure that whatever the terms we end up replacing these with do not offend anyone else as a result. Just making sure that this RFC solves what it is trying to solve.

I hope that community members who feel that any of the proposed terms are hurtful to them will raise those concerns as we discuss alternatives, and that the good faith effort to make this change will demonstrate to folks that feedback along those lines will be taken seriously.

kategengler:
I completely agree we should not be using the terms `whitelist` and `blacklist`.

I think it is possible that this API should be outright deprecated for removal, rather than only renamed. I've personally never encountered this API in use and understand there are other ways to prevent specific addons from being included in an app.

rwjblue:
> I think it is possible that this API should be outright deprecated for removal, rather than only renamed. I've personally never encountered this API in use and understand there are other ways to prevent specific addons from being included in an app.

Ya, this actually seems like a much better solution to me. I'm not opposed to renaming the options in the abstract, but this API just doesn't seem to need to exist at all.

text/0000-replace-blacklist-whitelist.md
stefanpenner: My only concern with `only` is that the list of addons can include add-ons which are not present. This lead me to `allowed`/ `disallow` but, that is a minor detail and could just be how I interpret `only`

stefanpenner:
@rwjblue there is actual value in this api, it isn't common but it has come up. I believe with some rigor elsewhere it could be made to go away, but that is likely more involved then the RFC in question is interested in exploring.

lukemelia:
> I think it is possible that this API should be outright deprecated for removal, rather than only renamed. I've personally never encountered this API in use and understand there are other ways to prevent specific addons from being included in an app.

ember-freestyle recommends using this API conditionally based on environment to avoid shipping freestyle code to production.

text/0000-replace-blacklist-whitelist.md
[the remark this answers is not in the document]
mydea: OK, I am fine with that - so I will remove the mention of the deprecation.

To cap this off, my thoughts on this: IMHO this change is a bit different from most other changes. Usually we would introduce a new API with changed behavior, and give devs time to migrate to the new behavior before deprecating it. However, in this case it is a one-to-one replacement of a rather niche functionality that is used at most once per application. That was my reasoning for proposing to deprecate it immediately, instead of requiring another RFC sometime later to deprecate it. IMHO it will potentially even introduce _more_ churn as new users might still be using the old terms (and not even know there are new ones), e.g. by copy-pasting code from somewhere or similar. But I respect your desire to follow the usual process :+1:

text/0000-replace-blacklist-whitelist.md
mydea: I am also OK with using a different term - I chose those to be consistent with current configuration options used in the Ember community, and because they made sense to me. 

I would prefer `only`/`except` over `allowed`/`disallow`, also because the latter can be tricky to get right (`allowed` vs. `allow` vs. `disallow` vs. `disallowed`).

Another option could be `only`/`ignore`, maybe.

mydea:
My use case (and the reason I stumbled over this API in the first place) is the same as for ember-freestyle, I have some in-repo addons with a bunch of code that should be completely ignored for production builds (including styleguides and some mocking stuff). 

I don't know about modifying `app.addons` (have neither encountered it anywhere so far, nor found any documentation on it), but it doesn't necessarily sound like a straightforward/better replacement to me. 

IMHO in this case, it makes sense to replace the terms right now and maybe think about removing it completely (or replacing it with a different functionality) later.

lenoraporter:
Making others feel comfortable isn't always easy and it definitely won't be convenient all the time. In any relationship, it is about compromise. Building software for **ALL** is also about compromise. It's a privilege to not think twice about Master/Slave or Whitelist/Blacklist. I wish I can do that. Although, I think there are many more important issues to tackle, I wholeheartedly believe in the journey of small wins. In the book Atomic Habits, James Clear talks about how small wins over time turn into habits and habits eventually create identity.

I understand that this change comes from a good place so I have no push back. Ember is an open-source community, so from my stand-point, we have to be realistic on what we can do as an open-source community to alleviate injustice. Right now, we all own the code and that's where we hold the most influence. If we can get to a place where new comers or old comers feel welcomed (Even if it is as "small" as a terminology change), I feel that's a win. Gradually, it'll become our identity and every choice going forward will consider **ALL** from the beginning. 

This reminds me of when I read Ember's contribution docs for the first time. It said:

> Write in complete sentences. Use a friendly tone of voice. Use American English. Use gender-neutral pronouns.

You've all set the standard for newcomers to come into this space with intention. Even Tomster has a few gender-neutral characters. It is in the fiber of Ember now without even thinking. That's what a small win does. I feel like we've been here before and I trust you all will make the right decision. Side Note: Hopefully, we can also think about diversifying the pool of decision makers so we can catch more issues as a team but that may be a small win for another time.

To put in my two cents, I like `AllowList` / `DenyList`. I can see `Accept` / `Ignore` as well.

sandstrom:
What do you guys think about the word *server*?

Ember is primarily a client-side framework which is good, but we do use e.g. `ember serve`. Perhaps we could use a less loaded word, such as "processor" or "assistant".

> Server comes from the Latin word servire "be a servant, be enslaved;", related to servus "slave," which is of uncertain origin.

miguelcobain:
> What do you people think about the word server?
> 
> Ember is primarily a client-side framework which is good, but we do use e.g. ember serve. Perhaps we could use a less loaded word, such as "processor" or "assistant".
> 
> Server comes from the Latin word servire "be a servant, be enslaved;", related to servus "slave," which is of uncertain origin.

This is what I meant when I said above that this RFC sets a precedent if accepted.
*If* this RFC is accepted, then I think removing the word "server" as @sandstrom suggested (and potentially many others) is the only coherent action to take.

We need to be prepared for the possibility to remove **a lot** of words.

sandstrom:
@miguelcobain I agree!

I made my previous comment to make a point. '*Server > servire > slave*' is fairly similar to '*Master key > master > slave*'. If one is offensive its hard not to argue that the other one is. Another common computer with similar origin is *bot/robot*, from *robota* which means *slave worker*.

Should we remove the word "server"? Stop using "dependabot"? 

Any terminology can be offensive with the "right" framing. Changing them may seem like an easy quick fix, but we're also going down the slippery slope of accepting the idea that certain words are, in themselves, problematic to the point where they cannot be uttered.

Also, different groups may feel offended by different things. Giving everyone the right to never be offended could make it impossible to say anything in the long run. Is that really where we want to go? Obviously this change won't put us on the path to totalitarianism, but I do think the question is valid and should be asked.

Who's going judge what constitutes an offensive word, and what groups should be considered.

For example, in Russia, it's offensive to [depict, talk about & illustrate gay relationships](https://en.wikipedia.org/wiki/Russian_gay_propaganda_law#Prosecutions_and_other_effects).

There was recently an investigation into Apple's inclusion of gay emoji in iOS 13. Apple didn't end up getting convicted, but the fact that the case was brought to a Russian court illustrates my point.

If Russian developers would be offended by a Tomster with a pride T-shirt, would you take that down?

***

image


https://www.amnesty.org.nz/russian-authorities-ban-putin-%E2%80%98gay-clown%E2%80%99-meme-fail-investigate-homophobic-killings

text/0000-replace-blacklist-whitelist.md
```diff
@@ -0,0 +1,97 @@
...
+
+Other projects have already taken similar steps, for example [Go](https://go-review.googlesource.com/c/go/+/236857/), [Android](<https://android-review.googlesource.com/q/topic:%22soong_inclusive_language%22+(status:open%20OR%20status:merged)>), [curl](https://github.com/curl/curl/pull/5546) or [PHPUnit](https://github.com/sebastianbergmann/phpunit/issues/4275).
+
+These terms are used in Ember CLI for rather advanced functionality, which should make providing a better alternative naming and deprecating the current one rather easy.
+
+## Detailed design
+
+You can use `blacklist` and `whitelist` to include/exclude certain addons from your build:
+
+https://ember-cli.com/user-guide/#whitelisting-and-blacklisting-assets
```
MelSumner, line 24: Please check to see if this same information is on the official ember-cli guides- cli.emberjs.com, since we're in the process of moving the documentation there. + @locks @jenweber

text/0000-replace-blacklist-whitelist.md
MelSumner: @wycats after discussion on Friday, do we want to go with `allowlist` and `disallowlist` ?

text/0000-replace-blacklist-whitelist.md
```diff
@@ -0,0 +1,97 @@
...
+
+// After
+let app = new EmberApp(defaults, {
+  addons: {
+    include: ["ember-cli-sass"],
+  },
+});
+```
+
+The old keys (`blacklist` and `whitelist`) should be aliased to the new ones. Functionally, they behave the same.
```
MelSumner, line 58: Please include a deprecation path for the terms- or specify that another RFC should be submitted for their deprecation and removal.

text/0000-replace-blacklist-whitelist.md
mydea: As far as I can see, there is no reference in the new CLI guides (https://github.com/ember-learn/cli-guides/search?q=blacklist&unscoped_q=blacklist).

There is a reference in the API, here: https://ember-cli.com/api/classes/EmberApp.html:

> * addons, defaults to { blacklist: [], whitelist: [] }

text/0000-replace-blacklist-whitelist.md
mydea: I added a note on deprecating this later (based on earlier feedback), hope that works!

lenoraporter:
@sandstrom This is a healthy conversation and all questions are valid! I'm sure no one will give you "heat" for voicing your opinions. I'm sure everyone is coming into this conversation with good intentions so I love your question.

In my opinion, this plays into the common feature development problem: _Business Need_ Versus _User Need_. In our case, the _Open Source Project Need_ Versus _User Need_ Versus _The World's Need_. So, who are our users? Are they concerned?

We clearly cater to a certain group of people and that seems to be English-speakers. If you look at our [Who's Using Ember.js page](https://emberjs.com/ember-users/), you would see mostly US-based companies. Most of our contributors are also US-based. We do have a few users that are not US-based. They are from places like the Netherlands, Germany, Spain, Sweden, London, Canada, Australia, and etc. So far, I've listed countries with a progressive stance on lots of issues. To be fair, they still need work especially around aboriginal-societies, structural racism, discrimination, immigration policies and etc. but that's not why we are here. We are specially talking about our choice in potentially offending someone by the use of certain technical terms. This Master/Slave conversation has been happening for a very long time. In the countries I've listed above, this context is extremely relevant. Why? Spain abolished slavery in 1811; however, it continued in the Caribbean Islands until 1873. The British Empire abolished slavery in 1833 (The debt to buy the slaves their freedom wasn't paid off until 2015 so **living British citizens** had to help to end the slave trade). Sweden banned slavery in 1847. The Netherlands abolished slavery in Dutch Caribbean colonies in 1861. US banned slavery in 1865. So, that means our users have a great context and awareness when it comes to the Master/Slave terminology. This user base has voiced their concerns around this issue! I've seen articles dating back to 2004 about the terms. Fun Fact: Harvard changed the title of its Residential Administrators from  ‘House Masters’ to ‘Faculty Deans’ in 2016. In 2003, Los Angeles officials asked their suppliers and contractors to stop using master/slave on computer equipment.

I'm saying all this to say, your question does not have a clear cut answer, however, it does have an answer when our user's/contributor's feedback comes into play. Like I've said before, Master/Slave and Whitelist/Blacklist has been a discussion for years now. This isn't new feedback for the tech industry. We can't predict every word that may end up being offensive; however, we can totally open our ears to actually HEAR and LISTEN to what is happening. This big influx is only happening because people are listening to voices they've never heard before (FINALLY) but these conversations were already happening.

I personally don't have an issue with `Ember Serve`. From my area in the world, I haven't heard my circles talk about the word serve. Serve isn't being used in a negative or derogatory way. Serve also doesn't give me the sense of captivity for more than 400 years, dehumanization, brutal treatment, or taking a group of people from their native countries and forcing them to perform free labor. Blacklist/Whitelist has lots of negative associations. Master/Slave is discussing a hierarchy that highlights a really dark past with majority of Ember's user group. Based on my quick analysis of our users (which may not be fully accurate), I see tons of US based companies/contributors. The US as a whole is rethinking race and actually trying to discuss it as a country. There is a discussion being had and old topics are being re-surfaced. The Whitelist/Blacklist Master/Slave topic is one of them. These groups have been very clear and they've specifically pointed at words that are offensive. No one told anyone to change the entire dictionary. I think the conversations have been pretty clear.

If we really want to know how our actual users feel, we could always open a feedback loop for these discussions.

Side note: Apple is an American multinational brand with its headquarters in the US. I hope they considered their Russian users along with the bulk of their user base and made the decision to keep the emoji. At some point, you just have to make a decision and stand firm behind it. I'm glad it went to court so it can be settled for good. That goes the same for us. We have to stand firm with the majority of our user base, the climate of the world, and make a decision. As time goes, we will make mistakes; however, if we leave one ear open to listen to our users concerns, we can make decisions from there. Small wins.

wifelette:
First, thanks Lenora, for doing the emotional labor and for your willingness to educate us, and lastly, for assuming good intent. We're a better community for your perspective, and appreciate you speaking up!

Moving on. Yes, the idea of a "slippery slope" is a thing. Yes, I/we can't provide KO-arguments to anyone who wants to say "but something different could offend everyone!" or "this isn't a racially charged term in my culture". But while not patently false, those are irrelevant ideas.

Right now, we have a use case where there's a thing we _know_ offends some people. Therefore we're going to do our best to fix it, as effectively and efficiently as we can. Bottom line.

Yes, we should in good faith try to make sure our _new_ solution doesn't offend anyone. And I think that happened above, seemingly successfully. And frankly, if we choose something that in three years we learn _is_ a problem, you'd better believe we'll have to change it all over again :p Humanity is constantly getting better, more aware, and hopefully as proven by this thread, more introspective.

One part of me wants to do the thing lots of other leaders have done: make it clear that this isn't a place for continued philosophical debate, and close the thread for comments. Typically though, that hasn't been the Ember way. So let me state in no vague terms that **the Ember project and leadership stands behind making this change**. We stand behind the BLM movement, the dire need for social change, behind our Black community members, and behind the need to just do the work (even if it's annoying). This is some of that work, and once it's done, we'll all be better for it.

For those of you who have gleaned some education from this thread, and for those of you who have generously provided it, I encourage you to continue those conversations off-thread. Discord, DMs, whatever works best. 

I'm asking that further comments be restricted to any additional implementation details, tooling, etc. The _premise_ however, is decided <3

sandstrom:
@wifelette I appreciate that you aren't locking the thread (the cheapest way to suppress contrarian ideas).

Maybe it's because I'm Swedish, but I don't understand this language purge. As a European I've always been troubled by US police violence, inequality of opportunity, gun proliferation, etc. But I don't see how you're making this any better by abolishing out-of-context words (I think you're making it worse). Electing a non-moron president and a less corrupt congress would seem like a more fruitful pursuit.

Even though I strongly disagree with this line of thinking I still like Ember a lot, and in general I think you and everyone else on the leadership team are doing an awesome job! 🐹

miguelcobain:
> So let me state in no vague terms that the Ember project and leadership stands behind making this change. We stand behind the BLM movement, the dire need for social change, (...)

I just wanted to point out that the concerns that me and @sandstrom raised are orthogonal to wether or not one supports the BLM movement, unlike @wifelette implied. One can support the BLM movement and still disagree with this RFC.

I really don't think this is a good path forward (on the contrary).
If this is accepted, I expect other subsequent *legitimate* RFCs to remove other words from the project. I was just preparing everyone for that possibility, just in case you didn't already know.

With that being said, this decision will have 0 impact in what i think about the awesomeness Ember and its community of talented people. I don't think I have anything else to add.
Hearing disagreeing opinions is sometimes hard, so thank you all for doing so in a respectful manner. That in itself speaks greatly about our community.

g13013:
With all due respect, I find this stupid !

patsy-issa:
@g13013 The conversation has been constructive until now please try to do the same, I would have expected more understanding from someone that comes from a country that used to be a french colony.

To add to the topic as I feel everyone's intentions have been to understand, digest the recent events within the US and try to do something, anything!

Racism and tragedies such as the murder of George Floyd are by no means isolated or new to our world, from the systematic racism that has made equal opportunity a myth across the globe, to the class struggle that has isolated communities regardless of colour.

Will changing the terminology fix racism? No, but it is a start.

I would like to reach out to anyone interested in making a difference and working towards a more conscious and inclusive world, while the outrage is global if anything is to change it will have to start local, hold your representatives accountable!

It is time for sustainable, conscious policies, be it left or right the end result has been the same, one need not look far for examples:

If you live in germany hold your representatives accountable: [german tanks used by turkey](https://www.bbc.com/news/world-europe-42795943)
If you live in the united states you, [ensure that the right to vote](https://www.democracynow.org/2020/6/22/headlines/kentucky_slashes_95_of_polling_places_ahead_of_primary_election)  is accessible in your state.
In my home country and across the arab world [there is so much we can do](https://en.wikipedia.org/wiki/Kafala_system), nay must do!

From my interactions with programming communities we are prone to detachment from the world, I remember a time when the change of the database terminology `master/slave`  was met with opposition.

The same arguments have been used `when I use it I do not mean it in x way`. This is not about you!

The point I am trying to make, the fight is not about landing the deprecation prematurely, we have guidelines and a process for introducing changes and this energy of pushing for immediate action, while remarkable, is misplaced.

Forgive me for carrying on the conversation in the thread and generating more noise, if anyone is interested in forming an actionable plan for ethics within code please do reach out through discord.

Thank you and be kind to one another.

simi:
> The conversation has been constructive until now please try to do the same, I would have expected more understanding from someone that comes from a country that used to be a french colony.

Any reason to make it personal?

wifelette:
I have hidden a couple responses that were _clearly_ not, as requested:

> further comments be restricted to any additional implementation details, tooling, etc. The premise however, is decided <3

Whether your thoughts affirm or disagree with the decision here, agree or disagree with the premise, etc, **please do not use this forum to express them further**. Feel free to reach out on other platforms to the folks here who earlier in the thread may have explicitly volunteered themselves, or to find a willing conversation partner on Discord or elsewhere. 

Please don't push me to _actually_ lock the thread by ignoring this request <3

g13013:
@patsy-issa so according to this logic, if hopefully  the enslavement is terminated on earth would you ask to remove the words "master/slave" from the dictionary ? I personally distinguish between a black respectable human being and a blacklist, it would be silly otherwise! anyways, black lives mattered, matter will matter to me, and changing blacklist to something else will change nothing but add more sensitivity and bring the debate to where it does not belong. 

And yes as my ancestors suffered from war crimes, I can say that I have some legitimity to say that I would never ask to remove the word "colony" from a source code (if it ever existed) as long as it doesn't directly designate a particular race, country or people or make a clear direct apology for colonialisme.

g13013:
@wifelette can you explain why is it offtopic to comment  to express an opinion on "Replace terms blacklist & whitelist" RFC "request for comment"

hergaiety:
@g13013 all `off-topic` comments are still available to be toggled to be seen if desired and by having them set as such will help focus further discussions toward refining and progressing this RFC. There's a desire to keep any RFC from falling into personal opinion limbo and it has been decided this is no longer the forum to have these debates.

Twice now there has been a request to limit the following discussion. Let's continue focused discussion so as to not require a locking of this thread.

g13013:
the problem is not on the subject now, the problem is on how this issue is being handled. anyways this is disappointing...

text/0000-replace-blacklist-whitelist.md
joukevandermaas: Maybe I'm late to the party, but how about `includeOnly` and `exclude`?

stefanpenner:
I'm surprised to see this having gone stale, it appeared to have many passionate contributors early. 

I would like to recommend, or empower someone to take this to completion as it would appear the general consensus is supportive of a name change.

MelSumner:
@stefanpenner it's on the agenda for Friday's core team meeting, there was a lull because I'm championing it and I had taken some vacation time. Thanks for checking in though!

MelSumner:
After discussion during today's framework core team meeting, we've decided to move this RFC into FCP. Thank you @mydea for doing this work. 👍

text/0000-replace-blacklist-whitelist.md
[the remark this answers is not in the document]
les2: seems like its deprecated as soon as the new fields are available. whether you warn people about the deprecation is another thing.

i'm using deprecation in the practical sense of "this field will go away in favor of this other thing".

$0.02

text/0000-replace-blacklist-whitelist.md
les2: RE `include` vs `whitelist`

A whitelist means "only these things" while "include" doesn't necessarily imply that the list is the only list of things added. you might be "including" to something else.

`allowlist` is the clearest alternative i've come across.

$0.02

mydea:
I have mostly been waiting for a consensus on the naming to appear, since I don't really care that much one way or another which one we end up choosing! It seems like `allowlist` and `disallowlist` are the front runners now, I guess.

webark:
The other general alternatives I have seen mentioned are `blocklist` (maybe too close to black) and `safelist` thought allow and disallow are fine. I think we should also come to a realization that (well, that I hope) we as a larger computing community, will standardize around some set words and terms (master/slave as well) and it's good to be on the forefront of this change, it will probably change one more time as standards evolve. And that's good, and ok.

webark:
one disadvantage to using allow and disallow is that when searching for "allowlist" you will also get results from disallowlist as well.

rwjblue:
We discussed this again in todays meeting, and think that it is ready to land. Thanks everyone for working so hard to stay focused and on topic!

rwjblue:
> one disadvantage to using allow and disallow is that when searching for "allowlist" you will also get results from disallowlist as well.

Ya, definitely true. We did consider this before deciding to move forward in todays meeting. It is absolutely an issue for using `disallowList`/`allowList` as a general purpose replacement for `blacklist`/`whitelist`, but ultimately this specific API is very rarely used so we didn't consider the lack of searchability for this API to be a major concern.

PR #645, opened 2020-07-03
Add ember-page-title to app default blueprint

KamiKillertO:
[Rendered](https://github.com/KamiKillertO/rfcs/blob/ember-page-title-rfc/text/0000-add-ember-page-title-addon.md)

### DIFF
--- a/text/0000-add-ember-page-title-addon.md
+++ b/text/0000-add-ember-page-title-addon.md
@@ -22,7 +22,7 @@ Adding this addon will provide developers a simple solution to achieve the [WCAG
 1. Make [`ember-page-title`](https://github.com/adopted-ember-addons/ember-page-title) an official Ember addon by transferring `ember-page-title` repo to the [Ember CLI org](https://github.com/ember-cli) (it's currently in the [Adopted Ember Addons org](https://github.com/adopted-ember-addons))
 2. Add the dependency to the app blueprint here: https://github.com/ember-cli/ember-cli/blob/master/blueprints/app/files/package.json#L19
 3. Update app blueprint to include `<HeadLayout />` at the top of [application.hbs](https://github.com/ember-cli/ember-cli/blob/master/blueprints/app/files/app/templates/application.hbs)
-4. Update route blueprint to include `{{page-title RouteName}}` at the top of the route template file. `RouteName` is the route name provided to the command `ember generate route`
+4. Update route blueprint to include `{{page-title RouteName}}` at the top of the route [template.hbs](https://github.com/emberjs/ember.js/blob/master/blueprints/route/native-files/__root__/__templatepath__/__templatename__.hbs) (this [template.hbs](https://github.com/emberjs/ember.js/blob/master/blueprints/route/files/__root__/__templatepath__/__templatename__.hbs) too) where `RouteName` is the name of the route provided to the `ember generate route` command
 
 ## How we teach this
 
--- a/text/0000-add-ember-page-title-addon.md
+++ b/text/0000-add-ember-page-title-addon.md
@@ -21,7 +21,7 @@ Adding this addon will provide developers a simple solution to achieve the [WCAG
 
 1. Move ember-page-title to the ember-cli org (?) (already in ember-adopted-addon org)
 2. Add the dependency to the app blueprint here: https://github.com/ember-cli/ember-cli/blob/master/blueprints/app/files/package.json#L19
-3. Update app blueprint to include `{{head-layout}}` on top of https://github.com/ember-cli/ember-cli/blob/master/blueprints/app/files/app/templates/application.hbs
+3. Update app blueprint to include `<HeadLayout />` on top of https://github.com/ember-cli/ember-cli/blob/master/blueprints/app/files/app/templates/application.hbs
 4. Update route blueprint to include `{{page-title "Route name"}}` on top of the route template file
 
 ## How we teach this
--- a/text/0000-add-ember-page-title-addon.md
+++ b/text/0000-add-ember-page-title-addon.md
@@ -1,0 +1,43 @@
+- Start Date: 2020-06-24
+- Relevant Team(s): Ember CLI
+- RFC PR: (after opening the RFC PR, update this with a link to it and update the file name)
+- Tracking: (leave this empty)
+- Authors: Benjamin Jegard
+
+# RFC to add ember-page-title to the default app blueprint
+
+## Summary
+
+Add https://github.com/adopted-ember-addons/ember-page-title as a default addon for the app blueprints. (addons too ?)
+
+## Motivation
+
+This RFC is part of the work made by Ember.js Accessibility Strike Team to ensure that newly created ember apps have no accessibility issues.
+
+Users with assistive technology rely on the page title to know if they are on the correct page of a website.
+Adding this addon will provide developers a simple solution to achieve the [WCAG Success Criterion 2.4.2: Page Titled](https://www.w3.org/WAI/WCAG21/Understanding/page-titled.html)
+
+## Detailed design
+
+1. Move ember-page-title to the ember-cli org (?) (already in ember-adopted-addon org)
+2. Add the dependency to the app blueprint here: https://github.com/ember-cli/ember-cli/blob/master/blueprints/app/files/package.json#L19
+3. Update app blueprint to include `{{head-layout}}` on top of https://github.com/ember-cli/ember-cli/blob/master/blueprints/app/files/app/templates/application.hbs
+4. Update route blueprint to include `{{page-title "Route name"}}` on top of the route template file
+
+## How we teach this
+
+Update the "Page Title" section in [Page Template Considerations](https://guides.emberjs.com/release/accessibility/page-template-considerations) to use `ember-page-title`.
+
+## Drawbacks
+
+- An additional dependency.
+
+## Alternatives
+
+## Unresolved questions
+
+None
+
+## References
+
+- [WCAG Success Criterion 2.4.2: Page Titled](https://www.w3.org/WAI/WCAG21/Understanding/page-titled.html)
--- a/text/0000-add-ember-page-title-addon.md
+++ b/text/0000-add-ember-page-title-addon.md
@@ -23,7 +23,6 @@ While there're other addons that provide the same capabilities it has been decid
 
 1. Make [`ember-page-title`](https://github.com/adopted-ember-addons/ember-page-title) an official Ember addon by transferring `ember-page-title` repo to the [Ember CLI org](https://github.com/ember-cli) (it's currently in the [Adopted Ember Addons org](https://github.com/adopted-ember-addons))
 2. Add the dependency to the app blueprint here: https://github.com/ember-cli/ember-cli/blob/master/blueprints/app/files/package.json#L19
-3. Update app blueprint to include `<HeadLayout />` at the top of [application.hbs](https://github.com/ember-cli/ember-cli/blob/master/blueprints/app/files/app/templates/application.hbs) as well as a helpful comment explaining what it is used for
 4. Update route blueprint to include `{{page-title RouteName}}` at the top of the route [template.hbs](https://github.com/emberjs/ember.js/blob/master/blueprints/route/native-files/__root__/__templatepath__/__templatename__.hbs) (this [template.hbs](https://github.com/emberjs/ember.js/blob/master/blueprints/route/files/__root__/__templatepath__/__templatename__.hbs) too) where `RouteName` is the name of the route provided to the `ember generate route` command
 
 ## How we teach this
@@ -35,7 +34,7 @@ While there're other addons that provide the same capabilities it has been decid
 ## Drawbacks
 
 - An additional dependency to new projects. But, no noticeable size differences it's only all +10kb of js for production builds (not gzipped, ~2kb js gzipped)
-- Add an extra helper `{{page-tile}}` and a new component `<HeadLayout />` (possible name clashing)
+- Add an extra helper `{{page-tile}}` (possible name clashing)
 - Slightly increases the learning curve. Increases the amount of code in a first-starter project, which could become overwhelming.
 - The addon will have to be maintained by the [Ember CLI org](https://github.com/ember-cli)
 
--- a/text/0000-add-ember-page-title-addon.md
+++ b/text/0000-add-ember-page-title-addon.md
@@ -37,6 +37,7 @@ While there're other addons that provide the same capabilities it has been decid
 - An additional dependency to new projects. But, no noticeable size differences it's only all +10kb of js for production builds (not gzipped, ~2kb js gzipped)
 - Add an extra helper `{{page-tile}}` and a new component `<HeadLayout />` (possible name clashing)
 - Slightly increases the learning curve. Increases the amount of code in a first-starter project, which could become overwhelming.
+- The addon will have to be maintained by the [Ember CLI org](https://github.com/ember-cli)
 
 ## Alternatives
 
--- a/text/0000-add-ember-page-title-addon.md
+++ b/text/0000-add-ember-page-title-addon.md
@@ -2,37 +2,41 @@
 - Relevant Team(s): Ember CLI
 - RFC PR: (after opening the RFC PR, update this with a link to it and update the file name)
 - Tracking: (leave this empty)
-- Authors: Benjamin Jegard
+- Authors: Benjamin Jegard, Melanie Sumner, Ricardo Mendes
 
-# RFC to add ember-page-title to the default app blueprint
+# Add ember-page-title to the app blueprint
 
 ## Summary
 
-Add https://github.com/adopted-ember-addons/ember-page-title as a default addon for the app blueprints. (addons too ?)
+Add [`ember-page-title`](https://github.com/adopted-ember-addons/ember-page-title) to the default blueprint for new Ember apps as a way to provide improved out-of-the-box (OOB) accessibility for Ember applications.
 
 ## Motivation
 
-This RFC is part of the work made by Ember.js Accessibility Strike Team to ensure that newly created ember apps have no accessibility issues.
+This RFC is part of the work made by the Ember.js Accessibility Strike Team to ensure that newly created ember apps have no accessibility issues.
 
 Users with assistive technology rely on the page title to know if they are on the correct page of a website.
 Adding this addon will provide developers a simple solution to achieve the [WCAG Success Criterion 2.4.2: Page Titled](https://www.w3.org/WAI/WCAG21/Understanding/page-titled.html)
 
 ## Detailed design
 
-1. Move ember-page-title to the ember-cli org (?) (already in ember-adopted-addon org)
+1. Make [`ember-page-title`](https://github.com/adopted-ember-addons/ember-page-title) an official Ember addon by transferring `ember-page-title` repo to the [Ember CLI org](https://github.com/ember-cli) (it's currently in the [Adopted Ember Addons org](https://github.com/adopted-ember-addons))
 2. Add the dependency to the app blueprint here: https://github.com/ember-cli/ember-cli/blob/master/blueprints/app/files/package.json#L19
-3. Update app blueprint to include `<HeadLayout />` on top of https://github.com/ember-cli/ember-cli/blob/master/blueprints/app/files/app/templates/application.hbs
-4. Update route blueprint to include `{{page-title "Route name"}}` on top of the route template file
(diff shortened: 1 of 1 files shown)


kellyselden:
Should we audit the API of the addon as part of this effort?

text/0000-add-ember-page-title-addon.md
```diff
@@ -0,0 +1,47 @@
...
+This RFC is part of the work made by the Ember.js Accessibility Strike Team to ensure that newly created ember apps have no accessibility issues.
+
+Users with assistive technology rely on the page title to know if they are on the correct page of a website.
+Adding this addon will provide developers a simple solution to achieve the [WCAG Success Criterion 2.4.2: Page Titled](https://www.w3.org/WAI/WCAG21/Understanding/page-titled.html)
+
+## Detailed design
+
+1. Make [`ember-page-title`](https://github.com/adopted-ember-addons/ember-page-title) an official Ember addon by transferring `ember-page-title` repo to the [Ember CLI org](https://github.com/ember-cli) (it's currently in the [Adopted Ember Addons org](https://github.com/adopted-ember-addons))
+2. Add the dependency to the app blueprint here: https://github.com/ember-cli/ember-cli/blob/master/blueprints/app/files/package.json#L19
+3. Update app blueprint to include `<HeadLayout />` at the top of [application.hbs](https://github.com/ember-cli/ember-cli/blob/master/blueprints/app/files/app/templates/application.hbs)
```
buschtoens, line 24:
```suggestion
3. Update app blueprint to include `<HeadLayout />` at the top of [application.hbs](https://github.com/ember-cli/ember-cli/blob/master/blueprints/app/files/app/templates/application.hbs) as well as a helpful comment explaining what it is used for
```

text/0000-add-ember-page-title-addon.md
```diff
@@ -0,0 +1,47 @@
...
+
+## How we teach this
+
+- Update the "Page Title" section in [Page Template Considerations](https://guides.emberjs.com/release/accessibility/page-template-considerations) to use `ember-page-title`.
+- Update code examples in [Building Pages](https://guides.emberjs.com/release/tutorial/part-1/building-pages/) to include uses of `{{page-title}}`. Also explain that updating the page title with the current page name will help users with assistive technology locate themself on the website.
+
+## Drawbacks
+
+- An additional dependency to new projects. But, no noticeable size differences it's only all +10kb of js for production builds (not gzipped, ~2kb js gzipped)
+- Add an extra helper `{{page-tile}}` and a new component `<HeadLayout />` (possible name clashing)
```
buschtoens, line 35: Another drawback, already kinda included in the last point, would be:

```suggestion
## Drawbacks

- An additional dependency to new projects. But, no noticeable size differences it's only all +10kb of js for production builds (not gzipped, ~2kb js gzipped)
- Add an extra helper `{{page-tile}}` and a new component `<HeadLayout />` (possible name clashing)
- Slightly increases the learning curve. Increases the amount of code in a first-starter project, which could become overwhelming.
```

Arguably though at least I personally always used to wonder how to properly set the page title pretty much a few days after I started using Ember, but didn't find proper / definitive ressoruces back then, so you could also sell it as an advantage, depending on how you look at it.

text/0000-add-ember-page-title-addon.md
```diff
@@ -0,0 +1,47 @@
...
+3. Update app blueprint to include `<HeadLayout />` at the top of [application.hbs](https://github.com/ember-cli/ember-cli/blob/master/blueprints/app/files/app/templates/application.hbs)
+4. Update route blueprint to include `{{page-title RouteName}}` at the top of the route [template.hbs](https://github.com/emberjs/ember.js/blob/master/blueprints/route/native-files/__root__/__templatepath__/__templatename__.hbs) (this [template.hbs](https://github.com/emberjs/ember.js/blob/master/blueprints/route/files/__root__/__templatepath__/__templatename__.hbs) too) where `RouteName` is the name of the route provided to the `ember generate route` command
+
+## How we teach this
+
+- Update the "Page Title" section in [Page Template Considerations](https://guides.emberjs.com/release/accessibility/page-template-considerations) to use `ember-page-title`.
+- Update code examples in [Building Pages](https://guides.emberjs.com/release/tutorial/part-1/building-pages/) to include uses of `{{page-title}}`. Also explain that updating the page title with the current page name will help users with assistive technology locate themself on the website.
```
buschtoens, line 30: We could also expand the `ember-welcome-page` addon to explain the contents of the fresh project a bit better.

text/0000-add-ember-page-title-addon.md
```diff
@@ -0,0 +1,47 @@
...
+- Relevant Team(s): Ember CLI
+- RFC PR: (after opening the RFC PR, update this with a link to it and update the file name)
+- Tracking: (leave this empty)
+- Authors: Benjamin Jegard, Melanie Sumner, Ricardo Mendes
+
+# Add ember-page-title to the app blueprint
+
+## Summary
+
+Add [`ember-page-title`](https://github.com/adopted-ember-addons/ember-page-title) to the default blueprint for new Ember apps as a way to provide improved out-of-the-box (OOB) accessibility for Ember applications.
```
NullVoxPopuli, line 11: does ember-page-title have a way to have a persistent leading dynamic segment? like if you want 
```
NullVoxPopuli | emberclear -- no unread
or
 
4 | NullVoxPopuli | emberclear

or

22 | emberclear // no active chat -- all unread

or

4 | NullVoxPopuli | 22 Unread total | emberclear
```

looking at the docs, it looks like, 

```
application route: 
```
{{page-title "emberclear"}}

{{#if (not (start-with this.router.currentRoute "chat"))}}
  {{page-title this.totalUnread " Unread total"
{{/if}}

```
chat/:id: 
```
{{page-title this.chattingWith.name}}

{{#if this.numUnread}}
  {{page-title this.numUnread }}
{{/if}}
```
?

Comment (NullVoxPopuli) [inline text/0000-add-ember-page-title-addon.md:11]: In the examples, I see: 

image

Is there a way to omit the `Segment > Segment` stuff? I wouldn't want a breadcrumb in the title all the time :D

**Comment by NullVoxPopuli** [comment]
> Should we audit the API of the addon as part of this effort?

@kellyselden I think so. There are certain capabilities I'd want to make sure are present, and I'd want to make sure that any addon we add "feels nice" to use 

---- 

I have no experience with this addon, and haven't used anything other than custom title manipulation, so predisposition-wise, I'm starting from scratch. :D 

Left some questions

File: text/0000-add-ember-page-title-addon.md (hunk 4 above)
Comment (NullVoxPopuli) [inline text/0000-add-ember-page-title-addon.md:11]: does the addon maintain a history of the titles?

Like, if I were to do a bunch of navigations where some, but not all page-title invocations have `replace=true`, does the title always reflect what is rendered?

Does that mean that the page title is dependent on render order?

What happens if I do?

{{#each this.breadcrumbs |breadcrumb|}}
  {{page-title breadcrumb.name}}
{{/each}}

Comment (KamiKillertO) [reply text/0000-add-ember-page-title-addon.md:11]: I'm not sure to totally understand the question. But I think the answer is yes 😅

Comment (KamiKillertO) [reply text/0000-add-ember-page-title-addon.md:11]: The separator can be specified for each use of `{{page-title}}` and you can also set the default separator in `config/environment.js` (everything is explained in the [README](https://github.com/adopted-ember-addons/ember-page-title#page-title-helper))

Comment (NullVoxPopuli) [reply text/0000-add-ember-page-title-addon.md:11]: it's not just the separator, I feel like there is duplication.

Like, I'd want `PostTitle | Posts | Ember.js`, not `Post Title > Posts | Ember.js > Blog > etc`

Comment (NullVoxPopuli) [reply text/0000-add-ember-page-title-addon.md:11]: yeah, based on the docs, I think so, too :D

Comment (KamiKillertO) [reply text/0000-add-ember-page-title-addon.md:11]: I think there's an issue on the doc page. All examples affect each others ><, for examples the part `Tim Evans (@ timmyce)` comes from the last example ><.

**Comment by btecu** [comment]
Why `ember-page-title` and not something like https://github.com/mike-north/ember-cli-document-title-northm (fork of https://github.com/kimroen/ember-cli-document-title)?

**Comment by NullVoxPopuli** [comment]
> Why ember-page-title and not something like https://github.com/mike-north/ember-cli-document-title-northm (fork of https://github.com/kimroen/ember-cli-document-title)?

While I have no allegiance to any of these addons, I think in the alternatives / prior art  and/or detailed design section, there should be some comparison between them? What makes one addon objectively better than the others and worth including by default?

**Comment by kellyselden** [comment]
Yeah I think we should be committed to whatever addon and API we choose. It has to make sense, and not just come along because the addon is popular. For instance. I see that the way to configure the addon is a key on `ENV` in config/environment.js. Is that still the canonical way to configure addons in an Octane world? The answer is probably yes, but I just want to make sure.

**Comment by KamiKillertO** [comment]
The first intent of the a11y strike team was to create an addon with the features of `ember-page-title` and `ember-document-title`. But, according to @MelSumner and a discussion she had with Yehuda (if remember correctly), `ember-page-title` is more aligned with the direction ember and front end frameworks are taking. Maybe @MelSumner can confirm  😊

**Comment by MelSumner** [comment]
> The first intent of the a11y strike team was to create an addon with the features of `ember-page-title` and `ember-document-title`. But, according to @MelSumner and a discussion she had with Yehuda (if remember correctly), `ember-page-title` is more aligned with the direction ember and front end frameworks are taking. Maybe @MelSumner can confirm  😊

Yes that’s correct! We felt ember-page-title was the most aligned with the current and future direction of the framework (a template-based solution).

File: text/0000-add-ember-page-title-addon.md (hunk 2 above)
Comment (KamiKillertO) [reply text/0000-add-ember-page-title-addon.md:35]: Even if we already have good documentation about page title, it's still worth mentioning it 👍 .

File: text/0000-add-ember-page-title-addon.md (hunk 4 above)
Comment (KamiKillertO) [reply text/0000-add-ember-page-title-addon.md:11]: I'm not sure, I'll make a dummy app to test ^^

**Comment by MelSumner** [comment]
@rwjblue reminder to file the relevant issues on the addon so we can move to FCP this RFC. 👍

**Comment by knownasilya** [comment]
@rwjblue are there any breaking changes required? Was planning on releasing a new major version this week.

File: text/0000-add-ember-page-title-addon.md [hunk 5]
```diff
@@ -0,0 +1,48 @@
...
+This RFC is part of the work made by the Ember.js Accessibility Strike Team to ensure that newly created ember apps have no accessibility issues.
+
+Users with assistive technology rely on the page title to know if they are on the correct page of a website.
+Adding this addon will provide developers a simple solution to achieve the [WCAG Success Criterion 2.4.2: Page Titled](https://www.w3.org/WAI/WCAG21/Understanding/page-titled.html)
+
+## Detailed design
+
+1. Make [`ember-page-title`](https://github.com/adopted-ember-addons/ember-page-title) an official Ember addon by transferring `ember-page-title` repo to the [Ember CLI org](https://github.com/ember-cli) (it's currently in the [Adopted Ember Addons org](https://github.com/adopted-ember-addons))
+2. Add the dependency to the app blueprint here: https://github.com/ember-cli/ember-cli/blob/master/blueprints/app/files/package.json#L19
+3. Update app blueprint to include `<HeadLayout />` at the top of [application.hbs](https://github.com/ember-cli/ember-cli/blob/master/blueprints/app/files/app/templates/application.hbs) as well as a helpful comment explaining what it is used for
```
raido, line 24: This will not be required after this gets done - https://github.com/adopted-ember-addons/ember-page-title/issues/167 - already in progress.

text/0000-add-ember-page-title-addon.md
```diff
@@ -0,0 +1,48 @@
...
+
+## How we teach this
+
+- Update the "Page Title" section in [Page Template Considerations](https://guides.emberjs.com/release/accessibility/page-template-considerations) to use `ember-page-title`.
+- Update code examples in [Building Pages](https://guides.emberjs.com/release/tutorial/part-1/building-pages/) to include uses of `{{page-title}}`. Also explain that updating the page title with the current page name will help users with assistive technology locate themself on the website.
+
+## Drawbacks
+
+- An additional dependency to new projects. But, no noticeable size differences it's only all +10kb of js for production builds (not gzipped, ~2kb js gzipped)
+- Add an extra helper `{{page-tile}}` and a new component `<HeadLayout />` (possible name clashing)
```
raido, line 35: We can remove `<HeadLayout />` part after this gets closed - https://github.com/adopted-ember-addons/ember-page-title/issues/167

raido (changes requested):

text/0000-add-ember-page-title-addon.md
KamiKillertO: @buschtoens what about

```
- Update [ember-welcome-page](https://github.com/ember-cli/ember-welcome-page) addon and [WelcomePage](https://github.com/ember-cli/ember-welcome-page/blob/master/addon/templates/components/welcome-page.hbs) component to explain the contents of the fresh project a bit better. 
```

text/0000-add-ember-page-title-addon.md
raido: 🔝  is done but there are 2 blueprint update issues still to do:

* https://github.com/adopted-ember-addons/ember-page-title/issues/174
* https://github.com/ember-fastboot/ember-cli-fastboot/issues/792

simonihmig:
> Yes that’s correct! We felt ember-page-title was the most aligned with the current and future direction of the framework (a template-based solution).

While I can relate to that, I think the RFC should explain that in some detail. Like there are a couple of other similar addons - which IIRC follow a route-based rather than template based API - and why do we pick this one and not one of the others?

Which brings me to another topic: meta tags. While this RFCs only mentions a11y concerns as a motivation - which is legit - another one is SEO, especially when used with FastBoot. But when thinking of SEO, you also want to account for proper meta tags. 

While this does not necessarily be tackled within the same RFC, I think we should have at least a broad idea about how to extend support to meta tags. So there a couple of addons (e.g. `ember-cli-meta-tags`, `ember-meta`) for this, which
* are also based on `ember-cli-head`
* follow a different route-based API
* also allow to set the title tag ⚠️

So when endorsing one addon in such a massive way by making it a default one, we should make sure there is a viable way to also support meta tags, that
* follows the same API philosophy (template vs. route)
* does not cause conflicts or overlapping concerns
* and as such guarantees a pleasant DX

Or at least we should make sure there are no reasons to later revert our decisions here because of this.

Another thing: what does making a "user-land" addon a default one mean in terms of maintenance guarantees? I am in no way suggesting this addon is not well maintained (frankly I have no idea, I haven't used it before), but the addon is part of the `addopted-ember-addons` *because* of lack of maintenance in the past I assume, so does adoptig this RFC basically mean this addons gets a maintenance guarantee from some official Ember team (core, cli?), in case current maintainers are not able to provide that anymore? If so, does it make sense to transfer the addon to the @emberjs org, to signal that stability guarantee to users?

text/0000-add-ember-page-title-addon.md
KamiKillertO: Great that's good news.
I'll wait until the new version is released to remove this.

KamiKillertO:
> While I can relate to that, I think the RFC should explain that in some detail. Like there are a couple of other similar addons - which IIRC follow a route-based rather than template based API - and why do we pick this one and not one of the others?

It's a good idea, I'm gonna add a small explanation 👍 .

> Another thing: what does making a "user-land" addon a default one mean in terms of maintenance guarantees? I am in no way suggesting this addon is not well maintained (frankly I have no idea, I haven't used it before), but the addon is part of the addopted-ember-addons because of lack of maintenance in the past I assume, so does adopting this RFC basically mean this addons gets a maintenance guarantee from some official Ember team (core, cli?), in case current maintainers are not able to provide that anymore? If so, does it make sense to transfer the addon to the @emberjs org, to signal that stability guarantee to users?

In the RFC, it's indicated that this addon should be transferred to the [Ember CLI org](https://github.com/ember-cli). It's true that making an addon part of the default blueprint involves that the official Ember team will have to maintain it somehow and we should be very careful about that. This question was raised while working on the [RFC 638](https://emberjs.github.io/rfcs/0638-interactive-app-creation.html) and now the RFC clearly indicates that before adding any new option using an external addon, we should [evaluate the maintenance cost](https://emberjs.github.io/rfcs/0638-interactive-app-creation.html#question-rubric) that might come with.

knownasilya:
I'm interested in the meta tags discussion, and am highly aware of the importance there as well. We just removed ember-cli-head from ember-page-title, so the meta support is no longer included because of that dependency. I think that area is worth discussing but unsure if it should be in this PR. Maybe something template based as well.

```hbs
{{page-meta (hash og:title='Test')}}
```

Not sure we can do something like that with special characters. Will have to see if `(hash 'og:title'='Test')` works.

gossi:
I agree with @simonihmig here. There is more than just `<title>`. I can relate to all his points. Indeed I found an old gist I put together that explains the whole problem around this: https://gist.github.com/gossi/c3d6fded03be001d8b9763a8219b0a5e It's from almost one year ago. Ignore the code, take the idea.

Actually, what I would expect from ember would be an API that the various addons can connect to. So ember is reigning `<head>` (or parts of it) and provides an API to that. Addons will use this API to exchange information. I as a consumer can choose an addon that provides the API **I** am looking for - not the one ember picks for me. For example:

I will write myself a decorator that I put on my routes, to control the title and it's hierarchy/information architecture of my site and another breadcrumb component (which I installed as an addon) that will use that generated data to control the breadcrumbs. The data lives in the "central head content repository".

knownasilya:
I think we can have both, an official service and helpers that let you change that without having to install addons. For accessibility it should be included out of the box, because otherwise it's not done, especially by junior devs.

text/0000-add-ember-page-title-addon.md
KamiKillertO: ember-page-title v6 has been released, so we can remove these lines 👍

MelSumner:
@KamiKillertO now that `ember-page-title` has a new release, I think we can update this RFC to reflect that, and work on getting this RFC shipped. 👍🏻

KamiKillertO:
Thanks @MelSumner 👍 , I've already updated the RFC to reflect `ember-page-title` changes

MelSumner:
We have discussed this in today's Framework Core Team meeting, and decided to move this RFC to Final Comment Period.

hergaiety:
_Just voicing support for this RFC_ and bringing the addon into the core app. I'm glad to see page titles will be less of an afterthought in the apps I work on every day and I'll no longer need to make a case to include another addon, as that case has already been made here! Great work to everyone involved :tada:

text/0000-add-ember-page-title-addon.md
```diff
@@ -0,0 +1,51 @@
...
+- Update the "Page Title" section in [Page Template Considerations](https://guides.emberjs.com/release/accessibility/page-template-considerations) to use `ember-page-title`.
+- Update code examples in [Building Pages](https://guides.emberjs.com/release/tutorial/part-1/building-pages/) to include uses of `{{page-title}}`. Also explain that updating the page title with the current page name will help users with assistive technology locate themself on the website.
+- Update [ember-welcome-page](https://github.com/ember-cli/ember-welcome-page) addon and [WelcomePage](https://github.com/ember-cli/ember-welcome-page/blob/master/addon/templates/components/welcome-page.hbs) component to explain the contents of the fresh project a bit better. The component should explain the presence of `{{page-title}}` and link to the "Page Title" section in [Page Template Considerations](https://guides.emberjs.com/release/accessibility/page-template-considerations)
+
+## Drawbacks
+
+- An additional dependency to new projects. But, no noticeable size differences it's only all +10kb of js for production builds (not gzipped, ~2kb js gzipped)
+- Add an extra helper `{{page-tile}}` (possible name clashing)
```
raido, line 37: A typo, page-tile -> page-title :)

text/0000-add-ember-page-title-addon.md
raido: Documentation page has been improved lately. This should no longer be an issue.

raido:
This seems to be still relevant issue - https://github.com/adopted-ember-addons/ember-page-title/issues/90 from accessibility perspective?

text/0000-add-ember-page-title-addon.md
raido: I'll add nested route + breadcrumbs test case to ember-page-title and we'll find out.

KamiKillertO:
I don't think 🤔.
Screen readers don't announce page title updates when there are no actual page navigations (navigation to a new HTML page), which is the case with ember applications (because of the router). This why the addon [a11y-announcer](https://github.com/ember-a11y/a11y-announcer) exist.

rwjblue:
Lets do it!

PR #648, opened 2020-07-15
Extend the supported asset file types to include commonly used formats svg and webp

st-h:
[rendered](https://github.com/st-h/rfcs/blob/image-support/text/0000-extend-supported-asset-file-types-svg-webp.md)

### DIFF
--- a/text/0000-extend-supported-asset-file-types-svg-webp.md
+++ b/text/0000-extend-supported-asset-file-types-svg-webp.md
@@ -1,0 +1,36 @@
+- Start Date: 2020-07-15
+- Relevant Team(s): Ember CLI
+- RFC PR: (after opening the RFC PR, update this with a link to it and update the file name)
+- Tracking: (leave this empty)
+
+# Extend the supported asset file types to include svg and webp (& possibly other)
+
+## Summary
+
+Support commonly used image formats svg and webp out of the box.
+
+## Motivation
+
+A default ember app should support commonly used image formats without requiring advanced changes to the config.
+
+## Detailed design
+
+Ember uses [broccoli-asset-rev](https://github.com/rickharrison/broccoli-asset-rev) to generate the appropriate urls within the ember application. The default configuration includes the following file types: `['js', 'css', 'png', 'jpg', 'gif', 'map']`. These days `svg` images are regularly used for vector graphics and `webp` is used in order to save bandwith (where supported). However, if these files are added to a new ember application, they are treated differently than the ones listed above. Fingerprinting is not enabled and related configurations like `prepend` are ignored. This can lead to anything from images not being displayed up to difficult to find issues like images being loaded from the wrong server, or broken cache invalidation.
+
+This RFC proposes to add `svg` and `webp` to the default file types handled by broccoli-asset-rev.
+
+## How we teach this
+
+Not necesary. Better image support - less things to teach.
+
+## Drawbacks
+
+Some users might have been unaware of the possibility to configure these file types to behave correctly and might have implemented other things based on that beheaviour. There might be a possibility for these things to break when the defaults are changed. To avoid this, a deprecation warning could be implemented before changing the defaults. However, it is unclear if and how many existing users would even be affected by such a change.
+
+## Alternatives
+
+Add more documentation. The cli documentation currently lists this option, but it is listed under `advanced-use`. As using a svg image is pretty much a trivial thing, this should be reflected accordingly within the docs.
+
+## Unresolved questions
+
+Do we need some sort of deprecation warning?
--- a/text/0000-extend-supported-asset-file-types-svg-webp.md
+++ b/text/0000-extend-supported-asset-file-types-svg-webp.md
@@ -3,7 +3,7 @@
 - RFC PR: (after opening the RFC PR, update this with a link to it and update the file name)
 - Tracking: (leave this empty)
 
-# Extend the supported asset file types to include svg and webp (& possibly other)
+# Extend the supported asset file types to include svg and webp
 
 ## Summary
 
@@ -11,11 +11,24 @@ Support commonly used image formats svg and webp out of the box.
 
 ## Motivation
 
-A default ember app should support commonly used image formats without requiring advanced changes to the config.
+A default Ember app should support commonly used image formats without requiring advanced changes to the config.
 
 ## Detailed design
 
-Ember uses [broccoli-asset-rev](https://github.com/rickharrison/broccoli-asset-rev) to generate the appropriate urls within the ember application. The default configuration includes the following file types: `['js', 'css', 'png', 'jpg', 'gif', 'map']`. These days `svg` images are regularly used for vector graphics and `webp` is used in order to save bandwith (where supported). However, if these files are added to a new ember application, they are treated differently than the ones listed above. Fingerprinting is not enabled and related configurations like `prepend` are ignored. This can lead to anything from images not being displayed up to difficult to find issues like images being loaded from the wrong server, or broken cache invalidation.
+Ember uses [broccoli-asset-rev](https://github.com/rickharrison/broccoli-asset-rev) to generate the appropriate urls within the Ember application. Ember's default value for `extensions` is currently the same as broccoli-asset-rev's:
+
+```javascript
+// Current default extensions
+['js', 'css', 'png', 'jpg', 'gif', 'map']
+```
+
+These days, `svg` images are regularly used for vector graphics and `webp` is used in order to save bandwidth (where supported). If `svg` and `webp` files are added to a new Ember application, they are treated differently than files with extensions listed above. Fingerprinting is not enabled and related configurations like `prepend` are ignored.
+
+This can lead to:
+
+- Images not being displayed
+- Images being loaded from the wrong server
+- Broken cache invalidation
 
 This RFC proposes to add `svg` and `webp` to the default file types handled by broccoli-asset-rev.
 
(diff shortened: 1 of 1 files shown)


text/0000-extend-supported-asset-file-types-svg-webp.md
```diff
@@ -0,0 +1,36 @@
+- Start Date: 2020-07-15
+- Relevant Team(s): Ember CLI
+- RFC PR: (after opening the RFC PR, update this with a link to it and update the file name)
+- Tracking: (leave this empty)
+
+# Extend the supported asset file types to include svg and webp (& possibly other)
```
ijlee2, line 6: Let's move "possibly others" into unresolved questions.
```suggestion
# Extend the supported asset file types to include svg and webp
```

text/0000-extend-supported-asset-file-types-svg-webp.md
```diff
@@ -0,0 +1,36 @@
...
+
+## Summary
+
+Support commonly used image formats svg and webp out of the box.
+
+## Motivation
+
+A default ember app should support commonly used image formats without requiring advanced changes to the config.
```
ijlee2, line 14:
```suggestion
A default Ember app should support commonly used image formats without requiring advanced changes to the config.
```

text/0000-extend-supported-asset-file-types-svg-webp.md
```diff
@@ -0,0 +1,36 @@
...
+
+## Detailed design
+
+Ember uses [broccoli-asset-rev](https://github.com/rickharrison/broccoli-asset-rev) to generate the appropriate urls within the ember application. The default configuration includes the following file types: `['js', 'css', 'png', 'jpg', 'gif', 'map']`. These days `svg` images are regularly used for vector graphics and `webp` is used in order to save bandwith (where supported). However, if these files are added to a new ember application, they are treated differently than the ones listed above. Fingerprinting is not enabled and related configurations like `prepend` are ignored. This can lead to anything from images not being displayed up to difficult to find issues like images being loaded from the wrong server, or broken cache invalidation.
```
ijlee2, line 18: While each sentence is clear, I think breaking the long paragraph into a few will help. What do you think about:

``````suggestion
Ember uses [broccoli-asset-rev](https://github.com/rickharrison/broccoli-asset-rev) to generate the appropriate urls within the Ember application. Ember's default value for `extensions` is currently the same as `broccoli-asset-rev`'s:

```javascript
// Current default extensions
['js', 'css', 'png', 'jpg', 'gif', 'map']
```

These days, `svg` images are regularly used for vector graphics and `webp` is used in order to save bandwidth (where supported). If `svg` and `webp` files are added to a new Ember application, they are treated differently than files with extensions listed above. Fingerprinting is not enabled and related configurations like `prepend` are ignored.

This can lead to:

- Images not being displayed
- Images being loaded from the wrong server
- Broken cache invalidation
``````

File: text/0000-extend-supported-asset-file-types-svg-webp.md [hunk 4]
```diff
@@ -0,0 +1,36 @@
...
+
+This RFC proposes to add `svg` and `webp` to the default file types handled by broccoli-asset-rev.
+
+## How we teach this
+
+Not necesary. Better image support - less things to teach.
+
+## Drawbacks
+
+Some users might have been unaware of the possibility to configure these file types to behave correctly and might have implemented other things based on that beheaviour. There might be a possibility for these things to break when the defaults are changed. To avoid this, a deprecation warning could be implemented before changing the defaults. However, it is unclear if and how many existing users would even be affected by such a change.
```
ijlee2, line 28:
```suggestion
Some users might have been unaware of the possibility to configure file extensions and might have implemented other things based on that behavior. There might be a possibility for these things to break when the defaults are changed.

To avoid this, a deprecation warning could be implemented before changing the defaults. It is unclear if and how many existing users would even be affected by such a change.
```

text/0000-extend-supported-asset-file-types-svg-webp.md
```diff
@@ -0,0 +1,36 @@
...
+
+## Alternatives
+
+Add more documentation. The cli documentation currently lists this option, but it is listed under `advanced-use`. As using a svg image is pretty much a trivial thing, this should be reflected accordingly within the docs.
```
ijlee2, line 32: In addition to adding the link to CLI Guides, I think I'd update the last sentence to avoid words like "pretty much" and "trivial", which can sound diminutive and less persuasive.

I understood the last sentence to mean, it may be common for a developer to add an svg image. Therefore, we could make it easier for the developer to find how to configure extensions in the Guides.

If this is what you intended with the sentence, maybe we can update it to reflect this?

```suggestion
Add more documentation. The [CLI documentation](https://cli.emberjs.com/release/advanced-use/asset-compilation/#fingerprintingandcdnurls) currently lists this option, but it is listed under `advanced-use`. As using a svg image is pretty much a trivial thing, this should be reflected accordingly within the docs.
```

ijlee2:
@st-h Thanks for the write-up. I left a few suggestions to make your writing more clear. I thought you already did a great job explaining the problem.

Here are a couple of questions that may need to be addressed in the write-up. I haven't thought them through so they may or may not end up being relevant.

- What will including `svg` and `webp` to the default `extensions` mean for addons like [`ember-svg-jar`](https://github.com/ivanvotti/ember-svg-jar)? What will addon authors need to do?
- What will `ember-cli-update` do when upgrading an Ember app or addon from one version to another? Does it keep the current `extensions` or change it to the new default?

st-h:
@ijlee2 thanks a lot for your comments.

- I am not really sure if this would even have any implications for addons like `ember-svg-jar`. I don't think this change will break anything, other than instructions like [this](https://github.com/ivanvotti/ember-svg-jar/blob/eb1290be0af25015bdc755fba95d595359e1e5f9/docs/svg-fingerprinting.md) can finally be deprecated (unless we need to warn users about fingerprinting).
- I personally would not update any extension that have been added to the app config. If one wants to be super safe, ember-cli could insert the current defaults before upgrading, if no config is present. But generally this change should be an improvement, rather than something which will cause issues.

ivanvotti:
@st-h @ijlee2  All good 👍 it's not going to break [ember-svg-jar](https://github.com/ivanvotti/ember-svg-jar).

mehulkar:
yes please!

text/0000-extend-supported-asset-file-types-svg-webp.md
```diff
@@ -0,0 +1,52 @@
...
+To avoid this, a deprecation warning could be implemented before changing the defaults and/or ember-cli-update could add the current formats to the config when upgrading, so behaviour will not change within an ember cli upgrade. However, it is unclear if and how many existing users would even be affected by such a change. It seems somewhat unlikely that this will actually be an issue, but it seems reasonable to think about any possible side effects.
+
+## Alternatives
+
+Add more documentation. The [CLI documentation](https://cli.emberjs.com/release/advanced-use/asset-compilation/#fingerprintingandcdnurls) currently lists this option, but it is listed under `advanced-use`. As using a svg image is pretty much a trivial thing, this should be reflected accordingly within the docs.
+
+## Unresolved questions
+
+Do we need some sort of deprecation warning?
+Are there other common file formats within web apps these days, that should be added as well?
```
jrjohnson, line 52: I have no idea if it is common, but I have `webmanifest` in our config as well.

st-h: for PWAs this should be common. However [mdn](https://developer.mozilla.org/en-US/docs/Web/Manifest) specifies it as `manifest.webmanifest`, but browser generally accept `manifest.json` as well.

st-h:
I am wondering, if the root cause actually is that asset-rev seems to handle both, fingerprinting and prepending something to the url. For fingerprinting it wouldn't be bad per se to not include all common files. But for prepending the url, this easily becomes an issue as for instance static domains won't work without carefully configuring all used file types.

Shouldn't this work in a way that all assets are prepended, but only those configured are fingerprinted?

st-h:
To move this RFC on, I think it is reasonable to include commonly used file types (which will be an improvement for most users) and discuss separating the config for fingerprinting and prefixing (or how broccoli-asset-rev handles things in general) in a dedicated RFC.

MelSumner:
@st-h I don't think this needs an RFC- I would submit a PR to the addon to add support for these asset file types.

st-h:
@MelSumner Thanks for your comment. I have been told on ember discord to open an RFC for this.
Personally I would not try to impose embers requirements on a general Broccoli plugin. And looking at the current state of this plugin, it is questionable if it even still is maintained. There has been some discussion in the dev-rfc about the future of that plug-in and possible replacements, but it contained too many internal details (that I am not aware of) to follow along.

MelSumner:
@st-h That's totally fine; we discussed this issue in Steering Committee (at a meta level- what process do we want to have around issues like new file formats _in general_), and we decided that we don't see this as a change that requires an RFC, and can be submitted as a PR to the appropriate repo. Hope this helps- please let me know if I can clarify anything further. 👍

MelSumner:
After discussing in our core framework team meeting, we have decided to move to close this RFC. Please chat with @rwjblue in Discord if there are any further questions. Thank you!

st-h:
Let's all pray to the gods of open source, that @rwjblue will never stop to take care so many ember related things

@MelSumner Did you talk about what the appropriate repo would be? Sorry, I just noticed your previous comment right now.

MelSumner:
The FCP to close timeframe has come to an end, so we are closing this RFC. Thank you!

Issue #612, opened 2020-04-03
Pre-RFC: (guid-for) helper


steveszc:
## The goal:

Provide a `guid-for` template helper for ergonomically generating unique IDs.
```hbs
<label for={{guid-for "toggle"}}>Toggle</label>
<input id={{guid-for "toggle"}} type="checkbox">
```

This idea originated from the Accessibility Strike Team's discussions of #595  

## Motivation

The primary use case for an `guid-for` helper is to programmatically associate labels and input elements using the label's `for` attribute and the input's `id` attribute. This pattern is required for accessibility in cases where it is not possible or convenient to nest input elements inside of their associated label ([details here](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label)). In addition to that use-case, this helper would be useful anytime the html `id` attribute needs to added to a dom element. 

The HTML spec requires that `id` attributes be unique in the whole document (https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id). Ember components provide the elementId attribute which could be used to derive unique ids within components, but this is no longer the case with Glimmer components.  

Since providing some faculty for generating unique IDs for dom elements can reasonably be considered a requirement for most web apps wishing to provide accessible form elements, it is reasonable for Ember to provide this at the framework-level. Ember already provides the [guidFor util in javascript](https://api.emberjs.com/ember/3.12/functions/@ember%2Fobject%2Finternals/guidFor), so it is reasonable for Ember to expose this via a helper within templates.  

## Approach

This helper could be included directly in ember as a built-in helper, or it could exist as an add-on that would be included in the default blueprint for new Ember apps. 

## Ergonomics

Developers should be able to use the `guid-for` helper to generate the same guid value every time the helper is called with a given argument within a given template invocation. But in another template (or within another invocation of the same template) that same argument should be expected to return a different guid. 

For example, in an app with a sign-in form and a password reset form, both existing in separate templates, I should be able to use `(guid-for "email-address")` in both templates and have a different guid generated. Additionally, I should be able to create a `CustomInput` component that renders a generic label/input pair using `(guid-for "input")`, and render that component multiple times on a page, and `(guid-for "input")` should generate a different guid in every instance of that component.

[...]

PR #659, opened 2020-08-25
unique-id helper

steveszc:
See [pre-RFC issue #612](https://github.com/emberjs/rfcs/issues/612)

[Rendered](https://github.com/steveszc/rfcs/blob/id-helper/text/0659-unique-id-helper.md)


text/0659-id-helper.md
```diff
@@ -0,0 +1,198 @@
...
+
+Some developers may not prefer the ergonomics of using the `{{id}}` with `let` blocks with-in template-only components.
+
+Optional usage of the `for` argument may make this slightly harder to teach and understand, and may lead to fragmented usage patterns between template-only components and class-backed templates.
+
+There is nothing about this proposal that could not be instead implemented in an add-on.
+
+## Alternatives
+
+1. Do nothing; developers can use backing classes for templates that require an ID, and either use elementId in classic components or import guidFor in glimmer components or via a hand-rolled helper.
```
MelSumner, line 176:
```suggestion
1. Do nothing; developers can use backing classes for templates that require an ID, and either use `elementId` in classic components or import `guidFor` in glimmer components or via a hand-rolled helper.
```

joukevandermaas:
I think calling this helper `{{id}}` is a problem for backwards compatibility. If older components have a local property called `id`, it will break if the template didn't use `this.` yet. Example:

```hbs
<button id={{id}}>my button</button>
```

This perfectly valid piece of code will now break, because `{{id}}` is now a helper. Of course, this can be addressed by changing the code:


```hbs
<button id={{this.id}}>my button</button>
```

But I would argue this type of issue can be pretty hard to track down. Maybe something like `unique-id` would work better?

sandstrom:
I think a less generic/general name would be better.

For example, I think the previously proposed `{{guid-for …}}` name is better. But other names would work too, my main complaint is that `{{id …}}` is too general.

bendemboski:
Agreed for both backwards compatibility and readability reasons. I'd propose `{{unique-id}}` as it's much less likely to conflict with anything else, and IMO pretty clearly describes what it does.

bendemboski:
Oh, haha, just read https://github.com/emberjs/rfcs/issues/612#issuecomment-674173647, which I hadn't before. I guess that's corroboration that `unique-id` will be readable/intuitive because I came up with the same name on my own 😆

steveszc:
> I think calling this helper `{{id}}` is a problem for backwards compatibility. If older components have a local property called `id`, it will break if the template didn't use `this.` yet. Example:
> 
> ```
> <button id={{id}}>my button</button>
> ```
> 
> This perfectly valid piece of code will now break, because `{{id}}` is now a helper. Of course, this can be addressed by changing the code:
> 
> ```
> <button id={{this.id}}>my button</button>
> ```
> 
> But I would argue this type of issue can be pretty hard to track down. Maybe something like `unique-id` would work better?

Yeah I agree this is a problem we need to solve, but I'm not sure picking a different name will actually solve the problem (though it may make it less likely to occur). Any other name we could come up with would have the same potential problem colliding with existing controller/component properties. 

Admittedly, other names we might choose (eg. 'unique-id`) may be less likely to to have this problem, but it is still valid and possible for an existing template to reference a property called `unique-id`.

The argument-less helper invocation proposed here may be problematic for this reason, and it will be complicated by [template strict mode](https://github.com/emberjs/rfcs/blob/master/text/0496-handlebars-strict-mode.md#3-no-implicit-invocation-of-argument-less-helpers), which will require mandatory parens to invoke an argument-less helper.

MelSumner:
> Admittedly, other names we might choose (eg. unique-id) may be less likely to to have this problem, but it is still valid and possible for an existing template to reference a property called unique-id.

I think this is something we could find out- let's do a code search in [Ember Observer](https://emberobserver.com/) and ask the community to weigh in on it (via Discord, Twitter, etc). That way we could choose something that won't be problematic for _most_ use cases. 

What do you think?

steveszc:
@MelSumner 
> I think this is something we could find out- let's do a code search in [Ember Observer](https://emberobserver.com/) and ask the community to weigh in on it (via Discord, Twitter, etc). That way we could choose something that won't be problematic for _most_ use cases.

Looks like `unique-id` will indeed be safer. A few large add-ons use `{{id}}`, including ember-concurrency, ember-bootstrap, and a host of materialize addons.
{{unique-id}} has 0 results on ember observer
{{id}} is found in 122 add-ons on Ember Observer

bendemboski:
> It also may be worth considering the names uid and guid for terseness

What's the benefit of terseness? I understand that some built-in helpers are pretty terse, but we have `{{link-to}}`, `{{query-params}}`, `{{has-block}}` and `{{has-block-params}}`, so I don't think something like `{{unique-id}}` would be an outlier. Also since it's generally expected to be used in a let statement, terseness seems even less important than, e.g. `{{eq}}` or `{{get}}` which are always going to be used in compound statements.

luxferresum:
In any case I strongly agree with the comments that `unique-id` would be much better.

However I *do* like the idea to also have the AST transforming variant in ember itself. I currently have an [addon](https://www.npmjs.com/package/@ember-lux/id-helper) that basically implements what is proposed as alternative here. And its IMHO much more ergonomic to not require additional nesting.

sukima:
This would save a lot of boilerplate for me. Without I do this quite _frequently_:

```js
  guid = guidFor(this);
}
```

```hbs
{{#let (concat this.guid "-input") as |id|}}
  <label for={{id}}>Name</label>
  <input id={{id}} …>
{{/let}}
```

steveszc:
I've just updated this RFC to use the helper name `unique-id` instead of `id`, per suggestions from feedback.

rwjblue:
:wave: We discussed this in today's core team meeting. Overall we are 👍 on the direction here, but would like to see a couple of (hopefully) small tweaks:

* Remove the `for=` API - This is mentioned in the alternatives section, we think that should be the default and we should not provide both forms of the API.
* Remove reference to `guidFor` - I think this largely falls out from removing the `for=` API since I _believe_ that is the only part that requires the concept of "object based stability".

Thanks again for pushing on this @steveszc, hopefully the last round of tweaks 🤞...

steveszc:
@rwjblue Thanks for taking the time to review with the core team! 

> 👋 We discussed this in today's core team meeting. Overall we are 👍 on the direction here, but would like to see a couple of (hopefully) small tweaks:
> 
>     * Remove the `for=` API - This is mentioned in the alternatives section, we think that should be the default and we should not provide both forms of the API.

If the default/only invocation option is the argument-less form, then I see two possible APIs that we'll need to choose between.

1. `{{unique-id}}` is a helper that returns a new unique value upon every invocation. This necessitates the usage of `let` blocks for re-use of an id value within a template.
2. `{{unique-id}}` is a "keyword" that returns the same value every time it is used within a template. This would be accomplished via an AST transform, as outlined in the alternatives section.

Did the core team meeting discussion assume that one of these two options would be the preferred API? It seems to me that this RFC will need to articulate which of those two APIs is being proposed.

text/0659-unique-id-helper.md
```diff
@@ -0,0 +1,196 @@
...
+
+See [pre-RFC issue #612](https://github.com/emberjs/rfcs/issues/612)
+
+## Motivation
+
+When working with HTML it is very common to need to create and reference [DOM IDs that are unique within the HTML document](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id). Classic Ember components provide the `elementId` attribute which can be used to construct unique ids within classic components, but `elementId` is not available within Glimmer components or route templates.
+
+There are several common use cases where a developer may need to generate a unique ID for use in a template:
+1. Associating `label` and `input` elements using the label's `for` attribute and the input's `id` attribute.
+2. Using WAI-ARIA attributes to improve accessibility (eg. aria-labelledby, aria-controls)
```
MelSumner, line 20:
```suggestion
2. Using WAI-ARIA attributes to improve accessibility (eg. `aria-labelledby`, `aria-controls`)
```

MelSumner:
IIRC, we want to be explicit in the RFC that use of the helper will return a unique ID on every invocation, and that we were okay with `let`.  There was some discussion about making it a keyword but that seemed more difficult and like it would have a non-zero impact on template imports. 

@steveszc does that help?

sandstrom:
Maybe I've missed something, but if all this helper does is return a unique id, why does it need to be in Ember? 

```js
export default buildHelper((_args, _options) => {
  return Math.round((Math.random() * 10 ** 20)).toString(36);
});
```

I thought the original idea was to make id generation for form id/for pairs easier. That would have been useful! (I'm still for that idea)

Now it's just a vanilla JS function wrapped in an Ember helper, not too different from `Ember.merge`, `Ember.k` and other "small helpers". If that's all we're doing here, isn't it better to start out with an addon and see if anyone starts using it?

rwjblue:
@steveszc 

> {{unique-id}} is a helper that returns a new unique value upon every invocation. This necessitates the usage of let blocks for re-use of an id value within a template.

Yes, this is the form we want to go with.

rwjblue:
> Maybe I've missed something, but if all this helper does is return a unique id, why does it need to be in Ember core?

Good question! Ultimately, it does **not** need any privileged access or other things that are only available internally (and therefore "need" to be in internals). The reason it is critical that we ship this by default (in one form or another) is that it is imperative that it be available as part of the main Ember programming model. I _personally_ don't care if that is via an addon that is included in the default stack (and it may well turn into that when we can actually use template imports; we could add a thing like `@ember/forms` or something) or if it is directly included in the emberjs/ember.js repo. Unless I've missed something, this RFC also doesn't specifically care about that either.

sandstrom:
@rwjblue Alright, I can see some benefits in shipping even the weaker version, but I would have preferred one that did more than simply generate a random string. 

But it's no biggie 😄

steveszc:
@rwjblue @MelSumner I believe the core team feedback has now been addressed.

MelSumner:
We have discussed this in today's Framework Core Team meeting, and decided to move this RFC to Final Comment Period.

hergaiety:
_Just voicing support_ for this RFC and the proposed helper. It'll be incredibly useful in the apps I work on every day and I'm glad to see it coming to the core framework. Thanks to everyone involved in getting it to this stage. :tada:

rwjblue:
One thing we have to make sure to do is that when we land this feature that we explicitly add tests for the "real world snippets" that this RFC mentions (to guarantee that they work as specified), but I guess that we all assumed that anyways....

rwjblue:
We chatted about this again in todays core team meeting, and are 👍 on moving forward!

chriskrycho:
Does an installable polyfill for this exist yet?

ctjhoa:
@chriskrycho 
I've published one
https://www.npmjs.com/package/ember-unique-id-helper-polyfill

bartocc:
What is the implementation status of this RFC in the framework?

NullVoxPopuli:
> What is the implementation status of this RFC in the framework?

unique-id landed in 4.4: https://github.com/emberjs/ember.js/releases/tag/v4.4.0

bartocc:
Thx a lot for the update @NullVoxPopuli 

I always find it hard to find where/when a RFC has made it to the framework…

For example, in the [actual rendered version of this RFC](https://github.com/steveszc/rfcs/blob/020194e64a8331d46686b0a2316fa5c68b23a0f8/text/0659-unique-id-helper.md) , I can see the following at the top

Capture d’écran 2022-06-13 à 10 30 16

Is this the right place to search for this info?

MrChocolatine:
@bartocc 
Personally I keep an eye on these 2 URLs:

- https://github.com/emberjs/ember.js/releases (subscribing to the repo helps a lot - [Watch menu > Custom > Releases](https://user-images.githubusercontent.com/47531779/173314381-f658a846-b86f-472e-8989-3a53601f7727.png))
- https://blog.emberjs.com

(viva la chocolatine :v:)

bartocc:
Thx @MrChocolatine , I do follow the blog and I had missed the new feature in 4.4, but the releases subscription is a great tool 👍

> (viva la chocolatine ✌️)

😆

PR #663, opened 2020-09-26
Update the README add RFC Stages

pzuraq:
Updates the process and templates to match [RFC 617](https://github.com/emberjs/rfcs/blob/master/text/0617-rfc-stages.md) (RFC Stages)

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,7 +82,8 @@ If that period is successful, with no new unaddressed concerns raised, the RFC
 is merged and is now in the [Accepted] stage.
 
 An RFC that does not achieve consensus to be accepted may enter a
-[Final Comment Period](#final-comment-periods-fcp) to [Close] the RFC.
+[Final Comment Period](#final-comment-periods-fcp) to move the RFC to the 
+[Closed] stage.
 
 ### Merged Proposals
 
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ of the community.
 
 --- 
 ## Quick links for Pull Requests for proposed and advancing RFCs
-[Quick links for Pull Requests for proposed and advancing RFCs]: #quick-links
+[Quick links for Pull Requests for proposed and advancing RFCs]: #quick-links-for-pull-requests-for-proposed-and-advancing-rfcs
 
 - [Proposed RFCs](https://github.com/emberjs/rfcs/pulls?q=is%3Apr+is%3Aopen+sort%3Aupdated-desc+label%3AS-Proposed)
 - [Exploring RFCs](https://github.com/emberjs/rfcs/pulls?q=is%3Apr+is%3Aopen+sort%3Aupdated-desc+label%3AS-Exploring)
@@ -39,7 +39,7 @@ This site shows the current state and text of all RFCs in Accepted or later stag
 [Table of Contents]: #table-of-contents
 
 - [Introduction](#ember-rfcs)
-- [Quick links](#quick-links)
+- [Quick links](#quick-links-for-pull-requests-for-proposed-and-advancing-rfcs)
 - [Table of Contents]
 - [How it works](#how-it-works)
 - [Stages]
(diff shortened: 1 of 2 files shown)


kategengler:
Thanks! We shouldn't merge this until we are ready to switch to RFC Stages.

README.md
```diff
@@ -79,26 +77,208 @@ add a link to the PR in the header of the RFC.
...
+### FCP to close
+
+An RFC may be closed or discontinued by the [core teams] after public discussion 
+has settled and comments have been made summarizing the rationale for closing. 
+The RFC will enter a "final comment period to close" lasting 7 days. At the end 
+of the "FCP to close" period, the PR will be closed.
+
+An RFC author may withdraw their own RFC by closing it themselves.
+
+## The RFC Stages
```
jenweber, line 111: Just for ease of community use/understanding, what do you think about moving the table above some of the other topics in the README?

jenweber (approved):
One thing that appears to be missing is the info about links to tracking things. I am not sure if that got lost in the shuffle/editing of the RFC itself, but for PRs that are merged, there should be some kind of tracking link that could go to a project board, Issue, PR, etc, just not necessarily the "RFCs Tracking" repo. The RFC has references to such a label that should be added as work progresses.

Otherwise looks great!

kategengler:
@jenweber I don't think this is meant to be a comprehensive implementation of the RFC

wagenet (approved):

wagenet:
I think this is a pretty crucial change since one major source of emo from people has been the claim that Ember is starting new features before properly wrapping up old ones. This would codify the process for that wrap-up and either make it clear that earlier features are wrapped up, or if they aren't, make it clear what work is necessary.

I'm starting a big push to grease the wheels of RFCs and PRs in general and I'm 100% on board with this.

README.md
```diff
@@ -1,20 +1,48 @@
...
 
-The "RFC" (request for comments) process is intended to provide a
-consistent and controlled path for new features to enter the framework.
+The process is intended to provide a consistent and controlled path for new
+features and changes to enter the framework. RFCs can be opened by any member
+of the community.
 
 [Active RFC List](https://github.com/emberjs/rfcs/pulls)
 
 [Approved RFC List](https://emberjs.github.io/rfcs/)
```
wagenet, line 19: These links will need to be changed since PRs are no longer the best way to check things. Also merged PRs are not necessarily approved after Staging is implemented. We should probably point to rfcs.emberjs.com.

wagenet (changes requested):
Looks good to me except the one issue I called out. We probably need to make sure to coordinate this with a blog post and we may also need rfcs.emberjs.com to be ready as well.

README.md
```diff
@@ -42,69 +70,256 @@ If you submit a pull request to implement a new feature without going
...
 feedback from the larger community, and the author should be prepared
-to revise it in response.
-* Find a champion on the relevant core team. The champion is responsible for 
-shepherding the RFC through the RFC process and representing it in core team 
+to revise it in response. The RFC is now in the [Proposed] stage.
+* Find a champion on the relevant core team. The champion is responsible for
+shepherding the RFC through the RFC process and representing it in core team
 meetings.
-* Update the pull request to add the number of the PR to the filename and 
+* Update the pull request to add the number of the PR to the filename and
```
wagenet, line 110: Do we still want this? This number is going to grow a lot more quickly if we have PRs to update RFCs.

kategengler: Would we stop numbering RFCs then? 

It shouldn't be any more onerous for the numbers growing quickly, but it is annoying that it needs to be added after opening the PR. RFCs keep the # from their initial PR to 'Accepted'

wagenet: I don’t know whether we should stop numbering or not, but there’s an already incorrect impression that RFC-800 means that there were 800 RFCs that were at least proposed by that point. This only gets worse with more actual PRs now. We could just pick the number sequentially when we actually merge. Possibly could even do this with automation. Not a big deal either way and not a blocker.

README.md
kategengler: I addressed this in the latest commit. lmk if you think that is enough.

wagenet: Looks great.

wagenet (approved):

PR #671, opened 2020-10-02
Stop Leaking Implementation Details of Built-in Components

chancancode:
[Rendered](https://github.com/emberjs/rfcs/blob/modernizing-built-in-components/text/0671-modernize-built-in-components-1.md)

### DIFF
--- a/text/0671-modernize-built-in-components-1.md
+++ b/text/0671-modernize-built-in-components-1.md
@@ -325,6 +325,327 @@ and tools alike, we do not intend to "reserve" the built-in component names, so
 if an app provides a component with the same name, they will take precedence
 over the built-in components.
 
+## Transition Path
+
+### Importing and Accessing Globals For Subclassing
+
+Example deprecation message:
+
+```
+Using Ember.Checkbox or importing from '@ember/component/checkbox' has been
+deprecated, install the `ember-legacy-built-in-components` addon and use
+`import { Checkbox } from 'ember-legacy-built-in-components';` instead.
+```
+
+* * *
+
+Example that causes deprecation:
+
+```js
+import Checkbox from '@ember/component/checkbox';
+//                   ~~~~~~~~~~~~~~~~~~~~~~~~~~~
+// Using Ember.Checkbox or importing from '@ember/component/checkbox' has been
+// deprecated, install the `ember-legacy-built-in-components` addon and use
+// `import { Checkbox } from 'ember-legacy-built-in-components';` instead.
+
+export class MyCheckbox extends Checkbox {
+  // ...
+}
+```
+
+Fixed example:
+
+```js
+import { Checkbox } from '@ember/legacy-built-in-components';
+
+export class MyCheckbox extends Checkbox {
+  // ...
+}
+```
+* * *
+
+Alternative example that triggers the deprecation:
+
+```js
+export function initialize(owner) {
+  owner.register(
+    'component:my-checkbox',
+    Ember.Checkbox.extend({ /* ... */ })
+//  ~~~~~~~~~~~~~~
+// Using Ember.Checkbox or importing from '@ember/component/checkbox' has been
+// deprecated, install the `ember-legacy-built-in-components` addon and use
+// `import { Checkbox } from 'ember-legacy-built-in-components';` instead.
+  );
+}
+```
+
+Fixed example:
+
+```js
+import { Checkbox } from '@ember/legacy-built-in-components';
+
+export function initialize(owner) {
+  owner.register(
+    'component:my-checkbox',
+    Checkbox.extend({ /* ... */ })
+  );
+}
+```
+
+* * *
+
+Additional prose in the deprecation guide:
+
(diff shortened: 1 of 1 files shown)


NullVoxPopuli (approved):

text/0671-modernize-built-in-components-1.md
```diff
@@ -0,0 +1,386 @@
...
+
+  Our preference would therefore be to pare down the existing components to
+  address the problems raised here. In the meantime, we will work on providing
+  primitives and capabilities that underpins these built-in components (such as
+  [router helpers](0391-router-helpers.md)), so the community can experiment
+  with alternatives and new patterns.
+
+  If and when they mature to the point of becoming widely accepted as best
+  practices in the ecosystem, we can revisit deprecating and removing any of
+  these built-in components as they become redundant. Braodly speaking, we do
```
alexlafroscia, line 380:
```suggestion
  these built-in components as they become redundant. Broadly speaking, we do
```

text/0671-modernize-built-in-components-1.md
```diff
@@ -0,0 +1,386 @@
...
+  By deprecating and removing individual features, as proposed in this plan, we
+  will be able to provide more directly actionable guidance in each case.
+
+* We can remove the built-in components altogether. This is a much bigger lift
+  that would require analyising all existing use cases and likely requires
+  designing new capabilities and features to fill some of those gaps.
+
+  Other than the problems outlined in this RFC, the built-in components mostly
+  gets the job done adequately, and we feel like there is not a lot of value in
+  taking on the project to reimagine them signaficantly and force the entire
```
alexlafroscia, line 369:
```suggestion
  taking on the project to reimagine them significantly and force the entire
```

text/0671-modernize-built-in-components-1.md
```diff
@@ -0,0 +1,386 @@
...
+
+  By deprecating and removing individual features, as proposed in this plan, we
+  will be able to provide more directly actionable guidance in each case.
+
+* We can remove the built-in components altogether. This is a much bigger lift
+  that would require analyising all existing use cases and likely requires
+  designing new capabilities and features to fill some of those gaps.
+
+  Other than the problems outlined in this RFC, the built-in components mostly
+  gets the job done adequately, and we feel like there is not a lot of value in
```
alexlafroscia, line 368:
```suggestion
  get the job done adequately, and we feel like there is not a lot of value in
```

alexlafroscia (commented):
Overall, I think it's a great idea to limit how these components can be interacted with and to avoid introducing a new "generation" of these built-in components.

With the exception of `LinkTo`, it's unclear to my what the benefit of even having these components as part of the Ember API rather than using the raw HTML elements directly, so I'm totally in agreement with not introducing new versions of them! I hope we deprecate them entirely one day, but can see how it's not worth the "churn" to do that right now.

chancancode:
> With the exception of `LinkTo`, it's unclear to my what the benefit of even having these components as part of the Ember API rather than using the raw HTML elements directly

I don't want this RFC to imply that it is the core teams/community's collective opinion that these components are obsolete/out-of-favor. I personally still use them and definitely find them generally useful and pleasant to use. Even after this RFC, they are still very part of the Ember first-class API, and no one should hesitate to use them, even when building new apps! We'll continue to teach them as the happy path of the learning materials too. 😄

That being said, it's totally fine and healthy to have a somewhat different opinion here, and I think eventually we will converge anyway – I don't necessarily think we will stick with this model forever. We should make sure we don't get in the way of making those alternative/experiments viable, and if there is anything that you need to make things better/easier, we should work it out to and make it happen (in separate proposals than this).

On the off chance that something here would make that situation worse compared to the status quo, definitely bring it up here!

text/0671-modernize-built-in-components-1.md
```diff
@@ -0,0 +1,386 @@
...
+  [router helpers](0391-router-helpers.md)), so the community can experiment
+  with alternatives and new patterns.
+
+  If and when they mature to the point of becoming widely accepted as best
+  practices in the ecosystem, we can revisit deprecating and removing any of
+  these built-in components as they become redundant. Broadly speaking, we do
+  not believe we are at that point yet, so in the meantime, we should not stop
+  improving what we have.
+
+## Unresolved questions
```
chancancode, line 384: Should `TextSupport` be available for import in the legacy addon?

chancancode:
We discussed this in today's framework team meeting, and we believe this proposal is ready for Final Comment Period. If you haven't had a chance to review it yet, this would be a good time to do it! 🎉

rwjblue (approved):
Awesome, thank you for writing this up @chancancode!

rwjblue:
Discussed this again yesterday, and we are all in favor of moving forward....

✋ 👍 ✋ 👍

st-h:
rendered link is broken. This seems to be the right one:
https://github.com/emberjs/rfcs/blob/master/text/0671-modernize-built-in-components-1.md

Issue #660, opened 2020-09-08
{{on}} modifier should not add event listeners if the passed function is undefined


NullVoxPopuli:
the On modifier's addEventListener helper function: https://github.com/emberjs/ember.js/blob/a493c6a840ae6e56a612237c2afa495ed2cdfd50/packages/%40ember/-internals/glimmer/lib/modifiers/on.ts#L211

_always_ adds an event listener (here: https://github.com/emberjs/ember.js/blob/a493c6a840ae6e56a612237c2afa495ed2cdfd50/packages/%40ember/-internals/glimmer/lib/modifiers/on.ts#L392 )

I think it'd help out _a lot_ of folks to make this change to `update`:

https://github.com/emberjs/ember.js/blob/a493c6a840ae6e56a612237c2afa495ed2cdfd50/packages/%40ember/-internals/glimmer/lib/modifiers/on.ts#L374

```diff
    // use prior state values for removal
    removeEventListener(element, eventName, callback, options);

    // read updated values from the state object
-   addEventListener(state.element, state.eventName, state.callback, state.options);
+   if (state.callback) {
+     addEventListener(state.element, state.eventName, state.callback, state.options);
+    }
```

buschtoens:
I already suggested that in the past and the recommendation was to conditionally apply the modifier once the RFC that allows passing around modifiers lands. 

I don't have the links at hand right now, because I'm on mobile.

NullVoxPopuli:
I think that could def help, but I believe it to be mostly orthogonal to this issue?

buschtoens:
I agree, but IIRC the line of reasoning was that not calling the modifier in the first place is more efficient then calling it and have it effectively no-op.

Wit the other RFC in place, you could basically use an `{{if}}` expression in place of the modifier:

```hbs
<button {{if @handler (modifier "on" "click" @handler)}}>
```

jelhan:
Wondering if this should be considered a bug? Attaching an event listener with a listener, which is neither a function nor implements `EventListener` interface is a bug. The `{{on}}` modifier should prevent this.

But it's another question if it should support `undefinded` or any other value, which isn't a valid listener. Maybe it should just fail early with a helpful error message? To be honest I'm tending towards that solution.

Silently not applying an event listener is error prune and hard to debug. I can imagine myself spending hours until I notice that there is a typo. 🙈

[...]

PR #676, opened 2020-10-20
Return noop function from {{fn}} helper if invoked without arguments

jelhan:
[Rendered](https://github.com/jelhan/rfcs/blob/argumentless-fn-helper-returns-noop-function/text/0676-argumentless-fn-helper-returns-noop-function.md)

Some preliminary discussion has happened in #660.

### DIFF
--- a/text/0000-argumentless-fn-helper-returns-noop-function.md
+++ b/text/0000-argumentless-fn-helper-returns-noop-function.md
@@ -1,0 +1,278 @@
+- Start Date: (fill me in with today's date, YYYY-MM-DD)
+- Relevant Team(s): (fill this in with the [team(s)](README.md#relevant-teams) to which this RFC applies)
+- RFC PR: (after opening the RFC PR, update this with a link to it and update the file name)
+- Tracking: (leave this empty)
+
+# Return noop function if {{fn}} helper is invoked without arguments
+
+## Summary
+
+This RFC introduces an argumentless version of the `{{fn}}` template helper.
+It should return a noop function.
+
+The RFC does not intend to change the existing meaning of the `{{fn}}` template
+helper if invoked with one or more arguments. Especially it does *not* propose
+to support `undefined` as value for the first positional argument. The `{{fn}}`
+template helper should continue to throw an assertion if invoked with another
+value than a function as the first positional argument.
+
+## Motivation
+
+This RFC is driven by two motivations:
+
+1. Provide a built-in way to get a noop function in a template.
+2. Simplify mental model around `{{fn}}` helper.
+
+### Use cases for noop functions in templates
+
+Noop function are commonly used in modern Ember. Currenlty the framework misses
+a built-in way to get one in a template. Developers need to either reach out
+for a userland tempalte helper like the [`{{noop}}` helper from
+ember-composable-helpers](https://github.com/DockYard/ember-composable-helpers#noop)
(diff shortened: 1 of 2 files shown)


text/0000-argumentless-fn-helper-returns-noop-function.md
```diff
@@ -0,0 +1,278 @@
...
+package or expose a noop function from backing JavaScript class.
+
+A noop function is mainly required in components that have an optional
+argument, which should be either passed down to another component as an action
+handler or used with an event listener registered through `{{on}}` modifier.
+The value of the optional argument is `undefined` if consumer does not set it
+but the component may and the `{{on}}` modifier do not support `undefined` but
+require a function.
+
+Let's have a look at two example use cases to illustrate the issue:
```
jrjohnson, line 41: Another place I frequently ready for `noop` is in integration tests when invoking a component with a passed action that isn't needed in the context of the specific test:
```
await render(hbs`<VocabulariesExpanded
  @vocabularies={{this.vocabularies}}
  @collapse={{noop}}
/>`);
```

jelhan: I get what you mean. But I also experienced that there is value in being more explicit about if the function should be called or if not. You may miss some bugs if you don't assert that a function is *not* executed if it should not be executed.

jrjohnson: Makes sense. I tent to treat that as a unique unit in an isolated test. Either way, I like empty `fn` a lot.

jelhan:
I added harmonization of public APIs of `{{array}}`, `{{hash}}` and `{{fn}}` helper as another motivation. I had lost that point in a late iteration on the draft before publishing it. Additionally I added providing a `{{noop}}` helper as an alternative. This was suggested by @kategengler in Discord.

amk221:
+1 for this.
The ember template lint rule would have to be updated? Because you'd get `Ambiguous path 'fn' is not allowed`

jelhan:
Added teaching section in 0fb24acb3490bb6eac286611fe7575ae76a3b4e3. Thanks to  @locks for pointing out that it's missing. To be honest I assumed that I had written it.

jelhan:
`lint-frontmatter` and `check-filename` CI jobs are failing with an infrastructure issue, which doesn't seem to be related to this merge request:

> Error: The head commit for this pull_request event is not ahead of the base commit. Please submit an issue on this action's GitHub repo.

pzuraq:
We discussed this in the core team meeting today, and the outcome was that with dynamic modifiers landing imminently, one of the motivating use cases for this feature, using `{{on}}` dynamically, will be solved in a way that is potentially better overall:

```js
<button {{(if @onClick on) "click" @onClick}}>
  {{yield}}
</button>
```

This is similar to the method that was mentioned in the RFC, but without the need for additional syntax (i.e. defining the semantics of `if` in modifier position). We also realized there were other motivating examples, and the consensus was that it would be good to flesh these out a bit more and also take some time to reflect on the new syntax and whether or not `{{fn}}` defaulting to no-op continues to make sense now that we (almost!) have it.

wagenet:
Per @pzuraq's comment it sounds like this may not be needed anymore. Also template imports will change this some too. If I'm mistaken, please let me know and I can reopen it and work to get it moving forwards again.

PR #678, opened 2020-10-27
Documenting Components

gossi:
[Rendered](https://github.com/gossi/rfcs/blob/documenting-components/text/0678-documenting-components.md)

### DIFF
--- a/text/0678-documenting-components.md
+++ b/text/0678-documenting-components.md
@@ -1,6 +1,6 @@
 - Start Date: 2020-10-27
 - Relevant Team(s): Ember.js, Learning (?)
-- RFC PR: (after opening the RFC PR, update this with a link to it and update the file name)
+- RFC PR: [#678](https://github.com/emberjs/rfcs/pull/678)
 - Tracking: (leave this empty)
 
 # Documenting Components
--- a/text/0000-documenting-components.md
+++ b/text/0000-documenting-components.md
@@ -1,0 +1,326 @@
+- Start Date: 2020-10-27
+- Relevant Team(s): Ember.js, Learning (?)
+- RFC PR: (after opening the RFC PR, update this with a link to it and update the file name)
+- Tracking: (leave this empty)
+
+# Documenting Components
+
+## Summary
+
+This RFC describes a way to document components in a way to connect it to tooling from the wider javascript ecosystem. The scope of this RFC is to cover documentation for **arguments**, **named blocks** and **yield** for components written in JavaScript, TypeScript and template-only.
+
+## Motivation
+
+The current preferred way to document your components is probably using [ember-cli-addon-docs](https://github.com/ember-learn/ember-cli-addon-docs), trying to use [storybook](https://storybook.js.org/), _older_ ember addons (such as [ember-freestyle](https://github.com/chrislopresto/ember-freestyle)), one of the [empress](https://github.com/empress) products or running your own system. The way these libraries extract information from your code is mostly in a _proprietary_ format (eg. `ember-cli-addon-docs` maintains its own standard and system of generating such documentation). For `ember-cli-addon-docs` this has proven to be volatile and maintenance is part of the respective projects. 
+
+The idea is to find a documentation standard that connects with the rest of javascript ecosystem rather than building one on its own and is recommended for usage within the ember community.
+
+### Example: `<Modal>` Component
+
+Let's start with an example to document a **template-only** `<Modal>` component with arguments and optional named blocks (and their yielded properties):
+
+```ts
+// addon/components/modal/index.d.ts
+import Component, { Template } from '@glimmer/component';
+
+export interface ModalArgs {
+  /** Trigger the visibility of the modal */
+  open: boolean;
+  /** Callback when the modal was opened */
+  opened: () => void;
+  /** Callback when the modal was closed */
+  closed: () => void;
+}
+
+export interface ModalBlocks {
+  default: [close: () => void];
+  header: [close: () => void];
+  body: [];
+  footer?: [];
+}
+
+export default class ModalComponent extends Component<ModalArgs, ModalBlocks> {}
+```
+
+which in turn we can use in these two variants (with and without named blocks):
+
+```hbs
+<Modal as |close|>
+  Here is the content of the modal
+  
+  <button {{on "click" close}}>close</button>
+</Modal>
+
+<Modal @open={{true}}>
+  <:header as |close|>
+    This is my title
+    
+    <button {{on "click" close}}>close</button>  
+  </:header>
+  
+  <:body>
+    Here is the content of the modal
+  </:body>
+</Modal>
+```
+
+## Detailed design
+
+The design is sectioned into _definitions_, _documentation API_, _examples_ and then summarized with _benefits_.
+
+### Definitions
+
+At first let's define terms that will be used throughout this RFC:
+
+- **Block**
+  For any block invocation of a component, this describes the contents in there. We can distinguish between _default_ block and named blocks.
+  This is defined in syntax section for [RFC 460 - Yieldable Named Blocks](https://emberjs.github.io/rfcs/0460-yieldable-named-blocks.html#syntax)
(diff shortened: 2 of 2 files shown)


jelhan (commented):
I'm very excited by this RFC! Thanks a lot for writing it.

Settling on a shared standard to document components API will be very helpful for the community in my opinion. I expect it to be adopted by existing and new tools quickly as typed templates, improved autocompletion in editiors and automatically generated API docs will improve developer experience a lot.

knownasilya:
cc @lifeart

text/0678-documenting-components.md
```diff
@@ -0,0 +1,326 @@
...
+  ```hbs
+  <article>
+    <header>{{yield @article.title to='header'}}</header>
+    <section>{{yield @article.body to='body'}}</section>
+  </article>
+  ```
+  
+### Documentation API
+
+Documentation will use the powerful typing system of TypeScript. For TypeScript this works out of the box. For template-only components, use a `my-component.d.ts` file. For JavaScript, you can choose between a `*.d.ts` file or jsdoc with extended tsdoc annotations (see examples below).
```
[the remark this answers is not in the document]
gossi: I think, there are two things to that:

1. Typescript language features as constructs for structured data
2. The doc dialect

Ad 1)

The goal is to reach for structured data and typescript is helping here. For non-native-typescript support, TS offers a dialect addition for (2), see here: https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html#typedef-callback-and-param (which is required to "upgrade" js to understand ts language features).

Ad 2)

The doc dialect is not part of this RFC. I believe this is something the community will settle on. There exists [tsdoc](https://github.com/microsoft/tsdoc), [api-documentor](https://api-extractor.com/pages/tsdoc/doc_comment_syntax/) (based on tsdoc) and [jsdoc](https://jsdoc.app/).
Apart from the requirements for js under (1), I think you are free to choose. At least from this RFC perspective. I have a feeling addons will rise for this ;)

pzuraq:
We discussed this RFC at the core team meeting on Friday, and were generally in favor of this direction in the long run. Using TypeScript and type systems in general to generate API documentation seems like a great innovation, and something we would like to support in Ember.

The issue lies in Ember also wanting to actually support TypeScript in the near future. The Typed Ember team is currently building out support for typed templates, and we're getting closer to being able to add official support in Ember itself, and to ship types with Ember. In order to do this, we will need to nail down the types for arguments, blocks, and attributes, and the chief concern here is that if we were to jump ahead and standardize types for _documentation_ purposes alone, we may end up shipping types that conflict with actual TS support later on.

We are already in this situation a bit with `@glimmer/component`. Since Glimmer.js itself shipped with types in v1, Glimmer components did as well, and the typed-Ember community has built up around those types. Changing this would already introduce breakage, but expanding the scope of those types could introduce even more, so for the time being, until the larger TS story is figured out, we don't think it would be a good idea to expand the types for documentation purposes.

All that said, we do still think that this RFC could advance, if it were to shift focus a bit.

All components in Ember _do_ a public API, which includes:

1. Arguments, both named and positional
2. Blocks, which are like callbacks that receive arguments (block params)
3. Attributes, and the element they are applied to

Today, we don't really teach this as the public API in any consolidated location, and we don't guide users to document these things at all. It would be good for us to add this definition, along with a detailed explanation of each, to the guides. 

We could also add some unopinionated documentation tooling that does not rely on types for the time being. For instance, we could add a generator that creates a `my-component.md` file, which includes a description of the args, blocks, and attributes/element that the component exposes. This would start us toward a more unified documentation story in general, and when the time comes to actually ship types with Ember, we would be a better place to ship types and tooling for components.

text/0678-documenting-components.md
```diff
@@ -0,0 +1,326 @@
...
+
+The interface for `@glimmer/component` needs extension for yields:
+
+```ts
+export default MyComponent extends Component<Input, Output> {}
+```
+
+The narrative is to give developers documentation slots for `Input` and `Output`. The `Input` will continue to work as is to document arguments and `Output` will be added to document blocks with their parameters.
```
pzuraq, line 136: I'm not sure that `Input` and `Output` is the correct framing here. Blocks aren't exactly outputs, or at least, they are not the _sole_ outputs of the component. The entire template within the component is output.

Blocks are more like _callbacks_ in JavaScript. For instance:

```js
this.store.fetchRecord('foo', 123).then(() => {
  // do things...
});
```  

The callback passed to this promise is similar to a block. It gets passed as an argument, and internally it gets called later on at some point, and its result is _slotted_ into a specific place. This is basically what blocks do as well.

To me, blocks are just a different form of input, namespaced so that they cannot collide with arguments.

lukemelia:
In evaluating an approach to documentation, it would be valuable to consider the path to expose documentation to javascript at runtime so that addons can use it to visualize the docs. We're currently giving ember-freestyle the ability to document components with interactive controls.

Here's the PR, which includes a short video of the feature in action: https://github.com/chrislopresto/ember-freestyle/pull/457

Also worth noting: we've been co-locating the file defining the freestyle example alongside the component file with the name `usage.hbs`. This pattern is working out pretty nicely so far.

It would great to be able to a) automatically consume the component documentation in the usage file to create the interactive widget without repeating the documentation, and/or b) have a generator that could generate a base usage.hbs for a component based on the docs that is ready for customization.

pzuraq:
@lukemelia so this is not necessarily the core team's view, but personally I really have liked the potential of Storybook.js for doing exactly that. I built a few proof of concepts about a year ago that exposed demo components using the [controls](https://storybook.js.org/docs/ember/essentials/controls) addon from Storybook, which were really great and easy to use.

Because Storybook is super pluggable too, it would be possible to write a plugin that reads TS types of JSDoc comments and creates a tab with API documentation for the component. This is also part of the motivation behind the recent [Argument Validation Primitives](https://github.com/emberjs/rfcs/pull/694) RFC, the thinking being that if we can make a statically analyzable high level format, then you could parse that straight into documentation via an addon. In addition, I created a `.glimmer.md` style file format which could be used to make interactive guides-style documentation, similar to what is available with ember-cli-addon-docs. This was all in an afternoon of hacking, so there was a very small amount of investment, with a very large amount of return.

The key issues are around how Ember integrates with Storybook currently, there are a number of issues (for instance, you have to boot up a whole Ember app for each example). But the changes we're working on for strict mode are bring us closer to being able to integrate very smoothly with it.

josemarluedke:
I'm really looking forward to being able to further extract information from Components to generate documentation. Today that's already possible when authoring in TypeScript (for components args), and the addition of [Argument Validation Primitives](https://github.com/emberjs/rfcs/pull/694) would allow extracting that when authoring in JS as well. What is missing is the story for blocks, and that's where I think this RFC helps a lot.

I have been working on a set of documentation tools and one of them is [glimmer-docgen-typescript](https://github.com/josemarluedke/glimmer-docgen-typescript). This tool uses TypeScript to figure out all the arguments in a Glimmer Component. This tool could be modified to allow for blocks as well on that story if defined.

Additionally, I'm very much looking to the day we have typed templates, so if this RFC is a dependency for that, I would love to get it right to avoid breaking changes.


Note: You can take a look at [Frontile](https://frontile.dev/)'s documentation to see glimmer-docgen-typescript in action.
On a side note about demos in markdown, checkout out [Docfy](https://docfy.dev/) for Ember ([Writing Demos](https://docfy.dev/docs/ember/writing-demos)).

text/0678-documenting-components.md
gossi: This mental model preceeds blocks. Not accurate any longer. Agree, callbacks are more the mechanics that reflects this.

gossi:
Thanks @pzuraq for your answer. From what I see, this RFC is already including a way for blocks, but is missing on attributes and element(s) and additionally even enhanced with argument validation.

What's best to advance from here?

gossi:
Thanks everyone for the feedback. I gonna close this one as this has moved over to #748.

Issue #377, opened 2018-09-18
Deprecate defaultStore located at `Route.store`


runspired:
See (private) docs for this intimate API that often trolls users of `ember-engines` and `ember-data`.

https://emberjs.com/api/ember/3.4/classes/Route/methods/store?anchor=store&show=inherited%2Cprivate

## Motivations

* address learning concerns around where `ember` ends and `ember-data` begins, specifically:
  * clarify that `Route.model()` is not named model "because it returns ember-data models"
  * clarify that `Route.model()` is not 1:1 with an ember-data model and remove confusion about whether routes are allowed to have more than one model
  * remove confusion around folks thinking that routes (and dynamic params in routes) must be 1:1 with API endpoints
  * squash bugs where routes have multiple dynamic params attempt to rely on `store.find`
  * squash bugs where dynamic params cannot have the same name
  * demystify how the `model()` hook works and demystify how to request data with `ember-data`
  * allow `ember-data` to drop the long vestigial `store.find` method
  * remove confusion around why `store` is available but not the right store in `engines` when folks forget to wire up store
  * remove a common reason developers believe ember-data is required
  * remove a common reason developers believe ember-data is ember
* enable `ember-data` to kill the eager initializer and eager instance-initializer necessary to override `defaultStore` (prevents tree shaking, is a perf hit when apps don't need ember-data to render the entry route, and limits ember-data's ability to evolve the Store's APIs)
* move to a pattern of explicit `store` injection, instead of magical injection on `routes` and `controllers`
* enable folks to write unit and integration tests that register the store without needing to first unregister the store due to the eager initializer ( this is a silent and very annoying bug to encounter )
* prepare for a world in which partitioning data can be a first class route concern
* prepare for a world in which request management is first class instead of spooky action at a distance
* prepare for a world in which `ember-data` may be adopted incrementally
* trim the default size of ember, which includes significant infrastructure to support defaultStore

For folks that want this magical behavior, an optional addon that reopens route with either a "DefaultStore" mixin or an "ember-data route store" mixin would be more in keeping with being leaner and more explicit.

runspired:
@rwjblue had done some exploration on deprecating this, unsure where it stalled but this prevents `ember-data` from cleaning up various initializer patterns which hinder the ability for folks to write tests that register a store without first unregistering the store service.

rwjblue:
Just saw the default store mentioned in a video linked in the most recent Ember Weekly: https://youtu.be/EcKaDu0xo_A

[...]

PR #680, opened 2020-11-02
Deprecate Implicit Injection on arbitrary Ember Framework objects

### LINKED ISSUE #508: Pre-RFC: Deprecate implicit injections (owner.inject)
I'm defining "implicit injections" as the type of injection where nothing exists in the class body (or inherited from the parent class), but a property is magically present anyways. These types of implicit injects are nearly always done from initializers or instance-initializers (both of which should also die in a 🔥, but that is another issue). Examples of these implicit injections are using this.store from within an arbitrary `Route` / `Service` / `Controller` / `Component` without first specifying `store: injectService()`. 

Implicit injections have long been a source of confusion for new comers and old-timers alike. 

> Where does this.store come from?!

Since the advent of `Ember.inject.service` (from waaaaayyy back in Ember 1.9.0!) we have had a much better way to bring in external collaborating services like this. Explicit injections are massively easier to reason about since you can easily see that the class itself (or its ancestor) defines the injection.

As mentioned above the most used implicit injection is `this.store` (though there are others) which might indicate that this is "Ember Data's problem", but due to the nature of these implicit injections there is really no way that `ember-data` can actually deprecate usage of these implicit injections. If `ember-data` removes the initializer that sets up the implicit injections, it is clearly a breaking change. `ember-data` has no way to ensure that the `store` property emits a deprecation (but only if not "clobbered" by `store: injectService()`) which means that it can't even be removed in a major version bump due to Ember's unique SemVer commitment (to not remove any undeprecated public APIs in a major version bump).

We must provide a mechanism to deprecate these implicit injections wholesale. They are bad, they make our programs harder to understand, 🔥🔥🔥.

---

Concrete implementation suggestion:

* Add an argument to `owner.inject` that allows the caller to specify that the injection is deprecated, which version it is deprecated until, a deprecation `id`, and a `url` for reading more about the deprecation. (For those of you following along, I've just described the `options` argument to `@ember/debug`'s `deprecate` method :wink:)
* In debug builds, when this deprecation options argument is passed, Ember will automatically setup a getter for the deprecated injected property on the target object when the object is created.
* Sometime later, issue a deprecation when the `owner.inject` is **not** provided that new extra "deprecations" object.

The goal of this pre-RFC is:

1. To circulate the idea of pushing this forward
2. Find an author that has the time to work with me on the final detailed design and put together the deprecation RFC
3. Help with implementation of the RFC once landed

snewcomer:
[Rendered](https://github.com/snewcomer/rfcs/blob/sn/owner-inject-deprecation/text/0680-implicit-injection-deprecation.md)

close #508

### DIFF
--- a/text/0680-implicit-injection-deprecation.md
+++ b/text/0680-implicit-injection-deprecation.md
@@ -90,11 +90,17 @@ and setter on the target. Whenever the property is accessed, we will check if th
 we will issue a deprecation in DEBUG builds with an until, id and url value to read more about this deprecation. In production, we will
 not issue this deprecation and continue assigning the property.
 
+To avoid this deprecation, you will need to explicitly add the injected property to the target object.  As an example, you can add
+`@service store` to your route or controller to avert this deprecation.
+
 
 ### 2. Deprecate `owner.inject`
 
 The first phase did not actually deprecate the "use" of `owner.inject`.  As a result, we need to deprecate
 it's use directly before removing.
+
+For users and or addons, you may be able to remove `owner.inject` to avoid this deprecation.  However, for libraries like `@ember/data`, we will not be able to
+remove until `@ember/data` v4.
 
 ### 3. Profit!
 
--- a/text/0680-implicit-injection-deprecation.md
+++ b/text/0680-implicit-injection-deprecation.md
@@ -10,7 +10,7 @@
 This RFC seeks to deprecate implicit injections on arbitrary Ember Framework objects. This is commonly done via `owner.inject` in an
 intializer or instance-initializer.
 
-A prevalent example of implicit injection is found in `ember-data` [injecting](https://github.com/emberjs/data/blob/4bd2b327c4cbca831f9e9f8bc6b497200a212f9b/packages/-ember-data/addon/setup-container.js) their default `store` into all
+A prevalent example of implicit injection is found in `ember-data` [injecting](https://github.com/emberjs/data/blob/4bd2b327c4cbca831f9e9f8bc6b497200a212f9b/packages/-ember-data/addon/setup-container.js) their default `@ember-data/store` into all
 `@ember/routing/route` and `@ember/controller` factory objects. Here is a pruned example of how this looks.
 
 ```js
@@ -47,19 +47,19 @@ Implicit injections have long confused developers onboarding into Ember projects
 explicit injection visible on the class body, developers start to wonder how certain properties got there.
 We can infer from a common implicit injection in `ember-data` to reason about some of the downsides this presents.
 If a project has `ember-data` installed, an initializer at runtime registers the `@ember-data/store` on both
-Ember.Route and Ember.Controller factory objects.  This is a nice convenience for a project with many routes
+`@ember/routing/route` and `@ember/controller` factory objects.  This is a nice convenience for a project with many routes
 that are used for data fetching. However, it leads to many inconveniences that hinder learning and advancement
 of the framework. This includes:
 
 1. By entangling specific objects with the whole dependency tree, we might be preventing incremental adoption.
    For example, `ember-data` has laid out its plan for adopting only specific packages in [Project Trim](https://github.com/emberjs/data/issues/6166).
-   However, because `store` is injected by default on all Ember.Route and Ember.Controller factory objects,
+   However, because `@ember-data/store` is injected by default on all `@ember/routing/route` and `@ember/controller` factory objects,
    a user cannot easily opt out of `@ember-data/store`. This is a common problem space in other communities as well.
    With previous JetBrains IDE installations, adding plugins required a reload. However, by removing a contructor function
    that injected all the dependencies at once, they were able to avoid reload after installation of a plugin.
 
 2. Eager initialization of the dependency tree from which the property came from prevents tree shaking and incurs a
-   performance hit. For example, if you app does not need `ember-data` to render the entry route, the `store` injection will
+   performance hit. For example, if you app does not need `ember-data` to render the entry route, the `@ember-data/store` injection will
    take up more CPU cycles than necessary, hurting common user facing metrics.
 
 3. Eager initialization makes it hard to users to write tests that need to take advantage of a stub.
@@ -68,7 +68,7 @@ of the framework. This includes:
    ensuring instance initializers run before each test.
 
 4. Injection on commonly used framework objects can lead to deviation of project specific styleguides. For instance,
-   injecting the `router` on the Ember.Component factory object will certainly lead to instances where the `router` is explicitly
+   injecting the `router` on the `@ember/component` factory object will certainly lead to instances where the `router` is explicitly
    injected and other instances where is it implicitly injected.
 
 5. Users have come to associate `Route.model()` with a hook that returns `ember-data` models in the absense of explicit injection.
@@ -87,7 +87,9 @@ removing APIs without a deprecation path. As a result, it will take us 3 phases 
 The first step will be issuing a deprecation on the target factory object that is receiving a property. This will surface the deprecation
 on user's owned objects including transitively through addons like `ember-data`. This can be accomplished by installing a native getter
 and setter on the target. Whenever the property is accessed, we will check if the property is explicitly injected. If it isn't,
-we will issue a deprecation with an until value, id and url to read more about this deprecation.
+we will issue a deprecation in DEBUG builds with an until, id and url value to read more about this deprecation. In production, we will
+not issue this deprecation and continue assigning the property.
+
 
 ### 2. Deprecate `owner.inject`
 
@@ -106,7 +108,7 @@ changes to their codebase. We need to consider this as we move through each phas
 The API docs would need to be overhauld in a few spots. First, we need to remove the [docs](https://guides.emberjs.com/release/applications/dependency-injection/#toc_factory-injections) about Factory injections.
 Second, we need to detail explicit injection where currently relying on implicit injection.  Many examples show fetching data
 via `this.store` but do not specifiy how `this.store` arrived as a property on the Route. Also apparent
-disclaimers for people visiting the `ember-data` docs that explicit injection of `store` is necessary
+disclaimers for people visiting the `ember-data` docs that explicit injection of `@ember-data/store` is necessary
 and has deviated from past behaviour is likely prudent.
 
 The default blueprints when generating a new Ember application will not change.
@@ -114,7 +116,7 @@ The default blueprints when generating a new Ember application will not change.
 ## Alternatives
 
 - Continue with use of `owner.inject` but overhaul docs and recommend explicit injection.
-- Provide application level super class with a default store injection provided by `ember-data` and other common examples in the community.
+- Provide application level super class with a default `@ember-data/store` injection provided by `ember-data` and other common examples in the community.
 - Allow injection on a specific factory with a single backing Class instead of a generic factory object (e.g. `route` vs `route.index`).
 
 ## Open Questions
--- a/text/0680-implicit-injection-deprecation.md
+++ b/text/0680-implicit-injection-deprecation.md
@@ -1,7 +1,7 @@
 - Start Date: 2020-10-04
 - Relevant Team(s): Ember.js
-- RFC PR: (after opening the RFC PR, update this with a link to it and update the file name)
-- Tracking: (leave this empty)
+- RFC PR: https://github.com/emberjs/rfcs/pull/680
+- Tracking: https://github.com/emberjs/rfcs/issues/508
 
 # Implicit Injection Deprecation
 
--- a/text/0680-implicit-injection-deprecation.md
+++ b/text/0680-implicit-injection-deprecation.md
@@ -55,7 +55,7 @@ of the framework. This includes:
    For example, `ember-data` has laid out its plan for adopting only specific packages in [Project Trim](https://github.com/emberjs/data/issues/6166).
    However, because `@ember-data/store` is injected by default on all `@ember/routing/route` and `@ember/controller` factory objects,
    a user cannot easily opt out of `@ember-data/store`. This is a common problem space in other communities as well.
-   With previous JetBrains IDE installations, adding plugins required a reload. However, by removing a contructor function
+   With previous JetBrains IDE installations, adding plugins required a reload. However, by removing a constructor function
    that injected all the dependencies at once, they were able to avoid reload after installation of a plugin.
 
 2. Eager initialization of the dependency tree from which the property came from prevents tree shaking and incurs a
--- a/text/0680-implicit-injection-deprecation.md
+++ b/text/0680-implicit-injection-deprecation.md
@@ -29,7 +29,7 @@ export default {
 ```js
 export default class PostRoute extends Route {
   // This proposal seeks to make this service injection explicit
-  // Currently users do no need to specify this injection if ember-data is installed
+  // Currently users do not need to specify this injection if ember-data is installed
   // @service store;
 
   model() {
--- a/text/0680-implicit-injection-deprecation.md
+++ b/text/0680-implicit-injection-deprecation.md
@@ -11,7 +11,7 @@ This RFC seeks to deprecate implicit injections on arbitrary Ember Framework obj
 intializer or instance-initializer.
 
 A prevalent example of implicit injection is found in `ember-data` [injecting](https://github.com/emberjs/data/blob/4bd2b327c4cbca831f9e9f8bc6b497200a212f9b/packages/-ember-data/addon/setup-container.js) their default `store` into all
-Ember.Route and Ember.Controller factory objects. Here is a pruned example of how this looks.
+`@ember/routing/route` and `@ember/controller` factory objects. Here is a pruned example of how this looks.
 
 ```js
 // app/initializers/store-inject.js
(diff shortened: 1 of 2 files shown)


text/0680-implicit-injection-deprecation.md
```diff
@@ -0,0 +1,127 @@
+- Start Date: 2020-10-04
+- Relevant Team(s): Ember.js
+- RFC PR: (after opening the RFC PR, update this with a link to it and update the file name)
```
rwjblue, line 3: Can you update with the PR link?

text/0680-implicit-injection-deprecation.md
```diff
@@ -0,0 +1,127 @@
...
+
+# Implicit Injection Deprecation
+
+## Summary
+
+This RFC seeks to deprecate implicit injections on arbitrary Ember Framework objects. This is commonly done via `owner.inject` in an
+intializer or instance-initializer.
+
+A prevalent example of implicit injection is found in `ember-data` [injecting](https://github.com/emberjs/data/blob/4bd2b327c4cbca831f9e9f8bc6b497200a212f9b/packages/-ember-data/addon/setup-container.js) their default `store` into all
+Ember.Route and Ember.Controller factory objects. Here is a pruned example of how this looks.
```
rwjblue, line 14: Let's use the module names, not the older global names.
```suggestion
`@ember/routing/route` and `@ember/controller` factory objects. Here is a pruned example of how this looks.
```

text/0680-implicit-injection-deprecation.md
```diff
@@ -0,0 +1,127 @@
...
+  }
+}
+```
+
+Ensuring a property is present on Ember's Framework objects should be explicitly defined. This will allow the Ember ecosystem to
+further progress the framework while easing the learning curve for all developers.
+
+## Motivation
+
+Implicit injections have long confused developers onboarding Ember projects. Without an
```
rwjblue, line 46:
```suggestion
Implicit injections have long confused developers onboarding into Ember projects. Without an
```

text/0680-implicit-injection-deprecation.md
```diff
@@ -0,0 +1,127 @@
...
+explicit injection visible on the class body, developers start to wonder how certain properties got there.
+We can infer from a common implicit injection in `ember-data` to reason about some of the downsides this presents.
+If a project has `ember-data` installed, an initializer at runtime registers the `@ember-data/store` on both
+Ember.Route and Ember.Controller factory objects.  This is a nice convenience for a project with many routes
```
rwjblue, line 50: Let's use module paths instead of globals

text/0680-implicit-injection-deprecation.md
```diff
@@ -0,0 +1,127 @@
...
+that are used for data fetching. However, it leads to many inconveniences that hinder learning and advancement
+of the framework. This includes:
+
+1. By entangling specific objects with the whole dependency tree, we might be preventing incremental adoption.
+   For example, `ember-data` has laid out its plan for adopting only specific packages in [Project Trim](https://github.com/emberjs/data/issues/6166).
+   However, because `store` is injected by default on all Ember.Route and Ember.Controller factory objects,
```
rwjblue, line 56: Please update to module paths instead of globals here

text/0680-implicit-injection-deprecation.md
```diff
@@ -0,0 +1,127 @@
...
+   performance hit. For example, if you app does not need `ember-data` to render the entry route, the `store` injection will
+   take up more CPU cycles than necessary, hurting common user facing metrics.
+
+3. Eager initialization makes it hard to users to write tests that need to take advantage of a stub.
+   This is often a silent and annoying hurdle that users have to get around when writing tests.
+   Moreover, this is especially apparent for users moving to Ember's new testing APIs in [RFC 232](https://github.com/emberjs/rfcs/blob/master/text/0232-simplify-qunit-testing-api.md) and [RFC 268](https://github.com/emberjs/rfcs/blob/master/text/0268-acceptance-testing-refactor.md),
+   ensuring instance initializers run before each test.
+
+4. Injection on commonly used framework objects can lead to deviation of project specific styleguides. For instance,
+   injecting the `router` on the Ember.Component factory object will certainly lead to instances where the `router` is explicitly
```
rwjblue, line 71:
```suggestion
   injecting the `router` on the `@ember/component` factory object will certainly lead to instances where the `router` is explicitly
```

text/0680-implicit-injection-deprecation.md
```diff
@@ -0,0 +1,127 @@
...
+Removing implicit injections is not possible until we first issue deprecations.
+This helps us maintain our strict semver commitment to not completing a major version bump and
+removing APIs without a deprecation path. As a result, it will take us 3 phases to remove `owner.inject`.
+
+### 1. Deprecate implicit injection on target object
+
+The first step will be issuing a deprecation on the target factory object that is receiving a property. This will surface the deprecation
+on user's owned objects including transitively through addons like `ember-data`. This can be accomplished by installing a native getter
+and setter on the target. Whenever the property is accessed, we will check if the property is explicitly injected. If it isn't,
+we will issue a deprecation with an until value, id and url to read more about this deprecation.
```
rwjblue, line 90: It would probably be good to mention that these getters/setters will only be setup in `DEBUG` builds, and in production builds we will assign the property just like we do now. The goal is to avoid any production penalty for supporting this deprecation.

text/0680-implicit-injection-deprecation.md
```diff
@@ -0,0 +1,127 @@
...
+5. Users have come to associate `Route.model()` with a hook that returns `ember-data` models in the absense of explicit injection.
+   While this can be true, it is not wholly true. New patterns of data loading are becoming
+   accepted in the community including opting to fetch data in a Component.
+
+
+## Detailed design
+
+Removing implicit injections is not possible until we first issue deprecations.
+This helps us maintain our strict semver commitment to not completing a major version bump and
+removing APIs without a deprecation path. As a result, it will take us 3 phases to remove `owner.inject`.
```
rwjblue, line 83: I think we need to explain a bit more about why we need multiple phases here (and in each phase). 

At the summary level, we should explain that it is impossible for `ember-data` to remove its `owner.inject` calls without being a breaking change (that could not have issued a deprecation for, since we couldn't tell if you **actually** used it).

text/0680-implicit-injection-deprecation.md
```diff
@@ -0,0 +1,127 @@
...
+
+### 1. Deprecate implicit injection on target object
```
rwjblue, line 85: Should mention _why_ we deprecate this first, and how you avoid the deprecation. Specifically, this would allow users to add `@service store;` explicitly as a way to prevent the deprecation and avoid relying on the implicit injection.

text/0680-implicit-injection-deprecation.md
```diff
@@ -0,0 +1,127 @@
...
+
+The first step will be issuing a deprecation on the target factory object that is receiving a property. This will surface the deprecation
+on user's owned objects including transitively through addons like `ember-data`. This can be accomplished by installing a native getter
+and setter on the target. Whenever the property is accessed, we will check if the property is explicitly injected. If it isn't,
+we will issue a deprecation with an until value, id and url to read more about this deprecation.
+
+### 2. Deprecate `owner.inject`
```
rwjblue, line 92: Should mention how folks (like ember-data) should mitigate this deprecation. One possible answer is that ember-data can't remove its `owner.inject` until after ember-data@4. This ordering is a bit nebulous at the moment...

text/0680-implicit-injection-deprecation.md
```diff
@@ -0,0 +1,127 @@
...
+
+The first phase did not actually deprecate the "use" of `owner.inject`.  As a result, we need to deprecate
+it's use directly before removing.
+
+### 3. Profit!
+
+Ember 4.0 will remove the ability to utilize `owner.inject` to inject arbitrary properties on Ember's Framework objects.
+
+It is important to consider the timeline of these three phases.  Many addons and apps will need to make minor and major
+changes to their codebase. We need to consider this as we move through each phase.
```
rwjblue, line 102: I agree that considering the phases is important, can you update to indicate relative timing (and likely "triggers")? As it stands now, we say that the timing is super important but don't actually say what the timing is or should be 😆

text/0680-implicit-injection-deprecation.md
```diff
@@ -0,0 +1,127 @@
...
+
+## Alternatives
+
+- Continue with use of `owner.inject` but overhaul docs and recommend explicit injection.
+- Provide application level super class with a default store injection provided by `ember-data` and other common examples in the community.
+- Allow injection on a specific factory with a single backing Class instead of a generic factory object (e.g. `route` vs `route.index`).
+
+## Open Questions
+
+- Should we simply replace the [docs](https://guides.emberjs.com/release/applications/dependency-injection/#toc_factory-injections) on Factory Injections without an alternative?
```
rwjblue, line 122: IMHO, yes.

rwjblue (commented):
Thanks for taking this on @snewcomer! I'm very much in favor!

text/0680-implicit-injection-deprecation.md
```diff
@@ -0,0 +1,127 @@
...
+export default {
+    name: 'store-inject',
+    initialize,
+};
+```
+
+```js
+export default class PostRoute extends Route {
+  // This proposal seeks to make this service injection explicit
+  // Currently users do no need to specify this injection if ember-data is installed
```
bertdeblock, line 32:
```suggestion
  // Currently users do not need to specify this injection if ember-data is installed
```

text/0680-implicit-injection-deprecation.md
```diff
@@ -0,0 +1,127 @@
...
+If a project has `ember-data` installed, an initializer at runtime registers the `@ember-data/store` on both
+Ember.Route and Ember.Controller factory objects.  This is a nice convenience for a project with many routes
+that are used for data fetching. However, it leads to many inconveniences that hinder learning and advancement
+of the framework. This includes:
+
+1. By entangling specific objects with the whole dependency tree, we might be preventing incremental adoption.
+   For example, `ember-data` has laid out its plan for adopting only specific packages in [Project Trim](https://github.com/emberjs/data/issues/6166).
+   However, because `store` is injected by default on all Ember.Route and Ember.Controller factory objects,
+   a user cannot easily opt out of `@ember-data/store`. This is a common problem space in other communities as well.
+   With previous JetBrains IDE installations, adding plugins required a reload. However, by removing a contructor function
```
richard-viney, line 58:
```suggestion
   With previous JetBrains IDE installations, adding plugins required a reload. However, by removing a constructor function
```

text/0680-implicit-injection-deprecation.md
BobrImperator: On a side note, I may be wrong here because of some misunderstanding or sth.

But I remember that there used to be some feature (that i have never relied on) where an implicit route would automagically load a model, is this a thing or is this something i just made up?

The question is, if it exists, how its going to be deprecated?

BobrImperator:
I misplaced my comment, so I'll quote it here.

https://github.com/emberjs/rfcs/pull/680#discussion_r518055729

> On a side note, I may be wrong here because of some misunderstanding or sth.

> But I remember that there used to be some feature (that i have never relied on) where an implicit route would automagically load a model, is this a thing or is this something i just made up?

> The question is, if it exists, how its going to be deprecated?

text/0680-implicit-injection-deprecation.md
snewcomer: Is this what you are referring to?

https://github.com/emberjs/rfcs/pull/557
https://github.com/emberjs/rfcs/issues/377

BobrImperator: Looks like it is it, thank you @snewcomer.
Was wondering how these behaviors play out in this RFC

snewcomer: As for the triggers, how does this sound?

1. Introduce deprecation on target
2. Next release introduce deprecation on owner.inject
4. Simmer until v4?

xg-wang:
One use case I have that does not seem easy to migrate is to inject a service that registers global route events to `route:application` in an `initializer`. This will eagerly instrument all route transitions without lookup the service in an `instance-initializer`, otherwise, I cannot mock the service in the test setup.

jelhan:
> One use case I have that does not seem easy to migrate is to inject a service that registers global route events to `route:application` in an `initializer`. This will eagerly instrument all route transitions without lookup the service in an `instance-initializer`, otherwise, I cannot mock the service in the test setup.

Not sure if I fully got it. But can't it be done using `RouterService.willChange` and `RouterService.didChange` events? You wouldn't need to inject anything to `ApplicationRoute` in that case.

xg-wang:
@jelhan I forgot to mention the use case if for an addon to provide a service that is aware of global router states. The service from the addon does use `this.router.on('routeDidChange')` in its constructor but services are lazily created on the first lookup.

jelhan:
> @jelhan I forgot to mention the use case if for an addon to provide a service that is aware of global router states. The service from the addon does use `this.router.on('routeDidChange')` in its constructor but services are lazily created on the first lookup.

Still not getting why you would need to inject the service in application route. Couldn't you call a setup method of the service in instance initializer directly?

pzuraq:
@xg-wang I’m not sure that this particular deprecation would affect that, because it should still be possible to lookup services with `owner.lookup()`. So it would still be possible for an initializer to lookup the service and set up some event handlers, if that’s what you’re describing.

If not, can you add some code examples of the case you’re concerned about in order to clarify?

rwjblue:
> If not, can you add some code examples of the case you’re concerned about in order to clarify?

@xg-wang - Maybe throw together a demo repo for us? I'm also unsure that this RFC directly affects your scenario, but without understanding the scenario better I can't be sure...

xg-wang:
@jelhan @pzuraq @rwjblue apology for the late response. I created https://github.com/xg-wang/test--router-injection-addon to explain what I meant.

The addon will need to prepare initial states, I have a `setupMyAddonTest` style ([setupTest](https://github.com/emberjs/ember-qunit#setup-tests)) test helper to fake. However, the `this.owner.register` won't work if an `instance-initializer` looks up the service.

Being injected into the main route, the service gets access to all application route transitions.

I hope the demo makes sense, or I can explain more. It's totally possible I was doing things hacky and silly here, please let me know if there's a better way for the use case.

xg-wang:
Talked to @pzuraq offline, the recommendation for this timing requirement in my demo addon is to look up the service in an `instance-initializer` and patch the service instance's methods in the test helper, instead of re-registering the class.

rwjblue:
We discussed this at todays team meeting, and are moving into final comment period.

text/0680-implicit-injection-deprecation.md
runspired: FTR we eagerly inject only because of defaultStore. Were defaultStore not present we would deprecate this ourselves. Deprecating implicit injection without deprecating defaultStore is surprising to me, I'm hoping there was a follow up RFC merged that I've missed 😅

snewcomer: I haven't seen `DS.defaultStore` anywhere for quite a while.  Does it still exist?

runspired: @snewcomer it's not an EmberData concept, it's an Ember concept. You even linked my issue for needing the framework to remove it in the RFC but I think you possibly misunderstood it as something that we (EmberData) needed to remove. It is currently located here in Ember's codebase: https://github.com/emberjs/ember.js/blob/master/packages/%40ember/-internals/routing/lib/system/route.ts#L2323

pzuraq: @runspired is I believe referring to this computed property: https://github.com/emberjs/ember.js/blob/master/packages/@ember/-internals/routing/lib/system/route.ts#L2319

We can and should also deprecate that property, though I think it will have to be for the 5.0 cycle. The deprecation warning should be enough to let user's know that they are using the deprecated functionality, and should explicitly inject `@store`, I think.

runspired: @pzuraq the issue is that if we don't eagerly inject ourself the `find` call will not be the same as having called `ember-data's` `store.find` and results in either this assert triggering or (for folks that happen to have a `find` method on their model) a very confusing and unexpected request triggering.

```
assert(
            `${classify(name)} has no method \`find\`.`,
            typeof modelClass.find === 'function'
          );
```

runspired: For folks that were previously making use of EmberData's injected store, they will get the deprecation and convert to using explicit service injection. For new apps in 4.0 and for places where devs simply forget to add in the explicit inject though the confusion will remain.

pzuraq: Right, exactly. I don't think this will be a huge issue as long as we add a deprecation that says loudly in the console:

```
You are using the deprecated `store` computed property. This computed property is not a service, and it has confusing and usually unexpected behavior. If you are using Ember Data or another addon that provides a `store` service, you should add `@service('store')` to your route in order to get access to the store service instead.
```

I do agree ideally that the deprecation should be sooner rather than later ideally, but we'll have to see where it fits in the timeline.

jelhan: When deprecating default store we should deprecate implicit record loading in default `model` hook as well. There is a open RFC to do so #557.

PR #682, opened 2020-11-09
Update templates with new frontmatter style

kategengler:
- Fix recently merged RFCs to have new style
- Add CI to check added RFCs for required frontmatter


rwjblue:
I would generally prefer to keep the check as "pending" (not passed or failed), is that possible?

kategengler:
In this, the check for historical RFCs is still pending, but only newly added RFCs will fail if they don't have those metadata fields (Stage, Start Date, RFC PR, and Relevant Team(s)). I'd like that to fail, even before Stages if fully implemented, so that we don't add new RFCs that don't match the converted style to real frontmatter, as we did with those two recent ones.

rwjblue (approved):

Issue #661, opened 2020-09-08
[deprecate] fallback lookup paths in ember-resolver


gabrielcsapo:
Currently we have multiple fallback lookup paths in ember-resolver.  As we continue to migrate towards a more strict future where imports are verified at build time (https://github.com/emberjs/rfcs/pull/454) it would be forward looking to remove non-strict lookups from ember-resolver to make codemods to migrate to; 
- Strict Mode (https://github.com/emberjs/rfcs/pull/496)
- the v2 addon format (https://github.com/emberjs/rfcs/pull/507)
- the slimming down the ember-cli build pipeline (https://github.com/emberjs/rfcs/pull/578)
- strict service lookups (https://github.com/emberjs/rfcs/pull/585) 

It would benefit the ecosystem to remove fallback logic that enables many-to-one lookups in favor of a one-to-one mechanism.

The knock on effects of this would be having a supported solution to solve the following;
- Where does Pod resolution go and is supported?

An example of what this resolver would look like after removing all fallback lookup paths would be https://github.com/stefanpenner/ember-strict-resolver.

rwjblue:
FWIW, I'm definitely in favor of a path forward here.

gabrielcsapo:
@rwjblue will work on an RFC to get a path forward planned out.

wagenet:
Looks like we have an RFC so I'll close this ticket.

Issue #651, opened 2020-07-29
Colocate route+template+controllers


mehulkar:
In #650, a few people mentioned that colocating routes/templates/controllers in a directory is the main reason they use pods. Making this work natively (similar to colocating components JS/HBS) without pods seems like a good next step.

Not clear what the technical lift is to make this happen, but seems like a natural progression. Requirements:

- works with and without usePods: true
- mix of colocated and non-colocated things works

(Note this doesn't intend to solve all of Module Unification, it is much more simple than that)

webark:
Would you see this as just covering the existing functionality and how it behaves? (route.js vs index.js) If it is "new functionality", how do you rectify the pattern of components where you have an "index.{js,hbs}" when you have a route and controller at the same level.

Also, does this include components nested in and alongside routes?

mehulkar:
> Would you see this as just covering the existing functionality and how it behaves? 

do you mean the functionality that pods provides? I can't say because i have not used that feature and don't know what all it entails. 

> how do you rectify the pattern of components where you have an "index.{js,hbs}" when you have a route and controller at the same level.

good question. I imagine folders are required and inside it, you have a `route.js`, `controller.js` and `template.hbs` file. I suppose `foo/route/index.js` is also possible, but i don't think that's valuable. 

> does this include components nested in and alongside routes?

nope. i think that should remain out of scope.

webark:
> I imagine folders are required and inside it, you have a route.js, controller.js and template.hbs file.

The main dilemma with this is it's 
a different convention then what is used for colocated components (which uses either my-component.{js,hbs} or my-component/index.{js,hbs}). now you could do the same with the route/template combo, but then having a my-route.controller.js (or￼ equivalent) is unexpected.

If i where starting a new app from scratch, i'd probably be fine with controllers being separated from the route js|hbs, and follow the convention in a similar and more coherent fashion.

lougreenwood:
I think one large, but seemingly small benefit of co-location which is often missed is the benefit to pull request reviews, and in turn code quality.

With component co-location, seeing all aspects of a component grouped together (except tests :trollface:) makes reviewing easier and in my experience, improves the quality of code review.

It would be nice to have a similar improvement for routers/controllers/templates too 👍

mydea:
Although not perfect, I could see something like this:

* `app/routes/my-route.js`
* `app/routes/my-route.hbs`
* `app/routes/my-route.controller.js`

That could work the same way as component co-location, so also allow:

[...]

PR #683, opened 2020-11-23
Deprecate Fallback Lookup Paths in ember-resolver

gabrielcsapo:
This is a follow up rfc created based on the feedback from issue #661.


mydea:
I generally think this is a good idea, my two cents to this:

* I believe a replacement for pods needs to be defined before doing that. We use pods for route/controller/templates, and would not like to move those to the classic architecture. See: https://github.com/emberjs/rfcs/issues/651 - would love to see some traction on that. If there is interest, I would also write up an RFC PR for that issue based on the ideas collected so far in the issue - it kind of quieted down and there was never any input from the core team on how they view the issue.
* I have tried the mentioned ember-strict-resolver and found a lot of problems with it in our app. Not sure if the idea of this RFC is to use that as basis, but if so, I think there is a lot that needs to be done on that front before it could be recommended. Happy to write up my concrete experiences, if that helps in the context of this PR?

gabrielcsapo:
@mydea thanks for your repy!

1. Totally agree there needs to be a pods replacement, the suggestion for moving forward makes this opt in and since the usage of pods seems constrained to host applications they wouldn't be affected. A suggestion came up when talking to @rwjblue was to have a shim fo pods instead of having that in the main resolver.
2. We had a similar experience, most of the issues came from our dependencies and our host app not having a consistent service or component lookup pattern. I would like to know what problems you experienced if you could do a writeup that would be great.

mydea:
The main issues I encountered were:

* Missing pod support (as mentioned)
* Missing support for nested co-location index components (e.g. app/components/my-component/index.js)
* Missing pluralization support. Noticed this with ember-can, which has e.g. `app/abilitites/my-thing.js`, but it wants to look up `ability` as `abilitys`. IMHO there should either be support for custom pluralization, or alternatively just do not pluralize unkown types at all - I would be fine with putting these into `app/ability/my-thing.js`.
* After I got nested co-location to work, components from addons using non-colocated structure stopped working for me. Had to manually handle those.
* That I had to manually map _all_ multi-word services was also rather unexpected and annoying.

For reference, here is the extended strict resolver that ended up working with our application. There is for sure some room for improvement, but it shows the steps I took to make it work at least:
https://gist.github.com/mydea/d890f3b83a4bf25b3c3a4f41f6a47e42

gabrielcsapo:
Thanks for doing this @mydea 

- I totally agree the lack of pods support is a known and I think would be great to deprecate pods support in favor of a more mainstream javascript solution to take advantage of open source tools without having to code them to understand the pods architecture. 
- I was not aware of the nested index components, what would be the benefit of doing this over the normal naming scheme?
- Looking at the gist we were able to avoid a lot of the component lookup issues as were already using https://github.com/rwjblue/ember-holy-futuristic-template-namespacing-batman which helped segment our code in in-repo addons very easily. 
- We had many downstream dependencies that we had to update to ensure that the service names were properly dasherized as a result we made a map and slowly removed items from the map to make the process gradual rather than blocking on it.

I think the biggest advantage of having less logic in the resolver is that it feels less magical. The thing that you want is the thing you get and teaching becomes incredibly straightforward. My favorite part of ember is that there is convention over configuration and this feels very conventionally as we lay out a 1:1 one of doing things instead of a many:1.

mydea:
Having nested components be at `app/component/my-component/index` can be very helpful when you have a lot of nested components. e.g. Having this:

* `app/compoments/my-component/index.hbs`
* `app/components/my-component/sub.hbs`

Makes it much easier to see all parts belonging together than e.g.

* `app/components/my-component/sub.hbs`
* ... 50 other folders/components
* `app/components/my-component.hbs`

Regarding the services, not sure I follow. All our services are already referenced in dasherized form, but I still need to map them in the resolver as they are somehow internally (?) looked up in camelcase form.

gabrielcsapo:
@mydea service lookups should not be transformed in the strict resolver case. If you are getting camel case lookups it is coming from having code that doesn't explicitly set the service name in the inject method.

```
import { inject as service } from "@ember/service";
```

```
@service
fooBar
```

Instead it should be referenced like:

```
@service("foo-bar")
fooBar
```

This was hard to track down until we enabled an internal eslint rule that enforced inject calls to ensure a value was provided as a dasherized string.

mydea:
Ah OK, that was absolutely not clear to me, and to be honest also is rather surprising. That kind of feels like just adding more friction to every service definition. I guess it _is_ more explicit, but I haven't seen this as a recommendation/best practice so far.

rwjblue:
> That kind of feels like just adding more friction to every service definition

Just to be clear, it is only required for multi-word services.

gabrielcsapo:
@rwjblue is there anyone who would be able to review this?

text/0683-deprecate-fallback-resolution-ember-resolver.md
```diff
@@ -0,0 +1,139 @@
...
+Templates
+ember-strict-resolver defines a stricter and therefore much faster version of these lookups, which results in a 1:1 instead of 1:N resolution.
+
+In the flagship web app at LinkedIn, adopting ember-strict-resolver resulted in the following improvements (per a TracerBench analysis):
+
+boot phase estimated difference -188ms [-195ms to -181ms]
+transition phase estimated difference -42ms [-57ms to -27ms]
+render phase estimated difference -23ms [-37ms to -9ms]
+
+## Detailed design
```
rwjblue, line 38: This section needs to explicitly state what is being proposed. As far as I can tell "strict resolution" is never defined. What does it mean? What would issue a deprecation? What would resolve to various files on disk?

text/0683-deprecate-fallback-resolution-ember-resolver.md
```diff
@@ -0,0 +1,139 @@
...
+| controller/foo/bar‌ ‌                                   | this.controllerFor(‘foo.bar’),‌ ‌this.controllerFor(‘foo/bar’)‌ ‌                                                                                                         | this.controllerFor(‘foo.bar’)‌ ‌                         |
+| component/foo/bar‌ ‌                                    | ‌{{foo/bar}}‌ ‌                                                                                                                                                           | {{foo/bar}}‌ ‌                                           |
+| component/foo/bar-baz‌ ‌                                | ‌{{foo/bar-baz}},‌ ‌{{foo/bar_baz}}‌ ‌                                                                                                                                    | ‌{{foo/bar-baz}}‌ ‌                                      |
+| (external-module-name)/addo‌n/components/foo/bar-baz‌ ‌ | {{component‌ ‌“(external-module-name)/foo/‌bar-baz”}}‌ ‌{{component‌ ‌“(external-module-name)/foo/‌bar_baz”}}‌ ‌{{component‌ ‌“(external-module-name)/foo/‌barBaz”}}‌ ‌ ‌ | {{component‌ ‌“(external-module-name)/foo/‌bar-baz”}}‌ ‌ |
```
rwjblue, line 54: I don't fully understand this table. What are the three columns trying to show? 

I think it would be best to talk about it from the perspective of the `service:foo` calls (since those are constant from the framework perspective). So for `service:fooBar` (which is what is used when you do `fooBar: injectService()`) the strict resolver would literally look for `app/services/fooBar.js` and the current resolver would look for `app/services/foo-bar.js`.

text/0683-deprecate-fallback-resolution-ember-resolver.md
```diff
@@ -0,0 +1,139 @@
...
+
+boot phase estimated difference -188ms [-195ms to -181ms]
+transition phase estimated difference -42ms [-57ms to -27ms]
+render phase estimated difference -23ms [-37ms to -9ms]
+
+## Detailed design
+
+As libraries migrate to the new use strict format older applications will not be negatively affected as the strict subset resolution is backwards compatible from libraries to applications.
+
+The major difference between the two resolvers can be easily linted against:
```
rwjblue, line 42: Linted against **how**? Is it limited to confirming that any `injectService` calls for camelcase property names have an argument?

rwjblue (changes requested):
Overall, I'm definitely in favor of moving in this direction. Thanks for working on this!

wagenet:
I'm going to try to bring this up with the core team and see if we can move it to a conclusion.

wagenet:
@chriskrycho should this be part of #832?

chriskrycho:
@wagenet I don’t think so; this is probably still worth doing if @gabrielcsapo wants to drive it forward, but separately from the Classic stuff. The resolver has its own path to slowly being replaced, but is separate from Classic issues.

achambers:
@ef4 Is this in any way related to the RFCs you're currently working on? Also seems Embroider related. Do you have any thoughts to add here?

ef4:
Yes, https://github.com/emberjs/rfcs/pull/938 wants to introduce a new resolver implementation. That makes it a good time to drop deprecated resolver features. So if this RFC can land, we can make the new resolver in 938 not support the things this RFC deprecates.

One bit of feedback for this RFC is that we can probably avoid using an optional feature, in favor of making ember-resolver a V2 addon and having dedicated import paths for the different implementations.

ef4:
Well, the work of deciding which things not to include still needs to happen. If #683 covers that work it makes it easier to do #938.

wagenet:
We agreed to move this to exploring at the RFC Review Meeting.

PR #685, opened 2020-11-28
New browser support policy

pzuraq:
[Rendered](https://github.com/emberjs/rfcs/blob/new_browser_support_policy/text/0685-new-browser-support-policy.md)


text/0685-new-browser-support-policy.md
```diff
@@ -0,0 +1,286 @@
...
+#### Note on the versions chosen
+
+The versions chosen here are the latest versions of Safari, which having just
+been released this year, do not have 100% adoption. As such, there are still a
+significant number of users who are using previous versions of Safari. However,
+by the time this support policy is enacted and a new major version of Ember is
+released, this number will likely have shifted significantly.
+
+In addition, since Ember will not drop support for any version of Safari until
+the next major release after this policy is enacted (v5), this decision strikes
```
snewcomer, line 187: RE: `v5`. 

1. My knowledge might be a little outdated or perhaps this is still up in the air.  Will ember versions >= `v4.0` remove IE support?  Is this the impetus for having this new browser support policy?

2. To clarify this comment about `v5`, are we stating that Safari version 13.1 and under will be supported for all of v4?  

If I have this right, then when looking across multiple versions of Ember, effectively `v4.0` removes IE support and `v5.0` removes Safari < 14 (but open to other removals depending on the landscape at the time).

pzuraq: Correct, v4 will implement this new policy, this is noted a bit later but it is not super prominent, might be good to make it more prominent and note it higher up.

Re: v5, the note here is assuming that we would remove support in v5, but technically that is just the _earliest_ we could remove support, since this policy comes into effect in v4. If we were to release v5 without any new RFCs in the meantime that alter the policy, then this policy would remain in effect, and we would still support Safari 14 in v5.

That said, I think it’s very likely we will as a community want to drop support for some Safari versions whenever a major version happens, so I think its safe to assume it will be done.

chriskrycho:
Overall I'm extremely 👍 on this.

I do have a concern about the choice of Safari 14 vs. 13, and suggest that it might make sense to make that a similar threshold-based mark for the start of any given major version. While I'm sympathetic to the reasons outlined here, I'm also aware that the app I support would start breaking for a non-trivial number of our users if we did this today, and I think it makes sense to evaluate on a per-major-release basis rather than simply choosing the latest. That also gives us flexibility around when we cut a major. The current approach implicitly assumes we cut a major somewhere at least half a year after a major operating system update, based on the adoption rate we've seen in the past for Safari releases (and some years are worse than others; Safari 12 hung on for a *lot* longer than many others b/c of iOS 13's infamous instability).

Having a carefully-chosen cutoff there, not dissimilar to the policy for evergreen browsers' rolling support on LTS releases, would solve the potential issue here I think. If something happens between now and Ember 4.0 such that we *don't* see a significant change in Safari 13 usage numbers, that lets us simply say "Well, 13 it is, at least for now!" and look forward to a hopefully-less-than-3-years-long commitment to supporting it.

That also suggests another alternative in the space, which would be good to call out in the **Alternatives**: to give Safari itself the same treatment as the evergreen browsers in terms of percentage of usage base.

pzuraq:
@chriskrycho I actually didn't intend to suggest that we would always choose the latest version of non-evergreen browsers whenever we cut a major version, I thought it would be more something that we would evaluate on a case-by-case basis as you suggested. I'll make that more clear in the prose.

In this particular case, I chose Safari 14 because:

1. It will be at least 6 months until we can cut the next major of Ember, giving us a large buffer for users to adopt it and
2. All of the new JS features in Safari 14 are fully polyfillable, which means that even though we don't technically support Safari 13, it should not break any time soon and apps should be able to continue supporting it in practice.

The main implications of this are that Ember does not need to provide its _own_ polyfills for features such as `BigInt`. We have taken the stance previously that if we use a browser feature that is not supported in some of our browsers, and requires a polyfill, then we _must_ use our own polyfill that is private to Ember, so as not to interfere with potential polyfills that users have added.

That said, I don't feel too strongly about this point, the features in Safari 14 are nice to have but not crucial, so if we want to support Safari 13 for this cycle instead I think that would be reasonable.

chriskrycho:
Ah, that clarification is helpful. I think a tweak to the text to make it that much more explicit—perhaps with some criteria about how we’d choose it?—would cover all of my concerns there.

I think it might also be worth noting expressly in the text the distinction between polyfillable and nonpolyfillable features, because that distinction is such a critical part of what made 🔥 DIAF IE11 🔥 support such a deadweight—vs. features which can be polyfilled so that even if you don’t technically support an older browser, users can bring their own polyfill and solve that issue for themselves on an as-needed basis.

NullVoxPopuli:
would polyfills be opt-in, or rather: "automatic" via a particular app's browserlist?

pzuraq:
Polyfills would continue to work as they do today, either by using the `includePolyfills` option in ember-cli-babel, or using an addon like `ember-cli-babel-polyfills`. I think we could have a separate RFC where we include one of these options in the default blueprint.

mydea:
I am very in favor of this. Regarding Safari, in one of our apps we had a decent number of users hanging on to older iOS versions. I  currently do not have any up to date stats, but we did have to support Safari 9 (iOS) when 11 was already out. So I think there should definitely be a slightly larger window of support for iOS, or  maybe a percentage based approach would also work there as well.

I also agree that it might also make sense to differentiate between polyfillable and non-polyfillable features. E.g. having a full support `iOS 14+`, and `iOS 12+ (polyfills required)` or something along this line. I would not have any issues with including my own polyfills for my custom browser support, this could be part of the release docs I suppose. Realistically this only applies to non-evergreen browsers (meaning Safari).

pzuraq:
While I think the polyfill-ability of features is something we should take into account when we update this policy and choose a new set of non-evergreen browsers to support, in addition to marketshare, the RFC is purposefully not setting permanent guidelines for making this decision. The point is that it is an arbitrary line-in-the-sand, and every time we update the support policy, we will just be drawing a new one. The criteria could change between now and the next update, for instance, and we should have the discussion at that point.

I also do not think that we should adopt a "works if you bring polyfills" policy, as supporting polyfills is its own can of worms. Different polyfills work differently, and sometimes they do not work perfectly (e.g. the `Symbol` polyfill). So, using an unsupported browser version with polyfills _may_ work, but the point is that we are not obligated to support it.

For this particular decision, I think the important thing is that the majority of Ember apps are able to update to Ember v4 _before_ the last LTS of Ember v3 is no longer supported for bugfixes. So, that would be 24 weeks after the release of v4.

If we assume that v4 is released around May 2021, then that would mean October/November of 2021 or so. By then, Safari 15 should be released (assuming Apple continues its yearly release cadence).

I still think, given that timeline, that supporting Safari 14 makes the most sense, especially considering we will not be able to drop support any time soon. It would also be reasonable to support Safari 13, potentially. Personally, I would not want to drop support any lower than that.

chriskrycho:
Strongly agreed that we shouldn't be supporting anything lower than 13, and the LTS point is one I hadn't fully taken onboard. I think given that 4.4 LTS would be mid-to-late fall, I think Safari 14 as the cutoff is perfectly reasonable.

text/0685-new-browser-support-policy.md
```diff
@@ -0,0 +1,293 @@
...
+
+1. Support for a given browser version is _never_ removed.
+2. Support is _added_ for any new versions of a browser that have been released
+   while the minor is still supported.
+
+So for instance, if we were to release Ember 4.4 with support for Chrome 92, 93,
+and 94, then bugfixes and security fixes for those versions of Chrome will be
+supported until 4.4 is no longer an LTS, following the LTS support matrix. In
+addition, if Chrome were to release version 95 while 4.4 is still actively
+supported, then 4.4 will also support Chrome 95.
```
rwjblue, line 165: How do you know a version is dropped? What does a user actually have to do to know a specfic chrome version they care about is no longer supported?

text/0685-new-browser-support-policy.md
```diff
@@ -0,0 +1,293 @@
...
+
+## Implementation
+
+This policy drops support for a major browser, and therefore can only be
+implemented with a major version bump.
+
+## How we teach this
+
+- A blog post explaining the policy with a link to this RFC will be posted.
+- Browser support policy will be linked in repo
```
rwjblue, line 271: What is the actual policy? This RFC likely contains all the information, but it doesn't seem like it is quite in the format that we would want?

text/0685-new-browser-support-policy.md
```diff
@@ -0,0 +1,293 @@
...
+
+Future RFCs that amend this policy to _remove_ support for a browser will
+_require_ a major version of Ember to implement.
+
+## Implementation
+
+This policy drops support for a major browser, and therefore can only be
+implemented with a major version bump.
+
+## How we teach this
```
rwjblue, line 268: Can you add an entries to the list below:

* for maintaining a concrete list of what is supported by a given minor version (and having that)
* to indicate that the release blog post will call out any dropped versions for that minor

text/0685-new-browser-support-policy.md
```diff
@@ -0,0 +1,293 @@
...
+  supported indefinitely, we will not be able to use any new browser features
+  added afterwards without another major version. In particular, the spec for
+  `WeakRef` has finally been stabilized, and this is something we'll likely want
+  to make use of in the not-to-distant future.
+
+  While this is true, it does not impact the immediate roadmap for Ember in the
+  next few years. When the time comes, we can update our browser support policy
+  again, and release a new major version.
+
+## Alternatives
```
rwjblue, line 288: * Drop IE11 but not add more definition to support (keeping it as a case by case determination by the core teams)

mydea:
Regarding iOS, I think it is important to note that this is not totally up to the users - there are older devices which cannot be updated to an iOS version that runs Safari 14+. Of course these are older models, but I _have_ seen people keep (happily) using them, and for them the only way forward would be to buy a new device. Especially with iPads I've met quite a few people still running _very_ old devices. We don't have to care about those edge cases, but we should keep it in mind.

In this concrete example, after looking it up, it probably doesn't matter that much as all devices that support iOS 13 also support iOS 14. So I guess we might as well require iOS 14 out of the box. So :+1: from me.

text/0685-new-browser-support-policy.md
```diff
@@ -0,0 +1,293 @@
...
+     use the latest Chrome an any supported version of Windows, macOS, Android,
+     etc. The browser is versioned independently from the operating system).
+   * Automatically update whenever a new version is available.
+
+2. **Non-evergreen browsers**, which are any browsers which do not meet the
+   criteria of evergreen browsers. For instance, Safari is a supported browser
+   whose version is tied to the version of macOS and iOS that users use. As
+   users will often wait to update their operating system, many users lag behind
+   on the version of Safari they are using, so it cannot be considered
+   evergreen.
```
mehulkar, line 102: Not sure how much it makes sense to include it here, but obligatory mention of Safari Tech Preview 😉.

mehulkar (commented):
Can we mention how this relates to `config/targets.js`?

text/0685-new-browser-support-policy.md
```diff
@@ -0,0 +1,293 @@
...
+   [statcounter](https://gs.statcounter.com/).
+
+This policy will allow Ember to generally support the most recent 2-3 versions
+of the browser, and account for versions which may become temporary transition
+points. For example, when MS Edge made the switch to the new Chromium-backed
+version, the last non-Chromium version saw higher usage for a slightly longer
+period of time as the update process was not as standard, and required an
+explicit opt-in. These transitionary periods are important, but generally
+speaking temporary, and once the usage drops below the threshold Ember can drop
+support at that point.
```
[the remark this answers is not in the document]
pzuraq: Added a note calling out the support change explicitly. Re: SemVer, I think that this policy effectively captures what we actually do today, rather than changing anything substantial. We already would avoid fixes for specific versions of really old Chrome or Firefox, for instance, or versions that have a bug that is fixed in a newer version (example in FireFox: https://github.com/emberjs/ember.js/issues/19115).

So, if anything, this policy widens our support a bit by saying that we _will_ support browsers for a given time period, especially on LTS versions of Ember, and for LTS versions of browsers.

text/0685-new-browser-support-policy.md
pzuraq: Every minor release of Ember will contain the list of minimum supported versions in the blog post. I'm going to expand the "how we teach this section" to include this.

text/0685-new-browser-support-policy.md
```diff
@@ -0,0 +1,293 @@
...
+
+        - Only Internet Explorer 11 is supported.
+
+- Mobile
+
+    1. All
+
+        - All mobile browsers may work but are not explicitly supported.
+
+Any browsers not listed here may work but are not explicitly supported.
```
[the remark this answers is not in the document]
pzuraq: Removed this, as I agree this wasn't really documented and just came from the default `targets.js` in the blueprint

text/0685-new-browser-support-policy.md
```diff
@@ -0,0 +1,293 @@
...
+- Mobile
+  - Safari: 14
+
+These versions will continue to be supported until support is explicitly dropped
+via a new RFC, and dropping support for any version of these browsers will
+require a new **major** release.
+
+In addition, all new major releases of these browsers will be supported as they
+are released. Like with evergreen browsers, support is retroactively added to
+any minor version of Ember which is still supported.
```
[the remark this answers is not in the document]
mixonic: Yeah crafting a statement like this at work recently we ended up with something like "the final patch release of the a major version".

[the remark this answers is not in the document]
pzuraq: The most recent draft explicitly calls this out, and uses language similar to that, the most recent patch release is the only supported version for both evergreen and non-evergreen.

pzuraq:
@mehulkar I added a mention about `config/targets.js` to the implementation section, I do think we should make them reflect the support policy by default. Re: Safari Tech Preview, maybe we should add a clause excluding beta versions from support in general? I think while we would obviously try to figure out issues from them, we wouldn't necessarily want to support them since they're more unstable.

text/0685-new-browser-support-policy.md
```diff
@@ -0,0 +1,432 @@
...
+
+#### Mobile
+
+| Chrome | Firefox | Safari |
+| ------ | ------- | ------ |
+| 87     | 83      | 14     |
+
+#### Headless
+
+| Chrome | Firefox  |
```
kellyselden, line 337: I'm confused by references to headless safari in some places and headless firefox is others. Did I miss the explanation somewhere?

locks:
The framework team discussed this RFC at the last meeting and decided to move it into FCP. Please leave your final comments!

text/0685-new-browser-support-policy.md
```diff
@@ -0,0 +1,416 @@
...
+  3. Microsoft Edge
+  4. Safari
+- Mobile
+  1. Google Chrome
+  2. Mozilla Firefox
+  3. Safari
+- Testing
+  1. Headless Chrome
+  2. Headless Safari
+
```
jelhan, line 80: What is the reason to not support headless Firefox for testing?

pzuraq: Sorry, this was a typo, it should be headless firefox not headless safari

sandstrom:
Overall positive! 💯  I have feedback on two points:

## Safari

But one issue with this is that it *almost* adjusts to the evergreen reality.

Although Safari isn't strictly evergreen, I'd argue that in practice it mostly is. Instead of "hard-coding" a Safari version into the browser support policy, with a major version bump requirement to drop it, I think it would make much more sense to say one of these things:

- We support Safari version N-2
- We support Safari released during the past 2 years.
- We support Safari versions with browser support > X percent of [mobile] market share

Hard-coding Safari 12 into the policy feels like a bad idea.

image


## Webviews

There is no mention of web views in this policy. For anyone using Ember to build mobile apps packaged up via a web view (for example [Capacitor](https://capacitorjs.com/) or [Cordova](https://cordova.apache.org/)) and delivered as a native app, the underlying web view renderer matters.

For iOS this is simple, since it's updated in lockstep with iOS. For Android I think it has an evergreen-like update cycle.

Is this something we'd like to mention in this RFC?

typeoneerror:
> There is no mention of web views in this policy. For anyone using Ember to build mobile apps packaged up via a web view (for example [Capacitor](https://capacitorjs.com/) or [Cordova](https://cordova.apache.org/)) and delivered as a native app, the underlying web view renderer matters.

I'd also like to express my desire to see this reflected in the policy. We use Ember with Capacitor and have noticed differences in WKWebView versus Mobile Safari at times on the same version of iOS. Would be good to know if support includes this or does not explicitly.

pzuraq:
@sandstrom Re: hardcoded Safari support, I understand where you're coming from and initially felt the same way when we began discussing the new policy. However, after a while we realized a few things:

1. Ember v5 is going to happen eventually, probably sooner rather than later (e.g. within the next 5 years). We have a lot of features to deprecate, if nothing else.
2. Looking even further into the future, all this policy means is that we will have to release a new major of Ember every few years indefinitely.
3. We can always revisit this policy in the future should we decide that we no longer want to release new major versions at a regular cadence like that.

I also think it's worth noting that picking explicit versions to support actually gives us _more_ flexibility in the long run. For instance, one of the upcoming JS features we really want to begin using sooner rather than later are WeakRefs. Safari has not implemented these, but should soon now that they are stabilized. With this policy, we could potentially pick a version of Safari that has support for these for the next major version, regardless of what the usage is currently at. I'm **_not_** saying we would do that, but the point is it actually gives us the option to do the cost benefit analysis at that time, rather than having a hard cutoff that says we must support the browser even if it only has `~0.26%` global usage.

Re: Webviews, I do think this is something we want to support. Do these browsers have independent names/versions? From what I can tell they do not show up in the usage statistics, so it's hard to know how we should reference them if we add them to the policy. Should we specify something like:

- Mobile
  - Safari/Safari WebView
  - Chrome/Chrome WebView

?

typeoneerror:
@pzuraq I'm triaging Sentry errors today and there a Capacitor-based Ember app comes through as `Mobile Safari UI/WKWebView`

sandstrom:
@pzuraq Those are valid points! 

As long as the bumping Ember's major version won't hold back dropping Safari support (which is the primary drawback with the policy as currently suggested) what you say makes sense.

That one is hard for me to answer, but as long as people on the core team are happy with (for example!) doing a major bump a few months after Safari with WeakRef is released, only to be able to drop it from supported browsers, what you say makes sense.

Whichever the direction this policy goes on Safari, it's great with a new browser support policy and to finally be able to drop IE11 is awesome! 💯 

Regarding web views, yeah it's basically "stock Safari" for iOS (no difference whatsoever) and on recent versions of Android it's basically the same as Chrome (vw is part of the User Agent string).

https://developer.chrome.com/docs/multidevice/user-agent/#webview-on-android

So it's not necessarily something that needs to be spelled out in a list, could also just mention it in a sentence, e.g. "we're also aiming to support the webview pair that correspond to the supported versions of Safari and Chrome".

text/0685-new-browser-support-policy.md
```diff
@@ -0,0 +1,416 @@
...
+  2. Mozilla Firefox
+  3. Microsoft Edge
+  4. Safari
+- Mobile
+  1. Google Chrome
+  2. Mozilla Firefox
+  3. Safari
+- Testing
+  1. Headless Chrome
+  2. Headless Safari
```
rwjblue, line 79: I think you mean 

```suggestion
  2. Headless Firefox
```

pzuraq:
@sandstrom @typeoneerror We discussed support for webviews at the core team meeting today, and the outcome was that we realized that adding explicit support for these requires a bit more in-depth discussion and probably a new RFC. Specifically, webviews can have non-standard features and as such we need to think through the implications of that, and add any caveats that we may need to add to the support policy.

So for now, they will not be in the explicitly supported list of browsers, with the intention of exploring support and what it would mean in a future RFC.

To be clear, this is basically formalizing the current state of things anyways - we don't do anything special for webviews at the moment, and they work well in general, so it's _highly_ unlikely that this would change. We did add a note saying that we would also specify the underlying browser engine version we support alongside the browsers in release notes, so that should make it easier to figure out which webview version matches up with the de-facto supported browser.

If either of you or anyone else is interested in writing the RFC for webview support, definitely reach out! I'm happy to help get that process jump started! 😄

sandstrom:
@pzuraq Thanks for the update!

As long as you've thought about this I'm happy. My objective regarding web views wasn't to necessarily have them included, only to raise the question on whether they should be or not. For RFCs, the way I see it my (and others) role is to come with feedback and ask relevant questions. In this case it seems the answer is "out of scope" and that's 100% fine.

(I did try to convince you to change your mind on Safari 12, but your counter points were good so I ended up changing my mind instead 😄)

***

For us this isn't a practical issue, I just wanted to mention webviews because we do deliver a mobile app built on Ember and webviews are a part of that. Basically, we've always been having more restrictive requirements (by quite a margin) compared to Ember. For example, we dropped IE11 support about a year ago.

For us, testing if things work in a webview is a trivial process and we expect it to continue to be an easy thing, even if web views aren't explicitly mentioned in the browser support policy.

In short, I don't think you need to write a separate RFC for web views. Time is probably better spent elsewhere.

locks:
Thank you everyone for participating, it's a crucial part of the process for it to work properly :) We discussed it again at the last core meeting and decided to move it forward. The concerns about webviews seem resolved and were discussed at the same meeting as @pzuraq mentioned, so it's time to merge!

BryanCrotaz:
Rendered links to a 404

pzuraq:
I believe this is because it was merged, and the branch was deleted. Updated to link to the merged text, we should consider automating that step.

text/0685-new-browser-support-policy.md
```diff
@@ -0,0 +1,417 @@
...
+updating, so they will effectively be locked in.
+
+These versions will be used for a variety of use cases, such as generating
+documentation and release blog posts (see below), and generating the default
+`config/targets.js` for new Ember apps and addons.
+
+### Deprecation
+
+We will add a deprecation to Ember CLI that shows when users have any
+unsupported browser in there `targets.json`, along with a guide for updating to
```
Bouke, line 274: “there” should be “their”

PR #686, opened 2020-11-28
Deprecate old manager capabilities versions

pzuraq:
[Rendered](https://github.com/emberjs/rfcs/blob/deprecate-old-manager-capabilities-versions/text/0686-deprecate-old-manager-capabilities-versions.md)

### DIFF
--- a/text/0686-deprecate-old-manager-capabilities-versions.md
+++ b/text/0686-deprecate-old-manager-capabilities-versions.md
@@ -1,12 +1,12 @@
 ---
 Stage: Accepted
-Start Date:
+Start Date: 2020-11-28
 Release Date: Unreleased
 Release Versions:
   ember-source: vX.Y.Z
   ember-data: vX.Y.Z
 Relevant Team(s): Ember.js
-RFC PR:
+RFC PR: https://github.com/emberjs/rfcs/pull/686
 ---
 
 # Deprecate Old Manager Capabilities
--- a/text/0686-deprecate-old-manager-capabilities-versions.md
+++ b/text/0686-deprecate-old-manager-capabilities-versions.md
@@ -93,6 +93,8 @@ are:
    ```js
    // before
    class CustomModifierManager {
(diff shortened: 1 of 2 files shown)


chriskrycho:
Yes, please!

text/0686-deprecate-old-manager-capabilities-versions.md
```diff
@@ -0,0 +1,163 @@
...
+   This can be updated to use the definition directly:
+
+   ```js
+   // after
+   class CustomModifierManager {
+     createModifier(Definition, args) {
+       return new Definition(args);
+     }
+   }
+   ```
```
rwjblue, line 111: Can you update these examples to call out the `capabilities` bit (which is required, but not included)?

text/0686-deprecate-old-manager-capabilities-versions.md
```diff
@@ -0,0 +1,163 @@
...
+
+   In general this should be avoided, however, and users who are writing
+   modifiers should instead use the value if they want it to be tracked by the
+   modifier.
+
+## Drawbacks
+
+- There will be some minor churn in the ecosystem as managers update, but this
+  is outweighed by the lowered maintenance burden in general for this
+  functionality.
```
rwjblue, line 157: FWIW, this is a significant issue. Since updating ember-modifer to use 3.22 capabilities is a breaking change **and** ember-cli does not currently support (not well at least) having multiple versions of a given package in the build we have to do a bit of hoop jumping to accomodate this.

Regardless, I think we should do this, but ^ is a very real issue that will need to be addressed before the actual deprecation lands.

rwjblue (commented):
Thanks for putting this together! I'm definitely in favor...

locks:
The framework team discussed this RFC at the last meeting and decided to move it into FCP. Please leave your final comments

locks:
The team discussed it again at the last meeting and decided to move forward and accept the RFC. Thank you for participating!

PR #689, opened 2020-12-23
Deprecate {{hasBlock}} and {{hasBlockParams}} in templates

pzuraq:
[Rendered](https://github.com/emberjs/rfcs/blob/deprecate-has-block/text/0689-deprecate-has-block.md)

### DIFF
--- a/text/0689-deprecate-has-block.md
+++ b/text/0689-deprecate-has-block.md
@@ -9,48 +9,54 @@ Relevant Team(s): Ember.js
 RFC PR: https://github.com/emberjs/rfcs/pull/689
 ---
 
-# Deprecate `{{hasBlock}}` in templates
+# Deprecate `{{hasBlock}}` and `{{hasBlockParams}}` in templates
 
 ## Summary
 
-Deprecate the `{{hasBlock}}` property in templates in favor of the
-`{{has-block}}` helper keyword.
+Deprecate the `{{hasBlock}}` and `{{hasBlockParams}}` properties in templates in
+favor of the `{{has-block}}` and `{{has-block-params}}` keywords.
 
 ## Motivation
 
 `{{hasBlock}}` is a way to check if component has a default block provided to
-it. It is effectively an alias to calling `{{has-block}}` without providing a
-template name, or calling `{{has-block "default"}}`. It is also not called, and
-acts like a template fallback lookup instead.
+it, and `{{hasBlockParams}}` is a way to check if that default block has block
+parameters. They are effectively aliases for calling `{{has-block}}` and
+`{{has-block-params}}` respectively, without providing a block name or with the
+block name `"default"`. They are also not called, and acts like a template
+fallback lookup instead.
 
 ```hbs
-{{! hasBlock can be referenced directly }}
-{{#if hasBlock}}
+{{! hasBlock and hasBlockParams can be referenced directly }}
+{{#if hasBlock}}{{/if}}
 
-{{/if}}
+{{#if hasBlockParams}}{{/if}}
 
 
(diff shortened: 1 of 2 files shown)


text/0689-deprecate-has-block.md
```diff
@@ -0,0 +1,102 @@
...
+## Summary
+
+Deprecate the `{{hasBlock}}` property in templates in favor of the
+`{{has-block}}` helper keyword.
+
+## Motivation
+
+`{{hasBlock}}` is a way to check if component has a default block provided to
+it. It is effectively an alias to calling `{{has-block}}` without providing a
+template name, or calling `{{has-block "default"}}`. It is also not called, and
```
locks, line 23: What do you mean by "it is also not called"?

bertdeblock:
👋 Should [hasBlockParams](https://api.emberjs.com/ember/3.23/classes/Ember.Templates.helpers/methods/hasBlockParams?anchor=hasBlockParams) be included in this RFC as well or should that be a separate one?

It seems:
```handlebars
{{#if hasBlockParams}}

{{/if}}
```
and
```handlebars
{{#if (has-block-params)}}

{{/if}}
```
accomplish the same thing as well.

chancancode:
Yep let's add `hasBlockParams` to the RFC as well!

chancancode:
We discussed this on the framework core team meeting today. We are all 👍 but agreed logistically it's easier to add `hasBlockParams` to this RFC first.

pzuraq:
Yeah, the only reason it was not part of this RFC was because the transform only documents that it works for `hasBlock` 🙃  will add it this week

ijlee2:
Hi, @pzuraq. Could you update the link for `Rendered`? I think it's broken cause the RFC title changed. Thanks.

text/0689-deprecate-has-block.md
```diff
@@ -0,0 +1,142 @@
...
+You may optionally pass a name to `(has-block)`, the name of the block to check.
+The name corresponding to the block that `{{hasBlock}}` represents is "default".
+Calling `(has-block)` without any arguments is equivalent to calling
+`(has-block "default")`.
+
+#### `{{hasBlockParams}}`
+
+The `{{hasBlockParams}}` property is true if the component was given a default block,
+and false otherwise. To transition away from it, you can use the `(has-block-params)`
+helper instead.
```
jelhan, line 96:
```suggestion
```suggestion
The `{{hasBlockParams}}` property is true if the component was invoked with block
params for the default block, and false otherwise. To transition away from it, you can
use the `(has-block-params)` helper instead.
```
helper instead.
```

jelhan (commented):
Both `{{has-block}}` and `{{has-block-params}}` are currently documented as `hasBlock` and `hasBlockParams` in [API docs](https://api.emberjs.com/ember/3.24/classes/Ember.Templates.helpers/methods/hasBlockParams?anchor=hasBlock). This might be very confusing when this deprecation lands.

rwjblue:
@jelhan - ya, totally agree! We'll need to ensure that the API docs are updated at the time the deprecation guide lands.

@pzuraq - Would you mind double checking the prose here to see if we need to explicitly call that out? I think it may be already included in the existing prose RE: "How we teach this", but would love a second set of 👀...

knownasilya:
Regarding implementing this, should this be a deprecation log on build? Since it's a build time transform?

chancancode:
I think either would be acceptable, but generally if we can do it at build time that would be preferable IMO.

rwjblue:
Agree. Doing it in the build gives us easier access to actual line / column numbers which is very helpful.

PR #690, opened 2020-12-23
Deprecate `{{attrs}}` in templates

pzuraq:
[Rendered](https://github.com/emberjs/rfcs/blob/deprecate-attrs-in-templates/text/0690-deprecate-attrs-in-templates.md)


rwjblue:
Lets do it! Moving into final comment period.

PR #691, opened 2020-12-23
Deprecate passing `classBinding` and `classNameBindings` as arguments

pzuraq:
[Rendered](https://github.com/emberjs/rfcs/blob/deprecate-legacy-class-binding-curly-invocation/text/0691-deprecate-class-binding-and-class-name-bindings.md)

### DIFF
--- a/text/0691-deprecate-class-binding-and-class-name-bindings.md
+++ b/text/0691-deprecate-class-binding-and-class-name-bindings.md
@@ -6,7 +6,7 @@ Release Versions:
   ember-source: vX.Y.Z
   ember-data: vX.Y.Z
 Relevant Team(s): Ember.js
-RFC PR:
+RFC PR: https://github.com/emberjs/rfcs/pull/691
 ---
 
 # Deprecate passing `classBinding` and `classNameBindings` as arguments
--- a/text/0691-deprecate-class-binding-and-class-name-bindings.md
+++ b/text/0691-deprecate-class-binding-and-class-name-bindings.md
@@ -128,7 +128,7 @@ After:
 
 ```hbs
 <MyComponent
-  @class="
+  class="
     {{if this.foo "bar"}}
     {{if this.some.boundProperty "bound-property"}}
     {{if this.isValid "is-valid" "is-invalid"}}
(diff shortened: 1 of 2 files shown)


text/0691-deprecate-class-binding-and-class-name-bindings.md
```diff
@@ -0,0 +1,162 @@
...
+At some point, the ability for merging on the instance itself was removed. In
+its place, a general template transform was added, which transformed the
+`classBinding` and `classNameBindings` args into the `class` argument, with the
+same semantics.
+
+```hbs
+{{some-component class="bar" classNameBindings="foo:truthy:falsy"}}
+
+{{! becomes }}
+{{some-component class=(concat "bar" (if this.foo "truthy" "falsy"))}}
```
CvX, line 47: Shouldn't there be a space in "bar" when using concat?

```suggestion
{{some-component class=(concat "bar " (if this.foo "truthy" "falsy"))}}
```

text/0691-deprecate-class-binding-and-class-name-bindings.md
```diff
@@ -0,0 +1,162 @@
...
+## Transition Path
+
+In general, users should convert to use angle bracket syntax to invoke their
+component, and then use standard interpolation within the `class` attribute of
+the component. In cases where this is not possible/desired, users can continue
+using curly invocation with `(concat)` to pass the value.
+
+## How We Teach This
+
+### Deprecation Guide
```
snewcomer, line 77: Is there an `until` for this deprecation?

text/0691-deprecate-class-binding-and-class-name-bindings.md
```diff
@@ -0,0 +1,162 @@
...
+    {{if this.isValid "is-valid" "is-invalid"}}
+  "
+>
+```
+
+Note that we are passing in the `@class` argument, not the `class` attribute.
+This is to maintain the existing behavior exactly, but in most cases it will be
+safe to convert to using the `class` attribute as well. It is only unsafe if you
+were directly referencing the value of the `@class` argument anywhere in your
+component.
```
chancancode, line 143: I suggest to flip this around and use the `class=` syntax in the example (which is what most people are going to look at and copy) and only mention the caveat in text. I think it's extremely unlikely that this will matter to real apps – you would have to know that `classBinding` and `classNameBindings` transforms into the `class` argument, plus it doesn't give you access to other classes added by the JS code. You could argue this is not really supported even.

text/0691-deprecate-class-binding-and-class-name-bindings.md
rwjblue: It will be until 4.0

chancancode:
We discussed this on the framework core team meeting today and decided to move this into the final comment period (assuming @pzuraq is going to apply my suggestion above – but it's a small enough detail that doesn't materially changes much about the RFC).

PR #692, opened 2020-12-23
Deprecate Array Observers

pzuraq:
[Rendered](https://github.com/emberjs/rfcs/blob/deprecate-array-observers/text/0692-deprecate-array-observers.md)

### DIFF
--- a/text/0692-deprecate-array-observers.md
+++ b/text/0692-deprecate-array-observers.md
@@ -6,7 +6,7 @@ Release Versions:
   ember-source: vX.Y.Z
   ember-data: vX.Y.Z
 Relevant Team(s): Ember.js
-RFC PR:
+RFC PR: https://github.com/emberjs/rfcs/pull/692
 ---
 
 # Deprecate Array Observers
--- a/text/0692-deprecate-array-observers.md
+++ b/text/0692-deprecate-array-observers.md
@@ -177,7 +177,7 @@ class ToStringArray extends ArrayProxy {
     return this.content.map((value) => value.toString());
   }
 
-  objectAtContent() {
+  objectAtContent(index) {
     return this._content.objectAt(index);
   }
 }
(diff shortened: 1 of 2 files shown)


text/0692-deprecate-array-observers.md
```diff
@@ -0,0 +1,442 @@
...
+lazy way, but this is a separate change that should not be tied to this specific
+deprecation.
+
+For the deprecation guide, see the transition path above.
+
+## Drawbacks
+
+- While array observers are not used commonly in public code, they are used by
+  several commonly used addons, such as `ember-collection`,
+  `ember-data-model-fragments`, and `ember-light-table`. Deprecating them will
```
snewcomer, line 422: Do we want to mention ember-data here?  Seems like this might be an intermediate refactor for hasMany objects.

https://github.com/emberjs/data/blob/623145d0009d683a50468ba9961431ae76ae6df6/packages/model/addon/-private/system/many-array.js#L164

pzuraq: Calling `arrayContentWillChange` and `arrayContentDidChange` is how an EmberArray _triggers_ array observers. Internally, Ember Data does not rely on array observers at all, but it must call these methods in case _others_ are using them.

This is a good point though, in that it's a similar situation to the one we're facing with implicit injections, Ember Data cannot remove these method calls until there are no more users, which means they cannot immediately comply with a deprecation. So, I think we should make it a phased deprecation, where `arrayContentWillChange` and `arrayContentDidChange` are fully deprecated in the next major version after array observers themselves are deprecated and removed.

rwjblue:
Moving into FCP

text/0692-deprecate-array-observers.md
```diff
@@ -0,0 +1,442 @@
...
+import { cached } from 'tracked-toolbox';
+
+class ToStringArray extends ArrayProxy {
+  @cached
+  get _content() {
+    return this.content.map((value) => value.toString());
+  }
+
+  objectAtContent() {
+    return this._content.objectAt(index);
```
kanongil, line 181: `index` is not defined here.

PR #695, opened 2020-12-26
Argument Default Primitives

pzuraq:
[Rendered](https://github.com/emberjs/rfcs/blob/add-default-args/text/0695-default-args.md)

### DIFF
--- a/text/0695-default-args.md
+++ b/text/0695-default-args.md
@@ -6,7 +6,7 @@ Release Versions:
   ember-source: vX.Y.Z
   ember-data: vX.Y.Z
 Relevant Team(s): Ember.js
-RFC PR:
+RFC PR: https://github.com/emberjs/rfcs/pull/695
 ---
 
 # Argument Default Primitives
--- a/text/0000-default-args.md
+++ b/text/0000-default-args.md
@@ -1,0 +1,556 @@
+---
+Stage: Accepted
+Start Date: 2020-12-26
+Release Date: Unreleased
+Release Versions:
+  ember-source: vX.Y.Z
+  ember-data: vX.Y.Z
+Relevant Team(s): Ember.js
+RFC PR:
+---
+
+# Argument Default Primitives
+
+## Summary
+
+Add a `getDefaultArgs` hook to component managers which can be used to provide
+default arguments to components.
+
+## Motivation
+
+With the introduction of Glimmer components and Template-only components which
+do not have an implicit backing class, Ember users no longer have a way to
+provide default arguments to components. Previously, they could do this by
+assigning a value to the property with the same name on the component class, and
+refer to the argument directly on the class instance:
+
+```js
+// app/components/my-component.js
+import Component from '@ember/component';
+
+export default class MyComponent extends Component {
+  someArg = 123;
+}
+```
+```hbs
+{{! app/components/my-component.hbs }}
+{{this.someArg}}
+```
+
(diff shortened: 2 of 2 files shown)


text/0695-default-args.md
```diff
@@ -0,0 +1,556 @@
...
+
+```js
+class MyManager {
+  getDefaultArgs(definition) {
+    return {
+      named: {
+        a: 1,
+        b: () => [],
+      },
+      positional: ['foo']
```
snewcomer, line 382: Do Glimmer components and their component managers support positional params?

pzuraq: Glimmer components do not support positional parameters, which is why the proposed `setDefaultArgs` API for Glimmer components only accepts named arg defaults. In general though, the VM does not really distinguish between components with and without positional parameters. You can invoke a Glimmer component with positional parameters, like so:

```hbs
{{my-glimmer-component @positional1}}
```

And the VM will pass that positional parameter to the component - there is just no way for the component to use that positional argument or refer to it in any way, since the underlying manager does not expose it. That's why in general, manager APIs receive both named and positional args.

text/0695-default-args.md
```diff
@@ -0,0 +1,556 @@
...
+This hook receives the definition of the component as its first and only
+argument, and returns an object containing default argument values. Valid
+default argument values are either:
+
+1. Primitive JavaScript values
+2. Frozen arrays or object
+3. Functions which return a default value
+
+The default value will be used for the corresponding named and positional
+arguments if the argument that was passed in is either `null` or `undefined`.
```
jelhan, line 225: What are the benefits to treat `null` the same as `undefined`? It seems to limit the component APIs significantly.

pzuraq: The main idea was just keeping it consistent between the way `null` and `undefined` are handled when interpolating in templates, so the rules are consistent in templates. I agree this could be limiting though, and in JS for instance, `??` is different than argument defaulting. `??` will coalesce if the value is `null` or `undefined`, whereas argument defaulting will use the default value _only_ if it is `undefined`.

Are there motivating use cases you can think of for when users would want to be able to pass `null`, and not trigger a default? It would be helpful if we could show such use cases in the motivation

ro0gr: Not a specific use case, but I believe these are just semantically different. I may want to tell a component, the value must be initialized with nothing.

jelhan: I find it difficult to come up with a realistic use case. I can image many components, which support `null` as a value for one of their arguments. So please excuse if that example may sound a little bit made up.

Let's imagine a `<Switch>` component, which three states: `true`, `false` or `null`. The component may be used in a poll to select the answers a) yes, b) no, c) I don't know. 

The component has a `@value` argument, which allows a consumer to pass in the state. For unknown reasons the developer of the component decides that the argument is optional and that `true` is a good default for it.

If the argument default primitives do not distinguish between `null` and `undefined`, this use case would not be supported. The default value would be used even if the consumer explicitly invokes the component with `@value={{null}}`.

mehulkar: We use null to opt out of using a default component name. This could probably be better modeled with named blocks though.

wagenet:
It's unclear that we still need this, especially with template imports. If I'm mistaken, please let me know and I'll reopen and work on getting this reviewed.

Issue #596, opened 2020-02-22
Hello GitHub Actions ✅, Bye Bye Travis 👋


buschtoens:
I just setup my first GitHub Actions for an Ember addon: [buschtoens/ember-link#291](https://github.com/buschtoens/ember-link/pull/291)
(Thanks @dfreeman! :beers:)

It was super simple and so far works perfectly. 😍

I propose we add something similar to [`.github/workflows/ci.yml`](https://github.com/buschtoens/ember-link/blob/master/.github/workflows/ci.yml) to the default blueprints and remove `.travis.yml` in exchange. One awesome side-effect of this is that it will _just work_, without the repo maintainer having to sign up with and link Travis for every new repo.

I'm not sure, whether this needs a fully-fledged RFCs, but I'd be happy to submit one and also champion the implementation.

All in favor say 👍 ? 😄

buschtoens:
We might also wanna add [`.dependabot/config.yml`](https://github.com/buschtoens/ember-link/blob/master/.github/workflows/ci.yml), while we're at it! 🤖 

I'm unsure, whether we should enable auto-merge by default though. Even though CI would be enabled by default now (thanks to GitHub Actions), not everyone always writes (good enough) tests. I could imagine that quite a few people would be annoyed, if their dependencies suddenly started changing beneath their feet. 😅 

Also here, happy to RFC this, if necessary.

nightire:
> We might also wanna add [`.dependabot/config.yml`](https://github.com/buschtoens/ember-link/blob/master/.github/workflows/ci.yml), while we're at it! 🤖

The link above should be: https://github.com/buschtoens/ember-link/blob/master/.dependabot/config.yml

buschtoens:
@nightire Good catch, thanks! I fixed the link.

buschtoens:
One concern that was raised: We should not default to any provider, or at least should make it convenient to choose.

Taking this into account, I would propose to not ditch Travis completely, but rather add a new flag / option to the blueprints, defaulting to GitHub Actions.

Alonski:
I had a similar thought a year or so ago. That we should maintain multiple configurations for different CI providers and allow adding them via blueprints.

sivakumar-kailasam:
I'd love to see this as a part of [ember-ci](https://github.com/mike-north/ember-ci) & include ember-ci in ember-cli so that we don't default to any provider but rather provide the blueprints to choose one.

alexlafroscia:
[...]

PR #696, opened 2021-01-04
Replace Travis CI with GitHub Actions in generated Ember CLI projects

kellyselden:
[Rendered](https://github.com/kellyselden/rfcs/blob/ember-cli-github-actions/text/0696-ember-cli-github-actions.md)

### DIFF
--- a/text/0696-ember-cli-github-actions.md
+++ b/text/0696-ember-cli-github-actions.md
@@ -1,0 +1,58 @@
+---
+Stage: Accepted
+Start Date: 2021-01-04
+Release Date: Unreleased
+Release Versions:
+  ember-source: vX.Y.Z
+  ember-data: vX.Y.Z
+Relevant Team(s): Ember CLI
+RFC PR: https://github.com/emberjs/rfcs/pull/696
+---
+
+# Replace Travis CI with GitHub Actions in generated Ember CLI projects
+
+## Summary
+
+Travis CI is no longer an easy setup for our users.
+GitHub Actions is the obvious replacement.
+
+## Motivation
+
+With the change of ownership, Travis CI is cracking down on open source usage.
+https://travis-ci.org has limited the number of builds for open source, which means in practice, most builds are queued for hours.
+https://travis-ci.org is also slated for shutdown soon.
+https://www.travis-ci.com for open source has been converted to free trial.
+After the initial credits are used, you must submit justification for more.
+There have been reports that they are refusing requests for more open source credits https://twitter.com/james_hilliard/status/1336081776691843072.
+
+## Detailed design
+
+We will replace the .travis.yml with a .github/workflows/ci.yml.
+It will be configured as close as possible to the existing Travis CI setup.
+Unlike Travis CI, there is no enable or authorization step.
+The presence of a file in .github/workflows/ is automatically recognized and run if hosted on GitHub.
+
+## How we teach this
+
+A blog post could be sufficient explaining this switch.
+Existing projects can continue to use Travis CI in its limited capacity.
+Those using ember-cli-update could choose to take the new GitHub Actions file or continue using the .travis.yml.
+I don't think we cover CI or Travis CI in our docs.
+
+## Drawbacks
+
+Some CI systems are agnostic to code hosting location, but I don't think you can use GitHub Actions without also placing the code there.
+This further ingrains the GitHub platform to people that may prefer a different code hosting provider.
+
+GitHub Actions also has a limit how much an open source project can use, but it resets every month.
+I haven't heard of anyone hitting the limit though.
+
+GitHub Actions doesn't support retrying individual jobs like Travis CI.
+This could impact some user's monthly budget.
+
+## Alternatives
+
+CircleCI and GitLab are other popular alternatives.
+Perhaps one would be better as a recommendation from us?
+
+We could also refrain from suggesting/including any CI config files.


text/0696-ember-cli-github-actions.md
```diff
@@ -0,0 +1,56 @@
...
+
+GitHub Actions also has a limit how much an open source project can use, but it resets every month.
+I haven't heard of anyone hitting the limit though.
+
+GitHub Actions doesn't support retrying individual jobs like Travis CI.
+This could impact some user's monthly budget.
+
+## Alternatives
+
+CircleCI and GitLab are other popular alternatives.
```
MichalBryxi, line 55: If there is going to be some pain with switching the platforms, it seems better to unify those systems under one roof. EmberJS (core) git projects are hosted on GitHub, so having CI there seems like a logical step.

kellyselden: But we don't say you have to put your personal project on GitHub, so I think the fact that we use internally is irrelevant.

Turbo87 (approved):

ijlee2:
A similar proposal (https://github.com/emberjs/rfcs/issues/596) was made several months ago, in the form of an issue. Do we want to combine the two somehow?

kellyselden:
Oh darn I didn't see that. Yes it seems like the same issue. I'd like to avoid the larger scope of that one. It mentions dependabot, code coverage, and new CLI options. This one can be completed much quicker with its limited scope.

text/0696-ember-cli-github-actions.md
```diff
@@ -0,0 +1,56 @@
...
+Some CI systems are agnostic to code hosting location, but I don't think you can use GitHub Actions without also placing the code there.
+This further ingrains the GitHub platform to people that may prefer a different code hosting provider.
+
+GitHub Actions also has a limit how much an open source project can use, but it resets every month.
+I haven't heard of anyone hitting the limit though.
+
+GitHub Actions doesn't support retrying individual jobs like Travis CI.
+This could impact some user's monthly budget.
+
+## Alternatives
```
jelhan, line 53: It misses the alternative to not include any CI configuration in the default blueprints. Instead some tooling could be provided to generate a CI configuration for the target provider (TravisCI, GitHub Actions, GitLab CI, Circle CI etc.). The script could be smart enough to provide sensitive defaults based on the Ember CLI version used. It could also support ugprades of existing CI configuration to the latest best practices. It could be even included in default blueprints and expose an Ember CLI command like `ember ci` or a blueprint like `ember generate ci`.

To be honest that would be my preferred solution. [`create-github-actions-setup-for-ember-addon`](https://github.com/jelhan/create-github-actions-setup-for-ember-addon) could be seen as a very early step in that direction.

sukima:
Alternative: Move the CI scripts into generators thus allowing the developer of a newly generated app to opt-in to the CI they prefer with commands like `ember g github-ci`, `ember g circle-ci`, or `ember g gitlab-ci`.

text/0696-ember-cli-github-actions.md
kellyselden: Good point. I added the alternative of not shipping a CI file. As for adding new functionality, I left that out to keep the scope small for this change, since Travis CI is basically inoperable now and fixing this ASAP would be best.

text/0696-ember-cli-github-actions.md
```diff
@@ -0,0 +1,58 @@
...
+
+CircleCI and GitLab are other popular alternatives.
+Perhaps one would be better as a recommendation from us?
+
+We could also refrain from suggesting/including any CI config files.
```
jelhan, line 58: This would mean removing `.travis.yml` from default blueprints without a replacement, wouldn't it? This would be my preferred short-term solution. Additionally we should experiment with alternative solutions to provide an easy setup of CI pipelines. But no need to do this kind of experiments in Ember CLI itself.

Turbo87: The reality is that the majority of Ember addons are living on GitHub, and that they have been using TravisCI by default so far. Not including a config means that new addons don't get CI by default, and existing addons might have their CI config removed. That would not be a good solution IMHO.

I agree with Kelly that the MVP here is replacing the TravisCI config with a default GitHub Action config, and if that isn't used because of using some other hosting provider then people can delete it.

This does not mean we should stop considering other options, like the generator that was mentioned. But we need a short-term solution now, and this RFC has a much smaller scope, which makes it much more likely that we can implement it soonish.

jelhan: > The reality is that the majority of Ember addons are living on GitHub, and that they have been using TravisCI by default so far.

I agree that this is true for open source addons. But I'm not sure if it is still true if also considering internal addons. We are making heavy use of private addons in the project I'm currently working on. Not using GitHub or TravisCI for it. Not sure how it is for others.

The argument also doesn't take applications into account. CI configuration is currently not only included in addon but also in application blueprints. We are not using GitHub but GitLab at work. I _guess_ many other teams do as well.

> Not including a config means that new addons don't get CI by default

Long time since I created a new addon, which was using TravisCI. But if I recall correctly maintainer needed to enable the TravisCI for the new project explicitly. I don't see a big difference between doing that manual step or running a setup script like `yarn create-github-actions-for-ember-addon`.

> and existing addons might have their CI config removed.

The motivation section of this RFC states that the existing CI configuration is unusable. So deleting the `.travis.yml` when updating Ember CLI blueprints should not make it worth. Additionally nothing prevents a maintainer from _not_ following that blueprint change.

> I agree with Kelly that the MVP here is replacing the TravisCI config with a default GitHub Action config, and if that isn't used because of using some other hosting provider then people can delete it.

The file will be recreated by `ember-cli-update` everytime it changes in default blueprints. For CI configuration this is atleast once per LTS version. It's easy to miss on upgrades. It quickly gets annoying. We stopped deleting the existing `.travis.yml` file at work therefore. @mehulkar raised the same concern in https://github.com/emberjs/rfcs/issues/596#issuecomment-632909864.

> This does not mean we should stop considering other options, like the generator that was mentioned. But we need a short-term solution now, and this RFC has a much smaller scope, which makes it much more likely that we can implement it soonish.

As far as I understand we agree that `.travis.yml` should be removed from default blueprints. We seems to disagree if a GitHub Actions workflow should be added as a replacement or not. Taking scope into account not proving a replacement as part of the blueprint is smaller than providing one.

Turbo87: > Taking scope into account not proving a replacement as part of the blueprint is smaller than providing one.

it not smaller at all, because "should we include a CI config at all?" is a much larger discussion than "should we include a CI config for the most used CI service?"

I'm aware that for internal apps and addons different rules apply, but I am primarily worried about public addons not running CI anymore because of the change you propose. If internal apps and addons break because they are not running CI anymore then that's an internal problem, but if public addons stop working then that becomes a problem of everyone.

kiwiupover:
@kellyselden thanks heaps for pushing this forward.

elwayman02:
Great proposal!

One caveat I would like to point out about GitHub Actions, however, is that it does not truly have an "allow-failure" option for testing against beta/canary builds, the way that Travis does. We will have to ensure that our replacement config accounts for this and also do our best as a community to pressure GitHub to support this feature. The best replacement I've seen so far is [`continue-on-error`](https://github.com/elwayman02/ember-user-activity/pull/341), which isn't exactly the same but should at least prevent the builds from failing. @kellyselden you will likely want to call this out somewhere in the RFC in terms of drawbacks.

Following up on other conversations above, I strongly prefer that we suggest and provide a solid default option for all projects (as this RFC proposes). There is no benefit to shipping the blueprints without any default CI configuration, because that just creates unnecessary friction. Ember and Ember-CLI as a community is built around shipping solid defaults. We don't refrain from shipping QUnit integrations just because someone might want to use Mocha, nor do we skip out on eslint or prettier because someone might choose another tool. If someone does not want to use the bundled CI config, including that file does not create reasonably more work for them than they were going to have to do anyway. They can simply delete the file and add their own. It does not make sense, however, for us to force ALL developers to have to take the extra step of explicitly creating their CI configs for every single project. It is such an integral part of a project that we would be doing the community a disservice by not including some sane default service configuration.

Given that GitHub's market share is is ~89% compared to [GitLab's market share](https://www.slintel.com/tech/source-code-management/gitlab-market-share#) of ~10%, we can comfortably conclude that the vast majority of projects are going to be using GitHub by default. As such, GitHub actions seems the natural successor to Travis for this RFC.

That said, I think it would be reasonable for motivated parties to write an RFC suggesting we build CI config generators for any number of common CI platforms. There is value to making it easy to configure CI even for non-default use-cases, but I think that should be a separate RFC altogether, as that would essentially be an enhancement on the current state of things regardless of whether we switch the default from Travis to GitHub Actions. After all, if we do nothing today, we could keep Travis as the default and build generators for GitHub Actions, CircleCI, Gitlab, or whatever other platforms are commonly used.

rwjblue (commented):
While I completely agree with the goal, I'd like to see this proposal move a _slightly_ different direction:

* Introduce a new `--ci-provider` flag to `ember new`/`ember addon`
* Commit to maintaining what we consider the most popular CI providers (I'm thinking TravisCI, GitHub Actions, GitLab CI, and CircleCI)
* Commit to allowing other CI providers (that are generally useful) to be maintained by the community (essentially "we will review and land if valid, but we do not guarantee that these work)
* These CI Providers would themselves be their own blueprints with a specific naming scheme (e.g. `ci-provider-<provider-name>`), this allows addons to ship with a host of additional providers that folks could use (including the ability for organizations to have internal addons with their own custom CI systems built in).
* Update the [interactive ember new](https://github.com/emberjs/rfcs/blob/master/text/0638-interactive-app-creation.md) RFC (which is not implemented yet), to include an additional prompt which defaults to GH Actions

I think this path provides much better extensibility (and some insulation from _this exact issue_ if GH Actions makes similar changes in the future). Using this strategy combined with `ember-cli-update`'s `config/ember-cli-update.json` will play **very nicely** for folks. 

@kellyselden - What do you think?

kellyselden:
Those are good ideas, but I don't think they help us get off Travis today. Travis is essentially unusable today, so I'd prefer shipping GitHub Actions, then work on extracting out blueprints. If we wait for the beautiful end goal, it may never get here.

rwjblue:
The thing I laid out doesn't have to take longer or delay the timeline at all. I'm mostly talking about _how_ to get to GH Actions by default, not the timing.

I'm happy to send a PR to your fork so we can take a look at it together.

elwayman02:
I like the idea for a `--ci-provider` flag as an enhancement, as long as we're still committing to using GitHub Actions as the sane default for when the flag isn't specified.

Ultimately, I agree with @kellyselden that there is a need to get rid of Travis as the default ASAP, and I'm not in favor of debating additional features that could be filed as separate RFCs to further enhance our CI functionality *after* landing this RFC. I think the changeset required to do what Kelly has proposed is simple, quick, and obvious, and the only thing that needs to be debated for it is whether GitHub Actions are the correct default. If we can all agree on that, then this RFC can move forward and we can continue to discuss the design of a flag, support for multiple CI providers, and so on.

I guess what I'm saying is, look at it this way. Today, we have Travis CI as the default. We could submit RFCs to enhance our CI integration in different ways, of which a `--ci-provider` flag is one example. Even if we do none of that, however, we still have a default config in our blueprint today. All this RFC needs to cover in scope is whether to change that default (and to which provider); anything else is an unnecessary broadening of scope that should be covered in another RFC, imo.

rwjblue:
> Ultimately, I agree with @kellyselden that there is a need to get rid of Travis as the default ASAP, and I'm not in favor of debating additional features that could be filed as separate RFCs to further enhance our CI functionality after landing this RFC. I think the changeset required to do what Kelly has proposed is simple, quick, and obvious, and the only thing that needs to be debated for it is whether GitHub Actions are the correct default. If we can all agree on that, then this RFC can move forward and we can continue to discuss the design of a flag, support for multiple CI providers, and so on.

I disagree. 😸

I think GitHub Actions is a perfectly reasonable default value, but supporting only a single provider fundamentally sets us up for this exact problem once again (vendor lock in, "favoritism", etc to boot). I'm opposed to supporting only a single provider, and would prefer to remove default CI setup all together (removing the "Travis is broken" issue) than continue the same mistake again.

Sorry. Once bitten, twice shy...

rwjblue:
Additionally, y'all are making a bit of a false dichotomy here. It isn't a choice of "just swap to GH Actions real quick or we will never land anything". The suggestion I mentioned above could easily be done within the same timeframe (e.g. 3.26'ish, which is the earliest any new RFC could land anyways).

jelhan:
> Today, we have Travis CI as the default. We could submit RFCs to enhance our CI integration in different ways, of which a `--ci-provider` flag is one example. Even if we do none of that, however, we still have a default config in our blueprint today. All this RFC needs to cover in scope is whether to change that default (and to which provider); anything else is an unnecessary broadening of scope that should be covered in another RFC, imo.

There is also the alternative to remove TravisCI configuration file from default blueprints _without_ providing an alternative _through_ the default blueprints. This would allow us to experiment with alternative solutions as community packages before.

I _think_ this RFC and the discussion ignores the complexity of _how_ to setup GitHub Actions as CI. It's easy to say that Ember CLI should include it. It's way more difficult to agree on a specific configuration. Let me give some examples for questions that came up when developing [create-github-actions-setup-for-ember-addon](https://github.com/jelhan/create-github-actions-setup-for-ember-addon):

- Should all jobs run in parallel? CI minutes are free for open source projects on GitHub Actions, you know.
- Or should they run in stages and fail fast to safe CI minutes? CI minutes are not free for everyone using GitHub Actions. It could get expensive for enterprise customers, you know.
- And what is about the environment? We may not pay for CI minutes but they have a climate cost for sure.
- And if we would run them in stages, what stages to use? Lint and primary tests together as we do currently in TravisCI? Or should we run lint as a separate stage before all others?
- How should we setup the caching? Should we cache `node_modules`? Or the global cache of the package manager? Or both? What is more performant? In which situations?

Not sure if we could consider all that questions (and many more) as implementation details.

jelhan:
I just came around another question. I'm sorry to be that guy who is questioning what seems to be the agreed way forward. But I fear that we are missing important points otherwise.

TravisCI must be explicit activated per-repository _additionally_ to creating the `.travis.yml` file. Is this the same for GitHub Actions if used in a private repository? Or does GitHub Actions run as soon as the repository contains a workflow? Wondering if there is the risk that it accidentally eats up CI minutes on private projects.

Some enterprise users complained in #596 that including a CI configuration in the blueprints can be annoying if it does not fit their needs. Especially as it's often recreated by Ember upgrades. This concern may also create costs.

elwayman02:
> TravisCI must be explicit activated per-repository _additionally_ to creating the `.travis.yml` file.

I don't think this is actually true anymore. I thought Travis switched (at least a year ago) to being able to automatically opt-in every repository as soon as you add the config file?

rwjblue:
> > TravisCI must be explicit activated per-repository additionally to creating the .travis.yml file.
> 
> 
> I don't think this is actually true anymore. I thought Travis switched (at least a year ago) to being able to automatically opt-in every repository as soon as you add the config file?

Ya, this is my understanding also.

rwjblue:
RE:

> Some enterprise users complained in #596 that including a CI configuration in the blueprints can be annoying if it does not fit their needs. Especially as it's often recreated by Ember upgrades. 

My proposal handles that concern nicely (`--no-ci-provider` or `--ci-provider=none`), as well as provide a very convenient way for enterprise customers to provide their own CI Provider that works well on their infrastructure (without that "it keeps coming back after every upgrade" problem).

rwjblue:
> I think this RFC and the discussion ignores the complexity of how to setup GitHub Actions as CI. It's easy to say that Ember CLI should include it. It's way more difficult to agree on a specific configuration.

Agree. The RFC should explicitly include the actual configuration file contents.

text/0696-ember-cli-github-actions.md
```diff
@@ -0,0 +1,58 @@
...
+It will be configured as close as possible to the existing Travis CI setup.
+Unlike Travis CI, there is no enable or authorization step.
+The presence of a file in .github/workflows/ is automatically recognized and run if hosted on GitHub.
+
+## How we teach this
+
+A blog post could be sufficient explaining this switch.
+Existing projects can continue to use Travis CI in its limited capacity.
+Those using ember-cli-update could choose to take the new GitHub Actions file or continue using the .travis.yml.
+I don't think we cover CI or Travis CI in our docs.
```
bertdeblock, line 40: We do mention the inclusion of a `.travis.yml` file [here](https://cli.emberjs.com/release/basic-use/folder-layout).

kiwiupover:
@kellyselden @rwjblue Where are we at with this RFC?

kellyselden:
In the nicest way possible: I don't have the motivation right now to change/increase the scope of this RFC. I can close it, or someone else can take it on.

NullVoxPopuli:
I can try updating things. ❤️

elwayman02:
I'm really not certain why this RFC is stalled like this, to be honest. I still don't see the blocking issues as actually blocking, but rather suggested enhancements that should be a separate RFC. If they were so easy to get done, it wouldn't be 6 months later with this very simple RFC still not merged and implemented. Can we move forward with the RFC as proposed, please? Travis is just completely broken on most community addons still using it, and people are struggling to manually convert their projects. We're doing active harm to the community by continuing to ship Travis as the default config.

NullVoxPopuli:
I'm gonna take a stab at suggesting a ci config tomorrow on here.

I also don't think we need to select which ci provider. That's scope creep and will bike shed this RFC for eternity 🙃

jelhan:
> I'm gonna take a stab at suggesting a ci config tomorrow on here.
> 
> Agree, @elwayman02. I also don't think we need to select which ci provider. That's scope creep and will bike shed this RFC for eternity 🙃

Are you suggesting to have a high-level CI configuration file, which is than turned by community tools into configuration for specific providers? That would be awesome.

Actually I tried something similar when implementing `create-github-actions-setup-for-ember-addon`. But it is not that easy if you want to support custom steps (code coverage check, deployment etc.) as well.

NullVoxPopuli:
> Are you suggesting to have a high-level CI configuration file, which is than turned by community tools into configuration for specific providers? That would be awesome.

I am not. Just a github-actions config. PR incoming to ember-cli momentarily

NullVoxPopuli:
https://github.com/ember-cli/ember-cli/pull/9569

For Both:
 - split linting and tests in to separate jobs to maximize parallel running to try to best decrease over all CI time
 - support `[skip ci]` magic commit messages
 - use Volta
 
For Apps:
 - build once for tests and production
 - cache dependencies between jobs, because GH Action's cache is faster than re-installing (even for small monorepos)
   - it's always easier to remove the cache and switch to install on every job that to figure out how to get the cache working
 - this makes it easier for folks to scale and add things like netlify deploy and deploy preview without relying on external deploy infrastructure (which can have predatory policy changes / send you bills when you've been on the free tier) -- example deploy config (not included in this PR): https://github.com/NullVoxPopuli/limber/blob/main/.github/workflows/ci.yml#L190-L276
 
 For Addons:
  - run on a weekly cron so that maintained projects can keep an eye on what things break over time (if any)

text/0696-ember-cli-github-actions.md
```diff
@@ -0,0 +1,58 @@
...
+
+## Drawbacks
+
+Some CI systems are agnostic to code hosting location, but I don't think you can use GitHub Actions without also placing the code there.
+This further ingrains the GitHub platform to people that may prefer a different code hosting provider.
+
+GitHub Actions also has a limit how much an open source project can use, but it resets every month.
+I haven't heard of anyone hitting the limit though.
+
+GitHub Actions doesn't support retrying individual jobs like Travis CI.
```
snewcomer, line 50: https://github.com/actions/runner/issues/432#issuecomment-879888732

Looks like it is rolling out!

bertdeblock:
What should happen with this RFC now that https://github.com/ember-cli/ember-cli/pull/9579 is merged?

snewcomer:
I think we can close it!

https://github.com/ember-cli/ember-cli/pull/9579

rwjblue:
Ya, seems good to me. Sorry this took so long folks.

PR #699, opened 2021-01-06
Add additional info in blueprint lifecycle hooks to support 3rd-party addons to provide a way to codemod blueprints to reduce copy-pasta

NullVoxPopuli:
[Rendered](https://github.com/nullvoxpopuli/rfcs/blob/codemod-blueprints/text/0699-codemod-blueprints.md)


text/0699-codemod-blueprints.md
```diff
@@ -0,0 +1,204 @@
...
+Release Versions: Leave as is
+Relevant Team(s): Fill this in with the [team(s)](README.md#relevant-teams) to which this RFC applies
+RFC PR: Fill this in with the URL for the Proposal RFC PR
+-->
+
+# Codemod blueprints
+
+## Summary
+
+Historically, blueprints are not extensible: they exist independently and con be
```
wagenet, line 27:
```suggestion
Historically, blueprints are not extensible: they exist independently and can be
```

wagenet:
@NullVoxPopuli I assume you'd like this to get a proper review?

NullVoxPopuli:
Yes! for extending blueprints, I still think codemodding the output would be stellar (and is the basis behind the [`ember-apply*`](https://github.com/NullVoxPopuli/ember-apply/) stuff

ef4:
I think it might make sense to close this and make this kind of extensibility a design goal for a next-gen blueprint system (something I've already been discussing with @NullVoxPopuli and others).

The problem with the existing blueprint system is that 
 - it leaks a lot of ember-cli internals that we don't want to maintain forever
 - those internals are extremely expensive to instantiate so the cli commands are very slow

These are the reasons that v2 addon do not (yet) have a way of including blueprints. When we add one, we want to make sure it doesn't share these problems.

It's not clear that we even need to be maintaining our own blueprint system, nothing about it is ember-specific and if somebody wants to survey the existing tools that are out there and suggest one as the new default, we could probably design a compatibility bridge that brings most existing blueprints across.

achambers:
@NullVoxPopuli Is this still relevant?

NullVoxPopuli:
I don't think so -- @mansona is leading the charge with blueprint updates

PR #702, opened 2021-01-10
Add eslint-plugin-qunit to ember-cli blueprint

bmish:
[Rendered](https://github.com/emberjs/rfcs/blob/a5ce66db52f8db49f07bba4ec4debc3b48a21110/text/0702-eslint-plugin-qunit.md).

[Implementation](https://github.com/ember-cli/ember-cli/pull/9392).

### DIFF
--- a/text/0702-eslint-plugin-qunit.md
+++ b/text/0702-eslint-plugin-qunit.md
@@ -8,17 +8,6 @@ Relevant Team(s): Ember CLI
 RFC PR: https://github.com/emberjs/rfcs/pull/702
 ---
 
-<!---
-Directions for above:
-
-Stage: Leave as is
-Start Date: Fill in with today's date, YYYY-MM-DD
-Release Date: Leave as is
-Release Versions: Leave as is
-Relevant Team(s): Fill this in with the [team(s)](README.md#relevant-teams) to which this RFC applies
-RFC PR: Fill this in with the URL for the Proposal RFC PR
--->
-
 # Add eslint-plugin-qunit to ember-cli blueprint
 
 ## Summary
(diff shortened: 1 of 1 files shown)


mehulkar:
It should only be added to a subset of files in `tests/` (using overrides), right? I think the node plugin gets a similar treatment?

bmish:
@mehulkar thanks for bringing this up. I agree that it would be good practice to limit the rules to the `tests/` folder with an override to reduce any potential unwanted impact (i.e. false positives, performance hit) (even though the rules already attempt to limit themselves to tests), so I have updated the RFC with this new configuration.

text/0702-eslint-plugin-qunit.md
```diff
@@ -0,0 +1,87 @@
...
+<!---
+Directions for above:
+
+Stage: Leave as is
+Start Date: Fill in with today's date, YYYY-MM-DD
+Release Date: Leave as is
+Release Versions: Leave as is
+Relevant Team(s): Fill this in with the [team(s)](README.md#relevant-teams) to which this RFC applies
+RFC PR: Fill this in with the URL for the Proposal RFC PR
+-->
```
rwjblue, line 20: Can probably delete this

text/0702-eslint-plugin-qunit.md
```diff
@@ -0,0 +1,87 @@
...
+}
+```
+
+## How we teach this
+
+We do not currently discuss linting in either guides.emberjs.com or cli.emberjs.com.
+
+Users will be able to find documentation for the new rules on the [eslint-plugin-qunit](https://github.com/platinumazure/eslint-plugin-qunit) GitHub page. Many IDEs will provide a link directly to the rule documentation on highlighted lint violations.
+
+Any violations of the new rules will be detected by `yarn lint:js` and in some cases autofixed by `yarn lint:js:fix`.
```
rwjblue, line 77:
```suggestion
Any violations of the new rules will be detected by `yarn lint` and in some cases autofixed by `yarn lint:fix`.
```

rwjblue (commented):
Sorry for missing this one! I think this is a good addition, will add to future agendas for conversation to see about moving forward.

rwjblue:
Discussed this today, and we are in favor of moving forward here.

PR #704, opened 2021-01-14
Deprecate Octane Optional Features

pzuraq:
[Rendered](https://github.com/emberjs/rfcs/blob/deprecate-optional-features/text/0704-deprecate-octane-optional-features.md)


rwjblue:
We discussed this at todays core team meeting, and are moving it into final comment period.

PR #706, opened 2021-01-14
Deprecate the Ember Global

pzuraq:
[Rendered](https://github.com/emberjs/rfcs/blob/deprecate-ember-global/text/0706-deprecate-ember-global.md)

### DIFF
--- a/text/0706-deprecate-ember-global.md
+++ b/text/0706-deprecate-ember-global.md
@@ -45,10 +45,11 @@ This ensures there is an _explicit_ dependency if it is used, and which will
 allow us to treeshake in the future if nothing imports `Ember`. In time, we will
 be able to deprecate the `Ember` object altogether.
 
-## Detailed design
+## Transition Path
 
 When the `Ember` object is defined on the global, we will create a getter for it
-that also issues a deprecation.
+that also issues a deprecation. Users who currently use the global will have to
+add `import Ember from 'ember';` at the top of the files in which they use it.
 
 ## How we teach this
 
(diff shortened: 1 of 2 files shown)


text/0706-deprecate-ember-global.md
```diff
@@ -0,0 +1,105 @@
...
+
+## Motivation
+
+The Ember global, available at `window.Ember` (or `globalThis.Ember` in all
+environments) is a mostly legacy API from before Ember adopted native ES
+modules. Modern Ember apps do not use the global, and in general it is mostly
+used by legacy code and occasionally in examples. However, it still exists and
+remains undeprecated.
+
+The primary motivation for deprecating the global object is for _tree shaking_
```
SergeAstapov, line 27: @pzuraq maybe I'm missing something, could you please describe how we could do tree shaking if we do `import Ember from 'ember';`? Would build system look which props get used from `Ember` objet imported that way (that *seem* like more complex that tree shaking based on imports)?

pzuraq: @SergeAstapov it's actually the opposite, if you do _not_ `import Ember from 'ember';` anywhere in your application, _then_ we can treeshake.

Today, we can't even if you do not import it explicitly, because any code anywhere could reference the Ember global dynamically.

SergeAstapov: @pzuraq thanks for clarification!

rwjblue:
We discussed this at todays core team meeting, and are moving it into final comment period.

xg-wang:
How is this going to affect `ember-inspector`? It accesses Ember app from the global https://github.com/emberjs/ember-inspector/blob/v4.3.4/ember_debug/main.js#L14

Ref: https://discord.com/channels/480462759797063690/486243207072710656/803662597835456542

pzuraq:
Ember Inspector uses a lot of intimate/private APIs, so we can expose something for it to read to get access to the things it needs. It doesn't need to be public, so we can figure out the details when the time comes.

PR #711, opened 2021-01-23
Deprecate Auto Location

### LINKED ISSUE #693: Deprecate Ember AutoLocation
Seems like PushState is supported in most browsers today:
https://caniuse.com/?search=pushState

So basically everyone using a browser supported by Ember will get `HistoryLocation` when they select `AutoLocation`.

The current detection code is this:

```ts
export function supportsHistory(userAgent: string, history: History): boolean {
  // Boosted from Modernizr: https://github.com/Modernizr/Modernizr/blob/master/feature-detects/history.js
  // The stock browser on Android 2.2 & 2.3, and 4.0.x returns positive on history support
  // Unfortunately support is really buggy and there is no clean way to detect
  // these bugs, so we fall back to a user agent sniff :(

  // We only want Android 2 and 4.0, stock browser, and not Chrome which identifies
  // itself as 'Mobile Safari' as well, nor Windows Phone.
  if (
    (userAgent.indexOf('Android 2.') !== -1 || userAgent.indexOf('Android 4.0') !== -1) &&
    userAgent.indexOf('Mobile Safari') !== -1 &&
    userAgent.indexOf('Chrome') === -1 &&
    userAgent.indexOf('Windows Phone') === -1
  ) {
    return false;
  }

  return Boolean(history && 'pushState' in history);
}

```

Would it make sense to deprecate `AutoLocation` and use `HistoryLocation` as the new default? 

https://api.emberjs.com/ember/release/classes/AutoLocation

We'd still want to retain `HashLocation` of course, so it can still be chosen.

In case anyone still need to switch between the two depending on availability they could import `modernizer` and check manually, before making their choice.

Let me know what you think! Good idea? Bad?

***

Somewhat related we could also remove `Ember.Location.registerImplementation` (been deprecated for several years now).

https://github.com/emberjs/ember.js/commit/087e21888a449a8086c69dbea9709677c798b0fa

sandstrom:
[Rendered](https://github.com/sandstrom/rfcs/blob/deprecate-auto-location/text/0711-deprecate-auto-location.md)

Closes #693

### DIFF
--- a/text/0711-deprecate-auto-location.md
+++ b/text/0711-deprecate-auto-location.md
@@ -69,13 +69,14 @@ if (
 
 ## Transition Path
 
-We'll begin with deprecating history, in next major release we
+We'll begin with deprecating auto, in next major release we
 set history as the default. Since auto is default today,
 and almost always resolve to history, most users won't need
-to do a thing and they'll still get history behavior in the
-next major release.
+to do a thing and they'll still get history location in the
+next major release (but set directly, without auto resolving
+to it).
 
-There are two groups to consider:
+There are two groups to consider though:
 
 1. People who need auto-detection (hash/history switching) and
 2. Those that have explicitly set location to 'auto'.
@@ -105,7 +106,7 @@ to anyway.
 ### Deprecated or already private (should also be removed)
 - Remove the `create` method on Location (replaced by registry)
 - `registerImplementation` method on `Location`
-- `supportsHistory` exported in location/utils.ts
+- `supportsHistory` function exported in location/utils.ts
 
 ## How We Teach This
 
--- a/text/0711-deprecate-auto-location.md
+++ b/text/0711-deprecate-auto-location.md
@@ -9,18 +9,7 @@ Relevant Team(s): Ember.js
 RFC PR: https://github.com/emberjs/rfcs/pull/711
 ---
 
-<!---
-Directions for above:
-
-Stage: Leave as is
-Start Date: Fill in with today's date, YYYY-MM-DD
-Release Date: Leave as is
-Release Versions: Leave as is
-Relevant Team(s): Fill this in with the [team(s)](README.md#relevant-teams) to which this RFC applies
-RFC PR: Fill this in with the URL for the Proposal RFC PR
--->
-
-# <RFC title>
+# Deprecate AutoLocation
 
 ## Summary
 
@@ -41,7 +30,12 @@ The Ember Router can use different mechanisms to serialize its state.
 The two common ones are 'history' and 'hash'. There is also a special
 location called 'auto' which does feature detection, preferring 'history'.
 
-Basically all browsers will get 'history':
+When the special 'auto' location was added (and made default), less than
+half of browsers supported the (then) new history location API. For seamless
+upgrade to history API when available, Ember created the special auto location.
+
+These days (many years later), basically all browsers will get 'history',
+since the History Location API is very widely supported, even by IE 11:
 https://caniuse.com/mdn-api_history_pushstate
 
 The check 'auto' is basically doing this:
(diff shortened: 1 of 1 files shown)


text/0711-deprecate-auto-location.md
```diff
@@ -0,0 +1,140 @@
...
+<!---
+Directions for above:
+
+Stage: Leave as is
+Start Date: Fill in with today's date, YYYY-MM-DD
+Release Date: Leave as is
+Release Versions: Leave as is
+Relevant Team(s): Fill this in with the [team(s)](README.md#relevant-teams) to which this RFC applies
+RFC PR: Fill this in with the URL for the Proposal RFC PR
+-->
```
rwjblue, line 21: Can probably delete this

text/0711-deprecate-auto-location.md
```diff
@@ -0,0 +1,140 @@
...
+
+# <RFC title>
```
rwjblue, line 23: Should have a title, maybe:

```suggestion
# Deprecate AutoLocation
```

text/0711-deprecate-auto-location.md
```diff
@@ -0,0 +1,140 @@
...
+all get it resolved to history anyway.
+
+## Background
+
+The Ember Router can use different mechanisms to serialize its state.
+The two common ones are 'history' and 'hash'. There is also a special
+location called 'auto' which does feature detection, preferring 'history'.
+
+Basically all browsers will get 'history':
+https://caniuse.com/mdn-api_history_pushstate
```
rwjblue, line 45: Might add some details about _why_ this exists at all (when we added AutoLocation, and made it default all of our browsers **didn't** support it).

text/0711-deprecate-auto-location.md
```diff
@@ -0,0 +1,140 @@
...
+```
+
+## Transition Path
+
+We'll begin with deprecating auto, in next major release we
+set history as the default. Since auto is default today,
+and almost always resolve to history, most users won't need
+to do a thing and they'll still get history location in the
+next major release (but set directly, without auto resolving
+to it).
```
rwjblue, line 77: I think you will have to call out changes to ember-cli blueprint (in order to avoid the deprecation), and that change will impact all users.

text/0711-deprecate-auto-location.md
```diff
@@ -0,0 +1,140 @@
...
+
+```js
+// app/router.js
+export default class Router extends EmberRouter {
+  location = (historyFeatureDetection() ? 'history' : 'hash');
+  // …
+}
+```
+
+For the second group, they simply need to replace their explicit choice
```
rwjblue, line 95:
```suggestion
For the second group, they need to replace their explicit choice
```

text/0711-deprecate-auto-location.md
```diff
@@ -0,0 +1,140 @@
...
+of 'auto' for 'history' in `locationType` in environment.js (or
+`location` in router.js), since history is what their 'auto' resolves
+to anyway.
+
+### What To Deprecate (and later remove)
+
+- The HistoryLocation class
```
rwjblue, line 102: Hmm, why would `HistoryLocation` be deprecated?

rwjblue: Maybe you meant `AutoLocation`?
```suggestion
- The AutoLocation class
```

text/0711-deprecate-auto-location.md
```diff
@@ -0,0 +1,140 @@
...
+- The method `detect` from the `Location` interface
+- Explicitly setting 'auto' as the preferred location.
```
rwjblue, line 104: I think we also need to deprecate **not** setting a location type at all (since that would force us to use 'auto').

rwjblue (changes requested):
Thanks for working on this!

text/0711-deprecate-auto-location.md
sandstrom: I tried and if `locationType: null` in environment.js the app won't boot. This is the error:

> Uncaught Error: Assertion Failed: Cannot call get with 'cancelRouterSetup' on an undefined object.

Same if I omit the key `locationType` altogether.

Do you still want me to mention deprecating not setting a location type? Maybe I'm missing some angle/scenario?

sandstrom:
@rwjblue Thanks! I've updated the text with your suggestions

text/0711-deprecate-auto-location.md
rwjblue: > I tried and if locationType: null in environment.js the app won't boot. This is the error:

Ya, we should fix it so it doesn't error 😄

sandstrom: Do you want me to change anything in the RFC related to this? 

In my view there is no need to deprecate not setting a locationType if that already throws an error (regardless of which one it throws).

If anything we could simply throw a more explicit error (e.g. "locationType" must be set).

rwjblue: > Do you want me to change anything in the RFC related to this?

No, I don't think the RFC needs more changes to accommodate that.

rwjblue (approved):

rwjblue:
Discussed with the core team, and moving this into final comment period.

PR #727, opened 2021-03-16
Deprecate `{{unbound}}` helper

bertdeblock:
[Rendered](https://github.com/bertdeblock/rfcs/blob/deprecate-unbound-helper/text/0727-deprecate-unbound-helper.md)

### DIFF
--- a/text/0000-deprecate-unbound-helper.md
+++ b/text/0000-deprecate-unbound-helper.md
@@ -1,0 +1,102 @@
+---
+Stage: Accepted
+Start Date: 2021-03-13
+Release Date: Unreleased
+Release Versions:
+  ember-source: vX.Y.Z
+  ember-data: vX.Y.Z
+Relevant Team(s): Ember.js
+RFC PR: 
+---
+
+# Deprecate `{{unbound}}` Helper
+
+## Summary
+
+This RFC proposes to deprecate the `{{unbound}}` template helper.
+
+## Motivation
+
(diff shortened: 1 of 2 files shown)


jdurand:
I think there is a good opportunity to teach people to use the `reads` decorator in this deprecation guide :

```javascript
// app/components/user.js

import { reads } from '@ember/object/computed';

export default class UserComponent extends Component {
  @reads('args.name') name;
}
```

bertdeblock:
@jdurand `reads` does not provide the same functionality though. `reads` still provides a one-way binding. If `args.name` changes, `name` will change as well. The `{{unbound}}` helper _removes_ the binding and only returns the initial value. Also in a Glimmer component, `reads` makes less sense as arguments are immutable anyways. Also, using computed macros in native classes is linted against.

pzuraq:
The `@reads` decorator is also incompatible with autotracking, in that you cannot read an autotracked getter unless you use `@dependentKeyCompat`:

```js
class Foo {
  @tracked bar;
 
  get baz() {
    return this.bar;
  }

  // This will not work
  @reads('baz') qux;
}
```

Computed macros will likely eventually be deprecated and either replaced with something simpler/not based on `computed`, or moved to an external library such as [macro-decorators](https://pzuraq.github.io/macro-decorators/).

FWIW, the [current iteration of the decorators proposal](https://github.com/tc39/proposal-decorators) would also mean we would have to add an `accessor` keyword for each getter, which really reduces the DX IMO:

```js
// app/components/user.js

import { reads } from '@ember/object/computed';

export default class UserComponent extends Component {
  @reads('args.name') accessor name;

  // vs
  get name() {
    return this.args.name;
  }
}
```

PR #728, opened 2021-03-18
Deprecate using tracked in classic classes

pzuraq:
[Rendered](https://github.com/emberjs/rfcs/blob/deprecate-tracked-classic-decorator/text/0728-deprecate-tracked-classic-decorator.md)

### DIFF
--- a/text/0000-deprecate-tracked-classic-decorator.md
+++ b/text/0000-deprecate-tracked-classic-decorator.md
@@ -1,0 +1,127 @@
+---
+Stage: Accepted
+Start Date: 2021-03-18
+Release Date: Unreleased
+Release Versions:
+  ember-source: vX.Y.Z
+  ember-data: vX.Y.Z
+Relevant Team(s): Ember.js
+RFC PR:
+---
+
+# Deprecate using tracked in classic classes
+
+## Summary
+
+Deprecates using `tracked` to create tracked properties in classic classes:
+
+```js
+const MyClass = EmberObject.extend({
+  someProp: tracked({ value: 123 }),
+});
+```
+
+## Motivation
+
+Using `tracked` within classic classes was seen as an interop step that would
+help to ease the transition from classic to native, at the same time as adopting
+Octane. However, it has seen little usage so far. In the ecosystem, there is not
+a single usage in any addon (according to Ember Observer), and in general it
+seems to serve a very niche case.
+
+Since this feature was implemented, we also realized that it may be very useful
+to use the function call form of `tracked()` to represent something other than a
+classic decorator. For instance, the [tracked-built-ins](https://github.com/pzuraq/tracked-built-ins)
+library allows users to create tracked versions of arrays, objects, maps, and
+sets with `tracked()`:
+
+```js
+import { tracked } from 'tracked-built-ins';
+
+class Foo {
+  @tracked value = 123;
+
+  obj = tracked({});
+  arr = tracked([]);
+  map = tracked(Map);
+  set = tracked(Set);
+  weakMap = tracked(WeakMap);
+  weakSet = tracked(WeakSet);
+}
+```
+
+This is a nice short hand which means users do not need to import classes like
+`TrackedObject`, `TrackedArray`, and so on for a reasonably frequent use case.
+
+Deprecating the usage of `tracked()` in classic classes would essentially allow
+us to reclaim its usage, and gives us the possibility of adding something like
+this API in the next major version of Ember. It also means we could do other
+things with the function call version of it, such as add additional args to pass
+to the decorator form.
+
+As a final note, this will also set us up to more cleanly separate out the
+`@glimmer/tracking` package from Ember, so that it can be built as a normal JS
+package using standard build tools and Embroider.
+
+## Transition Path
+
+### Deprecation Guide
+
+Using `tracked` as a classic decorator within classic classes has been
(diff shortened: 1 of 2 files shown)


NullVoxPopuli (approved):

rwjblue (commented):
This seems like a good thing to deprecate, but I'd rather see the deprecation be `until` 5.0 not 4.0.

pzuraq:
@rwjblue waiting until 5.0 would lock us into having to provide a shim around `@glimmer/tracking` until then, which I would prefer not to. Ideally, `@glimmer/tracking` could be completely independent of Ember sooner rather than later.

Does the alternative of exposing `trackedClassic()` from `@ember/object/compat` work as a short term solution? Effectively provide an easy-to-codemod alias for the existing functionality?

chancancode:
Is this really unused? I seem to remember us using it, though I would have to check to be really sure.

It also seems somewhat arbitrary to fall off a cliff on this particular feature and this feature *only*, I think we got a lot of millage out of the flexibility and simplicity of the decoupled migration model. It also breaks the general expectations about decorators in the framework (on this particular decorator *only*).

If it’s just about not loosing out on a good naming, this seems like too heavy-handed to me. Certainly there would be other just as good options like `new tracked(...)` (detect `new.target`) or just make it a different spelling (like `Tracked` to signify that it is a constructor).

I am also not sure if it’s actually a fundamentally good idea to overload this to mean two pretty different things. We did it with `{{action}}` and that’s kind of a mixed bag outcome (probably mostly bad). It would be somewhat harder to ask questions/google/lookup documentations, and things like typescript/vscode would have a somewhat less ideal UX.

pzuraq:
> I am also not sure if it’s actually a fundamentally good idea to overload this to mean two pretty different things. We did it with {{action}} and that’s kind of a mixed bag outcome (probably mostly bad). It would be somewhat harder to ask questions/google/lookup documentations, and things like typescript/vscode would have a somewhat less ideal UX.

@chancancode I believe this is an orthogonal point of discussion. Whether or not we decide to use `tracked({})` or `tracked([])`, it would be useful to have the option to, which is partially what this RFC is about.

rwjblue:
> @rwjblue waiting until 5.0 would lock us into having to provide a shim around @glimmer/tracking until then, which I would prefer not to. Ideally, @glimmer/tracking could be completely independent of Ember sooner rather than later.

I don't think this is correct? The definition of what the "invoke" does here (e.g. non-decorator use case) could absolutely be provided by `@glimmer/global-context` or any number of other things to avoid the coupling.

pzuraq:
That's a good point. It would be unfortunate to have to tie Ember in that way I think, specifically for a to-be-deprecated feature (classic classes), but it would work to solve that issue.

boris-petrov:
> In order to use tracked properties, you must convert to using a native class.

Does that mean that this will work:

```javascript
import Component from `@ember/component`;

class ClassicComponent extends Component {
  @tracked foo;
}
```

Or by `native class` you mean non-classic Ember stuff? If so, that should be re-worded.

pzuraq:
@boris-petrov that would work! This would only deprecating using the classic syntax, any classic classes that work as native classes would continue to work as native classes

sandstrom:
> Export a new value, trackedClassic, from @ember/object/compat. This would only be usable as a classic decorator in classic classes, and the current form of tracked() would still be deprecated.

Sounds like a good idea!

My gut feeling is that this is indeed rarely used, so the churn cost should be pretty low. Especially if `trackedClassic` would be around.

Speaking for our app, we've still got a good deal of classic components. We're gradually migrating, when we have other reasons to refactor code we'll rewrite with glimmer plus angle brackets and tracked. 

Just my view though, others may have another perspective.

runspired:
The classic syntax is indeed heavily used by many apps I've encountered, most commonly when migrating a very large app when the native class codemod cannot yet be run. It seems to be a transient middle state although a very useful one: the native class codemod doesn't take a lot of apps very far until they've cleaned up certain other patterns that cam take some time.

I don't know that i agree with overloading the signature but from the perspective of whether the classic syntax is used think the deprecation until should be the same as for classic syntax as many of these mid-migration apps are on latest Ember versions. Folks don't always upgrade right away and it would suck for there to be a mismatch here where they needed to downgrade or pin at an ember version because classic still worked but tracked on classic didn't.

wagenet:
Well this was not part of 4.0. Is the only thing blocking the version we do it in?

runspired:
@wagenet I think so

wagenet:
@chriskrycho would you think this falls under #832 or is it separate?

chriskrycho:
Roughly: classic classes themselves will go away. We could remove this earlier than those, though, so please do include it in that list!

wagenet:
I’ll add this to the #832 list but also leave it open for now.

chriskrycho:
This fell off my radar until I was doing a triage pass, but it pairs nicely with #876; I'm going to bring it to Framework this week. For good or for ill, it looks like it's going to end up targeting 6.0, since we are now in the 4.10 beta cycle and [we freeze deprecations at the `.10` minor release](https://rfcs.emberjs.com/id/0830-evolving-embers-major-version-process#freeze-deprecations-at-m10).

chriskrycho:
We talked about this today at Framework, and rather than pushing this direction (and deprecating things piecemeal) we are going to FCP to close—by getting rid of it holistically as part of what we're discussing around Classic on #832.

PR #730, opened 2021-03-23
RFC: Semantic Versioning for TypeScript Types

chriskrycho:
## Summary

Introduce a policy for semantic versioning for TypeScript types, which accounts for both the details of TypeScript's structural type system and the need to absorb breaking changes in the TypeScript compiler.

> This RFC proposes a definition of [Semantic Versioning][semver] for managing changes to TypeScript types, including when the TypeScript compiler makes breaking changes in its type-checking and type emit across a “minor”release. (Note that this RFC addresses *only* type checking and type emit, *not* the “transpilation”mode of the TypeScript compiler.)
> 
> While this RFC is being authored in the context of Ember.js’ [adoption of TypeScript as a first-class supported language][RFC], its recommendations are intentionally general, in hopes that these recommendations can be widely adopted by the broader TypeScript community.

[semver]: https://semver.org
[RFC]: https://github.com/emberjs/rfcs/pull/724

[Rendered text](https://github.com/chriskrycho/ember-rfcs/blob/semver-for-ts/text/0730-semver-for-ts.md)

## Notes

- For discussion of whether Ember should adopt TypeScript *at all*, see #724.
- For discussion of Ember's own support policy (for core packages and for supplemental packages), see the forthcoming RFC on the subject, for which this is required foundational material. In *this* thread, please focus on this design for SemVer for types.
- If you're a non-Ember user but have with an interest in this proposal, welcome! We'd love your feedback and input as we work to make a design here that benefits not only Ember but the entire TypeScript ecosystem!

## Background

This is adapted from the work done at https://github.com/typed-ember/ember-cli-typescript/pull/1158 – with significant changes made to the recommendations around the supported compiler version policy (recommending two alternative approaches) and adding details around strictness, module interop, and more.


text/0730-semver-for-ts.md
```diff
@@ -0,0 +1,912 @@
...
+A change to an exported symbol is *not* breaking when:
+
+-   a wholly new symbol is exported which was not previously exported and which does not share a name with another symbol of a different kind (type vs. value) with a symbol which was previously exported
+
+
+##### Interfaces, Type Aliases, and Classes
+
+A change to any type (user-constructable or not) is *not* breaking when:
+
+-   a new optional property is added to the type, since all existing code will continue working ([playground][new-optional-prop])
```
RyanCavanaugh, line 395: This isn't correct. Example
```ts
// User code
const obj = { name: "hello", size: 1 };
const p: MyInterface = obj;
```
```ts
// Library code
interface MyInterface {
  name: string;
  // Adding this breaks the user code
  size?: string
}
```

text/0730-semver-for-ts.md
```diff
@@ -0,0 +1,912 @@
...
+
+
+##### Functions
+
+For functions which return or accept user-constructable types, the rules specified for [Non-breaking Changes: Interfaces, Type Aliases, and Classes](#interfaces-type-aliases-and-classes-1) hold. Otherwise, a change to a function declaration is *not* breaking when:
+
+-   a function (including a class method or constructor) *accepts a less specific ("wider") type*, for example if it previously accepted only a `boolean` but now accepts `boolean | undefined`—since all existing user code will continue working and type-checking ([playground][wider-argument])
```
RyanCavanaugh, line 414: This isn't correct. Let's say you wrote this
```ts
// User code
const f: typeof libFunc = arg => {
  console.log(arg.toFixed());
};
```
```ts
// Library code

function libFunc(arg: number): void;
// Changing to this will break 'f':
// function libFunc(arg: number | string): void;
```

text/0730-semver-for-ts.md
```diff
@@ -0,0 +1,912 @@
...
+
+-   a function (including a class method) which *returns a more specific ("narrower") type*, for example if it previously returned `number | undefined` and now always returns `number`—since all user code will continue working and type-checking ([playground][narrower-return])
```
RyanCavanaugh, line 416: This isn't correct. You could have written this:
```ts
const arr = [libFunc()];
arr.push(undefined);
```
or
```ts
let m = libFunc();
if (cond) m = undefined;
```
This would also break code like
```ts
const r = func();
if (r === "") { // ...
```
when changing `func` from `string | number` to `number`

text/0730-semver-for-ts.md
chriskrycho: Ahh, this gets into the user-constructible type things, but it's a specific person of that which I had missed. Thank you!

text/0730-semver-for-ts.md
```diff
@@ -0,0 +1,912 @@
...
+
+-   a function (including a class constructor or method) makes a previously-required argument optional—since all existing user code will continue to work with it ([playground][optional-argument])
```
RyanCavanaugh, line 418: This isn't correct, since mocks of the the function or class would be new compile errors.

text/0730-semver-for-ts.md
```diff
@@ -0,0 +1,912 @@
...
+
+A change to any type (user-constructable or not) is *not* breaking when:
+
+-   a new optional property is added to the type, since all existing code will continue working ([playground][new-optional-prop])
+
+Any change to a non-user-constructable type is *not* breaking when:
+
+-   a `readonly` object property on the type becomes a *more specific ("narrower") type*, for example if it was previously `string | string[]` and now is always `string[]`—since all user code will continue working and type-checking ([playground][narrower-property]). Note that this includes a previously-optional property becoming required.
+
+-   a new required property is added to the object—since its presence does not require the consuming code to use the property at a type level ([playground][new-required-prop])
```
RyanCavanaugh, line 401: This isn't correct. The `"in"` operator can be used to discriminate unions. If you wrote
```ts
// User code
function fn(arg: { x: string } | LibType) {
  if ("x" in arg) {
    console.log(arg.x.substr(0));
  }
}
```
```ts
// Library code
type LibType = {
  somethingElse: boolean;
  // adding this breaks 'fn'
  x: number;
};
```

text/0730-semver-for-ts.md
```diff
@@ -0,0 +1,912 @@
...
+
+##### Interfaces, Type Aliases, and Classes
+
+A change to any type (user-constructable or not) is *not* breaking when:
+
+-   a new optional property is added to the type, since all existing code will continue working ([playground][new-optional-prop])
+
+Any change to a non-user-constructable type is *not* breaking when:
+
+-   a `readonly` object property on the type becomes a *more specific ("narrower") type*, for example if it was previously `string | string[]` and now is always `string[]`—since all user code will continue working and type-checking ([playground][narrower-property]). Note that this includes a previously-optional property becoming required.
```
RyanCavanaugh, line 399: Probably the nittiest of "this isn't correct", but fails for comparability reasons (`string | string[]` can be compared to `string | number`, but `string[]` can't)

text/0730-semver-for-ts.md
```diff
@@ -0,0 +1,912 @@
...
+
+-   a function (including a class method or constructor) *accepts a less specific ("wider") type*, for example if it previously accepted only a `boolean` but now accepts `boolean | undefined`—since all existing user code will continue working and type-checking ([playground][wider-argument])
+
+-   a function (including a class method) which *returns a more specific ("narrower") type*, for example if it previously returned `number | undefined` and now always returns `number`—since all user code will continue working and type-checking ([playground][narrower-return])
+
+-   a function (including a class constructor or method) makes a previously-required argument optional—since all existing user code will continue to work with it ([playground][optional-argument])
+
+-   changing a function from an arrow function declaration to `function` declaration, since it allows the caller to use `bind` or `call` meaningfully where they could not before
```
RyanCavanaugh, line 420: It's worth discussing method-syntax declarations here, since these introduce parameter bivariance. Again, failures can be induced, but this one is on the safer side.

text/0730-semver-for-ts.md
```diff
@@ -0,0 +1,912 @@
...
+
+#### Bug fixes
+
+As with runtime code, types may have bugs. We define a ‘bug' here as a mismatch between types and runtime code. That is: if the types allow code which will cause a runtime type error, or if they forbid code which is allowed at runtime, the types are buggy. Types may be buggy by being inappropriately *wider* or *narrower* than runtime.
+
+For example (noting that this list is illustrative, not exhaustive):
+
+-   If a function is typed as accepting `any` but actually requires a `string`, this will cause an error at runtime, and is a bug.
+
+-   If a function is typed as returning `string | number` but always returns `string`, this is a bug. It will not cause an error at runtime, since consumers must "narrow" the type to use it, and narrowing the type would not even be a breaking change. However, the type is incorrect, and it *will* require end users to do unnecessary work.
```
RyanCavanaugh, line 437: This is an interesting take and I think deserves some meditation.

Is a function that returns one of a fixed set of strings (e.g. `"POST"`, `"GET"`, or `"PATCH"`) but documented to return `string` incorrect?  By this definition, yes, and the types should be updated. *But* this paints the library into a bit of a corner because now adding `"DELETE"` to the return type is definitely a breaking change.

Some nuance is required to understand if the return type is "too wide" relative to the *actual runtime behavior* or "too wide" relative to the *implied contract of the function*. It comes down to intent - if you only return `string`s today but think it would make sense for the function to return `number`s in the future, then a too-wide return type is semantically correct per the API design even if it describes runtime scenarios that don't currently happen.

chriskrycho:
Thanks so much for the detailed feedback, @RyanCavanaugh! Deeply appreciated, and will update this to account for those in the days ahead!

text/0730-semver-for-ts.md
```diff
@@ -0,0 +1,912 @@
...
+To conform to this standard, a package must:
+
+- link to the final published version of this specification
+- specify the compiler support policy
+- specify the currently-supported versions of TypeScript
+- specify the definition of “public API” used by the library (e.g. “only documented types” vs. “all published types” etc.)
+- author and publish its types with `strict: true` and `noPropertyAccessFromIndexSignature: true` in its compiler configuration
+
+
+## How we teach this
```
MelSumner, line 514: @emberjs/learning-team-managers definitely want to have the input of the Learning Core Team here. Not sure if this information should be a separate repository.

chriskrycho: Input very welcome!

The reason I suggested to put it in a separate repo is that the goal is (explicitly!) for it not to remain an Ember-specific policy and design, much as semver.org is used by many projects. We need this, for our adoption process; but many others would benefit from it as well.

wagenet: It seems to me that the intent here is to actually build a specification that will hopefully be adopted well beyond Ember. I would love to see this happen and it seems to me that it would be a huge blocker to this goal if the specification only lived within Ember’s documentation.

wagenet: It strikes me that another way this process could go is to develop the repo independent of Ember first then make an RFC for adopting this practices within Ember. However, it’s going to be important for the project to have at least one major adoptee and if it can’t even get Ember to adopt then it’s pretty much a non-starter. So to me, that makes a good case for doing it the way this RFC has proposed.

wagenet: I think it’s also worth noting that this spec is mainly for package developers not end users. All end users really want to know is “we won’t break your types willy-nilly.” Of course, having an actual spec with details that they can read is useful, but I strongly suspect that most users won’t really care about the details. If we do it right, things will just work. Where the spec matters a lot more is for those making a package who actually have to make sure they don’t break stuff! All that to say, I don’t think there really needs to be a detailed explanation on the main Ember docs beyond “we adhere to TS SemVer.”

text/0730-semver-for-ts.md
```diff
@@ -0,0 +1,912 @@
...
+[3.5-breakage-plaground]: https://www.typescriptlang.org/play/?ts=3.5.1&ssl=1&ssc=27&pln=1&pc=40#code/GYVwdgxgLglg9mABAEwVAwgQwE4FMAK2cAtjAM64AUAlIgN4CwAUIonlCNkmLgO6KES5KpTxk4AGwBuuWgF4AfPWatWYyTJoBuFYgC+1HUz3NmEBGSiJiAT0GkKALgFEHuADx09SuSjRY8e2FtIA
+[3.5-mitigation-playground]: https://www.typescriptlang.org/play/?ts=3.5.1#code/GYVwdgxgLglg9mABAEwVAwgQwE4FMAK2cAtjAM64AUAlAFyKEnm4A8A3gL4B8ibAsAChEiPFBDYkYXAHcGRUhUqU8ZOABsAbrmqIAvD35DhI3Ks1VqAbkHCOVwR0GCICMlETEAnowW56P5nZuPRQ0LDwAxSsgA
+
+
+##### "Downleveling" types
+
+When a new version of TypeScript results in backwards-incompatible *emit* to to the type definitions, as they did in [3.7][3.7-emit-change], the strategy of changing the types directly may not work. However, it is still possible to provide backwards-compatible types, using the combination of [downlevel-dts] and [typesVersions]. (In some cases, this may also require some manual tweaking of types, but this should be rare for most packages.)
```
jeremy-w, line 771: Copy: Doubled “to”. Though “emit to type definitions” seems verbally weird and I’d expect more like “emission of type definitions”.

```suggestion
When a new version of TypeScript results in backwards-incompatible *emit* to the type definitions, as they did in [3.7][3.7-emit-change], the strategy of changing the types directly may not work. However, it is still possible to provide backwards-compatible types, using the combination of [downlevel-dts] and [typesVersions]. (In some cases, this may also require some manual tweaking of types, but this should be rare for most packages.)
```

text/0730-semver-for-ts.md
```diff
@@ -0,0 +1,912 @@
...
+      - ["Downleveling" types](#downleveling-types)
+      - [Opt-in future types](#opt-in-future-types)
+    - [Matching exports to public API](#matching-exports-to-public-api)
+
+
+## Motivation
+
+For TypeScript packages be good citizens of the broader, semantically-versioned JavaScript ecosystem, package authors need a useful definition of SemVer for TypeScript’s type system.
+
+This is somewhat more complicated than in other languages, even those other statically-typed languages with language-level SemVer guarantees (such as [Rust][rust-semver] and Elm), because TypeScript has an unusually flexible type system. In particular, its [structural type system][structural] means many more kinds of both breaking and non-breaking changes are possible than in languages with a [nominal type system][nominal]. Accordingly, this document proposes a definition of SemVer which accounts for the extra flexibility afforded by these features.
```
jeremy-w, line 72: Elm’s record types are structural IIUC. See for example the discussion of pattern matching in https://elm-lang.org/docs/records. I wonder how changes to them are treated by elm-package when automatically deciding the next package version?

jeremy-w (commented):
This is really great. I’m still somewhat confused by “user constructible”, and I’m curious where enums fit in. The practical, detailed, non-normative appendices are a big help.

chriskrycho:
Thanks for the feedback, @jeremy-w! (😬 at all the typo/wordo fixes!) I'll be integrating the suggestions etc. later this week. Great callout on `enum`s (`const` and regular); definitely a miss there. I believe enums, like classes, emit both type and value symbols—with the caveat that `const enum` gets converted to plain values at build time, and so doesn't have the possibility of runtime conflicts with the normal settings.

User-construct-ability requires a definition, I think, as well as a single choice of spelling. 😆 (Autocorrect converts "constructable" to "constructible," so it looks like that's what we'll be using.) As a brief summary for folks following this, though, the idea there is to provide package authors a way of publishing a type definition that's useful to end consumers while giving them the "out" of saying that the type is only constructible by the package itself, for example via a "factory"-style function.

As a motivating example from Ember, [the `Transition` type](https://api.emberjs.com/ember/3.25/classes/Transition) is a public type, but is only allowed to be constructed *by Ember*—specifically by the router. It's therefore public API, and removing fields or methods from it or changing them to a different type would be breaking. However, because only Ember's routing layer is *allowed* to construct a `Transition`, it's (mostly) safe to e.g. add a non-optional field to it. If a user has "constructed" a `Transition`, any breakage they experience at that point is on them. (I say "mostly" because I'm still chewing on Ryan's point about assignability safety for that case in general, but I *think* that the user-constructibility makes it safe? But I need to mull on it further.)

text/0730-semver-for-ts.md
chriskrycho: I had the same thought about Elm after I published this, and I may actually go poke at the relevant section of the build tooling to answer the question, as I would like to know! Unfortunately, the documentation of what Elm does [has gone missing](https://discourse.elm-lang.org/t/where-did-the-reference-for-package-semver-go/7122), and the info which existed there historically was [less detailed than I remembered](https://github.com/elm-lang/elm-package/commit/2076638d812d9dea641e6e695951495f523bb98f#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5L26-L48).

text/0730-semver-for-ts.md
jeremy-w: The use of typeof also suggests explicitly addressing how the semver contract (and public and user constructible documentation) applies to the anonymous types of exported values, like the typeof libFunc in this example.

jeremy-w:
That helps clear up the constructibility question a lot. “Client code should not attempt to directly create an instance of this type, but can work with instances provided by the library using their public API.” Explicitly stating the changes considered non-breaking is probably needed to avoid issues with breaking derived types as well, eg breaking union discrimination or extension with a conflicting type as raised in other comments.

text/0730-semver-for-ts.md
jeremy-w: The implementation looks to be at https://github.com/elm/compiler/blob/770071accf791e8171440709effe71e78a9ab37c/builder/src/Deps/Diff.hs#L128-L136. I think any change in the number of fields is treated as a major change, while there’s some notion of “compatible renamings”, but I’m not sure.

text/0730-semver-for-ts.md
jenweber: So, in terms of how we teach this, I like the idea of it being a separate repo when it's finished. Many successful specs are living documents that have changes over time, and I think it makes sense for those standards conversations to have a home that is not mixed in with conversation about other unrelated things (and vice versa). We could keep it as an RFC entry, but I think as a standalone it has a better chance of influencing the broader JS ecosystem to adopt the same things, versus creating their own RFC version of this with modifications.

An alternate approach is to make the separate repo now, and then link to it in this RFC as a proposal to adopt that document. That would make the Ember relationship a bit clearer, but I think the quality of feedback and iteration cycle really benefits from the full text being in the RFC. We just need to make sure there are clear, bold links to the other repo in this RFC later on.

jenweber:
I am a huge fan of having the SemVer definitions/expectations in writing somewhere. I don't have enough hands-on experience authoring public library types to comment on the specifics, but as an end user of TS, I can see how this is helpful. Thanks for writing it up!

tjjfvi:
It seems to me like many/most of these rules could be summed up by:
- Non-user-constructible types can be narrowed
- Non-user-consumable types can be widened
- Exports cannot be removed

And I'd be proponent of the following rules (not including them makes it almost impossible to change anything without causing a "breaking change"), though they don't always agree with what has been specified in the document:
- `typeof someValue` is not a user-constructible type
- Exports can be added

chriskrycho:
I like those first three much as summaries, @tjjfvi! Thank you for putting it so clearly.

As for the others, I think the `typeof` one could be repurposed/rephrased very slightly as a rule of thumb for consumers (though with the more specific spec text for the details). Adding exports is trickier for the reasons currently spelled out in the doc: it’s pretty easy to come up with cases where adding exports can break consumers. That said, a “don’t give things the same name as an import” rule of thumb for consumers would help here, and I can see an argument for simplifying the rules around value-only and type-only exports to simply say that unless *explicitly* specified by the package authors (thinking of cases where one would do this is weird, though) the name and type export are both considered “taken” if the other is exported.

***

(For lookers-on, I’ll be returning to this next week or the week after. I was busy wrapping up the quarter the last few weeks and am off this week!)

tjjfvi:
> Adding exports is trickier for the reasons currently spelled out in the doc: it’s pretty easy to come up with cases where adding exports can break consumers.

*Any* new export can break a consumer:

```ts
// myWrapperAroundLib.ts
export * from "lib"
export * from "./myUtils"

// myUtils.ts
export const something = 5;
```

If `lib` adds a `something` export, that will clobber the `something` export from `myUtils` in `myWrapperAroundLib.ts` and cause an error.

chriskrycho:
Ah, that’s a fair point, and one we need to explicitly address one way or another! I’ll mull on it further; at first blush I think that re-exports are an area the current spec design basically doesn’t account for at all!

wagenet:
Without having given it too much consideration, would it make sense to just ban `export *`?

tjjfvi:
`import *` can also have a similar problem

chriskrycho:
An update: I've been thinking on and talking with folks about especially the problems raised by Ryan, and I have the outline of some changes which I expect to land mid-to-late next week here, namely addressing the way that inference and *mutability* interact; that covers one subset of the issues Ryan addressed (though there are a couple others outstanding there, which neither I nor my regular collaborators here have yet come up with a robust solution for).

text/0730-semver-for-ts.md
```diff
@@ -0,0 +1,912 @@
...
+#### Non-breaking changes
+
+In each of these cases, some user code becomes *superfluous*, but it neither fails to type-check nor causes any runtime errors.
+
+
+##### Symbols
+
+A change to an exported symbol is *not* breaking when:
+
+-   a wholly new symbol is exported which was not previously exported and which does not share a name with another symbol of a different kind (type vs. value) with a symbol which was previously exported
```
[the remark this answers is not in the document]
chriskrycho: Your suggested reading here is the correct one. The distinction I was trying to get at is around the fact that this import might be *either* a type or value import:

```ts
import { Foo } from 'a-lib/foo';
```

So the *idea* was to capture the fact that you could introduce a new value symbol `Foo` (`const Foo = ...`) if you had previously exported only a type (`export interface Foo { ... }`), or vice versa, and that's distinct from a case where you had *no* export previously. The case where it shares a name with a symbol of the other type is discussed above.

text/0730-semver-for-ts.md
```diff
@@ -0,0 +1,912 @@
...
+    export function buildPerson(name: string, age: number): Person {
+      return new Person(name, age);
+    }
+    ```
+
+    </details>
+
+    This *cannot* be constructed outside the module. Note that it may be useful to provide corresponding test helpers for scenarios like this, since users cannot safely provide their own mocks.
+
+-   Document that users can create their own local aliases for these types, while *not* exporting the types in a public way. This has one of the same upsides as the use of the classs with a private brand: the type is not constructable other than via the module. It also shares the upside of being able to create your own instance of it for test code. However, it has ergonomic downsides, requiring the use of the `ReturnType` utility class and requiring all consumers to generate that utility type for themselves.
```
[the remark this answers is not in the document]
chriskrycho:
```suggestion
-   Document that users can create their own local aliases for these types, while *not* exporting the types in a public way. This has one of the same upsides as the use of the classes with a private brand: the type is not constructable other than via the module. It also shares the upside of being able to create your own instance of it for test code. However, it has ergonomic downsides, requiring the use of the `ReturnType` utility class and requiring all consumers to generate that utility type for themselves.
```

text/0730-semver-for-ts.md
chriskrycho:
```suggestion
When a new version of TypeScript results in backwards-incompatible change to *emitted type definitions*, as they did in [3.7][3.7-emit-change], the strategy of changing the types directly may not work. However, it is still possible to provide backwards-compatible types, using the combination of [downlevel-dts] and [typesVersions]. (In some cases, this may also require some manual tweaking of types, but this should be rare for most packages.)
```

text/0730-semver-for-ts.md
chriskrycho: This point is one I’ve mulled on a lot in the last two months. I think the way I need to phrase this in terms of *intent* rather than *implementation*. It's perfectly reasonable for an implementor to specify a wider type than the implementation currently uses precisely in support of future widening, esp. if they have an idea of how the implementation is planned to evolve over time. I would probably not *recommend* this for a post-1.0-style API, in that I think it's *helpful* to provide the minimal API—but that's a matter of judgment, rather than an actual bug.

text/0730-semver-for-ts.md
```diff
@@ -0,0 +1,912 @@
...
+
+    -   Given a *value*-only exported symbol, including `namespace` declarations, adding a *type* export with the same name as the *value* may break users' code: they may have imported the value and safely created a type of the same name. Their existing import will now cause a re-declaration conflict. Note that this is distinct from adding an entirely new type export where there was no type or value export previously, since the user could never accidentally introduce the conflict, and could work around the conflict using the `as` import specifier when introducing the import.
+
+    -   Given a *type*-only exported symbol, including `type`, `interface`, or `export type` for a type or value, adding a *value* export with the same name may break users' code: they may have imported the type and safely created a value of the same name. Their existing import will now cause a re-declaration conflict. Note that this is distinct from adding an entirely new value export where there was no type or value export previously, since the user could never accidentally introduce the conflict, and could work around the conflict using the `as` import specifier when introducing the import.
```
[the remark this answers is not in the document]
tjjfvi: No, because "type-only" imports still import it into the value space for `typeof` purposes:
```ts
import type { Component } from "react"
type X = typeof Component // ok
```

chriskrycho:
I've been mulling on the open points here for a while and it finally clicked for me this past weekend (I may just be slow!) that all of the inference + mutability issues ultimately come down to *variance* (as do *many* of the points I noted in the draft). For example, on function return types, I originally wrote this:

> a function (including a class method) which *returns a more specific ("narrower") type*, for example if it previously returned `number | undefined` and now always returns `number`—since all user code will continue working and type-checking

However, @RyanCavanaugh correctly pointed out that this would fail in this scenario:

```ts
const arr = [libFunc()];
arr.push(undefined);
```

The problem here is that `Array<T>` should be *invariant* (as e.g. the C♯ docs on variance note), because it's mutable/writable—and a `ReadonlyArray<T>` would be perfectly safe in this scenario. The usual rule of thumb is that read-only data types (“sources”) can be covariant, while write-only data types (“sinks”) can be contravariant, but mutable data types (which operate as both sources *and* sinks) must be invariant.

The best solution I've come up with here so far is:

1. Document this! We can *teach* the mental model.
2. Provide tooling which people can use to be safe about this.
    - For package authors, it should be possible to use API Extractor to analyze changes in the output and identify where variance comes into effect here.
    - For consumers of a package, we can provide a lint rule which handles this correctly (one can imagine a `@typescript-semver/plugin` etc.). Although many users will opt out of it for convenience, this will at least allow folks who *want* to be extra safe to "do the right thing", e.g. by declaring a given type as read-only, or as sufficiently-wide type to handle it.

I'll be integrating a discussion of variance into the text to account for this.

tjjfvi:
It's also worth noting that "user-constructible" and "user-consumable" flows through variance. E.g. if `(x: X) => Y` is a user-constructible type, `X` is a user-consumable type and `Y` is a constructible type.

text/0730-semver-for-ts.md
```diff
@@ -0,0 +1,912 @@
...
+
+
+
+#### Definitions
+
+<dl>
+<dt>Symbols</dt>
+<dd>
+
+There are two kinds of *symbols* in TypeScript: value symbols and type symbols. Value symbols are the symbols present in JavaScript: `let`, `const`, and `var` bindings; and `function` and `class` declarations. Type symbols are represented by `interface` declarations, `type` (type alias) declarations, and `class` declarations.
```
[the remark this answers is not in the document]
chriskrycho: I'll incorporate `enum` and `const enum` into the discussion here. In both cases, they introduce both type and value symbols.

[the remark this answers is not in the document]
chriskrycho: I incorporated those and `namespace` in 94af7e6.

text/0730-semver-for-ts.md
```diff
@@ -0,0 +1,912 @@
...
+
+No change to a type *documented as private* is a breaking change, whether or not the type is exported. *Documented as private* is defined in terms of the documentation norm of the package in question. Some packages may choose to specify that the public API consists of *documented* exports, in which case no published type may be considered public API unless it is in the documentation. Other packages may choose to say the reverse: all exports are public unless explicitly defined as private (for example with the `@private` JSDoc annotation, a note in the docs, etc.).
+
+</dd>
+
+<dt>User constructability:</dt>
+
+<dd>
+
+Exported types (interfaces, type aliases, and the type side of classes) may be defined by documentation to be user-constructable or *not*. For example, a package may choose to export an interface to allow users to name the type returned by a function, while specifying that the only legal way to construct such an interface is via the exported function: the type is *not* user-constructable. Alternatively, a package may export an interface or type alias explicitly for users to construct objects matching the type themselves: the type *is* user-constructable.
```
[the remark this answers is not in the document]
chriskrycho: I clearly need to do some digging and figure out why it's "-ible"/"-ibility" rather than "-able"/"-ability"! Fixed.

text/0730-semver-for-ts.md
```diff
@@ -0,0 +1,912 @@
...
+
+<dt>Functions:</dt>
+
+<dd>
+
+Unless otherwise specified, "functions" always refers interchangeably to: functions in standalone scope, whether defined with either `function` or an arrow; class methods; and class constructors.
+
+</dd>
+
+<dt>Privacy:</dt>
```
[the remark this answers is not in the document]
chriskrycho: I actually ended up going with using this as a way of defining "Public API"; see 6c6352f.

text/0730-semver-for-ts.md
```diff
@@ -0,0 +1,912 @@
...
+
+<dd>
+
+No change to a type *documented as private* is a breaking change, whether or not the type is exported. *Documented as private* is defined in terms of the documentation norm of the package in question. Some packages may choose to specify that the public API consists of *documented* exports, in which case no published type may be considered public API unless it is in the documentation. Other packages may choose to say the reverse: all exports are public unless explicitly defined as private (for example with the `@private` JSDoc annotation, a note in the docs, etc.).
+
+</dd>
+
+<dt>User constructability:</dt>
```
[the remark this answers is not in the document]
chriskrycho: I've pushed a revision to the text defining user-constructibility and public API in 892bc8b, but I'm not especially satisfied with the way I've phrased it. Further suggestions for revision most welcome.

chriskrycho:
Oh, that's an excellent addition. Much obliged.

text/0730-semver-for-ts.md
```diff
@@ -0,0 +1,912 @@
...
+
+
+### Conformance
+
+To conform to this standard, a package must:
+
+- link to the final published version of this specification
+- specify the compiler support policy
+- specify the currently-supported versions of TypeScript
+- specify the definition of “public API” used by the library (e.g. “only documented types” vs. “all published types” etc.)
```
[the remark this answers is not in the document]
chriskrycho: I've folded that discussion *into* the definition of public API, per your comments on that section of the definitions, so I think it's fine to leave it as a high-level bullet here – though perhaps it's worth including "user-constructibility" and "subclassing" as examples here since they're important?

text/0730-semver-for-ts.md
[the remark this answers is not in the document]
chriskrycho: Made this explicit in 8620856, crediting both of you for the discussion.

text/0730-semver-for-ts.md
chriskrycho: I always forget that method parameters are still bivariant; thanks. Addressed in 2a61af5.

text/0730-semver-for-ts.md
chriskrycho: Just integrated the notes here at 3a353db.

As an amusing aside for lookers-on: @jeremy-w and I are in a shared community online and we both independently got to the above link at about the same time and linked it to each other independently back in March. 😅 Love it! 🤜🏼🤛🏼

text/0730-semver-for-ts.md
```diff
@@ -0,0 +1,912 @@
...
+
+    -   Given a `namespace` export, changing it to a value-only export (that is, to an exported object) will break all nested type access, since types cannot be exported as nested members of a non-`namespace` values. (`namespace` exports cannot be directly converted to type-only exports.)
+
+-   changing an `interface` to a `type` alias will break any user code which used interface merging
+
+-   changing a `namespace` export to any other type will break any code which used namespace merging
+
+-   changing a `class` export to a type-only export will break any code which extended the class or constructed an instance of the class ([playground][class-to-type-only]), and changing a `class` export to a value-only export will break any code which referred to the class as a type ([playground][class-to-value-only])
```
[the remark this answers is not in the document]
chriskrycho: > These playground links suggest to me that this is observed rather than promised behavior. Is that the intended meaning? If so, should there be “as of TypeScript vX.Y” to anchor the observations?

I added a note in 68f2743 to clarify that it's a "sort of both" scenario here, since there is no spec for TS' type system—it's all implementation-defined, and I don't think that's likely ever to change (certainly not in the foreseeable future). I'll make a point to note the actual version for which these hold when the RFC is merged.

chriskrycho:
All right, I'm down to Ryan's last three (and by far the hardest) to address. Aiming to wrap them up in *some way* tonight.

Wanted to note on `export *` and `import *`, though:

- `export *` is totally an issue, and I'm going to push an update which addresses it, I think by forbidding it: I can’t actually see a scenario where a *different* library re-exports something legitimately like that. Normally, that kind of thing only shows up in a façade pattern for a library *itself*, or for an active collaborator which is willing to absorb those kinds of changes, e.g. Ember's re-exports of Glimmer APIs. (If someone has counter-examples, I’m open to them.)

- I don't *think* `import *` can be an issue, because per the expect it's necessary to name the imported namespace, which is the key difference between `import *` and `export *`.

text/0730-semver-for-ts.md
chriskrycho: I believe this is an example of a more general pattern of higher-order/type-level variance, though in a direction distinct from most of the others I’ve noticed, in that instead of making the discrimination *unnecessary* it makes it *useless*. :thinking:

chriskrycho:
We talked about this at today’s Framework Core team meeting and are moving it forward into Final Comment Period!

tjjfvi:
> `noPropertyAccessFromIndexSignature` -> `noUncheckedIndexedAccess`

Ah, that makes a lot more sense

tjjfvi:
I am an active member of the TypeScript community, and given (emphasis mine):
> (While this RFC is being authored in the context of [Ember.js’ adoption of TypeScript as a first-class supported language](https://github.com/emberjs/rfcs/pull/724), *its recommendations are intentionally general, in hopes that these recommendations can be widely adopted by the broader TypeScript community*.)

I feel that feedback from the TypeScript community at large is valuable.

---

I have significant concerns about the current content of the proposal, and though it may be workable for Ember, I think the current provisions would be unworkable for many TypeScript libraries.

Consider the imaginary package `frobnicate`, maintained by a friendly guy named Fred. It's a simple package, on the surface. Fred has chosen to follow Ember.js's Semantic Versioning for TypeScript Types in `frobnicate`. This is the source code of `frobnicate@1.0.0`:

```ts
export type FrobnicateOptions = {
  intensity: number,
  rng: () => number,
}

export function frobnicate(options: FrobnicateOptions) {
  // ...
}
```

This is perfectly reasonable code. The most notable thing about it is that it exports `FrobnicateOptions`, but this is usually considered good practice, as it allows users to use it in cases such as:
```ts
const frobnicateOptions: FrobnicateOptions = {
  intensity: 5,
  rng: () => Math.random() ** 2,
}
```

If `FrobnicateOptions` is not exported, users must resort to something like `Parameters<typeof frobnicate>[0]`, which is rather unattractive.

Soon, Fred gets a feature request – in the `rng` function, a user wants to be able to know some degree of state, without having to track it manually. This is a simple fix, Fred realizes; he can simply pass a number indicating how many times the function has been called before in this frobnication pass.

```ts
export type FrobnicateOptions = {
  intensity: number,
  rng: (callNumber: number) => number,
}

export function frobnicate(options: FrobnicateOptions) {
  // ...
}
```

Now, someone can create an `rng` such as `i => Math.random() * i`.

Fred is happy with this change, and seeks to release it as `frobnicate@1.1.0`.

However, to his horror, he realizes that **this would be a breaking change**:

> A change to any object type (user constructible or not) is breaking when:
> - a non-`readonly` object property's type changes in any way

[Source](https://github.com/chriskrycho/ember-rfcs/blob/semver-for-ts/text/0730-semver-for-ts.md#:~:text=A%20change%20to%20any,changes%20in%20any%20way%3A)

---

This would usually not be considered a breaking change, as `() => number` is assignable to `(x: number) => number`, and thus no type errors would occur in constructing code.

However, the current proposal treats all types as *user-consumable*, and so this change would break user code such as:
```ts
const options: FrobnicateOptions = {
  intensity: 5,
  rng: () => Math.random() ** 2,
}

options.rng()
```

Because now the type of `options.rng` expects a parameter.

---

(Note: I see mention of *user-consumability* in some of the above comments; was that once part of the document, and later removed, or was it proposed and then never implemented?)

I propose the following additions to the RFC:

A new concept, *user-consumability*.

Exported values from a library are, by default, user-consumable but not user-constructible, unless specifically noted otherwise in their type.

User-consumability of a type entails the following permissions:
- Reading properties of a value of the type
- Writing properties of a value of the type, if they are not `readonly`
- Calling a value of the type, if it is a function type

The terms *user-constructible* and *user-consumable* refer to ways of constructing and consuming the types without going through the library; for example, user-constructible types allow object literal construction, and user-consumable types allow property access.

In particular, "constructing" and "consuming" values by way of user-consumable library functions is always allowed.

Specifically:
- Users can "consume" a non-user-consumable type by passing it to a user-consumable function exported from the library
- Users can "construct" a non-user-constructible type by getting it from a user-consumable function exported from the library
- A user-consumable function may take non-user-constructible arguments, and its return type may be non-user-consumable.
- A user-constructible function may take non-user-consumable arguments, and its return type may be non-user-constructible.

---

(Terms other than *user-consumability* and *user-constructibility* may be worthwhile, as they are rather close to each other.)

tjjfvi:
Also, in the vein of requiring `noUncheckedIndexedAccess`, it may make sense to also require `exactOptionalPropertyTypes`

chriskrycho:
@tjjfvi this is excellent feedback—*exactly* what we’ve been hoping for! 💙 Do you mind opening a [Discussion](https://github.com/typed-ember/semver/discussions/categories/general) for both of these on [the repo for the spec](https://github.com/typed-ember/semver)? 🙏🏼

tjjfvi:
@chriskrycho Certainly; I didn't realize that that repo existed 😅. I've opened typed-ember/semver#3, typed-ember/semver#4, and typed-ember/semver#5.

PR #736, opened 2021-04-16
Clarify deprecation and interface in 0649

chriskrycho:
The previous version of the text suggested that the new interface with `for` and `since` would be mandatory *immediately*. That, however, would have been a breaking change which itself did not follow Ember's SemVer and deprecation rules! Clarify that invoking `deprecate` without the `for` or `since` options is deprecated until 4.0, and show this clearly in the provided TypeScript interfaces.

### DIFF
--- a/text/0649-deprecation-staging.md
+++ b/text/0649-deprecation-staging.md
@@ -189,7 +189,7 @@ interface DeprecationOptions {
 - for?: string;
 + for: string;
 - since?: {
-+ since?: {
++ since: {
     available: string;
     enabled: string;
   }
--- a/text/0649-deprecation-staging.md
+++ b/text/0649-deprecation-staging.md
@@ -121,8 +121,8 @@ interface DeprecationOptions {
   id: string;
   until: string;
   url?: string;
-+ for: string;
-+ since: {
++ for?: string;
++ since?: {
 +   available: string;
 +   enabled: string;
 + }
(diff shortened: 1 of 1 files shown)


text/0649-deprecation-staging.md
```diff
@@ -176,6 +179,23 @@ deprecate('Calling deprecate() without the \'since\' key is deprecated', false,
...
+
+```diff
+interface DeprecationOptions {
+  id: string;
+  until: string;
+  url?: string;
+- for?: string;
++ for: string;
+- since?: {
++ since?: {
```
jamescdavis, line 192:
```suggestion
+ since: {
```

text/0649-deprecation-staging.md
```diff
@@ -155,10 +155,13 @@ Stepping through the new options individually:
...
 
-If the `since` key is not provided (as is the case for existing deprecations),
-the deprecation is assumed to be `enabled`. This RFC also... wait for it... deprecates
-the usage of the `deprecate()` function without a `since` key. This meta deprecation
-will be enabled at the same time as it becomes available like this:
+  If the `since` key is not provided (as is the case for existing
+  deprecations), the deprecation is assumed to be `enabled`.
+
+This RFC also... wait for it... deprecates the usage of the `deprecate()`
+function without the `for` or `since` keys. This meta deprecation will be
```
jamescdavis, line 162: Shouldn't the example below also include a `for` key? (TBH, it's a little confusing as to what this example is attempting to demonstrate.)

chriskrycho: Yeah, this is what was there before and I was trying to make only minimal changes.

rwjblue (approved):

text/0649-deprecation-staging.md
mehulkar: The snippet below shows the implementation of a _missing_ `since` deprecation, using the syntax that is introduced in this RFC. It's... confusing to say the least 😢

PR #737, opened 2021-04-24
rfc: ember-data | state-machine-update

runspired:
Full Text: [Rendered](https://github.com/runspired/rfcs/blob/ember-data/state-machine-update/text/0737-ember-data-state-machine-updates.md)
Implementation PR: https://github.com/emberjs/data/pull/7475


runspired:
After discussion there is some disagreement over whether this is useful given once the RECORD_DATA_STATE and RECORD_DATA_ERRORS flags are activated we intend to deprecate the state-machine entirely. Adding new sub-states only to deprecate all of them soon thereafter feels somewhat awkward.

However, the boolean flag changes this proposes *are* states which need to be possible in the new world as well. We are going to investigate making the boolean flags work for the states proposed here without causing any changes to `stateName`, and ensure that they work in both worlds. We will revisit this with a mind to update the RECORD_DATA_STATE rfc if that cannot be achieved.

runspired:
Closing in favor of removal of the state machine entirely. The state machine has been refactored away from public APIs in emberjs/data#7258 and we will RFC a deprecation for `currentState.stateName`. The flags will remain public API on model and flag states can be made to correctly reflect truthiness in deletion / empty-state without this RFC.

PR #739, opened 2021-04-24
rfc: ember-data | deprecate-non-strict-relationships

runspired:
Full Text: [Rendered](https://github.com/runspired/rfcs/blob/ember-data/deprecate-non-strict-relationships/text/0739-ember-data-deprecate-non-strict-relationships.md)

### DIFF
--- a/text/0739-ember-data-deprecate-non-strict-relationships.md
+++ b/text/0739-ember-data-deprecate-non-strict-relationships.md
@@ -1,0 +1,195 @@
+---
+Stage: Accepted
+Start Date: 2021-04-23
+Release Date: Unreleased
+Release Versions:
+  ember-source: vX.Y.Z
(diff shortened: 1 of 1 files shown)


text/0739-ember-data-deprecate-non-strict-relationships.md
```diff
@@ -0,0 +1,195 @@
...
+*Before*
+
+```./before.js
+import Model, { belongsTo, hasMany } from '@ember-data/model';
+
+export default class Comment extends Model {
+  @belongsTo()
+  author;
+
+  @hasMany({ async: true, inverse: null })
```
rwjblue, line 47: I find it vaguely odd for the before to include both strict relationships (for this `hasMany`) and non-strict (for the `belongsTo` just above)

PR #740, opened 2021-04-24
rfc: ember-data | deprecate-non-strict-types

runspired:
Full Text: [Rendered](https://github.com/runspired/rfcs/commit/5d33bf84b99ea9b4874e08fcadca3daf2748a718)

### DIFF
--- a/text/0740-ember-data-deprecate-non-strict-relationships.md
+++ b/text/0740-ember-data-deprecate-non-strict-relationships.md
@@ -1,0 +1,84 @@
+---
+Stage: Accepted
+Start Date: 2021-04-23
+Release Date: Unreleased
+Release Versions:
+  ember-source: vX.Y.Z
+  ember-data: vX.Y.Z
+Relevant Team(s): ember-data
+RFC PR: https://github.com/emberjs/rfcs/pull/740
+---
+
+# EmberData | Deprecate Non Strict Types
+
+## Summary
+
+Deprecates when the `type` for a record provided by a user differs from the resolved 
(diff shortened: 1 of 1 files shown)


text/0740-ember-data-deprecate-non-strict-relationships.md
```diff
@@ -0,0 +1,84 @@
...
+RFC PR: https://github.com/emberjs/rfcs/pull/740
+---
+
+# EmberData | Deprecate Non Strict Types
+
+## Summary
+
+Deprecates when the `type` for a record provided by a user differs from the resolved 
+type, thereby removing the need to configure `ember-inflector` to treat `types` as `uncountable`
+in order to use plural model names, and removing the dasherization constraint.
```
hjdivad, line 18: I'd recommend clarifying what "resolved" means in this context.

hjdivad (approved):
definitely LGTM. Thanks for driving this @runspired

text/0740-ember-data-deprecate-non-strict-relationships.md
runspired: The resolved type means the on-disk file-path derived type. So for instance a model at `app/models/foos/otherUsers.js` has the resolved type `foos/otherUsers` while `app/models/foos/bar-user.js` has the resolved type `foos/bar-user`. For the latter example, `foos/barUser` and `foos/barUsers` could both currently be supplied to ember-data in most contexts today and still result in a successful lookup of the model. Said uses would be required to be refactored to use the resolved type.

I'll try to add a note to this effect to the RFC for clarification.

PR #741, opened 2021-04-24
rfc: ember-data | Deprecate Accessing Static Fields On Model Prior To Lookup

runspired:
Full Text: [Rendered](https://github.com/runspired/rfcs/blob/ember-data/deprecate-model-static-field-access-without-lookup/text/0741-ember-data-deprecate-model-static-field-access-without-lookup.md)


runspired:
see also #832

PR #744, opened 2021-05-04
RFC 0418 - Fix value of key `into`

MrChocolatine:
Just a small typo, `render()` uses `into: 'applcation'` while the above example shows the template `authenticated`.


MrChocolatine:
ping

PR #746, opened 2021-05-07
Enable Embroider

thoov:
[Rendered](https://github.com/emberjs/rfcs/blob/dbb7a7c5dc7a6698368f078aa66f97c4bddd4bcb/text/0000-enable-embroider.md)

### DIFF
--- a/text/0000-enable-embroider.md
+++ b/text/0000-enable-embroider.md
@@ -1,0 +1,190 @@
+---
+Stage: Accepted
+Start Date: 2021-04-27
+Release Date: Unreleased
+Release Versions:
+  ember-source: vX.Y.Z
+  ember-data: vX.Y.Z
+Relevant Team(s): Ember CLI
+RFC PR:
+---
+
+# Enable Embroider
+
+## Summary
+
+This RFC proposes enabling Embroider as the default build system for ember-cli.
+
+## Motivation
+
+The current build system provided by ember-cli (lib/ember-app.js) builds and combines your application and addon's code in a very dynamic manner. While this works, it comes with a cost of forcing additional complexities at build time while also bloating application sizes. Embroider moves away from this dynamic behavior by upgrading your addons to be statically analyzable into what is referred to as [V2 Packages](https://github.com/emberjs/rfcs/blob/master/text/0507-embroider-v2-package-format.md). Embroider can then operate on these packages to know exactly what is or is not needed to be a part of your application and can use that information to produce spec compliant JavaScript that all modern packagers (like Webpack) understand. At this point we can leverage the benefits of a particular packager such as Webpack's tree-shaking abilities and dynamic imports (to name a few).
+
+Additional information about the motivation and key features of Embroider can be [found here](https://github.com/embroider-build/embroider/blob/master/SPEC.md).
+
+## Detailed design
+
+This design will cover a few topics. The first discusses the explicit file changes that happen for your application to use Embroider. Next it will cover how both new applications move to Embroider and how existing applications will migrate to it. Lastly, it will cover some deprecated concepts that no longer apply under Embroider along with how we get community addon adoption.
+
+### Packages
+
+The following packages will be added to package.json:
+
+- @embroider/compat - Stage 1 is responsible for upgrading addons from their current implmentations ("V1") into a new format called "V2 Packages". This package also contains the entry point method that is invoked in your ember-cli-build.js file (shown below).
+- @embroider/core - Stage 2 is responsible for taking a collection of V2 packages and abstracting away the "emberness" into standard Javascript which all packagers can understand.
+- @embroider/webpack - Stage 3 of Embroider's design allows you to bring your own packager. By default Embroider ships with and recommends using Webpack
+
+### Build File Changes:
+
+Embroider replaces the logic provided by `ember-app.js`. By default it runs in a "safe" mode that provides the best backwards compatibility guarantees but prevents some optimizations from happening. This safe mode configuration is what we will ship by default:
+
+```diff
+-return app.toTree();
++const { Webpack } = require('@embroider/webpack');
++return require('@embroider/compat').compatBuild(app, Webpack);
+```
+
+### New application flow
+
+The above changes will be applied to the blueprint for `ember new`. Additionally, the build flag option of `--embroider` will exist so that the option will be recorded in `config/ember-cli-update.json` like:
+
+```json
+{
+  "schemaVersion": "1.0.0",
+  "packages": [{
+    "name": "ember-cli",
+    "version": "X.X.X",
+    "blueprints": [{
+      "name": "app",
+      "outputRepo": "https://github.com/ember-cli/ember-new-output",
+      "codemodsSource": "ember-app-codemods-manifest@1",
+      "isBaseBlueprint": true,
+      "options": [
+        "--embroider"
+      ]
+    }]
+  }]
+}
+```
+
+This will ensure that `ember-cli-update` only applies these changes to applications that are already on Embroider.
+
+### Existing application migration flow
+
+In order to make the migration as simple as possible, a codemod will be created. The codemod CLI will provide two commands: `preflight` and `migrate`.
+
+Example usage:
+
+`npx @embroider/cli preflight`
+`npx @embroider/cli migrate`
+
+#### Preflight
+
+The goal of this command is to give users an easy yes or no answer to the question: can I start migrating to Embroider? As an example, we can check if the project is on the minimal set of package versions of known "fixed" addons along with other static checks.
+
+#### Migrate
+
+The migrate command is purely to codemod the file changes above to get your app onto Embroider.
+
+### Deprecations
+
+While Embroider strives to be as backwards compatible as possible, some existing concepts will not directly apply under Embroider. These concepts largely rely around the fact that Embroider does not want to you care about the contents of your output. This means that concepts like fingerprints asset file names or modifing output paths of your dist assets will not be supported.
+
+#### Fingerprinting
+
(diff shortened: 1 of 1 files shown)


text/0000-enable-embroider.md
```diff
@@ -0,0 +1,190 @@
...
+
+#### Migrate
+
+The migrate command is purely to codemod the file changes above to get your app onto Embroider.
+
+### Deprecations
+
+While Embroider strives to be as backwards compatible as possible, some existing concepts will not directly apply under Embroider. These concepts largely rely around the fact that Embroider does not want to you care about the contents of your output. This means that concepts like fingerprints asset file names or modifing output paths of your dist assets will not be supported.
+
+#### Fingerprinting
```
thoov, line 92: Bringing over a comment thread from a [predraft version](https://github.com/thoov/rfcs/pull/1) of this RFC: 

@kategengler: 
Are other fingerprint options still supported?
 
@thoov:
Currently no. I can make this section state more explicitly that app.options.fingerprint are not enabled
 
@simonihmig:
I think we could (should?) point out that there is still quite a gap in the functionality, as Embroider will only fingerprint the basic JS and CSS files that it adds to index.html. broccoli-asset-rev would fingerprint arbitrary files, e.g. images, and rewrite all the references to those (as long as it can "see" the file paths, so they have to be literals, i.e. no dynamic concatenations logic or so), e.g in JS or CSS files. AFAIK there is no solution for this yet in Embroider-land.

Which is quite annoying as you cannot serve images with HTTP headers that enable long-term browser caching as you normally would. Or if you do, they are not invalidated when they change.

And people are probably quite used to this (the previous behavior), so when this does not work it would feel like a bug rather than a missing feature.

text/0000-enable-embroider.md
```diff
@@ -0,0 +1,190 @@
...
+
+#### Preflight
+
+The goal of this command is to give users an easy yes or no answer to the question: can I start migrating to Embroider? As an example, we can check if the project is on the minimal set of package versions of known "fixed" addons along with other static checks.
+
+#### Migrate
+
+The migrate command is purely to codemod the file changes above to get your app onto Embroider.
+
+### Deprecations
```
patocallaghan, line 88: I believe some of the existing Ember CLI build hooks used by addons are no longer supported in Embroider land. Can we be more explicit about what these hooks are, and if possible some guidance on how to migrate away from using them?

Should we also have official deprecations for these unsupported hooks?

timiyay:
Is the intent to introduce Embroider-by-default in Ember 4.0?

Our experience of many aborted attempts to start using Embroider is that the Ember addon ecosystem has some non-trivial incompatibilities, as @simonihmig mentions above.

Any packages related to SASS/CSS buildchains, in particular, have been showstoppers for us (`ember-component-css` and `ember-css-modules` are the main ones). `ember-cli-mirage` is also rife with dynamically-generated code that we need to work though.

All in all, is it viable to enable Embroider within v3 and expect it to be "not broken"?

NullVoxPopuli:
I have embroider on a few v3 projects and it works fine. Even with postcss+tailwind!

cah-briangantzler:
> 
> `ember-cli-mirage` is also rife with dynamically-generated code that we need to work though.
> 

Could you detail what you think some of that dynamically generated code is? Doing some work on ember-cli-mirage and would like to address those if possible.

timiyay:
@cah-briangantzler I'll circle back to our embroider PRs, revive it, and dig up the stack traces.

chriskrycho:
I’d just like to *very* strongly second @ef4 here:

> …the ideal amount of code you need to install a plugin is not zero lines. It's two lines (import it so you see where it comes from, and invoke it in the right place in your config). The difference may seem small, but it greatly eases learning, understanding, and empowering devs. People need to see how their choices come together. Having a single place that governs your build pipeline and seeing each of the dependencies that you chose interacting there is indeed more code to look at. But when you don't have it, the learnability of the system is far worse.

One of the *most* common complaints I hear, working as an infrastructure engineer at LinkedIn and supporting hundreds of engineers working on our app is that they can’t even figure out *how* to figure out how things fit together. We hire good engineers, many of whom come in *wanting* to learn and contribute, and a lot of them eventually give up.[^survivorship] The idea of “zero config” *sounds* nice, but “zero config” is much better approached as “the average app requires zero *manual* config to get up and running, *but all the config is available when you need to adjust it*”—and, equally, that all the config can be *followed*.

I would analogize this to the move we’ve made with Ember’s test infrastructure (with both QUnit and Mocha) over the last few years. The design of Ember’s test tooling in the 1.x era and most of the 2.x era had many fewer imports—but it was actually *harder* to work with in the long term as a result than our current system, which adds just enough explicitness to allow you to follow the chain when you need to. For the cost of a couple extra lines at the top of each test module, we gained the ability for people to understand *all* of the “magic” in the system. Getting rid of `moduleFor*` in favor of `setupTest` and `setupRenderingTest` was massively clarifying. Getting rid of special knowledge about kinds of registered items (components, routes, services, etc.) in favor of treating them all the same and allowing you to *choose* to layer in additional capabilities as appropriate (and providing good blueprints so you don’t have to understand that to get started) makes the system far easier to come up to speed on,t o change when you need your own capabilities, and (importantly) to change over time.

This doesn’t go forever. You can end up with *too much* explicitness for a given context. The goal is to be as (but only as) explicit as is appropriate for the domain we’re working in.[^from] We don’t need to be talking about memory allocation or handling CPU registers in our build config. But we *do* need the ability to defineall the pieces we’ve integrated into the build and how they relate to each other. And by the same token: it’s helpful as a ‘separation of concerns’ for a high-performance computing tool written in Rust to be *able* to reason about memory allocation… while having Cargo handle the vast majority of build work “transparently” (but configurably!).

---

One other note: I’ve spent much less time than Ed has with build tooling over the last few years. But I feel quite confident *precisely because* I lack his expertise that his claim about the ease of writing a Rollup or webpack integration is much, much higher than that of working with Broccoli plugins. I have bounced off of parts of Ember CLI and Broccoli *repeatedly*—enough to sometimes feel quite dumb as a result. By contrast, I’ve worked much more comfortably with both webpack and Rollup plugins, including in contexts where I was just building a small tool using one or the other from scratch. Add in the fact that in every case we will be speaking the *lingua franca* of spec-compliant JS, and I have *no* doubt that core ecosystem tooling will be *easier* to maintain, not harder, with this move.

[^survivorship]: It’s easy for folks who *have* gotten over the hurdles of learning to be dismissive of the challenges: this is [*survivorship bias*](https://en.wikipedia.org/wiki/Survivorship_bias).

[^from]: I’m borrowing this framing and language from Dave Herman, Yehuda Katz, and Aaron Turon.

jrjohnson:
I buy into that vision @ef4 / @chriskrycho. I'm not arguing with any of the points you've both made, they are excellent and good things. You're both talking about high level architecture and philosophy. Yay! Let's do this, it's great!

I'm talking about the practical concern that we're moving a lot of work to the edges without clear guidance on what that work is. I spend a significant amount of my OSS time updating addons and I can tell you that most of them are working very well in an almost completely unmaintained state. I fully support a move to standard build tooling and the less time I have to spend in AST explorer and debugger mode trying to worst through the build tree the better, but there is going to need to be a pretty big lift to get the top %% of addons on bored. I'm watching dedicated community members and maintainers try and get `ember-get-config` and `ember-cli-page-object` (for example) to work and it's not going super well. This is OK there will be pain in the transition and we can all pitch in and get this done, but if we have to do it all again when Webpack 6 is released it's going to be really draining.

I'm saying yes, let's go, but I'd like to see an update to this RFC or maybe just the communication around it that encourages a single path forward so we can all take the step together and then if something changes we'll all be ready to take the next step together. I just don't have a clear picture of exactly what *only addons that need to extend the build system* includes. It seems like a big chunk of addons I rely on need to do this now. Maybe they won't anymore, but I can't quite wrap my head around how to discover that.

chriskrycho:
The *vast* majority of addons shouldn't need any build time integration anymore. Many addons which “have a build” still won’t need build-time integration in the sense they have historically: they will just publish JS ready to be pulled in and used directly by consumers. (You also shouldn’t expect most addons to change their state of running on autopilot.) Looking at what e.g. `ember-cli-page-object` is doing, I don’t think any of the things it does in the build pipeline currently need to happen at all, and in fact they only exist because of a bad tendency in an earlier era of Ember to rewrite imports. 😑 The solution there, having spent only the last couple minutes looking at it, is almost certainly just to cut a breaking change which just uses spec-compliant, Node-resolveable import paths, and then there will be no need for build integration at all. That’s going to be the case for many of these.

`ember-get-config` is a *bit* less standard in that it's trying to provide something to an addon from the host. I can think of a variety of ways of doing that (including the one it recommends: using the DI container) which don't involve build integration. Many of these kinds of things can also just be provided via changes to APIs—the assumption that this kind of thing *should* be available this way just may not be a great assumption (even if it's sometimes a convenient one).

In yet other cases which have historically required build integration—think here of CSS Modules—we may be able to actually just drop *all* of our special sauce in favor of just taking advantage of things like the webpack css-modules-loader.

As these suggest, part of the trick here is reevaluating whether a given approach actually makes sense in the Embroider world. Sometimes they will, and it'll just involve a small tweak. Sometimes, they won't, and that's fine, too.

Net, it's not clear to me why an exact breakdown for every single migration out there—they’re all different!—should be a blocker for moving forward with this RFC. Tackling it via docs, guidance, etc. as we go forward: absolutely!

wagenet:
I'm not quite sure if Embroider is ready to be the default yet, but it's definitely moving that way. Let me bring this up with the core team again and see what the thought is here.

ef4:
Update from framework core review meeting: We intend to do this but we expect this RFC will be superseded by one that actually defines the new app format, with some details that aren't in here yet. The existing `compatBuild` based embroider leaves a lot of crufty features in place that we don't really want to keep.

Examples:
 - the HTML format refers to outputs of the build rather than inputs of the build
 - the config/environment.js system is needlessly complicated
 - users should be able to see and use standard config files for the underlying tools like babel, webpack or vite.

We expect to land each of these things in embroider first and then summarize them in an RFC to propose making that the default.

kategengler:
This has been superseded by #977

PR #748, opened 2021-05-13
RFC: Glimmer component `Signature` type

dfreeman:
Super-abridged summary: for TypeScript, this RFC proposes to change `GlimmerComponent`'s `Args` type parameter to a `Signature` type that can capture richer information about how a component can be invoked.

[Rendered text](https://github.com/dfreeman/ember-rfcs/blob/glimmer-component-signature/text/0748-glimmer-component-signature.md)

### DIFF
--- a/text/0748-glimmer-component-signature.md
+++ b/text/0748-glimmer-component-signature.md
@@ -13,7 +13,7 @@ RFC PR: https://github.com/emberjs/rfcs/pull/748
 
 In TypeScript, the `@glimmer/component` base class currently has a single `Args` type parameter. This parameter declares the names and types of the arguments the component expects to receive.
 
-This RFC proposes a change to that type parameter to become `Signature`, capturing more complete information about how components can be used in a template, including their expected **arguments**, the **blocks** they provide, and what type of **element(s)** they apply any received attributes and modifiers to.
+This RFC proposes a change to that type parameter to become `Signature`, capturing more complete information about how components can be used in a template, including their expected **arguments**, the **blocks** they accept, and what type of **element(s)** they apply any received attributes and modifiers to.
 
 This RFC is based in large part on prior work by [@gossi] and on learnings from [Glint].
 
--- a/text/0748-glimmer-component-signature.md
+++ b/text/0748-glimmer-component-signature.md
@@ -1,5 +1,5 @@
 ---
-Stage:
+Stage: Proposed
 Start Date: 2021-05-13
 Release Date: Unreleased
 Release Versions:
@@ -62,10 +62,7 @@ interface Signature {
     argName: ArgType;
     // ...
   };
-  BlockParams?: {
-    blockName: [param1: Param1Type, paramTwo: Param2Type, ...];
-    // ...
-  };
+  BlockParams?: [param1: Param1Type, paramTwo: Param2Type, /* ... */];
   Element?: AnElementType | null;
 }
 ```
@@ -104,14 +101,12 @@ export interface NoticeSignature {
     /** The kind of message displayed in this notice. Defaults to `info`. */
     kind: 'error' | 'warning' | 'info' | 'success';
   };
-  BlockParams: {
-    /**
-     * Any default block content will be displayed in the body of the notice.
-     * This block receives a single `dismiss` parameter, which is an action
-     * that can be used to dismiss the notice.
-     */
-    default: [dismiss: () => void];
-  };
+  /**
+   * Any default block content will be displayed in the body of the notice.
+   * This block receives a single `dismiss` parameter, which is an action
+   * that can be used to dismiss the notice.
+   */
+  BlockParams: [dismiss: () => void];
 }
 
 export default class Notice extends Component<NoticeSignature> {
@@ -129,7 +124,32 @@ A signature with no `Args` indicates that its component does not accept any argu
 
 ### `BlockParams`
 
-The `BlockParams` member dictates what blocks a component accepts and specifies what parameters, if any, it provides to those blocks.
+The `BlockParams` member dictates what blocks a component accepts and specifies what parameters, if any, it provides to those blocks. When `BlockParams` is a simple tuple type, that indicates that the component only yields to the default block:
+
+```ts
+BlockParams: [name: string];
+```
+
+```hbs
+{{yield "Tomster"}}
+```
+
+However, `BlockParams` may also be an object type indicating what parameters a component's named blocks provide:
+
+```ts
+BlockParams: {
+  header: [];
+  body: [item: T; index: number]
+}
+```
+
+```hbs
+{{yield to="header"}}
+
+{{#each items as |item index|}}
+  {{yield item index to="body"}}
+{{/each}}
+```
 
 A signature with no `BlockParams` indicates that its component never yields to any blocks.
 
@@ -242,9 +262,7 @@ The signature for this component might look like:
 ```ts
 export interface BlogPostSignature {
   Args: { post: Post };
-  BlockParams: {
-    default: [postTitle: string, postAuthor: string, postBody: string];
-  };
+  BlockParams: [postTitle: string, postAuthor: string, postBody: string];
 }
 ```
 
@@ -286,15 +304,33 @@ export default class FancyTable<T> extends Component<FancyTableSignature<T>> {
 }
 ```
 
+While the runtime design of named blocks currently only permits components to yield parameters out to them, community members have suggested possible ways of making blocks more akin to components themselves, potentially accepting `@args` or attributes, or even themselves accepting further nested blocks. Should such a design ever become reality, the shape of a signature might evolve to become more recursive. Today, however, there are many open questions about how such functionality would actually work for authors, and so we keep the proposed format here simple.
+
 ### `Element`
 
 The `Element` member of the signature declares what type of DOM element(s), if any, this component may be treated as encapsulating. That is, setting a non-`null` type for this member declares that this component may have HTML attributes applied to it, and that type reflects the type of DOM `Element` object any modifiers applied to the component will receive when they execute.
 
-A signature with no `Element` or with `Element: null` indicates that its component does not accept HTML attributes and modifiers at all.
+A signature with no `Element` or with `Element: null` indicates that its component does not accept HTML attributes and modifiers at all. While applying attributes or modifiers to such a component wouldn't produce a runtime error, it still likely constitutes a mistake on the author's part, similar to [passing an unknown key in an object literal][ecp].
 
 For example, [`{{animated-if}}`] would omit `Element` from its signature, as it emits no DOM content. Even if you invoked it with angle-bracket syntax, any attributes or modifiers you applied wouldn't go anywhere.
 
 On the other hand, [`<ResponsiveImage>`] would set `Element: HTMLImageElement`, as the element in its template that it ultimately spreads `...attributes` on to is an `<img>`.
+
+[`{{animated-if}}`]: https://ember-animation.github.io/ember-animated/docs/api/components/animated-if
+[`<ResponsiveImage>`]: https://github.com/kaliber5/ember-responsive-image#the-responsiveimage-component
+
+The `Element` member is of particular relevance for the modifiers that consumers can apply to a component. In a system using this information to provide typechecking, any modifiers applied to its component must be declared to accept the component's `Element` type (or a broader type) as its first parameter, or else produce a type error.
+
+- A component with `Element: Element` can only be used with modifiers that accept _any_ DOM element. Many existing modifiers in the ecosystem, such as `{{on}}` and everything in `ember-render-modifiers`, fall into this bucket.
+
+- A component with e.g. `Element: HTMLCanvasElement`, may be used with any general-purpose modifiers as described above _as well as_ any modifiers that specifically expect to be attached to a `<canvas>`.
+
+- A component whose `Element` type is a [union of multiple possible elements](#components-with-multiple-or-varying-elements) can only be used with a modifier that is declared to accept _all_ of those element types. This behavior is, in fact, the point—modifiers are essentially callbacks that receive the element they're attached to, and so the [normal considerations][variance] for typing callback parameters apply.
+
+[ecp]: https://www.typescriptlang.org/docs/handbook/interfaces.html#excess-property-checks
+[variance]: https://en.wikipedia.org/wiki/Covariance_and_contravariance_(computer_science)
+
+#### Components With Multiple or Varying Elements
 
 While it's not common, occasionally components might forward `...attributes` to different types of elements in their template:
 
@@ -306,10 +342,41 @@ While it's not common, occasionally components might forward `...attributes` to 
 {{/if}}
 ```
 
-For such cases, components can use a union type for their `Element`. In the case of the template above, the signature would have `Element: HTMLAnchorElement | HTMLSpanElement`.
+For such cases, components can use a union type for their `Element`. In the case of the template above, the signature would have `Element: HTMLAnchorElement | HTMLSpanElement`. Correspondingly, any modifiers used with such components would need to accept any of the possible types of DOM elements declared.
 
-[`{{animated-if}}`]: https://ember-animation.github.io/ember-animated/docs/api/components/animated-if
-[`<ResponsiveImage>`]: https://github.com/kaliber5/ember-responsive-image#the-responsiveimage-component
+Similarly, a component that may use `...attributes` on an `<a>` element or may not spread them at all might write: `Element: HTMLAnchorElement | null`. In such cases, it would be legal to use any modifiers that accept an `HTMLAnchorElement`, since they wouldn't ever be invoked for the `null` scenario.
+
+In cases where the distinction between possible elements is key to the functionality of the component and can be statically known based on the arguments passed in, the component author may choose to capture this as part of the signature at the expense of additional type-level bookkeeping.
+
+<details><summary>Gritty details</summary>
+The particular shape/value of arguments is something that varies from instance to instance of the component, and the standard tool in TypeScript for handling these cases is to introduce a _type parameter_ on the type(s) in question.
+
+For the `{{#if @destination}}` example above, the implementation might look like this:
+
+```ts
+interface MaybeLinkSignature<Destination extends string | undefined> {
+  Args: {
+    destination: Destination;
+    target?: string;
+  };
+  BlockParams: [];
+  Element: Destination extends string
+    ? HTMLAnchorElement
+    : HTMLSpanElement;
+}
+
+export default class MaybeLink<Destination extends string | undefined>
+  extends Component<MaybeLinkSignature<Destination>> {
+  // ...
+}
+```
+
+This would allow consumers to use a modifier that requires an `HTMLAnchorElement` on a `MaybeLink` if and only if the `@destination` arg they pass is definitely a string.
+
+Note, still, that general-purpose modifiers like `{{on}}` or `{{did-insert}}` would be usable with this component regardless of the type of `@destination`, or even if the author had simply typed `Element` as `HTMLAnchorElement | HTMLSpanElement` without the extra song-and-dance of explicitly capturing the `Destination` type.
+
+Finally, template analysis tools can provide escape hatches in the same vein as TypeScript's `@ts-ignore` and `@ts-expect-error` for these or any cases where consumers have information that library authors haven't encoded in the type system.
+</details>
 
 ## How we teach this
 
--- a/text/0748-glimmer-component-signature.md
+++ b/text/0748-glimmer-component-signature.md
@@ -1,12 +1,12 @@
 ---
 Stage:
-Start Date:
+Start Date: 2021-05-13
 Release Date: Unreleased
 Release Versions:
   ember-source: vX.Y.Z
   ember-data: vX.Y.Z
 Relevant Team(s): Ember.js
-RFC PR:
+RFC PR: https://github.com/emberjs/rfcs/pull/748
 ---
 
 ## Summary
--- a/text/0748-glimmer-component-signature.md
+++ b/text/0748-glimmer-component-signature.md
@@ -150,9 +150,7 @@ interface GlimmerComponentSignature {
     [argName: string]: unknown;
   };
   Blocks?: {
-    [blockName: string]:
-      | unknown[]
-      | { Positional: unknown[] }
+    [blockName: string]: unknown[]
   }
   Element?: Element | null;
 }
--- a/text/0748-glimmer-component-signature.md
+++ b/text/0748-glimmer-component-signature.md
@@ -54,7 +54,23 @@ The goal of this change is to enable tooling to provide a more complete picture 
 
 ## Detailed design
 
-Given a component with this template:
+This RFC proposes the structure for a `Signature` type that enables authors to answer the questions outlined above about their component.
+
+```ts
+interface Signature {
+  Args?: {
+    argName: ArgType;
+    // ...
+  };
+  BlockParams?: {
+    blockName: [param1: Param1Type, paramTwo: Param2Type, ...];
+    // ...
+  };
+  Element?: AnElementType | null;
+}
+```
+
+So, given a component with this template:
 
 ```hbs
 <div class="notice notice-{{@kind}}" ...attributes>
@@ -62,7 +78,7 @@ Given a component with this template:
 </div>
 ```
 
-This RFC proposes that a TypeScript backing class currently written like this:
+A TypeScript backing class currently written like this:
 
 ```ts
 import Component from '@glimmer/component';
@@ -105,17 +121,17 @@ export default class Notice extends Component<NoticeSignature> {
 
 By nesting the existing `Args` type under a top-level _signature_ type, we create a place to provide additional type information for other aspects of a component's behavior. One key element of this design is that eases future evolution of the signature, as including additional keys with new meaning won't be a breaking change.
 
-In addition to `Args`, two other signature members are proposed here: `BlockParams` and `Element`. A detailed explanation of each of these follows.
-
 ### `Args`
 
 The `Args` signature member works exactly as the top-level `Args` type parameter does in `@glimmer/component@1.x` prior to this RFC. That is, it determines the type of `this.args` in the component backing class, as well as acting as an anchor for human-readable documentation for specific arguments.
 
-A component that accepts no arguments can omit the `Args` key from its signature.
+A signature with no `Args` indicates that its component does not accept any arguments.
 
 ### `BlockParams`
 
 The `BlockParams` member dictates what blocks a component accepts and specifies what parameters, if any, it provides to those blocks.
+
+A signature with no `BlockParams` indicates that its component never yields to any blocks.
 
 The [yieldable named blocks RFC] and recent versions of the [Component guides] discuss blocks in some depth, but since named blocks in particular are relatively new to the community, brief definitions based on those in [RFC 678] are included here for clarity.
 
@@ -198,7 +214,7 @@ The [yieldable named blocks RFC] and recent versions of the [Component guides] d
      ```
 </details>
 
-The `BlockParams` type maps the block names a component accepts to a tuple type representing the parameters those blocks will receive. A component that never yields to any blocks may omit the `BlockParams` key from its signature.
+The `BlockParams` type maps the block names a component accepts to a tuple type representing the parameters those blocks will receive.
 
 As a concrete example, the `BlogPost` component in the [Block Parameters section] of the Ember guides looks like this:
 
@@ -272,7 +288,9 @@ export default class FancyTable<T> extends Component<FancyTableSignature<T>> {
 
 ### `Element`
 
-The `Element` member of the signature declares what type of DOM element(s), if any, this component may be treated as encapsulating. That is, setting a non-`null` type for this member declares that this component may have HTML attributes applied to it, and that type reflects the type of DOM `Element` object any modifiers applied to the component will receive when they execute. Omitting `Element` or explicitly declaring it as `null` indicates that a component does not accept HTML attributes and modifiers at all.
+The `Element` member of the signature declares what type of DOM element(s), if any, this component may be treated as encapsulating. That is, setting a non-`null` type for this member declares that this component may have HTML attributes applied to it, and that type reflects the type of DOM `Element` object any modifiers applied to the component will receive when they execute.
+
+A signature with no `Element` or with `Element: null` indicates that its component does not accept HTML attributes and modifiers at all.
 
 For example, [`{{animated-if}}`] would omit `Element` from its signature, as it emits no DOM content. Even if you invoked it with angle-bracket syntax, any attributes or modifiers you applied wouldn't go anywhere.
 
--- a/text/0748-glimmer-component-signature.md
+++ b/text/0748-glimmer-component-signature.md
@@ -151,6 +151,8 @@ BlockParams: {
 {{/each}}
 ```
 
+If a component also accepts both a default block _and_ other named blocks, it can specify the default block by name (`default: [...]`) in its `BlockParams` in exactly the same way a consumer of the component might use `<:default>` to pass a default block alongside named ones when invoking a component.
+
 A signature with no `BlockParams` indicates that its component never yields to any blocks.
 
 The [yieldable named blocks RFC] and recent versions of the [Component guides] discuss blocks in some depth, but since named blocks in particular are relatively new to the community, brief definitions based on those in [RFC 678] are included here for clarity.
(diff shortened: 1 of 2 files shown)


scottmessinger:
I've been using Glint (and loving it) and didn't realize `Element` was where `....attributes` goes. I'm +1 on a new name. Maybe `AttributesElement` or `ElementReceivingAttributes` or `ElementReceivingSplattributes`

NullVoxPopuli:
`ElementReceivingAttributes` or `AttributesElement` sounds good to me :100:

wagenet:
I think I prefer `AttributesElement`. It's clear enough (to me at least!) without feeling overly verbose.

wagenet:
I should also add that we've been using Glint for a few weeks and I love this approach. I'm fully on board.

jamescdavis:
How about `Splelement`?  😆

jamescdavis:
But seriously, I do agree that `Element` isn't super clear and I like `AttributesElement`. I could also go for `SplattributesElement` to make it even clearer (but still not too verbose). Splattributes may be an informal name for attributes spread on an element, but it's well known and has even made it into an RFC: #435.

lifeart:
`AppliedElementAttributes` | `TransferredElementAttributes` | `DOMAttributes` | `SupportedDOMAttributes`

chriskrycho:
Obviously we can bikeshed all day on this one. :joy: The two things I'd note:

- we should pick a name that is *relatively* short because TS (and JSDoc!) users will be typing it a lot
- we should be careful not to paint ourselves into a corner with the name—I can imagine splitting apart modifiers and other attributes at some point (something I've been pondering a bit after a bunch of Octane adoption experience), and we'd want to make sure our names allowed for that kind of change without *too* much confusion

chriskrycho:
Some feedback from a discussion with Framework Core about this on 2021/05/28, with my own commentary added in:

-   There's a tricky bit of overlap between *documenting* the possible "target element(s)" for a given item and providing an accurate *type* for those. For example, in the case where attributes may or may not be applied (depending on the arguments passed or other state in the system) it might make sense for documentation purposes to specify `Element: HTMLDivElement | null`, to make clear to consumers what they can(not) rely on in terms of `...attributes` placement. However, in that same scenario, passing *any* modifier will fail to type check, because modifiers must 

-   @chancancode was concerned about the fact that providing a type representing the union of potential targets for `...attributes` would cause some modifier applications at the component invocation site not to type check. For example if the component specifies `Element: HTMLMediaElement | HTMLDivElement`, then application of a modifier which requires `HTMLMediaElement` will fail to type check.

    This is not only true, it's also expected and desirable: indeed, we might say it's the whole point! Appplying a modifier to a component is the same as passing a callback to a function (and that's exactly how tools like Glint represent things under the hood).
    
    Accordingly, I'm recommending we update the text of the RFC in the following ways:

    -   Making the symmetry with callbacks explicit.
    -   Showing how a user invoking a component can handle this, e.g. by writing a modifier which is applicable regardless (e.g. by accepting `Element` and narrowing).
    -   Showing how a component author can handle it in the case where it can be statically known from the signature based on arguments (e.g. by providing a signature which is a union type and accepts the )
    -  Noting that tooling (like Glint) can (and indeed Glint already does) provide the same kinds of escape hatches which TS itself does, via `expect-error` or `ignore` pragmas.

-   In addition to the expected discussion about the right name for `BlockParams`, there was a discussion of whether we should provide nicer ergonomics for the most common case of having a single default block yielded from the consumer. In that case, folks suggested something like this might be nicer:

    ```diff
    interface Signature {
    + BlockParams: [string, number]
    - BlockParams: {
    -   default: [string, number]
    - }
    }
    ```

    I think we could make this work with e.g. Glint in a fairly straightforward way, where the short form is equivalent to the longer form. This also dovetails nicely with the other point raised as part of the discussion: how we might support further extensions to this design to support additions to blocks in the future. This is *not* something this RFC needs to support, in my opinion, other than to note that the design *is* extensible. We can imagine that having exactly the same kind of behavior as the shorthand for the `default` block, that is: being able to accept just a list of block parameters *or* being able to expand it into something else. (I'm using `Blocks` here, since `BlockParams` would be too narrow a name.)

    ```ts
    // simple `default` block yielded
    interface Signature {
      Blocks: [string];
    }

    // multiple named blocks:
    interface Signature {
      Blocks: {
        default: [string];
        header: [];
        closeButton: [{ onClose: () => void }]
      }
    }

    // a hypothetical future expansion supporting passing attributes; note the
    // symmetry with Signature itself
    interface Signature {
      Blocks: {
        default: [string];
        header: {
          Element: HTMLHeadingElement;
          Yields: []
        };
        closeButton: [{ onClose: () => void }]
      }
    }
    ```

-   There was also interest in avoiding the use of the `default` name for the default block specificiation where possible. My own point of view here is that this doesn't make sense: `default` is public API and the signature of a component written in TS should use exactly the same semantics as we've defined in general. This RFC is *not* the place to hash out future direction for block names, and the design proposed here could readily be codemodded in the future if some *other* RFC deprecated `default` or otherwise changed the public API.

text/0748-glimmer-component-signature.md
```diff
@@ -129,7 +124,32 @@ A signature with no `Args` indicates that its component does not accept any argu
...
+}
+```
+
+```hbs
+{{yield to="header"}}
+
+{{#each items as |item index|}}
+  {{yield item index to="body"}}
+{{/each}}
+```
```
chriskrycho, line 152: We should probably include how we handle `default` in the extended case as well. (Note that a "suggestion" doesn't work here b/c of the nested code blocks, so I'm just going to "quote" the suggested text instead.)

> However, `BlockParams` may also be an object type indicating what parameters a component's named blocks provide. In that case, the `default` block may be specified by name:
> 
> ```ts
> BlockParams: {
>   header: [];
>   default: [];
>   listItem: [item: T; index: number]
> }
> ```
> 
> ```hbs
> <h1>{{yield to="header"}}</h1>
> 
> <div class='body'>
>   {{yield}}
> </div>
> 
> <ul class='items'>
>   {{#each items as |item index|}}
>     <li class='item'>{{yield item index to="listItem"}}</li>
>   {{/each}}
> </ul>
> ```

text/0748-glimmer-component-signature.md
```diff
@@ -306,10 +342,41 @@ While it's not common, occasionally components might forward `...attributes` to
 {{/if}}
 ```
 
-For such cases, components can use a union type for their `Element`. In the case of the template above, the signature would have `Element: HTMLAnchorElement | HTMLSpanElement`.
+For such cases, components can use a union type for their `Element`. In the case of the template above, the signature would have `Element: HTMLAnchorElement | HTMLSpanElement`. Correspondingly, any modifiers used with such components would need to accept any of the possible types of DOM elements declared.
 
-[`{{animated-if}}`]: https://ember-animation.github.io/ember-animated/docs/api/components/animated-if
-[`<ResponsiveImage>`]: https://github.com/kaliber5/ember-responsive-image#the-responsiveimage-component
+Similarly, a component that may use `...attributes` on an `<a>` element or may not spread them at all might write: `Element: HTMLAnchorElement | null`. In such cases, it would be legal to use any modifiers that accept an `HTMLAnchorElement`, since they wouldn't ever be invoked for the `null` scenario.
```
chriskrycho, line 347: It's probably worth clarifying this slightly here by adding a note that ecosystem tooling which *consumes* this information should treat it safely accordingly.

chriskrycho:
Discussion in the 2021-06-25 Framework Core Team meeting: folks are generally happy with where this is at. @chancancode noted that he doesn't care for the name `BlockParams` because it doesn't fit the way he wants to think about teaching what a component invocation is, and he and @pzuraq proposed that we use `Blocks` with a (at least for now) notional desugaring from this:

```ts
interface SignatureA {
  Blocks: [string];
}

interface SignatureB {
  Blocks: {
    header: [];
    body: [string];
  }
}
```

to this:

```ts
interface SignatureA {
  Blocks: {
    default: {
      Params: [string]
    }
  };
}

interface SignatureB {
  Blocks: {
    header: {
      Params: [];
    };
    body: {
      Params: [string];
    };
  }
}
```

This would then be "forward-looking" in that it supports future expansion of the idea of `Blocks` without change, and it would also map to the way Godfrey (and perhaps others!) think of the model here.

I'm leaving this comment as representative of that discussion, *not* representative of my own opinion here. I expect we'll probably have some further mild discussion on this thread on that point, though!

chriskrycho:
For my part, I disagree with this approach pretty strongly, for two reasons:

1. I think that `Blocks` is basically unmotivated in terms of the actual behavior of the framework today
2. This *increases* the pedagogical burden for teaching people how to think about a component's signature.

I'll start with (2) and then circle back to (1). When reading `Blocks: [string, number]`, it's much less unclear compared to *either* `Yields` or `BlockParams` what the types in the tuple represent. (That goes for the expanded form with named blocks as well as the default case.) Teaching how to use TS with Ember components therefore would require teaching the (notional) desugaring into the expanded format. Minimally, our coverage in Glint as well in official docs would need to explain that the tuple of types for `Blocks` actually represent the parameters yielded into each block—which in turn would naturally raise the question of *why they aren't just named accordingly*. But the only motivation we can supply there the hypothetical possibility of future expansion in that direction.

That leads me to (1): while I think it's good to design in a way that allows us to expand in the future, I do not believe we should design this to have a *worse* pedagogical experience *now* for something which may or may not ever happen. What's more, the use of `BlockParams` (or `Yields`) now allows us to move to `Blocks` in the future should it become desirable, and in fact to do so in *exactly* the same kind of "desugaring". In particular, we should consider that it's possible we *never* expand blocks, in which case we would be stuck with a notional desugaring in our pedagogy that doesn't pay for itself whatsoever—whereas the migration path if we *do* expand blocks is trivially codemoddable to any such future expansion.

This asymmetry of costs means that I think very strongly that we should *not* use `Blocks` here. In my view, at most we should note that `BlockParams` (or `Yields`, which I still prefer as a more accurate name for what is actually happening in a template!) could easily be migrated to `Blocks` *iff* future work requires it.

lifeart:
There is a lot of confusion about it, especially, comparing with other frameworks and standards.

In templates we have "yield", but we have "has-block" helper,
in documentation it named "named block", but it may be "unnamed" (aka "default")

Syntaxis will be different for both.

In other frameworks, same behaviour has name "slot".

In web API, it's also "slot" - https://developer.mozilla.org/en-US/docs/Web/API/Element/slot

It's literally hell for any new developers.

chriskrycho:
@lifeart yeah, we all spent some time thinking about the "slot" nomenclature a while back, and ultimately came down against because the behavior of slots in Vue, Svelte, and the web API is different enough from what named blocks do that it doesn't map in well. I believe there was some discussion of this on the original named blocks RFC as well. In any case, we should make the names used in the types match the names used by the framework, so the only real options are some combination of "blocks" and "yields" and maybe "params" with those. (If we renamed them at the framework level, we would of course also rename them in the types, but that's a different discussion for sure.)

gossi:
I think there are two perspectives to it:

- As an author, I use the mechanism of _yield_ to temporarily give control back
- As a consumer, I use named(!) blocks to control the contents of my component

Params are a mechnanics to transport data from inside the component to the outside.

Also, my _direction of reading_ is an author writes the type signature _for_ the consumers, in the "language" of the consumers.

text/0748-glimmer-component-signature.md
```diff
@@ -0,0 +1,461 @@
...
+When a developer goes to invoke a component in a template, there are a variety of questions about its interface they need to be able to answer in order to determine how to use it correctly.
+
+1. What arguments does this component accept?
+   - Which arguments are required and which are optional?
+   - What does each argument do?
+   - What types of values am I expected to pass for a given argument?
+2. How can I nest content within this component?
+   - Does it accept a default block? Named blocks?
+   - In what context will my block content appear in the component's output?
+   - For any blocks I write, what parameters does the component expose to them?
```
jelhan, line 34: The current proposal does not seem to address all of these questions. The provided type information does not give a hint about the elements allowed within the block in all cases. E.g. if the component yields within a button element, a developer must only use [phrasing content](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Content_categories#phrasing_content), which is not [interactive](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Content_categories#interactive_content).

In most cases this could be derived from the `Element` type. But only the component also applies splat attributes to that element. This may not be desirable in all cases.

Also there is the case, in which a parent element, limits the permitted content of a child element. This is the case for any element, which allows transparent content (e.g. `<del>` and `<ins>`). In that case permitted content can not be derived from that element alone.

Not sure if that must be addressed by this RFC though. Maybe it is enough to explicitly name this limitation in the RFC.

text/0748-glimmer-component-signature.md
```diff
@@ -0,0 +1,461 @@
...
+export default class FancyTable<T> extends Component<FancyTableSignature<T>> {
+  // ...
+}
+```
+
+While the runtime design of named blocks currently only permits components to yield parameters out to them, community members have suggested possible ways of making blocks more akin to components themselves, potentially accepting `@args` or attributes, or even themselves accepting further nested blocks. Should such a design ever become reality, the shape of a signature might evolve to become more recursive. Today, however, there are many open questions about how such functionality would actually work for authors, and so we keep the proposed format here simple.
+
+### `Element`
+
+The `Element` member of the signature declares what type of DOM element(s), if any, this component may be treated as encapsulating. That is, setting a non-`null` type for this member declares that this component may have HTML attributes applied to it, and that type reflects the type of DOM `Element` object any modifiers applied to the component will receive when they execute.
```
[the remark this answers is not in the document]
chriskrycho: That’s correct. Note that the corresponding class is what’s used for regular HTML elements, too, so: `HTMLAnchorElement`, for an `a` tag, for example.

text/0748-glimmer-component-signature.md
chriskrycho: > The current proposal does not seem to address all of these questions. 

It does; the additional very interesting point you raise about children is an *additional* point. While we could revise the RFC to make it explicit, the answer is implicit in the semantics of templates and unaffected by this proposal: they are simply functions of HTML. This proposal is not trying to specify a way to guarantee that no invalid HTML can be written. It is simply interested in extending the set of things which can be statically known about it for documentation and type checking purposes.

It might be possible to add in a further layer of analysis to accomplish that kind of analysis at a later point, but it is not a current goal for this doc or even for Glint. And it’s worth more that with this proposal and Glint, we are already doing more than any other JS framework. (The one ecosystem I think has taken a swing at the “allowed children” problem is PureScript, though I don’t know if their exploration in that space ended up broadly used even there. It is a very difficult problem because there is no inherent relationship between parent and child elements in HTML syntactically, only semantically.)

chriskrycho:
Update for folks watching this issue as eagerly as I am: we *very* briefly discussed this in the today’s Framework Core team meeting and are planning to have it resolved within the next few weeks (hopefully, though of course this being software there are no guarantees, by the end of the month!). I fully expect we’ll come out of that with a design we’re all satisfied with, and ready to move forward. I’ll update again as soon as we get there!

chriskrycho:
# Spec Group Summary, 2022/03/08

A number of the Framework Core Team members interested in the topic (@wycats, @ef4, @locks, @chancancode, and I) met up today as part of our “spec group” meeting[^spec] to do a deep dive on this RFC and see if we could get consensus for a path forward, and I think we did! We started out by ID'ing things we knew we were agreed on going in:

1. The types and the teaching materials should agree: in terms of what we call things, and the mental model they provide.
2. The types we define (as with things we do with runtime code) need to be designed with an eye to future extensibility; we don’t want to lock ourselves in.

(Those were easy, because while we’ve had to work through *how* to implement those ideas, everyone has always shared those as starting points.)

The main things we needed to get alignment on the design for on were:

- What is the shape of the freely-extensible future design we can all get behind, even if it’s super verbose?
- How much "sugar" should we provide beyond that, and how should that sugar be packaged?

The other points we identified during conversation were:

- The Ember TypeScript team has seen over the years that introducing type utilities can seem more magical than they are, and here would also prefer to avoid introducing additional imports, *especially* type utilities.
- We want a notional `ComponentInvokable` to be an extension of the internal idea of an `Invokable` in general, and to be able to support all the things the VM supports with it, including both named and .
- `yield` as a keyword is not something we necessarily want to bake any *more* deeply into our teaching materials, so that’s a strong reason to avoid it here.
- While `Yields: [string]` is well-motivated—it’s very desirable to have sugar for that most common form of a single `default` block—it is also ambiguous: does `Yields: []` mean “there are no blocks” or “there is one block and it does not yield any params”? We don’t *require* that sugar to ship this, and per [Stroustrup’s rule](https://www.thefeedbackloop.xyz/stroustrups-rule-and-layering-over-time/) we can ship something somewhat more verbose and then use feedback during community adoption to finesse that part of the design.

The fully-expanded form of the type hierarchy that came out of that is:

```ts
// Roughly the shape of an existing definition within Glimmer
interface InvokableSignature {
  Args?: {
    named?: Record<string, unknown>;
    positional?: unknown[];
  };
}

interface InvokableComponentSignature extends InvokableSignature {
  // The `null` means "this does not use ...attributes"
  Element?: Element | null;
  Blocks?: {
    [blockName: string]: unknown[] | {
      positional?: unknown[];
    }
  }
}

interface InvokableGlimmerComponentSignature extends InvokableComponentSignature {
  Args?: {
    named?: Record<string, unknown>;
    positional?: [];
  }
}
```

We also agreed that it can make sense to then provide a narrower signature which de-sugars into that, more or less along the lines originally proposed here, for Glimmer Components, which we can then map back into the fully-expanded form using a type utility—the mapping is quite straightforward. This is the signature (where each item is optional all the way down):

```ts
interface GlimmerComponentSignature {
  Args?: Record<string, unknown>;
  Element?: Element | null;
  Blocks?: Record<string, unknown[]>
}
```

All the optional fields here and above indicate what required for a signature: *nothing*. All of these fields are totally optional.

We will include in the RFC (and, more importantly, in the teaching materials!) an explanation of this. The underlying tooling in Glint uses the public signature type as a convenient way to express what the contract for the type is, but beyond that it doesn't directly use the signature: it plucks the data off to wire it up.

With those changes, let’s see how this would affect a couple of the examples from the RFC text. The `NoticeSignature` example:

```ts
export interface NoticeSignature {
  Element: HTMLDivElement;
  Args: {
    /** The kind of message displayed in this notice. Defaults to `info`. */
    kind?: 'error' | 'warning' | 'info' | 'success';
  };

  /**
   * Any default block content will be displayed in the body of the notice.
   * This block receives a single `dismiss` parameter, which is an action
   * that can be used to dismiss the notice.
   */
  BlockParams: [dismiss: () => void];
}
```

Here is how that would look with the changes suggested by today’s discussion:

```ts
export interface NoticeSignature {
  Element: HTMLDivElement;
  Args: {
    /** The kind of message displayed in this notice. Defaults to `info`. */
    kind?: 'error' | 'warning' | 'info' | 'success';
  };

  /**
   * Any default block content will be displayed in the body of the notice.
   * This block receives a single `dismiss` parameter, which is an action
   * that can be used to dismiss the notice.
   */
  Blocks: {
    default: [dismiss: () => void];
  };
}
```

Note that this is a pretty small change! We’d similarly update all text about `BlockParams` to discuss `Blocks` instead.

Similarly, the `FancyTableSignature` example might look like this:

```ts
export interface FancyTableSignature<T> {
  Args: {
    /** The items to be displayed in the fancy table, each corresponding to one row. */
    items: Array<T>
  };
  Blocks: {
    /** Any header contents for the table, broken into cells. */
    header: [];

    /** Content to be rendered for each row, receiving the corresponding item and its index. */
    row: [item: T, index: number];
  };
}
```

For both `Args` and `Blocks`, it is straightforward to make them support this “sugar” for the `Record` (for `Args`) and the array/tuple (for `Blocks`) as well as an expanded form which names the `named` and `positional` bits explicitly, using a simple union type.

I’ve had a preliminary chat with @dfreeman and @jamescdavis, and they were also broadly amenable to this design. Our current goal is to make sure there aren’t any other concerns any of us spot with this, and to have an updated version for this week’s Framework Core meeting.l

@dfreeman also confirmed ([playground][pg]) that we can do everything we need to for this expansion-based design. Two notes here:

1. We can simplify *one* layer of this when we stop supporting Glimmer Component v1.0 with Glint (though we expect that to be quite a long time from now!).
2. This should not be a performance issue, despite the complexity of the type: TS should only be evaluating it once per component definition, *not* once per invocation.

One remaining open question we aim to resolve in an upcoming Framework Core team meeting is whether we need a further layer in `Blocks` for accuracy in the fully expanded form:

```ts
interface InvokableComponentSignature {
  // other details elided
  Blocks: {
    [name: string]: {
      Params: {
        positional: unknown[]
      }
    }
  }
}
```

This would match the structure of the top-level `Args`, and would also be explicit and provide an appropriate place to hook in future pieces of a `Block` signature. (This would not meaningfully change the design or the complexity of the mapping between the short form and the long form.)

[pg]: https://www.typescriptlang.org/play#code/JYWwDg9gTgLgBAbzgUwB5mQYxgFQJ4YDyAZnAL5zFQQhwDkaG2AtDAcnQNwBQ3AJlgA2AQyjI4mCADsAzvACi4NgC44AVynAAjmvEy8IAEYRBPNhjiKwbQoYBWWeAF5EcANpW2AXQD8qmFC65Dzc5uIA4sgwhFDygjLIADy2dgA0cADS6QBiwoKChsKYANYAfHAuGSioMMhSfDJwxch4EKQpcD5wKW4ZXnCqufmFJSEA9ABUE3DhwABudXDCUnhL+RAA7sh8cDIAFtAwe8s7xNC0bUu7wADmUsIwamLpAjJqN6KNwPAwEHDfjWIany1TAJ22cDCcAmY1C7Es6BOAGVbvdHmJEjhyi4ENw4JZBMgQHUYKoAERxIkksnVWr1RrNVqkHCdOA4NwUwnEqQwMn9VRSYGmPFwACCUBuMlUjMuLLQdIacDJ4slNIAPkrKdzeXANWSAEKCCAlGQ0sZjOAASUaR2AjWE1zuDyeyB8IvxXWVEtNtLqiplzLg5rgABEIMgbXs7Y60S64McFnAAAYqmRJuACYjAKTbN34-Os9le1X9eV+xpIABywmJfD8cAg9kcnDgAAUIDJvsBpHl6xpilJNlI3P0KMHxeIji04EapDdjvU1mJhHw8HmC-mulWa9tVJForF4kki6m+ek6NXa3R0p48ClHKUW+3OzBu-dBHuojE4glMRzT1457Pl2PaCNe7heOUZDuhuqjbrWqgnt6fJPh2IHvqoI7kDB+JwXAl67pYSh3k22CoS+b55Jho4wXhBF8Ih5HoVREHBCKhrGsUUpKhxJo0mW9JNC0sowVuOH5m4vHFP8UhCUybIclJppeF4iGKUafFeJJGnFKWNTluoUgDkOI7iQWW5tmhr6gWpBo6cp2mcTRG6wQpdmcQ5UleDwLnQQWqi3veZHcGQ4wWgAAjcgigMSUBjJI4DSCScAAIwAHQAAzcNmtRQMQRTiIQgh8EimAHCYqaICKZwQKochQNmNw8NB3DBmSwhqEc0CNZQ5wPDS9y1swoiSsw0iCHg2U8sgeUFXASIHLAC58KmhBSBNVX4qmeEwTVdUBI1PnYS1wYLFAxgJOkhidTOyDCHMPUQDdjDRZg3yWRRoFTbl+WYOI9Gretqy4lt3pwTB9Hgy5e27Adc5HfifnHbwp0zRd4iDdsw3emNQNwAAFEm8iIvUKJOuiyDpvsT3FXAhjiA6g4QGADaybaMgAJTfTNv3iNkQqAxtINimDm35sB1kYRBCP4TuDFiwWMP1YdMFI6FIpaiSApCkdSkBcRQUwM1vCMI4+BEMQiTE2CpOos6GJFSVZUQBV3qlKU+Mc2lvzyDoeTm8gJCJPz+SC3gHsczwpvYAHQfW8idsU4kC2HMtYfu573sQL7aj++wQch4I6ee1H6Bm-nlvx7b5MuokAPemtE0Z17Pt+4IseW4XxeR9w0e4BXVskyVie113DdA83Wc53nFvBwL49NyXKMWu1nWLT1ZxQCA-UZsg+XAvAhg6azE3c7Nf3zYtRwnEpjfA+x9lQ-mmYdYIpLuMSMgyMINzIPtDVzm8iKaCJ0LRnXRrvfeb86bH3GqsQmVdh41zEFTA4wIdj0yuEzFm0hIRRk5mfXmcBC63zxsLTWPJtb5COttVwkMiLWBIg4MiH1mIfmlmxfEesFaQNfu-YWBYJaUXYW4JWcMbhAILGrY2vcy4xwHogsmsYMQpyWjfeyd9J6t1zu3AeJCNETyXn3Dug8bZIOUUkfRHlNERynm3ExViTSaKXq1FeHUuoAJuL1LeO9kBcmSnAwhc1VHX3qBQ6IZCNb+MoXAAAEjgAAsgAGQAMLLDmMIGQ4SZGo3Oh2cQfiqQ8hPvAomQ8lH20prsNBtNMGMxdjgtm+CuY5R5nNQu4S75i24YFUiRsRS0PgoRXpzCjasMlixLC6t8ThNUPE5JaSpAZKydE-pLVjEKPKSPFRV9lqdInrY7RM9A6dyFPsxePcNmz0UdsyxZzVk2Mzkc3Rs8OkPMMT3VxlAhSrEgGAYEDxtjpCBPkVYjBwQ7E7Mg8QCCtnQtQTTDBDM4DYNZngu0LTprn3EPIM6eBbRzn1D-PxYtBkQzlnhMRnjOHiyssIzC3wiTayMDNSRNLZlxMSak9JmTskPw8k-fEL8D6CtpZ9KWbgMmCF0HVAwxhBBeFVsAkKJs5H92uXCixVs8UEpuES3+ggtHZ3sQonVUZCXEsNSXIAA


[^spec]: which we’ll chat more about going forward, because it is *not* exclusive to Core folks, though it is anchored by Core folks

text/0748-glimmer-component-signature.md
```diff
@@ -0,0 +1,619 @@
...
+  ember-data: vX.Y.Z
+Relevant Team(s): Ember.js
+RFC PR: https://github.com/emberjs/rfcs/pull/748
+---
+
+## Summary <!-- omit in toc -->
+
+In TypeScript, the `@glimmer/component` base class currently has a single `Args` type parameter. This parameter declares the names and types of the arguments the component expects to receive.
+
+This RFC proposes a change to that type parameter to become `Signature`, capturing more complete information about how components can be used in a template, including their expected **arguments**, the **block parameters** they provide, and what type of **element(s)** they apply any received attributes and modifiers to.
```
chriskrycho, line 16:
```suggestion
This RFC proposes a change to that type parameter to become `Signature`, capturing more complete information about how components can be used in a template, including their expected **arguments**, the **blocks** they provide, and what type of **element(s)** they apply any received attributes and modifiers to.
```

text/0748-glimmer-component-signature.md
```diff
@@ -0,0 +1,619 @@
...
+  ember-data: vX.Y.Z
+Relevant Team(s): Ember.js
+RFC PR: https://github.com/emberjs/rfcs/pull/748
+---
+
+## Summary <!-- omit in toc -->
+
+In TypeScript, the `@glimmer/component` base class currently has a single `Args` type parameter. This parameter declares the names and types of the arguments the component expects to receive.
+
+This RFC proposes a change to that type parameter to become `Signature`, capturing more complete information about how components can be used in a template, including their expected **arguments**, the **blocks** they provide, and what type of **element(s)** they apply any received attributes and modifiers to.
```
dfreeman, line 16:
```suggestion
This RFC proposes a change to that type parameter to become `Signature`, capturing more complete information about how components can be used in a template, including their expected **arguments**, the **blocks** they accept, and what type of **element(s)** they apply any received attributes and modifiers to.
```

dfreeman:
Thank you @chriskrycho, both for your work on building a consensus and for updating the RFC text with the results!

As Chris mentioned, I'm on board with this direction and super excited to see this move forward! One observation I have about the proposed expanded structure for blocks:

```ts
interface MyComponentSignature {
  Blocks: {
    default: {
      Positional: [...];
    };
  };
}
```

As it stands, the `[...]` there reads to me as describing a `default` "positional block" (whatever that would mean), because nothing about "params" or "parameters" ever appears in the path to that type.

If the Framework Core team decides in favor of the additional layer of nesting, that would solve the issue (and also has the nice property of making the structure for block params mirror that of `Args`)

```ts
interface MyComponentSignature {
  Blocks: {
    default: {
      Params: {
        Positional: [...];
      };
    };
  };
}
```

If we opt _not_ to go that route, though, I'd like to advocate for an alternate name for the `Positional` member, even if just something slightly longer like `PositionalParams`.

chriskrycho:
**TL;DR:** we have consensus to do exactly what @dfreeman suggested [above][comment]. We'll present the updated RFC at this week's Framework Core team meeting, with every expectation of moving to FCP since we've addressed all outstanding concerns with this! :tada:

[comment]: https://github.com/emberjs/rfcs/pull/748#issuecomment-1063848106

---

In last week’s Framework team meeting and today’s Spec meeting, we further discussed the tradeoffs for both teaching and ultimate design (including possible future directions for blocks)—with an eye to minimizing the number of pitfalls in the short term while making sure that we don't see any blockers for the long-term directions we're aware of. The net of that is the combination of our discussion from last week's spec meeting for the "short" form (what folks will normally author) and then exactly what @dfreeman suggested [in his most recent comment][comment] for the fully-expanded form.

For the short form users will normally write:

```ts
interface GlimmerComponentSignature {
  Args?: {
    [name: string]: unknown;
  };
  Blocks?: {
    [blockName: string]: unknown[];
  };
  Element?: Element | null;
}
```

And the fully-expanded form:

```ts
interface InvokableSignature {
  Args?: {
    Named?: {
      [name: string]: unknown;
    };
    Positional?: unknown[];
  };
}

interface InvokableComponentSignature extends InvokableSignature {
  // The `null` means "this does not use ...attributes"
  Element?: Element | null;
  Blocks?: {
    [blockName: string]: {
      Positional?: unknown[];
    }
  }
}
```

One key point we noted about the pedagogy here: we will need to be explicit that in the case where someone is *extending* another signature, the normal rules of function variance apply for block params: they are, after all, function arguments.

This design leaves us the flexibility to expand how we conceptualize blocks in the future, as well as to represent types as functions in the future if or when it makes sense to do so. At the same time, it keeps the footprint for *today* given the constraints users operate under *now*.

We also had what I think was a fairly productive conversation about possible future directions we might want to take around components, including blocks, and how that would work with Glint's design (very well, and indeed possibly *more* easily than today!).

chriskrycho:
**TL;DR:** we have consensus to do exactly what @dfreeman suggested [above][comment]. We'll present the updated RFC at this week's Framework Core team meeting, with every expectation of moving to FCP since we've addressed all outstanding concerns with this! :tada:

[comment]: https://github.com/emberjs/rfcs/pull/748#issuecomment-1063848106

---

In last week’s Framework team meeting and today’s Spec meeting, we further discussed the tradeoffs for both teaching and ultimate design (including possible future directions for blocks)—with an eye to minimizing the number of pitfalls in the short term while making sure that we don't see any blockers for the long-term directions we're aware of. The net of that is the combination of our discussion from last week's spec meeting for the "short" form (what folks will normally author) and then exactly what @dfreeman suggested [in his most recent comment][comment] for the fully-expanded form.

For the short form users will normally write:

```ts
interface GlimmerComponentSignature {
  Args?: {
    [name: string]: unknown;
  };
  Blocks?: {
    [blockName: string]: unknown[];
  };
  Element?: Element | null;
}
```

And the fully-expanded form:

```ts
interface InvokableSignature {
  Args?: {
    Named?: {
      [name: string]: unknown;
    };
    Positional?: unknown[];
  };
}

interface InvokableComponentSignature extends InvokableSignature {
  // The `null` means "this does not use ...attributes"
  Element?: Element | null;
  Blocks?: {
    [blockName: string]: {
      Positional?: unknown[];
    }
  }
}
```

One key point we noted about the pedagogy here: we will need to be explicit that in the case where someone is *extending* another signature, the normal rules of function variance apply for block params: they are, after all, function arguments.

This design leaves us the flexibility to expand how we conceptualize blocks in the future, as well as to represent types as functions in the future if or when it makes sense to do so. At the same time, it keeps the footprint for *today* given the constraints users operate under *now*.

We also had what I think was a fairly productive conversation about possible future directions we might want to take around components, including blocks, and how that would work with Glint's design (very well, and indeed possibly *more* easily than today!).

NullVoxPopuli:
why pascal case? instead of the existing camel case?

chriskrycho:
It’s in the RFC. 😉

chriskrycho:
We agreed to move this to FCP to merge in today's Framework Core team meeting!

simonihmig:
First, I love this! 😍

> These interfaces are not importable, because importing them does not give any value to consumers: use of extends does not add any constraints (because of the optional-ity discussed below).

What I don't quite get is the reason for this...

When I author the signature of a component, I would like
* to get autocompletion. E.g. I could type an uppercase `B`, and the IDE would expand this to `Blocks`. Hey, I'm old, I don't want to remember things! :wink:
* rule out as much as possible the likelihood for typos. Even if all the (root level) keys are optional, having the interface constrained would at least prevent me from adding *wrong* keys, e.g. accidentally writing `Block`. 

Take the following (intentionally broken) example:

```
export interface NoticeSignature {
  Element: 'foo';
  Args: 'bar';
  /**
   * Any default block content will be displayed in the body of the notice.
   * This block receives a single `dismiss` parameter, which is an action
   * that can be used to dismiss the notice.
   */
  Block: {
    default: [dismiss: () => void];
  };
}

export default class Notice extends Component<NoticeSignature> {
  // ...
}
```

If you paste that into the TS playground linked in the RFC, TS wouldn't complain at all. For `Args` this is less an issue, as my wrong typing will trigger errors as soon as I actually use `this.args`, but would it be true for `Block[s]`?

Had I been able to to write this as `export interface NoticeSignature extends GlimmerComponentSignature` then I would immediately see all my silly errors. 

We don't need to enforce users doing that, but why are we "actively" preventing it?

simonihmig:
> The problem with providing a base interface is that extend-ing one interface in another is different from implement-ing an interface in a class declaration. 

Oh geez, you are totally right! I was confusing that terribly somehow... 🙈

rwjblue:
Great work getting this over the line, thank you for the awesome discussion here folks!

Let's do this!!!! 💪 🎉 💯

PR #750, opened 2021-05-26
RFC: Deprecate Ember.assign

nlfurniss:
[Rendered](https://github.com/emberjs/rfcs/blob/168cc1b6209af84eed2a69d685db28c637b3f085/text/0750-deprecate-ember-assign.md)


text/0000-deprecate-ember-assign.md
```diff
@@ -0,0 +1,52 @@
...
+## Summary
+
+Deprecate `Ember.assign` starting in v4.0. Now that Ember is dropping support for IE11, we no longer need `Ember.assign` as a polyfill since `Object.assign`
+is available in all browsers that Ember v4.0+ supports ([CanIUse](https://caniuse.com/mdn-javascript_builtins_object_assign), [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign#browser_compatibility)).
+
+## Motivation
+
+The polyfill is no longer necessary and is a small amount of code that can be removed and remove another Emberism. Apps and addons can use `Object.assign` or object destructring depending on their browser support targets.
+
+## Transition Path
```
rwjblue, line 34: I think it would be good to detail the path forward by a few categories:

* Libraries - If you still support IE11 (e.g. existing Ember 3.0 usage), you probably should do `const assign = Object.assign || Ember.assign;`
* Apps - Once you've dropped IE11 support, just update `Ember.assign` directly to `Object.assign`

text/0000-deprecate-ember-assign.md
```diff
@@ -0,0 +1,52 @@
...
+# Deprecate Ember.assign
+
+## Summary
+
+Deprecate `Ember.assign` starting in v4.0. Now that Ember is dropping support for IE11, we no longer need `Ember.assign` as a polyfill since `Object.assign`
+is available in all browsers that Ember v4.0+ supports ([CanIUse](https://caniuse.com/mdn-javascript_builtins_object_assign), [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign#browser_compatibility)).
+
+## Motivation
+
+The polyfill is no longer necessary and is a small amount of code that can be removed and remove another Emberism. Apps and addons can use `Object.assign` or object destructring depending on their browser support targets.
```
rwjblue, line 32:
```suggestion
The polyfill is no longer necessary and is a small amount of code that can be removed and remove another Emberism. Apps and addons can use `Object.assign` or object destructuring depending on their browser support targets.
```

rwjblue (commented):
Thanks for picking this up!

text/0000-deprecate-ember-assign.md
nlfurniss: I'm proposing to only add this deprecation to Ember 4.0, so consuming apps will already have dropped IE11 support and Ember 3.x users won't see it.

Should I change that and assume the deprecation will land in Ember 3.x?

rwjblue: It seems very common that an addon may support apps that are using Ember 3.x and 4.x at the same time, or that the addon isn't ready to do a breaking change release (to drop IE11 support) just yet.

nlfurniss: Makes sense. Added some IE11/Ember 3.x verbage

text/0750-deprecate-ember-assign.md
```diff
@@ -0,0 +1,59 @@
+---
+Stage: Accepted
+Start Date: 2021-05-26
+Release Date: Unreleased
+Release Versions:
+  ember-source: vX.Y.Z
+  ember-data: vX.Y.Z
+Relevant Team(s): Ember.js
+RFC PR: https://github.com/emberjs/rfcs/pull/750
+---
```
chancancode, line 10: This seems incorrect

chancancode:
This was discussed at today's framework team meeting and we think this is ready for Final Comment Period! (The header still need to be updated for this to be merged, assuming it completes the FCP.)

text/0750-deprecate-ember-assign.md
chriskrycho: This is exactly the format specified by the combination of the process [outlined in the README](https://github.com/emberjs/rfcs#the-process) and [specified in the instructions](https://raw.githubusercontent.com/emberjs/rfcs/master/0000-template.md). If that needs tweaking, the instructions in the template are what we should tackle!

chancancode: The format is correct but the information was copied from a different RFC

nlfurniss: Which information? I keep flipping between this and the template to copy and I’m not getting it

chancancode: Ah. I see why. I was looking at the “Rendered” link which pointed to a specific outdated commit, which had an incorrect header that links to a different RFC issue. Can you fix the link in the PR description? Sorry for the confusion

nlfurniss: Ahh ok. Fixed

PR #753, opened 2021-06-15
Adds Logging and Debugging Context API

pzuraq:


text/0753-logging-and-debugging-context-api.md
```diff
@@ -0,0 +1,189 @@
...
+- We could use a different import path than `@glimmer/debug`. This could be
+  useful, as we may want to reserve that namespace for APIs like `assert` and
+  `deprecate` which _will_ be stripped from production, and co-mingling stripped
+  and non-stripped APIs could be confusing.
+
+- We could use a separate tree specifically for debugging context/information.
+  This would likely be very costly from a performance perspective, and would
+  likely be prohibitive.
+
+- We could only ship this API in development and not production, relaxing the
```
NullVoxPopuli, line 186: could there be an option to opt in to stripping this or other apis(?) out only in production?

pzuraq: I think the cost here would be minimal enough that it won't really matter, and I think there could be broader repercussions to consider if we allowed that. For instance, some apps/addons may assume it exists, and stop working if it does not. I also worry about increasing implementation cost, and increased cost for Embroider to add another config option. That said, I'm open to the idea, would definitely like to see what others think about it.

text/0753-logging-and-debugging-context-api.md
```diff
@@ -0,0 +1,189 @@
...
+
+Debugging context generators for a given object are set via
+`setDebuggingContextGenerator`. When you call `getDebuggingContext` on a
+particular object, it first checks to see if a generator was associated directly
+with that object. If one wasn't, it then checks the prototypes of that object.
+This way, debugging context generators can be associated with classes instead of
+individual objects. When one is found, it is called and passed the object that
+`getDebuggingContext` was called on.
+
+### setDebuggingContextGenerator
```
rwjblue, line 142: Can you update to include information about what happens if `setDebuggingContextGenerator` is called twice?

text/0753-logging-and-debugging-context-api.md
```diff
@@ -0,0 +1,189 @@
...
+      console.log(`Fetch request failed! This occured in:\n\n ${context}`);
+    }
+  }
+}
+```
+
+## Motivation
+
+Currently, it can be pretty difficult to understand what caused an error within
+an Ember application. Errors can occur almost anywhere at any time. They can be
```
snewcomer, line 45: This seems extremely useful for ember-data!

wagenet:
The general idea here seems really useful. Is there still interest in moving this forwards?

PR #757, opened 2021-07-21
Default Modifier Manager

NullVoxPopuli:


NullVoxPopuli:
gonna try to update this RFC this weekend with all the things I learned from the default-helper-manager RFC and phrase things from first-principles' perspectives, and then we can get this RFC movin!!!!

NullVoxPopuli:
Updated!

text/0757-default-modifier-manager.md
```diff
@@ -0,0 +1,276 @@
...
+
+# Default Modifier Manager
+
+## Summary
+
+Anything that can be in a template has its lifecycle managed by the manager pattern.
+Today, we have known managers for `@glimmer/component`, `@ember/helper`, etc.
+But what happens when the VM encounters an object for which there is no manager?,
+such as a plain function? This RFC explores and proposes a default behavior for
+those unknown scenarios when it comes to _modifiers_.
```
chriskrycho, line 20: I'd suggest summarizing what we're actually doing here: this should be a TL;DR for the rest of it.

That might look something like this:

---

Enable standalone functions to be used a modifiers, by introducing a default *modifier manager*.

Today, when defining a modifier to set scroll position when a condition is true—

```hbs
<div {{scroll when=@someCondition}}></div>
```

—we need to import `modifier` and define it this way:

```js
// app/modifiers/scroll.js
import { modifier } from 'ember-modifier';

export default modifier(function scroll(el, _, { when: shouldScroll }) {
  if (shouldScroll) {
    el.scrollTop(0);
  }
});
```

With the default modifier manager proposed in this RFC, we can simply write a function which receives the element and the arguments directly:

```js
// app/modifiers/scroll.js
export default function scroll(el, { when: shouldScroll }) {
  if (shouldScroll) {
    el.scrollTop(0);
  }
}
```

This complements the introduction of a [default helper manager][rfc-0756], and is especially nice when combined with strict mode templates via [ember-template-imports][eti] (or as proposed in [First Class Component Templates][fcct]), where functions.

[rfc-0756]: https://github.com/emberjs/rfcs/pull/756
[eti]: https://github.com/ember-template-imports/ember-template-imports
[fcct]: https://github.com/emberjs/rfcs/pull/779

---

This way you tell people what they *get* from this!

text/0757-default-modifier-manager.md
```diff
@@ -0,0 +1,276 @@
...
+
+```js
+import Component from '@glimmer/component';
+import { tracked } from '@glimmer/tracking';
+
+export default class Example extends Component {
+  @tracked numberOfClicks = 0
+
+  // when the number of clicks is odd, add a JS movement effect / animation
+  wiggly = (element) => {
```
chriskrycho, line 66: It'd be good to call out that this is the modifier with something like `// MODIFIER DEFINITION` or something.

text/0757-default-modifier-manager.md
```diff
@@ -0,0 +1,276 @@
...
+
+
+</details>
+
+
+## Detailed design
+
+_A Default Manager is not something that can be chosen by the user, but is baked in to the framework
+as a default so that a user doesn't have to build something to use a non-framework-specific variant
+of the three constructs: Helpers, Modifiers, and Components._
```
chriskrycho, line 145: Couple quick notes on this:

- You're not advocating for a default component manager anymore, so "three constructs" and "Components" probably need to go.
- The actual motivation is that the user doesn't have to use special framework-provided tools to build these, but can just use normal JS constructs in places where it makes sense.

text/0757-default-modifier-manager.md
```diff
@@ -0,0 +1,276 @@
...
+ - reduce boilerplate
+ - be easily portable to JS for developers' mental model of how template and JS interact.
+
+Which results in:
+ - default to positional parameters
+ - all named arguments are grouped into an "options object" as the last parameter.
+   this happens to align with the _syntax_ of modifier invocation where named arguments may not appear
+   before the last positional argument.
+
+#### Example with mixed params
```
chriskrycho, line 158: For all of these examples, I would suggest finding actual modifier use cases to demonstrate them. That makes the case much stronger. The scroll example I gave is a decent one; you might also grab some of the examples from the guides!

text/0757-default-modifier-manager.md
```diff
@@ -0,0 +1,276 @@
...
+type FnArgs<Args extends Arguments = Arguments> =
+  | [Element, ...Args['positional'], Args['named']]
+  | [Element, ...Args['positional'], {}];
+
+type FunctionModifier<
+  Args extends Arguments = Arguments,
+  ModifierArgs extends FnArgs<Args> = FnArgs<Args>
+> =
+  | ((...args: ModifierArgs) => void) // setup, with no teardown
+  | ((...args: ModifierArgs) => () => void) // setup with teardown
```
chriskrycho, line 206: For TS purposes, it'd be better to fold these into a single function type with a union on the return type:

```suggestion
  ((...args: ModifierArgs) => void | () => void)
```

Then use a comment to explain or such.

text/0757-default-modifier-manager.md
```diff
@@ -0,0 +1,276 @@
...
+```ts
+import type { CapturedArguments as Arguments } from '@glimmer/interfaces';
+
+type FnArgs<Args extends Arguments = Arguments> =
+  | [Element, ...Args['positional'], Args['named']]
+  | [Element, ...Args['positional'], {}];
+
+type FunctionModifier<
+  Args extends Arguments = Arguments,
+  ModifierArgs extends FnArgs<Args> = FnArgs<Args>
```
chriskrycho, line 203: I think we can simplify this a fair bit, and we also need to capture the case (not common, but does exist) which *only* takes the element.

```ts
type FnArgs<Args extends Arguments = Arguments> =
   | [Element, ...Args['positional'], Args['named']]
   | [Element, ...Args['positional'], {}]
   | [Element];

type FunctionModifier<ModifierArgs extends FnArgs> =
  ((...args: ModifierArgs) => void | (() => void));
```

[(playground)](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBDAnmApnA3nAwgQzDAVyhQBMBBKAcwJBQDsYBnOHZi62h5gXzgDMoEEHADkAAUoAbULSgB6YAxRQ+OAMYpGIgNwAoXUlRwAYnXaMAPObgoAHjHok2VGvSZwAvHHauuAPk9dOGCAHzgAbQBRSRROGAAaOAA6FPNwkUhGYBhgCDocSREAXUS0kXzaEmKioNCI6Ni3RJSksszs3PzCkoxuGuC4MKiYuKK9A2Q0YwI6NRy8gFkIEmA+YGULJZW15Ws7BzonEzMqRgCPWoAKS5acU4AuOC3V9ahzAEpPAIA3CGASQZwa6fDw-P4kd7vcZqPKMeCMAAWEAIkhIAGU1IJJJJHtNZvM6M8dlALOEABIAFQWABkGnFEpgAO4I+iPABGEAgMRwdDgfXOQJQkgZcGZrLgiORqIxWMkfPej1+-y+GFqqyBkpR6MxXMkn3QtWCQqSjB12IpEDAnjgAAY9MFuLpHboYXQ4XAYVzcTM5p0ia9SZSaQAFO44ShQPAIuluIoCy5CkEBA1GyQms2SC1WrwARht+b0zt0JBQakkdzQfB9BIQOAA1pp-cpLiBli9lN78X628TA1TqaHIxGozGGHGFXAlSQ9DB642e69Lp69TO54wm1BLprpRmoUA)

text/0757-default-modifier-manager.md
```diff
@@ -0,0 +1,276 @@
...
+> =
+  | ((...args: ModifierArgs) => void) // setup, with no teardown
+  | ((...args: ModifierArgs) => () => void) // setup with teardown
+
+class A {
+  myModifierA: FunctionModifier = (element, a, b, c, options) => {
```
chriskrycho, line 209: The reason the aren't actually helpful: there's no value on the TS side to providing a form where you can simply say "any old modifier" but *not* actually specify what its elements are. You end up with basically useless types within the body if you do that.

In general, we (meaning on the Typed Ember/Glint side) also won't recommend people import and use `FunctionModifier`. We will just recommend that they define a function which has the right form, and Glint will let them know if it doesn't pass! But for the sake of supporting this where folks *do* want it, the version I showed above gets the job done.

text/0757-default-modifier-manager.md
```diff
@@ -0,0 +1,276 @@
...
+
+On the [Template Lifecycle, DOM, and Modifiers](tldm) page,
+we'll want to insert a section on custom modifiers towards the end, and maybe swap out the
+existing documentation referencing `ember-modifier` and/or `@ember/render-modifiers`.
+We'll also want examples of how to manage modifiers with tracked data, maybe when to use a
+local/private modifier, rather than something you'd put in `app/modifiers`.
+Maybe a whole page is added to the guides on modifiers, their philosophy, when to use them, etc.
+
+While the overall API for modifiers in general has been "not perfect", there has been a fair amount of
+experimentation with modifiers since the 3.11 - 3.16 (pre-octane) era. Function modifiers are
```
chriskrycho, line 243: I'd also note here that there has also been extensive usage of `ember-modifier` *since* the start of the Octane era, which has proven that the function-based modifier design is solid for its use cases, and that all the open questions are specifically around class-based modifiers with their lifecycle management.

text/0757-default-modifier-manager.md
```diff
@@ -0,0 +1,276 @@
...
+[tldm]: https://guides.emberjs.com/release/components/template-lifecycle-dom-and-modifiers/
+[helpers-polyfill]: https://github.com/NullVoxPopuli/ember-functions-as-helper-polyfill
+
+## Drawbacks
+
+
+There could be some awkwardness around the last argument
+passed to the function, as the type signature may not match how the call-site expects
+the signature to be. Projects like [Glint](https://github.com/typed-ember/glint#readme)
+would be essential for helping with clarity here.
```
chriskrycho, line 267: I'm not clear what you mean by this! Can you explain a bit more?

chriskrycho (changes requested):

text/0757-default-modifier-manager.md
NullVoxPopuli: ooo yes. and with autotracking in modifiers, element-only _is_ common!

text/0757-default-modifier-manager.md
NullVoxPopuli: I like that plan!

text/0757-default-modifier-manager.md
NullVoxPopuli: yea!

text/0757-default-modifier-manager.md
NullVoxPopuli: added!

text/0757-default-modifier-manager.md
NullVoxPopuli: Added!

text/0757-default-modifier-manager.md
NullVoxPopuli: Done!, thanks!

text/0757-default-modifier-manager.md
NullVoxPopuli: Updated!, thanks!

text/0757-default-modifier-manager.md
NullVoxPopuli: added

text/0757-default-modifier-manager.md
NullVoxPopuli: added!

NullVoxPopuli:
Made an addon to try it out: https://github.com/NullVoxPopuli/ember-functions-as-modifiers-polyfill

chriskrycho:
Hey @NullVoxPopuli, we chatted about this briefly at today’s Framework Core team meeting, and we all agree that the basic idea is good, but folks had some thoughts about what the shape of the API should be, as well as how it fits in timing-wise with #779, which substantially ups the importance and perceived urgency of this design. We’ll be in touch to work through some of those details with you in a “spec” working group context to hammer down those details.

Thanks again for getting the ball rolling here!

wagenet:
> We’ll be in touch...

Narrator: They were not in touch.

---

In all seriousness though, what needs to be done here? @chriskrycho

chriskrycho:
I believe we actually have been: @wycats has chatted with @NullVoxPopuli about a bunch of related pieces here (and for Resources), based in part on the Starbeam research, and I believe they’ve made some progress there. We should revisit current status though!

I believe the main consideration around function-based modifiers and a default manager is the exact shape of the teardown handling. The current design just returns a teardown function. I think the idea from Starbeam is to return a slightly richer object which includes that.

wagenet:
@chriskrycho to be fair, I did think it was likely that there was some conversation, it’s just not visible here.

wycats:
@chriskrycho said:

> I believe the main consideration around function-based modifiers and a default manager is the exact shape of the teardown handling. The current design just returns a teardown function. I think the idea from Starbeam is to return a slightly richer object which includes that.

This is about right. The longer version is that Starbeam's design supports multiple setup blocks (each of which can run either in "layout" / pre-paint timing or in normal / idle timing), and each of which can have its own teardown.

Most but not all of this can theoretically be built on top of modifier manager. We don't need to add those features now, but it's nice for the "just a function" API design to leave space for future extensions along these lines.

Another consideration from Starbeam: Starbeam modifiers can be invoked with a "ref" that gets filled in dynamically with an element, and once the ref is filled in, the modifier's lifecycle begins. 

Ember doesn't strictly need that design, because our lifecycle is largely managed through the implementation of the template runtime. However, if we want to support invoking modifiers directly, we might want to consider having them take a reference rather than an element. It's not strictly necessary, because you can always call the modifier in the same place as you fill in the element reference.

That said, for Starbeam, the extra indirection allows the reference to be filled in by the internals of third-party code (like another framework), and this may be useful for Ember as well.

I don't think any of this is a blocker to making some tactical progress on this issue, but I suppose @NullVoxPopuli and I should work together to update this proposal to match the Starbeam design and see what people think 😄

MelSumner:
@wycats @NullVoxPopuli nudge to work on this RFC and update it. <3

NullVoxPopuli:
Gonna copy some disco from Discord so it doesn't get lost:

-----

last we schemed, we were thinking about ditching ember-modifier in favor of resources -- something similar to: https://reactive.nullvoxpopuli.com/functions/resource_modifier.modifier.html

tho, in general, we need native support for resource builders so the modifier import wouldn't be needed

there are some questions about timing as well, related to 
- https://github.com/emberjs/rfcs/pull/883
- and https://github.com/emberjs/rfcs/pull/957

cc @runspired @wycats @ef4

one main thing we need for any of this to be ergonomic is to have a chained invocation (at some timing) in the rendering layer.
like:
```gjs
// modifier
function myModifier(element, ...args) {
  // abstraction that handles cleanup here 
}

// resource
function ConnectionStatus() {
  // abstraction that handles cleanup here
}

// these are the same, authoring wise, except for their signature and usage in syntax:
<template>
  <div {{myModifier}}>{{Clock}}</div>
</template>
```
where:
```
  // abstraction that handles cleanup here

```

could be:
```js
function myModifier(element, ...args) {
  return resource(({ on }) => {
    on.cleanup(() => {});
  });
}

function ConnectionStatus() {
  return resource(({ on }) => {
    on.cleanup(() => {});
    
    return () => someValue;
  });
}
```

or 
```js
function myModifier(element, ...args) {
  // setup here 
  
  return {
    /* custom properties for reactivity or something */,
    /* handle for the framework to destroy, like on.cleanup */
    [Symbol.dispose]() {
      // cleanup here
    }
  }
}

function ConnectionStatus() {
  // setup here 
  
  return {
    // special getter? or do we require properties?
    // (eliminating the possibility that resources today 
    //  have where they can represent plain values
    //  (I would not like losing that 😅 )
    get value() {
      return someValue;
    },
    /* custom properties for reactivity or something */,
    /* handle for the framework to destroy, like on.cleanup */
    [Symbol.dispose]() {
      // cleanup here
    }
  }
}
```
using `Symbol.dispose` could one day be the _implementation_ of resources -- but we still have to explore timing semantics, as `on` gives us more API/timing surfaces to explore (as well as a synchronization phase, which can't currently be implemented in ember)
<details><summary>an idea if/when dispose lands</summary>
  to get plain values back, maybe we could do some... prototype hacks.
  
```js
function ConnectionStatus() {
  // setup here 
  
  const result = someValue;
  
  result[Symbol.dispose] = () => { /* cleanup */ };
  
  return result;
}
```

idk, probably terrible.
  
</details>



the chained invocation is needed because we need to call the factory function, and then "do something" with the thing it returns.

this would either be using the HelperManager or ModifierManager syntems based on usage in the template (edited)
in any case, it'd be very hard to get away from any imports for modifiers unless you didn't need cleanup at all

tho, an importless modifier could be one that uses native disposables, like above -- but we don't even have `Symbol.dispose` yet in browsers, so we can't really do that yet.

NullVoxPopuli:
2026 update -- I think this'll be useful to re-startup once Disposables ship -- we probably wouldn't need to use "resources" as a user-facing utility for creating the modifier. can just be a function with dispose as shown above. 

- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/dispose
- https://caniuse.com/mdn-javascript_builtins_symbol_dispose

We are currently missing support from Safari

PR #758, opened 2021-07-21
Default Component Manager

NullVoxPopuli:
[Rendered](https://github.com/NullVoxPopuli/rfcs/blob/default-component-manager/text/0758-default-component-manager.md)

### DIFF
--- a/text/0758-default-component-manager.md
+++ b/text/0758-default-component-manager.md
@@ -1,0 +1,134 @@
+---
+Stage: Accepted
+Start Date: 2021-05-17
+Release Date: Unreleased
+Release Versions:
+  ember-source: vX.Y.Z
+  ember-data: vX.Y.Z
+Relevant Team(s): Ember.js, Learning
+RFC PR: https://github.com/emberjs/rfcs/pull/758
+---
+
+# Default Component Manager
+
+## Summary
+
+Anything that can be in a template has its lifecycle managed by the manager pattern.
+Today, we have known managers for `@glimmer/component`, `@ember/helper`, etc.
+But what happens when the VM encounters an object for which there is no manager?,
+such as a plain function? This RFC explores and proposes a default behavior for
+those unknown scenarios when it comes to _components_.
+
+## Motivation
+
+The addon, [ember-could-get-used-to-this](https://github.com/pzuraq/ember-could-get-used-to-this)
+demonstrated that it's possible to use plain functions for helpers and modifiers.
+Components on the other hand, have had no such "plain" / "vanilla" support / adoption from addons.
+Maybe it's worth exploring what default behavior components should have.
+
+This could reduce the number of dependencies folks need to worry about when building components.
+Using native classes eliminates
+
+## Detailed design
+
+Each implemented meta-manager needs a default manager specified. The logic for choosing when to use
+a default manager is, at a high-level:
+
+```
+if (noExistingManagerFor(X)) {
+  return defaultManager;
+}
+```
+Where X is, for the purposes of this RFC, a Component.
+
+_A Default Manager is not something that can be chosen by the user, but is baked in to the framework
+as a default so that a user doesn't have to build something to use a non-framework-specific variant
+of the three constructs: Helpers, Modifiers, and Components._
+
+
+The default component manager could be similar to the component manager for `@glimmer/component`,
+but be compatible with vanilla classes, such as:
+```js
+class MyComponent {
+  @tracked count = 0;
+}
+```
+Since the introduction of `@ember/destroyable`, we no longer _need_ a specific class for helping
+with destruction.
+
+If a component wanted to have a destruction hook, it could register one itself during construction:
+```js
+import { registerDestructor } from '@ember/destroyable';
+
+export default MyComponent {
+  constructor(...args) {
+    super(...args);
+
+    registerDestructor(this, () => this.myDestroy)
+  }
+
+  myDestroy() { /* ... */ }
+}
+```
+
+and the component manager to support this could look like:
+
+```js
+import { setComponentManager } from '@ember/component';
+import { destroy } from '@ember/destroyable';
+
+class DefaultComponentManager {
+  constructor(owner) { this.owner = owner; }
+
+  createComponent(ComponentClass, args) {
+    return new ComponentClass(this.owner, args);
+  }
+
+  getContext(component) {
+    return component;
+  }
+
+  destroyComponent(component) {
+    destroy(component);
+  }
+}
+
+// side-effect -- this file needs to be imported for the component manager to be installed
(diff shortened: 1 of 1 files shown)


lifeart:
wondering, could we support functions as default components? https://github.com/rwjblue/ember-functional-component

NullVoxPopuli:
My personal opinion is that functions imply that we'd need too many new concepts to be a default. People would expect to be able to have tracked state, an instance / ref of some sort, a way to manage lifececle (hooks, effects, etc)

text/0758-default-component-manager.md
```diff
@@ -0,0 +1,138 @@
...
+those unknown scenarios when it comes to _components_.
+
+## Motivation
+
+The addon, [ember-could-get-used-to-this](https://github.com/pzuraq/ember-could-get-used-to-this)
+demonstrated that it's possible to use plain functions for helpers and modifiers.
+Components on the other hand, have had no such "plain" / "vanilla" support / adoption from addons.
+Maybe it's worth exploring what default behavior components should have.
+
+This could reduce the number of dependencies folks need to worry about when building components.
```
chriskrycho, line 29: If this is aimed at reducing the number of imports, then it has an interesting side effect: it makes the “base case” *more* for people to type. It's extremely rare in my experience to want a backing class which *doesn't* care at all about `args`. Given that, the proposed delta for your average component would be from today’s default—

```js

  // whatever local and derived state
}
```

—to this:

```js

export default class MyComponent {
  constructor(owner, args) {
    this.owner = owner;
    this.args = args;
  }

  // whatever local and derived state
}
```

This is a substantial *increase* in lines of code, and the result is a substantial *decrease* in the basic safety: you can accidentally end up just stomping `this.args` somewhere by writing `this.args = 'lol';`: something Glimmer component protects you from.

text/0758-default-component-manager.md
```diff
@@ -0,0 +1,138 @@
...
+Using native classes eliminates the reliance on a "framework thing" and further
+lessens the appearance of "magic through object-oriented inheritance".
```
chriskrycho, line 31: I *very strongly* disagree here. This is actually much *more* "magical": in this design, the framework calls a class constructor with specific arguments because it happens to be invoked in a specific way. It's *not* equivalent to function-based helpers or modifiers in my view:

- helpers are just functions which *you* pass all the arguments to
- modifiers are just functions which you pass all the arguments to as well, modulo the `element`—but notionally, sticking them on/applying them to the element *is* passing it that

To be sure, I can see a way to rationalize this in terms of saying that a component template is simply a specific kind of callback to the rendering engine (in a real sense that *is* what it is, of course), but it’s very strange notionally, and seems to me to move *away* from the actual value of explicitness that the Glimmer component subclass gives us.

I also don't think that there's anything particularly object-oriented about this use of inheritance, nor that it's in the realm of anti patterns that we *do* see from inheritance-based designs. There is effectively no *behavior* we're worried about here: just some boilerplate elimination via setting up the `owner` and `args`—and, importantly, doing it *safely* so that they don't get accidentally stomped later.

Insofar as one potential goal might be getting rid of the `willDestroy` hook, we can do that separately: it's perfectly reasonable to consider deprecating that from Glimmer `Component`, and it's also basically trivial to codemod to.

text/0758-default-component-manager.md
```diff
@@ -0,0 +1,138 @@
...
+
+Additionally, a goal of baseclass-less-classes as components is to have 0 potential API
+conflicts with userspace.
```
chriskrycho, line 34: In general, it's rare to want to name something as abstract as `Component`, but in the few cases where it *does* make sense, you can simply take advantage of the fact that ES Modules allow renaming imports (*especially* easy for default exports):

```js
import GlimmerComponent from '@glimmer/component';

export default class Component extends Component {
  // ...
}
```

Accordingly, I don't think this is actually a valuable goal. Names are actually useful and good things to provide, and they provide us great hooks for teaching, for docs, etc.!

chriskrycho (commented):
I spent some time thinking about this today (in relation to whether we would want to consider doing it as part of #779), and I ended up with a lot more questions than answers here.

Beyond the points addressed inline, it's simply not clear to me what value this actually adds for end users. Consider: for the case when you just want some lightweight local tracked state, this might seem appealing:

```js

export default class MyComponent {
  @tracked isNeat = false;

  toggle = () => {
    this.isNeat = !this.isNeat;
  };

  <template>
    <p>{{if this.isShowing "neat!" "not neat"}}</p>
    <button type='button' {{on "click" this.toggle}}>
      toggle neatness
    </button>
  </template>
}
```

This is fine, so far as it goes (though again: see my questions and comment sinline), but in *many* such cases, you're just as well off using something like my [ember-simple-track-helper](https://www.npmjs.com/package/ember-simple-track-helper):

```js
import { track } from 'ember-simple-track-helper';

<template>
  {{#let (track false) as |neatness|}}
    <p>{{if neatness.value "neat!" "not neat"}}</p>
    <button type='button' {{on "click" (fn neatness.update (not neatness.value))}}>
      toggle neatness
    </button>
  {{/let}}
</template>
```

…or the equivalent in something local and dedicated, or, again, just subclassing Glimmer `Component`.

text/0758-default-component-manager.md
NullVoxPopuli: that's a good point :thinking:

NullVoxPopuli:
@chriskrycho many many many thanks for your very thorough and detailed feedback. I think, after noodlin' on this for the past few months, and after reading the more recent RFCs, and especially your `<template>` RFC, and with where we all want to see ember go in the near future, I'm in agreement with your points.

I'm going to close this RFC, and it was 100% the weakest of my default manager RFCs. lol
_but_, I did want the option brought up and discussed -- and I can't believe I totally forgot about assigning owner/args on the class if a plain class is used. _oof_.

Also, this is :fire: 
```jsx
import { track } from 'ember-simple-track-helper';

<template>
  {{#let (track false) as |neatness|}}
    <p>{{if neatness.value "neat!" "not neat"}}</p>
    <button type='button' {{on "click" (fn neatness.update (not neatness.value))}}>
      toggle neatness
    </button>
  {{/let}}
</template>
```

chriskrycho:
Glad it was useful input! And I’m super excited for your default modifier manager RFC as well as grateful for your working already landing (and polyfilling!) the default helper manager RFC.

One other point for posterity if/as this comes up in the future: there is also zero even *potential* win here for TS users, because you need some way to hook into the broader system so tools like Glint can understand them. @dfreeman and I spent some time looking at this today, and the net of our investigation is that you end up *needing* at minimum an interface to implement which requires `args`… so you have just as many imports and *more* lines of code in the implementation (even with TS’s shorthand for constructors).

Issue #434, opened 2019-01-19
Component helper in JS


mehulkar:
It's possible to use the `{{component}}` helper in templates to specify a dynamic component name. This allows all kinds of polymorphic approaches to rendering data, which is awesome. However, as of now, it is not easily possible to bind a different set of arguments to dynamic components, which means that your components need to have the exact same API.

It's debatable whether or not this is a good idea, but if there was an equivalent to the component helper in JS, we would be able to construct a component with a different set of arguments.

The goal is to be able to do something like this:

```hbs
{{#each this.items as |item|}}
  {{component item.component}}
{{/each}}
```

```js
// app/controllers/index.js
export default Controller.extend({
  items: computed('model', () => {
    return this.model.map(item => {
      let component;
      if (item.type === 'foo') {
        component = createComponent('special-component', { arg1: item.name, arg2: item.type });
      } else {
        component = createComponent('default-component', { otherArg: item.name });
      }
      return Object.assign(item, { component });
    });
  });
});
```

This lets us use separate components based on `item.type`, but allows for flexibility in their API.

sdhull:
I have also wanted the ability to do `fooComponent = component('foo-display', {foo, bar});` in .js files. Usually I've figured out tricksy ways to work around the fact that this is impossible (typically by using the `{{component}}` helper in a parent .hbs context and passing it down) but I'm sure there are places where it would be clearer to simply instantiate the component renderer where it's needed.

I'd like to add that it would be awesome if instances of `CurriedComponentDefinition` were callable like a normal function (so that it might be renderable by 3rd party libraries). This is probably impossible but a guy can dream.

sdhull:
Maybe
```js
import { componentRenderer } from '@ember/component';
```
I'm not really sure what where it would be appropriate to export such a thing. I also think it needs some sort of rendering context to work...?

chancancode:
As I mentioned in the other thread, the difficulty here is passing _bound_ arguments. Something like `component(“foo-display”, { foo, bar })` is only going to work if you expect foo and bar to be const/unbound, which isn’t always going to be appropriate.

Herriau:
@chancancode I wouldn't expect `foo` or `bar` to be bound. Talking from experience of the various use cases where we have leveraged our homemade `createCurriedComponentDefinition()` (see emberjs/ember.js#17509), we have rarely felt the need for any of the curried properties to be bound. 

[...]

PR #762, opened 2021-08-11
invokeHelper argument-based thunks

NullVoxPopuli:


text/0762-invokeHelper-property-thunks.md
```diff
@@ -0,0 +1,176 @@
...
+  });
+
+  get calculatorValue() {
+    return getValue(this.calculator);
+  }
+}
+
+// Assume there is a Helper manager registered that knows what to do with this
+// and getValue(this.calculator) returns an instance of this class
+// (instead of calling compute like on the default Helper class)
```
NullVoxPopuli, line 62: I have a reproduction of this in a stackblitz, but, you'd need to do some work to get it running :cry: 
https://github.com/stackblitz/webcontainer-core/issues/294

NullVoxPopuli:
Add a "How we teach this". Rearrange the Motivation / Summary a bit.

ef4:
I'm not fully convinced this solves the problem better than what can already be done with existing primitives.

It's the author of the helper who knows whether the entangling of argument tracking matters or not. But it's the caller of the helper who would be obligated to pick the right invocation pattern. So this doesn't make it possible for helper authors to solve their own problem reliably.

Instead, a helper author who cares about fine-grained autotracking of several different values could design their API to make that mandatory by making thunks or tracked properties explicitly part of what you're expected to pass in. 

For example, nothing stops the author from accepting an argument that it itself a thunk:

```js
invokeHelper(this, SomeHelper, () => ({
  firstArg: () => theTrueValue,
}))
```

Or accepting an argument with tracked properties:

```js
class HelperArgs {
  @tracked someArgument;
  @tracked someOtherArgument;
}

// ---

let args = new HelperArgs();
invokeHelper(this, TheHelper, () => args);
args.someArgument = 1; // fine-grained invalidation visible to the helper.
```

NullVoxPopuli:
How would something like this be possible if we don't have argument-based thunks? https://github.com/emberjs/rfcs/issues/434#issuecomment-878324807

```js
// wrapper around curry because glimmer components only take named args
function component(Klass, thunk) {
  return curry(Klass, () => {
    let named = thunk(); // all args are consumed, regardless of usage
    return { named };
  });
}

let curried = component(MyComponent, () => ({ fruit: this.pepper }))
```

NullVoxPopuli:
unless... we make a public API for the Lazy Args Proxy class that I've seen duplicated in a couple places :thinking:

Like, 
```js
let args = argsProxyFor({
  positional: [
    () => this.a,
    () => this.b,
  ],
  named: {
    c: () => this.c,
    d: () => this.d,
  }
});

args.named.c // consumes and evaluates the this.c
args.positional[0] // consumes and evaluates this.a

// ...

// example of something I want for components / modifiers / helpers / etc.
// (not a goal of this RFC, but something I do want to work towards)
function component(Klass, namedArgumentThunks) {
  return curry(Klass, argsProxyFor({ named: namedArgumentThunks }));
}

// usage
curriedComponent = component(MyComponent, {
  c: () => this.c,
  d: () => this.d,
};
``` 
This is still verbose, but would allow curried components to have lazily consumed args

ef4:
What I was saying above is that the author of a helper can design their arguments to be independently lazy so their consumption can be tracked independently. 

That is different from being able to invoke an arbitrary *existing* helper in a way that provides the laziness.

And I think that's OK because it's the author of the helper who even knows whether some arguments will remain unconsumed in some important cases. If they know it's important, they can design the arguments to accommodate it.

> I still think we need a way to build out these lazy-evaluating argument objects.

I'm pretty sure it doesn't require anything from the framework, just native `Proxy`.

Finally, an alternative to currying-in-JS would be to implement argument splatting plus GJS or equivalent, which would give a very clear one-liner way to do currying:

```js
let curriedComponent = <template><MyComponent @title=curriedValue ...attributes ...arguments /></template>
```

And this case avoids all the complexities of visible thunks, which is in one sense the whole part of our declarative template language.

NullVoxPopuli:
thanks!
I'm going to explore both Proxy and getter -using thunks for resources.

>  argument splatting plus GJS

this is very exciting, and if we added block splatting, that would solve a [ton of design problems I'm currently having](https://github.com/emberjs/rfcs/pull/460#issuecomment-902961346) creating components.

```jsx
let curriedComponent = <template>
  <MyComponent @title=curriedValue ...attributes ...arguments ...blocks />
</template>
```

rwjblue:
After some discussion with the core team today, I am moving this RFC into "FCP to Close".

Based on some conversations that @NullVoxPopuli and I had in discord, using getters for the named properties should allow the laziness desired here without changes to the system (and "its just JavaScript" 🎉 ).

```js
export default class PlusOne extends Component {
  plusOne = invokeHelper(this, RemoteData, () => {
    return {
      positional: [],
      named: {
        get foo() { return this.args.qux; }
      }
    }
  });
}
```

NullVoxPopuli:
This "works",
```js
      plusOne = invokeHelper(this, RemoteData, () => {
        return {
          positional: [],
          named: {
            foo: () => this.startingNumber,
          },
        };
      });
```
but it means that all accesses to `foo` must now be `foo()`

rwjblue:
Ya, sorry about that. I think the syntax would be:

```js
export default class PlusOne extends Component {
  plusOne = invokeHelper(this, RemoteData, () => {
    return {
      positional: [],
      named: {
        get foo: () => this.args.qux,
      }
    }
  });
}
```

PR #763, opened 2021-08-13
Asset Import Spec

ef4:
[Rendered](https://github.com/emberjs/rfcs/blob/asset-importing-spec/text/0763-asset-importing-spec.md)

### DIFF
--- a/text/0763-asset-importing-spec.md
+++ b/text/0763-asset-importing-spec.md
@@ -1,0 +1,200 @@
+---
+Stage: Accepted
+Start Date: 2021-08-13
+Release Date: Unreleased
+Release Versions:
+  ember-source: vX.Y.Z
+  ember-data: vX.Y.Z
+Relevant Team(s): Ember.js, Ember CLI
+RFC PR: https://github.com/emberjs/rfcs/pull/763
+---
+
+<!---
+Directions for above:
+
+Stage: Leave as is
+Start Date: Fill in with today's date, YYYY-MM-DD
+Release Date: Leave as is
+Release Versions: Leave as is
+Relevant Team(s): Fill this in with the [team(s)](README.md#relevant-teams) to which this RFC applies
+RFC PR: Fill this in with the URL for the Proposal RFC PR
+-->
+
+# Asset Import Spec
+
+## Summary
+
+This RFC defines standard semantics for what it means to depend on files that are not Javascript or CSS, like images, fonts, and other media. It proposes that when your code needs to refer to an asset, you should import it, which will give you back a valid URL:
+
+```js
+import myImage from './hello.png';
+class extends Component {
+  myImage = myImage
+}
+```
+```hbs
+<img src={{this.myImage}} />
+```
+
+## Motivation
+
+Apps and addons often need to refer to images, fonts, etc from within their JS, HBS,CSS, or HTML files. These references need to remain correct even as both the referring file and the referred-to file go through a build that might include fingerprinting and optimization.
+
+Today, the ecosystem mostly relies on [broccoli-asset-rev](https://github.com/ember-cli/broccoli-asset-rev). But broccoli-asset-rev has some major architectural problems and it doesn't take advantage of the newer capabilities we have in ember-auto-import and embroider.
+
+The biggest problem with broccoli-asset-rev is that it cannot achieve full correctness because it tries to detect inter-file references within arbitrary source code that is not statically analyzable. It's easy to accidentally refactor working code into code that works in development and test and then fails in production! For example, assuming broccoli-asset-rev is configured to fingerprint PNG assets, this refactor will look correct until you get into production and then have broken images:
+
+```diff
+class extends Component {
+-  iconURL = "/icons/happy.png"
++  get iconURL() {
++    return `/icons/${this.args.emotion}.png`
++  }
+}
+```
+
+The regular expressions that broccoli-asset-rev uses to do reference rewriting are effectively un-editable at this point. Any possible change you'd make to them breaks somebody's app.
+
+Another problem with broccoli-asset-rev is that it's "push based", meaning it handles files because they exist, rather than handling files because somebody is actually consuming them. It can't detect unused assets and it can't detect missing assets at build time. It exposes the underlying implementation of how assets get delivered, making it hard to change later without breaking app code.
+
+And because broccoli-asset-rev exposes assumptions about the exact layout of the final build output, it's not compatible with Embroider. Embroider allows different Javascript packagers to optimize the app delivery in any spec-compliant way. For example, small images might get inlined as `data:` URIs, CSS `@import()` might get inlined or not, etc. All these concerns really need to be integrated into the Javascript packager, because asset handling affects app-wide hashing & fingerprinting.
+
+In contrast, a pull-based design that lets code declare what assets it needs and then not worry about how those assets will get delivered is safer and easier to change in the future.
+
+## Detailed design
+
+Inter-file references take many forms. Here is a guide to inter-file references organized by which kind of file is doing the referring ("From Type") and which type of file it's referring to ("To Type"). Only the row containing **Proposed** is new in this RFC:
+
+From Type | To Type | Example Syntax | Semantics
+--------------------|-----------------------|--------|---
+JS | JS | `import { stuff } from 'thing';`| ECMA with NodeJS resolving rules
+JS | CSS | `import 'thing/style.css';` | [Embroider Spec CSS Rule](https://github.com/emberjs/rfcs/blob/master/text/0507-embroider-v2-package-format.md#css) (1)
+JS| Other Asset (2) | `import url from 'thing/icon.png';`| **Proposed** URL Rule (3)
+CSS | CSS | `@import("thing/style.css")` | [W3C](https://drafts.csswg.org/css-cascade/#at-import) plus NodeJS resolving rules
+CSS | Other | `url("thing/icon.png");` | W3C plus NodeJS resolving rules
+HTML | JS | `<script src="./thing.js"></script>` | W3C
+HTML | CSS | `<link rel="stylesheet" href="./thing.css"/>` | W3C
+HTML | Other | `<img src="./thing.png" />` | W3C
+
+
+1. Importing a file with an explicit `.css` extension guarantees that the given CSS will be loaded into the DOM before your module executes. We do not define any exported values, this is purely for side-effect. This rule is not part of the present RFC, it was already in RFC 507.
+2. "Other Asset" means this rule applies to any JS import of a path that ends in an explicit file extension that is not `.js` or `.css`, meaning:
+    ```js
+    function isOther(theImportedPath) {
+      let extension = /\.([^.\/]+)$/.exec(theImportedPath)?.[1]?.toLowerCase();
+      return extension && !['js', 'css'].includes(extension);
+    }
+    ```
+
+3. Importing any "Other Asset" file gives you back a `default` export containing a valid URL to that file.
+
+### What about references from HBS?
+
+This RFC doesn't treat HBS separately because as far as the module system is concerned, HBS files are just JS files. They're transpiled into JS before any module traversal can happen. This RFC doesn't define any static syntax for use in templates because we should use whatever syntax is designed to go with [Strict Mode](https://github.com/emberjs/rfcs/blob/master/text/0496-handlebars-strict-mode.md). For example, in GJS you could say:
+
+```js
+import myImageURL from './my-image.png';
+<template>
+  <img src={{myImageURL}} />
+</template>
+```
+
+Today, the recommendation should be to use the JS part of a component to import assets and pass the resulting URLs to the HBS part of the component.
+
+Using only Ember public API, it would also be possible to write an optional preprocessor that allows you to experiment with any authoring format you like for accessing assets from HBS. For example, you could rewrite this:
+
+```hbs
+{{!-- for illustration purposes only, this
+      hypothetical "import-asset" helper is
+      not being proposed by this RFC, it's
+      just something you could layer on top. --}}
+<img src={{import-asset "./thing.png"}} />
+```
+
+To this:
+```js
+// this line is the only thing in this file that is newly proposed by this RFC
+import url0 from './thing.png';
+// all the rest of this is existing low-level API from other merged RFCS:
+import { setComponentTemplate } from '@ember/component';
+import { precompileTemplate } from '@ember/template-compilation';
+import templateOnlyComponent from '@ember/component/template-only';
+export default setComponentTemplate(
+  precompileTemplate('<img src={{url0}} />', { scope: () => ({ url0 }) }),
+  templateOnlyComponent()
+);
+```
+
+### Extensibility
+
+There are many alternative competing ways to assign extra semantics to imports of non-JS files. Examples include:
+
+ - importing `.graphql` files, which get preprocessed into a Javascript value, possibly with the side effect of tracking that this particular query is in use.
+ - doing "CSS in JS" where an imported CSS file comes back as a Javascript value.
+
+This RFC is not in conflict with these patterns, because they should be understood as authoring formats that can be compiled into a spec-compliant format for compatibility with the rest of the build system and ecosystem. In both the above cases, a preprocessor would rewrite these files to Javascript, and then they get completely standard handling from that point forward.
+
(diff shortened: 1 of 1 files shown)


NullVoxPopuli:
[Rendered](https://github.com/emberjs/rfcs/blob/asset-importing-spec/text/0763-asset-importing-spec.md) (post file rename)

text/0763-asset-importing-spec.md
```diff
@@ -0,0 +1,200 @@
...
+HTML | CSS | `<link rel="stylesheet" href="./thing.css"/>` | W3C
+HTML | Other | `<img src="./thing.png" />` | W3C
+
+
+1. Importing a file with an explicit `.css` extension guarantees that the given CSS will be loaded into the DOM before your module executes. We do not define any exported values, this is purely for side-effect. This rule is not part of the present RFC, it was already in RFC 507.
+2. "Other Asset" means this rule applies to any JS import of a path that ends in an explicit file extension that is not `.js` or `.css`, meaning:
+    ```js
+    function isOther(theImportedPath) {
+      let extension = /\.([^.\/]+)$/.exec(theImportedPath)?.[1]?.toLowerCase();
+      return extension && !['js', 'css'].includes(extension);
```
NullVoxPopuli, line 85: any reason to exclude ts from here?

NullVoxPopuli: I suppose this same caveat/note [here](https://github.com/emberjs/rfcs/pull/763/files#diff-38fadf17712b2606dfa948983f8d8d89d28c66e71b346d7c7b5bccc5808bf446R93) applies to ts

chancancode:
How would importing assets _from_ CSS work? eg images and fonts?

mydea:
Do I understand correctly, that if I'd want dynamic filenames I'd have to continue using `/public`? E.g. something like this:

```js
export default class FlagComponent  extends Component {
  get url() {
    return `/assets/flags/${this.args.country}.png`;
  }
}
```

I know that also does not "just work" with the current broccoli-asset-rev fingerprinting, but might be interesting to mention how dynamic asset names would/should work under the new format (which I am generally def. in favor of!). Depending on the use case, I guess either an exhaustive map (if you don't have too many options) would work fine, but for larger lists of options (e.g. country flags) it might be a bit "overkill":

```js
import flagDe from 'my-app/flags/de.png';
import flagAt from 'my-app/flags/at.png';
import flagUk from 'my-app/flags/uk.png';

const flagMap = {
  de: flagDe,
  at: flagAt,
  uk: flagUk
};

export default class FlagComponent extends Component {
  get url() {
    return flagMap[this.args.country];
  }
}
```

text/0763-asset-importing-spec.md
```diff
@@ -0,0 +1,200 @@
...
+In contrast, a pull-based design that lets code declare what assets it needs and then not worry about how those assets will get delivered is safer and easier to change in the future.
+
+## Detailed design
+
+Inter-file references take many forms. Here is a guide to inter-file references organized by which kind of file is doing the referring ("From Type") and which type of file it's referring to ("To Type"). Only the row containing **Proposed** is new in this RFC:
+
+From Type | To Type | Example Syntax | Semantics
+--------------------|-----------------------|--------|---
+JS | JS | `import { stuff } from 'thing';`| ECMA with NodeJS resolving rules
+JS | CSS | `import 'thing/style.css';` | [Embroider Spec CSS Rule](https://github.com/emberjs/rfcs/blob/master/text/0507-embroider-v2-package-format.md#css) (1)
```
jelhan, line 71: How would that play together with [native CSS modules](https://www.chromestatus.com/feature/5948572598009856) recently shipped as an experimental feature in Chrome?

I know there is much discussion around that spec. I don't have enough insights myself into bundling in general, CSS-in-JS etc. But I think we should consider how the decision made here may affect a migration path.

ef4: Thanks for pointing that out, I had not been following that proposal. It relies on [import assertions](https://github.com/tc39/proposal-import-assertions) syntax, which is currently in stage3 at tc39.

Because that proposal relies on new syntax, it wouldn't conflict with anything in this proposal as written. We could adopt that proposal when it stabilizes. But it's worth asking if we shouldn't just jump directly to that proposal. Mostly that comes down to practical questions of how likely it is to stabilize without further changes and whether enough existing tooling supports it.

For example, webpack seems to be able to parse the new syntax, but doesn't let you write rules dependent on the import assertions, as far as I can tell.  TypeScript doesn't have an implementation yet. Unsure about eslint.

text/0763-asset-importing-spec.md
ef4: I would say the "Extensibility" section applies. Compiles-to-javascript languages aren't covered by this proposal because we're only standardizing what happens in the Javascript.

Also, TypeScript in particular forbids an explicit `.ts` extension in its import statements, or at least it used to. They rely on the ambiguity to keep the import valid after everything has converted from ts to js.

text/0763-asset-importing-spec.md
jelhan: > Because that proposal relies on new syntax, it wouldn't conflict with anything in this proposal as written.

Thanks for clarifying.

> But it's worth asking if we shouldn't just jump directly to that proposal.

I'm not sure if we can at this point in time. If I got the proposal right, importing CSS does not have side-effects. Instead the default export is a [constructable style sheet](https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleSheet), which must be applied explicitly to either `document` or a shadow root. A style sheet can be applied using `adoptedStyleSheets` API. But that one is not supported by Firefox and Safari yet [accordingly to caniuse](https://caniuse.com/mdn-api_document_adoptedstylesheets).

ef4:
> How would importing assets _from_ CSS work? eg images and fonts?

I should be more explicit about it in the text, but what I meant by:

From Type | To Type | Example Syntax | Semantics
-- | -- | -- | --
CSS | CSS | @import("thing/style.css") | W3C plus NodeJS resolving rules
CSS | Other | url("thing/icon.png"); | W3C plus NodeJS resolving rules

is that we promise that those syntaxes will obey their original W3C semantics, with the addition of the ability to do node-style package resolution (so that the imported paths have a consistent meaning between here and in JS).

Like with JS, how we actually implement the semantics is left unspecified. `@import` for example is totally allowed to be inlined, or not. `url()` could end up containing a data URI, etc.

Providing the option to fine-tune the implementation is left up to each specific bundler. In a classic build, that would be ember-auto-import and its webpack.

text/0763-asset-importing-spec.md
ef4: Whether browsers implement is not really the limiting factor, because we can transpile / polyfill as needed.

The limiting factor is whether developer tooling works well enough on your code as-authored that you'd want to use these features in it.

text/0763-asset-importing-spec.md
```diff
@@ -0,0 +1,200 @@
...
+
+Stage: Leave as is
+Start Date: Fill in with today's date, YYYY-MM-DD
+Release Date: Leave as is
+Release Versions: Leave as is
+Relevant Team(s): Fill this in with the [team(s)](README.md#relevant-teams) to which this RFC applies
+RFC PR: Fill this in with the URL for the Proposal RFC PR
+-->
+
+# Asset Import Spec
```
stefanpenner, line 23: When it comes to static assets, such as images, it is  common for them to be "internationalized". Do you have thoughts on how this proposal could play with that?

For example, let's imagine i have the following localized versions of `hello.png`:

* `hello.en-us.png`
* `hello.fr-fr.png`

Where the form is something like: `hello.<%locale%>.png` or maybe `<%locale%>/hello.png`

text/0763-asset-importing-spec.md
```diff
@@ -0,0 +1,200 @@
...
+
+And because broccoli-asset-rev exposes assumptions about the exact layout of the final build output, it's not compatible with Embroider. Embroider allows different JavaScript packagers to optimize the app delivery in any spec-compliant way. For example, small images might get inlined as `data:` URIs, CSS `@import()` might get inlined or not, etc. All these concerns really need to be integrated into the JavaScript packager, because asset handling affects app-wide hashing & fingerprinting.
+
+In contrast, a pull-based design that lets code declare what assets it needs and then not worry about how those assets will get delivered is safer and easier to change in the future.
+
+## Detailed design
+
+Inter-file references take many forms. Here is a guide to inter-file references organized by which kind of file is doing the referring ("From Type") and which type of file it's referring to ("To Type"). Only the row containing **Proposed** is new in this RFC:
+
+From Type | To Type | Example Syntax | Semantics
```
stefanpenner, line 68: Today when folks use `.wasm` [one approach is](https://developer.mozilla.org/en-US/docs/WebAssembly/Loading_and_running) is to use `fetch` as follows:

```js
WebAssembly.instantiateStreaming(fetch('simple.wasm'), importObject)
.then(results => {
  // Do something with the results!
});
```

This also runs into the same finger printing + reachability problems described in this RFC.

My concern arises when attempting to apply this RFCs solution to the `wasm` problem. This RFC would recommend:

```js
import simple from './simple.wasm';

WebAssembly.instantiateStreaming(fetch(simpl), importObject)
.then(results => {
  // Do something with the results!
});
```

Which I suspect would work today, BUT it wouldn't surprise me if `import simple from './simple.wasm'` [eventually becomes an official way to load and import `wasm`](https://github.com/WebAssembly/esm-integration/tree/master/proposals/esm-integration), at which the RFC's approach would be incompatible....

stefanpenner: To expand this beyond `wasm`, I am not super comfortable with the idea of creating "meaning" to the default export of a currently unsupported module FileType. `wasm` above is an example that may very likely become supported, and I would be extremely surprised if we don't see further evolution in this area.

To mitigate said risk, we may want to consider adding this capability too `@embroider/macros`. Specifically, in additional to `importSync` `importURL` could be created.

```js
import { importURL } from '@embroider/macros';

WebAssembly.instantiateStreaming(importURL('simple.wasm'), importObject)
.then(results => {
  // Do something with the results!
});
```

or 

```js
import { importURL } from '@embroider/macros';

importURL('thing/icon.png');
importURL('thing/style.css');
importURL('.'); // either illegal & throws at build/lint or implies the current module MUST be addressable independently of a larger bundle
```

stefanpenner: Just to add some increased specificity, ultimately this RFC is about being able to gain access to the `import.meta.url` for a given `ModuleSpecifier` a form of module/bundler/runtime reflection but in a statically analyzable manner.

Given that, a more pedantic name could be:

```js
import { importMeta } from '@embroider/macros'
importMeta(ModuleSpecifier).url
```

text/0763-asset-importing-spec.md
```diff
@@ -0,0 +1,200 @@
...
+-->
+
+# Asset Import Spec
+
+## Summary
+
+This RFC defines standard semantics for what it means to depend on files that are not JavaScript or CSS, like images, fonts, and other media. It proposes that when your code needs to refer to an asset, you should import it, which will give you back a valid URL:
+
+```js
+import myImage from './hello.png';
```
stefanpenner, line 30: Do we need an async + dynamic variant? today we not only have `import` but `importSync` and `async function import`

text/0763-asset-importing-spec.md
stefanpenner: Ok some further snooping it looks like `node.js` has an experimental feature which I think is an even closer fit to what this RFC is exploring:

Specifically: [`import.meta.resolve(specifier[, parent])`](https://nodejs.org/api/esm.html#esm_import_meta_resolve_specifier_parent)

`import.meta.resolve` is async, so for our needs `import.meta.resolveSync` variant would be important.
One could implement `import.meta.resolveSync` via code-mode or more conservatively as an `@embroider/macros` macro:

```js
import { resolveSync } from '@embroider/macros';
```

```hbs
<img src={{resolve-sync 'thing.jpg'}} >
```



----

Although it seems fishy at first to introduce `import.meta.resolveSync` it is actually pretty aligned with the design of `import.meta` as it is designed to be augmented by a host, as one can see with Nodes `import.meta.resolve`.

knownasilya: Lots of good thoughts here that I agree with as well.

ef4: Yeah, this is pretty convincing, I think we can iterate in this direction.

text/0763-asset-importing-spec.md
ef4: In JS this would take advantage of dynamic import with a template string literal argument (which is already allowed in the embroider spec) or the `importSync` macro (which works in classic apps too).

In CSS, I don't think we would support it directly and you'd need to either work in JS or use a preprocessor on the CSS that expands outs the substitutions. I'm pretty sure CSS vars can't be interpolated into URLs, but if they can we could respect that.

stefanpenner: So TL;DR `importSync` would be used to add the right dependency from JS -> CSS?

wagenet:
What's the status on this? Should we get more formal review?

achambers:
Context: @kategengler has some updates to the RFC here: https://github.com/emberjs/rfcs/pull/973

ef4:
I have merged updates from @kategengler and taken this out of draft, please take a look again as we think it's ready to advance.

simonihmig:
On the implementation side, how does this RFC relate to https://github.com/embroider-build/ember-auto-import/issues/565? My understanding would be, that a good part of this RFC's implementation is to get that eai feature to land, the other part is https://github.com/webpack/webpack/issues/16693 and some additional default webpack config to set up the asset loader, is that fair to say? 

On the eai feature, as it is currently envisaged, it would require users to opt-in by specifying a pattern for all the stuff that eai should handle, that would have traditionally been part of the classic build. But it seems according to the RFC importing anything (that's not a module or CSS I guess?) is supposed to "just work", correct?

ef4:
@simonihmig this RFC no longer proposes using import statements for assets. It proposes `import.meta.resolve()`.

The implementation of that feature in classic apps might very well be handled by ember-auto-import.

simonihmig:
@ef4 I understand the recent changes in this RFC. But I still feel there is a bit of overlap between this RFC and the proposed eai feature, that could lead to some inconsistencies...

For example, under this RFC you would be able to `import.meta.resolve('./style.css')` if you wanted to, right? If you don't want just the URL, but the side-effect of _applying_ the CSS, you could do `import './style.css'`, but _only_ for v2 addons and Embroider apps. For v1 addons or classic apps, that would be possible with that eai feature, but as it stands only with some additional configuration.

For a more complete map of which kind of asset handling is possible:

| Expression                            | v1 addon | v2 addon | classic app | Embroider app |
|---------------------------------------|----------|----------|-------------|---------------|
| `import.meta.resolve('./logo.png')`   | 🟢        | 🟢        | 🟢           | 🟢             |
| `import.meta.resolve('./style.css')`  | 🟢        | 🟢        | 🟢           | 🟢             |
| `import logo from './logo.png'` [^1]      | 🟠🟡       | 🟡        | 🟠🟡          | 🟡             |
| `import './style.css'`                | 🟠        | 🟢        | 🟠           | 🟢             |
| `import.meta.resolve('./data.yml')`   | 🟢        | 🟢        | 🟢           | 🟢             |
| `import parsedData from './data.yml'` | 🟠🟡       | 🟡        | 🟠🟡          | 🟡             |
[^1]: this example is superseded by this RFC, but still possible to set up

🟢 works out of the box, according to this RFC / RFC507 (v2 spec)
🟡 needs custom webpack loader configuration
🟠 needs https://github.com/embroider-build/ember-auto-import/issues/565 and custom config

The 🟡  part of needing custom config for importing unexpected file types is expected, and not my point here. But my point is about the 🟠 part of being able to import your own assets, where v1 addons/classic apps need that eai feature with a custom config (which btw is not backed by an RFC), while this "just works" for v2 addons/Embroider apps. 

This RFC makes `import.meta.resolve()` work symmetrically across all types of addons/apps. Shouldn't we strive for similar symmetry for the other cases?

Implementation-wise, I would _think_ this shouldn't be a big problem when we have that eai feature, given that it was never possible to import anything but JavaScript for v1 addons and classic apps, so requiring a config is not really need. We could just enable eai to handle any (local, relative) import without any explicit config, when that import refers to anything that is not JavaScript, couldn't we? Which would make the table above symmetric again.

Does that make sense? And if so, should this be a short follow-up RFC?

balinterdi:
[Rendered](https://github.com/emberjs/rfcs/blob/master/text/0763-asset-importing-spec.md) (for some reason, the URL has changed again).

PR #772, opened 2021-11-10
Deprecate Bower Support

### LINKED ISSUE #610: [Ember CLI] Deprecate bower APIs
More context at https://github.com/ember-cli/ember-cli/issues/9110.

Ember CLI still contains internal APIs to support bower. These should be deprecated for later removal.

bertdeblock:
[Rendered](https://github.com/bertdeblock/rfcs/blob/deprecate-bower-apis/text/0772-deprecate-bower-support.md)

It could very well be that a lot of stuff is missing in this RFC, but hopefully it can get the ball rolling on deprecating Bower support.

Closes #610.

### DIFF
--- a/text/0000-deprecate-bower-apis.md
+++ b/text/0000-deprecate-bower-apis.md
@@ -1,0 +1,251 @@
+---
+Stage: Accepted
+Start Date: 2021-10-21
+Release Date: Unreleased
+Release Versions:
+  ember-source: vX.Y.Z
+  ember-data: vX.Y.Z
+Relevant Team(s): Ember CLI
+RFC PR:
+---
+
+# Deprecate Bower APIs
+
+## Summary
+
+This RFC proposes to deprecate the following Bower APIs:
+
+- [Blueprint::addBowerPackageToProject](https://ember-cli.com/api/classes/blueprint#method_addBowerPackageToProject)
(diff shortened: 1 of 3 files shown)


backspace:
The rendered link is broken for me, maybe it should be updated to [this](https://github.com/bertdeblock/rfcs/blob/deprecate-bower-apis/text/0772-deprecate-bower-apis.md)? Thanks for the work on this 🎉

bertdeblock:
@backspace Updated. Thanks!

ef4:
This looks good to me. 

I think the other key thing to deprecate is the part of the build pipeline that incorporates files from the bower directory (`/bower_components` by default). A good place to emit the deprecation warning would be here: https://github.com/ember-cli/ember-cli/blob/e747ac4c21a8e4a37e158a226dfa8ac048541b1f/lib/broccoli/default-packager.js#L632

And the warning could be something like:

> Building bower packages has been deprecated. You have bower packages in ${bowerDirectory}.

The deprecation guide would tell people to switch to either:

 1. Copy the bower package contents into `/vendor`, which would let you use it much the same way as before, just without bower.
 2. Consume the dep via NPM instead and use ember-auto-import instead of `app.import`.

bertdeblock:
@ef4 Thanks for the feedback! Updated accordingly.

rwjblue:
We discussed this in the ember-cli team meeting today, and are moving this into final comment period.

kategengler:
We decided to merge in CLI meeting today as there has been no new discussion.

PR #774, opened 2021-11-14
Deprecate implicit record loading in Ember Route

### LINKED ISSUE #773: Deprecate default store and record loading in Ember route
With #508 and ultimately removing the ember-data `store` injection [here](https://github.com/emberjs/data/pull/7667), the use of the default store👇 is uncommon and in some (many?) cases, [error prone](https://github.com/emberjs/data/pull/7744).  This is further complicated that Ember Data Model class instances don't have a `find` method, thus causing an assertion.

We should deprecate the default store to 

https://github.com/emberjs/ember.js/blame/017b11e2f58880869a5b8c647bf7f3199fc07f26/packages/%40ember/-internals/routing/lib/system/route.ts#L1729-L1760

### Motivation
Confusion.  Comment from an RFC...

> BIG fan of this RFC. A few weeks ago I spent half a day trying to figure out why our app was doing an unexpected request, until I stumbled upon this gem of functionality

### Notes

This RFC does not intend to remove the [default](https://github.com/emberjs/ember.js/blame/017b11e2f58880869a5b8c647bf7f3199fc07f26/packages/%40ember/-internals/routing/lib/system/route.ts#L1376) `model` hook and `findModel`.  Although this is entirely possible to expand this RFC to include this if people so desire.  Otherwise, I'd be happy to open a new RFC to accomplish this.

snewcomer:
[Rendered](https://github.com/snewcomer/rfcs/blob/implicit-record-route/text/0774-implicit-record-route-loading.md)

close #773 
close #557


text/0774-implicit-record-route-loading.md
```diff
@@ -0,0 +1,85 @@
...
+  }
+}
+```
+
+In this example, the `model` hook is not defined.  However, Ember will attempt to try a few things before rendering this route's template.
+
+1. If there is a `store` property on your route, it will attempt to call it's `find` method.  Assuming you have `ember-data` installed, you may be expecting this. The arguments will be extracted from the params.
+  a. For example, if a dynamic segment is `:post_id`, there exists logic to split on the underscore and and find a record of type `post`.  
+
+2. As a fallback, it will attempt to defined a `find` method and use your `Model` instance's find method to fetch.  If no `find` method exists, an assertion is thrown.
```
bertdeblock, line 35:
```suggestion
2. As a fallback, it will attempt to define a `find` method and use your `Model` instance's find method to fetch.  If no `find` method exists, an assertion is thrown.
```

text/0774-implicit-record-route-loading.md
```diff
@@ -0,0 +1,85 @@
+---
+Stage: Initial
```
bertdeblock, line 2:
```suggestion
Stage: Accepted
```
I think.

runspired (approved):

wagenet:
@scottmessinger @runspired is there anything more that we need here other than an "official" review?

runspired:
@wagenet this is a framework team item not a data team item though Scott did pick it up from @NullVoxPopuli it seems to try to get it over the line. I think this became of increased importance because the troll behavior lead to increased bug reports and confusion once injections were eliminated for 4.x

NullVoxPopuli:
ah right! this supersedes https://github.com/emberjs/rfcs/pull/557 <-- I'll close this

snewcomer:
100%.  I’d be eager to do the legwork once merged if someone else didn’t pick it up.  Unsure if it needs a “sponsor”.  Let me know if someone can bring it up in the core team meeting!

ef4:
Framework core discussed this and everyone is in favor of the deprecation. 

A suggestion is that the design could use an [optional feature](https://github.com/emberjs/ember-optional-features) that disables the old bad behavior. This would let people clear the deprecation by enabling the feature, rather than requiring them to create new empty model hooks and potentially new route files.

snewcomer:
@ef4 Thanks and updated with your suggestion for the optional feature ([PR](https://github.com/emberjs/rfcs/pull/774/commits/a3d9b56cd7eec1360bc4bcbed13e4fb4c2725f78)).  What is the process for merging this RFC?

ef4:
> What is the process for merging this RFC?

Core framework team needs to put it into final comment period, wait the customary week so people have a chance to comment, and then merge it. 

I'll make sure it comes up at the next meeting.

wagenet:
@snewcomer looks like it may have gotten dropped on accident, sorry! I've officially added it to the agenda for our next meeting.

wagenet:
@snewcomer Sorry for the delay. Everything got thrown off with the holidays. I'll try to get this back on track.

wagenet:
I'm moving this to FCP. It needs a few tweaks for our new stages setup and will need to be retargeted at v6. I can help with this.

wagenet:
Relatedly, we should also deprecate the default store property in routes.

text/0774-implicit-record-route-loading.md
```diff
@@ -0,0 +1,89 @@
...
+  }
+}
+```
+
+In this example, the `model` hook is not defined.  However, Ember will attempt to try a few things before rendering this route's template.
+
+1. If there is a `store` property on your route, it will attempt to call it's `find` method.  Assuming you have `ember-data` installed, you may be expecting this. The arguments will be extracted from the params.
+  a. For example, if a dynamic segment is `:post_id`, there exists logic to split on the underscore and find a record of type `post`.  
+
+2. As a fallback, it will attempt to define a `find` method and use your `Model` instance's find method to fetch.  If no `find` method exists, an assertion is thrown.
```
wagenet, line 35:
```suggestion
2. As a fallback, it will attempt to define a `find` method and use your `Model` instance's find method to fetch.  If a `Model` cannot be found or if the found `Model` does not have a `find` method, an assertion is thrown.
```

text/0774-implicit-record-route-loading.md
```diff
@@ -0,0 +1,89 @@
...
+    // do stuff
+  }
+}
+```
+
+## Alternatives
+
+- Continue to provide fallback fetching behaviour but ensure no `assert` is called for users that have neither a store nor a `Model` with a `find` method.
+
+## Open Questions
```
wagenet, line 84: Let's just remove this if there aren't any.

wagenet (dismissed):
Looks good! In addition to the inline comments, can you add a "Drawbacks" section? I think it's sufficient to just note that when the implicit behavior is nice it saves typing and now you have to type more.

text/0774-implicit-record-route-loading.md
```diff
@@ -0,0 +1,89 @@
...
+
+By removing this behaviour, we will encourage developers to explicitly define what data is being fetched and from where.
+
+## Detailed design
+
+We will issue a deprecation to [`findModel`](https://github.com/emberjs/ember.js/blob/017b11e2f58880869a5b8c647bf7f3199fc07f26/packages/%40ember/-internals/routing/lib/system/route.ts#L1376) notifying the user that if they want to continue this behaviour of attempting to fetch a resource implicitly, they should try and replicate with their own explicitly defined `model` hook. 
+
+In addition, we will include an [optional feature](https://github.com/emberjs/ember-optional-features) to disable this feature and clear the deprecation.
+
+In v5.0.0, we will remove `findModel` and logic to determine arguments for this method.  This will not remove returning the `transition` context when no `model` hook is defined.
```
wagenet, line 49:
```suggestion
In v6.0.0, we will remove `findModel` and logic to determine arguments for this method.  This will not remove returning the `transition` context when no `model` hook is defined.
```

Windvis:
I think this also means that the [`routes-segments-snake-case rule`](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/routes-segments-snake-case.md) can be removed from the recommended ruleset?

I've been disabling that rule on every project I work on since we don't depend on the implicit record loading and it would be nice if we could remove it from our eslint config file 😄.

PR #776, opened 2021-11-16
Author Built-In Blueprints in TypeScript

cafreeman:
In order to support both TypeScript and JavaScript as official languages in Ember, we should enable blueprints to be written in TypeScript and (optionally) transformed to JavaScript. We should also convert all of the existing built-in blueprints to TypeScript to take advantage of this new functionality and lay the groundwork for future TypeScript inclusion efforts.

[rendered](https://github.com/cafreeman/rfcs/blob/typescript-blueprints/text/0776-typescript-blueprints.md)


rwjblue:
Thanks for working on this @cafreeman!

wagenet:
This seems like a great idea! I look forward to reviewing it in detail.

wagenet (commented):
I'm 100% on board with this and, at this point, don't have any concerns or further suggestions!

kategengler:
"How we teach this" should probably cover authoring blueprints, in as much as they are covered on the cli guides.

rwjblue:
We discussed this in today's ember-cli team meeting, and are generally in favor of moving forward. There were a few things called out during that discussion that we think needs to be updated:

* "How we teach this" should include information about updating the blueprint authoring guides (see https://github.com/emberjs/rfcs/pull/776#issuecomment-983940805)
* We should explicitly add the `--typescript` flag (supporting blueprint generation with typescript if the underlying blueprint was actually authored in typescript). We can also call out that adding `"typescript": true` to `.ember-cli` is a convenient way  to ensure that users in a project get typescript blueprints whenever possible (it could even suggest that the ember-cli-typescript blueprint be updated to add that for folks automatically).

@cafreeman - Once those tweaks are made, I think this will be ready for another look.

cafreeman:
Thanks for the feedback @kategengler and @rwjblue! I've just pushed an update that should address both issues that were brought up. Let me know if you've got any other comments!

rwjblue:
Thanks for the updates @cafreeman!! We discussed this at the last ember-cli team meeting, and are moving this into final comment period.

muziejus:
What is the plan for exposing typeScript in the ember documentation? Will the above be taught with a rationale for why the ts files look the way they do, with extra `interface`s? Or is the question of documenting the ts dx in Ember left to `ember-cli-typescript`, despite the fact that the ts blueprints will be part of Ember proper?

muziejus:
More concretely, does this mean that the component blueprint will look something like this:

```
<%= importStatement %>

<% if isTypeScriptProject interfaceStatement %> // or whatever

export default <%= defaultExport %>
```

and that will yield, for `ember g component foo`,

```

interface FooComponentArgs {}

export default class FooComponent extends Component<FooComponentArgs> {}
```

or 

```

export default class FooComponent extends Component {}
```

depending on the presence of the flag? If that's the case, doesn't this imply that "using ember with typescript" becomes part of the ember story, and not the `ember-cli-typescript` story?

Issue #781, opened 2021-12-16
Ember builds without conflicts of globals


simonihmig:
# *This is a draft of an RFC, for which I'd like to get some early feedback. Will then finalize and submit it as a real RFC.*

## Summary

Due to the use of the global namespace, especially for its AMD-style module system, Ember apps have problems coexisting 
with other Ember apps or any AMD-based JavaScript on the same page. This RFC proposes a way to prevent all potentially
conflicting usage of these globals.

## Motivation

The most common case for running an Ember app is to run it as a pure isolated SPA (Single Page Application), i.e. to have 
it *own* the HTML page it runs on. This is what you get out of the box when running `ember build`: a (mostly empty) 
index.html, that contains the build assets that run the Ember app inside that empty page. 

But another use case is to [embed it into an existing page](https://guides.emberjs.com/release/configuring-ember/embedding-applications/),
so that it runs alongside other JavaScript (let's call it "3rd party" from the perspective of the Ember app). While that
*might* work fine in some cases, due to Ember's still present reliance on globals, especially its AMD-style module 
system that is actually *incompatible* with the original one of [RequireJS](https://requirejs.org/), there is a high chance of non-trivial conflicts.

Let's look at a few examples:

### Embedding an Ember app into a 3rd party HTML page

You might want to embed an Ember app on a 3rd party page (i.e. one that you don't control), for example
* a messaging/chat widget embedded on a corporate/sales website (think of Intercom)
* a product configurator embedded on an e-commerce site (that's what we do)
* a messaging board app on a customer support website (think of Discourse, but embedded)

In these cases, if the 3rd party page uses RequireJS itself, only one implementation (Ember vs. RequireJS) of the globally 
defined `define()` and `require()` functions will "win", and the other's module loading will inherently be broken, in 
the worst case both.

This is not uncommon, for example [Magento](https://magento.com/), the most wide-spread e-commerce solution worldwide, 
[uses RequireJS](https://devdocs.magento.com/guides/v2.4/javascript-dev-guide/javascript/requirejs.html), so effectively
you cannot use an Ember app on its pages without *somehow* tackling that conflict.

### Micro Frontends

A rather new pattern in frontend development is the analogy of microservices for the frontend world, coined 
[Micro Frontends](https://martinfowler.com/articles/micro-frontends.html), where you develop, test and deploy parts of 
your frontend independently, and compose them together for a seamless user experience.

[...]

PR #784, opened 2022-01-07
Introduce `no-globals` optional feature

simonihmig:
[Rendered](https://github.com/simonihmig/rfcs/blob/no-globals/text/0784-no-globals.md)

---

For the record: there has been an early draft and some discussion around it here: https://github.com/emberjs/rfcs/issues/781. Special thanks to @bendemboski for contributing a lot of details about the ember-electron use case!

### DIFF
--- a/text/0784-no-globals.md
+++ b/text/0784-no-globals.md
@@ -1,0 +1,494 @@
+---
+Stage: Accepted
+Start Date: 2022-01-07
+Release Date: Unreleased
+Release Versions:
+ember-source: vX.Y.Z
+ember-data: vX.Y.Z
+Relevant Team(s): Ember.js, Ember CLI
+RFC PR: https://github.com/emberjs/rfcs/pull/784
+---
+
+# `no-globals` optional feature
+
+## Summary
+
+Due to the use of the global namespace, especially for its AMD-style module system, Ember apps have problems coexisting
+with other Ember apps or any AMD- or CJS-based JavaScript within the same page/context. This RFC proposes a way to prevent 
+all potentially conflicting usage of these globals.
+
+## Motivation
+
+The most common case for running an Ember app is to run it as a pure isolated SPA (Single Page Application), i.e. to have
+it *own* the HTML page it runs on. This is what you get out of the box when running `ember build`: a (mostly empty)
+index.html, that refers to the build assets that run the Ember app inside that empty page.
+
+But another use case is to [embed it into an existing page](https://guides.emberjs.com/release/configuring-ember/embedding-applications/),
+so that it runs alongside other JavaScript (let's call it "3rd party" from the perspective of the Ember app). While that
+*might* work fine in some cases, due to Ember's still present reliance on globals, especially its AMD-style module
+system that is actually *incompatible* with the original one of [RequireJS](https://requirejs.org/), there is a high
+chance of non-trivial conflicts.
+
+Let's look at a few examples:
+
+### Embedding an Ember app into a 3rd party HTML page
+
+You might want to embed an Ember app on a 3rd party page (i.e. one that you don't control), for example
+* a messaging/chat widget embedded on a corporate/sales website (think of Intercom)
+* a product configurator embedded on an e-commerce site (that's what we do)
+* a messaging board app on a customer support website (think of Discourse, but embedded)
+
+In these cases, if the 3rd party page uses RequireJS itself, only one implementation (Ember vs. RequireJS) of the globally
+defined `define()` and `require()` functions will "win", and the other's module loading will inherently be broken, in
+the worst case both.
+
+This is not uncommon, for example [Magento](https://magento.com/), the most wide-spread e-commerce solution worldwide,
+[uses RequireJS](https://devdocs.magento.com/guides/v2.4/javascript-dev-guide/javascript/requirejs.html), so effectively
+you cannot use an Ember app on its pages without *somehow* tackling that conflict.
+
+### Micro Frontends
+
+A rather new pattern in frontend development is the analogy of microservices for the frontend world, coined
+[Micro Frontends](https://martinfowler.com/articles/micro-frontends.html), where you develop, test and deploy parts of
+your frontend independently, and compose them together for a seamless user experience.
+
+The same issue arises when one of those micro frontends happens to use RequireJS, or two or more frontends are actual
+Ember apps. While in the latter case we don't have to deal with different implementations of `define()` and `require()`,
+those functions close over their module registry, so whatever functions win (as being the globally defined one), they will
+only be able to see *their* registered modules, the modules of the other app will not exist anymore. And we are not even
+speaking about conflicting versions of the same modules (say `@ember/component`). So in any way, that scenario is bound
+to fail miserably.
+
+### Running an Ember app in a node.js environment
+
+Node.js introduced its own module system, CommonJS (CJS), which happens to also define a global `require()` function.
+When an Ember app runs in a node.js environment, it is able to `require()` and run node.js code. This obviously leads to
+conflicts with Ember's AMD-style `require()`.
+
+In practice this is a problem for [ember-electron](https://ember-electron.js.org/), which is able to package an Ember app
+as a native Desktop app, and enable it to access (native) node.js modules. To work around the conflict, `ember-electron`:
+
+* injects [a script](https://github.com/adopted-ember-addons/ember-electron/blob/464b13c76a89b9803cf27fcd974a798e05cb109c/public/shim-head.js) into `index.html` before the app and vendor scripts that renames `window.require` (the Node.js require) to `window.requireNode`
+* injects [a script](https://github.com/adopted-ember-addons/ember-electron/blob/464b13c76a89b9803cf27fcd974a798e05cb109c/vendor/wrap-require.js) into `vendor.js` (after the Ember code) that replaces `window.require` (which is now the `loader.js` require) with a custom `require` function that tries requiring via both module loaders (Node.js and `loader.js`)
+
+With this RFC implemented, there would still be some ambiguity where bundled app code calls `require()`, as that could refer
+both to Ember's AMD version of it and to node's CJS. But using the already [recommended `requireNode()`](https://ember-electron.js.org/docs/guides/development-and-debugging#require-and-requirenode-) 
+instead would resolve that ambiguity. The other case however, where 3rd party node modules expect the global `require()`
+to be the CJS version and therefore forcing ember-electron into overloading Ember's global `require()` to also support node 
+modules, would be solved by this RFC, as Ember wouldn't override node's global `require()` anymore.
+
+## Detailed design
+
+The guiding principle suggested in this RFC is to make Ember apps practically side effect free, with respect to the
+usage of the browser's global namespace (`window`/`globalThis`).
+
+That means it should be discouraged to
+* read custom variables from the global namespace
+* write anything to the global namespace
+
+While this can only be a recommended but not enforceable best practice for user-land code, for Ember and its wider build
+tooling ecosystem this should become a rule. That specifically includes:
+* Ember.js (its runtime code)
+* Ember-CLI
+* ember-resolver
+* ember-auto-import
+* Embroider
+* ember-engines
+
+Where this is practically not possible, it should be ensured that there is no realistic chance of a conflict, by
+* naming global variables in a unique way, e.g. scoped to the app's name (in `package.json`), so that two different Ember
+  apps cannot cause conflicts
+* ideally also adding a non-deterministic part like a random string that changes *per build*, that is only known to
+  privileged code (as listed above), so user-land code cannot rely on the existence of these globals.
+
+While the `Ember` global has already been deprecated and removed from Ember 4 as part of [RFC 706](https://emberjs.github.io/rfcs/0706-deprecate-ember-global.html),
+there are still other variables that Ember apps leak into the global namespace as described below, most notably Ember's
+AMD-like module system, and as such pose a risk for conflicts.
+
+### Optional feature `no-globals`
+
+The changes proposed here should ideally only require changes in Ember's build tooling. For idiomatic Ember code, especially
+code that only uses ES modules, it is a non-breaking change. Also code that uses Ember's AMD-style module system
+directly like `require()` or `require.entries` should continue to work without any changes in most cases, as long as the
+code is built by Ember-CLI (see below).
+
+> Although the AMD-like module API might not be strictly a public API of Ember, but it is often used in the wild to be at
+> least regarded as "intimate API". So for the scope of this RFC, we will consider it equivalent to public API.
+
+But there might be cases where people are doing things differently, and code - even after the build tooling changes
+outlined below - still effectively relies on the globally defined AMD-like module API. For this reason we may not just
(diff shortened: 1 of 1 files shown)


text/0784-no-globals.md
```diff
@@ -0,0 +1,494 @@
...
+  return EmberENV;
+})(window.EmberENV || {}, {"FEATURES":{},"EXTEND_PROTOTYPES":{"Date":false},"_APPLICATION_TEMPLATE_WRAPPER":false,"_DEFAULT_ASYNC_OBSERVERS":true,"_JQUERY_INTEGRATION":false,"_TEMPLATE_ONLY_GLIMMER_COMPONENTS":true});
+```
+</details>
+
+But as you can see there, you can also inject those by setting a globally defined `window.EmberENV`.
+
+In `no-globals` mode though, the latter behavior will be disabled. Only the variables built into `vendor.js` are taken
+into account. Implementation wise this can happen in the same way as described above, by defining `EmberENV` within the
+IIFE (and not using the explicit `window.` namespace).
```
rwjblue, line 449: I think it would be better to have Ember leverage `@embroider/macros` `getConfig` or `getOwnConfig` inside Ember's own code where we currently read `window.EmberENV` instead of leveraging the IIFE style.

rwjblue (commented):
Broadly, this seems great! I've left a few inline comments, but overall I'd like to see us move away from requiring the IIFE solution (in favor of providing an API more broadly that folks can leverage to do the app's own `require`/`define` as needed directly).

simonihmig:
> overall I'd like to see us move away from requiring the IIFE solution 

See my other inline comment about this. The idea was to let any direct use of `require()` continue to work...

> (in favor of providing an API more broadly that folks can leverage to do the app's own require/define as needed directly).

Not sure I get what you mean here. Who is "folks"? Like I would see user-land code to only use ESM, and eventually `@embroider/macros` (like `importSync()`) for the special cases. But not expose anything new here as a public API for *users*. 

Or do you refer to what I called the privileged addons, so they use some private API to get references to define/require? But again, this would require either breaking user-land code, or transforming it also. The IIFE was meant to let user code continue to work, without transformation...

Which brings me to the overarching question of how deep we should start to dig into the *implementation details* of this RFC. I did all this IIFE exploration only to prove that it's feasible to make this work *somehow* (I literally changed the `/dist` code by hand to see that the approach works), while intentionally stating in the RFC that those details are not normative. The TLDR of the publicly visible changes for users is basically: there is a new optional feature, that when enabled prevents leaking globals. For users everything should continue work, except for some exotic edge cases (code that is not processed through the normal build pipeline and which is relying on those globals). Everything else are implementation details.

Or do you think the RFC must specify the *actual* implementation path?

text/0784-no-globals.md
```diff
@@ -0,0 +1,494 @@
...
+must make sure they are scoped to the current app, so they pose no risk of conflicts.
+
+So for `vendor.js` this is what the basic change could look like:
+
+
+<details>
+<summary>vendor.js</summary>
+
+```diff
++ (function(){
```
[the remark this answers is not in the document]
simonihmig: Also note that I mentioned under "Alternatives" that we could also drop all support of user-land code calling the AMD functions. Which would be more invasive, breaking some addon's code. Still no breaking change, as that is opt-in only.

Then indeed we don't need that IIFE nor any addon code transformations...

text/0784-no-globals.md
```diff
@@ -0,0 +1,494 @@
...
+* Ember.js (its runtime code)
+* Ember-CLI
+* ember-resolver
+* ember-auto-import
+* Embroider
+* ember-engines
+
+Where this is practically not possible, it should be ensured that there is no realistic chance of a conflict, by
+* naming global variables in a unique way, e.g. scoped to the app's name (in `package.json`), so that two different Ember
+  apps cannot cause conflicts
```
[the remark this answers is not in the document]
rwjblue: In general, folks should think of an engine as another app (in that it's self contained/isolated). I was mostly wondering if we should advise folks to scope engine related globals to the engine name or to a combination of the app name and the engine name.

Totally fine to not have an answer, I don't think we have to have concrete guidance here for that (we can make an issue over on ember-engines when this RFC merges)

text/0784-no-globals.md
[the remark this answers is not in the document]
rwjblue: > But wouldn't that break any addon code that still uses e.g. require()?

Yes 😈 , but that's the point of creating an optional feature anyways (it's opt-in, we are allowed to limit semantics).

---

> The idea here was to prevent leaking those into the global namespace, while still allowing all existing addon code that uses these things in the wild (most notably ember-resolver) to still work. That is as long as that addon code is processed through the normal Ember CLI build process, i.e. it lands within the IIFE. When you call require from somewhere else like an external <script>, that would cease to work, hence the optional feature to allow this "intimate API" to break in such rare cases.

Ya, I think we should provide a global that is app scoped (e.g. `globalThis._appNameGlobals.require` and `globalThis._appNameGlobals.define`) instead of continuing to use `define`/`require` unscoped. This would require some tweaks to the babel setup (e.g. we'd need a custom version of https://github.com/babel/babel/blob/main/packages/babel-plugin-transform-modules-amd/src/index.ts or send a PR to allow the `define`/`require` hard coded constants to change).

---

> Note that I am referring to addons directly calling require() etc., like [here](https://github.com/miragejs/ember-cli-mirage/blob/d99571d9010011976b757f40cca8450c71dcb070/packages/ember-cli-mirage/addon/utils/read-modules.js#L62), and not regular ESM imports that happen to be transformed to define/require during the build.

I actually think this case is **not** a good example. Specifically because that is not using `require` global, it is using `import require from 'require';`:


https://github.com/miragejs/ember-cli-mirage/blob/d99571d9010011976b757f40cca8450c71dcb070/packages/ember-cli-mirage/addon/utils/read-modules.js#L8

^ pattern should still continue to work just fine (because whatever loader happens to be running, will respond to that import with the correctly scoped `require`).

rwjblue:
> Not sure I get what you mean here. Who is "folks"? Like I would see user-land code to only use ESM, and eventually @embroider/macros (like importSync()) for the special cases. But not expose anything new here as a public API for users.

Ya makes sense! I think I agree with you that 99.9% of users won't have to ever care what we name the app specific globals (from my comment above that might be `globalThis._appName.require`/`globalThis._appName.define`), but I do think it should be _possible_ for that 0.01% case to be written. However, I think my other comments address this as well: users in that edge case can leverage `getConfig` to find out the scoped property name.

> Which brings me to the overarching question of how deep we should start to dig into the implementation details of this RFC.

I think that we need to explain enough of how it will work for folks to be able to have a mental model. In my mind that "simple" mental model is that when you opt-in to the optional feature, we don't use `define`/`require` globals at all and instead we put those in a scoped globals bucket for the app itself (like I mentioned above). We don't have to get into all the places that will need to be updated/fixed for that to work though.

text/0784-no-globals.md
chancancode: It seems fine to still support _reading_ from `window.EmberENV` even in no-globals mode? Otherwise we need to propose _something_ for the Ember Inspector to [enable the debug render tree](https://github.com/emberjs/ember-inspector/blob/15f49e5ab04e738ec540550001ad8a1b1537eb70/skeletons/web-extension/boot.js#L13).

chancancode: Actually, come to think about it, I think the inspector in its current form/architecture will most likely just completely break in no-globals mode as it relies on being able to get access to the things you are trying to no leak with this RFC. Probably not a good enough reason to hold this back, but it's definitely a drawback that should be noted. We can overhaul the inspector in theory but it will require more design than is in this RFC currently, and it's probably not fair to have this RFC solve that problem, plus I am not sure if anyone has time to work on a big inspector project. In the short/medium term, I think the inspector will probably just not work with `no-globals` mode. Is that an acceptable drawback?

chancancode: (cc @rwwagner90)

text/0784-no-globals.md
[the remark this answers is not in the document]
simonihmig: > Yes 😈 , but that's the point of creating an optional feature anyways (it's opt-in, we are allowed to limit semantics).

Yes, which is basically what I say here, right? 👇

> Also note that I mentioned under "Alternatives" that we could also drop all support of user-land code calling the AMD functions. Which would be more invasive, breaking some addon's code. Still no breaking change, as that is opt-in only.

At least when they are called as globals. Which the current form of the RFC still tries to support...

So to get this 100% clear: you are suggesting that we do this alternative approach, i.e. stop supporting the global use of `require()` & Co as globals all together, right? 

This does mean that we need to "fix" a lot more of the core addons, like `ember-resolver`, which AFAICT does use `require()` as a global [here](https://github.com/ember-cli/ember-resolver/blob/master/addon/resolvers/classic/index.js)

> I actually think this case is not a good example. Specifically because that is not using require global

Oh, you are right, I missed that! It's not so easy to search for examples of addons doing `require()` using emberobserver's code search, given that you get a bazillion hits for legit use of `require()` in node-land... 🙈

>  pattern should still continue to work just fine (because whatever loader happens to be running, will respond to that import with the correctly scoped require)

While that is possible, I wonder if we *actually* want that. Probably out of scope for this RFC, but shouldn't we eventually try to get rid of any usage of loader.js stuff at all? 

Like in a future where we have Polaris and Embroider by default, shouldn't we tell users that the only way to import stuff is via ESM imports (and `@embroider/macros`)? So that we can *eventually* get to a point where we could get rid of any specific runtime module system like loader.js, when we have just ESM imports at stage 3, then throw that at webpack, which then emits whatever it emits. 

So maybe turning on `no-globals` is actually a good moment to make *all* of `require()` stop working for unprivileged code?

simonihmig:
> but I do think it should be possible for that 0.01% case to be written.

Can you give an example of a legit use case where you would absolutely need this, and cannot express this in terms of regular imports or the `importSync` macro?

If we make this possible for unprivileged code, aren't we then basically introducing a new public API, that we need to support? Or at least kind of replacing one intimate API with another one? See also my other inline comment about whether to support `import require from 'require'`...

> users in that edge case can leverage getConfig to find out the scoped property name

Speaking of the implementation detail of how to get hold of the scoped version of `define` and `require`, are you sure we actually need a "runtime" (in the sense that it goes into app code) thing like `getConfig`? I was rather thinking about something that is available for the build system, i.e. in node.js. So something that you could use for example in `ember-auto-import` [here](https://github.com/ef4/ember-auto-import/blob/29587ab84358f53ffbb886bb323f1d5aea3b99e2/packages/ember-auto-import/ts/webpack.ts#L77-L87).

text/0784-no-globals.md
simonihmig: Oh, that's a good point! I didn't think about that. Certainly needs to be mentioned. I don't know enough about the Inspector to assess how much needs to change to support `no-globals`. I guess all those possible ways to coordinate between those privileged tools what the scoped version of `require()` & Co are (e.g. `getConfig`) will not work for the Inspector, as it is a standalone app that does not take part in the build system of the *app*...

Maybe there still could be some escape hatch like defining `window.__ember_scoped_globals_dont_ever_use_this_in_user_code = 'myEmberApp_somerandomhash'`, when the app has `require` (and `define`) scoped as window.myEmberApp_somerandomhash.require`?

> Is that an acceptable drawback?

I would still say yes. You can still define the flag conditionally, like enable it for test and prod, while disabling it for dev!

wagenet:
It looks like this still has a good bit of interest so I'll see what I can do to keep it moving forwards.

ef4:
I think the framing of this RFC is too focused on what not to do ("don't use globals") when the problem would also go away if we clarify what *to* do instead.

I would rather solve the same problems by clarifying exactly how does one boot an Ember app, starting from only web standards like ES modules. It's a problem that our boot process is currently implementation-defined. This RFC is about changes in that implementation. But the implementation details should be invisible and shouldn't require an RFC at all! We need to get the implementation details behind public API and web standards.

jamesarosen:
> We need to get the implementation details behind public API and web standards.

I'm interested to hear what standards you might have in mind. I haven't seen any for "how to boot an application within the DOM." The closest things I can think of are

1. [Web Components](https://developer.mozilla.org/en-US/docs/Web/API/Web_components), which defines `constructor`, `connectedCallback`, `disconnectedCallback`, `adoptedCallback`, and `attributeChangedCallback` as the lifecycle API.
2. The [single-spa application lifecycle](https://single-spa.js.org/docs/building-applications), which defines `bootstrap`, `mount`, and `unmount` as the lifecycle API. single-spa isn't a web standard, but it is reasonably popular.

In either case, there would have to be some non-module entrypoint JS (`<script type="text/javascript">`) to kick things off. Everything after that could be a `<script type="module">`.

ef4:
> In any case, there would have to be some non-module entrypoint JS (<script type="text/javascript">) to kick things off. 

Why non-module? I suspect it can be all modules all the way down.

> I'm interested to hear what standards you might have in mind.

Something like:
 - in an ES module
 - import a particular function from ember and call it with arguments that include
   - an HTMLElement to render into
   - some description of the ES modules that define your application, so that it can load them when it want to
   - and maybe something like a [Location](https://github.com/emberjs/ember.js/blob/87c7fc72a3112936ae701a6768072d31caf4730d/packages/%40ember/routing/location.ts#L62) that mediates all access to the current URL.

jamesarosen:
> Why non-module? I suspect it can be all modules all the way down.

You're absolutely right. `<script type="module" src="/ember-entrypoint.js">` works great.

> Something like…

This is very similar to the single-spa API. Each application is defined as a single top-level module that exports `{ boostrap, mount, unmount }`.

They are registered with the application loader with this API:

```js
registerApplication({
  name: '@organization/app2',
  app: () => import('/src/app2/main.js'),
  activeWhen(location) => location.pathname.startsWith('/app2'),
  customProps(appName, location) => ({}),
);
```

or the HTML version of the same:

```html
<script type="importmap">
  {
    "imports": {
      "@organization/app2": "/src/app2/main.js"
    }
  }
</script>
<route path="/app2">
  <application name="@organization/app2"></application>
</route>
```

simonihmig:
> I would rather solve the same problems by clarifying exactly how does one boot an Ember app, starting from only web standards like ES modules. It's a problem that our boot process is currently implementation-defined. This RFC is about changes in that implementation. But the implementation details should be invisible and shouldn't require an RFC at all! We need to get the implementation details behind public API and web standards.

Sure, I fully agree. But the goal of this RFC was not a full-scale major change, but rather a first stepping stone towards that future, like making the AMD implementation not leak into the outside world, while still preserving compatibility inside the Ember world. The use of AMD leaks so much into the ecosystem, that it is not really an implementation details as it should, but rises to the point of being a public API, that's why I introduced a first opt-in feature to scale that down by making it invisible from the outside world. I did also have in mind to maybe deprecate it as part of this RFC, but chose to leave that for a later RFC.

Having said that, I don't think it makes much sense to keep working on this RFC at this point, so I will close this in favor of https://github.com/emberjs/rfcs/pull/938, which I think goes far beyond what is proposed here, so that stepping stone is not really needed anymore. Even what that RFC might not cover the use cases brought up here (idk, does it?), and when to cover those we still need another RFC, it probably makes sense to define that one on top of #938.

PR #788, opened 2022-01-13
in-implementation updates to the default helper manager RFC, 756

NullVoxPopuli:
During implementation, a couple things came up:
 - we don't do a great job of documenting when autotracking happens outside of getters  -- therefore, added an example of autotracking in a helper to the examples section -- as composing helpers _in_ classes that have access to tracked data can be a powerful encapsulation tool for library and app authors alike
 
  - we need a minor tweak to how arguments are passed to the helper function to _enable_ default arguments for helpers relying on positional args
  
  Much thanks to @chancancode for binging these things up.
  
Also, Idk how this is supposed to work with staged RFCs, :shrug: I guessed (feedback welcome, as always)

[Rendered](https://github.com/NullVoxPopuli/rfcs/blob/helper-manager-updates/text/0756-helper-default-manager.md)

### DIFF
--- a/text/0756-helper-default-manager.md
+++ b/text/0756-helper-default-manager.md
@@ -67,13 +67,15 @@ of the three constructs: Helpers, Modifiers, and Components._
 The desired usage of a plain function in a template should be:
  - convenient
  - reduce boilerplate
- - be easily portable to JS for developers' mental model of how template and JS interact.
+ - be easily portable to JS for developers' mental model of how template and JS interact
+ - support normal JavaScript idioms and existing JavaScript functions (from lodash, etc).
 
 Which results in:
  - default to positional parameters
  - all named arguments are grouped into an "options object" as the last parameter.
    this happens to align with the _syntax_ of helper invocation where named arguments may not appear
    before the last positional argument.
+ - if no named arguments are passed in the template, no "options object" is passed to the JS call.
 
 #### Example with mixed params
 
@@ -112,6 +114,26 @@ class A {
 }
 ```
 
+#### Example with consuming tracked data defined outside of the helper
+
+This works with the `helper` and `Helper` from `@ember/component/helper`, as well as plain functions.
+
+```hbs
+<output>{{this.multiply 4}}</output>
+
+<button {{on 'click' this.increment}}>Increment</button>
+```
+```js
+class A {
+  @tracked multiplicand = 5;
+
+  multiply = (passed) => passed * this.multiplicand;
+
+  increment = () => this.multiplicand++;
+}
+```
+When the button is clicked, the text in `<output>` will update, even though the `multiplicand` is not passed to the helper.
+
 #### Example Default Helper Implementation
 
 The implementation for the this function-handling helper-manager could look like this:
@@ -136,8 +158,6 @@ class FunctionHelperManager {
 
     if (Object.keys(args.named).length > 0) {
       argsForFn.push(args.named);
-    } else {
-      argsForFn.push({});
     }
 
     return fn(...argsForFn);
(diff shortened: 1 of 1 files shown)


text/0756-helper-default-manager.md
```diff
@@ -156,6 +176,57 @@ setHelperManager(() => DEFAULT_HELPER_MANAGER, Function.prototype);
  - to register this helper manager, it should occur during app boot so developers do not need to import anything to
    trigger the `setHelperManager` call
 
+**Update 2022-01-13, during implementation**
+During the implementation of this feature, https://github.com/glimmerjs/glimmer-vm/pull/1348
+an issue came up where this RFC doesn't _explicitly_ say what to do with undefined arguments and default arguments.
+
+For example, it was originally assumed that the _last arg_ passed to the helper-function would always be the
+named arguments -- "options-style". This is problematic for positional-only APIs where someone may want to do
```
chancancode, line 184:
```suggestion
In this RFC, we proposed that the _last argument_ passed to the helper-function would be the
named arguments as a POJO. It was previously assumed that this argument will _always_ be
passed to the helper, even when there are no arguments present in the invocation, in which
case an empty object (`{}`) will be passed.

This is problematic for JavaScript functions relying on optional (positional) arguments. For
example:
```

text/0756-helper-default-manager.md
```diff
@@ -156,6 +176,57 @@ setHelperManager(() => DEFAULT_HELPER_MANAGER, Function.prototype);
...
+```js
+function myHelper(a, b = 3) {
+  return a * b;
+}
+```
+_with only_ positional arguments, this works as expected, but if named arguments get involved,
+```hbs
+{{myHelper 1 optionA=true}}
+```
+then the named args are passed in place of `b`, which is expected to be a number.
```
chancancode, line 200: (This is not quite right with the original proposal)

```suggestion
With the original approach, the first invocation would work as intended, but the second invocation
result in `NaN` unexpectedly. (This is because the function would be called as `myHelper(1, {})`.)

For this reason, we propose to amend the RFC such that the options argument is only passed if
there is at least one named argument in the invocation. In the example above, the function will be
called as `myHelper(1, 2)` and `myHelper(1)`, respectively.
```

text/0756-helper-default-manager.md
```diff
@@ -156,6 +176,57 @@ setHelperManager(() => DEFAULT_HELPER_MANAGER, Function.prototype);
...
+    // potentially assume named arguments
+    options = maybeOptions;
+    positional = reversed.reverse();
+  } else {
+    // maybe a number was passed
+    positional = args;
+  }
+
+  // ...
+}
```
chancancode, line 227: You can adjust this accordingly based on the changes above. I think the mainly we should talk (in sentences, not paragraphs) about:

* The drawbacks
  * The consistency issue
  * Hazards when mixing defaulted positional arguments with named
  * This needs to be preserved if we add a spread operator/operation in handlebars
* The bottomline
  * Defaulted positional arguments are very common in functions, especially ones you import from libraries like loadash
  * Mixing defaulted positional arguments with options is relatively uncommon because they have the same hazard in plain JavaScript
  * Therefore this is probably worth it

text/0756-helper-default-manager.md
```diff
@@ -108,6 +108,26 @@ class A {
...
 ```
 
+#### Example with consuming tracked data defined outside of the helper
+
+This works with the `helper` and `Helper` from `@ember/component/helper`, as well as plain functions.
+
+```hbs
+<output>{{this.multiply 4}}</output>
+
+<button {{on 'click' this.increment}}>Increment</button>
```
chancancode, line 118: This isn't really technically a helper per-se, the first example is probably good enough

text/0756-helper-default-manager.md
NullVoxPopuli: Updated! thanks!

wycats:
Notes from our meeting:

- [ ] Motivation: "it looks like a JavaScript call"
- [ ] Idiom Constraint: We want to support normal JavaScript idioms from Handlebars. This means that we need to support optional positionals, and we also would like to turn named arguments into a options object. However, in idiomatic JavaScript, it is difficult to have optional positional arguments *and* named arguments without doing a lot of manual checking. We are intentionally leaning into the JavaScript call idioms *over* the Handlebars call idioms, which *do* support that pattern.

text/0756-helper-default-manager.md
```diff
@@ -156,6 +176,68 @@ setHelperManager(() => DEFAULT_HELPER_MANAGER, Function.prototype);
  - to register this helper manager, it should occur during app boot so developers do not need to import anything to
    trigger the `setHelperManager` call
 
+**Update 2022-01-13, during implementation**
```
rwjblue, line 179: I find this odd to be in it's own section here, why isn't this updating the detailed design just above (or at least _also_ updating the detailed design)?

NullVoxPopuli:
I no longer have the energy to put in to this, so if someone wants to take it over, that'd be much appreciated

text/0756-helper-default-manager.md
NullVoxPopuli: Because it made sense to me, and i wanted to progression of discovery to be known. I haven't been aware of any 'proper' way to do these kinds of updates 🙃

chancancode:
I think this is currently hung up on the text is still not very clear to everyone who read it/didn't capture everything we want to convey with the change succinctly. I can eventually wrap this up (or someone else could) but it's not a huge priority for me personally, because the feature was [landed](https://github.com/emberjs/ember.js/issues/20054) anyway. It's important to capture this in the RFC as the source of truth but it's not a blocker, though it's important to do it while we still have the context.

text/0756-helper-default-manager.md
```diff
@@ -108,6 +108,26 @@ class A {
...
+#### Example with consuming tracked data defined outside of the helper
+
+This works with the `helper` and `Helper` from `@ember/component/helper`, as well as plain functions.
+
+```hbs
+<output>{{this.multiply 4}}</output>
+
+<button {{on 'click' this.increment}}>Increment</button>
+```
+```js
```
MrChocolatine, line 120:
```suggestion
```ts
```

Because of

> multiply = (passed: number) => passed * this.multiplicand;

wagenet:
@chancancode do your thoughts remain the same here?

chancancode (approved):

PR #793, opened 2022-02-11
EmberData | Polymorphic Relationship Support

yratanov:
Please don't judge me, it's my first RFC :)

Some discussion here: https://github.com/emberjs/data/issues/6728

[rendered](https://github.com/yratanov/rfcs/blob/polymorphic-relations-without-inheritance/text/0793-polymporphic-relations-without-inheritance.md)

### DIFF
--- a/text/0793-polymporphic-relations-without-inheritance.md
+++ b/text/0793-polymporphic-relations-without-inheritance.md
@@ -52,12 +52,13 @@ Now consider that this same post and comment class also both need to be "viewabl
 
 ```js
 import Model, { belongsTo, hasMany } from '@ember-data/model';
+import Mixin from "@ember/object/mixin";
 
-class TagModel extends Model {
+class TaggableMixin extends Mixin {
   @belongsTo('taggable', { polymorphic: true }) taggable;
 }
 
-class ViewModel extends Model {
+class ViewableMixin extends Mixin {
   @belongsTo('viewable', { polymorphic: true }) viewable;
 }
 
@@ -72,25 +73,87 @@ class CommentModel extends Model.extends(TaggableMixin, ViewableMixin) {
 
 This RFC proposes to allow explicitly declaring polymorphic traits similar to [rails conventions](https://guides.rubyonrails.org/association_basics.html#polymorphic-associations).
 
+### API
+
+The inverse relationship, if one exists, of a polymorphic relationship must be concrete (e.g. not also polymorphic).
+
+The polymorphic relationship **MUST** explicitly declare itself as polymorphic and **MUST** explicitly declare it's inverse as either a key or `null`. All records that satisfy the polymorphic type must declare a matching inverse via that key.
+
+Polymorphic relationships take the form:
+
+\```ts
+hasMany(baseType: string, options: { polymorphic: true, inverse: string | null }) 
+belongsTo(baseType: string, options: { polymorphic: true, inverse: string | null }) 
+\```
+
+Concrete Relationships which fulfill polymorphic relationships take the form:
+
+\```ts
+hasMany(recordWithPolymorphicRelationship: string, options: { inverse: string, as: string }) 
+belongsTo(recordWithPolymorphicRelationship: string, options: { inverse: string, as: string }) 
+\```
+
+So for instance, given a field named "tagged" that is a polymorphic hasMany with baseType "taggable", and inverse key "tags" the following would be the polymorphic relationship and concrete relationship.
+
+\```ts
+// polymorphic relationship
+class Tag extends Model {
+   @hasMany("taggable", { polymorphic: true, inverse: "tags" }) tagged;
+}
+// an inverse concrete relationship
+class Post extends Model {
+   @hasMany("tag", { inverse: "tagged", as: "taggable" }) tags;
+}
+\```
+
+### Example
+
 ```js
 import Model, { belongsTo, hasMany } from '@ember-data/model';
 
 class TagModel extends Model {
-  @belongsTo('taggable', { polymorphic: true }) taggable;
+  @belongsTo('taggable', { polymorphic: true, inverse: 'tag' }) taggable;
 }
 
 class ViewModel extends Model {
-  @belongsTo('viewable', { polymorphic: true }) viewable;
+  @belongsTo('viewable', { polymorphic: true, inverse: 'views' }) viewable;
 }
 
 class CommentModel extends Model {
-  @hasMany('views', { inverse: 'viewable' }) views;
-  @hasMany('tag', { inverse: 'taggable' }) tags;
+  @hasMany('views', { inverse: 'viewable', as: 'viewable' }) views;
+  @hasMany('tag', { inverse: 'taggable', as: 'taggable' }) tags;
 }
 ```
 
 `taggable` and `viewable` in this example should be typed as `extends Model`, so any model can take that spot. 
 JSON:API has enough information (`type` and `id`) to set this relation.
+
+### Polymorphic Relationships without Inverses
+
+- polymorphic relationships need not have an inverse, in which case they must specify `inverse: null`.
+
+### Base Types
+
+Every polymorphic relationship has an implicit "base type" that represents the entity on the other side of the relationship.
+
+In the below example, the polymporphic relationship implies the existence of the `base type` "taggable".
+
+\```
+class Tag extends Model {
+  @hasMany("taggable", { inverse: "tags", polymorphic: true }) taggables;
+}
+\```
+
+It is only required that your app implement a model (or if using instantiateRecord and the schema service implement handling for that type) if your API will return entities whose type is the base type. If the base type is abstract and never directly interacted with, it is not required to implement a Model or other associated support.
+
+### Limitations
+
+- polymorphic to polymorphic relationships are not allowed. One side of the relationship MUST be a concrete entity type or have no inverse.
+- this includes self-referential relationships
+
+### Deprecation Design
+
+Situations in which polymorphism is not configured by the above mechanism but in which EmberData would have previously detected and attempted to do the right thing will be deprecated. This means effectively that mixin based and inheritance based polymorphism will print a deprecation *only when* the corresponding relationships are not also configured for polymorphism correctly. By this means users are free to continue using mixins and inheritance if they so choose.
 
 ## How we teach this
 
@@ -98,7 +161,7 @@ JSON:API has enough information (`type` and `id`) to set this relation.
 re-organized or altered? Does it change how Ember is taught to new users
 at any level?
 
-The docs should be changed to promote the new way.
+Polymorphism was never explicitly documented or supported by EmberData. With this RFC, we would add an official declaration of support and add a section on polymorphism to both the docs for model, docs for the schema service and to the guides.
 
 ## Drawbacks
 
@@ -106,10 +169,12 @@ None
 
 ## Alternatives
 
-> What other designs have been considered? What is the impact of not doing this?
+We could require a more explicit API that declares the full-intent and full-resolvability on one or even both sides. The advantage of both sides is both less work to do on the part of the ember-data to determine validity and increased clarity to the reader in that both sides show the config.
+
+However, This can have some drawbacks in that as an API grows to have more entities that satisfy a polymorphic relationship the config becomes increasingly and unnecessarily large and prevents runtime additions to the schema.
 
 > This section could also include prior art, that is, how other frameworks in the same domain have solved this problem.
 
 ## Questions
 
-It might be considered to use `as` instead of `inverse` to make it stand out as polymorphic relation, but if it works without it, I think it's better to keep `inverse`.
+None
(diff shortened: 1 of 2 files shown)


text/0793-polymporphic-relations-without-inheritance.md
```diff
@@ -0,0 +1,135 @@
...
+
+# Polymorphic relations without inheritance
+
+## Summary
+
+Allow setting polymorph relations without inheritance and mixins.
+
+## Motivation
+
+In pre-octane Ember you could set polymorphic relations using two ways: inheritance or mixins. Mixins are deprecated and impossible to use with native classes, so the only option is to use inheritance:
```
runspired, line 31: Mixins aren't officially deprecated so lets not say this. They can be used with native classes via either a decorator approach or a double-extend ala 

```ts
export default class Post extends Model.extend(TaggableMixin) {

}
```

They are discouraged and we do want to simplify the programming model though. The motivation here is that polymorphic relationships should be possible without mixins or inheritance.

text/0793-polymporphic-relations-without-inheritance.md
```diff
@@ -0,0 +1,135 @@
...
+class TaggableModel extends Model {
+  @hasMany('tag', { inverse: 'taggable' }) tags;
+}
+
+// real models with tags
+class CommentModel extends TaggableModel {}
+class PostModel extends TaggableModel {}
+```
+
+Which fells apart when you need your model to be a target of several polymorphic relations.
```
runspired, line 50: Will be more clear if we explicitly extend the same setup as above.

> Now consider that this same post and comment class also both need to be "viewable", and that "viewable" is a trait also shared by other models that are not "taggable". Today this is only achievable via mixins.

Then show the mixin based setup.

text/0793-polymporphic-relations-without-inheritance.md
```diff
@@ -0,0 +1,135 @@
...
+}
+
+// How to make Comment and Post viewable?...
+```
+
+You can do some dirty tricks making it inherit from each other  `ViewableModel extends TaggableModel`. Or even by adding new base model class, like this:
+
+```js
+
+import Model, { belongsTo, hasMany } from '@ember-data/model';
```
runspired, line 70: I don't think we need this section to get the point across

text/0793-polymporphic-relations-without-inheritance.md
```diff
@@ -0,0 +1,135 @@
...
+  @hasMany('tag', { inverse: 'taggable' }) tags;
+}
+
+```
+
+While it solves the problem, it seems like a misuse of inheritance and too much trickery involved.
+
+## Detailed design
+
+I propose to introduce new API, similar to what [rails have](https://guides.rubyonrails.org/association_basics.html#polymorphic-associations).
```
runspired, line 93:
```suggestion
This RFC proposes to allow explicitly declaring polymorphic traits similar to [rails conventions](https://guides.rubyonrails.org/association_basics.html#polymorphic-associations).
```

runspired (commented):
Overall this is headed in the right direction and mirrors what we've been discussing as the path forward for some time (simply that if the relationship definition on one model satisfies the definition on the inverse that declares itself polymorphic then that is sufficient to allow the record to satisfy the relationship). I'll leave more detailed feedback in a comment.

runspired:
Some prior discussion and art:

- https://github.com/emberjs/data/pull/7491
- https://github.com/emberjs/data/pull/7421

Overall this is the direction the team has wanted to take this for sometime. I think there are a few key discussion points for us all to hash out.

1) would a much more explicit API be better here?  This could include declaring intent on one or even both sides. The advantage of both sides is both less work to do on the part of the ember-data to determine validity and increased clarity to the reader in that both sides show the config. Example below with a Many:Many relationship (tags can belong to many taggables, taggables can have many tags). This can have some drawbacks in that as an API grows to have more "taggable" entities (per the example below) the config may becomes increasingly large unnecessarily. It is likely enough to the right-hand-side (Comment.tags) to just say that it satisfies the left (Tag.taggables), the left likely does not need to declare what things do satisfies it, which keeps the API flexible and allows for on-demand runtime additions to the schema.

```ts
// note how we explicitly declare what types can satisfy this relationship in "allowedTypes"
class Tag extends Model {
  @hasMany("taggable", { inverse: "tags", polymorphic: true, allowedTypes: ["post", "comment"] }) taggables;
}

// note how we explicitly set the "base type" the relationship is satisfying here
// this side is not polymorphic as only a Tag satisfies the relationship.
class Comment extends Model {
  @hasMany("tag", { inverse: "taggables", as: "taggable" }) tags;
}
```

2) how does ember-data treat `taggable` if there is no model for the type (and thus no schema for it). Do we require that a user specify something via the schema service even if using @ember-data/model? If using @ember-data/model, do we require that a base model for that "trait" always exist just to hold that schema? Do we assume a missing inverse schema definition for a polymorphic relationship will just be created on-demand later and allow nothing to be specified?

text/0793-polymporphic-relations-without-inheritance.md
```diff
@@ -0,0 +1,135 @@
...
+Release Versions: Leave as is
+Relevant Team(s): Ember Data
+RFC PR: https://github.com/emberjs/rfcs/pull/793
+-->
+
+# Polymorphic relations without inheritance
+
+## Summary
+
+Allow setting polymorph relations without inheritance and mixins.
```
runspired, line 27: We should also explicitly deprecate Mixin and Inheritance based polymorphism as part of this PR. That deprecation would target 5.0 and would only be printed if the relationship definition was not sufficient for determining polymorphism without making use of checks of the prototype chain. E.g. users can continue to use mixins and inheritance to sprinkle in polymorphism if they so desire; however, the mechanism will be simply the actual definition of the relationship and anywhere that fails to be sufficient will print a deprecation asking the user to update the relationship definition to one that is sufficient.

yratanov:
Thank you @runspired! I've applied your feedback.

--------------- 

> would a much more explicit API be better here? This could include declaring intent on one or even both sides. The advantage of both sides is both less work to do on the part of the ember-data to determine validity and increased clarity to the reader in that both sides show the config. Example below with a Many:Many relationship (tags can belong to many taggables, taggables can have many tags). This can have some drawbacks in that as an API grows to have more "taggable" entities (per the example below) the config may becomes increasingly large unnecessarily. It is likely enough to the right-hand-side (Comment.tags) to just say that it satisfies the left (Tag.taggables), the left likely does not need to declare what things do satisfies it, which keeps the API flexible and allows for on-demand runtime additions to the schema.

I really don't like having `allowedTypes`. Model with polymorphic relation shouldn't care what exact type is passed to it imo.

> how does ember-data treat taggable if there is no model for the type (and thus no schema for it). Do we require that a user specify something via the schema service even if using @ember-data/model? If using @ember-data/model, do we require that a base model for that "trait" always exist just to hold that schema? Do we assume a missing inverse schema definition for a polymorphic relationship will just be created on-demand later and allow nothing to be specified?

Sorry, I'm not familiar with Ember Data internals, I'd assume that if `inverse` key is set, then we have to require it to be defined in the taggable model.

```js
class Tag extends Model {
  @hasMany("taggable", { inverse: "tags", polymorphic: true }) taggables;
}

class Comment extends Model {
}

tag.taggable = comment; // => Error: Comment model doesn't have "tags"

// ======

class Tag extends Model {
  @hasMany("taggable", { polymorphic: true }) taggables;
}

tag.taggable = comment; // => All good
```

text/0793-polymporphic-relations-without-inheritance.md
```diff
@@ -0,0 +1,115 @@
...
+
+The docs should be changed to promote the new way.
+
+## Drawbacks
+
+None
+
+## Alternatives
+
+> What other designs have been considered? What is the impact of not doing this?
```
runspired, line 109:
```suggestion
We could require a more explicit API that declares the full-intent and full-resolvability on one or even both sides. The advantage of both sides is both less work to do on the part of the ember-data to determine validity and increased clarity to the reader in that both sides show the config.

However, This can have some drawbacks in that as an API grows to have more entities that satisfy a polymorphic relationship the config becomes increasingly and unnecessarily large and prevents runtime additions to the schema.
```

text/0793-polymporphic-relations-without-inheritance.md
```diff
@@ -0,0 +1,115 @@
...
+
+> This section could also include prior art, that is, how other frameworks in the same domain have solved this problem.
+
+## Questions
+
+It might be considered to use `as` instead of `inverse` to make it stand out as polymorphic relation, but if it works without it, I think it's better to keep `inverse`.
```
runspired, line 115:
```suggestion
None
```

text/0793-polymporphic-relations-without-inheritance.md
```diff
@@ -0,0 +1,115 @@
...
+
+## Detailed design
+
+This RFC proposes to allow explicitly declaring polymorphic traits similar to [rails conventions](https://guides.rubyonrails.org/association_basics.html#polymorphic-associations).
+
+```js
+import Model, { belongsTo, hasMany } from '@ember-data/model';
+
+class TagModel extends Model {
+  @belongsTo('taggable', { polymorphic: true }) taggable;
```
runspired, line 79:
```suggestion
  @belongsTo('taggable', { polymorphic: true, inverse: 'tag' }) taggable;
```

text/0793-polymporphic-relations-without-inheritance.md
```diff
@@ -0,0 +1,115 @@
...
+}
+
+class ViewModel extends Model {
+  @belongsTo('viewable', { polymorphic: true }) viewable;
```
runspired, line 83:
```suggestion
  @belongsTo('viewable', { polymorphic: true, inverse: 'views' }) viewable;
```

text/0793-polymporphic-relations-without-inheritance.md
```diff
@@ -0,0 +1,115 @@
...
+}
+
+class CommentModel extends Model {
+  @hasMany('views', { inverse: 'viewable' }) views;
```
runspired, line 87:
```suggestion
  @hasMany('views', { inverse: 'viewable', as: 'viewable' }) views;
```

text/0793-polymporphic-relations-without-inheritance.md
```diff
@@ -0,0 +1,115 @@
...
+  @hasMany('tag', { inverse: 'taggable' }) tags;
```
runspired, line 88:
```suggestion
  @hasMany('tag', { inverse: 'taggable', as: 'taggable' }) tags;
```

text/0793-polymporphic-relations-without-inheritance.md
```diff
@@ -0,0 +1,115 @@
...
+  @hasMany('views', { inverse: 'viewable' }) views;
+}
+
+```
+
+## Detailed design
+
+This RFC proposes to allow explicitly declaring polymorphic traits similar to [rails conventions](https://guides.rubyonrails.org/association_basics.html#polymorphic-associations).
+
```
runspired, line 74:
```suggestion

### API

The inverse relationship, if one exists, of a polymorphic relationship must be concrete (e.g. not also polymorphic).

The polymorphic relationship **MUST** explicitly declare itself as polymorphic and **MUST** explicitly declare it's inverse as either a key or `null`. All records that satisfy the polymorphic type must declare a matching inverse via that key.

Polymorphic relationships take the form:

\```ts
hasMany(baseType: string, options: { polymorphic: true, inverse: string | null }) 
belongsTo(baseType: string, options: { polymorphic: true, inverse: string | null }) 
\```

Concrete Relationships which fulfill polymorphic relationships take the form:

\```ts
hasMany(recordWithPolymorphicRelationship: string, options: { inverse: string, as: string }) 
belongsTo(recordWithPolymorphicRelationship: string, options: { inverse: string, as: string }) 
\```

So for instance, given a field named "tagged" that is a polymorphic hasMany with baseType "taggable", and inverse key "tags" the following would be the polymorphic relationship and concrete relationship.

\```ts
// polymorphic relationship
class Tag extends Model {
   @hasMany("taggable", { polymorphic: true, inverse: "tags" }) tagged;
}
// an inverse concrete relationship
class Post extends Model {
   @hasMany("tag", { inverse: "tagged", as: "taggable" }) tags;
}
\```

### Example

```

runspired:
@yratanov I updated the RFC to fill out bits of context and design it needed as well as to update a few bits of the proposed design. To answer your above questions:

> I really don't like having allowedTypes. Model with polymorphic relation shouldn't care what exact type is passed to it imo.

I think everyone is in agreement, but it's an alternative to consider. I've noted this in the alternatives.

> Sorry, I'm not familiar with Ember Data internals, I'd assume that if inverse key is set, then we have to require it to be defined in the taggable model.

This wasn't what I meant. Let's say that you have a polymorphic relationship to an abstract entity (taggable), do we require that `taggable` ever exist on its own? or if Comment and Post satisfy Taggable can they exist without there being an explicit Taggable entity? I've answered this in the RFC updates but roughly it comes down to "you only need to implement Taggable if your API is going to return things with the type `'taggable'`"

> Do we need `as` if we have `inverse` and/or should we rename `inverse` to `as`.

We need both. `inverse` specifies the key, `as` specifies the abstract type.

PR #795, opened 2022-02-14
ember-data: Return a Promise from Model.save()

snewcomer:
Partial dup of https://github.com/emberjs/rfcs/pull/362#issuecomment-553646135 with improvements.  Specifically the addition of derived state to the promise wrapper.

[rendered](https://github.com/snewcomer/rfcs/blob/sn/model-save/text/0795-ember-data-return-promise-save.md)

### DIFF
--- a/text/0795-ember-data-return-promise-save.md
+++ b/text/0795-ember-data-return-promise-save.md
@@ -1,0 +1,70 @@
+---
+Stage: Proposed
+Start Date: 2022-02-13
+Release Date: Unreleased
+Release Versions:
+  ember-source: vX.Y.Z
+  ember-data: vX.Y.Z
+Relevant Team(s): ember-data 
+RFC PR: https://github.com/emberjs/rfcs/pull/795
+---
+
+# Return a Promise from Model.save()
+
+## Summary
+
+Model.save() will return an RSVP.Promise wrapper instead of a PromiseObject.
+
+## Motivation
+
+* The API documentation already documents the return value as a Promise.
+* Remove dependency on promise proxies
+* Async Consistency - The PromiseObject encourages a usage pattern of sometimes-async, sometimes-sync behavior. We want to [reduce zalgo](http://blog.izs.me/post/59142742143/designing-apis-for-asynchrony).
+* API consistency - While the PromiseObject proxies properties, the PromiseObject does not proxy methods on the underlying model. For instance, if you try to call a method like destroyRecord on the model, you will get a “not a function” error because the method is called on the Proxy and not the underlying model. This encourages users for Ember Data to reach into private implementation details of the promise proxy, such as using proxy.get('content') to call methods.
+* Enable new functionality in Ember Data while making backwards compatibility with older versions of Ember easier.
+* The documentation of using PromiseProxy APIs has been replaced over time in the guides and API documentation to use more typical JavaScript usage of Promises, such as async/await, instead of PromiseProxies.
+
+## Detailed design
+
+Introduce a new feature flag, `DS_MODEL_SAVE_PROMISE`, Model.save() will return an an RSVP.Promise wrapper that resolves with the model if the save was successful, or rejects with an error if the save fails. Moreover, `isPending`, `isResolved`, and `isRejected` will be exposed as derived state on the promise.  When disabled, a PromiseObject will be returned to keep today's behavior, but accessing properties on the PromiseObject or calling non-Promise functions (.then, .finally, .catch) on the PromiseObject will issue a deprecation warning.
+
+Deprecation Plan:
+
+The deprecations can be implemented by adding deprecations around the existing PromiseObject. However, we can't include the deprecations in the actual PromiseObject class because [it is the base class for the relationship proxies](https://github.com/emberjs/data/blob/master/addon/-legacy-private/system/promise-proxies.js#L85). Instead, we can create and use a new DeprecatedPromiseObject for the DS.Model.save() function. It functions like a PromiseObject but deprecates the following behavior:
+
+    * Accessing properties on the PromiseObject returned from DS.Model.save()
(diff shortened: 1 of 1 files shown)


sandstrom:
Would it make more sense to return a vanilla Promise? 

Most of RSVP is already available in native Promise. 

(I'm guessing RSVP will eventually be deprecated too)

snewcomer:
@sandstrom I like your suggestion. Might be something we want to RFC to get off of RSVP now that we are past IE11.  I'll talk to Chris tomorrow!

snewcomer:
There are a few hairy parts where we were relying on private RSVP API's to guard against dangling promises.  Specifically [here](https://github.com/emberjs/data/blob/master/packages/store/addon/-private/system/store/finders.js#L6)...So alongside this PR, we are going to introduce a deprecation for RSVP.

text/0795-ember-data-return-promise-save.md
```diff
@@ -0,0 +1,70 @@
...
+
+We do not believe this requires any update to the Ember curriculum. API documentation does not need to be updated as it is already documented as returning a promise. This behavior of accessing properties through the PromiseProxyMixin specifically for DS.Model.save() is not documented in the guides.
+
+## Drawbacks
+
+For users relying on this behavior, they may have to refactor their code to access model properties and methods, especially if they were reaching for private APIs.
+
+## Alternatives
+
+The impact of not doing this prevents further changes in Ember Data.
```
runspired, line 68: alternatives also include:

- returning just a promise
- returning a promise but also exposing a template/js helper in a new package that wraps any promise exposing these flags providing broader utility without imposing this promise-state management on ember-data itself.

runspired (approved):

runspired:
Added final comment period, planning on merging monday considering we previously moved this same RFC to FCP in #362

jelhan:
> That said I would like to revisit the decision to add the state-flags. I think there's a strong case for it being a separate util, even if an official one.

I agree that an util function would be way better - especially for teaching. But I don't see how such an util function could be implemented. As far as I know there is no way to [synchronously determine a JavaScript Promise's state](https://stackoverflow.com/questions/30564053/how-can-i-synchronously-determine-a-javascript-promises-state).

snewcomer:
@jelhan There has been a few approaches I have taken (the first being used with promise many arrays).

https://github.com/emberjs/data/pull/7868/files#diff-8aca4274066862134d43e99891338fd7bbd479fbb186cf0ab2f721fc55436a37

https://github.com/snewcomer/ember-stateful-promise/blob/main/addon/utils/stateful-promise.js

I'm unsure of what these two examples don't cover or if you were thinking about something else.  One downside to the util approach is for users to fish though their code to replace.  However, in conjunction with https://github.com/emberjs/rfcs/pull/796, I could see us not concerning ourselves with the derived state.

jelhan:
> @jelhan There has been a few approaches I have taken (the first being pulled from promise many array implementation).
> 
> https://github.com/emberjs/data/pull/7868/files#diff-8aca4274066862134d43e99891338fd7bbd479fbb186cf0ab2f721fc55436a37
> 
> https://github.com/snewcomer/ember-stateful-promise/blob/main/addon/utils/stateful-promise.js
> 
> I'm unsure of what these two examples don't cover or if you were thinking about something else. One downside to the util approach is for users to fish though their code to replace. However, in conjunction with #796, I could see us not concerning ourselves with the derived state.

Thanks a lot for sharing this examples. I used the `ember-stateful-promise` library as an example to show the issue: https://github.com/snewcomer/ember-stateful-promise/pull/30

The state of a Promise, which has been resolved _before_ constructing a `StatefulPromise`, is reported wrongly as `isRunning` until next tick. As far as I know this limitation is by design and can not be fixed until ECMAScript provides an API to synchronously determine the state of a Promise.

I think that's the reason why @runspired proposes to have a `WeakMap` to provide a stable return for the same promise. That would resolve the issue of having two _different_ states reported for the same promise - depending if the stateful promise instance state was created before or after the Promise settled. But the state may still be wrong for one tick.

I fear that this API limitation would be very confusing to consumers. Let me illustrate using one _very_ confusing edge case based on the API proposed by @runspired in https://github.com/emberjs/rfcs/pull/795#issuecomment-1046318284:

```js
const promise = post.save();
await promise;

const state = promiseState(promise);
console.log('isPending', state.isPending); // true
console.log('isResolved', state.isResolved); // false

// do something async, which is totally unrelated,
// but causes later code to be executed in the next tick
await Promise.resolve();

console.log('isPending', state.isPending); // false
console.log('isResolved', state.isResolved); // true
```

:exploding_head:

snewcomer:
@jelhan Really nice points.  Just a few notes...

1. The stateful-promise export is a promise, so I would expect it would be isRunning until the next tick.  A promise within a promise of sorts.

2. I can see how the latter example you provided **could** be an issue.  But in practice with async programming, more often than not, wouldn't be.

```
const statefulPromise = promiseState(promise);
await unrelated promise;
const result = await statefulPromise;

result // correct!
```

But I guess it would really trip people up if they used those state flags for non UI purposes.

```
const statefulPromise = promiseState(promise);
await unrelated promise;

if (statefulPromise.isRunning) {
  const result = await statefulPromise;
  result // correct!
}
```

So I agree with your take and runspireds.

runspired:
@jelhan while it is true that potential for it to be wrong by one-tick exists, the WeakMap solution effectively solves that problem for the 99% case.

```ts
const promise = record.save();
const state = promiseState(promise);
// all state flags are correct before the await
await promise;
// all state flags are still correct after the await
// all state 2 flags are immediately correct, because state2 === state
const state2 = promiseState(promise);
```

 For the 1% case where someone fails to call promiseState until after the promise has resolved, there are two outcomes:
 
 1) the incorrect pending state is non-consequential since it gets fixed on the next tick (I suspect the majority of 1% cases would fall into this bucket)
 2) the incorrect pending state matters but is easily addressed by ensuring promiseState is called on the outcome since that was what was desired.

PR #796, opened 2022-02-20
RFC: ember data deprecate RSVP.Promise for native Promises

snewcomer:


text/0796-ember-data-deprecate-rsvp.md
```diff
@@ -0,0 +1,76 @@
...
+
+With the removal of IE11 from our build infrastructure and browser requirements, we can now safely remove `RSVP.Promise`. From the [docs](https://github.com/tildeio/rsvp.js/)
+
+- Specifically, it is a tiny implementation of Promises/A+. It works in node and the browser (IE9+, all the popular evergreen ones).
+
+RSVP was Ember's polyfill since an early v1.0.0 version during a time when native Promises existed in any browser.  In addition, Promises have been supported since Node 0.12.
+
+By removing `RSVP.Promise` in favor of native Promises, we can drop an unnecessary dependency for both client side and server side fetching of data.
+
+According to [bundlephobia](https://bundlephobia.com/package/rsvp@4.8.5), this would allow us to remove a significant chunk of dependency weight.
```
mehulkar, line 28: 10kb last I checked! https://twitter.com/mehulkar/status/1460864208203960325?s=21

runspired (approved):

text/0796-ember-data-deprecate-rsvp.md
```diff
@@ -0,0 +1,76 @@
...
+      return new RSVPPromise((resolve, reject) => {
+        ...
+      });
+    }
+  },
+```
+
+## How we teach this
+
+We do not believe this requires any update to the Ember curriculum. API documentation may be needed to remove traces of `RSVP.Promise`.
```
locks, line 58: This needs to include the information that goes into the deprecation guide.
You will also need to document the optional feature, especially if you want to flip the availability down the line, right?

locks (commented):
Some minor tweaks!

text/0796-ember-data-deprecate-rsvp.md
```diff
@@ -0,0 +1,76 @@
...
```
sukima, line 58: It has been my experience that many feel that RSVP offers two features not available in native Promises: `.hash()` and test waiters support (`await settled()` support).

If this is not that case—

1. Can we add to the "How we teach this" how to support `.hash()` or offer an alternative using native promises.
2. Can we add ti the "How do we teach this" an explanation why native promises are test-waiter compatible and that the integration that RSVP offered is no longer needed.

runspired: @sukima

1) whether we return a native promise or an rsvp promise does not change anyone's ability to use `hash`.
2) good call out, it's unnecessary because the test waiters aren't waiting on RSVP Promises in this case either, they are waiting on various other things, which ember-data already installs waiters for.

sukima: Oh! This RFC is for Ember-data  I thought it was for Ember proper  gotcha. :+1:

PR #804, opened 2022-03-06
Add release versions for RFCs released between 3.25 and 4.1

SergeAstapov:
Recent RFCs didn't have updated stage, release date, release versions.

Added such details for RFCs released between Ember.js versions 3.25 and 4.1

### DIFF
--- a/text/0236-deprecation-ember-string.md
+++ b/text/0236-deprecation-ember-string.md
@@ -1,5 +1,9 @@
 ---
+Stage: Accepted
 Start Date: 2017-07-14
+Release Date: Unreleased
+Release Versions:
+  ember-source: vX.Y.Z
 Relevant Team(s): Ember.js
 RFC PR: https://github.com/emberjs/rfcs/pull/236
 Tracking: https://github.com/emberjs/rfc-tracking/issues/26
--- a/text/0772-deprecate-bower-support.md
+++ b/text/0772-deprecate-bower-support.md
@@ -1,10 +1,9 @@
 ---
-Stage: Accepted
+Stage: Ready for Release
 Start Date: 2021-10-21
 Release Date: Unreleased
 Release Versions:
-  ember-source: vX.Y.Z
-  ember-data: vX.Y.Z
+  ember-cli: vX.Y.Z
 Relevant Team(s): Ember CLI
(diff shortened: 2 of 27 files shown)


text/0236-deprecation-ember-string.md
```diff
@@ -1,5 +1,9 @@
 ---
+Stage: Released
 Start Date: 2017-07-14
+Release Date: 2021-02-08
+Release Versions:
+  ember-source: v3.25.0
```
locks, line 6: This deprecation hasn't been fully released.

text/0772-deprecate-bower-support.md
```diff
@@ -1,10 +1,9 @@
 ---
-Stage: Accepted
+Stage: Ready for Release
 Start Date: 2021-10-21
 Release Date: Unreleased
 Release Versions:
-  ember-source: vX.Y.Z
-  ember-data: vX.Y.Z
+  ember-cli: v4.3.0
```
locks, line 6: ember-cli v4.3 has not been released yet.

SergeAstapov: @locks it was in beta already, so it's in sort of "almost there" state. removed for now

text/0236-deprecation-ember-string.md
SergeAstapov: @locks thanks! Didn't know that, removed!

kategengler:
We haven't fully implemented the RFC Stages RFC, which is part of why none of these are filled out. The new process will include a point for the team to agree that it is time to move an RFC to 'Ready for release', so I wouldn't want to make that change here. Moving to `released` will also require verifying the entire RFC is complete.

SergeAstapov:
> We haven't fully implemented the RFC Stages RFC, which is part of why none of these are filled out. The new process will include a point for the team to agree that it is time to move an RFC to 'Ready for release', so I wouldn't want to make that change here. Moving to `released` will also require verifying the entire RFC is complete.

Got it! Thank you for the explanation!

locks:
I've reached out to @SergeAstapov and we'll be coordinating as we implement RFC stages!

wagenet:
I think at least some of these should probably be marked as Recommended assuming they're well documented.

wagenet:
I've merged this with my work in #835. Thanks!

PR #806, opened 2022-03-11
Engine Router Helpers

SergeAstapov:
[Rendered RFC](https://github.com/SergeAstapov/rfcs/blob/engine-router-helpers/text/0806-engine-router-helpers.md)

### DIFF
--- a/text/0806-engine-router-helpers.md
+++ b/text/0806-engine-router-helpers.md
@@ -1,12 +1,12 @@
 ---
 Stage: Accepted
-Start Date: 
+Start Date: 2022-03-11
 Release Date: Unreleased
 Release Versions:
   ember-source: vX.Y.Z
   ember-engines: vX.Y.Z
 Relevant Team(s): Ember.js
-RFC PR: 
+RFC PR: https://github.com/emberjs/rfcs/pull/806
 ---
 
 <!--- 
--- a/text/0000-engine-router-helpers.md
+++ b/text/0000-engine-router-helpers.md
@@ -1,0 +1,273 @@
+---
+Stage: Accepted
+Start Date: 
+Release Date: Unreleased
+Release Versions:
+  ember-source: vX.Y.Z
+  ember-engines: vX.Y.Z
+Relevant Team(s): Ember.js
+RFC PR: 
+---
+
+<!--- 
+Directions for above: 
+
+Stage: Leave as is
+Start Date: Fill in with today's date, YYYY-MM-DD
+Release Date: Leave as is
+Release Versions: Leave as is
+Relevant Team(s): Fill this in with the [team(s)](README.md#relevant-teams) to which this RFC applies
+RFC PR: Fill this in with the URL for the Proposal RFC PR
+-->
+
+# Engine Router Helpers
+
+## Summary
+
+This RFC intends to:
+
+- deprecate `<LinkToExternal />` component provided by `ember-engines` package.
+- introduce a new template helper `{{path-for}}` provided by new `ember-engines-router-helpers` package.
+
+## Motivation
+
+`<LinkToExternal />` is the primary way to link to a path that isn't below the mount point of the Ember Engine.
+
+`<LinkToExternal />` is a thin wrapper around `<LinkTo />` component and this worked well in Ember.js pre v4.
+Extension of `<LinkTo />` component has been deprecated in 3.27.0 version of Ember.js per [RFC #0671](./0671-modernize-built-in-components-1.md) and removed in 4.0.0 hence we need a new approach.
+
+There is another problem this RFC seeks to solve.
+Large scale application authors often create an addon(s) to contain reusable components  that can be rendered either in the context of the host app or Ember Engine.
+In such scenario an extra guard needs to be introduced as `<LinkToExternal />` component  available only in the context of Ember Engine and does not exist in the context of regular application, so it's common to see such pattern:
+
+```hbs
+{{#if @isEngine}}
+  <LinkToExternal @route="contact">Contact Us</LinkToExternal>
+{{else}}
+  <LinkTo @route="contact">Contact Us</LinkTo>
+{{/if}}
+```
+
+## Detailed design
+
+Changes proposed in scope of this RFC span several packages:
+
+* updates to `ember-engines`
+* updates to `ember-source`
+* introduction of new addon `ember-engines-router-helpers`
+
+### `ember-engines-router-helpers` addon
+
+This RFC proposes to introduce new addon `ember-engines-router-helpers` so the template helper `{{path-for}}` would be available for use within Ember Engine as well as in libraries with components that may be shared between conventional apps and Engines.
+
+Template helper `{{path-for}}` should detect the execution context and return different value if the owner is an instance of the `ApplicationInstance` or `EngineInstance`.
+
+If the owner is an instance of the `ApplicationInstance` class, the helper should return passed in route name "as is" without modification. This approach allows to safely use helper in shared libraries which may be used both in the conventional Ember app or Ember Engine.
+
+If the owner is an instance of the `EngineInstance` class, and it's non routable engine, then return value from `externalRoutes` map.
+
+If the owner is an instance of the `EngineInstance` class, and it's routable engine, then there are two scenarios:
+
+* if provided route exists in `externalRoutes` map, then return value form the map.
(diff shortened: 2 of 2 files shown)


runspired:
Overall I like the direction, I'm not sure an optional `external=true` param for the helper does anything though. I think we either don't add that at all or we add it and enforce it (even though we can easily determine the correct thing to do without it).

I personally lean towards not having it, because developers working on engines should already be aware that in addition to their routes file they have an external routes map. IIRC Collisions are already not allowed between a key in the routes map and a key provided in the external routes hash; however, if that's not enforced then by all means we should enforce the explicit API and use it to handle collisions.

anehx:
I really like where this is going, especially the removal of engines related code from `ember-source`. However, I do have some questions:

1. Wouldn't the removal of `namespaceRoute` in `ember-source` result in **every** `<LinkTo />` used in an engine having to use `{{path-for}}` to work? Even engine internal ones, or am I missing something?
2. Why not include the proposed helper in the `ember-engines` package? Having an `ember-engines-route-helpers` package doesn't really make sense to me if it contains such basic and often used logic.

runspired:
@anehx I think you've got the point but didn't realize it!

> 1. Wouldn't the removal of namespaceRoute in ember-source result in every <LinkTo /> used in an engine having to use {{path-for}} to work? Even engine internal ones, or am I missing something?

Yes! The idea is that in an engine you always use path-for, and this helper is then what does the namespacing. By making ember-source not do the engine prefixing for us we also open the door for other DX patterns besides the one in this RFC. For instance, because LinkTo now takes any fully qualified path, one could decide to use link-to directly without a helper for paths that are always provided by the main app, and you could build a helper just to provide the engine namespace. That would look something like:

```hbs
<LinkTo @route="home">Home</LinkTo>
<LinkTo @route="{{engine-namespace}}.dashboard">Dashboard</LinkTo>
```

This is really the crux of this RFC: make ember-source not care (because there isn't a strong reason for it to care) so that the DX can be nicely explored. The engines team will provide a default helper(s) but more than one valid approach would be possible and could offer a desirable dev experience.

> 2. Why not include the proposed helper in the ember-engines package? Having an ember-engines-route-helpers package doesn't really make sense to me if it contains such basic and often used logic.

Because that would force any addon that needs to make use of the helper to also declare ember-engines as a dependency, which is not required. This proposal solves a common complaint with engines in that components in an addon cannot make use of `LinkTo` in a way that allows them to be shared between apps and engines reliably. `path-for` changes this.

anehx:
@runspired I realized that I missed step 2 in "how we teach this" - it all makes sense now. But to avoid such confusion I'd suggest not limiting the intro sentence of "how we teach this" to `<LinkToExternal />`. Also, the explaination to why we need an addon for this makes perfect sense, thanks!

I really like this approach and hope this will be pushed forward so we can finally have ember 4 support in engines, thanks for your efforts!

anehx:
> Overall I like the direction, I'm not sure an optional `external=true` param for the helper does anything though. I think we either don't add that at all or we add it and enforce it (even though we can easily determine the correct thing to do without it).
> 
> I personally lean towards not having it, because developers working on engines should already be aware that in addition to their routes file they have an external routes map. IIRC Collisions are already not allowed between a key in the routes map and a key provided in the external routes hash; however, if that's not enforced then by all means we should enforce the explicit API and use it to handle collisions.

I'm with you on this. I think having such an option without enforcing it leads to more confusion than really helping the developers and enforcing it would make the most basic of all use-cases tedious to write.

text/0806-engine-router-helpers.md
```diff
@@ -0,0 +1,273 @@
...
+
+* updates to `ember-engines`
+* updates to `ember-source`
+* introduction of new addon `ember-engines-router-helpers`
+
+### `ember-engines-router-helpers` addon
+
+This RFC proposes to introduce new addon `ember-engines-router-helpers` so the template helper `{{path-for}}` would be available for use within Ember Engine as well as in libraries with components that may be shared between conventional apps and Engines.
+
+Template helper `{{path-for}}` should detect the execution context and return different value if the owner is an instance of the `ApplicationInstance` or `EngineInstance`.
```
buschtoens, line 63: While such a template helper makes for convenient usage in templates (and should be kept for that purpose), I think it's crucial that this functionality is also exposed on a service. Otherwise it will be impossible or needlessly complicated to invoke it from JS/TS.

The helper could then just delegate to that service.

text/0806-engine-router-helpers.md
```diff
@@ -0,0 +1,273 @@
...
+
+  This is not viable long term solution as any changes made to `<LinkTo />` component in the Ember.js codebase would need to be backported to `@ember/legacy-built-in-components` which is not planned to be done.
+
+* Deprecate and remove `externalRoutes` mapping.
+
+  This would require providing *full route name* to `<LinkTo />` component *including* engine mount point. This option is not viable as it breaks isolation principle of Ember Engines whereas every engine should be self-scoped micro-frontend.
+
+* Bake in these implementation details into Ember.js codebase.
+
+  Concerns discussed in this RFC are specific to Ember Engines and there is no need to ship this functionality to every user of Ember.js.
```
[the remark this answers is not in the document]
anehx: So basically what `ember-engines-router-service` does plus wrapping `<LinkTo />` to use said service?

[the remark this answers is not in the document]
buschtoens: Somewhat, but not quite. [`ember-engines-router-service`](https://github.com/villander/ember-engines-router-service#readme) does not unify the internal and external route namespaces to drop `*External` methods. Therefore it is not interoperable with the standard `RouterService`, as you still have to invoke e.g. `transitionToExternal(externalRouteName)` in order to navigate to an external route.

The two linked issues contain a discussion regarding the pros and cons of this namespace merger.

If `ember-engines-router-service` were to merge the namespaces, such that you can use `transitionTo(externalRouteName)` to navigate to an external route, then yes: The only remaning change is refactoring / wrapping / polyfilling `<LinkTo>` to only use the `RouterService` public API.

I think that refactoring `<LinkTo>` to only rely on public `RouterService` API is a sensible change either way. And if `ember-engines` ships its own `EngineRouterService` as `service:router`, then there's no need for additional dependencies and no need for third-party authors to worry about engine compatibility, because it will work out of the box.

text/0806-engine-router-helpers.md
runspired: I agree with the sentiment that there should also be a service, I'll need to reflect a little on whether an engine specific version of router vs a service with a different name is better/worse. Thanks for the prior art links!

SergeAstapov: @runspired @buschtoens as a prior art, there is https://github.com/ember-engines/ember-engines/issues/779, followed by PR https://github.com/ember-engines/ember-engines/pull/669.

@villander also released that functionality as separate addon https://github.com/villander/ember-engines-router-service

I do agree that it's essential to have to have such functionality exposed on a service which we can tackle independently from this RFC by pushing https://github.com/ember-engines/ember-engines/issues/779 and https://github.com/ember-engines/ember-engines/pull/669.

SergeAstapov:
> Overall I like the direction, I'm not sure an optional `external=true` param for the helper does anything though. I think we either don't add that at all or we add it and enforce it (even though we can easily determine the correct thing to do without it).
> 
> I personally lean towards not having it, because developers working on engines should already be aware that in addition to their routes file they have an external routes map. IIRC Collisions are already not allowed between a key in the routes map and a key provided in the external routes hash; however, if that's not enforced then by all means we should enforce the explicit API and use it to handle collisions.

Was thinking on this for couple days and I can't find strong argument to enforce optional `external=true` param.
In another part of the RFC we speak about simplifying developing components in shared libraries which to avoid shared lib know if it's internal or external route which contradicts enforce that argument.

And I strongly agree having it *optional* makes things confusing for developers.

I'll update RFC removing `external` param.

runspired:
We discussed this today in our meeting and decided

1.) we will open a new RFC targeting just the changes to `ember-source` required to give us the flexibility to explore solutions
  - LinkTo no longer cares about engines
  - LinkTo only uses public router service APIs
  - Router service has a public import path `import RouterService from "@ember/routing/service";`
 
 2.) We will follow that RFC up with a new RFC for Engines to replace this one that 
   - deprecates `LinkToExternal`
   - proposes some variation of an engines router service implementation and/or helpers
   
  Thank you everyone for the discussion!

PR #808, opened 2022-03-24
Amend RFC 0724 to include template type checking

chriskrycho:
RFC #724 originally intentionally excluded template type checking via Glint from its designation of "official support" for Ember. However, in the time since the RFC was authored, there have been two significant changes:

1.  Glint itself has matured significantly, with no known major issues at this time (though plenty of polish still to be done).

2.  The *major* design issues in Ember needed to unblock Glint have been resolved:

    - resolution of template identifiers (components, helpers, and modifiers): RFC #779
    - a Glimmer Component type signature which exposes information about the blocks and the target for `...attributes`: RFC #748

Although there remain a number of smaller design questions to resolve, this is sufficient for us to treat type-checked templates as a viable part of our story, and given *viability*, we think this is *necessary*.

### DIFF
--- a/text/0724-road-to-typescript.md
+++ b/text/0724-road-to-typescript.md
@@ -122,9 +122,7 @@ Additionally, this RFC does not propose changing the *default* experience from J
 
 #### Typed Templates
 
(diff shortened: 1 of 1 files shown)


chriskrycho:
This is now in Final Comment Period!

chriskrycho:
We talked about this again at today's Framework Team meeting and are merging it! 🎉

Issue #652, opened 2020-08-04
Allow modifiers to interact with the element before it is insert into the DOM


jelhan:
Modifier Manager should provide a hook, which has access to the DOM element but is executed *before* the element is insert into the DOM.

The `installModifier` hook is executed *after* the element is insert into the DOM:

> This hook has the following timing semantics:
> 
> Always
> 
> -    called after all children modifier managers installModifer hook are called
> -    called after DOM insertion
>
> https://github.com/emberjs/rfcs/blob/master/text/0373-Element-Modifier-Managers.md#installmodifier

The `createModifier` hook does not have access to the DOM element. Accordingly to [Modifier Manager RFC](https://github.com/emberjs/rfcs/blob/master/text/0373-Element-Modifier-Managers.md#createmodifier) "called as discovered during DOM construction". Not sure if the element is created yet and just not passed in or does not even exist yet.

I have two use cases for a hook, which has access to the DOM element but is executed *before* it's insert into the DOM:

1. A `set-attribute` and `set-property` modifier, which sets values as attributes or properties on the element before it's insert into the DOM.
2. The existing [`ember-style-modifier`](https://github.com/jelhan/ember-style-modifier), which manipulates the CSS styles of an element using CSSOM.

For both of them running before the element is insert into the DOM is important from a performance perspective. If executed *after* the element has been insert into the DOM the browser may do an unnecessary layout calculation in between. E.g. if `Element.getBoundingClientRect()` or a similar function is called in between.

It's even worse as there isn't any garantuee if the modifier is executed in the same tick as the DOM insertion:

> May or May Not
> 
> -    be called in the same tick as DOM insertion

If it's executed in the next queue the Browser paint the state before the modifiers has been run. This may be visible to the end-user as an inconsistent and flickering UI.

Especially for `set-attribute` and `set-property` the timing is also important if considering [custom elements](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_custom_elements). Custom elements are part of the web components feature set. Similar to `@ember/component` they have lifecycle hooks. One of them is the `connectedCallback`. It's executed when the element is insert into the DOM.

[...]

PR #811, opened 2022-03-29
Element Modifiers

SergeAstapov:
This RFC supersedes the original [RFC #353 "Modifiers"](https://github.com/emberjs/rfcs/pull/353)


text/0811-element-modifiers.md
```diff
@@ -0,0 +1,232 @@
...
+   (Note that here we should *only* teach the [newly-redesigned API][v3.2.0], not
+   the deprecated previous APIs. This will simplify our teaching significantly,
+   since we can describe how it matches the `Helper` APIs.)
+
+## Drawbacks
+
+- If we merge [RFC #757 "Default Modifier Manager][rfc-0757],
+  it may seem redundant with this RFC.
+- There is the potential for confusion between the `ember-modifier` package and
+  the `@ember/modifier` package which ships as part of `ember-source`.
```
NullVoxPopuli, line 212: what if ember-modifier is re-exported from `@ember/modifier`? adding `Modifier` and `modifier` to the exports from `@ember/modifier`?

SergeAstapov: @NullVoxPopuli this is meant by `Integrate `ember-modifier` into the `@ember/modifier` package directly.` in "Alternatives".
I can expand that to be more specific

NullVoxPopuli: yeah, like, I think there are some pretty good ergonomics to that alternative, and I'm curious what others think.

Calling that is good, I totally forgot about the similar import paths :sweat_smile: 


(good RFC, @SergeAstapov !!! <3)

SergeAstapov: @chriskrycho may know more about the potential downsides of this approach.

From my perspective, this would be logical to have imports like

```js
import Modifier, { modifier } from '@ember/modifier';
```

as it would follow what we do with helpers, where we have imports like

```js
import Helper, { helper } from '@ember/component/helper';
```

NullVoxPopuli: tangent, but I think the `/component/` part of the helper imports is weird :sweat_smile:

chriskrycho: This is one of the questions I intend to talk through at the Framework core team meeting on Friday! One of the reasons is that there is open design question going forward: should this kind of thing be in `@ember/*` or `@glimmer/*`?

NullVoxPopuli: oo, good point. probably makes more sense for `@glimmer/*`, tbh -- even though ember feels like the place to be _only_ because it feels more maintained. :see_no_evil:

text/0811-element-modifiers.md
```diff
@@ -0,0 +1,232 @@
...
+RFC PR: Fill this in with the URL for the Proposal RFC PR
+-->
+
+# Element Modifiers
+
+## Summary
+
+This RFC introduces the concept of user defined element modifiers and proposes
+adding [ember-modifier](https://github.com/ember-modifier/ember-modifier)
+to the blueprints that back `ember new` and `ember addon`, providing an
```
bertdeblock, line 29: > and `ember addon`

It's not super clear if this means that `ember-modifier` will become a direct dependency of a new addon or if `ember-modifier` will become a dev dependency of a new addon's dummy app. I'm assuming it's the latter?

text/0811-element-modifiers.md
```diff
@@ -0,0 +1,232 @@
...
+
+This RFC supersedes the original [RFC #353 "Modifiers"][rfc-0353].
+
+## Motivation
+
+Ember Octane introduced Glimmer Components as a replacement for Classic
+Components. They are simpler, more ergonomic, and more declarative. In contrast
+with Classic Components, Glimmer Components don't have any element/DOM based
+properties or hooks giving access to DOM. This was an intentional move as
+component baking class gets disconnected from DOM manipulation.
```
bertdeblock, line 40:
```suggestion
component backing class gets disconnected from DOM manipulation.
```

chriskrycho:
**Status:** this was on the agenda for this week’s Framework Core Team meeting, but we had a couple of longer discussions about other open RFCs before we got to it. I expect we’ll be able to cover it next week.

rwjblue (commented):
I personally would love to see this come from `@glimmer/modifier`, what do folks think?

text/0811-element-modifiers.md
```diff
@@ -0,0 +1,232 @@
...
+## Drawbacks
+
+- If we merge [RFC #757 "Default Modifier Manager][rfc-0757],
+  it may seem redundant with this RFC.
+- There is the potential for confusion between the `ember-modifier` package and
+  the `@ember/modifier` package which ships as part of `ember-source`.
+
+## Alternatives
+
+- Introduce [RFC #757 "Default Modifier Manager][rfc-0757]
```
Windvis, line 216: I think this is the direction we should go in. [ember-modifier's function-based modifiers](https://github.com/ember-modifier/ember-modifier#function-based-modifiers) feel like a stop-gap solution for the default modifier manager and it would be better to simply focus on that RFC instead. It would also be consistent with the already merged ["Default helper manager" RFC](https://github.com/emberjs/rfcs/pull/756) and is one less thing we would have to import.

I realize adding the `modifier` function would be the modifier equivalent of the already existing [`helper` function](https://api.emberjs.com/ember/4.3/functions/@ember%2Fcomponent%2Fhelper/helper), but I think that should be deprecated once the default helper manager is shipped :smile:.

The scope of this RFC could then be reduced to only the class based modifier API of ember-modifiers which ideally, as mentioned in the RFC and by others, would be exported from `@glimmer/modifier` or `@ember/modifier`.

Thanks for writing this RFC :heart:!

chriskrycho:
Some notes on last week’s Framework team meeting and one-off discussions—

1. We should tweak this to highlight the design motivation _as such_. This will be a couple small tweaks to the text, which I’ll suggest in a review today or tomorrow. Compliments to @SergeAstapov, though, who did a great job spelling out the motivation in those terms *already*: it means the tweaks here are small and minimal. This distinction is important because we are *not* establishing a norm of just adding popular packages to the default blueprint—we are adding this one specifically because it addresses a fundamental gap in the story for authoring Octane apps, and because it already has a well-rationalized API matching the framework.

2. We do want to go ahead and just add `ember-modifier` to the default blueprint. ***This is explicitly and intentionally adding some incoherence to Ember as we begin pushing toward Polaris, where we expect to resolve that incoherence.*** Specifically, we will now have imports for both `@ember/modifier` and `ember-modifier` available, and folks will encounter that regularly as they start adopting strict mode via First-Class Component Templates and need to `import { on } from '@ember/modifier';`. If they happen to also *define* a modifier locally, that can look a bit strange (I actually call this out explicitly in my EmberConf talk, as you’ll see next week.)

(2) is related why we’re not just taking the path @Windvis suggested (though that is also a space we want to iterate on): we see a path to rationalizing modifiers, effects (of which modifiers are a special case), helpers, resources, etc. into a single unified design. Continuing to ship `ember-modifier` allows us to *separately* ship `@glimmer/modifier` in the future (per @rwjblue’s suggestion) building on that unification, or to adopt the existing `ember-modifier` design there and eventually make a breaking change, or other variations in the space.

The key is that we have a minimal and incremental path forward by adopting `ember-modifier`:

- there is *zero* churn for existing users: they just keep using the same library
- we can move to either of `@ember/modifier` or `@glimmer/modifier` at will in the future, with a full set of options for exactly *what* we move there at that time
- we can also tackle the question of function-based modifiers _a la_ #757 separately, and that can inform what we do or don't add to `@glimmer/modifier` etc.

I'll review shortly with the recommended tweaks, and we’ll hopefully talk about it again at this week’s Framework meeting!

text/0811-element-modifiers.md
chriskrycho: @sergeastapov as @bertdeblock notes, we should clarify this. I believe it will be added to `devDependencies` of apps and `dependencies` of addons, per the usual (same as e.g. `@glimmer/component`.

text/0811-element-modifiers.md
```diff
@@ -0,0 +1,232 @@
...
+## Detailed design
+
+The necessary changes to `ember-cli` are relatively small since we only need
+to add the dependency to the `app` blueprint, and the `addon` blueprint will
+inherit it automatically.
+
+This has the advantage (over including it as an implicit dependency), that
+apps and addons that don't want to use it for some reason can opt out by
+removing the dependency from their `package.json` file.
+
```
chriskrycho, line 189: We should add a comment here explicitly calling out that "add an ecosystem addon to the blueprint" is not our normal operating model. Instead, it is a case where we have carefully rationalized that design *with an eye to doing this*, and we *know* it is introducing additional "incoherence" to the programming model.

```suggestion

**Notes:**

1.  This is *not* the usual path for delivering features into Ember: we do not
    generally introduce community addons directly into the blueprint. In this
    specific case, we think it is the right move:
        
        -   The addon has intentionally been reworked to rationalize it in
            terms of the rest of the Octane programming model, explicitly with an eye to adoption in this way.

        -   We do *not* want to introduce `@glimmer/modifier` with exactly this
            API, at least not yet, because we believe we may want to introduce
            a slightly updated API from that package *without* requiring
            breaking changes in the future.

2.  We are intentionally introducing more "incoherence" to the programming
    model with this addition. (Or rather, we are acknowledging the *existing*
    incoherence in the ecosystem, with an eye to making progress on resolving
    it!) In particular, as the community begins adopting strict mode templates
    via First-Class Component Templates in the months ahead, they will often
    end up importing from both `ember-modifier` and `@ember/modifier`:

    ```js
    import { on } from '@ember/modifier';
    import { modifier } from 'ember-modifier';

    const playWhen = modifier((el, [shouldPlay]) => {
      if (shouldPlay) {
        el.play();
      } else {
        el.pause();
      }
    });

    <template>
      <audio
        src={{@src}}
        {{on "error" @onBadLoad}}
        {{playWhen @shouldPlay}}
      />
    </template>
    ```

    As noted in (1) above, we have ideas on how to resolve this, but are
    intentionally not blocking on those in favor of unlocking this key
    functionality for Ember users in the Octane programming model *today*.

```

chriskrycho (changes requested):
One comment about the suggestions I'm making here:

**This is explicitly *not* setting a precedent for adding *other* Ember ecosystem addons to the default blueprint, and should not be taken to invite RFCs to that end.** The Framework Core team discussed the need for this primitive and the best path forward and *then* asked @SergeAstapov to write this RFC.

text/0811-element-modifiers.md
bertdeblock: Are you sure about adding it to `dependencies` for addons? Most addons don't ship modifiers, so I would prefer to keep it under `devDependencies`. Both `@glimmer/*` packages are also included under `devDependencies`.

chriskrycho: Ah, that's a good point. Should be in `devDependencies` of both.

rwjblue:
> We do want to go ahead and just add ember-modifier to the default blueprint.

Hmm, this seems somewhat strange to me (and I missed the core team meeting referenced). Can you explain why? Is your expectation that **every app** will create at least one modifier of it's own (that's not 100% obvious to me)?

I think I'd vaguely prefer to add a blueprint for `ember g modifier` in `ember-source`, and have that check to ensure that `ember-modifier` is installed properly...

text/0811-element-modifiers.md
SergeAstapov: thanks @bertdeblock and @chriskrycho! This was clarified in the "Motivation" section:

> This RFC seeks to fill this gap in Ember.js' development mental model by providing `ember-modifier` in the blueprint. `ember-modifier` will be added to `devDependencies`, same as e.g. @glimmer/component.

text/0811-element-modifiers.md
SergeAstapov: Thanks @bertdeblock! this was applied

text/0811-element-modifiers.md
SergeAstapov: @chriskrycho thank you! these notes were added to the RFC

chriskrycho:
@rwjblue 

> Hmm, this seems somewhat strange to me (and I missed the core team meeting referenced). Can you explain why? Is your expectation that every app will create at least one modifier of it's own (that's not 100% obvious to me)?

1. I think the answer is *yes*, most apps will want custom modifiers; but that’s an intuition rather than something I know for a fact, of course.
2. How is this different from (a) adding a `@glimmer/modifier` which is in the default blueprint or (b) things like having `@ember/component/helper` in the default?

> I think I'd vaguely prefer to add a blueprint for `ember g modifier` in `ember-source`, and have that check to ensure that `ember-modifier` is installed properly...

Without (necessarily) implying argument, how is that an improvement over just shipping the capability out of the box?

---

Is the concern about adding bloat to the generated app build if it's included by default, or something else? Assuming we ship it as a v2 addon (pulling in the upcoming v4 version of the release, likely this week), which will in turn get resolved by way of `ember-auto-import` (if I recall correctly), doesn't that resolve that concern?

chriskrycho:
Oh: the reason for adding `ember-modifier` rather than `@glimmer/modifier` was twofold:

1. @wycats noted that the work he has been doing with Starbeam indicates we may want a somewhat further-refined version of the modifier API in the future, and that he would prefer to avoid having to deal with a migration story of different versions of `@glimmer/modifier` vs. being able to tell people to migrate *from* `ember-modifier` *to* `@glimmer/modifier` piecemeal over time.

2. @kategengler noted that, if we want to avoid *that* churn, and we *also* don’t want to put it in `@ember/modifier` (which no one really wanted), then the actual *minimal* amount of churn here is simply to add `ember-modifier` as is.

chriskrycho:
A data point on the expected usage level: `ember-modifier` has a similar number of downloads/month as `@glimmer/tracking`. This would not by itself be a signal to make it official, but it *is* a good data point for whether "most" apps will use it.

rwjblue:
> How is this different from (a) adding a @glimmer/modifier which is in the default blueprint 

I agree, I don't think it is fundamentally different (other than the current design doesn't really support Glimmer.js or GlimmerX users, for _roughly_ no good reason).

> How is this different from (b) things like having @ember/component/helper in the default?

We don't have these in the default blueprint though (because they are not real packages).

---

After thinking about this a bit more, I think my main issue is that we aren't being prescriptive enough in this RFC. Some questions:

* Do we add this to the addon blueprint? If so, is it a devDep? Is it a prod dependency? Is it a devDep + peer dependency? Either way, the RFC needs to explicitly call out what we want. My personal preference would be to **not** add it to the addon blueprint (because I doubt "most addons will use modifiers"), but to ensure that `ember g modifier` in an addon works properly (where IMHO "properly" means devDep + peerDep).
* IMHO, if we are going to go the way this RFC is calling out (adding `ember-modifier` to the blueprint) then ember-modifier needs to be transitioned to a V2 addon before this RFC is implemented. If we don't, then we are making https://github.com/ember-cli/ember-cli/issues/9827 worse **and** knowingly adding it to the blueprint of all apps when we have to do a major "soon" is pretty bad...

SergeAstapov:
@rwjblue for the first question, currently RFC says:

> This RFC seeks to fill this gap in Ember.js' development mental model by providing `ember-modifier` in the blueprint. `ember-modifier` will be added to `devDependencies`, same as e.g. @glimmer/component.

we may edit it to make 

> This RFC seeks to fill this gap in Ember.js' development mental model by providing `ember-modifier` in the blueprint. `ember-modifier` will be added to `devDependencies` of both app and addon blueprints, same as e.g. `@glimmer/component`.

As for the second question, `ember-modifier` was already converted to v2 in https://github.com/ember-modifier/ember-modifier/pull/296, so the planned `ember-modifier@4` @chriskrycho mentioned and it will be Embroider native.

chriskrycho:
@rwjblue 

> My personal preference would be to not add it to the addon blueprint (because I doubt "most addons will use modifiers"), but to ensure that ember g modifier in an addon works properly (where IMHO "properly" means devDep + peerDep).

Wouldn’t it make more sense (both from an ease-of-implementation POV and an ease-of-figuring-it-out-to-maintain-it POV) to just leave it as a blueprint supplied by `ember-modifier`? What’s the situation where you would want `ember g modifier` to work in an addon which *doesn’t* have it as a dependency?

rwjblue:
@chriskrycho 

> Wouldn’t it make more sense (both from an ease-of-implementation POV and an ease-of-figuring-it-out-to-maintain-it POV) to just leave it as a blueprint supplied by ember-modifier?

I definitely agree that the simplest implementation is to "just" provide `modifier` blueprint in `ember-modifier` like you said, but it means that folks upgrading ember-source still don't necessarily get any help generating modifiers. Yes, I know this is obvious, but folks commonly update ember-source independently of other things; and if this RFC is merged we are saying that "the default stack includes support for ergonomically generating modifiers" and the fact of _how_ we do that (adding `ember-modifier` to the app/addon blueprints) is just an implementation detail.

>  What’s the situation where you would want ember g modifier to work in an addon which doesn’t have it as a dependency?

I will provide an alternate question / example: `ember g component` works (and generates a `@glimmer/component`) even if `@glimmer/component` is not installed, what is wrong with that setup?

Also, to be clear, I think it would be 100% fine to actually do both:

* have ember-modifier actually own it's blueprint (assuming we codify what it should do RE: peerDependencies like I mentioned in prior comments)
* create an ember-source blueprint for `modifier` that ensures `ember-modifer` is installed and just run it's blueprint. This would still ensure the capability for generating modifiers "out of the box", and allow the ember-modifier package to own it's own implementation details. Seems win/win to me...

chriskrycho:
@rwjblue I think you have an intuition/some experience/a lot of knowledge here that isn't obvious to me reading the conversation! *Why* could we as a community not absorb changes there? I'm asking because, assuming Embroider v2 addon format versions of these packages, it seems to me that the only reason the complication would exist *is* if we don't just add them to the package and make blueprints do special things (otherwise we would just cut a major and people could upgrade piecemeal: it is only Classic builds' Highlander Rules which would break that?)… but I assume that's because I'm missing something that is obvious to you from having worked on this problem space. Can you expand on why this is true?

> our community just can never absorb a major release of those packages (because it would be roughly impossible to even know what dependencies need to be updated/bumped/etc and which are already compatible)

rwjblue:
@chriskrycho 

> Why could we as a community not absorb changes there? I'm asking because, assuming Embroider v2 addon format versions of these packages, it seems to me that the only reason the complication would exist is if we don't just add them to the package and make blueprints do special things (otherwise we would just cut a major and people could upgrade piecemeal: it is only Classic builds' Highlander Rules which would break that?)… but I assume that's because I'm missing something that is obvious to you from having worked on this problem space. Can you expand on why this is true?

The current RFC prose states that `ember-modifier` will be in `devDependencies`, if we go with that (and don't do any of the changes I mentioned in https://github.com/emberjs/rfcs/pull/811#issuecomment-1124013993) addons would start having `addon/modifiers/**` and there would be no dependency relationship (`dependencies`/`peerDependencies`) that the package manager could leverage. We have exactly the same issue (as I mentioned in https://github.com/emberjs/rfcs/pull/811#issuecomment-1124016278) with `@glimmer/component` and `@glimmer/tracking`. Addons sometimes list those as dependencies, some times as only a devDep, and other times as both devDep + peerDep. Leaving this decision to individual addon authors to figure out, is leaving a massive (loaded!) footgun in place. 

We **must** address this with official guidance, and this RFC is the right place to start. The two simple options are: make it a dependency, make it a devDep/peerDep. Both of those have trade offs, I personally think peer dep is the right relationship here, but at the very least this RFC must say _something_ about this.

jelhan:
> We must address this with official guidance, and this RFC is the right place to start. The two simple options are: make it a dependency, make it a devDep/peerDep. Both of those have trade offs, I personally think peer dep is the right relationship here, but at the very least this RFC must say something about this.

I agree that this must be addressed. But I'm not sure if this RFC is the right place to do so. It seems to not only affect newly added `ember-modifier` package. As you mentioned it also affects existing `@glimmer/component ` and `@glimmer/tracking` packages. For me it seems to deserve its own RFC.

chriskrycho:
@jelhan it may indeed need its own RFC, but we need to figure that out *before* introducing this unless we want to possibly cause a high degree of churn across the ecosystem. Note that it is specifically the act of introducing it to the default blueprint, with the need for guidance around what to do, which risks that churn: until we do that, it's in the same boat as any other addon in the ecosystem (and there are no planned API changes for when we bring it in, as it stands!).

jelhan:
> @jelhan it may indeed need its own RFC, but we need to figure that out _before_ introducing this unless we want to possibly cause a high degree of churn across the ecosystem. Note that it is specifically the act of introducing it to the default blueprint, with the need for guidance around what to do, which risks that churn: until we do that, it's in the same boat as any other addon in the ecosystem (and there are no planned API changes for when we bring it in, as it stands!).

I fully agree.

Let me propose the following:

1. Create a RFC how dependencies like these (`@glimmer/component`, `@glimmer/tracking`, `ember-modifier` etc.) should be handled in addons in general.
2. Scope this RFC (and #812) to only add the dependencies to app blueprints for now to unblock it.
3. Have a follow-up RFC to add `ember-modifier` to addon blueprints until the general question is solved.

rwjblue:
> For me it seems to deserve its own RFC.

Totally agree! 

> Let me propose the following:
> 
> * Create a RFC how dependencies like these (@glimmer/component, @glimmer/tracking, ember-modifier etc.) should be handled in addons in general.
> * Scope this RFC (and https://github.com/emberjs/rfcs/pull/812) to only add the dependencies to app blueprints for now to unblock it.
> * Have a follow-up RFC to add ember-modifier to addon blueprints until the general question is solved.

This seems like a reasonable step forward to me, but it's important to me that we don't loose track of actually _doing_ that RFC. Additionally, this RFC should be updated to explicitly state that the addon blueprint will not include `ember-modifier` (I know this is probably what you meant in the second bullet there, but I want to be very explicit there since the addon blueprint "extends" from the app blueprint).

chriskrycho:
@SergeAstapov do you have time to update this RFC this week? I think @rwjblue's concern was the only blocking concern here so we should be able to nominate this for FCP again once that's resolved!

wagenet:
Looks like this is waiting on @SergeAstapov, correct?

chriskrycho:
That's correct!

SergeAstapov:
> @SergeAstapov do you have time to update this RFC this week? I think @rwjblue's concern was the only blocking concern here so we should be able to nominate this for FCP again once that's resolved!

sorry for the delay, added the requested tweak in commit https://github.com/emberjs/rfcs/pull/811/commits/808da756585338f7badceff969490982b621f51f

@chriskrycho @rwjblue please let me know if this does not match exactly what you were thinking/proposing.

chriskrycho (approved):
This seems good to me! I'll see if @rwjblue can make some time to give it a look next week, and I expect we will put it into FCP at next week's Framework Core team meeting. Thank you! 👍🏼

chriskrycho:
We discussed this at Framework today and decided to put it into Final Comment Period, with intent to merge it! Thanks, @SergeAstapov!

jelhan:
> We discussed this at Framework today and decided to put it into Final Comment Period, with intent to merge it! Thanks, @SergeAstapov!

What is the status on this? Label for final comment period has not been added. And it has not been merged.

chriskrycho:
Whoops! We've been talking about a bunch of things and just forgot to circle back to this. I'll do an async check on Monday and it should get merged then!

chriskrycho:
Folks have been busy, but I have merging this at the top of the agenda for this week's Framework team meeting; it should get merged then!

Panman82:
Could RFC #652 impact this RFC?

wagenet:
When we merge, we'll have to manually trigger the advancement PR since this one affects two RFC files.

jelhan:
> ~Could RFC #652 impact this RFC's API design?~ Sorry for the noise, I don't think it is super impactful. Any API additions from #652 would have to be implemented in several areas anyway. zipper_mouth_face

We expect that we will add additional timing capabilities as part of addressing #652. I hope publishing such a RFC in the next week.

The new low-level primitives providing more granular control of modifier timing would need to be exposed in high-level APIs provided by `ember-modifier` _somehow_. But how such a high-level API could looks like, could and should be explored in follow-up PRs. I don't see it as a blocker for this one. Especially as `ember-modifier` package in blueprints is only an interim step towards a `@glimmer/modifier` package.

PR #813, opened 2022-04-16
<template> layering proposal

chancancode:
[Rendered](https://github.com/emberjs/rfcs/blob/template-layering/text/0813-template-layering.md)

### DIFF
--- a/text/0813-template-layering.md
+++ b/text/0813-template-layering.md
@@ -577,29 +577,38 @@ defined as:
       1. The property name must be a valid JavaScript identifier and its
          value must be a reference to the variable with the same name. This
          implies that the JavaScript shorthand property syntax can be used,
-         which is encouraged but optional. Alternatively,
-      2. When the `template()` function is called from within a `static` block
-         inside a class body, the property name can also be a valid JavaScript
-         private field name (i.e. a valid JavaScript identifier prefixed by the
-         `#` character), which must be quoted. Its value must be a function
-         expression that:
-         1. Has exactly one parameter, and
-         2. Returns the current value of the corresponding private field (with
-            the same name as the object key) on the object passed in the
-            parameter.
+         which is encouraged but optional.
 
-4. When, and only when, the `template()` function is called from within a
-   `static` block inside a class body, then a third argument must be supplied
-   to the call, and this argument must be the `this` keyword.
+4. In order to associate a template with a component class:
+
+   1. The `template()` function must be called from within a `static` block on
+      the class, where it must be the only statement in the block, and
+   2. A third argument must be supplied to the call, and it must be the `this`
+      keyword, and
+   3. The specifications for the object literal described in 3.2. is expanded
+      to permit the following:
+      1. The property name can also be a valid JavaScript private field name
+         (i.e. a valid JavaScript identifier prefixed by the `#` character),
+         which must be quoted. Its value must be a function expression that:
+         1. Has exactly one parameter, and
+         2. When called with an object instance of the same class as parameter,
+            it must return the value of the corresponding private field on the
+            object.
 
 5. No additional arguments can be passed.
+
+6. Comments are permitted in any positions allowable by JavaScript, and any
+   conforming tool must be configured to ignore comments when applying these
+   rules. However, there is no guarantee that the comments will be preserved in
+   the output, and even if they are preserved, there are no guarantee around
+   the placements of these comments in the output.
 
 For the tagged template string literal form, idiomatic usages are defined as:
 
 1. The string literal must not contain any interpolations.
 2. For class association, the tagged template string must be placed inside a
-   `static` block within the class body, where it must be the only statement in
-   the block.
+   `static` block on the class, where it must be the only statement in the
+   block.
 
 Some examples of idiomatic usages:
 
@@ -635,6 +644,15 @@ template("<Hello />", function () {
 class Foo {
   static {
     template("<Hello />", () => ({ Hello }), this);
+  }
+}
+
+class Bar {
+  static {
+    // Note: this does meet the requirements for associating the template with
+    // the surrounding `Bar` class (rule 4). However, it is otherwise perfectly
+    // legal to have an *unassociated* template() call inside a static block.
+    this.bar = template("<Hello />");
   }
 }
 ```
@@ -679,20 +697,14 @@ template("<Hello />", () => ({ Hello }), class Foo {});
 
 template("<Hello />", () => ({ Hello }), Foo);
 
-class Bar {
+class Foo {
   static template = template("<Hello />", () => ({ Hello }));
 }
 
-class Baz {
+class Bar {
   static {
     debugger;
     template`<Hello />`;
-  }
-}
-
-class Bat {
-  static {
-    let template = template`<Hello />`;
   }
 }
 ```
--- a/text/0813-template-layering.md
+++ b/text/0813-template-layering.md
@@ -544,44 +544,62 @@ variable, then any indirect invocations are not considered an idiomatic use.
 For the function call form of the `template()` function, idiomatic usages are
 defined as:
 
-1. The first argument must be a single standalone inline string literal, which
-   must be:
-   1. Specified with single quotes, or
-   2. Specified with double quotes, or
-   3. Specified with backticks but without a tag and no interpolations.
-2. The second argument, if provided, must be a single standalone inline
-   anonymous function expression that:
-   1. Specified with either the `function` keyword or arrows, and
-   2. Does not have any modifiers (e.g. `async`, `function*`, etc), and
-   3. Does not take any arguments, and
-   4. Contains exactly the following function body:
-      1. A single return statement, which must return a standalone inline
-         object literal (a "POJO") that:
-         1. Does not contain duplicate property names, and
-         2. Does not contain computed property names, and
-         3. Does not contain any getters, setters or methods, and
-         4. Does not use the spread operator, and
-         5. Contains only name-value paris where the the value is a bare
-            JavaScript identifier (variable reference) and the name matches the
-            identifier exactly (i.e. cases where the shorthand property syntax
-            would have been allowed), and
-         6. Can optionally be specified using shorthand property syntax. Note
-            that because of the previous restriction this is always possible.
-      2. For arrow function expressions, the single expression shorthand is
-         also permitted, provided that the expression is an object literal
-         surrounded in a single pair of parenthesis that matches the above
-         specification.
-3. The third argument, if provided, must be a JavaScript identifier or the
-   `this` keyword.
-4. No more than three arguments can be passed.
-5. Comments are not permitted anywhere inside the arguments list.
+1. Unless otherwise specified:
 
-For the tagged template string form, idiomatic usages are defined as:
+   1. All string literals inside the call (arguments, property keys, etc) must:
+      1. Be quoted with single quotes, or
+      2. Be quoted with double quotes, or
+      3. Be quoted with backticks but without a tag and without interpolations.
+   2. All object literals ("POJOs") inside the call must:
+      1. Not contain duplicate property names, and
+      2. Not contain computed property names, even if the computed name is just
+         a string literal, and
+      3. Not contain any getters, setters or methods, and
+      4. Not use the spread operator.
+   3. All function expressions must:
+      1. Be specified with either the `function` keyword or arrows, and
+      2. Be anonymous, and
+      3. Not have any modifiers (e.g. `async`, `function*`, etc), and
+      4. Have exactly one statement in the function body, which must be a
+         `return` statement. For arrow function expressions, this implies the
+         single-expression implicit return shorthand syntax can be used, which
+         is encouraged but optional.
 
-1. The string must not contain any interpolations.
-2. For class association, the tagged template string must be placed in a static
-   initializer block where it is the only statement. Comments are not permitted
-   inside the block.
+2. The first argument to the call must be a string literal containing the
+   template source.
+
+3. The second argument to the call is optional. If provided, it must be a
+   function expression that:
+
+   1. Has no parameters, and
+   2. Returns an object literal where every name-value pair must be in the form
+      of:
+      1. The property name must be a valid JavaScript identifier and its
+         value must be a reference to the variable with the same name. This
+         implies that the JavaScript shorthand property syntax can be used,
+         which is encouraged but optional. Alternatively,
+      2. When the `template()` function is called from within a `static` block
+         inside a class body, the property name can also be a valid JavaScript
+         private field name (i.e. a valid JavaScript identifier prefixed by the
+         `#` character), which must be quoted. Its value must be a function
+         expression that:
+         1. Has exactly one parameter, and
+         2. Returns the current value of the corresponding private field (with
+            the same name as the object key) on the object passed in the
+            parameter.
+
+4. When, and only when, the `template()` function is called from within a
+   `static` block inside a class body, then a third argument must be supplied
+   to the call, and this argument must be the `this` keyword.
+
+5. No additional arguments can be passed.
+
+For the tagged template string literal form, idiomatic usages are defined as:
+
+1. The string literal must not contain any interpolations.
+2. For class association, the tagged template string must be placed inside a
+   `static` block within the class body, where it must be the only statement in
+   the block.
 
 Some examples of idiomatic usages:
 
@@ -674,10 +692,9 @@ class Baz {
   }
 }
 
-// This is also a syntax error in JavaScript anyway
 class Bat {
   static {
-    return template`<Hello />`;
+    let template = template`<Hello />`;
   }
 }
 ```
--- a/text/0813-template-layering.md
+++ b/text/0813-template-layering.md
@@ -13,9 +13,11 @@ RFC PR: https://github.com/emberjs/rfcs/pull/813
 
 ## Summary
 
-This RFC propose to introduce a `template()` function as an alternative for
-the `<template>` syntax extension defined in [RFC #779][rfc-779] using
-standard JavaScript:
+This RFC propose to introduce a `template()` function as an alternative for the
+`<template>` syntax extension defined in [RFC #779][rfc-779] using standard
+JavaScript:
+
+[rfc-779]: ./0779-first-class-component-templates.md
 
 ```js
 import { template } from "@ember/template-compilation";
@@ -37,17 +39,48 @@ export class Bar {
 }
 ```
 
-[rfc-779]: ./0779-first-class-component-templates.md
+Idiomatic usages of the `template()` function can be pre-compiled by a build
+step if desired, but a runtime implementation can also be used via an opt-in
+described below.
+
+We also propose an alternative form that uses the `template()` function as a
+tag function for JavaScript tagged template strings. However, this variant does
+not have a runtime implementation and must be handled by a build step:
+
+```js
+import { template } from "@ember/template-compilation";
+import Hello from "my-app/components/hello";
+
+// <template><Hello></template> becomes...
+export default template`<Hello />`;
+
+// export const Foo = <template><Hello /></template> becomes...
+export const Foo = template`<Hello />`;
+
+// export class Bar {
+//   <template><Hello /></template>
+// }
+export class Bar {
+  static {
+    template`<Hello />`;
+  }
+}
+```
+
+In addition, we also propose to add a `@ember/template-compilation/runtime`
+module as an opt-in to enable runtime template compilation.
 
 ## Motivation
 
-[RFC #779](rfc-779) introduced a first-class component syntax feature that has
+[RFC #779][rfc-779] introduced a first-class component syntax feature that has
 numerous benefits, among which are:
 
-1. Opt-in to strict mode (RFC #496)
-2. Eliminating the need for runtime name-based resolutions
-3. Ability to use imported components/helpers/modifiers
+1. Opt-in to strict mode ([RFC #496][rfc-496])
+2. Eliminating the need for name-based runtime resolutions
+3. Ability to use imported template constructs
 4. Ability to interact with the surrounding JavaScript scope
+
+[rfc-496]: ./0496-handlebars-strict-mode.md
 
 However, one of the drawbacks is that it requires a custom syntax extension
 (`<template>`) and a custom file format (`.gjs/.gts`), which requires a build
@@ -65,12 +98,12 @@ not acceptable:
 
 2. There may be editors and environments that are impossible for us to support,
    either because they don't have an extension system at all, or those systems
-   do not expose the capabilities we need. For example, CodeSandbox, CodPen,
+   do not expose the capabilities we need. For example, CodeSandbox, CodePen,
    TypeScript playground, etc.
 
 3. When publishing packages to NPM, it is important and desirable to publish
    only standard `.js` files that does not require further processing by the
-   consumer (see: [v2 Addon Format](./0507-embroider-v2-package-format.md)).
+   consumer (see [v2 Addon Format](./0507-embroider-v2-package-format.md)).
 
 4. There are use cases that prefer or require components that are directly
    runnable in the browser without a build step, such as bug report templates,
@@ -174,6 +207,19 @@ extension and the low-level primitives.
   `.gjs` format (such as Discord and GitHub today), but where it's desirable to
   still retain the highlighting for the JavaScript/TypeScript portions.
 
+This RFC also _recommends_ that the addon spec (v2.1?) to be updated to
+consider using `template(hbs-source, pre-compiled-scope-function)` as the
+primary publishing format for component templates and consider reducing and
+deprecating the need for other alternative mechanisms (such as shipping `.hbs`
+files). However, this is only a general recommendation, and a future RFC or
+amendment is needed to explore that further.
+
+It is notable that this update is only needed for the purpose of clarifying
+the recommended/preferred way of shipping templates in addons. However, because
+`template()` is just standard JavaScript, there is nothing stopping addons from
+adopting the format as long as their target Ember versions support the feature
+(or that they depend on a suitable polyfill for the feature).
+
 ## Detailed design
 
 This RFC proposes to re-specify the `<template>` language extension into a
@@ -182,7 +228,7 @@ de-sugaring into `template()` calls:
 1. Top-level declaration
 
    ```
-   import Foo form 'somewhere';
+   import Foo from "somewhere";
 
    <template>
      <Foo />
@@ -192,18 +238,21 @@ de-sugaring into `template()` calls:
    becomes
 
    ```js
-   import { template } from '@ember/template-compilation';
-   import Foo form 'somewhere';
+   import { template } from "@ember/template-compilation";
+   import Foo from "somewhere";
 
-   export default template(`
+   export default template(
+     `
      <Foo />
-   `, () => ({ Foo }));
+   `,
+     () => ({ Foo })
+   );
    ```
 
 2. Expression
 
    ```
-   import Foo form 'somewhere';
+   import Foo from "somewhere";
 
    export const Bar = <template><Foo /><template>;
    ```
@@ -211,8 +260,8 @@ de-sugaring into `template()` calls:
    becomes
 
    ```js
-   import { template } from '@ember/template-compilation';
-   import Foo form 'somewhere';
+   import { template } from "@ember/template-compilation";
+   import Foo from "somewhere";
 
    export const Bar = template(`<Foo />`, () => ({ Foo }));
    ```
@@ -220,7 +269,7 @@ de-sugaring into `template()` calls:
 3. Class
 
    ```
-   import Foo form 'somewhere';
+   import Foo from "somewhere";
 
    export default class Bar {
      <template><Foo /><template>
@@ -230,8 +279,8 @@ de-sugaring into `template()` calls:
    becomes
 
    ```js
-   import { template } from '@ember/template-compilation';
-   import Foo form 'somewhere';
+   import { template } from "@ember/template-compilation";
+   import Foo from "somewhere";
 
    export default class Bar {
      static {
@@ -241,8 +290,13 @@ de-sugaring into `template()` calls:
    ```
 
    **Note**: In static initializer block isn't required here (though it is a
-   standard JavaScript feature). The `template()` function can wrap around the
-   class, or be applied after the class has been declared.
+   standard JavaScript feature). The `template()` function can also be applied
+   after the class has been declared. If available, the static initializer
+   block allows the closest placement to the `<template>` equivalent (nested
+   inside the class body) which is why it is preferred.
+
+   **Note**: When used in this form, the function returns/passes through the
+   target supplied in the third argument (the component class).
 
 In these small snippets, the scope bindings may look very verbose compared to
 the size of the templates, but in real-world templates, the ratio will improve.
@@ -256,10 +310,15 @@ should be able to reason about `<template>` with this desugaring in mind.
 
 ### Tagged Template Literals
 
+Alternatively, the `template()` function can also be used as a tag function for
+a JavaScript tagged template string. This alternate form automatically captures
+the lexical scope variable bindings. However, this form does not have a runtime
+implementation and must be processed by a build step.
+
 1. Top-level declaration
 
    ```
-   import Foo form 'somewhere';
+   import Foo from "somewhere";
 
    <template>
      <Foo />
@@ -269,8 +328,8 @@ should be able to reason about `<template>` with this desugaring in mind.
    becomes
 
    ```js
-   import { template } from '@ember/template-compilation';
-   import Foo form 'somewhere';
+   import { template } from "@ember/template-compilation";
+   import Foo from "somewhere";
 
    export default template`<Foo />`;
    ```
@@ -278,7 +337,7 @@ should be able to reason about `<template>` with this desugaring in mind.
 2. Expression
 
    ```
-   import Foo form 'somewhere';
+   import Foo from "somewhere";
 
    export const Bar = template`<Foo />`;
    ```
@@ -286,8 +345,8 @@ should be able to reason about `<template>` with this desugaring in mind.
    becomes
 
    ```js
-   import { template } from '@ember/template-compilation';
-   import Foo form 'somewhere';
+   import { template } from "@ember/template-compilation";
+   import Foo from "somewhere";
 
    export const Bar = template`<Foo />`;
    ```
@@ -295,7 +354,7 @@ should be able to reason about `<template>` with this desugaring in mind.
 3. Class
 
    ```
-   import Foo form 'somewhere';
+   import Foo from "somewhere";
 
    export default class Bar {
      <template><Foo /><template>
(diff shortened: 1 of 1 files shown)


text/0813-template-layering.md
```diff
@@ -0,0 +1,332 @@
...
+   }
+   ```
+
+   becomes
+
+   ```js
+   import { template } from '@ember/template-compilation';
+   import Foo form 'somewhere';
+
+   export default @template(`<Foo />`, () => ({ Foo }) class Bar {
```
NullVoxPopuli: can / should this RFC mention the pre-decorators version of class decorator usage?

I think it'd be:

```js
export default template(`<Foo />`, () => ({ Foo }))(class Bar {
});
```

text/0813-template-layering.md
```diff
@@ -0,0 +1,332 @@
...
+Also, with default helper manager, you can pretty much do anything along those
+lines anyway with an inline closure that you invoke inside the template.
+
+## Alternatives
+
+**TBD**
+
+## Unresolved questions
+
+**TBD**
```
NullVoxPopuli, line 343: would we have the ability to use this at runtime?

I imagine this `template` function _always_ converts to the underlying format, but in apps, a build step could swap out the implementation -- allowing `template` to exist at runtime, if imported and used dynamically, such as the use cases of:
 - https://github.com/NullVoxPopuli/ember-repl
 - [limber.glimdown.com/](https://limber.glimdown.com/)

text/0813-template-layering.md
chancancode: Yeah we can definitely bikeshed what is the most ergonomic version of this should be. I don't super love the decorator approach either, I just pushed a commit for a version using a static initializer block (matching the template/backtick version)

text/0813-template-layering.md
chancancode: Yes that's the idea, see:

> It is designed to be pre-processed at build time where possible, just like the `<template>` feature and today's "official" `hbs` tag. However, in cases where this is not possible, there will be a runtime implementation available provided the template compiler is also available at runtime.

chancancode: Which is why it's important that the `template()` version doesn't do any magic that is not actually possible at runtime.

text/0813-template-layering.md
NullVoxPopuli: well, my motivation here is, I think aligned with the motivation of this RFC, to potentially allow shipping of ember "stuff" without transpilation.  Who knows how long it'll take before decorators are native, ya know?

text/0813-template-layering.md
NullVoxPopuli: YAAAS

text/0813-template-layering.md
chancancode: What do you think about the static initializer block version? I like that it keeps it inside the class body, matching `<template>` more closely. It always does not _intrinsically_ require the use of static initializer blocks if it's not supported in your environment, it's basically just a "three arguments version" of the function and you can wrap it around the class if you want to.

chancancode:
By the way, one of the things that I think was a missed opportunity is we didn't discuss whitespace-folding rules for `<template>` (which isn't too late to do still). We should take the opportunity to explore the possibility of auto-folding whitespaces or at least auto trimming the indentation from the tag? That has implications around how careful the de-suagring needs to be when it comes to matching the whitespace/indentation. That said, I think it is feasible to totally match match the whitespaces/indentations/lines exactly with this design in a transformation, if that turned out to be important.

text/0813-template-layering.md
chancancode: I pushed an update making the static initializer block version the preferred one, and moved the class decorator to the alternatives section.

nightire:
> We should take the opportunity to explore the possibility of auto-folding whitespaces or at least auto trimming the indentation from the tag

In that case, it should provide an option to preserve the whitespace for particular purposes like rendering code blocks, etc.

NullVoxPopuli:
> it should provide an option to preserve the whitespace for particular purposes like rendering code blocks, etc.

idk about that -- I think doing something like `stripIndent` from https://www.npmjs.com/package/common-tags is straight-forward~ish, and does what everyone wants, while still _allowing_ the use of invisible characters to sensibly make it in to the output.

_but_, if we want to bike shed, _I_ think it'd be _stellar_ to have an option to strip all non-single space invisible characters -- But!!! with how this proposal is laid out, I don't think we need to support that, because we can do:
```js
import { stirpIndent } from 'common-tags';

const MyComponent = template(stripIndent`
  <p>
     hi
  </p>
`, {});
```
which, as long as we design the build-time transform to evaluate the first arg to template (somehow? I'm fuzzy on this -- anyone know how it'd work? could lead to security problems? (I can only think of `eval`)), stripIndent would make the template go from:
~~~
`
  <p>
    hi
  </p>
`  
~~~ 
to
~~~
`<p>
  hi
</p>
`  
~~~

text/0813-template-layering.md
NullVoxPopuli: I have internal conflict. So like,

I'm used to having to do:
```js
setComponetnTemplate(hbs`my stuff`, MyClass);
```
which _could_ look like very similar to the decorator usage of template (which, maybe for other readers, class decorators are syntax sugar for functions that take a class and return a class)
```
@setComponentTemplate(hbs`my stuff`)
class MyClass {

}
```

<details><summary>implementation</summary>

```js
export function setComponentTemplate(...args) {
  if (args.length === 2) {
    _setComponentTemplate(...args);
    
    return args[1];
  }
  
  if (args.length === 1) {
    return (Klass) => {
      _setComponentTemplate(args[0], Klass);
      
      return Klass;
    }
  }
  
  throw new Error(`Unsupported arity: ${args.length}`);
}

```

</details>

Now, for the static initializer block -- I had never seen these in use before now, so that's exciting (TIL, etc) -- I can't _comfortably_ have an informed opinion on them yet though, only gut reactions.

And my gut reaction is that the static initialize block achieves the intent we want:
 - a component's template is not dynamic (or rather, you're in for a world of pain if you make it dynamic)
 - it's clear that the template is the same for every instance of the component
 - the association between template and component is also clear (whereas in can be missed with things like `setComponentTemplate`
 
 And what I'm wary of:
  - people unfamiliar with static initializers may be like "wtf? what is `this` in a static context", "why is JS weird?" (lol), "not `this` again" -- to which I say, _learn the language you're using, ask questions, etc,_
  - it's more typing / more blocks than just calling the function (but with the function, you have the same problem as `setComponentTemplate`, where the association is less obvious.
  
  I'm actually just realizing that we can use static initializers with `setComponentTemplate` today... which I think I may want to explore in my own stuff:
  
  ```js
  class Demo extends Component {
    static {
      setComponentTemplate(hbs`hello @name`, this);
    }
  }
  ```


So.... I think I'm in _favor_ of static initializers _because_ of the implied association between template and component.

text/0813-template-layering.md
```diff
@@ -30,9 +30,10 @@ export const Foo = template(`<Hello />`, () => ({ Hello }));
 // export class Bar {
 //   <template><Hello /></template>
 // }
-export
-@template(`<Hello />`, () => ({ Hello })
-class Bar {
+export class Bar {
+  static {
```
NullVoxPopuli, line 34: I was initially against this, but I worked through it here
https://github.com/emberjs/rfcs/pull/813#discussion_r851770215

and now I'm hugely in favor :100:

text/0813-template-layering.md
```diff
@@ -0,0 +1,345 @@
...
+
+## Detailed design
+
+This RFC proposes to re-specify the `<template>` language extension into a
+de-sugaring into `template()` calls:
+
+1. Top-level declaration
+
+   ```
+   import Foo form 'somewhere';
```
michaelrkn, line 185: from, not form :)

text/0813-template-layering.md
```diff
@@ -0,0 +1,345 @@
...
+   <template>
+     <Foo />
+   <template>
+   ```
+
+   becomes
+
+   ```js
+   import { template } from '@ember/template-compilation';
+   import Foo form 'somewhere';
```
michaelrkn, line 196: from, not form :)

chancancode:
I added a bunch more details, including the runtime opt-in, a restricted set of syntax that is guaranteed to be statically analyzable, and the question about whitespace handling. I marked it as ready for review/discussion now.

@NullVoxPopuli @nightire about the questions you asked (re: whitespace), maybe read the new section on that and we can continue from there. Short answer is no, I don't think we want to create an arbitrary macro ecosystem here and in any case, solutions like `stripIndent` doesn't work for `<template>`. I agree there are use cases where you care about whitespace, but the current/default rules doesn't really do anything for you in those cases either.

text/0813-template-layering.md
```diff
@@ -0,0 +1,856 @@
...
+configuration is also not friendly to tree-shaking/code-splitting, as it forces
+the template compiler to be included into the main/initial bundle, when it may
+only be needed in one of the infrequently used routes.
+
+This RFC takes the opportunity to improve on this by proposing a new module –
+`@ember/template-compilation/runtime`. When this module is imported, either as
+a "side-effect" import statement, or if its exported values are imported, then
+it serves as an opt-in for enabling runtime template compilation.
+
+For the time being, we propose that this module should have a single named
```
NullVoxPopuli, line 650: as someone who's worked with runtime compilation a bit, it's _very_ useful to have all the utilities used to make `compileTemplate`: https://github.com/NullVoxPopuli/ember-repl/blob/main/addon/hbs.ts#L51

for example, building a component for block-based translations: https://github.com/RuslanZavacky/ember-intl-component/blob/try-upgrade/addon/components/i18n/index.js#L44

NullVoxPopuli: by limiting the re-exports, this re-blocks us from implementing component + block based translations.

cc @ruslanzavacky

NullVoxPopuli: clarification (less strong wording):
re-blocks:
_makes us continue to rely on :shrug:-compatible versions of `@glimmer/syntax`

NullVoxPopuli: `template` overall, makes this much nicer though:

```js
import { template } from '@ember/template-compilation/runtime';
import { getTemplateLocals } from '@glimmer/syntax'; // would prefer this be re-exported as well

// ...

  htmlComponent(i18nString) {
    i18nString = i18nString.replace(/\[\[\[(.*?)\]\]\]/g, '{{yield to="$1"}}');

    let owner = getOwner(this);
    let locals = getTemplateLocals(i18nString);
    let scope = {};

    for (let local of locals) {
      let kebabName = dasherize(local);
      let componentFactory = owner.factoryFor(`component:${kebabName}`);

      scope[local] = componentFactory.class;
    }

    return template(i18nString, () => scope); // no need for a while library for this step (bye ember-repl!)
}    
```

text/0813-template-layering.md
```diff
@@ -0,0 +1,856 @@
...
+export – `template`, which is a re-export of the `template()` function we
+described above. It has the same semantics as `template()` from the main module
+except it is always invalid to use this imported function as a tag, so linters
+and TypeScript can trivial flag this invalid usage.
+
+The purpose of the re-exported `template()` from the runtime module is to
```
NullVoxPopuli, line 656: will we throw errors if the non-runtime module makes it into a build? (even if accidentally somehow, where a linter wouldn't be able to help?)

or, does it matter?

If the non-runtime module _happens_ to make it into the build, is it the exact same as the runtime module?

what are the drawbacks of only having one module instead of also having the runtime module?

text/0813-template-layering.md
chancancode: I have a few questions, but does

1. `import { compile } from '@ember/template-compilation'` not work for you today for some reason? (If so not sure re-exporting that really helps?)
2. Can you not fill the same use cases with `template`/components? In what situations does it super matter that it returns a raw template as opposed to a template-only component?

text/0813-template-layering.md
chancancode: It won't. `template()` from the runtime module is just a re-export so yeah they are exactly the same. The reason to import from the main module (plus sticking with the statically analyzable subset, both of which you should do 95% of the time), is for your code to be agnostic to where the compilation happens. The reason to import from the runtime module is to make it clear that the template should never be processed at build time.

NullVoxPopuli: so is it just a hint to tooling? does it instead make sense to provide a way to analyze usage of `template` within a module to determine if it should be left in or compiled away?

chancancode: Sure. It's more than a guarantee than a hint, and it's not just to tools, it also makes it clear to whoever is reading the code that it is deliberate. You can look at the "idiomatic usages" section. The idea is we should implement all the same rules in the linter and warn you about non-idiomatic usages. And the instructions will tell you to either not do the thing that made it not statically analyzable, or to switch over to use the runtime import to make it clear that it's deliberate. However it's "just a lint". Even if you don't change the import and let it ship to the browser, it would still work (but you will probably get an additional warning).

text/0813-template-layering.md
```diff
@@ -0,0 +1,857 @@
...
+JavaScript, for use in environments where the custom syntax extension is not
+available, and also as a publishing format for addons on NPM:
+
+[rfc-779]: ./0779-first-class-component-templates.md
+
+```js
+import { template } from "@ember/template-compilation";
+import Hello from "my-app/components/hello";
+
+// <template><Hello></template> becomes...
```
SolPier, line 27: Should't it be `<template><Hello /></template>` ?
If so, the typo is in the next block also.

text/0813-template-layering.md
```diff
@@ -0,0 +1,857 @@
...
+step if desired, but a runtime implementation can also be used via an opt-in
+described below.
+
+We also propose an alternative form that uses the `template()` function as a
+tag function for JavaScript tagged template strings. However, this variant does
+not have a runtime implementation and must be handled by a build step:
+
+```js
+import { template } from "@ember/template-compilation";
+import Hello from "my-app/components/hello";
```
SolPier, line 53: I wonder if this import wouldn't be detected as inactive by the syntax hyghlighter :thinking:

text/0813-template-layering.md
NullVoxPopuli: yup

SolPier (commented):
Just a couple of notes, nothing deep.
Happy to use this feature !
Syntax support in the editor s definitely needed.

text/0813-template-layering.md
chancancode: I pushed some updates that clarified the motivations for the RFC. This is definitely not intended to _replace_ the `<template>` proposal.

Regarding your specific question, yeah, it is likely that without a lot of additional configurations/integrations the editor will probably not be able syntax highlight the content of the code block. Variables that are referenced only from within the template blocks will likely be (incorrectly) marked as unused.

These drawbacks are enumerated in RFC #779. Perhaps counterintuitively, the experience from the prototyping phase was that it would be _more_ difficult to properly support these things inside a `.js`/`.ts` file, because the `.js`/`.ts` language support plugins generally does not provide a way to do that for (what appears to be) arbitrary string literals. That's why the `<template>`+ custom file type approach was chosen instead. We'd have to do the work of wrapping/forking the JS/TS language support plugins _anyway_, and overall we believe it will be a better experience/more practical to have these Ember-specific plugins target Ember-specific file formats.

This RFC does not suggest that we revisit any of these, but just acknowledging that there are situations in which it is not feasible/possible to use/support the custom file format, and it will still be important to have _some_ (even if degraded) form of the `<template>` feature/programming model available in those environments.

text/0813-template-layering.md
chancancode: I think we resolved this in a side conversation @NullVoxPopuli? I don't think we actually know of any use cases where it is very important that `compile()` is used instead of `template()`, which is the only other export from that module.

text/0813-template-layering.md
```diff
@@ -0,0 +1,984 @@
...
+TypeScript support, other than having a few overloaded signatures/purpose on
+the same `template()` function may make things a little cumbersome, but not
+prohibitively so. The signature of the function can be described as such:
+
+```ts
+// These types are defined in other RFCs
+interface Signature {
+  /* ... */
+}
+interface Component<S extends Signature> {
```
chriskrycho, line 815: Note: the `extends` thing actually causes a bunch of annoying issues in terms of what it does/n't support (for good reasons around interface merging) and also doesn't actually add any value (because they're type-level extensions so TS will happily let you typo away: `Arsg` is a totally legitimate thing to extend `Signature` with!), so:

```suggestion
interface Component<S> {
```

text/0813-template-layering.md
```diff
@@ -0,0 +1,984 @@
...
+  /* ... */
+}
+
+type Scope = () => Record<string, unknown>;
+
+function template<S extends Signature>(source: string): Component<S>;
+function template<S extends Signature>(
```
chriskrycho, line 822:
```suggestion
function template<S>(source: string): Component<S>;
function template<S>(
```

text/0813-template-layering.md
```diff
@@ -0,0 +1,984 @@
...
+  source: string,
+  scope: Scope
+): Component<S>;
+function template<S extends Signature, C extends Component<S>>(
```
chriskrycho, line 826:
```suggestion
function template<S, C extends Component<S>>(
```

text/0813-template-layering.md
```diff
@@ -0,0 +1,984 @@
...
+  source: string,
+  scope: Scope,
+  target: C
+): C;
+
+// For the build-time tagged template literal extension. By not specifying
+// additional arguments here, TypeScript actually enforces that the tagged
+// string literal cannot have any interpolation expressions.
+function template<S extends Signature>(
```
chriskrycho, line 835:
```suggestion
function template<S>(
```

text/0813-template-layering.md
```diff
@@ -0,0 +1,984 @@
...
+  source: TemplateStringsArray
+): Component<S>;
+```
+
+One nice thing about this design is that there is a natural place for the type
+(generic) argument for the template's signature, negating the need for the
+special `<template[Signature]>` syntax.
```
chriskrycho, line 842: This actually does *not* solve the issue from a general authoring perspective, though, I think? After all, we're not going to be teaching people to write it this way in general, and we’re *definitely* not going to be implementing Glint support for it.

text/0813-template-layering.md
```diff
@@ -0,0 +1,984 @@
...
+documented. It also does not align with the direction we are headed, where we
+encourage using the modules system and imports to express dependencies. This
+configuration is also not friendly to tree-shaking/code-splitting, as it forces
+the template compiler to be included into the main/initial bundle, when it may
+only be needed in one of the infrequently used routes.
+
+This RFC takes the opportunity to improve on this by proposing a new module –
+`@ember/template-compilation/runtime`. When this module is imported, either as
+a "side-effect" import statement, or if its exported values are imported, then
+it serves as an opt-in for enabling runtime template compilation.
```
chriskrycho, line 774:
```suggestion
This RFC takes the opportunity to improve on this by proposing a new module –
`@ember/template-compilation/runtime`. We will not compile out the `template()`
function imported from this module. Additionally, importing the module as a
"side-effect" import statement will also serve as an opt-in for including it for
dynamic runtime template compilation.
```

(I tripped over the original phrasing here and am suggesting this after chatting about it with @chancancode in Discord.)

text/0813-template-layering.md
```diff
@@ -0,0 +1,984 @@
...
+
+The teaching recommendations in [RFC #779][rfc-779] shall remain in full force.
+We continue to recommend teaching materials to be focused around `<template>`
+and `.gjs`. The `template()` feature is recommended only in standard JavaScript
+environments where adoption of `<template>` and `.gjs` is not feasible.
+
+Because `template()` is a standard JavaScript API that has a "real" import
+location, there is a natural place to document and describe its behavior in the
+API docs. This should be the primary reference of the feature, and should also
+where we document the idiomatic subset as well.
```
chriskrycho, line 854: We should use that as a hook for describing the whole end-to-end bits in the API docs, including `<template>`, too, since as you note below it's the actual desugaring for that.

text/0813-template-layering.md
```diff
@@ -0,0 +1,984 @@
...
+default behavior of the browser.
+
+However, even with that in mind, the default conclusion for the whitespace
+handling in `<template>` tags is _still_ not useful, _especially_ when the
+author is trying to preserve significant whitespace for something like code
+samples. This is because `<template>` tags often appears in JavaScript
+contexts that already has leading indentation (such as inside a class body), so
+in practice, if we preserve the whitespace exactly as it appears inside the
+`<template>` tag, it still would not do what the author was trying to
+accomplish.
```
chriskrycho, line 945: Strongly agree, and in line with the discussions we've had offline, I think we should probably apply the `HEREDOC`/etc. rules: *leading* indentation up to the indentation of the line on which the opening `<template>` appears should be stripped.

text/0813-template-layering.md
```diff
@@ -0,0 +1,984 @@
...
+  the `<template>` feature and today's "official" `hbs` tag. However, in cases
+  where this is not possible, there will be a runtime implementation available
+  provided the template compiler is also available at runtime.
+
+- For environments where it is possible to configure the development and build
+  tools to recognize the format, the tagged template literal form can be used,
+  which automatically captures the lexical scope variable bindings. This
+  directly replaces [ember-template-imports][ember-template-imports]'s "hbs
+  backtick" format but brings its naming into alignment with the `<template>`
+  feature.
```
chriskrycho, line 207: What would be an environment which can handle *that* preprocessing step but which *couldn't* handle `<template>`?

text/0813-template-layering.md
```diff
@@ -0,0 +1,984 @@
...
+Once this RFC is accepted, any future changes (including deprecations) to the
+primitive APIs referenced in RFC #779 should have no direct bearing on the
+`<template>` or `template()` features.
+
+### Tagged Template Literals
+
+Alternatively, the `template()` function can also be used as a tag function for
+a JavaScript tagged template string. This alternate form automatically captures
+the lexical scope variable bindings. However, this form does not have a runtime
+implementation and must be processed by a build step.
```
chriskrycho, line 333: This *does not* actually "work" in any of the target environments described above. For example, in CodePen, the code would parse, but you wouldn't be able to *use* it at all: The template part of it would be completely inert! I’m curious what you see this providing for users?

chriskrycho (commented):
I am a very strong 👍🏼 on the introduction of the target output, of updating the addon v2 spec to target it, and of the runtime `template()` import!

I am extremely confused (even after a few discussions with folks!) about what value the alternative tagged template literal `template` provides here compared to `<template>`, since in both cases it does nothing without a transformation. (I’m quite willing to be persuaded here, I just am not yet seeing the value there!)

text/0813-template-layering.md
chancancode: Yep, it should probably say it gives you a place to emit into. But what I was trying to say (badly) is that it does not require specifically designing something in this API that does that equivalent of `<template[Signature]>` because it already naturally exists

chriskrycho: Ah, yep, that’s clearer and I agree: We still need surface syntax for it, but it’s nice that there’s a natural home for it in the compile target. 👍🏼

chancancode:
About the tagged template string literal version, I think it's a useful feature to have that can stand on its own, but also fits into our bigger plan around the content-tags-in-JavaScript format.

But before that, I think we are in broad agreement that `<template>` is going to be the primary format. I think #779 (and the blog post) made good arguments about why that is necessary, and I believe this RFC made it pretty clear it's not about re-opening that conclusion. It remains the case that we will recommend `<template>` and use it throughout our teaching materials, etc, and when everything is implemented, it is what is going to give users the best experience, both because that's what we are going to focus on building tools for, but also because of the intrinsic difficulties of adding functionalities into existing JS language support.

However, even with a degraded experience, it is still useful to have a version of the feature using only legal JavaScript syntax. Broadly, I can think of two kind of use cases:

1. Where you are able to run a full/normal ember build but have little ability to customize the editing experience
2. Where you are uninterested in actually running the code but it is important that the code parses/highlights correctly

[codesandbox.io](https://codesandbox.io) is probably a good example of the first case. It allows running arbitrary build code in a Node/remote container environment, but the editor UI is implemented in the browser and offers little in the way of an editor customization/plug-in system. In this scenario, if you were to use the `<template>`/`.gjs` format, you will get no syntax highlight, and any other editor features that are normally available to JS code. On the other hand, using the "template backtick" version in this proposal, you will still not get syntax highlighting for the templates, and you may have to disable or ignore some unused variable warnings, but otherwise you will still be able to use the normal JS features.

We may or may not be able to work with the codesandbox maintainers to get this particular case resolved quickly, but the general point is that "tooling support" is a long tail, and I am confident that we will get the major ones ticked off pretty quickly and provide a superb experience in those environments, there will be a bunch of places where the support trails behind or where we just cannot support due to restrictions outside of our control.

As the `<template>` programming model become mainstream, it will become the main way Ember users learn about, think about and write components, and everything else will slowly fade into unfamiliar/"obsolete" territories over time, in the same way that there are probably plenty of Ember users that never wrote a "classic Ember class" anymore. It is important to that we offer a comparable programming model for these environments so users who operate in them are not forced to drop down to a completely different programming model, whether that means "the old way of doing things" or "a much lower-level way of doing things".

In my opinion, having this available as an option will also help with adoption. As editor support is still being worked on, some of these benefits apply today for editors that are still unsupported, and if some of your teammates aren't using the "mainstream" environments that could be a blocker to adoption as well. The "template backtick" version can help fill the gaps in the meantime, provides the same programming model (in terms of thinking about writing components) and is 100% codemod-able into the `<template>` syntax when desired.

For the second type of use cases, the RFC text alluded to a few possibilities, but mostly it amounts to syntax highlighting in blog posts, gists, that kind of stuff (so codepen actually isn't a good example of that).

Again I think these reasons are good and important enough that the feature can stand on its own, but also this layering approach fits into the broader "plan" ([sketch](https://wycats.github.io/polaris-sketchwork/content-tag.html)) around content-tags-in-JS.

The general idea is that we write a generalized specification that enables things like `<template>` tags to work (in theory, it can be proposed as a language feature for JavaScript itself). In the generalized version of the feature is that:

* `<foo>...</foo>` desugars into <code>foo`...`</code>
* `<foo bar="baz">...</foo>` desugars into <code>foo({ bar: "baz" })`...`</code>
* When inside a class body, it desugars into a static initializer block:
   ```
   class MyClass {
     <foo bar="baz">...</foo> // => static { foo({ bar: "baz" })`...` }
   }
   ```

...and that's it. It doesn't synthesize a default export, doesn't automatically add imports, etc.

The point of this is to specify a general-purpose syntactic transformation/desugaring for the feature into standard JavaScript, while not specifying what the syntax actually _does_. This allows us to write a shared tool for this "first-stage" transformation that can be used by anyone interested in the feature (say, adding supporting `<style>` tags in component classes), but then could use standard tools (Babel, etc) to implement the tag-specific logic/semantics (e.g. pulling out the styles into an external stylesheet file).

In terms of implementing the `<template>` feature, the ideal layering would be:

1. Transform them into "template backtick"
2. In Babel, transform "template backtick" into idiomatic `template()` with the scope closure
3. In Babel, transform idiomatic `template()` calls into the low-level primitives

We don't necessarily have to immediately go out of our way to rewrite code that already works, but that conceptual layering/desugaring should work.

text/0813-template-layering.md
chancancode: I think I cleared this up some more in the most recent update

text/0813-template-layering.md
```diff
@@ -0,0 +1,1001 @@
...
+not intend to revisit its recommendations. However, there remain cases where
+this drawback is highly undesirable or simply not acceptable:
+
+1. Until we have implemented good editor integration, the developer experience
+   of using `.gjs`/`.gts` could be abysmal (without any syntax highlighting or
+   completion at all, etc). Even after we ship good editor integration for the
+   mainstream editors, this will likely remain the case for a subset of less
+   used editors that we did not or could not prioritize supporting.
+
+2. There may be editors and environments that are impossible for us to support,
```
MelSumner, line 101: This is not unique to this RFC but applies to Ember in general. As such, I am not sure it's relevant to include as a supporting reason.

text/0813-template-layering.md
```diff
@@ -0,0 +1,1001 @@
...
+However, one of the drawbacks is that it requires a custom syntax extension
+(`<template>`) and a custom file format (`.gjs/.gts`), which requires a build
+step and other custom tooling.
+
+RFC #779 makes a good case for why this is necessary and preferable to the
+alternatives. We stand by the rationales given in RFC #779, and this RFC does
+not intend to revisit its recommendations. However, there remain cases where
+this drawback is highly undesirable or simply not acceptable:
+
+1. Until we have implemented good editor integration, the developer experience
```
MelSumner, line 95: [question] I thought we accepted that editor integration (to include syntax highlighting etc) was a blocker to implementation (with the idea that it would drive implementation). When did that change?

NullVoxPopuli:
>  to ensure we never did anything like what this RFC is proposing. 

how do you mean?

this RFC is _essential_ for a lot of reasons.

chriskrycho:
@NullVoxPopuli your response here is part of why I had actually suggested breaking this into two separate RFCs: *dynamic/client-side template compilation* is indeed finicky, but it does not require the `template` tagged template literal; it only needs the `template()` import to which that compiles (and indeed the tagged template literal form is just as inert as `<template>` is without a Babel transform operating on it!). What @MelSumner is alluding is that it was a lot of work and a very narrow path we walked to consensus around `<template>`, which very carefully *excluded* tagged template literals as an *authoring format*.

This remains my concern as well: for all that the RFC indicates caution around messaging, I think it’s important that it go further—it’s totally fine by me if the tagged template literal `template` is a legal intermediate “desugaring” of the `<template>` tag (in view of generalizing the capability and making it viable for non-Ember users to adopt the same semantics!) which is then further compiled into our `template()` invocation. However, I think very strongly that we should *not teach it as an alternative in any way*: only as the desugaring.

I do not (yet) find the motivation offered here for the tagged template literal *as even a secondary authoring format* (which it is explicitly offered as here) to come close to warrant reopening that very, very difficult consensus thing.

Mind: this *really* comes down to “How Do We Teach This?” from my POV because, again, I’m on board with it as a well-rationalized desugaring as part of a multi-stage build pass. I think it makes an enormous difference in terms of the relationship to the existing consensus we worked hard to achieve whether or not we treat the tagged template literal form as something we should encourage users to author under any circumstances or not. I think we can probably have consensus on it quite easily as an intermediate compilation formation, and I am not sure if or whether we could get to consensus for it *at all* as even a secondary authoring format.

In other words, this—

> In my opinion, having this available as an option will also help with adoption. As editor support is still being worked on, some of these benefits apply today for editors that are still unsupported, and if some of your teammates aren't using the "mainstream" environments that could be a blocker to adoption as well. The "template backtick" version can help fill the gaps in the meantime, provides the same programming model (in terms of thinking about writing components) and is 100% codemod-able into the `<template>` syntax when desired.

—is the crux of the issue. @chancancode's RFC proposes adding a secondary, but *real*, authoring format for first-class component templates, *not simply an update to the compilation pipeline*, and that is the part which is controversial.

NullVoxPopuli:
> which very carefully excluded tagged template literals as an authoring format.
> it’s totally fine by me if the tagged template literal template is a legal intermediate “desugaring”

Ye! To me, and what I'm most excited for, is that 
this wrapper API (roughly, squinting) can be _maintained by the framework_
(it currently lives here: https://github.com/NullVoxPopuli/ember-repl/blob/main/addon/hbs.ts )

<details><summary> example implementation of `template` </summary>

```ts

import { precompileJSON } from '@glimmer/compiler';
import { getTemplateLocals } from '@glimmer/syntax';
import { setComponentTemplate } from '@ember/component';
import templateOnlyComponent from '@ember/component/template-only';
import { createTemplateFactory } from '@ember/template-factory';

import { nameFor } from './utils';

/**
 * compile a template with an empty scope
 * to use components, helpers, etc, you will need to compile with JS
 *
 * (templates alone do not have a way to import / define complex structures)
 */
export function compileHBS(
  template: string,
  options: Omit<CompileTemplateOptions, 'moduleName'> = {}
) {
  let name = nameFor(template);
  let component: undefined | unknown;
  let error: undefined | Error;

  try {
    component = setComponentTemplate(
      compileTemplate(template, { moduleName: name, ...options }),
      templateOnlyComponent(name)
    );
  } catch (e) {
    error = e;
  }

  return { name, component, error };
}

interface CompileTemplateOptions {
  moduleName: string;
  scope?: Record<string, unknown>;
}

export function template(source: string, scope) {
  let locals = getTemplateLocals(source);
  let moduleName = '(dynamically compiled template)';
  let options = {
    strictMode: true,
    moduleName,
    locals,
    isProduction: false,
    meta: { moduleName },
  };

  // Copied from @glimmer/compiler/lib/compiler#precompile
  let [block, usedLocals] = precompileJSON(source, options);

  // some assertion here about locals missing from `scope`
  assert('...', ...);

  let blockJSON = JSON.stringify(block);
  let templateJSONObject = {
    id: moduleName,
    block: blockJSON,
    moduleName,
    scope,
    isStrictMode: true,
  };

  let factory = createTemplateFactory(templateJSONObject);

  return factory;
}
```

</details>

> I think very strongly that we should not teach it as an alternative in any way

100% on board with this. `template` is solely a power-user tool (dynamic or compileless environment, etc)

> but real, authoring format for first-class component templates, not simply an update to the compilation pipeline, and that is the part which is controversial.

I'm not 100% sure what's controversial, even after reading this -- but maybe I can try re-phrasing and we can see how close I am? (I like lists)

:heavy_check_mark:  `template()` as dynamic tool 
:heavy_check_mark: `template()` as an _option_ to be an intermediary compile format
:no_entry_sign:  `template()` as a component authoring tool apps addons etc (static, exists on disk)
:bulb: `template()` is not a feature for the average user, and is mostly a power-user and tooling.. tool

is this close?

text/0813-template-layering.md
```diff
@@ -0,0 +1,1001 @@
...
+
+- It is designed to have the same semantics as the `<template>` feature (such
+  as the strict-mode opt-in, returning a template-only component when not
+  attached to a class), providing the same high-level programming model for
+  authoring components in standard JavaScript environments.
+
+- It requires manually supplying the lexical scope variable bindings (the
+  "scope function").
+
+- It uses a [static initializer block][static-block] to associate the template.
```
snewcomer, line 188: Based on reading the Class code snippets, the proposal is to use the static initializer block to side effect the association of the template?  Or is there actual static initialization on the class itself?  Perhaps inconsequential, but was hoping to enhance my understanding of this proposal.

text/0813-template-layering.md
```diff
@@ -0,0 +1,1001 @@
...
+
+It is not technically impossible to accomplish these same goals in standard
+JavaScript environments. After all, the `<template>` feature is specified using
+primitives that already exist. Users in standard JavaScript environments can
+just use those primitives directly.
+
+However, that is not an ideal outcome. `<template>` is a user-facing feature,
+and is poised to (or at least well-positioned to) become the main way Ember
+users read, write and reason about components in the next edition of Ember when
+the feature is fully rolled out.
```
snewcomer, line 132: `<template>` vs `template()`.  What prevents the latter from wide adoption and use in the community?  Do we just expect users to prefer the format in rfc 779?

snewcomer (commented):
👏

chancancode:
I would put it a bit differently. I don't think "authoring format" and "primitives" are as real/concrete as a distinction than we try to make it sound.

* * *

By the standard we are using here, I think you could say the `class {}` syntax in JavaScript is the "authoring format" and the function-as-a-constructor-plus-prototype-manipulation paradigm is the "primitive". Likewise the getter/setter syntax can be called the "authoring format" and `Object.defineProperty` would be the "primitive".

You could _technically_ say that there is nothing "preventing" the primitives from gaining wide adoption – it's not like you get a warning when you use the primitives and they didn't make you type `Object.dangerouslyDefineProperty()`. _Technically_ it would also be correct to say that TC39 just "expected" users will prefer the nicer "authoring format". But I think most would agree that's a strange angle to analyze the situation.

In those examples, I think it would be wrong (at least overly strong) to say the primitives should never be recommended under any circumstances. It is true that you would generally prefer to use the high-level syntax, it feels better to use, and tools like TypeScript gives you better support. However, occasionally you find yourself in strange situations where the high-level features just didn't work for you. Maybe you are dynamically modifying a class, writing a decorator-esq thing, etc.

Perhaps you should reconsider whether you really should be approaching the problem that particular way, and sometimes that leads you to better alternative designs. But there are also less common but totally valid reasons for needing to do that. It's true that when you reach for the primitives, it tends to get a bit awkward, is likely to raise some eyebrows on code reviews, and you probably need to write a comment explaining yourself. But other than that, the primitives are totally "real" and have exactly the same semantics. If you have the need for them, you can reach for them, you can write the code, you can find the docs on MDN, you can choose to save it on disk, or not. All of the "can"s here doesn't mean "recommended", but it also doesn't mean "never ever recommended" or "we didn't mean to tell anyone about these".

* * *

We can also look at some examples from the Ember ecosystem.

In the Octane time frame, we shipped two features that changes the "authoring experience" for components – making template-only components have outer-HTML semantics (#278) and component template co-location (#481).

The intended design is for components to be a fairly lightweight concept in Octane. At the most basic form, all they are is a way to make reusable markup snippets. You can easily create one buy making a `.hbs` file in the components folder, copy some markup into it and you can now reuse it anywhere. Need to do something sightly different on each invocation? There are arguments and blocks for that purpose. Need to do more than that? Add a class when you need one. In this world, "template-only components" aren't special at all, they are just what a basic bare bones component that didn't happen to need a class (yet). It's a subtle but important paradigm shift and you can see this in action in the Octane tutorial and I think most would agree that felt great.

When we first proposed/shipped #278, template-only components are specified in terms of the user-facing feature – it is a bare `.hbs` template file on disk without a corresponding `.js` file at the expected filesystem location. The feature can be enabled by an addon (which became `@ember/optional-features`). The rest are all private implementation details.

This made sense, as that's how most users would interact with this feature on a day-to-day basis. However, over time, we realized that there are a small but real use cases or environments where these specifications didn't make sense or is not reliable/acceptable. For example, addon authors cannot generally count on the flag being enabled in consuming apps, the addon opt-in and filesystem specifications didn't necessarily make sense for Ember Twiddle and other exotic/non-Ember-CLI environments.

Therefore, when we did #481, which was primarily about the user-facing component template co-location feature, we also introduced the primitives at the same time, like `setComponentTemplate()` and `templateOnlyComponent()`. While the RFC was pretty clear that we didn't intend for users/app authors to use these APIs directly, the feature was specified in terms of these APIs, so that if you did find yourself in a situation where you need to replicate the same behavior, these primitive APIs would give you the same end result.

I think that divide held up reasonably well overtime. If you were to survey 100 random Ember users, most of them probably have never heard of or used these APIs. And yet, there just are use cases where they are useful and necessary (the `<template>` feature being one of them!), some of which we didn't even anticipate at the time.

Another thing that happened, TypeScript (Glint) users ended up writing `templateOnlyComponent()` on the regular, which I am personally not thrilled about. (This is not a dig at the amazing work by the TS/Glint team did to enable template type checking!) The API was really meant as a low-level tool and isn't really a sensible concept for app authors to have to learn about. The whole point of this aspect of the Octane design is that a "template-only" component isn't a distinct concept, it's just a regular component but without a class.

But at the end of the day, the confusion we were trying to prevent was pretty bounded and manageable. We would have preferred to present the thoughtful design to all of our users, but it's not like the programming model is a fragile house of cards that falls apart immediately if you look at it under the wrong light. Perhaps it would have been better if we worked with the TypeScript/Glint sooner to develop an alternative that preserved the intended design (which we _are_ doing with `<template>` in Polaris), but it was also a far cry from table-flipping incoherence. Until then, it's on balance a good thing the workaround exists, even if it's sub-optimal, and that the TS/Glint could do what they needed to do without being blocked.

With all that said, I definitely do _not_ regret including `templateOnlyComponent()` in the RFC. I am happy we didn't ship only the user-facing feature and leaving other use cases outside of the intended happy path in the dust. I am happy we didn't design the primitives so narrowly that they were only usable in the blessed use cases. I am happy we didn't make them artificially cumbersome to use just because they were intended as compilation targets and we can get away with making them extra verbose and awkward. (If we had done that, we would probably have just made TypeScript users type the very cumbersome thing on the regular and have changed nothing else!) I am happy that the TS team didn't have to propose another adjacent primitive that only worked for them, just to have @NullVoxPopuli come back to propose something else for their REPL thing. I am glad we documented these APIs and didn't intentionally make them hard to find. We chose not to teach them in the guides, because there were no need or them at the time (which is more or less by design), but I am happy that we didn't legislate into the RFC that they shall never be mentioned in the guides. 🙃

Another example from the Octane time frame was that we introduced the `@tracked` decorator. That worked great for the most part but there were situations where decorators just didn't work. Maybe you are in an environment that doesn't support them. Maybe you are not writing a class. Maybe you have an unbounded set of dynamic keys. And at the time, the status quo was that you either use `@tracked` or you can use the classic object model (`Ember.get`, `Ember.set`, computed properties, etc), which really sucked when you happen to not be on the happy path, which points to a "missing middle" layer of API.

So we did #669 to introduce the primitives, which we intentionally chose to do rather than simply solving the most common symptom of the problem by introducing tracked built-ins (which is a great set of APIs are on the way of being standardized into Polaris, but they couldn't solve all the problems/scenarios). With the benefit of hindsight, IMO it would have been better if they were rolled out at the same time with `@tracked`, but better late than never. There were technically nothing preventing the primitives from gaining wide adoption instead of the decorators in that case either, but we don't generally think that's something we need to worry about. In fact, the RFC went out of its way to show how the `@tracked` decorator could be implemented with the primitives – it even specifically recommended that we include that in the in-depth guides!

* * *

So back to this RFC. The intention here is to fill out the "missing middle" APIs for the `<template>` feature, and I am happy to categorize them as primitives. Some of those niches/needs came up here and there in the original RFC comment thread, some of them were left as work for the future (which I am picking up here), some were too niche/lacked enough interest to solve at the time.

I think the context were just pretty different. In the work leading up to #779, the focus was designing and standardizing the primary end-user facing feature. That made sense, and that's what matters the most at the end of the day. I am happy with the choices we made there, and it makes sense that we prioritize the most important day-to-day use cases and focus on how we could make that the best possible experience. Like `@tracked`, it's okay that it doesn't solve all the possible use cases, and it wouldn't make sense to accommodate niche use cases at the expense of a compromised experience for the day-to-day usages. But from the past experiences, these can be solved by composing the high-level features in well-designed layers of primitives, which is what I am trying to accomplish here.

In an edition, each individual feature is just a single puzzle piece that forms a bigger picture.

With Octane, there were a bunch of individual changes the the component programming model – outer-HTML template-only components, defaulting to template-only components in the generators, co-location, angle bracket invocations, splattributes, `@namedArguments`, `@model`, Glimmer Components, modifiers... Each of them have their own benefits and can stand on their own, but the most important thing is the overall paradigm shift – components are a lightweight abstraction for reusing markup, like functions but for HTML, with the ability to parametrize them with arguments and blocks, and the ability to abstract state and behavior using classes, etc.

In Polaris, `<template>` is certainly a very important (critical, even) feature, but there is a more important bigger picture paradigm shift we are trying to accomplish here – we want to move away from name-based resolutions, we want imports, we want to break down the barriers between JavaScript and templates (as in – "ease of refactoring between JS and templates", not "blur the line and make JS code and templates the same thing"). `<template>` is how most users would get those benefits (but it's not the only thing that matters, other features like default managers are part of that picture too), and it's important that we make that experience as polished as possible and get as many users onboard as possible.

But at the end of the day, what is even more important is to get _everyone_ on board with the _new components programming model_, and we cannot do that with `<template>` being the only way to access that world. Just like `@tracked`, we must acknowledge that it has its limitations and there are situations where it is not going to work. And just like with `@tracked`, we can solve that by layering the APIs so that there will be something else to fallback to, whether they are used directly or for someone else to build something on top of those layers that is fitting for the unique situation we didn't account for.

And to be clear, I am not talking about presenting these as "alternative" as in "beef, chicken or vegetables". It's more like "you can get the binaries for this tool, or install it with `cargo`, or build it from source yourself". Or perhaps something closer to home: use native `<button>` or `<select>` tags if you could, because they do everything right for you, and if you think you couldn't, think again harder because there is probably some way to approach the problem differently, but otherwise, what they do for you can be described and broken down into the lower-level aria roles, aria attributes and other behavioral specifications so you can compose those to do the right thing in your use case. 

With `<template>`, yes, it's already built on primitives, but those primitives represent the legacy worldview that doesn't really match the new Polaris paradigm (such as working with "raw templates" as a distinct concept from components), so it's worth replacing them with something that matches the `<template>` programming model.

Now, is it possible that someone would take the primitives and use it in ways "we don't like"? Sure? Does that mean it's a bad idea to expose the primitives? I don't think so. First of all, we should probably have less opinion on what experiments others choose to do in their free time and how people want to write their code 😛 But more importantly, if people are motivated enough to experiment/deviate from the happy path, they are going to do it anyway. By providing more strategically placed branching-off points/off-ramps, we actually get to share more and keeping everyone closer together and headed in the same overall direction. If the only available branching-off point is all the way down at the bottom and they have to rebuild a lot of the same things and remake a bunch of the decisions we already made, that's when you end up getting people veering way off course.

Case in point, whether you personally prefer it or not, projects like Emblem exists and some people like to use them. We don't have to go out of our way to support them, but where possible we should work towards layering the work so that it is possible for someone who is motivated enough to make it work can leverage a lot of the work we put into making `<template>` work (hand wave: something something transforming into the hbs before the rest of the pipeline "sees" the templates). Otherwise, they may end up deciding it's not worth the effort and decide to stick with standalone template files plus some kind of `{{import}}` helpers. And I think _that_ would be a worse situation overall. (Not claiming this RFC solves the Emblem problem, though along with the content-tag proposal I think it gets us much closer.)

* * *

So, specifically about the tagged template literal part of the layering. I am happy to call it an intermediate format.

I think there are use cases for it where that makes sense, and honestly I think I made a pretty good case for it with the codesandbox scenario. There just are places where running a build is not an issue but the editor experience is the bottleneck. We can agree that this is not a super common situation and that this is not a massively popular/important use case, and that's a good thing! Otherwise, we will be in even deeper trouble! I don't think that's not really the point and the bar we need to clear for these middle/primitives APIs. We landed `templateOnlyComponent()` originally just because we needed something for addon authors, and the other use cases were mostly discovered organically and that's totally fine.

I also stipulated that some users (because of their editor preferences, or because they want to be an early adopter) may want to adopt the programming model but is constrained by the editing experience. I think that's just a special case of the same situation that is perhaps more concrete and easily to relate to, but now the concern is "woah, now that's _too_ useful and uncanny, it solves that problem almost _too well_ and people may get comfortable with it and not want to switch", so that kind of feels like a catch 22 situation in terms of justifying whether this feature is actually useful or not. 🙃

As far as why isn't just having `template()` good enough, I think it's the same reason why just having `precompileTemplate()` isn't good enough. The factor here being the scope closure. It is cumbersome and hard to get right. In situations where you can't avoid it, it's better than nothing, but when possible, we should prefer to steer people (even "power users") away from having to write it, either by hand or by re-implementing the logic to generate one. While the high-level of "capture any referenced lexical variables" is easy to say, we _still_ have unresolved semantics questions around the finer details of how exactly this works (which we should resolve asap!), and it would be unfortunate if those details need to be widely understood.

Even if we ignore the writing-by-hand cases, it's still good for us to layer our tooling that way so that a third-party babel transform (say, something to do with Emblem or some other experimentation that requires splicing in components) can be inserted before ours and emit the backtick code and let our plugin handle generating the correct scope closure, rather than having to target the `template()` directly and have to re-implement the logic.

I have more thoughts but this is getting long and it's getting late. The bottomline is, I agree this is probably mostly a "How Do We Teach This?" issue, but I don't think it's as simple as what words we do or do not say in the docs. I think "authoring format" vs "primitives" is kind of a red herring, in the real world they are not going to be so black and white. When we introduce APIs, of course we expect some people to actually use them. We can be clear about our intentions, provide good guidelines, clear documentation for their purpose, carrots, etc, but at the end of the day we need to trust those who are writing the code to pick the right tools for their situations given our guidance. We can't be everyone's code parents.

So, I think it makes more sense to focus on what future we are intending to build and make that clear – `<template>` is the user-facing high-level feature that we recommend to everyone and focus our resources on, while "template backtick" and `template()` are the de-sugaring/primitives that it is built on; it's not an either-or-depending-on-your-preference choice, it's not `--pods` (or maybe saying it _is_ like pods is the better deference, take your pick! 😄); we are not going to have a toggle on the guides to switch between the two; we are probably not going to have a generator flag that you can save to `.ember-cli`. I think we can focus on the making `<template>` a good experience for most and the core Ember community has a pretty good track record of sticking with the happy path that we recommend (sometimes maybe too much! see: the "packages in the  blueprint" issue), and I don't think we need to (or is appropriate for us to) police what other alternatives/experimental things people want to do.

Happy to have the conversations at the meeting tomorrow and I am sure that would change things too, so I'll leave the rest for later.

habdelra:
I'm really excited about this RFC. I'm working on a project that where one of our steps is to compile JS using babel at runtime. Our js includes glimmer templates. One of the challenges that we are running into is that the various babel plugins for class property and decorator support can alter a simple class into something that is almost unrecognizable. For instance if you have a class that looks like:
```js
export class Person extends Card {
  @field firstName = contains(StringCard);
  @field lastName = contains(StringCard);
  static embedded = class Embedded extends Component<typeof this> {
    <template>
      <@fields.firstName/> <@fields.lastName/>
    </template>
  }
}
```
it will result in this after a babel compile with the standard presets (where we use the `'ember-template-imports/lib/preprocess-embedded-templates'` to preprocess the `<template>` tags):
```js
.
.
export let Person = (_class = (_class2 = class Person extends Card {
  constructor(...args) {
    super(...args);

    _initializerDefineProperty(this, "firstName", _descriptor, this);

    _initializerDefineProperty(this, "lastName", _descriptor2, this);
  }

}, _defineProperty(_class2, "embedded", (_GLIMMER_TEMPLATE = __GLIMMER_TEMPLATE(`
      <@fields.firstName/> <@fields.lastName/>
    `, {
  strictMode: true
}), class Embedded extends Component {
  constructor(...args) {
    super(...args);

    _defineProperty(this, _GLIMMER_TEMPLATE, void 0);
  }

})),
.
.
```
sadly that moves the `__GLIMMER_TEMPLATE` tags into syntax that is not supported by `ember-template-imports/src/babel-plugin`, The ember-template-imports plugin can only (understandably) support a few different syntaxes:
```
/**
 * Supports the following syntaxes:
 *
 * const Foo = [GLIMMER_TEMPLATE('hello')];
 *
 * export const Foo = [GLIMMER_TEMPLATE('hello')];
 *
 * export default [GLIMMER_TEMPLATE('hello')];
 *
 * class Foo {
 *   [GLIMMER_TEMPLATE('hello')];
 * }
 */
```
I think providing a `template()` function with a runtime meaning would be really helpful to these sorts of scenarios where we need to interact with compilers, like babel, where it may perform these complex but valid transformations on the underlying javascript.

wagenet:
I'd like to try to catch up on this, but maybe someone who has been following along can summarize where this stands.

ef4:
My understanding here is that when @gitKrystan and @wycats did the low-level integration work for ESLint *et al* their finding was that having a JS representation like this was less useful than standardizing an AST-based representation. If that's correct, we may want to replace this RFC with one that standardizes the AST representation of `<template>`.

gitKrystan:
> My understanding here is that when @gitKrystan and @wycats did the low-level integration work for ESLint _et al_ their finding was that having a JS representation like this was less useful than standardizing an AST-based representation. If that's correct, we may want to replace this RFC with one that standardizes the AST representation of `<template>`.

Here are some issues we've been having that might be alleviated with a standardized AST representation:

https://github.com/gitKrystan/prettier-plugin-ember-template-tag/issues/43
https://github.com/ember-cli/eslint-plugin-ember/issues/1659
https://github.com/ember-cli/eslint-plugin-ember/issues/1750

ef4:
Discussion from core framework review meeting:

 - the more explicit form here is still desirable as the standard "vanilla" form that, for example, addons should publish to NPM. It's better than `setComponentTemplate` because `setComponentTemplate` leaks the concept of "template" separately from "component", which is not a thing we want to keep. It's also shorter and requires less imports.
 - whether or not this form is actually used in the implementation of tooling is not really the important part. Instead for tooling we probably want a library that emits an AST-based representation of the template tag.

PR #814, opened 2022-04-20
Add Stylelint To Ember blueprints

elwayman02:


text/0814-stylelint.md
```diff
@@ -0,0 +1,138 @@
...
+
+>Ember applications may choose to use CSS, SCSS, LESS, or other approaches for styling. How do we 
+>reconcile our default linting config?
+
+Today, the de facto default for Ember apps via the blueprints is CSS, so our default configuration 
+should strive to support that use-case. `stylelint` has readily-available support for popular 
+preprocessors built-in, and there are plugins like [`stylelint-scss`](https://github.com/stylelint-scss/stylelint-scss) 
+that add rules specific to those ecosystems for teams to opt into.
+
+>Why name the scripts `lint:style` instead of `lint:css`?
```
bmish, line 120: I would vote for `lint:css` instead of `lint:style`. Why?

1. The [ember-cli blueprint](https://github.com/ember-cli/ember-cli/blob/master/blueprints/app/files/package.json) already has `lint:js` and `lint:hbs` scripts that are named after the most-common file extension for each. Mixing `lint:style` in with these creates an inconsistency that I'm sure will break people's muscle memory as they try to remember the names of these scripts.
2. The existing `lint:js` script can already lint both JS/TS files, as ESLint is capable of linting both when configured properly. `lint:css` would function the same way in that it could lint different types of CSS files like SCSS/SASS when configured properly.

text/0814-stylelint.md
```diff
@@ -0,0 +1,138 @@
+---
+Stage: Accepted
+Start Date: 2022-04-19
+Release Date: Unreleased
+Release Versions:
+ember-source: vX.Y.Z
+ember-data: vX.Y.Z
```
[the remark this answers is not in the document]
SergeAstapov: probably even just `ember-cli` as none of the blueprints in `ember-source` have anything to do with CSS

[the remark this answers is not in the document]
elwayman02: ah yes I just left that alone cuz the template instructions said not to touch that section of the RFC.

https://github.com/emberjs/rfcs/blame/master/0000-template.md#L18

text/0814-stylelint.md
elwayman02: Yea I went back and forth on this and I'm open to community consensus here. I wasn't aware that `lint:js` is considered the standard for TS linting over adding a `lint:ts` command. Do you have thoughts here @chriskrycho?

text/0814-stylelint.md
```diff
@@ -0,0 +1,138 @@
...
+Relevant Team(s): Ember.js
```
bertdeblock, line 8: I think this should be `Ember CLI`?

text/0814-stylelint.md
chriskrycho: Yeah, we use `lint:js` for all JS-ish files. We use something like `test:types` in general for the type checking pass.

I think `lint:css` makes good sense.

(We’re also going to need to rethink these eventually when first-class component templates become the default, but that’s a little ways out yet and shouldn’t block this at all.)

jherdman:
I recently removed Stylelint from our applications at Precision Nutrition. There was significant push back from our team on this tool. We found that:

1. Setting up Stylelint to work with our existing stack was non-trivial. We have a fairly old, and complicated, PostCSS setup that necessitated a fair bit of massaging to appease Stylelint.
2. Resolving conflicts was non-trivial for our existing, and large, CSS code base.
3. Stylelint's TODO functionality is poor comparative to ember-template-lint, you basically end up with a wall of warnings that is _never_ resolved. It demoralizes the developer. 
4. We're transitioning to a Tailwind oriented approach. This more or less makes Stylelint meaningless in the long run for us.
5. Though it did catch some nasty bugs and help with best practices, resolving these issues was non-trivial and largely not worth the effort. "If it ain't broke, don't fix it."

I know that these RFCs ultimately afford the escape valve of simply excluding said packages from one's stack, but I thought I'd add some colour with our experience. All of this is to say I'm sure this tool helps some people, but I'm not sure it's as universal a win as the other linting tools cited.

text/0814-stylelint.md
```diff
@@ -0,0 +1,138 @@
...
+
+There are many rules, configs, and plugins available for stylelint and prettier, but we are 
+intentionally avoiding being prescriptive on anything that isn't nearly universally agreed upon 
+by the stylelint, prettier, and Ember communities.
+
+### Blueprint Changes
+
+In general, it is recommended that all blueprints provided by addons should satisfy the default 
+linting configuration of a new Ember application. As such the blueprints provided by ember-cli, 
+ember-source, and ember-data will be updated as needed to ensure that they satisfy these new
```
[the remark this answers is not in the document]
elwayman02: They likely aren't - I specifically included the language "as needed" to indicate that any and all core blueprints should be reviewed and updated to ensure they pass the new linting.

text/0814-stylelint.md
```diff
@@ -0,0 +1,138 @@
...
+linting rules requirements.
+
+#### `package.json` scripts
+
+The `app` and `addon` blueprints will be updated to add style-related linting scripts:
+
+```json title="package.json"
+{
+  "scripts": {
+    "lint:style": "stylelint .",
```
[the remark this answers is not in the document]
elwayman02: Good callout. I think this is because Stylelint supports CSS-in-JS, which we probably don't need to support in our default configuration. An ignore file would likely be difficult to maintain, so we can scope this call to the most common filetypes: CSS, SCSS, SASS, LESS

elwayman02:
> I recently removed Stylelint from our applications at Precision Nutrition. There was significant push back from our team on this tool. We found that:
> 
> 1. Setting up Stylelint to work with our existing stack was non-trivial. We have a fairly old, and complicated, PostCSS setup that necessitated a fair bit of massaging to appease Stylelint.
> 2. Resolving conflicts was non-trivial for our existing, and large, CSS code base.
> 3. Stylelint's TODO functionality is poor comparative to ember-template-lint, you basically end up with a wall of warnings that is _never_ resolved. It demoralizes the developer.
> 4. We're transitioning to a Tailwind oriented approach. This more or less makes Stylelint meaningless in the long run for us.
> 5. Though it did catch some nasty bugs and help with best practices, resolving these issues was non-trivial and largely not worth the effort. "If it ain't broke, don't fix it."
> 
> I know that these RFCs ultimately afford the escape valve of simply excluding said packages from one's stack, but I thought I'd add some colour with our experience. All of this is to say I'm sure this tool helps some people, but I'm not sure it's as universal a win as the other linting tools cited.

These concerns are already addressed in the Drawbacks section: https://github.com/elwayman02/rfcs/blob/stylelint/text/0814-stylelint.md#drawbacks

text/0814-stylelint.md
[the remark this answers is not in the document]
elwayman02: Or, should we explicitly only support CSS with the default config since that's what Ember apps are configured for, even though stylelint can support other file types without additional plugins/configuration?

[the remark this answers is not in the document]
bertdeblock: The stylelint config suggested in this RFC only works for plain `.css` files afaict, so including `scss`, `sass` and `less` seems unnecessary to me. If you do try to lint e.g. a `scss` file with the suggested config, you'd get:

```
When linting something other than CSS, you should install an appropriate syntax, e.g. "postcss-scss", and use the "customSyntax" option
```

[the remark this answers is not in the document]
bertdeblock: I still think we either need to make the path even more specific or introduce a `.stylelintignore` file (similar to `.eslintignore` and `.prettierignore`). At the moment, CSS files in the following folders would be linted as well:
- `coverage`
- `vendor`
- `dist`
- `ember-try` folders
- ...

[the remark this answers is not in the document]
elwayman02: I agree that we should probably do both - we'll need the stylelintignore but also should scope it to css files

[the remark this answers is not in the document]
elwayman02: I've restricted the scope to css file types and added a lint ignore section based on the existing ignore files.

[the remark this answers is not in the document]
elwayman02: It's possible the suggested ignore file is slightly broader than it needs to be, but I'd favor consistency with other ignore files over removing something just because we don't *expect* it to ever contain CSS files.

text/0814-stylelint.md
```diff
@@ -0,0 +1,180 @@
...
+
+#### `package.json` scripts
+
+The `app` and `addon` blueprints will be updated to add style-related linting scripts:
+
+```json title="package.json"
+{
+  "scripts": {
+    "lint:css": "stylelint \"**/*.{css}\"",
+    "lint:css:fix": "stylelint \"**/*.{css}\" --fix",
```
bmish, line 119: To avoid repeating this stylelint command/glob in two scripts, we can use this:

```suggestion
    "lint:css:fix": "npm-run-all \"lint:css --fix\"",
```

npm-run-all which is already used in other scripts allows us to avoid directly referencing npm or yarn (as we don't know which one the user is using).

text/0814-stylelint.md
```diff
@@ -0,0 +1,180 @@
...
+will be updated as needed to ensure that they satisfy these new linting requirements.
+
+#### `package.json` scripts
+
+The `app` and `addon` blueprints will be updated to add style-related linting scripts:
+
+```json title="package.json"
+{
+  "scripts": {
+    "lint:css": "stylelint \"**/*.{css}\"",
```
bmish, line 118: Should the braces include the possibly CSS extensions?
```suggestion
    "lint:css": "stylelint \"**/*.{css,sass,scss}\"",
```

bertdeblock: The Stylelint config suggested in this RFC only supports plain `css` files.
If you do try to lint e.g. a `sass` file with this config, you would get:

```
When linting something other than CSS, you should install an appropriate syntax, e.g. "postcss-scss", and use the "customSyntax" option
```

Therefore, it seemed more logical to me to only include `css`.

SergeAstapov: 2 cents from my side on this:
out of the box ESLint ember config (as well as Prettier) includes only `*.js` files. If addon or app author decides to add TypeScript support, ESLint config needs to be edited to run ESLint for `*.ts` files.

Hence leaving only `*.css` would be in line with what we do for JavaScript by requiring app/addon author to add sass/scss/etc. file extension should it be added to the codebase.

bmish: If we only support pure CSS out-of-the-box, then I suppose covering only CSS here is fine.

However, it would definitely be nicer if we could avoid mention file extensions at all somehow, the same way JS linting just uses `eslint .`.

bertdeblock: Would be nicer indeed, but don't think we can at the moment. Stylelint seems to include everything it can find when running `npx stylelint .`.

elwayman02: Yes, in my testing calling `stylelint .` produces very weird results, so we need to scope it to the file extension.

elwayman02:
Any other thoughts/concerns about moving forward with this RFC?

bertdeblock:
@elwayman02 I will bring this one up during the next CLI meeting.

wagenet:
@bertdeblock where does this stand?

bertdeblock:
Sorry for the delay. I've added this RFC to next week's agenda.

kategengler:
We discussed this in the cli meeting today and decided to move it to final comment period. We'd like to hear more feedback on this topic, if there is any, but agree that it is broadly inline with other linting we do.

void-mAlex:
in all of the projects that we work with, this would be almost a no-op due to the use of tailwindcss; I can see this potentially adding a few extra seconds to the runtime of a lint:* command

I wonder if a better approach would be to offer people a way to opt into stylelint via something like ember-apply, as I can be sure we will remove the packages on upgrade, in order to ever so slightly lessen the burden of packages needing to be kept up to date.

on a final note I believe that the proposal should also add the config file `.stylelint.js` to the `.eslintrc.js` overrides section where all the other .config files live

kategengler:
@void-mAlex If you remove the scripts in `package.json` for stylelint (would remove any extra time for the `lint` command) and the related packages, and use `ember-cli-update` to upgrade in the future, they shouldn't come back. That said, there does seem to be a maintained plugin for stylelint config. for tailwindcss.

bertdeblock:
@void-mAlex I think your last concern is covered [here](https://github.com/elwayman02/rfcs/blob/stylelint/text/0814-stylelint.md#linting-stylelints-config).

void-mAlex:
> @void-mAlex I think your last concern is covered [here](https://github.com/elwayman02/rfcs/blob/stylelint/text/0814-stylelint.md#linting-stylelints-config).

I did read that but somehow didn't make the connection even searched for eslintrc to double-check I didn't miss it. you are correct though.

NullVoxPopuli:
can we have a flag to opt out?

void-mAlex:
> @void-mAlex If you remove the scripts in `package.json` for stylelint (would remove any extra time for the `lint` command) and the related packages, and use `ember-cli-update` to upgrade in the future, they shouldn't come back. That said, there does seem to be a maintained plugin for stylelint config. for tailwindcss.

as far as I can tell the only thing it does is disable stylelint rules when you use tailwindcss and its special @ syntax
```
rules: {
    'at-rule-no-unknown': [
      true,
      {
        ignoreAtRules: [
          'tailwind',
          'apply',
          'layer',
          /** tailwindcss v1, v2 */
          'variants',
          'responsive',
          'screen',
        ],
      },
    ],
    'function-no-unknown': [
      true,
      {
        ignoreFunctions: ['theme'],
      },
    ],
  },
```
in case you're talking about `stylelint-config-tailwindcss` which is the one I found so it's not really useful as the point is to have the styles generated based on what you actually use in the template
so `app.css` has 3 lines in it

text/0814-stylelint.md
```diff
@@ -0,0 +1,180 @@
+---
+Stage: Accepted
+Start Date: 2022-04-19
+Release Date: Unreleased
+Release Versions:
+ember-cli: vX.Y.Z
+Relevant Team(s): Ember CLI
+RFC PR: https://github.com/emberjs/rfcs/pull/814
```
wagenet, line 8: Need to update front matter

wagenet (dismissed):

text/0814-stylelint.md
elwayman02: Can you explain what needs to be updated (or suggest the changes)? This comment isn't clear to me.

wagenet: The format of the YAML part at the top has changed. See https://raw.githubusercontent.com/emberjs/rfcs/master/0000-template.md

ef4:
> can we have a flag to opt out?

If you mean a flag while running the blueprint, I don't think that would be bad and it could be an implementation detail. I don't think the RFC needs to commit one way or the other.

text/0814-stylelint.md
kategengler: This should be the frontmatter:
```
---
stage: accepted
start-date: 2022-04-19
release-date:
release-versions:
teams:
  - cli
prs:
  accepted: https://github.com/emberjs/rfcs/pull/814
project-link:
---
```

kategengler:
Framework team discussed this today and have decided to merge. The actual merge is pending technicalities (updates to the frontmatter).

kategengler:
@elwayman02 Thank you!

jelhan:
> Framework team discussed this today and have decided to merge. The actual merge is pending technicalities (updates to the frontmatter).

What is the status on this one? Still pending for required changes to frontmatter?

bmish:
I've opened up an implementation for this RFC in:

* https://github.com/ember-cli/ember-cli/pull/10122

PR #819, opened 2022-05-12
Public router service

SergeAstapov:
[Rendered](https://github.com/SergeAstapov/rfcs/blob/public-router-service/text/0000-public-router-service.md)

### DIFF
--- a/text/0000-public-router-service.md
+++ b/text/0000-public-router-service.md
@@ -34,9 +34,9 @@ Today, users of [`ember-engines`](https://github.com/ember-engines/ember-engines
 custom components, helpers and services for linking. This leaves the experience feeling sub-par, in addition to introducing
 extra maintenance friction attempting to keep these primitives aligned with those in `ember-source`.
 
-By providing public a import for `RouterService`, `ember-engines` would be able to provide its own router service
-implementation extending from the base implementation. Combined with minor changes to the engines-specfici behavior of
-`link-to`,  this change would allow us to fill the existing DX gap when it comes to navigating between an Engine
+By providing a public import for `RouterService`, `ember-engines` would be able to provide its own router service
+implementation extending from the base implementation. Combined with minor changes to the engines-specific behavior of
+`link-to`, this change would allow us to fill the existing DX gap when it comes to navigating between an Engine
 and a host App *and* allow engines to eliminate use of the legacy components required to support 4.x versions of Ember.
 
 These proposed changes would further allow `ember-engines` to deprecate the `<LinkToExternal />`
@@ -90,6 +90,32 @@ Once this RFC gets implemented, above snippet could be replaced with
 
 ### `ember-source` package
 
+This proposal does not seek to make any changes to the default blueprint.
+App author may or may not introduce overridden `RouterService` via `app/services/router.js`.
+
+If app author decide to do so and provide extended `RouterService`, it may look like:
+
+```js
+// app/services/router.js
+import RouterService from '@ember/routing/service';
+
+export default class extends RouterService {
+    // ...
+}
+```
+
+In order to make this possible, method `buildRegistry` of `Application` class will be updated to have logic like so:
+
+```js
+import { RouterService } from '@ember/-internals/routing';
+
+// ...
+
+if (!registry.has('service:router')) {
+  registry.register('service:router', RouterService);
+}
+```
+
 #### `<LinkTo />` component
 
 The use of `-routing` service in `<LinkTo />` component would be replaced with the `router` service.
@@ -104,26 +130,28 @@ Currently, those methods of `-routing` service used:
 * `transitionTo`
 * `isActiveForRoute`
 
-Additionally any engines specific code or assertions for url and path generation would be removed, delegating
- these responsibilities instead to the router service. For instance, `<LinkTo />` should not be aware of `mountPoint`.
+Additionally, any engines specific code or assertions for url and path generation would be removed, delegating
+these responsibilities instead to the router service. For instance, `<LinkTo />` should not be aware of `mountPoint`.
 
 #### `RouterService`
 
-Provide public import via `@ember/routing/router-service`.
+Provide public import via `@ember/routing/service`.
 
 #### Router Helpers
 
-The changes outlined here improve the DX for engines for the not-yet implemented url helpers specified by [RFC#391](https://emberjs.github.io/rfcs/0391-router-helpers.html). Since those helpers will also utilize the routing service,
+The changes outlined here improve the DX for engines for the not-yet implemented url helpers
+specified by [RFC#391](https://emberjs.github.io/rfcs/0391-router-helpers.html).
+Since those helpers will also utilize the routing service,
 they would be compatible with both internal and external engines paths immediately.
 
 ### `ember-engines` package
 
 While not the scope of this RFC to detail the exact specifics, once these changes are implemented, `ember-engines`
-will provide a `router` service that extends from `@ember/routing/router-service`.
+will provide a `router` service that extends from `@ember/routing/service`.
 
 ```javascript
-# ember-engines/addon/services/router.js
-import RouterService from '@ember/routing/router-service';
+// ember-engines/addon/services/router.js
+import RouterService from '@ember/routing/service';
 
 export default class extends RouterService {
   // ...
@@ -148,10 +176,15 @@ Transition path needs to be published on [ember-engines.com](https://ember-engin
 
 1. Replace use of `<LinkToExternal @route="external-route" />` component
    with `<LinkTo @route="external-route" />` in every engine codebase.
-2. Users using `ember-engines-router-service` would need to replace any of it's usage with
+2. Users using `ember-engines-router-service` would need to replace any of its usage with
    the `RouterService` provided by `ember-engines`. This addon was created to solve part of
    this problem previously, and would likely be updated to become a true polyfill for the ember-engines
    routing service.
+
+Ember Guides should be updated to rationalize and explain the difference between `app/router.js` and `app/services/router.js`.
+The confusion between those two distinct yet different entities exists even today.
+However, it's less a problem today as there is only one thing named `router.js` can exist in the app today and
+ability to create another such file (even with different purpose and semantics) may be confusing.
 
 ## Drawbacks
 
@@ -169,7 +202,7 @@ Some prior discussion on this point is [viewable here](https://github.com/ember-
 In the case of this RFC, it would similarly become the responsibility of the engine's routing service to prefix such a url with the
 engine's mount path as appropriate, or decide to error, to preserve encapsulation.
 
-In practice we feel the flexibility and improved DX of this solution outweighs these drawbacks, and intend to
+In practice, we feel the flexibility and improved DX of this solution outweighs these drawbacks, and intend to
 continue enforcing the same constraints via the engines provided service.
 
 ## Alternatives
--- a/text/0000-template.md
+++ b/text/0000-template.md
@@ -1,0 +1,188 @@
+---
+Stage: Accepted
+Start Date: 
+Release Date: Unreleased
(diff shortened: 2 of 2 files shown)


chriskrycho:
Thanks for writing this up! One question I have: is the idea that the router subclass could override *all* parent methods? Or should some of them be `private` (in TS terms and also in actual TS implementation)?

SergeAstapov:
@chriskrycho as currently only methods listed as part of public API are actually implemented in `RouterService`,  I assume we should keep all the methods as public and overridable as `ember-engines` would need to do those overrides.

runspired:
> is the idea that the router subclass could override all parent methods? Or should some of them be private (in TS terms and also in actual TS implementation)?

Confirm what @SergeAstapov said, currently there is no private API in the router service, and Ideally it stays that way (basically as more of an interface).

This has the practical effect that at least (for now) overridden methods should eventually call `super` in order to have actual routing things take effect, minimizing the potential for folks to get too ---creative--- with what they might do as the signature and intent will have to remain aligned.

text/0819-public-router-service.md
```diff
@@ -0,0 +1,242 @@
...
+These proposed changes would further allow `ember-engines` to deprecate the `<LinkToExternal />`
+component provided by `ember-engines` package. As a result, there will be less
+engine specific concepts and code would be more portable between regular applications and engines.
+
+Once this RFC gets implemented, following will be possible in user defined Ember Engines:
+
+* links to external routes in templates within the engine rendered via `<LinkTo />` component:
+
+  ```hbs
+  <LinkTo @route="external-route">Go Somewhere</LinkTo>
```
villander, line 51: I'm still thinking that we should use `<LinkTo @route="external.home">Go Somewhere</LinkTo>` where `external` (hypothetical name) means a prefix that it's an external route. It's good for ergonomics because as a developer we don't want to go to externalRoutes definition and check if it's a internal or external route. I see the importance of deprecating `<LinkToExternal/>` but I don't want to lose the readability that it means an external link without poking around the code. 

Also it solve the drawback on this RFC 😄

text/0819-public-router-service.md
```diff
@@ -0,0 +1,242 @@
...
+  As this may be considered temporary solution to solve Ember 4 compatibility for `ember-engines`,
+  this is not viable long term solution as any changes made to `<LinkTo />` component in the Ember.js codebase
+  will need to be backported to `@ember/legacy-built-in-components`.
+
+* Re-implement `<LinkTo />` functionality in `ember-engine`.
+
+  This is not viable long term solution as any changes made to `<LinkTo />` component in the Ember.js codebase
+  would need to be backported to `ember-engines`.
+
+* Deprecate and remove `externalRoutes` mapping.
```
villander, line 222: 🚨 It breaks the isolation principle around ember-engines, where every engine can access any outside contexts, it going in the opposite direction of code isolation.

text/0819-public-router-service.md
```diff
@@ -0,0 +1,242 @@
...
+## How we teach this
+
+Ember Engines documentation provided at [ember-engines.com](https://ember-engines.com)
+should be updated with `<LinkTo />` replacing all the references of `<LinkToExternal />` component.
+
+Transition path needs to be published on [ember-engines.com](https://ember-engines.com) containing following steps:
+
+1. Replace use of `<LinkToExternal @route="external-route" />` component
+   with `<LinkTo @route="external-route" />` in every engine codebase.
+2. Users using `ember-engines-router-service` would need to replace any of its usage with
```
villander, line 179:
```suggestion
2. Users using [ember-engines-router-service](https://github.com/villander/ember-engines-router-service) would need to replace any of its usage with
```

text/0819-public-router-service.md
```diff
@@ -0,0 +1,242 @@
...
+  ```
+
+* using `RouterService` methods for redirects to the external routes:
+
+  ```javascript
+  # addon/components/my-component.js
+  export default class extends Component {
+    @service router;
+    @action clickLink () {
+      this.router.transitionTo('external-route');
```
villander, line 61: Also, I'd like to have something like that to don't lose the explicit outside context on engines 

```ts
import { external } from 'ember-engines/helper'
...
@action 
clickLink () {
   this.router.transitionTo(external('external-route'));
 }
```

villander (commented):
I really loved this RFC, thanks for doing that @SergeAstapov, but I think we need to keep the readability and isolation principle on top of our head to keep moving on in a direction to provide compatibility without losing ergonomy and purpose 😄

text/0819-public-router-service.md
runspired: @villander this would prevent addons from utilizing LinkTo, which I think makes it a non-starter. It might be interesting to bring the external-routes map into `routes.js` at some point to address this though so that there's only one location where valid routes may be defined. This would be outside the scope of this RFC.

text/0819-public-router-service.md
runspired: right, that's why though this is an alternative it's not the suggested path of the RFC

text/0819-public-router-service.md
runspired: It feels unnecessary to impose an extra function call on users to mark something as external when it has already been defined as external in their router configuration (see also comments above on `external.`)

text/0819-public-router-service.md
villander: Looking for this perspective I agree with you @runspired I think we just need to find some way to don't make engines more confusing and harder to teach than it already is.

text/0819-public-router-service.md
villander: It depends, I kinda agree with you that mapping those things on the router configuration clarifies things on a certain level, but I love readable and explicit code. How would you realize just looking for a ton of `transitionTo` in your code? imagine you have at least 50 occurrences how many are linking outside context? Correct me if I'm wrong, IMO is necessary back in forth on the router configuration to realize those things.

text/0819-public-router-service.md
villander: Surely, just wanted to highlight and provide a reminder/support to the next reviews 😃

chriskrycho:
Re: "no private API"—I want to be clear here about two things:

1. I don’t have a vested interest in the particular decision we make here! I just want to make sure we make it carefully and with clear understanding of what we're committing to. 😄 

2. Directly related: the fact that everything has been public *so far* has been uninteresting just insofar as it *has not* been subclass-able and therefore not overridable. Supporting *subclasses overriding methods* has very different semantic and maintenance implications than public *call-ability*. (This comes up in any kind of formal analysis of types, whether dynamic or static; [ask me how I know](https://semver-ts.org). 😂) In particular, public *call-ability* does not have the same implications for upholding internal-only invariants that *subclass-ability* does.

Net, I think that it's worth being very explicit and intentional about this question and I would actually like to see it addressed explicitly in the RFC. I'd also like to see some actual analysis of [the existing public API](https://api.emberjs.com/ember/4.3/classes/RouterService): do we *actually* want to say that it's fine for subclasses to override the definition of `currentURL`, for example? Or of `location`, or the timing and behavior of firing `routeWillChange` and `routeDidChange` events? Maybe, but that's not *obvious* to me!

Accordingly, I think it would be helpful to delineate, for each of these (perhaps at the category level if appropriate, but at a per-item level if that is more appropriate in other cases) whether it should be overridable by subclasses or not—

Methods:

- [isActive](https://api.emberjs.com/ember/4.3/classes/RouterService/methods/isActive?anchor=isActive)
- [recognize](https://api.emberjs.com/ember/4.3/classes/RouterService/methods/recognize?anchor=recognize)
- [recognizeAndLoad](https://api.emberjs.com/ember/4.3/classes/RouterService/methods/recognizeAndLoad?anchor=recognizeAndLoad)
- [refresh](https://api.emberjs.com/ember/4.3/classes/RouterService/methods/refresh?anchor=refresh)
- [replaceWith](https://api.emberjs.com/ember/4.3/classes/RouterService/methods/replaceWith?anchor=replaceWith)
- [transitionTo](https://api.emberjs.com/ember/4.3/classes/RouterService/methods/transitionTo?anchor=transitionTo)
- [urlFor](https://api.emberjs.com/ember/4.3/classes/RouterService/methods/urlFor?anchor=urlFor)

Properties:

- [currentRoute](https://api.emberjs.com/ember/4.3/classes/RouterService/properties/currentRoute?anchor=currentRoute)
- [currentRouteName](https://api.emberjs.com/ember/4.3/classes/RouterService/properties/currentRouteName?anchor=currentRouteName)
- [currentURL](https://api.emberjs.com/ember/4.3/classes/RouterService/properties/currentURL?anchor=currentURL)
- [location](https://api.emberjs.com/ember/4.3/classes/RouterService/properties/location?anchor=location)
- [rootURL](https://api.emberjs.com/ember/4.3/classes/RouterService/properties/rootURL?anchor=rootURL)

Events:

- [routeDidChange](https://api.emberjs.com/ember/4.3/classes/RouterService/events/routeDidChange?anchor=routeDidChange)
- [routeWillChange](https://api.emberjs.com/ember/4.3/classes/RouterService/events/routeWillChange?anchor=routeWillChange)

---

Bonus *doh* comment: it can’t actually use `private` in TS for the implementation details, because it *is* public for call-ability. But per both general good norms and per the SemVer for TS Types spec, it’s perfectly fine for us to *define* what the public API supports in terms of subclassing!

chancancode:
(FYI we didn't get to this agenda topic this week because #811 and #812 took up most of the time, will try again next week)

chancancode:
I don't have a ton of context on this, but maybe [this](https://github.com/emberjs/ember.js/issues/18422#issuecomment-534564808) will be important/useful to y'all since you have been thinking about it?

text/0819-public-router-service.md
```diff
@@ -0,0 +1,242 @@
...
+    }
+  }
+  ```
+
+**Bonus: improved DX for addons**
+
+This RFC seeks to provide a solution that also solves another DX issue faced by engines developers.
+
+Authors of large scale applications often create an addon(s) to contain reusable components that can
+be rendered either in the context of the host App or Engine. In such scenarios an extra guard needs
```
NullVoxPopuli, line 71: this is also solved by using hrefs instead of route/model/query args of the default link-to.
with an `a` tag with `preventDefault`d click handler, we can use the router service to detect when an href is in the route, or in an external app, and appropriately "do the right thing.

I like this approach because it means we:
 - don't need extra components for each type of usage
 - uses the existing href knowledge that everyone already understands with `<a>`
 - aligns with:  (the spirit of): https://github.com/emberjs/ember.js/pull/19271 and https://github.com/emberjs/rfcs/pull/391
 
 here is a demo: https://codesandbox.io/s/custom-link-component-dgbxl?file=/app/components/link.hbs
 (this demo is incomplete, but gives the gist).
 (except codesandbox doesn't work so... :shrug: (this sandbox has been untouched since 3.21, and the dependency checker is preventing start up... idk what's going on. ))
 
 where improvements could be made here where `<LinkTo>` is a bit better at:
  - query-param only transitions
  - sticky query params

NullVoxPopuli: though, unifying on LinkTo is still a good improvement!

runspired: This RFC is what would make those router helpers and the href approach work correctly for engines as well.

wagenet:
It looks like this has some interest so we should work to keep it moving.

villander:
Can we move this forward? 

@runspired @chriskrycho @wagenet @chancancode

achambers:
Can we move this to Exploring?

wagenet:
We had some back and forth in the Review meeting today around whether this should move to Exploring or not. There's definitely agreement that this is a problem we need to resolve, but that it's probably not going to be something that we can resolve until the new router design is more finalized.

knownasilya:
Is this still in limbo?

PR #821, opened 2022-05-24
RFC: public types for Owner, Transition, RouteInfo

chriskrycho:
From the summary:

> Introduce public import locations for type-only imports which have previously had no imports, and fully specify their public APIs for end users:
> 
> - `Owner`, with `TypeOptions` and `FactoryManager`
> - `Transition`
> - `RouteInfo` and `RouteInfoWithAttributes`

[rendered RFC](https://github.com/emberjs/rfcs/blob/public-types/text/0821-public-types.md)

### DIFF
--- a/text/0821-public-types.md
+++ b/text/0821-public-types.md
@@ -322,7 +322,11 @@ These types need two updates to the existing API documentation:
 
   And:
 
-  > While the `RouteInfo` interface can be imported to name the type for use in your own code, it is not user-implementable. The only supported way to get a `RouteInfo` is using the `from` or `to` properties on an instance of [the `Transition` class](https://api.emberjs.com/ember/4.3/classes/Transition)
+  > While the `RouteInfo` interface can be imported to name the type for use in your own code, it is not user-implementable. The only supported way to get a `RouteInfo` is using a public API which provides it, including but not limited to:
+  >
+  > - the `from` or `to` properties on an instance of [the `Transition` class](https://api.emberjs.com/ember/4.3/classes/Transition)
+  > - the `currentRoute` property on [the `RouterService` class](https://api.emberjs.com/ember/4.4/classes/RouterService)
+  > - the `child` or `parent` properties on another `RouteInfo` instance, or as returned from the `find()` method on a `RouteInfo`
 
 
 ### Blog post
--- a/text/0821-public-types.md
+++ b/text/0821-public-types.md
@@ -41,7 +41,7 @@ Additionally, the lack of a public import or contract for the `Owner` interface 
 
 ### `Owner`
 
-`Owner` is a user-constructible interface, with an intentionally-minimal subset of the existing `Owner` API, aimed at what we *want* to support for `Owner` in the future:
+`Owner` is a **non-user-constructible** interface, with an intentionally-minimal subset of the existing `Owner` API, aimed at what we *want* to support for `Owner` in the future:
 
 ```ts
 export default interface Owner {
@@ -59,7 +59,12 @@ export default interface Owner {
 }
 ```
 
-It is the default import from a new module, `@ember/owner`. With it come two other new types: `RegisterOptions` and `Factory`.
+`Owner` is the default import from a new module, `@ember/owner`. With it come two other new types: `RegisterOptions` and `Factory`.
+
+`Owner` is non-user-constructible because constructing it correctly also requires the ability to provide a factory manager.[^existing-owner-usage]
+
+[^existing-owner-usage]: Existing usage of the Owner interface this way (e.g. setting custom owners for tests) mostly falls under the "intimate API" rules, and will likely be deprecated after a future introduction of createOwner() hook so that that there is a public API way to get the required type.
+
 
 #### `RegisterOptions`
 
--- a/text/0821-public-types.md
+++ b/text/0821-public-types.md
@@ -215,6 +215,19 @@ function takesTransition(theTransition) {
 }
 ```
 
+#### `getOwner` and `setOwner`
+
+Both of the existing `getOwner` and `setOwner` functions now make *much* more sense as named exports from `@ember/owner`. They will also now take `owner` as the type of their argument explicitly:
+
+```ts
+export function getOwner(object): Owner | undefined;
+export function setOwner(object: unknown, owner: Owner): void;
+```
+
+The existing exports from `@ember/application` will become re-exports of these functions. The existing exports will also be deprecated, with the deprecation becoming "Available" no earlier than the first minor *after* an LTS containing the updated import location. For example, if `getOwner` and `setOwner` are available to import from `@ember/owner` in Ember v4.7, the deprecation for the imports from `@ember/application` would not be deprecated until at least Ember v4.9, after Ember v4.8 LTS.[^timeline]
+
+[^timeline]: This is in line with our normal approach to deprecation: it allows the addon ecosystem to absorb the change via LTS support releases, so that consuming apps are not flooded with deprecations without recourse.
+
 
 ### `RouteInfo`
 
--- a/text/0821-public-types.md
+++ b/text/0821-public-types.md
@@ -19,7 +19,7 @@ RFC PR: https://github.com/emberjs/rfcs/pull/821
 
 Introduce public import locations for type-only imports which have previously had no imports, and fully specify their public APIs for end users:
 
-- `Owner`, with `TypeOptions` and `FactoryManager`
+- `Owner`, with `TypeOptions` and `Factory`
 - `Transition`
 - `RouteInfo` and `RouteInfoWithAttributes`
 
@@ -49,7 +49,7 @@ export default interface Owner {
 
   register<T>(
     fullName: string,
-    factory: FactoryManager<T>,
+    factory: Factory<T>,
     options?: TypeOptions
   ): void;
 
@@ -57,18 +57,18 @@ export default interface Owner {
 
   resolveRegistration(
     fullName: string
-  ): FactoryManager<unknown> | object | undefined;
+  ): Factory<unknown> | object | undefined;
 }
 ```
 
-It is the default import from a new module, `@ember/owner`. With it come two other new types: `TypeOptions` and `FactoryManager`.
+It is the default import from a new module, `@ember/owner`. With it come two other new types: `TypeOptions` and `Factory`.
 
-#### `TypeOptions`
+#### `RegisterOptions`
 
-`TypeOptions` is a user-constructible interface:
+`RegisterOptions` is a user-constructible interface:
 
 ```ts
-export interface TypeOptions {
+export interface RegisterOptions {
   instantiate?: boolean | undefined;
   singleton?: boolean | undefined;
 }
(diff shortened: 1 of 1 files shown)


text/0821-public-types.md
```diff
@@ -0,0 +1,358 @@
...
+These types need two updates to the existing API documentation:
+
+- Specify the modules they can be imported from (as noted above).
+- Specify that these types are *not* meant to be implemented by end users. For example:
+
+  > While the `Transition` interface can be imported to name the type for use in your own code, it is not user-implementable. The only supported way to get a `Transition` is by using Ember's routing APIs, like [the `transitionTo` method](https://api.emberjs.com/ember/4.3/classes/RouterService/methods/transitionTo?anchor=transitionTo) on [the router service](https://api.emberjs.com/ember/4.3/classes/RouterService).
+
+  And:
+
+  > While the `RouteInfo` interface can be imported to name the type for use in your own code, it is not user-implementable. The only supported way to get a `RouteInfo` is from an of [the `Transition` class](https://api.emberjs.com/ember/4.3/classes/Transition).
```
MrChocolatine, line 308: > The only supported way to get a `RouteInfo` is from an of the `Transition` class

Missing words.

Did you mean:
> The only supported way to get a `RouteInfo` is by using `from` and `to` of the `Transition` class

MrChocolatine (commented):
Awesome, thanks @chriskrycho.

text/0821-public-types.md
```diff
@@ -0,0 +1,358 @@
...
+
+  hasRegistration(fullName: string): boolean;
+
+  resolveRegistration(
+    fullName: string
+  ): FactoryManager<unknown> | object | undefined;
+}
+```
+
+It is the default import from a new module, `@ember/owner`. With it come two other new types: `TypeOptions` and `FactoryManager`.
```
SergeAstapov, line 64: @chriskrycho out of curiosity, why not the existing `@ember/application` module which provides `getOwner`/`setOwner` functions?

chriskrycho: A good question! The reason is twofold:

1. Historically we've tried to treat a single "major type" as the responsibility of a given module, and supplied it as the default export from that module, and in fact this is explicitly defined as a design constraint in #176.

2. `@ember/application` supplies `Application` and `ApplicationInstance`, but those are themselves subclasses of `Engine` and `EngineInstance` respectively, and `Engine` and `EngineInstance` *implement* the `Owner` interface, so that would be a very strange module relationship (and would provide some complications for our build tooling to disentangle too).

That said, I had forgotten that `getOwner` and `setOwner` come from `@ember/application`. If we are introducing `@ember/owner`, maybe we should move `getOwner` and `setOwner` there, and make the versions in `@ember/application` (deprecated) re-exports?

SergeAstapov: > That said, I had forgotten that `getOwner` and `setOwner` come from `@ember/application`. If we are introducing `@ember/owner`, maybe we should move `getOwner` and `setOwner` there, and make the versions in `@ember/application` (deprecated) re-exports?

that's interesting option, I kinda like it as we could separate owner vs. app/engine concepts and make them come from different modules

text/0821-public-types.md
```diff
@@ -0,0 +1,358 @@
...
+
+[0274]: https://rfcs.emberjs.com/id/0724-road-to-typescript
+[0800]: https://rfcs.emberjs.com/id/0800-ts-adoption-plan
+[spec]: https://www.semver-ts.org
+
+Additionally, the lack of a public import or contract for the `Owner` interface has been a long-standing problem for *all* users, but especially TypeScript users, and given the APIs provided for e.g. the Glimmer Component class where `owner` is the first argument, the pervasive use of `getOwner` in low-level library code, etc., it is important for TypeScript users to be able to use an `Owner` safely, and for JavaScript users to be able to get autocomplete etc. from the types we ship.
```
wagenet, line 35: `Owner` is the optional first argument for `EmberObject` as well.

text/0821-public-types.md
```diff
@@ -0,0 +1,358 @@
...
+
+  resolveRegistration(
+    fullName: string
+  ): FactoryManager<unknown> | object | undefined;
+}
+```
+
+It is the default import from a new module, `@ember/owner`. With it come two other new types: `TypeOptions` and `FactoryManager`.
+
+#### `TypeOptions`
```
wagenet, line 66: Is this actually a good name?

text/0821-public-types.md
```diff
@@ -0,0 +1,358 @@
...
+## Detailed design
+
+**Note:** For the terms "user-constructible" and "user-subclassable" see the [Semantic Versioning for TypeScript Types spec][spec].
+
+### `Owner`
+
+`Owner` is a user-constructible interface, with an intentionally-minimal subset of the :
+
+```ts
+export default interface Owner {
```
rwjblue, line 47: Should also include `factoryFor`

text/0821-public-types.md
```diff
@@ -0,0 +1,358 @@
...
+    fullName: string,
+    factory: FactoryManager<T>,
+    options?: TypeOptions
+  ): void;
+
+  hasRegistration(fullName: string): boolean;
+
+  resolveRegistration(
+    fullName: string
+  ): FactoryManager<unknown> | object | undefined;
```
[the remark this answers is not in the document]
chriskrycho: Excellent; will swap those!

text/0821-public-types.md
```diff
@@ -0,0 +1,358 @@
...
+
+
+#### `FactoryManager`
+
+`FactoryManager` is an existing concept available to users via [the `Engine#factoryFor` API][ff].[^factory-name] The public API to date has included only two fields, `class` and `create`, and we maintain that in this RFC. The result is this user-constructible interface:
+
+[ff]: https://api.emberjs.com/ember/4.3/classes/EngineInstance/methods/factoryFor?anchor=factoryFor
+
+```ts
+export interface FactoryManager<Class> {
```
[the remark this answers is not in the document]
chriskrycho: I'm *entirely* fine with going back to `Factory` here, FWIW; we should make a point to update the docs for `factoryFor` accordingly! Will push an update to that effect.

text/0821-public-types.md
chriskrycho: This is just me, but if it comes from the `@ember/owner` module/package, my thought is that we should just call it `RegisterOptions` and let the inherent namespace effect of modules do its job (and if folks want to explicitly name it that way they can). There are (at least per ripgrep!) no *other* `RegisterOptions` anywhere within Ember so we don't even have the possible auto-import overlap issue there.

rwjblue: Sounds good (and still better than `TypeOptions`, which sounds way too generic)

rwjblue:
@chriskrycho - How can we ensure that folks don't try to leverage these new (type only) imports in a runtime capacity? I don't see that being addressed...

text/0821-public-types.md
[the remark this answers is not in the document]
chriskrycho: Updated to use `Factory` in c71d685!

chriskrycho:
@rwjblue it's called out as part of the definition as well as in the **How We Teach This** section for `Transition` and `RouteInfo`; see [here](https://github.com/emberjs/rfcs/blob/a7cc41f7accee0c08a4cbbd1904dd2673819c559/text/0821-public-types.md#transition-routeinfo-and-routeinfowithattributes). We will also want to double down on that idea as part of the blog post called for in that section, since it's a new concept to not only the Ember community but the JS community at large.

text/0821-public-types.md
chriskrycho: Updated in a7cc41f to move them and specify when and how we'll deprecate the re-exports.

text/0821-public-types.md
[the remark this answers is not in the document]
chriskrycho: @rwjblue this is almost as bad as `Engine` and `EngineInstance` and their `Application` friends: it turns out that we have `FactoryManager` *extensively* throughout, and that there are *two* types in play:

- `Factory`, which is [what `register` takes](https://github.com/emberjs/ember.js/blob/master/packages/%40ember/-internals/runtime/lib/mixins/registry_proxy.ts#L17) and also [what `lookup` (may) return](https://github.com/emberjs/ember.js/blob/master/packages/%40ember/-internals/runtime/lib/mixins/container_proxy.ts#L15) (though in the latter case, not in a way we can currently provide useful types for 😭—it just has to be `unknown`)
- `FactoryManager`, which is what `factoryFor` returns (exceedingly confusingly in my opinion!)

Why in the world `factoryFor` returns the thing which manages factories for things is likely a :sob: moment in the depths of time, but that's what it actually is. I'm open to names for the latter, though `FactoryForResult` is 😬 to me.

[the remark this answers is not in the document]
chriskrycho: I'm actually going to call it `FactoryFor`! It is indeed a type which relates two other types, so that name makes reasonably good sense here.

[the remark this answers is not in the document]
chriskrycho: …okay, having read the implementation *and* the docs repeatedly, I now: have literally no confidence that my idea of how these types are related or how they are supposed to be used is correct. [`Factory`](https://github.com/emberjs/ember.js/blob/55ffe6326f11efcaeb278cdf7e0f86543daa9f04/packages/%40ember/-internals/owner/index.ts#L16-L22) and [`FactoryManager`](https://github.com/emberjs/ember.js/blob/master/packages/%40ember/-internals/container/lib/container.ts#L462) have *very* similar APIs, but `FactoryManager#class` is a `Factory` whereas `Factory#class` is a "class" which can either be any old object (?) or specifically a [`FactoryClass`](https://github.com/emberjs/ember.js/blob/55ffe6326f11efcaeb278cdf7e0f86543daa9f04/packages/%40ember/-internals/owner/index.ts#L12-L14), which is just any object with `positionalParams` of a certain type?

Writing this down for now and will come back to trying to actually get something reasonable after discussing with folks who know this more (@rwjblue, @wagenet, etc.) tomorrow. 😂

text/0821-public-types.md
```diff
@@ -0,0 +1,375 @@
...
+
+- Specify the modules they can be imported from (as noted above).
+
+- Specify that these types are *not* meant to be implemented by end users. For example:
+
+  > While the `Transition` interface can be imported to name the type for use in your own code, it is not user-implementable. The only supported way to get a `Transition` is by using Ember's routing APIs, like [the `transitionTo` method](https://api.emberjs.com/ember/4.3/classes/RouterService/methods/transitionTo?anchor=transitionTo) on [the router service](https://api.emberjs.com/ember/4.3/classes/RouterService).
+
+  And:
+
+  > While the `RouteInfo` interface can be imported to name the type for use in your own code, it is not user-implementable. The only supported way to get a `RouteInfo` is using the `from` or `to` properties on an instance of [the `Transition` class](https://api.emberjs.com/ember/4.3/classes/Transition)
```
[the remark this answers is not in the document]
chriskrycho: Good call! Fixed in ca1328d.

text/0821-public-types.md
```diff
@@ -0,0 +1,375 @@
...
+`Owner` is a user-constructible interface, with an intentionally-minimal subset of the :
+
+```ts
+export default interface Owner {
+  lookup(name: string): unknown;
+
+  register<T>(
+    fullName: string,
+    factory: Factory<T>,
+    options?: TypeOptions
```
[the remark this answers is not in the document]
chriskrycho: Thanks! 21ccda4

chriskrycho:
I have just pushed, in 5b9bc9c, a proposed name for the thing currently named `FactoryManager` in Ember's internal types: `FactoryFor`. It now at least has a well-defined type… and we can bike-shed the name, because I *hate* that name!

chriskrycho:
We talked about this at today's Framework meeting and agreed that it's ready for FCP, and the Typed Ember team agrees, so this RFC is now in FCP!

text/0821-public-types.md
```diff
@@ -0,0 +1,412 @@
...
+  lookup(name: string): unknown;
+
+  register<T>(
+    fullName: string,
+    factory: Factory<T>,
+    options?: RegisterOptions
+  ): void;
+
+  hasRegistration(fullName: string): boolean;
+
```
chriskrycho, line 54: The *final* comment is: this is redundant with `factoryFor`.

```suggestion
```

chriskrycho:
We talked about this at today's Framework team and are merging it. Types will be updated on DefinitelyTyped ASAP!

PR #833, opened 2022-07-29
[RFCs App Rendering] Reformat frontmatter

mansona:
This is part of the work that we are doing to improve the rendering and tracking experience for RFCs while we're building https://rfcs.emberjs.com/ 

As part of our build step we have to "[process the RFCs](https://github.com/ember-learn/rfcs-app/blob/main/process-rfcs.js)" because of inconsistent data, and to make it more computer-readable which allows us to add new app features. If we can merge this PR then we would not need to process the data any more 🎉


.github/workflows/ci.yml
```diff
@@ -12,32 +12,16 @@ jobs:
       - uses: actions/checkout@v2
 
       - name: Checkout tools repo
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
         with:
-          repository: emberjs/rfcs-tooling
+          repository: mansona/emberjs-rfcs-tooling
+          ref: new-format-frontmatter
```
mansona, line 18: I can drop this change once https://github.com/emberjs/rfcs-tooling/pull/4 is merged 👍

0000-template.md
```diff
@@ -1,23 +1,31 @@
...
-RFC PR: 
+teams: # delete teams that aren't relevant
+  - cli
+  - data
+  - framework
+  - learning
+  - steering
+  - typescript
+proposal-pr:
+tracking-link:
```
kategengler, line 16: Why add back tracking link? In theory it should be unnecessary because the stages (especially 'Ready for Release') are the tracking.

text/0001-transform-attribute-meta-parameter.md
```diff
@@ -1,8 +1,11 @@
 ---
-Start Date: 2014-08-14
-RFC PR: https://github.com/emberjs/rfcs/pull/1
```
kategengler, line 3: Losing some historical data on the left here, where other fields were filled out.

0000-template.md
```diff
@@ -1,23 +1,31 @@
...
-Relevant Team(s): 
-RFC PR: 
+teams: # delete teams that aren't relevant
+  - cli
+  - data
+  - framework
+  - learning
+  - steering
+  - typescript
+proposal-pr:
```
kategengler, line 15: I think we will probably want to have a bit of metadata for each stage's PR

kategengler (changes requested):
A lot of historical data is lost in this conversion -- not sure how much we should care, but it should be reasonably easy to leave unknown fields (I believe my version of a conversion script did so). Also the previous `Tracking` field isn't even making it into `tracking-link`

0000-template.md
wagenet: I believe the idea here is when an RFC is merged this would point to one of the following:
- Quest Issue
- PR
- Github Project

In any of these cases the RFCs app would be able to use the Github API to check completion progress.

kategengler: Ah, my understanding of the Stages RFC was the goal to replace the `Tracking link` with tracking of the progress done on the PR to move the RFC from `Accepted` to `Ready for Release`

kategengler: If we want to keep, I think we should make it an array field so there can be many links

wagenet: I'm somewhat flexible on this and open to whatever will ultimately work best.

text/0001-transform-attribute-meta-parameter.md
mansona: so this idea of an "Ember Issue" doesn't look like something that we do much nowadays right? How would you feel about keeping the information about the issues but just in the body of the RFC rather than exposing it in the frontmatter? I'm just concerned about supporting something that we aren't using any more

0000-template.md
mansona: So this is a bit of a crossover between two efforts that are going on here. There was the RFC Stages RFC that talks about the PRs to move from one stage to the next but there is also the Ember Learning Initiative that I proposed (way back in 2019) to communicate the progress of RFCs using an "app". Imagine the next phase of https://rfcs.emberjs.com 

The idea of having a tracking link here is to have something that is keeping track of the next steps, as @wagenet mentions I have in mind that it could be an Issue a PR or a Github project depending on the preference of the person/team that is implementing the RFC. With a single tracking link we are able to use this information and show it on the app as some sort of progress-bar percentage (or something) which is one of the key goals of what I'm trying to do with my Initiative.

I hope this makes sense, let me know if you have any more questions and I can try to explain my point of view a bit better 👍

kategengler: We previously had tracking links and used the RFC Tracking Repo. Stages is supposed to be a replacement of sort that would require the teams to update it to move the RFC forward -- the various links and status would be collected in the PR to the `Ready for Release` stage. https://github.com/emberjs/rfcs/blob/master/text/0617-rfc-stages.md#changes-to-rfc-meta

text/0001-transform-attribute-meta-parameter.md
wagenet: This might be a bad idea, but to keep this moving forward, could we stick this in an "additional-data" section?

kategengler: I think that would be great -- I just don't want to lose what we put there.

wagenet: Basically, we just need to mark them as non-standard fields. The UI could choose to show them in some generic way if we wanted, but I agree about not wanting to lose information.

0000-template.md
wagenet: Good point!

0000-template.md
mansona: I have updated this based on the Discord discussion. Now we have a `prs` object that will have keys that will each point to a PR for advancement to that stage. I have added a `project-link` as a compromise to allow projects to track progress in the website 👍

0000-template.md
mansona: this is now done 👍

text/0001-transform-attribute-meta-parameter.md
mansona: this is now fixed. I've methodically gone through every single field (with my little script) and made sure that it is represented in a `meta` object (essentially just stuff that we will ignore but we're keeping for reference) 👍

.github/workflows/ci.yml
mansona: this has been dropped 👍

0000-template.md
```diff
@@ -1,23 +1,30 @@
...
-Release Versions: Leave as is
-Relevant Team(s): Fill this in with the [team(s)](README.md#relevant-teams) to which this RFC applies
-RFC PR: Fill this in with the URL for the Proposal RFC PR
+stage: Leave as is
+start-date: Fill in with today's date, 2032-12-01T00:00:00.000Z
+release-date: Leave as is
+release-versions: Leave as is
+teams: Include only the [team(s)](README.md#relevant-teams) for which this RFC applies
+proposal-pr: Fill this in with the URL for the Proposal RFC PR
+tracking-link: Leave as is
```
kategengler, line 27: this has been removed or renamed

0000-template.md
```diff
@@ -1,23 +1,30 @@
...
 Directions for above: 
 
-Stage: Leave as is
-Start Date: Fill in with today's date, YYYY-MM-DD
-Release Date: Leave as is
-Release Versions: Leave as is
-Relevant Team(s): Fill this in with the [team(s)](README.md#relevant-teams) to which this RFC applies
-RFC PR: Fill this in with the URL for the Proposal RFC PR
+stage: Leave as is
+start-date: Fill in with today's date, 2032-12-01T00:00:00.000Z
```
kategengler, line 22: The linter doesn't require the time, maybe we shouldn't?

text/0001-transform-attribute-meta-parameter.md
```diff
@@ -1,8 +1,15 @@
...
-Ember Issue: https://github.com/emberjs/data/pull/4086
-
+start-date: 2014-08-14T00:00:00.000Z
```
kategengler, line 2: This metadata order is different than the template, do we care? I do find it less readable

0000-template.md
```diff
@@ -1,23 +1,30 @@
...
+teams: # delete teams that aren't relevant
+  - cli
+  - data
+  - framework
+  - learning
+  - steering
+  - typescript
+prs:
+  accepted: # update this to the PR that you propose your RFC in
+project-link:
```
kategengler, line 15: Why not add later? This has no spec and no instructions

0000-template.md
mansona: sorry I missed the comment block directly inside the content of the markdown 🙈 I'll fix this now

0000-template.md
mansona: We discussed this a bit in the Discord chat, essentially we want to mark those dates as UTC to prevent a whole host of issues down the line. I was under the impression that we concluded this wasn't a blocker 🤔 

If the issue is that the linter doesn't require this format are you asking me to add this to the linter?

text/0001-transform-attribute-meta-parameter.md
mansona: I don't see why metadata order matters 🤔 and what do you mean by "readable" in this context? Really we will be hoping to read most RFCs in the context of the app https://rfcs.emberjs.com/id/0001-transform-attribute-meta-parameter

right now that should be the default for all merged RFCs and in the future it will work for proposed ones too 👍

kategengler: Ok -- I'll probably always going to files, but so long as the order stays as it is in the template for new RFCs (and thus ones we will be editing), it seems fine.

0000-template.md
kategengler: If you want it to always have the time part, you should definitely add it to the linter. I'm sure most authors will write `YYYY-MM-DD` no matter what is in instructions, otherwise.

wagenet (approved):

0000-template.md
wycats: Out of curiosity, what's a concrete example of a problem that would arise in this app, and which we would avoid by using this format?

PR #847, opened 2022-08-26
Deprecate support for `ember-cli-qunit` and `ember-cli-mocha` when generating test blueprints

bertdeblock:
[Rendered RFC](https://github.com/bertdeblock/rfcs/blob/deprecate-support-for-ember-cli-qunit-and-ember-cli-mocha-when-generating-test-blueprints/text/0847-deprecate-support-for-ember-cli-qunit-and-ember-cli-mocha-when-generating-test-blueprints.md)

### DIFF
--- a/text/0847-deprecate-support-for-ember-cli-qunit-and-ember-cli-mocha-when-generating-test-blueprints.md
+++ b/text/0847-deprecate-support-for-ember-cli-qunit-and-ember-cli-mocha-when-generating-test-blueprints.md
@@ -1,0 +1,114 @@
+---
+stage: accepted
+start-date: 2022-08-26T00:00:00.000Z
+release-date:
+release-versions:
+teams: # delete teams that aren't relevant
+  - cli
+  - data
+  - framework
+prs:
+  accepted: https://github.com/emberjs/rfcs/pull/847
+project-link:
+---
+
+<!---
+Directions for above:
+
+stage: Leave as is
+start-date: Fill in with today's date, 2032-12-01T00:00:00.000Z
+release-date: Leave as is
+release-versions: Leave as is
+teams: Include only the [team(s)](README.md#relevant-teams) for which this RFC applies
+prs:
+  accepted: Fill this in with the URL for the Proposal RFC PR
+project-link: Leave as is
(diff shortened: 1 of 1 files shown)


bertdeblock:
It seems we also generate different test files depending on the `ember-mocha` version installed in the app:
https://github.com/emberjs/ember.js/blob/master/blueprints/test-framework-detector.js#L31-L41

Should we also deprecate support for `ember-mocha` versions below `v0.14.0`?

What's the status actually of `ember-mocha`? Development seems to have stalled.

ef4:
This seems fine and uncontroversial.

kategengler:
This was discussed in today's ember-cli meeting and we have decided to move it into Final Comment Period.

text/0847-deprecate-support-for-ember-cli-qunit-and-ember-cli-mocha-when-generating-test-blueprints.md
```diff
@@ -0,0 +1,114 @@
+---
+stage: accepted
+start-date: 2022-08-26T00:00:00.000Z
+release-date:
+release-versions:
+teams: # delete teams that aren't relevant
+  - cli
+  - data
+  - framework
```
wagenet, line 9: This needs to be updated, right?

kategengler: I believe this list is correct in this case. There are blueprints in both data and cli that will be affected.

runspired: confirm, which is why I shipped it. Consider my review to be data's ack that blueprint changes are to be made.

wagenet: 👍

kategengler:
The framework team discussed this today (with a large subset of ember-cli team) and we've decided to merge since no new discussion has happened since FCP.

PR #849, opened 2022-08-30
data: updates to the polymorphic relationships rfc

runspired:
[Rendered](https://github.com/emberjs/rfcs/blob/9de548960294e63d63f0b56e5a5fd2eeafea380a/text/0793-polymporphic-relations-without-inheritance.md)

While working on implementing polymorphic relationship support we discovered both that the restriction on polymorphic-to-polymorphic was unnecessary *and* that clarification of how to avoid needing to implement the abstract-type class was required. This updates the RFC text to reflect these things.

cc @yratanov @snewcomer

### DIFF
--- a/text/0793-polymporphic-relations-without-inheritance.md
+++ b/text/0793-polymporphic-relations-without-inheritance.md
@@ -84,22 +84,20 @@ This RFC proposes to allow explicitly declaring polymorphic traits similar to [r
 
 ### API
 
-The inverse relationship, if one exists, of a polymorphic relationship must be concrete (e.g. not also polymorphic).
-
 The polymorphic relationship **MUST** explicitly declare itself as polymorphic and **MUST** explicitly declare it's inverse as either a key or `null`. All records that satisfy the polymorphic type must declare a matching inverse via that key.
 
 Polymorphic relationships take the form:
 
 ```ts
-hasMany(baseType: string, options: { polymorphic: true, inverse: string | null })
-belongsTo(baseType: string, options: { polymorphic: true, inverse: string | null })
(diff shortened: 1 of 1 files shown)


text/0793-polymporphic-relations-without-inheritance.md
```diff
@@ -84,59 +84,149 @@ This RFC proposes to allow explicitly declaring polymorphic traits similar to [r
...
-In the below example, the polymporphic relationship implies the existence of the `base type` "taggable".
+In the below example, the polymporphic relationship implies the existence of the `abstract type` "taggable".
 
+```ts
+class Tag extends Model {
+  @hasMany("taggable", { async: false, inverse: "tags", polymorphic: true }) taggables;
+}
 ```
+
+It is only required that your app implement a model (or if using instantiateRecord and the schema service implement handling for that type) if your API will return entities whose type is the abstract type. If the abstract type is never directly interacted with, it is not required to implement a Model or other associated support.
```
yratanov, line 132: > If the abstract type is never directly interacted with

What does it mean?

runspired: This is actually carry over from the original rfc (I cleaned it up to refer to abstract vs base as it was more clear)

what it means is that if Post implements Taggable but no attempt to fetch something with type taggable and no data received will have type taggable then it’s never necessary to have a Model named Taggable

yratanov: super! Thank you for clarification!

ef4:
This looks ready to merge. Core framework looked it over briefly but I think it's up to the data team to merge it.

PR #853, opened 2022-09-22
Amend RFC 0821: add Resolver public type

chriskrycho:
This was missed during the original work on [0821: Public API for Type-Only Imports](https://rfcs.emberjs.com/id/0821-public-types/), and caught while doing the work for emberjs/ember.js#20180. At last week’s Framework Core Team meeting, we agreed that it makes the most sense to add as an amendment to 821.

---

As a meta note, we expect to see this more often as we move to following [RFC 0617: RFC Stages](https://rfcs.emberjs.com/id/0617-rfc-stages). 0821 is currently *accepted* and will become *released* once we finish implementing it.

### DIFF
--- a/text/0821-public-types.md
+++ b/text/0821-public-types.md
@@ -395,23 +395,28 @@ function takesRouteInfoWithAttributes(routeInfoWithAttributes) {
 
 ### `Resolver`
 
-The resolver is a contract implemented by libraries outside Ember itself, such as `ember-resolver`, `ember-strict-resolver`, and any number of custom resolvers which exist in apps across the ecosystem. It has never had public documentation, but is fully public API. It is a user-constructible interface with the following definition (using the `Factory` and `FullName` types exported from the new `@ember/owner` module), available as a new named type-only export from the `@ember/engine` package:
+The resolver is a contract implemented by libraries outside Ember itself, such as `ember-resolver`, `ember-strict-resolver`, and any number of custom resolvers which exist in apps across the ecosystem. It has never had public documentation, but is fully public API. It is a user-constructible interface with the following definition (using the `Factory` and `FullName` types exported from the new `@ember/owner` module):
 
 ```ts
-import EmberObject from '@ember/object';
-import type { Factory, FullName } from '@ember/owner';
-
-type KnownForTypeResult<Name extends string> = {
+export type KnownForTypeResult<Name extends string> = {
   [fullName in `${Name}:${string}`]: boolean | undefined;
 };
 
-export interface Resolver extends EmberObject {
+export interface Resolver {
   knownForType?: <Name extends string>(type: Name) => KnownForTypeResult<Name>;
   lookupDescription?: (fullName: FullName) => string;
   makeToString?: (factory: Factory<object>, fullName: FullName) => string;
   normalize?: (fullName: FullName) => string;
   resolve(name: string): Factory<object> | object | undefined;
 }
+```
+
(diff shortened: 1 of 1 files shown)


wagenet (approved):

text/0821-public-types.md
```diff
@@ -392,6 +393,28 @@ function takesRouteInfoWithAttributes(routeInfoWithAttributes) {
...
+
+```ts
+import EmberObject from '@ember/object';
+import type { Factory, FullName } from '@ember/owner';
+
+type KnownForTypeResult<Name extends string> = {
+  [fullName in `${Name}:${string}`]: boolean | undefined;
+};
+
+export interface Resolver extends EmberObject {
```
dfreeman, line 408: Does a `Resolver` implementation have to be an `EmberObject`, or do the existing implementations just happen to be?

text/0821-public-types.md
```diff
@@ -392,6 +393,28 @@ function takesRouteInfoWithAttributes(routeInfoWithAttributes) {
 ```
 
 
+### `Resolver`
+
+The resolver is a contract implemented by libraries outside Ember itself, such as `ember-resolver`, `ember-strict-resolver`, and any number of custom resolvers which exist in apps across the ecosystem. It has never had public documentation, but is fully public API. It is a user-constructible interface with the following definition (using the `Factory` and `FullName` types exported from the new `@ember/owner` module), available as a new named type-only export from the `@ember/engine` package:
```
dfreeman, line 398: Is the thinking that since engines are the base entity that accept a resolver, that's the package the type should come from? My kneejerk reaction is that the concept feels more closely tied to `@ember/owner`, but my mental model may be off.

dfreeman (commented):
A couple of questions, but more for my own edification than anything. I didn't know 3/5 of those hooks existed!

text/0821-public-types.md
chriskrycho: Mmm, good catch. I don’t *think* it has to be, and I made the mistake of pulling this over from the existing `ember-resolver` types—that one is, of course, but I don’t think the contract does or should require it. Will update!

text/0821-public-types.md
chriskrycho: That was my own kneejerk reaction as well. I picked `@ember/engine` only because it’s the place that it’s associated with and I usually group “support” modules with their types. I could see an argument for it coming from `@ember/owner` or even `@ember/owner/resolver`, too—I don’t have a strong preference here and so just picked a reasonable-ish place as a starting point for exactly this discussion.

dfreeman: In terms of the public API, it is the engine (or application) that's getting constructed with a `Resolver`, so I definitely see where you're coming from. I think what's throwing me off is the knowledge that, under the covers, the key purpose of the resolver is to power the `Owner` (which, concretely, is an engine/application instance; but abstractly we've decided that's its own concept in `@ember/owner`).

Ultimately I don't feel strongly about it, just wanted to make sure I understood the reasoning

chriskrycho:
We talked about this at Framework Core today, and everybody agreed with @dfreeman's comments above (and my intuition): it really ought to be in `@ember/owner`. I'm moving it there, and then putting it into FCP. 🎉

text/0821-public-types.md
chriskrycho: I brought this up with folks at Framework today, and they agreed with our intuition. 👍🏼

chriskrycho:
Meant to merge this a few weeks ago!

PR #856, opened 2022-10-08
Default workspace settings of code editors

kaermorchen:
Default workspace settings will help new developers to start working on Ember projects.

[Rendered](https://github.com/kaermorchen/rfcs/blob/default-workspace-settings/text/0000-default-workspace-settings.md)


Windvis:
I think it would be better to look at something like [`ember-apply`](https://github.com/NullVoxPopuli/ember-apply) instead. Not everyone in a team will use VS Code and I really dislike that these kind of files are part of a project's repo. 

Adding more conditional complexity to ember-cli itself increases the maintenance burden on the cli team and any updates to these files are bound to the ember-cli release cycle as well. It feels safer to keep it separate so each team can tweak and ship its own default configs if needed.

MrChocolatine:
> I think it would be better to look at something like [`ember-apply`](https://github.com/NullVoxPopuli/ember-apply) instead. Not everyone in a team will use VS Code and I really dislike that these kind of files are part of a project's repo.
> 
> Adding more conditional complexity to ember-cli itself increases the maintenance burden on the cli team and any updates to these files are bound to the ember-cli release cycle as well. It feels safer to keep it separate so each team can tweak and ship its own default configs if needed.

I completely agree with what @Windvis said. In my opinion this should not be part of a JS framework.

jrjohnson:
These are great recommendations @kaermorchen, if you added them to `ember-apply` I'd recommend them for everyone on my team. I do tend to agree that while VSCode is super popular right now it's probably not a good idea to push this stuff into `ember-cli` as the most popular editor has changed at least 3x since I started working with ember 😁

kaermorchen:
Thanks @Windvis and all for the opinions and suggestions they are good for me.

I have taken a look at `ember-apply` and it seems a perfect solution for this issue.

PR #861, opened 2022-11-16
Update Glimmer Template Truthiness Semantics

bakerac4:
[Rendered](https://github.com/Gavant/rfcs/blob/master/0000-update-glimmer-template-truthiness-semantics.md)

Provide a path for Glimmer templates to switch away from classic Handlebars truthiness to JavaScript’s truthiness. This will align our templating language with users' expectations and make teaching easier. It will also allow TypeScript features like type narrowing to work as users would expect it to with keywords like `{{and}}` or `{{or}}`.


0000-update-glimmer-template-truthiness-semantics.md
```diff
@@ -0,0 +1,106 @@
...
+prs:
+  accepted: Fill this in with the URL for the Proposal RFC PR
+project-link: Leave as is
+-->
+
+# Update Glimmer Template Truthiness Semantics
+
+## Summary
+
+Update truthiness semantics from handlebar's version to javascript's version to allow for typescript to work as users would expect it to
```
chriskrycho, line 33:
```suggestion
Provide a path for Glimmer templates to switch away from classic Handlebars truthiness to JavaScript’s truthiness. This will align our templating language with users' expectations and make teaching easier. It will also allow TypeScript features like type narrowing to work as users would expect it to with keywords like `{{and}}` or `{{or}}`.
```

boris-petrov:
I'm confused, is this connected to [this](https://github.com/jmurphyau/ember-truth-helpers/issues/141) issue? Isn't updating these semantics going to break billions of apps?

chriskrycho:
1. No, because assuming we do this, we are going to have a migration plan that allows the community to migrate incrementally.
2. Billions of apps 🤣

0000-update-glimmer-template-truthiness-semantics.md
```diff
@@ -0,0 +1,106 @@
...
+## Summary
+
+Provide a path for Glimmer templates to switch away from classic Handlebars truthiness to JavaScript’s truthiness. This will align our templating language with users' expectations and make teaching easier. It will also allow TypeScript features like type narrowing to work as users would expect it to with keywords like `{{and}}` or `{{or}}`.
+
+## Motivation
+
+Due to handlebars supporting a different truthiness semantics than javascript, typescript cant function like users would expect it to. 
+By aligning HTMLBars version of truthiness semantics with the wider javascript community, type checking will function as users expect. 
+
+To give an example of why this is an issue we can use the `{{and}}` operator. Right now we have a choice without changing the implementation of truthiness semantics in our templates.
```
wagenet, line 40:
```suggestion
To give an example of why this is an issue we can use the `{{and}}` operator. Under the current Glimmer semantics, we can handle TypeScript in one of two ways:

File: 0000-update-glimmer-template-truthiness-semantics.md [hunk 3]
```diff
@@ -0,0 +1,106 @@
...
+1) `{{and}}` compiles to `&&` which will allow TS to lie to us about what is safe and what isn't (and will cause compile time errors) but WILL allow for type guards and narrowing to work
+2) `{{and}}` compiles to `and()` which will allow truthy and falsy values to respect handlebars version of truthy and falsy, but wont allow for the types to be correct
```
wagenet, line 42:
```suggestion
2) `{{and}}` compiles to `and()` which will allow truthy and falsy values to respect handlebars version of truthy and falsy, but wont allow for type guards and narrowing
```

0000-update-glimmer-template-truthiness-semantics.md
```diff
@@ -0,0 +1,106 @@
...
+# Update Glimmer Template Truthiness Semantics
+
+## Summary
+
+Provide a path for Glimmer templates to switch away from classic Handlebars truthiness to JavaScript’s truthiness. This will align our templating language with users' expectations and make teaching easier. It will also allow TypeScript features like type narrowing to work as users would expect it to with keywords like `{{and}}` or `{{or}}`.
+
+## Motivation
+
+Due to handlebars supporting a different truthiness semantics than javascript, typescript cant function like users would expect it to. 
+By aligning HTMLBars version of truthiness semantics with the wider javascript community, type checking will function as users expect.
```
wagenet, line 38:
```suggestion
By aligning Glimmer's version of truthiness semantics with the wider javascript community, type checking will function as users expect. 
```

wagenet (approved):
Looks great overall. Beyond my comments, I think there's a bit more minor editing work than can be done, mainly around consistent capitalization of JavaScript and TypeScript.

0000-update-glimmer-template-truthiness-semantics.md
```diff
@@ -0,0 +1,106 @@
...
+
+### {{gt}}
+Binary operation. Throws an error if not called with exactly two arguments. Equivalent of `>` in Javascript.
+
+### {{gte}}
+Binary operation. Throws an error if not called with exactly two arguments. Equivalent of `>=` in Javascript.
+
+### Migration Plan
+1. Ship comparison helpers `and`, `or`, `not` with existing equality and truthiness semantics (they currently match `if` and `unless` truthiness semantics)
+2. Ship types which match that and document why they work they way they do. Acknowledge that this wont work for TS users in they way they would expect and why.
```
jrjohnson, line 87:
```suggestion
2. Ship types which match that and document why they work the way they do. Acknowledge that this wont work for TS users in the way they would expect and why.
```

chriskrycho:
We talked a little bit about this at a high level at the Framework team meeting today (everyone is excited), and I'll review it early next week with the input from that!

0000-update-glimmer-template-truthiness-semantics.md
```diff
@@ -0,0 +1,106 @@
...
+3. Ship ability to opt into new behavior for `and`, `or`, `not`, `if` and `unless` (Following JS truthiness semantics)
```
jelhan, line 88: How would that opt-in look like? Would be great if you could do more in detail for the migration story in general.

jelhan (commented):
Thanks a lot for this RFC! I'm very much in favor of following native JavaScript more closely.

0000-update-glimmer-template-truthiness-semantics.md
```diff
@@ -0,0 +1,106 @@
...
+
+### {{or}}
+Takes at least two positional arguments. Raises an error if invoked with less than two arguments. It evaluates arguments left to right, returning the first one that is truthy (by javascript's definition of truthiness) or the right-most argument if all evaluate to falsy. This is equivalent to the {{or}} helper from ember-truth-helpers. Doing this would allow `{{or}}` to compile down to `||` and give us the correct type safety and narrowing that users would expect.
+
+### {{not}}
+Unary operator. Raises an error if invoked with more than one positional argument. If the given value evaluates to a truthy value (by javscript's definition of truthiness), the false is returned. If the given value evaluates to a falsy value (by javascript's definition of truthiness) then it returns true. This is equivalent to the {{not}} helper from ember-truth-helpers. Doing this would allow `{{not}}` to compile down to `!` and give us the correct type safety and narrowing that users would expect.
```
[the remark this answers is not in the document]
wagenet: Maybe still strange? I'm unsure. We definitely should support `{{not undefined}}` but I'm not sure about `{{not}}`.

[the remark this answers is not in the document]
jelhan: I would tend to not throw if invoked without arguments (`{{(not)}}`). But from typing perspective the argument should be required. And we might want to have a linting rule forbidding argumentless `{{(not)}}`.

I don't see much value in throwing. Implementation complexity should be the same. But we should use types and linting to push people towards `{{false}}` instead.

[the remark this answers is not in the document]
wagenet: I like @jelhan's solution here.

wagenet:
We should discuss what it takes for us to accept this at the next RFC Review.

ef4:
Two main bits of feedback:

1. We need a more detailed migration plan. Does this use optional-features? Is the opt-in per-package or global? If it's global, how are people realistically supposed to upgrade without breaking their addons? What are addon authors supposed to do (they can't control when their consuming apps flip a global opt-in)? If it's per-package, how does that actually work at runtime (is there something visible in the actual compiled templates to say "this one uses the new boolean semantics"?).

2. An important "Alternatives" topics is the idea of adding Javascript expression syntax to templates, something that we've discussed at length and people generally like as an idea, though no written design exists yet. Since that would also introduce new replacements for truth helpers, ones that would necessarily have javascript semantics, it seems like an area of overlap and it could make doing this RFC easier if we also did that one.

I have to say, I've been working in Glint quite a lot already and I've yet to hit a realistic example where this semantic gap actually came up. I would be curious to see people's examples. That doesn't mean I don't care about closing it, but I do think it's relevant to how we teach this and how we prioritize the migration.

chriskrycho:
The places it's likely to come up are exceptionally rare, but likely to be *very* annoying for people when they do happen, just insofar as they're a case where TS/Glint tells you they're safe and then you get a runtime error from it. Implicit in @ef4's comment is that Glint *does* let folks opt in (and our docs will better reflect that hopefully this week!).

That said, using a JS-expression-syntax would *also* solve this, and has the possibility a clear and straightforward opt-in at a template authoring level (e.g. a `<template lang='something-good-here'>` or some such along with an optional-feature to flip the default), and cleanly solves the problem from a Glint POV because *there is no problem* in that world: you just get TS semantics for free. Given the likely non-trivial work of migrating the truthiness semantics of existing Handlebars *vs.* switching to a JS expression syntax, my own personal preference would be extremely strongly toward the latter.

kategengler:
We have a mismatch between how JavaScript logic and handlebars logic which is now more apparent with Glint and <template> tag. This RFC has stalled, however, because of the complications of changing the existing meaning of logic in the templates. We've also found this mismatch is not a problem as often as would have been expected. We feel the easier way to address this is to find a way to allow JavaScript expressions in templates.

NullVoxPopuli:
moving to FCP-to-close, because for `{{ }}`, there isn't really a path to change truthiness incrementally across the addon/app boundaries.

ef4:
I would add that, empirically, we think the motivation for this isn't strong enough to hit people with a breaking change. 

In principle, we can agree handlebars truthiness is a quirk, but in practice I don't really see it causing pain. Even after glint adoption -- where typescript is interpreting your templates with JS truthiness -- I have never seen a real app bug caused by the gap.

PR #864, opened 2022-11-22
Deprecate Ember.A

jrjohnson:
[rendered](https://github.com/jrjohnson/ember-rfcs/blob/deprecate-ember-a/text/0864-deprecate-ember-a.md)

Remove `Ember.A()` as the functionality it provides is no longer needed in the post octane reactivity system and with the removal of array prototype extensions in #848.


text/0864-deprecate-ember-a.md
```diff
@@ -0,0 +1,94 @@
...
+prs:
+  accepted: https://github.com/emberjs/rfcs/pull/864
+project-link:
+---
+
+# Deprecate Ember.A()
+
+## Summary
+
+Remove `Ember.A()` as the functionality it provides is no longer needed in the post octane reactivity system and with the removal of array prototype extensions in #848.
```
wagenet, line 20:
```suggestion
Remove `Ember.A()` as the functionality it provides is no longer needed in the post Octane reactivity system and with the removal of array prototype extensions in #848.
```

text/0864-deprecate-ember-a.md
```diff
@@ -0,0 +1,94 @@
...
+
+## Motivation
+
+`Ember.A()` provides a way for addon authors or application owners who disable array prototype extensions to add reactivity to arrays (through `pushObject()`, `removeObject()`) and to access prototype extension convenience methods such as `filterBy()` on an any array-like object. Both of these paradigms have been deprecated in Ember and `Ember.A()` should follow suit as it can break things in unexpected ways and interferes with progress in [Ember](https://github.com/emberjs/ember.js/blob/4339725976299b24c69fb9dfbf13d18bf9917130/packages/@ember/-internals/utils/lib/ember-array.ts) and [Ember Data](https://github.com/emberjs/data/blob/47a71ca1538ba9e2d7dfa01bf048a2db897bdf5f/packages/store/addon/-private/record-arrays/identifier-array.ts#L381-L401) where special care must be taken to prevent or compensate for usage.
```
wagenet, line 24: I think it may just be sufficient to say that maintaining behavior that diverges from a true native Array unnecessarily adds to maintenance and potential confusion. It's actually possible to make Ember.Array work and not cause problems for Ember Data (I think), but it would just be nicer to have it gone!

text/0864-deprecate-ember-a.md
```diff
@@ -0,0 +1,94 @@
...
+
+Along with it's usefulness `Ember.A()` has a significant API gotcha in that it doesn't return a new instance of the array-like object it is passed, it modifies that object and returns is such that `A(array) === array` this can create the dreaded spooky-action-at-a-distance effect where suddenly an object which was `Array` can become an `Ember.NativeArray`. An example of this is using `{{includes "Zoey" this.mascots}}` from [ember-composable-helpers](https://github.com/DockYard/ember-composable-helpers#includes) will modify `this.mascots` and add `lastObject` to the API. Re-ordering the template code at a later date will result in a mysterious failure that can be extremely frustrating to understand.
```
wagenet, line 26: I think we may also be able to fix this, but it's definitely clear that Ember.A does weird stuff!

text/0864-deprecate-ember-a.md
```diff
@@ -0,0 +1,94 @@
...
+  <template>
+    {{#if (includes "Zoey" this.mascots)}}
+      Zoey Rules!
+    {{/if}}
+    Last Mascot: {{this.lastMascot}}
+  </template>
+}
+```
+
+Deprecating `Ember.A()` will provide a signal to the addon community to move away from both array prototype extensions and array reactivity which requires using the Emberism `pushObject`/`removeObject` to track updates.
```
wagenet, line 45: :100:

text/0864-deprecate-ember-a.md
```diff
@@ -0,0 +1,94 @@
...
+      Zoey Rules!
+    {{/if}}
+    Last Mascot: {{this.lastMascot}}
+  </template>
+}
+```
+
+Deprecating `Ember.A()` will provide a signal to the addon community to move away from both array prototype extensions and array reactivity which requires using the Emberism `pushObject`/`removeObject` to track updates. 
+
+Now that array prototype extensions have been deprecated it is quite possible that usage of `Ember.A()` will increase significantly as a defensive coding strategy to maintain access to the family of methods Ember adds to all arrays. This should be immediately discouraged as `Ember.A()` is not a long term viable solution and other better alternatives exist.
```
wagenet, line 47: This is a very good reason to deprecate Ember.A, ASAP.

text/0864-deprecate-ember-a.md
wagenet, line 20: I think we also may should deprecate `Ember.NativeArray` as part of this. I'd also like to deprecate `Ember.Array` altogether, but I'm not sure if that should be a separate thing or not. At the very least it would require the work we're doing here.

text/0864-deprecate-ember-a.md
```diff
@@ -0,0 +1,94 @@
...
+
+Deprecating `Ember.A()` will provide a clear signal to move away from this construct and rely on these more robust modern alternatives.
+
+### Transitional Feature Flags
+
+Transition away from `Ember.A()` should be aided by a pair of feature flags:
+
+#### Wrap Passed Value
+
+Instead of returning the passed value `Ember.A()` would return a native proxy to wrap the original object without modifying it. This would remove the most confusing part of the API and allow apps to opt in early and clear any unexpected issues before `Ember.A()` is fully removed.
```
wagenet, line 69: I've started doing some of this work in https://github.com/emberjs/ember.js/pull/20245. This change shouldn't actually require a flag since it's how Ember.A claims to work. However, we could add a short-lived flag to make it easier to validate the behavior in some real apps. Once we successfully validated we'd just remove the flag and have it always on. I don't think we'd have to publicize it in this case.

Once we successfully validate this, I'd like to add individual deprecations to the custom methods on the wrapper. That would help us to quickly identify cases where we're actually relying on the Ember.A behavior.

text/0864-deprecate-ember-a.md
```diff
@@ -0,0 +1,94 @@
...
+
+## Drawbacks
+
+`Ember.A()` has been a huge part of building Ember Addons for a long time and it is present in many popular addons. Removing this feature will cause a significant amount of churn in otherwise stable addons and require a significant community effort to replace.
+
+## Alternatives
+
+### Change Confusing API
+
+Instead of modifying the passed value `Ember.A()` could return a new object which contains a reference to the original value. This would avoid the `A(array) === array` issue and make it easier to detect if an object had been passed through `A()` and to work with the original value if necessary. This would still be a breaking change as many apps (often unknowingly) rely on the behavior of an object having been passed through `A()`.
```
wagenet, line 92: I think we're already planning on doing this. It wouldn't be considered breaking since the docs always said that Ember.A returns a new object. I think (I hope) it's rare for people to actually rely on the mutating behavior.

wagenet (commented):
This is a great start!

text/0864-deprecate-ember-a.md
jrjohnson: Credit to @runspired, I stole this from https://github.com/emberjs/rfcs/pull/848#issuecomment-1235883364

text/0864-deprecate-ember-a.md
```diff
@@ -0,0 +1,94 @@
...
+
+Along with it's usefulness `Ember.A()` has a significant API gotcha in that it doesn't return a new instance of the array-like object it is passed, it modifies that object and returns is such that `A(array) === array` this can create the dreaded spooky-action-at-a-distance effect where suddenly an object which was `Array` can become an `Ember.NativeArray`. An example of this is using `{{includes "Zoey" this.mascots}}` from [ember-composable-helpers](https://github.com/DockYard/ember-composable-helpers#includes) will modify `this.mascots` and add `lastObject` to the API. Re-ordering the template code at a later date will result in a mysterious failure that can be extremely frustrating to understand.
```
MrChocolatine, line 26:
```suggestion
Along with its usefulness `Ember.A()` has a significant API gotcha in that it doesn't return a new instance of the array-like object it is passed, it modifies that object and returns it such that `A(array) === array`, this can create the dreaded spooky-action-at-a-distance effect where suddenly an object which was `Array` can become an `Ember.NativeArray`. An example of this is using `{{includes "Zoey" this.mascots}}` from [ember-composable-helpers](https://github.com/DockYard/ember-composable-helpers#includes), which will modify `this.mascots` and add `lastObject` to the API. Re-ordering the template code at a later date will result in a mysterious failure that can be extremely frustrating to understand.
```

text/0864-deprecate-ember-a.md
```diff
@@ -0,0 +1,94 @@
...
+
+## Transition Path
+
+### Replacement APIs
+
+As discussed in #848 better alternatives exist for the APIs provided by `Ember.A()`:
+
+> For convenient methods like `filterBy`, `compact`, `sortBy` etc., the replacement functionalities already exist either through native array methods or utility libraries like [lodash](https://lodash.com), [Ramda](https://ramdajs.com), etc.
+
+>For mutation methods (like `pushObject`, `removeObject`) or observable properties (like `firstObject`, `lastObject`) participating in the Ember classic reactivity system, the replacement functionalities also already exist in the form of immutable update style with tracked properties like `@tracked someArray = []`, or through utilizing `TrackedArray` from `tracked-built-ins`.
```
MrChocolatine, line 57:
```suggestion
> For mutation methods (like `pushObject`, `removeObject`) or observable properties (like `firstObject`, `lastObject`) participating in the Ember classic reactivity system, the replacement functionalities also already exist in the form of immutable update style with tracked properties like `@tracked someArray = []`, or through utilizing `TrackedArray` from [tracked-built-ins](https://github.com/tracked-tools/tracked-built-ins).
```

text/0864-deprecate-ember-a.md
```diff
@@ -0,0 +1,94 @@
...
+
+`Ember.A()` is not covered in the guides, but it is present in [many](https://emberobserver.com/code-search?codeQuery=import%20%7B%20A%20%7D%20from%20%27%40ember%2Farray%27%3B) [many](https://emberobserver.com/code-search?codeQuery=Ember.A) addons. 
+
+Usage of convenience APIs can be easily replaced using the [no-array-prototype-extensions](https://github.com/ember-cli/eslint-plugin-ember/blob/4820f0fb286e40872a77d687618be56999f23704/docs/rules/no-array-prototype-extensions.md) codemod.
+
+For reactivity addon authors will need to rely forcing tracking as `this.mascots = [...this.mascots, 'Rock & Roll']` or through utilizing `TrackedArray` from [tracked-built-ins](https://github.com/tracked-tools/tracked-built-ins).
+
+## Drawbacks
+
+`Ember.A()` has been a huge part of building Ember Addons for a long time and it is present in many popular addons. Removing this feature will cause a significant amount of churn in otherwise stable addons and require a significant community effort to replace.
```
MrChocolatine, line 86:
```suggestion
`Ember.A()` has been a huge part of building Ember addons for a long time and it is present in many popular addons. Removing this feature will cause a significant amount of churn in otherwise stable addons and require a significant community effort to replace.
```

text/0864-deprecate-ember-a.md
jrjohnson: ✅

text/0864-deprecate-ember-a.md
jrjohnson: I'm not sure if I need to express this in the RFC or if this is just for information / reference. I agree with most of it, though I don't share your faith that no one is relying on this behavior. I think it is probably fair to say that the number of apps *intentionally* relying in this is low, however I think the side effect may have turned into a feature for many.

jrjohnson:
Thanks @wagenet and @MrChocolatine, I've incorporated your feedback.

text/0864-deprecate-ember-a.md
```diff
@@ -0,0 +1,94 @@
...
+
+# Deprecate Ember.A()
+
+## Summary
+
+Remove `Ember.A()` as the functionality it provides is no longer needed in the post Octane reactivity system and with the removal of array prototype extensions in #848.
+
+## Motivation
+
+`Ember.A()` provides a way for addon authors or application owners who disable array prototype extensions to add reactivity to arrays (through `pushObject()`, `removeObject()`) and to access prototype extension convenience methods such as `filterBy()` on an any array-like object. Both of these paradigms have been deprecated in Ember and `Ember.A()` should follow suit as maintaing this behavior of an array like object that differs from a true native array in important ways is a significant burdeon without clear benifits.
```
MrChocolatine, line 24:
```suggestion
`Ember.A()` provides a way for addon authors or application owners who disable array prototype extensions to add reactivity to arrays (through `pushObject()`, `removeObject()`) and to access prototype extension convenience methods such as `filterBy()` on an any array-like object. Both of these paradigms have been deprecated in Ember and `Ember.A()` should follow suit as maintaining this behavior of an array like object that differs from a true native array in important ways is a significant burden without clear benefits.
```

text/0864-deprecate-ember-a.md
```diff
@@ -0,0 +1,94 @@
...
+#### Noop Ember.A
+
+Currently if array prototype extensions is enabled `Ember.A()` is a `noop` and returns the passed value unmodified. It should be possible for application developers to turn off array prototype extensions and maintain this behavior of `Ember.A()`. This will make bugs easier to find and allow applications to more quickly prevent prototype pollution of the `Array` API before `Ember.A()` is fully removed.
+
+
+## How We Teach This
+
+`Ember.A()` is not covered in the guides, but it is present in [many](https://emberobserver.com/code-search?codeQuery=import%20%7B%20A%20%7D%20from%20%27%40ember%2Farray%27%3B) [many](https://emberobserver.com/code-search?codeQuery=Ember.A) addons. 
+
+Usage of convenience APIs can be easily replaced using the [no-array-prototype-extensions](https://github.com/ember-cli/eslint-plugin-ember/blob/4820f0fb286e40872a77d687618be56999f23704/docs/rules/no-array-prototype-extensions.md) codemod.
```
bmish, line 80:
```suggestion
Usage of convenience APIs can be linted against and largely autofixed/codemodded to native JavaScript by the [ember/no-array-prototype-extensions](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/no-array-prototype-extensions.md) lint rule.
```

jrjohnson: I like the text change, but don't like linking to `master` for docs in a file that will probably live on much longer than either the `master` branch or the path to this specific file. How about `https://github.com/ember-cli/eslint-plugin-ember/blob/v11.0.0/docs/rules/no-array-prototype-extensions.md`? to go to a specific version tag?

bmish: Got it. I just updated the URL to use the latest commit since there have been recent improvements to that rule doc.

jrjohnson:
@wagenet / @chriskrycho anything I should be doing to push this along or advocate for the change?

wagenet:
@jrjohnson Thanks for checking in on this. We talked about this in core today and I think this is going to have to be part of a more comprehensive plan for deprecating classic and @ember/component. I still think it's likely we'll be able to use this, but we just have to fit it in to the broader plan.

jrjohnson:
@wagenet I don't quite understand why this deprecation would be held up by that process. I'm not worried about my work here, take it or leave it, but I am worried about the proliferation of `Ember.A` that will be caused by the deprecation of array prototype extensions. I think accepting this RFC, even if the actual deprecation and removal happens as part of a larger integrated plan further down the line, is an important signal not to use this method to re-add prototypes to arrays in a way that is very difficult to un-tangle and will cause significant future pain.

wagenet:
My understanding is that there are still places where EmberArray still ends up being somewhat necessary, especially in the classic model. To have a deprecation without a full replacement is going to be pretty painful. I admit that I don't love our current situation, but my inclination is that it's actually better to risk more use of Ember.A in the short term if it means we can have a good plan to get rid of it properly in the long term.

jrjohnson:
That makes sense for `EmberArray`, but I don't understand why that prevents deprecating (or making internal only) `Ember.A()`. In every case I've seen the replacement for `Ember.A()` is either octane (by transforming to getters or using tracked state) or the solutions in the already accepted array prototypes deprecation. The full replacement already exists for this, or if it doesn't I hope someone can provide an example so I can add it to this RFC because that is super relevant.

ef4:
This was discussed this week in the [feature design (aka spec) meeting](https://github.com/emberjs/core-notes/blob/master/spec-meeting/2023/2023-10-12.md) where the suggestion was to keep Ember.A undeprecated during the transition away from the classic object model while deprecating all the actual non-standard array behaviors themselves. This could help ease the transition so that only code that's actively relying on the old system sees deprecations, and code that was treating Ember.A as a normal array won't need to be changed.

wagenet:
I think our problem here is that this didn't go far enough. We should just rip off the bandaid and deprecate _all_ of Ember Array at once.

PR #870, opened 2022-11-23
Always run new RFC checks so CI can pass

wagenet:


kategengler:
I think there are parts of the workflow that rely upon this not running for all directories. It needs some testing.

wagenet:
At least some of these items are required for CI to be green, so if we don't do this specific thing, we need to find another way to resolve it. See https://github.com/emberjs/rfcs/pull/867 for the issue.

kategengler:
I think it's not unreasonable to merge PRs that don't touch RFCs without green CI, until we work this out. None of the CI jobs do anything for changes not under /text and PRs that don't change RFCs are rare.

kategengler:
@sandstrom Newly opened RFCs will have a label of `S-Proposed`. There are some [quick links to filter the list in the README](https://github.com/emberjs/rfcs#quick-links-for-pull-requests-for-proposed-and-advancing-rfcs). 

But, I suggest that you should also be interested in RFCs moving to Ready for Release, Released, and Recommended: advancing to these stages means the RFC is progressing and there may be changes at these later stages. There are FCPs for both Ready for Release and Recommended. 

I suggest filtering to exclude PRs that are in `Draft`.

PR #883, opened 2022-12-10
Modifier manager timing capability

jelhan:
[Rendered](https://github.com/jelhan/rfcs/blob/modifier-manager-timing-capability/text/0000-modifier-manager-timing-capabilities.md)

> Element modifiers enable developers to customize a HTML element. Currently modifiers are always installed after the browser painted the element. This could lead to a flash of unstyled content and negatively impact performance due to double paint. This RFC proposes adding a second timing option allowing modifiers to run before the browser painted the element.

### DIFF
--- a/text/0000-modifier-manager-timing-capabilities.md
+++ b/text/0000-modifier-manager-timing-capabilities.md
@@ -28,15 +28,15 @@ project-link: Leave as is
 
 ## Summary
 
-Element modifiers enable developers to customize a HTML element. Sometimes it is important _when_ a customization is applied to an element for performance reasons.
-
-Element modifiers only support one timing so far. They always run _after_ the browser has finished pending work, including rendering the element on the screen. This RFC proposes adding a second timing option allowing modifiers to execute work _before_ browser rendered the element.
+Element modifiers enable developers to customize a HTML element. Currently modifiers are always installed _after_ the browser painted the screen. This could lead to a flash of unstyled content and negatively impact performance due to double paint. This RFC proposes adding a second timing option allowing modifiers to run _before_ the browser painted the screen.
 
 ## Motivation
 
-Delaying modifier install hook execution until browser finished the work, prevents modifiers from blocking rendering, which may cause stuttering animations and slow feedback to user interactions otherwise. This is an important performance optimization for modifiers, which do _not_ affect the layout of an element.
+Delaying modifier install hook execution until browser painted the screen, prevents modifiers from blocking rendering. Adding work into critical rendering path of the browser needs to be done carefully. If the browser is not able to execute all work within one rendering frame, the browser starts dropping frame, which could cause stuttering animations and slow feedback to user interactions. Delaying modifier Installation is an important performance optimization for most modifiers. But it may cause issues for modifiers, which either affect the layout of an element or add interactivity to it.
 
-But it may cause performance issues for modifiers, which affect the layout of an element. As the browser rendered the element _before_ the modifier is installed, modifying layout of an element in a modifier causes the element to be rendered twice.
+A modifier may be used to style an element. Painting the element _before_ the modifier is installed causes a flash of unstyled content. If this could be noticed by the user depends on many factors - including the performance of the device and other work executed in the same rendering frame.
+
+Additionally, it causes a double paint issue. The browser needs to paint the same element twice: Once before and once after the modifier is installed. This contradicts the performance considerations, which leads to the timing decision on the first place.
 
 Modifiers, which affect the layout of an element, are common. Example use cases, which have been published as open source, includes:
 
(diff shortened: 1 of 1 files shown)


chriskrycho:
Seems good. 👍🏼 I think implementing support for this in `ember-modifier` can readily be done in a backwards-compatible way, likely just by migrating to the new timing equivalent to the existing `installModifier` as the default and introducing a second pair of imports (function- and class-based) which use the new timing semantics. All existing modifiers will Just Work™ that way, but can opt into using the `installModifierOnLayout` hook:

```js
import { modifier, Modifier } from 'ember-modifier/name-TBD';
```

We need to think about how to name that, of course, but you can see how that approach works nicely from a design POV: it means the default is to use the existing semantics (which is what most users *should* use) while the more costly, but sometimes useful, semantics.

That said, I think it will be important to think about the relationship between the other possible timing semantics as well for the evolution of that library: before-insertion seems like it can just be another timing mode users will be able to opt into, but template-invocation-ordering is an orthogonal condition which seems likely to be cross-cutting, and it isn't obvious how to guarantee it given existing semantics about when and how we do re-renders—trying to enforce that guarantee isn't particularly natural in the context of Glimmer templates today.

jelhan:
>  I think it will be important to think about the relationship between the other possible timing semantics as well for the evolution of that library: before-insertion seems like it can just be another timing mode users will be able to opt into

Yes. That's the idea. The most challenging part for that is to not complicate server-side rendering support even more. For server-side rendering a `HTMLElement` is not available. We only have js-dom. To not paint us into a corner, I try to address server-side rendering in the next RFC before coming to before insert semantics. But it paves the way for it.

> template-invocation-ordering is an orthogonal condition which seems likely to be cross-cutting, and it isn't obvious how to guarantee it given existing semantics about when and how we do re-renders—trying to enforce that guarantee isn't particularly natural in the context of Glimmer templates today.

In the discussion I had with @wycats we tend to not support it. So far grabbing a reference to an element to be used later e.g. with `{{in-element}}` was named as only use case. We think that use case could be supported by other low-level primitives better. @wycats proposes adding a special syntax for grabbing a reference to an element to Glimmer templates itself in https://github.com/emberjs/rfcs/issues/652#issuecomment-1198551957.

jelhan:
> I think implementing support for this in ember-modifier can readily be done in a backwards-compatible way, likely just by migrating to the new timing equivalent to the existing installModifier as the default and introducing a second pair of imports (function- and class-based) which use the new timing semantics.

While this might be a good interim solution, I don't see it as a great high-level API for long-term. It limits modifiers to schedule work for either of the timings. But does not expose the power of scheduling some work for _on layout_ and some other work for _on idle_.

An event-driven API may fit better with added flexibility. It might look like this:

```js
modifier.on('layout', (element) => {
  // do some work, which needs to be done before browser paints the element 
});

modifier.on('idle", (element) => {
  // do some other work, which could not cause a flash of unstyled content issue
});

modifier.on('idle', () => {
  // could split up the work in logically independent chunks to structure the code
});

modifier.on('destroy', () => {
  // teardown logic
});
```

But deciding on a high-level API is out of scope for this RFC. Thanks to low-level primitives provided by modifier manager, we can experiment with those and gather experience before committing to a specific high-level API.

chriskrycho:
The text needs some clarification, then: as it stands the two capabilities/hooks appear to be *installation*-specific and that strongly implies that only one of them runs. (The text doesn’t explicitly say that but it’s what I inferred from the name!) If *both* hooks run, then I agree that we’d want to approach it differently; but we should update the hook names to be clear about that.

NullVoxPopuli:
> An event-driven API may fit better with added flexibility. It might look like this:

coming from Starbeam, and having worked with resources a little bit, I think I'd prefer a unified API for _all things_™

for example (and with the caveat that this is very hand-wavy (except for the resource example, which comes from the [docs](https://www.starbeamjs.com/api/core/resource.html))):

resources:
```js
import { Cell, Resource } from "@starbeam/universal";

const Stopwatch = Resource(({ on }) => {
  const start = Cell(new Date());
 
  const interval = setInterval(() => {
    start.set(new Date());
  }, 1000);
 
  on.cleanup(() => {
    clearInterval(interval);
  });
 
  return start;
})
```

and an eventual modifier would be a resource that takes the element as an argument:
```js
import { Modifier } from 'somewhere';

const Style = Modifier(({ element, on }) =>{
  on.layout(() => { /* set the style attribute */ });
  on.idle(() => { /* wiggle using the WebAnimation API */ });
  on.cleanup(() => { /* remove changes */ });
});
```

While this looks _about the same_, if you squint, I think it provides better ergonomics for editors and language servers.

jelhan:
> The text needs some clarification, then: as it stands the two capabilities/hooks appear to be _installation_-specific and that strongly implies that only one of them runs. (The text doesn’t explicitly say that but it’s what I inferred from the name!)

I agree. The current naming is misleading.

Having more than one hook, which is executed on first render does not fit well in the existing lifecycle model for modifiers in general. It currently has the following:

1. Create (one time)
2. Install (one time)
3. Update (zero to many times)
4. Destroy (one time)

Maybe we need to have something like this instead:

1. Create (one time)
2. Setup on layout (zero to one time)
3. Setup on idle (zero to one time)
4. Update (zero to many times)
5. Destroy (one time)

Maybe we want to provide a reference to the update lifecycle hook in that case as well. It would not fit well in the mental model if a modifier _must_ implement one of the setup hooks just to grab a reference to the element.

jelhan:
I renamed the hooks to `setupModifierOnLayout` and `setupModifierOnIdle`. Additionally I added a clarification that the same modifier manager can implement both hooks. @chriskrycho, I hope this addresses the confusion you raised.

NullVoxPopuli:
how has exploration been going? This is highly needed! <3

wagenet:
@NullVoxPopuli I think part of our challenge here is making sure that we're actually going to do the work. Is this something you'd be interested in helping with.

NullVoxPopuli:
yes, It is -- partially to help with graphics and chart rendering performance, and partially to help out with ergonomics and performance in how ember-headless-table is wired

machty:
Another nice use case for this: merging tailwind classes for component library use cases: https://twitter.com/buschtoens/status/1658483457000103938

PR #889, opened 2022-12-15
Deprecate @ember/error

wagenet:
[Rendered](https://github.com/emberjs/rfcs/blob/deprecate-ember-error/text/0889-deprecate-ember-error.md)


chriskrycho (approved):
DO IT.

When we say "super easy codemod," by the way, we mean:

```sh
npx jscodeshift -t https://astexplorer.net/#/gist/5f88840599701bae9452cea55e86641e/b9641f4cb132946c885b6138284b52a82d5cffe4/raw/transform.js <glob for files>
```

SergeAstapov:
@wagenet tbh not a concern. maybe even not need to be explicitly stated. Ember Data just would need to account for that I assume

wagenet:
Per the core team meeting today, we're moving this to FCP to merge.

PR #894, opened 2023-01-07
Add eslint-plugin-import to ember-cli blueprint

bmish:
This RFC proposes adding [eslint-plugin-import](https://github.com/import-js/eslint-plugin-import) to the ember-cli blueprint.

TODO:

1. I talked to @wycats about this 2 years ago, and he said this linter was very important to add, but mentioned there were some issues that would need to be solved with it. I'm hoping he or others can elaborate on additional issues that we need to solve. 
2. In the RFC, I describe some issues with the [import/no-unresolved](https://github.com/import-js/eslint-plugin-import/blob/HEAD/docs/rules/no-unresolved.md) rule, and consider whether we can use [eslint-import-resolver-ember](https://github.com/gabrielcsapo/eslint-import-resolver-ember) to solve them. Based on some initial testing, I haven't gotten things working yet, and will need feedback/advice on this. CC: @gabrielcsapo


text/0894-eslint-plugin-import.md
```diff
@@ -0,0 +1,174 @@
...
+release-date: Leave as is
+release-versions: Leave as is
+teams: Include only the [team(s)](README.md#relevant-teams) for which this RFC applies
+prs:
+  accepted: Fill this in with the URL for the Proposal RFC PR
+project-link: Leave as is
+suite: Leave as is
+-->
+
+# Add eslint-plugin-import to ember-cli blueprint
```
NullVoxPopuli, line 28: I have a config for no-extraneous-imports, which we need, but i don't know if plugin-import supports it.

Why plugin-import over plugin-n?

bmish: What do you mean exactly? The ember-cli blueprint already includes [eslint-plugin-node](https://github.com/mysticatea/eslint-plugin-node). [eslint-plugin-n](https://github.com/eslint-community/eslint-plugin-n) is intended to be a temporary fork of eslint-plugin-node, until the eslint-community team is able to take over eslint-plugin-node.

NullVoxPopuli: eslint-plugin-node is unmaintained and doesn't support ESM or exports.

eslint-plugin-n has rules (such as no-extraneous-import), that work well within browser code -- because every imported module _must_ be declared in the package.json -- this rule is very good

bmish: Replacing eslint-plugin-node with eslint-plugin-n, at least temporarily, is worth considering given that the owner of eslint-plugin-node has disappeared. But eslint-plugin-import has many other distinct rules beyond these Node plugins. Are you suggesting eslint-plugin-import isn't worthwhile over the existing linting?

NullVoxPopuli: Looking over this list:
```
plugin:import/recommended
```
https://github.com/import-js/eslint-plugin-import -- config here: https://github.com/import-js/eslint-plugin-import/blob/main/config/recommended.js#L8

this does actually still look useful, though the set of recommended rules seems actually kinda small.

we should do a separate RFC for eslint-plugin-n (dropping eslint-plugin-node, and adding additional rules for browser code)

Disregard my original comment!

This RFC still brings up a ton of other valid stuff that we should def implement.
Incremental improvements is key! (no need to have perfection out of the gate -- though, I think it's weird that eslint-plugin-import still has no analogous rule to no-extraneous-import :upside_down_face: )

bertdeblock: FYI https://github.com/ember-cli/ember-cli/pull/10060

NullVoxPopuli: I'm outdated!

text/0894-eslint-plugin-import.md
```diff
@@ -0,0 +1,174 @@
...
+    },
+  ],
+}
+```
+
+Note that these configurations enable only a handful of key rules from the plugin's 40+ rules. We may optionally choose to individually enable additional rules that we determine to be particularly valuable for Ember apps, either in this RFC, or in the future. But regardless, getting the plugin with its basic linting setup into place right now is worthwhile on its own, as this will also make it much easier for users to enable rules on their own too.
+
+### Resolver
+
+There are a few issues to fix with the [import/no-unresolved](https://github.com/import-js/eslint-plugin-import/blob/HEAD/docs/rules/no-unresolved.md) rule.
```
NullVoxPopuli, line 90: Did some testing on this plugin today in my own https://github.com/NullVoxPopuli/eslint-configs

`import/no-unresolved` does not understand `package.json#exports` -- which is a bit of a deal breaker for modern packages.

I tested with `vitest`:

```
eslint-configs/vitest.config.ts
  1:30  error  Unable to resolve path to module 'vitest/config'  import/no-unresolved
# ...
❯ cat node_modules/vitest/package.json | jq '.exports';
{
  ... ✂️...
  "./config": {
    "types": "./config.d.ts",
    "require": "./dist/config.cjs",
    "import": "./dist/config.js"
  }
  
```

So far, all V2 addons also use `package.json#exports`

ef4:
I think it's probably not possible to achieve reliable detection of unresolvable imports in traditional apps and addons. People are allowed to emit arbitrary AMD modules from a whole bunch of possible places in the build, across all your addons and their addons, and those are all imports that work at runtime but are totally opaque to a static linter.

In embroider we are stabilizing around having one standardized module resolver that would potentially be able to power a lint rule, but even there when we're looking at a v1 addon or v1 app (and at this point all apps are v1 apps because we haven't done a v2 app format RFC), we *never* emit resolver failures, because when we run out of possible static places to look we need to assume there might be a runtime answer, so we emit a runtime lookup.

V2 addons *do* follow strict enough rules that an unresolvable import lint rule can work reliably there. The [list of traditional ember-source provided imports](https://github.com/embroider-build/embroider/blob/01575a4d3a86b8b329c75024aa77b2539a6dad5d/packages/shared-internals/src/ember-standard-modules.ts#L14) is known and hard-coded, and everything else is required to be statically resolvable.

It's possible we'd want to be similarly strict in a v2 app format, as long as we can explain what the upgrade path is such that apps can upgrade even if not every addon has done so.

NullVoxPopuli:
I'm good with disabling `import/no-unresolved` :100: the lint that has helped me out the most is `import/no-extraneous-dependencies` (or `n/no-extraneous-imports`) -- it helps folks properly manage their package.json, which can get very messed up in npm and yarn projects.

I'd even be a fan of splitting up the RFC _per lint rule_, if needed, because the plugin clearly has some good rules that we can use today and can help guide folks towards better configured projects -- all without waiting on v2 stuff

text/0894-eslint-plugin-import.md
bmish: Thanks for testing that. I have updated the RFC to disable the `import/no-unresolved` lint rule.

bmish:
@ef4 thanks for the context. I have updated the RFC to disable the `import/no-unresolved` lint rule to get around the current infeasibility of reliably resolving imports. We can always come back to it later if it becomes possible, but this unblocks getting the rest of eslint-plugin-import into place.

wagenet:
We've moved this to Exploring.

ef4:
I still don't think we can safely implement this until after something like https://github.com/emberjs/rfcs/pull/938 because there's too much stuff that is valid and working but not detectable as a real dependency, even beyond just the standard ember-provided ones.

